PR #3, opened 2019-03-08
Handle overridden Transport channels.

ditac:
Description of changes:

The security plugin assumes that transport channel is always of a particular type. The performance analyzer plugin wraps the transportChannel into a different object and this breaks the transport plugin. This change retrieves the underlying transportChannel via reflection.


src/main/java/com/amazon/opendistroforelasticsearch/security/transport/OpenDistroSecurityRequestHandler.java
```diff
@@ -126,12 +127,20 @@ protected void messageReceivedDecorate(final T request, final TransportRequestHa
               && !transportChannel.getChannelType().equals("PerformanceAnalyzerTransportChannelType")) {
                throw new RuntimeException("Unknown channel type "+transportChannel.getChannelType());
            }
+           String channelType = transportChannel.getChannelType();
 
-           getThreadContext().putTransient(ConfigConstants.OPENDISTRO_SECURITY_CHANNEL_TYPE, transportChannel.getChannelType());
+           if(transportChannel.getChannelType().equals("PerformanceAnalyzerTransportChannelType")) {
```
hardik-k-shah: we can reuse channelType string here.

if(channelType.equals("PerformanceAnalyzerTransportChannelType")) {

PR #5, opened 2019-03-11
Clarify OpenID Connect (OIDC) support

praseodym:
This plugin supports the modern OpenID Connect (OIDC) protocol, not the older XML-based OpenID 2.0 protocol. This PR updates the README to clarify this.

Ideally, references in code would also be updated to refer to “OIDC” instead of just “OpenID”, but that’s likely to be a breaking change.


hardik-k-shah:
Thanks @praseodym for taking care of this changes.

PR #23, opened 2019-04-07
Handle attributes when impersonating user

### LINKED ISSUE #22: Impersonated user has no attributes
After impersonating a user using the `opendistro_security_impersonate_as` header, all of my document-level-security rules that relied on user attributes exploded.

Additionally, the attributes were not visible on a query to `_opendistro/_security/authinfo` - user name, roles, etc... all looked right, but `custom_attribute_names` was empty.

This occurred for me on both the 0.7.0 and 0.7.1 docker images.

I'll submit a pull request with a unit test that exposes this behavior, plus a patch that fixes it, shortly.

turettn:
After impersonating a user using the `opendistro_security_impersonate_as` header, all of my document-level-security rules that relied on user attributes exploded.

Additionally, the attributes were not visible on a query to `_opendistro/_security/authinfo` - user name, roles, etc... all looked right, but `custom_attribute_names` was empty.

This pull request includes a unit test that checks for this behavior, along with a fix so that user attributes are handled the same way roles are on impersonation.

Closes #22.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/auth/internal/InternalAuthenticationBackend.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/auth/internal/InternalAuthenticationBackend.java
@@ -35,6 +35,7 @@ import java.nio.CharBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 
 import org.bouncycastle.crypto.generators.OpenBSDBCrypt;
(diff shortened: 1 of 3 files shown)


turettn:
@hardik-k-shah @shivdoshi @prashar If this needs changes, please let me know.

FWIW, it recently got imported into floragunncom/search-guard@cddc94d5086a00146d2c100beb20aaa7ffbe4021.

hardik-k-shah:
Sorry for delay in this. 
Yes, we are building and running test on this at our end.  We will merge it or provide feedback after it. 
Thanks.

PR #84, opened 2019-07-10
Predefined roles for alerting.

lucaswin-amzn:
This change defines 3 new default read only roles to easily allow integration with the [opendistro alerting plugin](https://github.com/opendistro-for-elasticsearch/alerting)

The three roles are as follows:
- **alerting_view_alerts**: Allows a user to view alerts. No monitors or destinations, also does not allow users of this role to acknowledge alerts.
- **alerting_crud_alerts**: Allows a user to view and acknowledge alerts. User cannot view monitors or destinations.
- **alerting_full_access**: Provides full functionality of the alerting plugin.

Testing done manually using a single user for each role.

### DIFF
--- a/securityconfig/roles.yml
+++ b/securityconfig/roles.yml
@@ -205,3 +205,6 @@ alerting_full_access:
     '?opendistro-alerting*':
       '*':
         - CRUD
+    '?opendistro-alerting-alert*':
+      '*':
+        - CRUD
--- a/securityconfig/roles.yml
+++ b/securityconfig/roles.yml
@@ -181,3 +181,27 @@ own_index:
     '${user_name}':
       '*':
         - INDICES_ALL
+
+# Allows users to view alerts
+alerting_view_alerts:
+  readonly: true
+  indices:
+    '?opendistro-alerting-alert*':
+      '*':
+        - READ
+
+# Allows users to view and acknowledge alerts
+alerting_crud_alerts:
+  readonly: true
+  indices:
+    '?opendistro-alerting-alert*':
+      '*':
+        - CRUD
+
+# Allows users to use all alerting functionality
+alerting_full_access:
+  readonly: true
+  indices:
+    '?opendistro-alerting*':
+      '*':
+        - CRUD


lucaswin-amzn:
> As you are explicitly marking them readonly, why not adding them into static config?
> 
> I would also like to know reason behind adding them here and not creating these roles through Alerting plugin itself.
> These roles are specially for alerting feature and if user doesn't want to install alerting plugin, these are extra roles staying there in system

1. These roles are read only due to avoid modification to them and expose more than just alerting features since they are meant to provide only access to alerting.
2. Not adding them to static config as this is for opendistro-0.7.0.1, will add them to the static config which was created in the current release of 1.0.0
3. The overhead for these roles would be minimal in the security plugin, while we could create them in the alerting plugin it would be much easier to ship security with these installed.

Most downloads of opendistro contain all plugins, such as the docker image to my understanding. Hence shipping out of the box with them here would make sense in my opinion.

securityconfig/roles.yml
```diff
@@ -181,3 +181,27 @@ own_index:
...
+# Allows users to view and acknowledge alerts
+alerting_crud_alerts:
+  readonly: true
+  indices:
+    '?opendistro-alerting-alert*':
+      '*':
+        - CRUD
+
+# Allows users to use all alerting functionality
+alerting_full_access:
```
elfisher: Does this span both alerts and monitors/destinations?

securityconfig/roles.yml
```diff
@@ -181,3 +181,27 @@ own_index:
...
+# Allows users to view alerts
+alerting_view_alerts:
+  readonly: true
+  indices:
+    '?opendistro-alerting-alert*':
+      '*':
+        - READ
+
+# Allows users to view and acknowledge alerts
+alerting_crud_alerts:
```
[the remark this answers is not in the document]
elfisher: This is only for acknowledging alerts and reading them.

securityconfig/roles.yml
elfisher: This does span alerts, monitors, and destinations.

securityconfig/roles.yml
```diff
@@ -181,3 +181,27 @@ own_index:
...
+  indices:
+    '?opendistro-alerting-alert*':
+      '*':
+        - CRUD
+
+# Allows users to use all alerting functionality
+alerting_full_access:
+  readonly: true
+  indices:
+    '?opendistro-alerting*':
```
hardik-k-shah: Can we explicitly mentioned entire index name (use * only for timestamp)?

lucaswin-amzn:
Closing this in favor of a PR against master which will be cherry picked to other releases.

PR #96, opened 2019-07-18
Adding Kibana readonly role

allenyin96:
Ran unit tests jenkins and all passed.

### DIFF
--- a/securityconfig/roles.yml
+++ b/securityconfig/roles.yml
@@ -6,7 +6,7 @@ _meta:
 kibana_read_only:
   reserved: true
 
-rest_api_access:
+security_rest_api_access:
   reserved: true
  
 # Allows users to view alerts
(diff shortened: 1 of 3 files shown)


securityconfig/roles.yml
```diff
@@ -2,6 +2,13 @@ _meta:
   type: "roles"
   config_version: 2
 
+# Restrict users so they can only view visualization and dashboard on kibana
+kibana_read_only:
+  reserved: true
+
+rest_api_access:
```
elfisher: we use `security_rest_api_access` in the elasticsearch.yml. Lets add the following comment. "The security REST API access role is used to assign specific users access to change the security settings through the REST API."

allenyin96: Updated the PR to reflect this change.

PR #151, opened 2019-10-09
Add new header for requests going to Protected Indices

lucaswin-amzn:
Add a new header for requests coming to protected indices.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/OpenDistroSecurityIndexSearcherWrapper.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/OpenDistroSecurityIndexSearcherWrapper.java
@@ -146,8 +146,7 @@ public class OpenDistroSecurityIndexSearcherWrapper extends IndexSearcherWrapper
                 return true;
             }
         }
-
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/OpenDistroSecurityIndexSearcherWrapper.java
```diff
@@ -138,13 +138,17 @@ protected final boolean isBlockedIndexRequest() {
...
+
+        if (user != null && caller != null) {
+            final Set<String> securityRoles = evaluator.mapRoles(user, caller);
+            if (WildcardMatcher.matchAny(allowedRoles, securityRoles)) {
+                return true;
+            }
         }
-        final Set<String> securityRoles = evaluator.mapRoles(user, caller);
-        if (WildcardMatcher.matchAny(allowedRoles, securityRoles)) {
+        if ("true".equals(HeaderHelper.getSafeFromHeader(threadContext, ConfigConstants.OPENDISTRO_PROTECTED_INDICES_CONF_REQUEST_HEADER))) {
```
dinusX: We should probably do this only if "user == null", which means that only requests that don't have user context (originating from internal cron jobs) will have access.

dinusX:
After an internal discussion, we decided to not push this change.
Closing.

PR #174, opened 2019-12-03
extended proxy authenticator to pass additional attributes via header

jcantrill:
This PR introduces and extended proxy authenticator to allow proxies to pass additional user attributes that can be consumed when using document level security.  Configured like:
```
    http:
      xff:
        enabled: true
        remoteIpHeader: 'x-forwarded-for'
        trustedProxies: '.*'
        internalProxies: '.*'
    authc:
      authentication_domain_proxy:
        enabled: true
        order: 0
        http_authenticator:
          challenge: false
          type: extended-proxy
          config:
            user_header: 'x-proxy-remote-user'
            attr_header_prefix: 'x-proxy-ext-'
        authentication_backend:
          type: noop
```
This allows a trusted proxy to provide request headers:
```
x-proxy-ext-namespace: foo
x-proxy-ext-namespace: bar
x-proxy-ext-bu: makemoney
```
that are made available as user attributes:
```
attr.proxy.namespace: foo,bar
attr.proxy.bu: makemoney
```
This is a change to master comparable to https://github.com/opendistro-for-elasticsearch/security/pull/85

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigModel.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigModel.java
@@ -23,6 +23,7 @@ import com.amazon.opendistroforelasticsearch.security.auth.limiting.UserNameBase
 import com.amazon.opendistroforelasticsearch.security.http.HTTPBasicAuthenticator;
 import com.amazon.opendistroforelasticsearch.security.http.HTTPClientCertAuthenticator;
 import com.amazon.opendistroforelasticsearch.security.http.HTTPProxyAuthenticator;
+import com.amazon.opendistroforelasticsearch.security.http.proxy.HTTPExtendedProxyAuthenticator;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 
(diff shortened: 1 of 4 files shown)


jcantrill:
cc @allenyin96  @richm

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigModel.java
```diff
@@ -77,6 +78,7 @@ public DynamicConfigModel() {
 
         authImplMap.put("basic_h", HTTPBasicAuthenticator.class.getName());
         authImplMap.put("proxy_h", HTTPProxyAuthenticator.class.getName());
+        authImplMap.put("extednd-proxy_h", HTTPExtendedProxyAuthenticator.class.getName());
```
allenyin96: Seems like a typo. Should be "extended-proxy_h"

jimmyjones2:
@jcantrill Should the role header be accepted too, like the original proxy authentication?

jcantrill:
> @jcantrill Is the role header be accepted too, like the original proxy authentication?

Since this is an extension of the original class I would expect the previous headers to be accepted as well.

PR #179, opened 2019-12-11
CI workflow with Github Actions

sujithvm:
Integrating CI workflow with Github actions.
This workflow will be triggered for any pull request or push to master branch.
Workflow will run only after it is merged to master. 

Tasks which will be implemented in future PRs
- Checkstyle enforcement

Tasks not implemented yet due to approvals
- Codecov code coverage with Jacoco 

Testing:
https://github.com/sujithvm/security/commit/45ce53e2122ab37862a5a1f0e3719360d9795a47/checks?check_suite_id=354132934

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,52 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
+      - '*'
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: opendistroforelasticsearch/security-maven:v1
+
+    steps:
+
+    - name: Checkout security-parent
+      uses: actions/checkout@v1
+      with:
+        repository: opendistro-for-elasticsearch/security-parent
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+  pull_request:
+    branches:
+      - '*'
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
allenyin96: Do we know what's installed on this machine? Or is it better to run the build in a docker container if that's possible? So that we have more control over the build environment.

sujithvm: It is the base image of ubuntu 18.04 LTS. I believe it is possible to run in a docker container. As this is a standalone Java project and has no system dependencies (is there any system dependency?) it should be fine running without a container.

sujithvm: I am on the same page to use docker image for building (which uses open jdk version). I have updated the PR.

PR #193, opened 2020-01-07
Use BASE_DIR in demo install script

epol:
In `install_demo_configuration.sh` was hardcoded the default directory
`/usr/share/elasticsearch`, this won't work if elasticsearch is
installed on a custom location (or if it used from the tar.gz for
testing).

Using the `BASE_DIR` variable we can get more flexibility from the script.

### DIFF
--- a/tools/install_demo_configuration.sh
+++ b/tools/install_demo_configuration.sh
@@ -113,17 +113,12 @@ SUDO_CMD=""
 ES_INSTALL_TYPE=".tar.gz"
 
 #Check if its a rpm/deb install
-if [ -f /usr/share/elasticsearch/bin/elasticsearch ]; then
+if [ "/usr/share/elasticsearch" -ef "$BASE_DIR" ]; then
     ES_CONF_FILE="/usr/share/elasticsearch/config/elasticsearch.yml"
 
     if [ ! -f "$ES_CONF_FILE" ]; then
         ES_CONF_FILE="/etc/elasticsearch/elasticsearch.yml"
     fi
-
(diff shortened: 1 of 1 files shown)


tools/install_demo_configuration.sh
```diff
@@ -113,17 +113,17 @@ SUDO_CMD=""
...
 
     if [ ! -f "$ES_CONF_FILE" ]; then
         ES_CONF_FILE="/etc/elasticsearch/elasticsearch.yml"
     fi
 
-    ES_BIN_DIR="/usr/share/elasticsearch/bin"
-    ES_PLUGINS_DIR="/usr/share/elasticsearch/plugins"
-    ES_MODULES_DIR="/usr/share/elasticsearch/modules"
-    ES_LIB_PATH="/usr/share/elasticsearch/lib"
+    ES_BIN_DIR="$BASE_DIR/bin"
```
sujithvm: @epol Thank you for your contribution.
These values are already set above so I was wondering if it is possible to delete these lines instead. Thanks!

epol: Thank you for your review. I'm going to write a better edit since this one may have been breaking the rpm/deb logic.
I'll update this pull request as soon as i'm done.

epol:
After the review I noticed that my commit was breaking the logic that checks if elasticsearch has been installed from packages (deb/rpm).

I rewrote the patch simply by changing this check:

- before we were checking just if /usr/share/elasticsearch/bin/elasticsearch esisted
- now we check if we are running the script from that directory

This should take care of both cases also in machines where elasticsearch has been installed both as package and tar.

PR #197, opened 2020-01-10
Getting ready for opendistro-1.4 release

allenyin96:
Changes:

1. Minor changes to reflect ES API changes.
2. Removing support of OpenSSL for Java 12+

This means: 
1. For java versions before 12, the behavior stays the same.
2. For java 12+, we use JDK SSL, and if the user enables OpenSSL with Java 12+, we give a message saying that we don't support OpenSsl with Java12+ and continue to use JDK SSL

Tested with Java 1.8, 11 & 12

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@
     <groupId>com.amazon.opendistroforelasticsearch</groupId>
     <artifactId>opendistro_security</artifactId>
     <packaging>jar</packaging>
-    <version>1.3.0.0</version>
+    <version>1.4.0.0</version>
     <name>Open Distro Security for Elasticsearch</name>
     <description>Open Distro For Elasticsearch Security</description>
     <inceptionYear>2015</inceptionYear>
@@ -67,7 +67,7 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <maven.compiler.release>8</maven.compiler.release>
 
-        <elasticsearch.version>7.3.2</elasticsearch.version>
+        <elasticsearch.version>7.4.2</elasticsearch.version>
 
         <!-- deps -->
         <netty-native.version>2.0.25.Final</netty-native.version>
@@ -97,7 +97,7 @@
         <url>https://github.com/opendistro-for-elasticsearch/security</url>
         <connection>scm:git:git@github.com:opendistro-for-elasticsearch/security.git</connection>
         <developerConnection>scm:git:git@github.com:opendistro-for-elasticsearch/security.git</developerConnection>
-        <tag>1.3.0.0</tag>
+        <tag>1.4.0.0</tag>
     </scm>
 
     <issueManagement>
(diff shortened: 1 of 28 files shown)


pom.xml
```diff
@@ -34,7 +34,7 @@
     <groupId>com.amazon.opendistroforelasticsearch</groupId>
     <artifactId>opendistro_security</artifactId>
     <packaging>jar</packaging>
-    <version>1.3.0.0</version>
+    <version>1.4.0.0</version>
```
sujithvm: Since we have released the artifacts for 1.4.0.0, we should be updating this version to `1.5.0.0-SNAPSHOT` which reflects the development for supporting elasticsearch 7.5.0. What do you think ?

allenyin96: I agree that we should change the version to 1.5.0.0-SNAPSHOT in the master branch, but I think we should do that after we merge this PR into master and cut a 1.4 branch from it.

sujithvm: Okay sounds good. I was confused seeing the artifacts already released. Approved the PR.

src/main/java/com/amazon/opendistroforelasticsearch/security/tools/OpenDistroSecurityAdmin.java
```diff
@@ -448,6 +449,7 @@ public static int execute(final String[] args) throws Exception {
                 .put(SSLConfigConstants.OPENDISTRO_SECURITY_SSL_TRANSPORT_ENFORCE_HOSTNAME_VERIFICATION_RESOLVE_HOST_NAME, !nrhn)
                 .put(SSLConfigConstants.OPENDISTRO_SECURITY_SSL_TRANSPORT_ENABLED, true)
                 .put(SSLConfigConstants.OPENDISTRO_SECURITY_SSL_TRANSPORT_ENABLE_OPENSSL_IF_AVAILABLE, useOpenSSLIfAvailable)
```
hardik-k-shah: we don't need this line anymore.

src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/OpenDistroSecuritySSLPlugin.java
```diff
@@ -82,6 +82,7 @@
 //For ES5 this class has only effect when SSL only plugin is installed
 public class OpenDistroSecuritySSLPlugin extends Plugin implements ActionPlugin, NetworkPlugin {
 
+    public static final boolean OPENSSL_SUPPORTED = false;
```
hardik-k-shah: we should check this runtime. 
If java12 is available, this flag should be false otherwise it should be true. 

We should also check how we are building and packaging with? is it Java 8, 11 or 12?

allenyin96: Made the changes. With github action, we are building the package with Java 11

src/main/java/com/amazon/opendistroforelasticsearch/security/tools/OpenDistroSecurityAdmin.java
allenyin96: Deleted.

PR #198, opened 2020-01-10
[Optimization] Implement faster version of implies type perm

DmitryOlshansky:
Code for checking permission match was very contrived and hidden a big performance problem we have spotted in our installation for log analytics.

In short getResolvedIndexPattern method is very costly as it goes to ElasticSearch Metadata store takes the tree map of _all_ of indices/aliases and elaborately filters things out, then it proceeds to extract aliases and more. 

In the original code this method is called for every matching index times every index pattern. In our system we have long series of rotated indices so it might be why we have triggered this problem.

Another optimization is to avoid creating temporary hash sets with complicated logic and translate the whole code to a combination of a few nested matchAll + matchAny algorithms.

Does it all matter ? Yes, in our case impliesTypePerm took about 16-17% of total CPU time, indexing of documents with Lucene (to compare) would use ~29% with the rest being GC and other thread pools. Now it's down to around 1.6%. Next obvious place to tackle is Wildcard matcher but with an order of magnitude lesser gain.

P.S. I also removed some of commented code for dealing with types no sure if it's needed - just for reference to keep branches for older versions of ES in sync? If so I'll restore it.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
@@ -47,7 +47,6 @@ import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.common.transport.TransportAddress;
 
 import com.amazon.opendistroforelasticsearch.security.resolver.IndexResolverReplacer.Resolved;
-import com.amazon.opendistroforelasticsearch.security.securityconf.ConfigModelV6.SecurityRole;
 import com.amazon.opendistroforelasticsearch.security.securityconf.impl.SecurityDynamicConfiguration;
 import com.amazon.opendistroforelasticsearch.security.securityconf.impl.v7.ActionGroupsV7;
 import com.amazon.opendistroforelasticsearch.security.securityconf.impl.v7.RoleMappingsV7;
@@ -936,26 +935,26 @@ public class ConfigModelV7 extends ConfigModel {
         }));
     }
 
-    static final class ResolvedIndexPattern{
+    static final class IndexPatternsAndPermissions {
         public String[] pattern;
         public Set<String> perms;
-        public ResolvedIndexPattern(String[] pattern, Set<String> perms) {
+        public IndexPatternsAndPermissions(String[] pattern, Set<String> perms) {
             this.pattern = pattern;
             this.perms = perms;
         }
     }
 
-    private static boolean impliesTypePerm(Set<IndexPattern> ipatterns, Resolved resolved, User user, String[] actions,
+    private static boolean impliesTypePerm(Set<IndexPattern> ipatterns, Resolved resolved, User user, String[] requestedActions,
                                            IndexNameExpressionResolver resolver, ClusterService cs) {
-        Set<String> resolvedIndices = resolved.getAllIndices();
-        List<ResolvedIndexPattern> resolvedIndexPatterns = ipatterns
+        Set<String> resolvedRequestedIndices = resolved.getAllIndices();
+        List<IndexPatternsAndPermissions> indexPatternsAndPermissions = ipatterns
                 .stream()
-                .map(p -> new ResolvedIndexPattern(p.getResolvedIndexPattern(user, resolver, cs), p.perms))
+                .map(p -> new IndexPatternsAndPermissions(p.getResolvedIndexPattern(user, resolver, cs), p.perms))
                 .collect(Collectors.toList());
-        return resolvedIndices
+        return resolvedRequestedIndices
                 .stream()
                 .allMatch(index ->
-                        Arrays.stream(actions).allMatch(action -> resolvedIndexPatterns
+                        Arrays.stream(requestedActions).allMatch(action -> indexPatternsAndPermissions
                                 .stream()
                                 .anyMatch(rip -> WildcardMatcher.matchAny(rip.pattern, index) && WildcardMatcher.matchAny(rip.perms, action))
                         )
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
@@ -936,38 +936,30 @@ public class ConfigModelV7 extends ConfigModel {
         }));
     }
 
-    private static boolean impliesTypePerm(Set<IndexPattern> ipatterns, Resolved resolved, User user, String[] actions,
-            IndexNameExpressionResolver resolver, ClusterService cs) {
-        Set<String> matchingIndex = new HashSet<>(resolved.getAllIndices());
-
-        for (String in : resolved.getAllIndices()) {
-            //find index patterns who are matching
-            Set<String> matchingActions = new HashSet<>(Arrays.asList(actions));
-            //Set<String> matchingTypes = new HashSet<>(resolved.getTypes(-));
-            for (IndexPattern p : ipatterns) {
-                if (WildcardMatcher.matchAny(p.getResolvedIndexPattern(user, resolver, cs), in)) {
-                    //per resolved index per pattern
-                    //for (String t : resolved.getTypes(-)) {
-                        //for (TypePerm tp : p.typePerms) {
-                            //if (WildcardMatcher.match(tp.typePattern, t)) {
-                                //matchingTypes.remove(t);
-                                for (String a : Arrays.asList(actions)) {
-                                    if (WildcardMatcher.matchAny(p.perms, a)) {
-                                        matchingActions.remove(a);
-                                    }
-                                }
-                            //}
-                        //}
-                    //}
-                }
-            }
-
-            if (matchingActions.isEmpty() /*&& matchingTypes.isEmpty()*/) {
-                matchingIndex.remove(in);
-            }
+    static final class ResolvedIndexPattern{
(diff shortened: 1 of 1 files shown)


DmitryOlshansky:
I tried to minimize the scale of changes. Hope you folks can take a look and comment on it.

hardik-k-shah:
Thanks for your contribution. We are looking into this. We will do perf testing internally with your change and come back to you.  By the time, can you share your performance testing setup and results?

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
```diff
@@ -936,38 +936,30 @@ private static String toQuotedCommaSeparatedString(final Set<String> roles) {
...
+        public ResolvedIndexPattern(String[] pattern, Set<String> perms) {
+            this.pattern = pattern;
+            this.perms = perms;
         }
+    }
 
-        return matchingIndex.isEmpty();
+    private static boolean impliesTypePerm(Set<IndexPattern> ipatterns, Resolved resolved, User user, String[] actions,
+                                           IndexNameExpressionResolver resolver, ClusterService cs) {
+        Set<String> resolvedIndices = resolved.getAllIndices();
```
hardik-k-shah: can we rename it to resolvedRequestedIndices? (as these are indices requested from user.) It will help code more readable down into method where we added lot more complexity.

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
```diff
@@ -936,38 +936,30 @@ private static String toQuotedCommaSeparatedString(final Set<String> roles) {
...
+        public String[] pattern;
+        public Set<String> perms;
+        public ResolvedIndexPattern(String[] pattern, Set<String> perms) {
+            this.pattern = pattern;
+            this.perms = perms;
         }
+    }
 
-        return matchingIndex.isEmpty();
+    private static boolean impliesTypePerm(Set<IndexPattern> ipatterns, Resolved resolved, User user, String[] actions,
```
hardik-k-shah: Can we rename `actions` to `requestedActions`? (As these are actions requested by user and not coming from roles definition) It will help understanding code better.

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
```diff
@@ -936,38 +936,30 @@ private static String toQuotedCommaSeparatedString(final Set<String> roles) {
...
-                            //}
-                        //}
-                    //}
-                }
-            }
-
-            if (matchingActions.isEmpty() /*&& matchingTypes.isEmpty()*/) {
-                matchingIndex.remove(in);
-            }
+    static final class ResolvedIndexPattern{
```
hardik-k-shah: Can we rename it `ResolvedIndexPattern` to `indexPatternsAndPermissions` ? (as it has patterns and permissions)

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
```diff
@@ -936,38 +936,30 @@ private static String toQuotedCommaSeparatedString(final Set<String> roles) {
...
+            this.pattern = pattern;
+            this.perms = perms;
         }
+    }
 
-        return matchingIndex.isEmpty();
+    private static boolean impliesTypePerm(Set<IndexPattern> ipatterns, Resolved resolved, User user, String[] actions,
+                                           IndexNameExpressionResolver resolver, ClusterService cs) {
+        Set<String> resolvedIndices = resolved.getAllIndices();
+        List<ResolvedIndexPattern> resolvedIndexPatterns = ipatterns
```
hardik-k-shah: Can we rename `resolvedIndexPatterns` to `indexPatternsAndPermsFromRole`? (as these are resolved index patterns and permissions from role)

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
DmitryOlshansky: starting with lowercase? Or did you mean `IndexPatternsAndPermissions` ?

DmitryOlshansky:
>  By the time, can you share your performance testing setup and results?

It's rather involved but I'll try to post the exact reproducer soon. 
Just to give you a hint on the setup:
hundreds of logs.something-000001 ... logs.something-000XXX

each chain has an alias logs.something

now that alias points to the whole chain with is_write_index: false, and the last one has is_write_index: true

I just tried to reproduce this by having 300 logstash-001 .. logstash-300 indices each having is_write_index: false alias of logstash-alias. Then I started indexing a few gigs of http logs into logstash-301 that has logstash-alias is_write_index: true.

Now this seems to regress performance extremely with over 70% of CPU spent in opendistro security code paths, BUT it's not in impliesTypePerm. Something is different this time around. Still digging and trying to get closer to our setup. Still I believe this is something to consider, by bet is that resolving *both*  role patterns *and* alias destination patterns leads to N^2 matching problem in this case.

Meanwhile I'll upload all of renamings.

DmitryOlshansky:
@hardik-k-shah Not related to this particular PR but what do you guys think about redoing pattern matching from loops over strings/patterns to some sort of DFA (deterministic finite-state automation)?

If it's not on someone's todo list already I'd give it a shot myself.

DmitryOlshansky:
@hardik-k-shah  I guess that's not what you asked for but here goes:
https://github.com/opendistro-for-elasticsearch/security/pull/258

this time a bigger patch, and I'd better to check if we are affected by this second problem (we should be as I modeled our environment as described above).

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
```diff
@@ -936,38 +935,30 @@ private static String toQuotedCommaSeparatedString(final Set<String> roles) {
...
+        List<IndexPatternsAndPermissions> indexPatternsAndPermissions = ipatterns
+                .stream()
+                .map(p -> new IndexPatternsAndPermissions(p.getResolvedIndexPattern(user, resolver, cs), p.perms))
+                .collect(Collectors.toList());
+        return resolvedRequestedIndices
+                .stream()
+                .allMatch(index ->
+                        Arrays.stream(requestedActions).allMatch(action -> indexPatternsAndPermissions
+                                .stream()
+                                .anyMatch(rip -> WildcardMatcher.matchAny(rip.pattern, index) && WildcardMatcher.matchAny(rip.perms, action))
```
vrozov: rename `rip` to `indexPatternsAndPermissions` (or less preferable `ipap`).

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
```diff
@@ -936,38 +935,30 @@ private static String toQuotedCommaSeparatedString(final Set<String> roles) {
...
-                        //}
-                    //}
-                }
-            }
-
-            if (matchingActions.isEmpty() /*&& matchingTypes.isEmpty()*/) {
-                matchingIndex.remove(in);
-            }
+    static final class IndexPatternsAndPermissions {
+        public String[] pattern;
```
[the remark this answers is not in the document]
DmitryOlshansky: Done

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
DmitryOlshansky: used ipap as the scope is 3 lines and I believe in short names for short scopes.

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
```diff
@@ -936,38 +935,30 @@ private static String toQuotedCommaSeparatedString(final Set<String> roles) {
...
+            this.pattern = pattern;
+            this.perms = perms;
         }
+    }
 
-        return matchingIndex.isEmpty();
+    private static boolean impliesTypePerm(Set<IndexPattern> ipatterns, Resolved resolved, User user, String[] requestedActions,
+                                           IndexNameExpressionResolver resolver, ClusterService cs) {
+        Set<String> resolvedRequestedIndices = resolved.getAllIndices();
+        List<IndexPatternsAndPermissions> indexPatternsAndPermissions = ipatterns
```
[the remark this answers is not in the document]
DmitryOlshansky: Done

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
```diff
@@ -936,38 +935,30 @@ private static String toQuotedCommaSeparatedString(final Set<String> roles) {
...
-                            //}
-                        //}
-                    //}
-                }
-            }
-
-            if (matchingActions.isEmpty() /*&& matchingTypes.isEmpty()*/) {
-                matchingIndex.remove(in);
-            }
+    static final class IndexPatternsAndPermissions {
```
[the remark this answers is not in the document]
DmitryOlshansky: Also covered

PR #209, opened 2020-01-22
Set cluster.routing.allocation.disk.threshold_enabled default value

### LINKED ISSUE #208: `cluster.routing.allocation.disk.threshold_enabled` default set to false
For some reason, the default value of this variable is set to `false` while default value for Elasticsearch is `true`.
That variable controls the balance mechanism of shards between nodes of a cluster.

vincenzodnp:
This PR fix #208 by setting `cluster.routing.allocation.disk.threshold_enabled` default value (_true_), according to Elasticsearch default value

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/security/test/helper/cluster/ClusterHelper.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/security/test/helper/cluster/ClusterHelper.java
@@ -358,7 +358,7 @@ public final class ClusterHelper {
                 .put("transport.tcp.port", tcpPort)
                 .put("http.port", httpPort)
                 //.put("http.enabled", true)
-                .put("cluster.routing.allocation.disk.threshold_enabled", false)
+                .put("cluster.routing.allocation.disk.threshold_enabled", true)
                 .put("http.cors.enabled", true)
                 .put("path.home", ".");
     }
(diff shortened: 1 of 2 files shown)


vincenzodnp:
Any update on this? Thanks

vincenzodnp:
Any change to include this in the next opendistro releases?

vrozov:
How this settings affects security? If it is `true` by default, is it necessary to set it explicitly? I would prefer that security plugin does not enforce any value.

dinusX:
Can you add more details on what exactly this change is fixing ? What issues are you seeing when it set to false?

So far the change is impacting only tests and the script used for generating demo certificates and config.

vincenzodnp:
Hello @dinusX and @vrozov 
This PR will fix an issue we had at @amazeeio.
The default value in "official" Elasticsearch is `true` and it allows to balance the number of shards across all nodes of the cluster.
If you set it as `false` (as we got from the official docker image of Opendistro), the shards are not spread across all the nodes but randomly. This means we got nodes with a higher number of shards compared to others and we need to re-allocate shards to other nodes to avoid disk space issues.

vrozov:
IMO, the setting needs to be removed from both places and security plugin should not enforce it to be `true` or `false`.

debu99:
samething on node.max_local_storage_nodes

Tarasovych:
I wish this PR will be merged just to keep OpenDistro default settings in sync with https://www.elastic.co/guide/en/elasticsearch/reference/7.x/modules-cluster.html#disk-based-shard-allocation

vrozov:
The PR is not ready to be merged till all comments are resolved.

sujithvm:
Fixed by #1067 . Thank you for the contribution

PR #224, opened 2020-02-04
Removed OpenSSL for 7.4

debjanibnrj:
* Tested with ODFE7.4
* Sanity tested for security plugin


src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/OpenDistroSecuritySSLPlugin.java
```diff
@@ -82,7 +82,8 @@
 //For ES5 this class has only effect when SSL only plugin is installed
 public class OpenDistroSecuritySSLPlugin extends Plugin implements ActionPlugin, NetworkPlugin {
 
-    public static final boolean OPENSSL_SUPPORTED = PlatformDependent.javaVersion() < 12;
+    // Not supporting OPENSSL for ES7.4+
+    public static final boolean OPENSSL_SUPPORTED = false;
```
sujithvm: If we are not supporting open ssl we need to remove the assoicated open ssl code too from the project?

debjanibnrj: This is a temporary workaround since the current openssl library does not seem to be working for ES7.4+.

PR #234, opened 2020-02-17
Fix to use inner channel when channel is not direct or transport type

sujithvm:


src/main/java/com/amazon/opendistroforelasticsearch/security/transport/OpenDistroSecurityRequestHandler.java
```diff
@@ -126,16 +126,9 @@ protected void messageReceivedDecorate(final T request, final TransportRequestHa
...
-                try {
-                    Method getInnerChannel = wrappedChannelCls.getMethod("getInnerChannel", null);
-                    TransportChannel innerChannel = (TransportChannel)(getInnerChannel.invoke(transportChannel));
-                    channelType = innerChannel.getChannelType();
-                } catch (NoSuchMethodException ex) {
-                    throw new RuntimeException("Unknown channel type " + channelType + " does not implement getInnerChannel method.");
-                }
+            if (!channelType.equals("direct") && !channelType.equals("transport")) {
+                TransportChannel innerChannel = getInnerChannel(transportChannel);
+                channelType = innerChannel.getChannelType();
```
sujithvm: I am not sure if we should channel to innerChannel here.

src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/transport/OpenDistroSecuritySSLRequestHandler.java
```diff
@@ -81,6 +82,11 @@ public final void messageReceived(T request, TransportChannel channel, Task task
             channel.sendResponse(exception);
             throw exception;
         }
+
+        String channelType = channel.getChannelType();
+        if (!channelType.equals("direct") && !channelType.equals("transport")) {
+            channel = getInnerChannel(channel);
```
sujithvm: Also wondering if this check should come after the following `if (!"transport".equals(channel.getChannelType()))` check

debjanibnrj: Please include comments about why this check needs to be there.

src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/transport/OpenDistroSecuritySSLRequestHandler.java
```diff
@@ -81,6 +82,11 @@ public final void messageReceived(T request, TransportChannel channel, Task task
             channel.sendResponse(exception);
             throw exception;
         }
+
+        String channelType = channel.getChannelType();
+        if (!channelType.equals("direct") && !channelType.equals("transport")) {
```
debjanibnrj: Wasn't this originally - (!"netty".equals(channel.getChannelType()) && !"direct".equals(channel.getChannelType())) based on https://github.com/opendistro-for-elasticsearch/security-ssl/pull/2/files.

Why are we checking for !channelType.equals("transport")?

sujithvm: Maybe its changed from elastic side https://github.com/elastic/elasticsearch/pull/40074

sujithvm: netty is renamed to transport

src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/transport/OpenDistroSecuritySSLRequestHandler.java
sujithvm: Sure. Will add comments. Can you verify the logic ?

debjanibnrj: Logic seems fine for PA. As an extra measure we can verify that the returned channel from getInnerChannel is of channelType transport.

src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/transport/OpenDistroSecuritySSLRequestHandler.java
debjanibnrj: Thanks for clarification

src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/transport/OpenDistroSecuritySSLRequestHandler.java
sujithvm: throw a runtime error if its not basic and transport ?

debjanibnrj: I think so, esp since we're expecting the Wrapper classes `getInnerChannel` to return a transport channel.

sujithvm: We are getting a transport channel but the issue is with the channelType, custom implementations may pass

src/main/java/com/amazon/opendistroforelasticsearch/security/transport/OpenDistroSecurityRequestHandler.java
debjanibnrj: Afaik its fine to change channel to innerChannel here. 

In our previous PR's we override the channelType value instead of the channel value but the idea behind the two seem to be the same - https://github.com/opendistro-for-elasticsearch/security-ssl/pull/2/files; https://github.com/opendistro-for-elasticsearch/security/pull/3/files.

PR #253, opened 2020-02-28
Auditlog refactoring 

sujithvm:
- Created AuditConfig to represent audit configuration settings. This helps in injecting new settings during hot-reloading.

- Created ComplianceConfig to represent audit compliance settings. This helps in injecting new settings during hot-reloading.

- Created AsyncStoragePoolConfig to represent storage pool settings. This helps in injecting new settings during hot-reloading.

- Implemented AuditMessageBuilder pattern to generate audit message keeping AuditMessage clean for sink providers. 

- Refactored AuditCategory to represent all categories and create EnumSet to represent collection of categories. Previously strings were passed around.  Added tests.

- Refactored filter methods into a separate utility class AuditFilter to separate out validator util functions. 

- Merged AbstractAuditLog into AuditLogImpl (adheres to AuditLog interface). 

- Factored out AuditMessage request resolver and compliance resolver into separate classes. (This will be injected into AuditLogImpl class).


src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
```diff
@@ -0,0 +1,38 @@
...
+
+        categories
+                .stream()
+                .filter(Objects::nonNull)
+                .map(category -> category.toUpperCase())
+                .forEach(category -> {
+                    try {
+                        set.add(AuditCategory.valueOf(category));
+                    } catch (IllegalArgumentException ex) {
+                    }
```
debjanibnrj: throw/log exception?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditLogImpl.java
```diff
@@ -38,179 +53,468 @@
...
+        }
+    }
+
+    @Override
+    public void logFailedLogin(final String effectiveUser,
+                               final boolean securityAdmin,
+                               final String initiatingUser,
+                               final TransportRequest request,
+                               final Task task) {
+        if (!enabled) return;
```
debjanibnrj: since save already checks for enabled is there any advantage of having the enabled check for each audit log?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditLogImpl.java
```diff
@@ -38,179 +53,468 @@
...
+
+        List<AuditMessage> msgs = requestResolver.resolve(AuditCategory.AUTHENTICATED, getOrigin(), action, null,
+                effectiveUser, securityAdmin, initiatingUser, getRemoteAddress(), request, task, auditConfig, null);
+
+        msgs.forEach(this::save);
+    }
+
+    @Override
+    public void logSucceededLogin(String effectiveUser, boolean securityAdmin, String initiatingUser, RestRequest request) {
+        if (!enabled) return;
```
debjanibnrj: any advantage of having this check for each method? Save function is already checking for enabled

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessageBuilder.java
```diff
@@ -0,0 +1,350 @@
...
+    public static final String REMOTE_ADDRESS = "audit_request_remote_address";
+    public static final String REST_REQUEST_PATH = "audit_rest_request_path";
+    //public static final String REST_REQUEST_BODY = "audit_rest_request_body";
+    public static final String REST_REQUEST_PARAMS = "audit_rest_request_params";
+    public static final String REST_REQUEST_HEADERS = "audit_rest_request_headers";
+    public static final String TRANSPORT_REQUEST_TYPE = "audit_transport_request_type";
+    public static final String TRANSPORT_ACTION = "audit_transport_action";
+    public static final String TRANSPORT_REQUEST_HEADERS = "audit_transport_headers";
+    public static final String ID = "audit_trace_doc_id";
+    //public static final String TYPES = "audit_trace_doc_types";
```
debjanibnrj: any reason why these are commented out?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessageBuilder.java
```diff
@@ -0,0 +1,350 @@
...
+import java.nio.file.Path;
+import java.nio.file.attribute.FileTime;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+public class AuditMessageBuilder {
```
debjanibnrj: If you're planning on using the Builder pattern to create AuditMessage objects, you need to implement this as a public static nested class of AuditMessage. See [link1](https://howtodoinjava.com/design-patterns/creational/builder-pattern-in-java/) [link2](https://dzone.com/articles/the-builder-pattern-for-class-with-many-constructo)

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/AuditConfig.java
```diff
@@ -0,0 +1,168 @@
...
+                ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_WRITE_IGNORE_USERS,
+                defaultIgnoredUsers);
+
+        List<String> ignoreAuditRequests = getConfigList(
+                settings,
+                ConfigConstants.OPENDISTRO_SECURITY_AUDIT_IGNORE_REQUESTS,
+                Collections.emptyList());
+
+
+        AuditConfig config = new AuditConfig(
```
debjanibnrj: minor: you can return here

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/AuditConfig.java
```diff
@@ -0,0 +1,168 @@
...
+                disabledRestCategories,
+                disabledTransportCategories);
+        return config;
+    }
+
+    private static List<String> getConfigList(final Settings settings,
+                                              final String key,
+                                              final List<String> defaultList) {
+        List<String> list = new ArrayList<>(settings.getAsList(key, defaultList));
+        if (list.size() == 1 && "NONE".equals(list.get(0))) {
```
debjanibnrj: why are we allowing users to pass NONE? how is this different from an empty list?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
```diff
@@ -0,0 +1,38 @@
...
+    COMPLIANCE_EXTERNAL_CONFIG,
+    COMPLIANCE_INTERNAL_CONFIG_READ,
+    COMPLIANCE_INTERNAL_CONFIG_WRITE;
+
+    public static EnumSet<AuditCategory> parse(final List<String> categories) {
+        EnumSet<AuditCategory> set = EnumSet.noneOf(AuditCategory.class);
+        if (categories == null)
+            return set;
+
+        categories
```
debjanibnrj: Can also be written as
```
 return categories.stream()
                .filter(Objects::nonNull)
                .map(category -> AuditCategory.valueOf(category.toUpperCase()))
                .collect(Collectors.toCollection(() -> EnumSet.noneOf(AudityCategory.class)));
```

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/sink/SinkProviderTLSTest.java
```diff
@@ -129,15 +128,14 @@ private SSLContext createSSLContext() throws Exception {
...
-		Assert.assertTrue(in, in.contains(AuditMessage.UTC_TIMESTAMP));
-		Assert.assertTrue(in, in.contains(Category.FAILED_LOGIN.name()));
+		Assert.assertTrue(in, in.contains(AuditMessageBuilder.FORMAT_VERSION));
+		Assert.assertTrue(in, in.contains(AuditMessageBuilder.CATEGORY));
+		Assert.assertTrue(in, in.contains(AuditMessageBuilder.REMOTE_ADDRESS));
+		Assert.assertTrue(in, in.contains(AuditMessageBuilder.ORIGIN));
+		Assert.assertTrue(in, in.contains(AuditMessageBuilder.REQUEST_LAYER));
+		Assert.assertTrue(in, in.contains(AuditMessageBuilder.TRANSPORT_REQUEST_TYPE));
+		Assert.assertTrue(in, in.contains(AuditMessageBuilder.UTC_TIMESTAMP));
+		Assert.assertTrue(in, in.contains(AuditCategory.FAILED_LOGIN.name()));
```
debjanibnrj: why .name() for this one?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
sujithvm: This will throw IllegalArgumentException exception if invalid category string is passed in the List.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
sujithvm: Will log error. Should not throw exception in my opinion

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessageBuilder.java
sujithvm: This would make the interface to be 
`AuditMessage.AuditMessageBuilder()....` 

Why particular reason why this is needed ?

debjanibnrj: Thats how the Builder pattern works. We just use Builders to construct other objects. We want an object of type AuditMessage, not of type AuditMessageBuilder. Whenever AuditMessage.Builder....build() is a common way to implement and use Builders. 

https://stackoverflow.com/questions/5007355/builder-pattern-in-effective-java

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditLogImpl.java
sujithvm: This would help in early exit instead of processing/creating audit message and then failing.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessageBuilder.java
sujithvm: Alright! Thanks for the links :)

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/sink/SinkProviderTLSTest.java
sujithvm: This is a category type (part of enum). Its converting to string type and then asserting

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/AuditConfig.java
sujithvm: Good question. I retained what existed already.
YAML 
1) list: []
2) list: ["NONE"]  
3) list: "NONE" 

should all behave the same for now.
Only advantage I see is how users can explicitly say "NONE" instead of a list.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
debjanibnrj: Won't a try/catch block resolve that?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/AuditConfig.java
debjanibnrj: Thanks for the explanation. We should determine if we want to support NONE as a valid option in the future. According to me an empty list should suffice. Will let @hardik-k-shah  and @dinusX chime in with their opinions.

sujithvm: https://opendistro.github.io/for-elasticsearch-docs/docs/security-audit-logs/#exclude-categories 
We probably need the above if we allow users to set NONE. But yeah if we force the users to use [] then we wont need it.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
sujithvm: It would quit the looping for remaining categories instead of just ignoring the current one

debjanibnrj: Cool thanks for following up.

sujithvm:
Closing this PR and opening a new one. CI seems broken

PR #256, opened 2020-02-28
Allow SuperAdmin to update/delete/add read only config such as action…

palashhedau:
Allow SuperAdmin to update/delete/add read only config such as action_groups, roles, roles_mapping, internal_users

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
@@ -115,9 +115,7 @@ public class InternalUsersApiAction extends PatchableResourceApiAction {
         }
 
         // check if resource is writeable
-        Boolean readOnly = configurationSettings.v2().getAsBoolean(username + "." + ConfigConstants.CONFIGKEY_READONLY,
-                Boolean.FALSE);
-        if (readOnly) {
+        if (!isReadOnlyAndAccessible(configurationSettings.v2(), username)) {
             forbidden(channel, "Resource '" + username + "' is read-only.");
             return;
         }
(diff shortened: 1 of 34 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -115,9 +115,7 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
         }
 
         // check if resource is writeable
-        Boolean readOnly = configurationSettings.v2().getAsBoolean(username + "." + ConfigConstants.CONFIGKEY_READONLY,
-                Boolean.FALSE);
-        if (readOnly) {
+        if (!isReadOnlyAndAccessible(configurationSettings.v2(), username)) {
```
sujithvm: Can you verify if  resources are stored by `username + "." + ConfigConstants.CONFIGKEY_READONLY ` and this works as expected

palashhedau: Yes, looks like this mechanism in place 
example -  v2={"admin.attributes.attribute1":"value1","admin.attributes.attribute2":"value2"

palashhedau: Its just the way of lookup in the settings, my code doesn't interfere with the logic change for the lookup

sujithvm: Correct me if i m wrong. It looks like in v6, read only internal user resource is suffixed with `.ConfigConstants.CONFIGKEY_READONLY`. Dont we need to check for this also in `isReadOnlyAndAccessible` function?

palashhedau: There is a function isReadOnly which has this check and isReadOnlyAndAccessible is calling this isReadOnly function.

The one thing I noticed is all other api were making use of this function but not internalusers[code duplication]. now internalusers is also making use of isReadOnly via isReadOnlyAndAccessible

sujithvm: Alright. I see isReadOnly has implemented the same

PR #257, opened 2020-02-29
Allow SuperAdmin to update/delete/add read only config such as action…

palashhedau:
Functionality :- Allow SuperAdmin to update/delete/add read only config such as action_groups, roles, roles_mapping, internal_users

Fixed readonly check for AccountAPI


src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -115,9 +115,7 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
         }
 
         // check if resource is writeable
-        Boolean readOnly = configurationSettings.v2().getAsBoolean(username + "." + ConfigConstants.CONFIGKEY_READONLY,
```
sujithvm: Just want to check if this is verified. Are internal users used store in any different format in 6.x.x ?

PR #258, opened 2020-03-01
[Optimization] Faster resolve index pattern - fixes bulk requests against aliases to long index series with 1 write index

DmitryOlshansky:
So continuing with my PR https://github.com/opendistro-for-elasticsearch/security/pull/198

I decided to reproduce performance problem and have hit a much bigger one.

Test setup is 
300 indices of 10 documents each, logstash-100 ... logstash-400 , all have alias logstash-alias pointing to them (is_write_index: false)

1 target index logstash-1 (alias logstash-alias, is_write_index: true)

Profiler https://github.com/jvm-profiling-tools/async-profiler is attached as follows:
```
./profiler.sh -d 30 -e cpu -f ~/new.svg -i 900000 `ps aux | grep elastic | awk '{ print $2; exit }'`
```

The dataset is one of http_logs files from ES Rally benchmark tool, see https://github.com/elastic/rally-tracks/tree/master/http_logs

(I used tar.gz distro and removed perf analyzer plugin at some point b/c I thought it was adding overhead, not really but it's missing from flame graphs for this reason)

The following series of flames is how the big chunk of CPU time on the *left*  slowly melts from ~65 to 5%.

original

latest

latest-2
latest-3

latest-4

latest-5

The last 2 flamegraphs include my patch for impliesTypePerm because at that point impliesTypePerm finally become a significant portion of CPU time.

Full interactive flamegraphs are in zip file.

[flame-graphs.zip](https://github.com/opendistro-for-elasticsearch/security/files/4272522/flame-graphs.zip)

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
@@ -319,7 +319,8 @@ public final class IndexResolverReplacer implements DCFListener {
         final AtomicBoolean isIndicesRequest = new AtomicBoolean();
 
         getOrReplaceAllIndices(request, new IndicesProvider() {
-            // resolve cache helps us big time on bulk requests
+            // resolve cache - keep track of already completed resolve index requests
+            // help us big time for bulk requests
             final Set<IndexResolveKey> resolvedBefore = new HashSet<>();
 
             @Override
@@ -327,7 +328,7 @@ public final class IndexResolverReplacer implements DCFListener {
                 final IndicesOptions indicesOptions = indicesOptionsFrom(localRequest);
                 final boolean isSearchOrFieldCapabilities = localRequest instanceof FieldCapabilitiesRequest || localRequest instanceof SearchRequest;
                 final IndexResolveKey key = new IndexResolveKey(indicesOptions, isSearchOrFieldCapabilities, original);
-                // skip the whole thing if we have seen this exact resolveIndexPatterns result
+                // skip the whole thing if we have seen this exact resolveIndexPatterns request
                 if (!resolvedBefore.contains(key)) {
                     final Resolved iResolved = resolveIndexPatterns(key.opts, key.isSearchOrFieldCapabilities, key.original);
                     resolvedBefore.add(key);
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
@@ -226,18 +226,16 @@ public final class IndexResolverReplacer implements DCFListener {
         else {
 
             ClusterState state = clusterService.state();
-
-            final SortedMap<String, AliasOrIndex> lookup = state.metaData().getAliasAndIndexLookup();
-            final List<String> aliases = lookup.entrySet().stream().filter(e -> e.getValue().isAlias()).map(e -> e.getKey())
-                    .collect(Collectors.toList());
-
             Set<String> dateResolvedLocalRequestedPatterns = localRequestedPatterns
                             .stream()
                             .map(resolver::resolveDateMathExpression)
                             .collect(Collectors.toSet());
             //fill matchingAliases
-            matchingAliases = aliases
+            final SortedMap<String, AliasOrIndex> lookup = state.metaData().getAliasAndIndexLookup();
+            matchingAliases = lookup.entrySet()
                     .stream()
+                    .filter(e -> e.getValue().isAlias())
+                    .map(Map.Entry::getKey)
                     .filter(alias -> WildcardMatcher.matchAny(dateResolvedLocalRequestedPatterns, alias))
                     .collect(Collectors.toSet());
 
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
@@ -32,7 +32,6 @@ package com.amazon.opendistroforelasticsearch.security.resolver;
 
 import java.io.IOException;
 import java.io.Serializable;
-import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -49,6 +48,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
 
+import com.google.common.base.Objects;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.elasticsearch.action.ActionRequest;
@@ -161,7 +161,7 @@ public final class IndexResolverReplacer implements DCFListener {
         return false;
     }
 
-    private Resolved resolveIndexPatterns(final IndicesOptions indicesOptions, final Object request, final String... requestedPatterns0) {
+    private Resolved resolveIndexPatterns(final IndicesOptions indicesOptions, final boolean isSearchOrFieldsCapabilities, final String... requestedPatterns0) {
 
         if(log.isTraceEnabled()) {
             log.trace("resolve requestedPatterns: "+ Arrays.toString(requestedPatterns0));
@@ -179,8 +179,7 @@ public final class IndexResolverReplacer implements DCFListener {
 
         final RemoteClusterService remoteClusterService = OpenDistroSecurityPlugin.GuiceHolder.getRemoteClusterService();
 
-        if(remoteClusterService.isCrossClusterSearchEnabled() && request != null
-                && (request instanceof FieldCapabilitiesRequest || request instanceof SearchRequest)) {
+        if(remoteClusterService.isCrossClusterSearchEnabled() && isSearchOrFieldsCapabilities) {
             remoteIndices = new HashSet<>();
             final Map<String, OriginalIndices> remoteClusterIndices = OpenDistroSecurityPlugin.GuiceHolder.getRemoteClusterService()
                     .groupIndices(indicesOptions, requestedPatterns0, idx -> resolver.hasIndexOrAlias(idx, clusterService.state()));
@@ -209,7 +208,6 @@ public final class IndexResolverReplacer implements DCFListener {
         }
 
         final Set<String> matchingAliases;
-        final Set<String> matchingIndices;
         final Set<String> matchingAllIndices;
 
         if (isLocalAll(requestedPatterns0)) {
@@ -217,14 +215,13 @@ public final class IndexResolverReplacer implements DCFListener {
                 log.trace(Arrays.toString(requestedPatterns0) + " is an LOCAL ALL pattern");
             }
             matchingAliases = Resolved.All_SET;
-            matchingIndices = Resolved.All_SET;
             matchingAllIndices = Resolved.All_SET;
 
         } else if (!remoteIndices.isEmpty() && localRequestedPatterns.isEmpty()) {
             if (log.isTraceEnabled()) {
                 log.trace(Arrays.toString(requestedPatterns0) + " is an LOCAL EMPTY request");
             }
-            return new Resolved.Builder().addOriginalRequested(Arrays.asList(requestedPatterns0)).addRemoteIndices(remoteIndices).build();
+            return new Resolved(Collections.emptySet(), Collections.emptySet(), Collections.emptySet(), Sets.newHashSet(requestedPatterns0), remoteIndices);
         }
 
         else {
@@ -232,25 +229,22 @@ public final class IndexResolverReplacer implements DCFListener {
             ClusterState state = clusterService.state();
 
             final SortedMap<String, AliasOrIndex> lookup = state.metaData().getAliasAndIndexLookup();
-            final Set<String> aliases = lookup.entrySet().stream().filter(e -> e.getValue().isAlias()).map(e -> e.getKey())
+            final List<String> aliases = lookup.entrySet().stream().filter(e -> e.getValue().isAlias()).map(e -> e.getKey())
+                    .collect(Collectors.toList());
+
+            Set<String> dateResolvedLocalRequestedPatterns = localRequestedPatterns
+                            .stream()
+                            .map(resolver::resolveDateMathExpression)
+                            .collect(Collectors.toSet());
+            //fill matchingAliases
+            matchingAliases = dateResolvedLocalRequestedPatterns
+                    .stream()
+                    .filter(pattern -> WildcardMatcher.matchAny(pattern, aliases))
                     .collect(Collectors.toSet());
 
-            matchingAliases = new HashSet<>(localRequestedPatterns.size() * 10);
-            matchingIndices = new HashSet<>(localRequestedPatterns.size() * 10);
-            matchingAllIndices = new HashSet<>(localRequestedPatterns.size() * 10);
-
-            //fill matchingAliases
-            for (String localRequestedPattern : localRequestedPatterns) {
-                final String requestedPattern = resolver.resolveDateMathExpression(localRequestedPattern);
-                final List<String> _aliases = WildcardMatcher.getMatchAny(requestedPattern, aliases);
-                matchingAliases.addAll(_aliases);
-            }
-
-
-            List<String> _indices;
+            Set<String> _indices;
             try {
-                _indices = new ArrayList<>(
-                        Arrays.asList(resolver.concreteIndexNames(state, indicesOptions, localRequestedPatterns.toArray(new String[0]))));
+                _indices = Sets.newHashSet(resolver.concreteIndexNames(state, indicesOptions, localRequestedPatterns.toArray(new String[0])));
                 if (log.isDebugEnabled()) {
                     log.debug("Resolved pattern {} to {}", localRequestedPatterns, _indices);
                 }
@@ -259,36 +253,13 @@ public final class IndexResolverReplacer implements DCFListener {
                     log.debug("No such indices for pattern {}, use raw value", localRequestedPatterns);
                 }
 
-                _indices = new ArrayList<>(localRequestedPatterns.size());
-
-                for (String requestedPattern : localRequestedPatterns) {
-                    _indices.add(resolver.resolveDateMathExpression(requestedPattern));
-                }
-
+                _indices = dateResolvedLocalRequestedPatterns;
             }
 
-            final List<String> _aliases = WildcardMatcher.getMatchAny(localRequestedPatterns.toArray(new String[0]), aliases);
-
-            matchingAllIndices.addAll(_indices);
-
-            if (_aliases.isEmpty()) {
-                matchingIndices.addAll(_indices); //date math resolved?
-            } else {
-
-                if (!_indices.isEmpty()) {
-
-                    for (String al : _aliases) {
-                        Set<String> doubleIndices = lookup.get(al).getIndices().stream().map(a -> a.getIndex().getName()).collect(Collectors.toSet());
-                        _indices.removeAll(doubleIndices);
-                    }
-
-                    matchingIndices.addAll(_indices);
-                }
-            }
+            matchingAllIndices = _indices;
         }
 
-        return new Resolved.Builder(matchingAliases, matchingIndices, matchingAllIndices, null, requestedPatterns0, remoteIndices)
-                /*.addTypes(resolveTypes(request))*/.build();
+        return new Resolved(matchingAliases, matchingAllIndices, Collections.emptySet(), Sets.newHashSet(requestedPatterns0), remoteIndices);
 
     }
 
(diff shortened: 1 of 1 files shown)


hardik-k-shah:
Thanks for coming up with more details and improvement. We will review this and update here.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -324,18 +320,26 @@ public Resolved resolveRequest(final Object request) {
         final Resolved.Builder resolvedBuilder = new Resolved.Builder();
         final AtomicBoolean isIndicesRequest = new AtomicBoolean();
         getOrReplaceAllIndices(request, new IndicesProvider() {
+            // resolve cache helps us big time on bulk requests
+            final ConcurrentHashMap<IndexResolveKey, Resolved> cache = new ConcurrentHashMap<>();
```
vrozov: @DmitryOlshansky Possibly I miss something, is it necessary to use thread safe `Map` for the `cache`?

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -31,14 +31,7 @@
 package com.amazon.opendistroforelasticsearch.security.compliance;
 
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
```
vrozov: Please add `java.util.Collection` to the list of imports and remove `java.util.*`. Consider changing Java IDE preferences not to use `*` imports.

src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -30,14 +30,7 @@
 
 package com.amazon.opendistroforelasticsearch.security.privileges;
 
-import java.util.ArrayList;
```
vrozov: The same as `ComplianceConfig.java`

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -161,7 +163,7 @@ private static final boolean isLocalAll(final String... requestedPatterns) {
         return false;
     }
 
-    private Resolved resolveIndexPatterns(final IndicesOptions indicesOptions, final Object request, final String... requestedPatterns0) {
+    private Resolved resolveIndexPatterns(final IndicesOptions indicesOptions, final Boolean isSearchOrFieldsCapabilities, final String... requestedPatterns0) {
```
vrozov: avoid unnecessary boxing: `Boolean` -> `boolean`

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -208,49 +209,41 @@ private Resolved resolveIndexPatterns(final IndicesOptions indicesOptions, final
...
             ClusterState state = clusterService.state();
 
             final SortedMap<String, AliasOrIndex> lookup = state.metaData().getAliasAndIndexLookup();
-            final Set<String> aliases = lookup.entrySet().stream().filter(e -> e.getValue().isAlias()).map(e -> e.getKey())
-                    .collect(Collectors.toSet());
-
-            matchingAliases = new HashSet<>(localRequestedPatterns.size() * 10);
-            matchingIndices = new HashSet<>(localRequestedPatterns.size() * 10);
-            matchingAllIndices = new HashSet<>(localRequestedPatterns.size() * 10);
+            final List<String> aliases = lookup.entrySet().stream().filter(e -> e.getValue().isAlias()).map(e -> e.getKey())
```
vrozov: I don't understand `Set` to `List` conversion. Please explain why is it necessary and what is the significance of duplicates in the `List`?

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
DmitryOlshansky: Indeed I was thinking about it ever since I did this patch. What thrown me off was that atomic boolean, but no amount of reviewing code suggests any concurrency going on here.

It would be best to put boolean inside provider struct itself and pass a reference then check it. 

Obviously I expect great saving in heap by switching to plain old Hasap

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
DmitryOlshansky: There are no duplicates in aliases for the simple reason it originates as Set.

The problem is collecting from a Set to Set create a new Set which is significantly more expansive then collecting to a list (or array for that matter).

Use of set may stand for 2 things:

1. We want to clarify in our interface that no duplicates are to be expected and (typically) that order of elements is not insignificant.
2. We want to remove duplicates from original source so we create a Set.

Even while the 1st is generally fine it sometimes has to stay just List and with clarifying documentation about uniqueness of items in performance sensitive code.

This is the case of 1st that got extended to internal and even scope-local places where it serves no purpose except for defensive programming. And internal state is a poor place for such practices, if need be we'd use assertions.
It's not without cost as each time we re-create a Set from another set we'd got to add all elements to a new set, recomputing hash codes (in case of HashSet) or rebalancing the Tree (in case of TreeSet).

DmitryOlshansky:
@vrozov  I tried to address all of your comments, see if there is anything else catching your eye. Thanks!

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
vrozov: For this particular `Set` -> `List` conversion, the source is indeed `Set`, but in many other places the source is not a `Set` and, I guess, is subject to duplicates.

What is the expected size of those sets/lists? It is not data, it is security metadata and I would not expect it to exceed 100K.  With 100K the difference between creating `Set` and `List` is negligible.

DmitryOlshansky: The numbers above speak for themselves I think.

Also the ratio of time to build a hash set vs time to build a list is never close to 1. HashSet is always ~X times more expansive especially with strings as list building will not even look at characters but hash set has to compute a hash for each of them, possibly multiple times due to rehash.

The absolute value may be fairly low, but this operation here is done many thousands of times. Building a million of hash sets of size 10 is way more expansive then lists of size 10. As I already did the profiling and presented the numbers, I suggest you to do the same and see if it indeed will not make a difference

vrozov: I do not want to say that collecting to `Set` is faster than collecting to `List`, but 

- `Set` provides additional functionality and it is not clear to me that the functionality of a `Set` is not required in all cases where `Set` was replaced with a `List`.
- The difference between collecting 1M times of an array of size 10 that contains strings with the length of 100 to `Set` vs `List` is negligible (200ms vs 150ms on my laptop).

DmitryOlshansky: As far as cases go - that is the precise problem with curent defensively programmed code - I had to reverse engineer what really needs to be a set and what not. If missed something feel free to point out. 

First I want to warn you of danger of doing correct microbenchmarks esp on Java, a platform with powerful JIT compiler featuring a number of agressive dynamic optimizations.
Thankfully for Java there is JMH, a tool built specifically for this purpose and used by JVM engineers as well.

Here is a complete benchmark:
```
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Param;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.State;

import java.util.List;
import java.util.Set;
import java.util.stream.Stream;


@State(Scope.Benchmark)
public class MyBenchmark {

    public Set<String> data = Stream.iterate(0, x -> x +1)
            .limit(10).map(x -> "logs-some-system"+x)
            .collect(Collectors.toSet());

    @Benchmark
    public List<String> collectToList() {
        return data.stream().filter(x -> x.startsWith("logs")).collect(Collectors.toList());
    }

    @Benchmark
    public Set<String> collectToSet() {
        return data.stream().filter(x -> x.startsWith("logs")).collect(Collectors.toSet());
    }
}
```

And the output on my laptop, sadly the numbers for sets have bigger deal of error margin maybe I was running something else on the laptop, I will re-run it later to check again.
```
REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
experiments, perform baseline and negative tests that provide experimental control, make sure
the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
Do not assume the numbers tell you what you want them to tell.

Benchmark                   Mode  Cnt        Score        Error  Units
MyBenchmark.collectToList  thrpt   25  5873007.810 ±  90055.803  ops/s
MyBenchmark.collectToSet   thrpt   25  3838232.740 ± 468729.614  ops/s

```

This gives me rough estimate  it is collecting a set of 10 short strings with simple filter to set is 50% slower  then to list. Asymptotic for larger sets are interesting too.

DmitryOlshansky: Made strings a bit longer and 100 items per set. 
Here are the numbers:
```
Benchmark                   Mode  Cnt       Score      Error  Units
MyBenchmark.collectToList  thrpt   50  597804.200 ± 9090.985  ops/s
MyBenchmark.collectToSet   thrpt   50  240782.714 ± 3631.795  ops/s
```
Now sets are ~0.40 times as fast as lists, or conversely list ~2.5x faster.

P.S. I forgot that Strings in Java cache the hashCode so the effect of hashing is dimnished in the benchmark (because it's the same literals over and over) but in Opendistro with HashSets left and right the strings likely to have hashCode cached by that time as well.

vrozov: My point is not that there are cases where `Set` is required and cases where `Collection` or `List` is sufficient. The point is that the time spent in collecting to `List` or `Set` is a small percentage compared to the time spent in other operations and changing `Set` to `Collection` is not well justified even if collecting to `Set` takes 2x or 3x more time collecting to `List` or any other collection.

Sure, we can go more scientific, correct and accurate route and use JMH (I am quite familiar with JMH, hot spot JVM compiler, JIT and vectorization), but it  produces similar (I am interested in the order of magnitude) results:

```
package com.amazon.opendistroforelasticsearch;

import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;
import org.openjdk.jmh.annotations.Param;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;

import java.nio.charset.Charset;
import java.util.Arrays;
import java.util.HashSet;
import java.util.Random;
import java.util.Set;
import java.util.concurrent.TimeUnit;

@State(Scope.Benchmark)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@Warmup(iterations = 5, time = 100, timeUnit = TimeUnit.MILLISECONDS)
@Measurement(iterations = 5, time = 100, timeUnit =  TimeUnit.MILLISECONDS)
public class CollectorsBenchmark {

    @Param({"10"})
    private int SIZE;

    @Param({"100"})
    private int LENGTH;

    private String[] array;
    private Set<String> set;
    private final Charset charset = Charset.forName("UTF-8");

    @Setup
    public void setup() {
        Random random = new Random();
        array = new String[SIZE];
        set = new HashSet<>(SIZE);
        for (int i = 0; i < SIZE; i++) {
            byte[] string = new byte[LENGTH];
            random.nextBytes(string);
            array[i] = new String(string, charset);
            set.add(array[i]);
        }
    }

    @Benchmark
    public void benchmarkSetToList() {
        set.stream().collect(Collectors.toList());
    }

    @Benchmark
    public void benchmarkSetToSet() {
        set.stream().collect(Collectors.toSet());
    }

    @Benchmark
    public void benchmarkArrayToList() {
        Arrays.stream(array).collect(Collectors.toList());
    }

    @Benchmark
    public void benchmarkArrayToSet() {
        Arrays.stream(array).collect(Collectors.toSet());
    }
}
```
```
Benchmark                                 (LENGTH)  (SIZE)  Mode  Cnt    Score    Error  Units
CollectorsBenchmark.benchmarkArrayToList       100      10  avgt   25   63.526 ±  5.563  ns/op
CollectorsBenchmark.benchmarkArrayToSet        100      10  avgt   25  149.283 ± 16.372  ns/op
CollectorsBenchmark.benchmarkSetToList         100      10  avgt   25  121.944 ± 21.390  ns/op
CollectorsBenchmark.benchmarkSetToSet          100      10  avgt   25  193.206 ± 18.709  ns/op
```

In this particular case, `String` is a key of the `lookup` (`SortedMap`) so its hash code is computed and cached by the time keys are collected to a `Set` or a `List`.

DmitryOlshansky: > The point is that the time spent in collecting to List or Set is a small percentage compared to the time spent in other operations and changing Set to Collection is not well justified even if collecting to Set takes 2x or 3x more time collecting to List or any other collection.

This is quite interesting as I recall you prompted me to avoid boxing a boolean and to use collect to array instead of collect to List. Was the reason puristic following of rules or was “prefer more efficient version” your motivation? 

In this particular case (object is local to this function) the semantic gain of having Set instead of List is ZERO but it costs more. If you absolutely want it - I’ll put it back.

Most of other places are also either internal helpers for this file or the callsite just iterates the elements. 

Please point out specific places where you believe having Set is semantically important to justify the extra cost in performance and I’ll roll them back.

vrozov: The difference is that boxing a boolean or collecting to a `List` instead of an array does not provide any additional functionality compared to collecting to a `List` instead of a `Set`. The same applies to the usage of `ConcurrentHashMap` where `HashMap` is sufficient. It is necessary to guarantee the correctness of data first and then see how it can be obtained more efficiently. In addition, using `List` instead of a `Set` may be OK for now, but may lead to bugs in the future.

My concern is not specific to this particular code change, but more to `Resolved.getAllIndices()` API change. Sorry for the confusion, I should post my comment there. I would be OK to change `Set` to `List` or an array for a local variable as long as it does not cause API change.

In the future, I would consider changing to a lazy evaluation of underlying collections (switching to `Stream` or `Iterator`).

DmitryOlshansky: > In the future, I would consider changing to a lazy evaluation of underlying collections (switching to Stream or Iterator).
 
Amen.

> My concern is not specific to this particular code change, but more to Resolved.getAllIndices() API change. Sorry for the confusion, I should post my comment there. I would be OK to change Set to List or an array for a local variable as long as it does not cause API change.

That I agree with. Indeed public API-level concerns are important. I’ll change that place to use Sets.

vrozov: Please rebase against master during the next round of changes to simplify PR review

DmitryOlshansky:
@vrozov  Squashed and rebased on top of master. Also minimized the scope of changes, most sets got restored.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -209,48 +208,43 @@ private Resolved resolveIndexPatterns(final IndicesOptions indicesOptions, final
...
+                            .map(resolver::resolveDateMathExpression)
+                            .collect(Collectors.toSet());
             //fill matchingAliases
-            for (String localRequestedPattern : localRequestedPatterns) {
-                final String requestedPattern = resolver.resolveDateMathExpression(localRequestedPattern);
-                final List<String> _aliases = WildcardMatcher.getMatchAny(requestedPattern, aliases);
-                matchingAliases.addAll(_aliases);
-            }
-
+            matchingAliases = dateResolvedLocalRequestedPatterns
```
vrozov: I would expect a collection of aliases, not patterns, please check.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -316,26 +287,61 @@ public boolean replace(final TransportRequest request, boolean retainMode, Strin
...
-                final Resolved iResolved = resolveIndexPatterns(indicesOptions, localRequest, original);
-                resolvedBuilder.add(iResolved);
-                isIndicesRequest.set(true);
-
-                if(log.isTraceEnabled()) {
-                    log.trace("Resolved patterns {} for {} ({}) to {}", original, localRequest.getClass().getSimpleName(), request.getClass().getSimpleName(), iResolved);
+                final boolean isSearchOrFieldCapabilities = localRequest instanceof FieldCapabilitiesRequest || localRequest instanceof SearchRequest;
+                final IndexResolveKey lookup = new IndexResolveKey(indicesOptions, isSearchOrFieldCapabilities, original);
+                // skip the whole thing if we have seen this exact resolveIndexPatterns result
+                if (cache.get(lookup) == null) {
```
vrozov: There is no need to call `cache.get(...)` (double get).

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -316,26 +287,61 @@ public boolean replace(final TransportRequest request, boolean retainMode, Strin
         }, false);
     }
 
+    static final class IndexResolveKey {
```
vrozov: nit: please declare the class and all member variables `private`.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -316,26 +287,61 @@ public boolean replace(final TransportRequest request, boolean retainMode, Strin
...
         if(log.isDebugEnabled()) {
             log.debug("Resolve aliases, indices and types from {}", request.getClass().getSimpleName());
         }
 
         final Resolved.Builder resolvedBuilder = new Resolved.Builder();
         final AtomicBoolean isIndicesRequest = new AtomicBoolean();
+
         getOrReplaceAllIndices(request, new IndicesProvider() {
+            // resolve cache helps us big time on bulk requests
+            final HashMap<IndexResolveKey, Resolved> cache = new HashMap<>();
```
vrozov: nit: use `Map` interface in the declaration.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
DmitryOlshansky: In fact it needs to call get, but there is no need to use the result down the line...
Switched to Set and renamed "cache" to resolvedBefore.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
DmitryOlshansky: Done

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
DmitryOlshansky: Yeah, that's a good catch.. Test are passing which looks dangerous.
Switched around to collect aliases.

vrozov: Can you add this to the list of issues?

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
vrozov: With `Map` it is not necessary to call both `get()` and `computeIfAbsent()` and I agree that it is not necessary to maintain `Map` and `Set` provides the required functionality.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -49,6 +47,7 @@
 import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
 
+import com.google.common.base.Objects;
```
vrozov: nit: move it next to `com.google.common.collect.Sets` import.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -316,26 +284,61 @@ public boolean replace(final TransportRequest request, boolean retainMode, Strin
...
+            if (o == null || getClass() != o.getClass()) return false;
+            IndexResolveKey that = (IndexResolveKey) o;
+            return isSearchOrFieldCapabilities == that.isSearchOrFieldCapabilities &&
+                    Objects.equal(opts, that.opts) &&
+                    Arrays.equals(original, that.original);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hashCode(opts, isSearchOrFieldCapabilities) + 31*Arrays.hashCode(original);
```
vrozov: nit: not sure if `Objects.hashCode()` is the best option here. I would rather use `opts.hashCode()` and `Boolean.hashCode(isSearchOrFieldCapabilities)`.

DmitryOlshansky:
@vrozov As per comments I'll split this in 2 PRs:
- one to add already visited mask set to IndexResolverReplacer
- another one to simplify Resolved class and its builder

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
DmitryOlshansky: https://github.com/opendistro-for-elasticsearch/security/issues/307

DmitryOlshansky:
resolve mask goes here:
https://github.com/opendistro-for-elasticsearch/security/pull/309

DmitryOlshansky:
And the rest of surgery on Resolved  goes here:
https://github.com/opendistro-for-elasticsearch/security/pull/310

DmitryOlshansky:
First part is in, thanks guys.
I’m closing this one - let’s continue with the second part in its own PR.

PR #291, opened 2020-03-11
Allowing hot-reload certificates for 0.8 and below

palashhedau:
Functionality - Allowing hot-reload certificates for 0.8 and below

This will allow users to update the SSL certs without restarting elasticsearch

Removed settings from Security as they are present in SSL Plugin


palashhedau:
Tests run: 78, Failures: 0, Errors: 0, Skipped: 4

PR #294, opened 2020-03-14
[RFC] Better pattern matching

DmitryOlshansky:
*Issue #, if available:*
Should create one I guess?

*Description of changes:*

There is a lot of implicit pattern matching code in OpenDistro Security that relies on calling ad-hoc functions on sets, arrays and other collections of strings and their combinations. This situation makes it hard to optimize pattern matching and even more importantly the intent of variables and code is less clear (is this a set of strings or can it contain patterns, if so which ones?).

This is work in progress and I'd like to collect feedback early and make use of your excellent CI pipeline ;)


src/main/java/com/amazon/opendistroforelasticsearch/security/support/wildcard/Wildcard.java
```diff
@@ -0,0 +1,70 @@
+package com.amazon.opendistroforelasticsearch.security.support.wildcard;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.stream.Collectors;
+
+// Case-sensitivity and other options are defined by concrete Wildcard implementation
+public interface Wildcard {
```
vrozov, line 9: I don't think it is necessary to introduce `Wildcard` interface. Combine `Wildcard` with `WildcardMatcher` class, especially that `WildcardMatcher` is a more suitable name, IMO. Make `WildcardMatcher` abstract class that implements `Predicate`. All other `Wildcard` classes should be inner `private static` classes of `WildcardMatcher` class.

src/main/java/com/amazon/opendistroforelasticsearch/security/support/wildcard/Wildcard.java
```diff
@@ -0,0 +1,70 @@
...
+    boolean matches(String candidate);
+
+    default boolean matchesAny(Collection<String> candidates) {
```
vrozov, line 12: `matchesAny` -> `matchAny` to be consistent with `WildcardMatcher`.

src/main/java/com/amazon/opendistroforelasticsearch/security/support/wildcard/Wildcard.java
```diff
@@ -0,0 +1,70 @@
...
+        return candidates.stream().anyMatch(this::matches);
+    }
+
+    default boolean matchesAny(String[] candidates) {
+        return Arrays.stream(candidates).anyMatch(this::matches);
+    }
+
+    // TODO: make serializable, hashable etc.
+    Wildcard ANY = new Wildcard() {
```
vrozov, line 21: Needs to implement `matchAny` and assert if `matches` (`test`) is called.

src/main/java/com/amazon/opendistroforelasticsearch/security/support/wildcard/Wildcard.java
```diff
@@ -0,0 +1,70 @@
...
+    // TODO: make serializable, hashable etc.
+    Wildcard NONE = new Wildcard() {
+        @Override
+        public boolean matches(String candidate) {
+            return false;
+        }
+    };
+
+    // This may in future use more optimized techniques to combine multiple wildcards in a single automaton
+    static Wildcard caseSensitiveAny(Collection<String> patterns) {
```
vrozov, line 37: use flag for case sensitivity: `any(Collection<String> patterns, boolean isCaseSensitive)`

src/main/java/com/amazon/opendistroforelasticsearch/security/support/wildcard/Wildcard.java
DmitryOlshansky: Yeah, Wildcard -> WildcardMatcher is good idea.

src/main/java/com/amazon/opendistroforelasticsearch/security/support/wildcard/Wildcard.java
DmitryOlshansky: Ok

src/main/java/com/amazon/opendistroforelasticsearch/security/support/wildcard/Wildcard.java
DmitryOlshansky: Care to elaborate about matches(test) ?

src/main/java/com/amazon/opendistroforelasticsearch/security/support/wildcard/Wildcard.java
DmitryOlshansky: I believe case sensitive is a property of pattern not of matching. Much like JDK Pattern class.

src/main/java/com/amazon/opendistroforelasticsearch/security/support/wildcard/Wildcard.java
vrozov: Please ignore my comment regarding `assert`, it may still be a valid call.

src/main/java/com/amazon/opendistroforelasticsearch/security/support/wildcard/Wildcard.java
vrozov: It is a property of `WildcardMatcher` and there is no difference whether it is designated by the method name `caseSensitiveAny(patterns)` or by the flag `any(pattern, true)`.

DmitryOlshansky: Or you mean the factory method, yeah makes sense

DmitryOlshansky:
@vrozov Should be much simpler and full-scale now. Time to review the whole lot, thanks for the early feedback.

src/main/java/com/amazon/opendistroforelasticsearch/security/support/wildcard/Wildcard.java
DmitryOlshansky: Done

legacy/securityconfig_v6/config.yml
```diff
@@ -207,7 +207,7 @@ opendistro_security:
             # Filter to search for users (currently in the whole subtree beneath userbase)
             # {0} is substituted with the username 
             usersearch: '(uid={0})'
-            # Skip users matching a user name, a wildcard or a regex pattern
+            # Skip users matching a user name, a WildcardMatcher or a regex pattern
```
vrozov, line 210: There is a large number of `wildcard` to `WildcardMatcher` renames that need to be reverted back. Please undo those changes to continue with the PR review.

DmitryOlshansky: Yeah, botched up mass-renames in a hurry. Will look into the tests as well and ping you back.

DmitryOlshansky: All done, AFAICT.

DmitryOlshansky:
Waiting on the test run results, but otherwise it should be in the reviewable shape again. @vrozov

legacy/securityconfig_v6/config.yml
vrozov: @DmitryOlshansky Please double check. For example `legacy/securityconfig_v6/config.yml` should not refer to WildcardMatcher.

legacy/securityconfig_v6/roles.yml
```diff
@@ -17,7 +17,7 @@
 # for the request to be successful. If no role match then the request will be denied. Currently a match must happen within
 # one single role - that means that permissions can not span multiple roles. 
 
-# For <permission>, <indexname or alias> and <type> simple wildcards and regular expressions are possible. 
+# For <permission>, <indexname or alias> and <type> simple wildcards and regular expressions are possible.
```
vrozov, line 20: Either don't delete trailing spaces or delete them for the entire file. I'd prefer to handle this in a separate commit/PR.

DmitryOlshansky:
There are still quite a few failing tests, I'm slowly fixing that a few cases at a time.
The biggest problem is to make sure serialization works and is compatible with the old version.

vrozov:
@DmitryOlshansky Please review #458. I rebased your commits against master and added changes that I propose in a separate commit.

vrozov:
Merged as part of #458.

PR #303, opened 2020-03-20
AuditLog: Refactor to use thread pool config

sujithvm:
Changes:
- Creating a AsyncStoragePoolConfig which is injected to AsyncStoragePool 

No new functionality added.


src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AsyncStoragePoolConfig.java
```diff
@@ -0,0 +1,40 @@
+package com.amazon.opendistroforelasticsearch.security.auditlog.config;
+
+import com.amazon.opendistroforelasticsearch.security.support.ConfigConstants;
+import org.elasticsearch.common.settings.Settings;
+
+public class AsyncStoragePoolConfig {
```
vrozov: Will this class be used outside of `AsyncStoragePool`? The current usage is limited to `AsyncStoragePool`, why not to make it `static` inside `AsyncStoragePool`?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AsyncStoragePool.java
```diff
@@ -72,11 +59,18 @@ public void submit(AuditMessage message, AuditLogSink sink) {
 		}
 	}
 
-	private ThreadPoolExecutor createExecutor(final int threadPoolSize, final int maxQueueLen) {
+	private ThreadPoolExecutor createExecutor(final AsyncStoragePoolConfig config) {
```
vrozov: Not sure why this method is necessary, can't it be inlined in the constructor?

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/ThreadPoolSettingsTest.java
```diff
@@ -15,21 +15,21 @@
...
 
-	@Test
-	public void testNoMultipleEndpointsConfiguration() throws Exception {
-		Settings settings = Settings.builder().loadFromPath(FileHelper.getAbsoluteFilePathFromClassPath("auditlog/endpoints/sink/configuration_no_multiple_endpoints.yml")).build();
-		AuditMessageRouter router = createMessageRouterComplianceEnabled(settings);
-		Assert.assertEquals(5, router.storagePool.threadPoolSize);
-		Assert.assertEquals(200000, router.storagePool.threadPoolMaxQueueLen);
-	}
+    @Test
+    public void testNoMultipleEndpointsConfiguration() throws Exception {
```
vrozov: Should this test be moved to `RoutingConfigurationTest`?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AuditMessageRouter.java
```diff
@@ -51,7 +52,7 @@
 
 	public AuditMessageRouter(final Settings settings, final Client clientProvider, ThreadPool threadPool, final Path configPath) {
 		this.sinkProvider = new SinkProvider(settings, clientProvider, threadPool, configPath);
-		this.storagePool = new AsyncStoragePool(settings);
+		this.storagePool = new AsyncStoragePool(AsyncStoragePoolConfig.getConfig(settings));
```
vrozov: Is it necessary to change the signature of the `AsyncStoragePool` constructor? Can it continue to accept `Settings` and create `AsyncStoragePoolConfig` inside?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AsyncStoragePoolConfig.java
```diff
@@ -0,0 +1,40 @@
...
+    private static final int DEFAULT_THREAD_POOL_SIZE = 10;
+    private static final int DEFAULT_THREAD_POOL_MAX_QUEUE_LEN = 100 * 1000;
+
+    private final int threadPoolSize;
+    private final int threadPoolMaxQueueLen;
+
+    public AsyncStoragePoolConfig(int threadPoolSize, int threadPoolMaxQueueLen) {
```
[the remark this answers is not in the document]
sujithvm: Public constructor is not needed for this PR but it will be made public soon in upcoming PRs as config will no longer be generated from Settings.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AsyncStoragePoolConfig.java
```diff
@@ -0,0 +1,40 @@
...
+        if (threadPoolSize <= 0) {
```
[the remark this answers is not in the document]
sujithvm: I believe it is a very common practice? 
Creating a thread pool with negative value will complain and we can go two ways from there:
- Complain and crash the process
- Use default values. 

I will log a warning for this case.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AsyncStoragePoolConfig.java
```diff
@@ -0,0 +1,40 @@
...
+            threadPoolSize = DEFAULT_THREAD_POOL_SIZE;
+        }
+
+        if (threadPoolMaxQueueLen <= 0) {
```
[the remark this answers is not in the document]
sujithvm: Same as above

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AsyncStoragePool.java
sujithvm: This class will be subscribed to a config listener in future PR which will update for any config changes which will call this function again. Hence moved it to be passed in a method.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AuditMessageRouter.java
sujithvm: I believe it is a good idea to pass only whats needed to a class.

sujithvm:
The goal is to consolidate configurations that can be hot-reloaded. 
These refactors are necessary for that.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AsyncStoragePoolConfig.java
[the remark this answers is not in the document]
vrozov: My take is to complain (log error) and crash. IMO, silently switching to a default value will cause confusion. Logging a warning may go unnoticed.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AsyncStoragePoolConfig.java
sujithvm: Yes. It will be used outside AsyncStoragePool to notify config changes. Moreover having it static inner class will make an outer class dependent on inner class. Eg: AsyncStoragePool constructor accepting inner class AsyncStoragePoolConfig dependency.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AsyncStoragePoolConfig.java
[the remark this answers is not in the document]
sujithvm: I m slightly not sure, we should throw an exception and make it crash.
- Reading from elasticsearch settings asks for a default value if not configured.
- Users who don't care to make a specific config setting, this will crash for them too instead of relying on defaults.
- This would make thread pool setting mandatory (may need to consult product and update docs).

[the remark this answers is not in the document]
vrozov: If one of the settings is not present it needs to be replaced with a default value, there should not be any crash. In case an invalid value is passed to the constructor, it should be OK to throw an exception and fail to construct `AsyncStoragePoolConfig` object.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AsyncStoragePoolConfig.java
vrozov: Will it only notify `AsyncStoragePool` about config changes? The class name `AsyncStoragePoolConfig ` suggests that it is specific to `AsyncStoragePool` and can't be used with other thread pools. In this case, I don't see anything wrong with making it more explicit and making `AsyncStoragePool` constructor accept an instance of an inner **static** class. It is simply a way to organize naming.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AsyncStoragePool.java
vrozov: To support config changes, it would be necessary to properly handle existing pool shutdown and creation of a new thread pool. What if there are pending tasks to execute? What if a running task takes too long to complete? This method is likely to require significant refactoring and new functionality before it can support config change.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AsyncStoragePoolConfig.java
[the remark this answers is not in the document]
sujithvm: Alright

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AsyncStoragePool.java
sujithvm: Thread pool size can be dynamically changed. However dynamically changing the queue length will be more challenging as the current queue must depleted as LinkedBlockingQueue size cant be changed once. For sure. it will need more work. 
This PR just cleans up exisiting code.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AsyncStoragePoolConfig.java
sujithvm: I will rename it to ThreadPoolConfig to make it more generic if we want to use config for other thread pools.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AuditMessageRouter.java
vrozov: Yes, but on the other side, now `AuditMessageRouter` is exposed to `Settings`, `AsyncStoragePool` and in addition to `AsyncStoragePoolConfig`.

sujithvm: Those also will undergo changes.

sujithvm: All required dependencies will be injected for all the classes

sujithvm: I really want to keep scope small for PRs as previous PRs went un-attended for weeks.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AsyncStoragePool.java
vrozov: Please clarify, if you want to create another Executor on configuration change, it would be necessary to deal with the issues I mentioned. If configuration change will modify an existing Executor, this function can't be reused as it creates new instance of `ThreadPoolExecutor`. I'd recommend to inline the function, especially that it looks like it was supposed to be `static` (it takes member variable `config` as an argument), but also requires access to `log` that is not static.

sujithvm: I was answering your concern of handling existing queue when there is a config change. I passed this as a param because createExecutor looked like a candidate that can be made separate. I will make this static.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/ThreadPoolConfig.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public int getThreadPoolMaxQueueLen() {
+        return threadPoolMaxQueueLen;
+    }
+
+    public static ThreadPoolConfig getConfig(Settings settings) {
+        int threadPoolSize = settings.getAsInt(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_THREADPOOL_SIZE, DEFAULT_THREAD_POOL_SIZE);
+        int threadPoolMaxQueueLen = settings.getAsInt(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_THREADPOOL_MAX_QUEUE_LEN, DEFAULT_THREAD_POOL_MAX_QUEUE_LEN);
+
+        if (threadPoolSize <= 0) threadPoolSize = DEFAULT_THREAD_POOL_SIZE;
```
vrozov: The check is not necessary. If there are no settings, it will return a default value, otherwise, it should check validity.

sujithvm: A lot of tests are setting this value to be 0 with the expectation to use the default thread pool

vrozov: There are only 3 tests that incorrectly set the value to be 0. If they want to rely on the default value, they should either set it explicitly to the default value or not to set it at all.

sujithvm: I m really torn on changing this because I agree with you but removing this will change the functionality 
- previous: if set <= 0 or not defined, use default 
- current: if set <= 0, complain with exception. if not defined, use default.

Changing functionality is something normally I wouldn't do in a refactor. 
There are actually 40+ tests explicitly setting: opendistro_security.audit.threadpool.size to 0. 
If you think changing the functionality is the correct way, let me know. I will change it.

vrozov: I suggest changing the behavior to avoid the possibility of overruling explicitly set by a user settings. IMO, tests need to be changed anyway to avoid using `"opendistro_security.audit.threadpool.size"` string instead of `OPENDISTRO_SECURITY_AUDIT_THREADPOOL_SIZE` constant. If a test needs to use **not a default value** it should use `ConfigConstants.OPENDISTRO_SECURITY_AUDIT_THREADPOOL_SIZE`.

sujithvm:
Fixes https://github.com/opendistro-for-elasticsearch/security/issues/311

hardik-k-shah:
we should do profiling on default size 10 for opendistro_security.audit.threadpool.size

PR #304, opened 2020-03-20
AuditLog: Refactor to use AuditCategory enum

sujithvm:
Changes:
- Pull out AuditCategory from AuditMessage 
- Use EnumSet for all categories instead of List of Strings 

No new functionality added.

Fixes: https://github.com/opendistro-for-elasticsearch/security/issues/323

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
@@ -130,15 +130,29 @@ public abstract class AbstractAuditLog implements AuditLog {
         restAuditingEnabled = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_REST, true);
         transportAuditingEnabled = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_TRANSPORT, true);
 
-        disabledRestCategories = AuditCategory.parse(getConfigList(
-                settings,
-                ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_REST_CATEGORIES,
-                DEFAULT_DISABLED_CATEGORIES));
+        final List<String> disabledRestCategoriesList = settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_REST_CATEGORIES, DEFAULT_DISABLED_CATEGORIES);
 
-        disabledTransportCategories = AuditCategory.parse(getConfigList(
-                settings,
-                ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_TRANSPORT_CATEGORIES,
-                DEFAULT_DISABLED_CATEGORIES));
+        if (disabledRestCategoriesList.size() == 1 && "NONE".equalsIgnoreCase(disabledRestCategoriesList.get(0))) {
+            disabledRestCategories = EnumSet.noneOf(AuditCategory.class);
+        } else {
+            disabledRestCategories = AuditCategory.parse(disabledRestCategoriesList);
+        }
+
+        if (!disabledRestCategories.isEmpty()) {
+            log.info("Configured categories on rest layer to ignore: {}", disabledRestCategories);
+        }
+
+        final List<String> disabledTransportCategoriesList = settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_TRANSPORT_CATEGORIES, DEFAULT_DISABLED_CATEGORIES);
+
+        if (disabledTransportCategoriesList.size() == 1 && "NONE".equalsIgnoreCase(disabledTransportCategoriesList.get(0))) {
+            disabledTransportCategories = EnumSet.noneOf(AuditCategory.class);
+        } else {
+            disabledTransportCategories = AuditCategory.parse(disabledTransportCategoriesList);
+        }
+
+        if (!disabledTransportCategories.isEmpty()) {
+            log.info("Configured categories on transport layer to ignore: {}", disabledTransportCategories);
+        }
 
         logRequestBody = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_LOG_REQUEST_BODY, true);
         resolveIndices = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_RESOLVE_INDICES, true);
@@ -179,16 +193,6 @@ public abstract class AbstractAuditLog implements AuditLog {
         }
 
         this.excludeSensitiveHeaders = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_EXCLUDE_SENSITIVE_HEADERS, true);
-    }
-
-    private static List<String> getConfigList(final Settings settings,
-                                              final String key,
-                                              final List<String> defaultList) {
-        List<String> list = settings.getAsList(key, defaultList);
-        if (list.size() == 1 && "NONE".equals(list.get(0))) {
-            return Collections.emptyList();
-        }
-        return list;
     }
 
     @Override
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
@@ -132,7 +132,7 @@ public abstract class AbstractAuditLog implements AuditLog {
 
         final List<String> disabledRestCategoriesList = settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_REST_CATEGORIES, DEFAULT_DISABLED_CATEGORIES);
 
-        if (disabledRestCategoriesList.size() == 1 && "NONE".equalsIgnoreCase(disabledRestCategoriesList.get(0))) {
+        if (disabledRestCategoriesList.isEmpty() || (disabledRestCategoriesList.size() == 1 && "NONE".equalsIgnoreCase(disabledRestCategoriesList.get(0)))) {
             disabledRestCategories = EnumSet.noneOf(AuditCategory.class);
         } else {
             disabledRestCategories = AuditCategory.parse(disabledRestCategoriesList);
@@ -144,7 +144,7 @@ public abstract class AbstractAuditLog implements AuditLog {
 
         final List<String> disabledTransportCategoriesList = settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_TRANSPORT_CATEGORIES, DEFAULT_DISABLED_CATEGORIES);
 
-        if (disabledTransportCategoriesList.size() == 1 && "NONE".equalsIgnoreCase(disabledTransportCategoriesList.get(0))) {
+        if (disabledTransportCategoriesList.isEmpty() || (disabledTransportCategoriesList.size() == 1 && "NONE".equalsIgnoreCase(disabledTransportCategoriesList.get(0)))) {
             disabledTransportCategories = EnumSet.noneOf(AuditCategory.class);
         } else {
             disabledTransportCategories = AuditCategory.parse(disabledTransportCategoriesList);
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
@@ -25,6 +25,7 @@ import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -67,7 +68,6 @@ import org.elasticsearch.threadpool.ThreadPool;
 import org.elasticsearch.transport.TransportRequest;
 
 import com.amazon.opendistroforelasticsearch.security.auditlog.AuditLog;
-import com.amazon.opendistroforelasticsearch.security.auditlog.impl.AuditMessage.Category;
 import com.amazon.opendistroforelasticsearch.security.compliance.ComplianceConfig;
 import com.amazon.opendistroforelasticsearch.security.dlic.rest.support.Utils;
 import com.amazon.opendistroforelasticsearch.security.support.Base64Helper;
@@ -80,6 +80,8 @@ import com.flipkart.zjsonpatch.JsonDiff;
 import com.google.common.io.BaseEncoding;
 
 public abstract class AbstractAuditLog implements AuditLog {
+
+    private static final List<String> DEFAULT_DISABLED_CATEGORIES = Arrays.asList(AuditCategory.AUTHENTICATED.toString(), AuditCategory.GRANTED_PRIVILEGES.toString());
 
     protected final Logger log = LogManager.getLogger(this.getClass());
     protected final ThreadPool threadPool;
@@ -97,9 +99,8 @@ public abstract class AbstractAuditLog implements AuditLog {
     private List<String> ignoredComplianceUsersForRead;
     private List<String> ignoredComplianceUsersForWrite;
     private final List<String> ignoreAuditRequests;
-    private final List<String> disabledRestCategories;
-    private final List<String> disabledTransportCategories;
-    private final List<String> defaultDisabledCategories = Arrays.asList(Category.AUTHENTICATED.toString(), Category.GRANTED_PRIVILEGES.toString());
+    private final EnumSet<AuditCategory> disabledRestCategories;
+    private final EnumSet<AuditCategory> disabledTransportCategories;
     private final List<String> defaultIgnoredUsers = Arrays.asList("kibanaserver");
     private final boolean excludeSensitiveHeaders;
 
@@ -129,27 +130,15 @@ public abstract class AbstractAuditLog implements AuditLog {
         restAuditingEnabled = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_REST, true);
         transportAuditingEnabled = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_TRANSPORT, true);
 
-        disabledRestCategories = new ArrayList<>(settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_REST_CATEGORIES, defaultDisabledCategories).stream()
-                .map(c->c.toUpperCase()).collect(Collectors.toList()));
+        disabledRestCategories = AuditCategory.parse(getConfigList(
+                settings,
+                ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_REST_CATEGORIES,
+                DEFAULT_DISABLED_CATEGORIES));
 
-        if(disabledRestCategories.size() == 1 && "NONE".equals(disabledRestCategories.get(0))) {
-            disabledRestCategories.clear();
-        }
-
-        if (disabledRestCategories.size() > 0) {
-            log.info("Configured categories on rest layer to ignore: {}", disabledRestCategories);
-        }
-
-        disabledTransportCategories = new ArrayList<>(settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_TRANSPORT_CATEGORIES, defaultDisabledCategories).stream()
-                .map(c->c.toUpperCase()).collect(Collectors.toList()));
-
-        if(disabledTransportCategories.size() == 1 && "NONE".equals(disabledTransportCategories.get(0))) {
-            disabledTransportCategories.clear();
-        }
-
-        if (disabledTransportCategories.size() > 0) {
-            log.info("Configured categories on transport layer to ignore: {}", disabledTransportCategories);
-        }
+        disabledTransportCategories = AuditCategory.parse(getConfigList(
+                settings,
+                ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_TRANSPORT_CATEGORIES,
+                DEFAULT_DISABLED_CATEGORIES));
 
         logRequestBody = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_LOG_REQUEST_BODY, true);
         resolveIndices = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_RESOLVE_INDICES, true);
@@ -184,44 +173,34 @@ public abstract class AbstractAuditLog implements AuditLog {
             log.info("Configured Users to ignore for write compliance events: {}", ignoredComplianceUsersForWrite);
         }
 
-
-
         ignoreAuditRequests = settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_IGNORE_REQUESTS, Collections.emptyList());
         if (ignoreAuditRequests.size() > 0) {
             log.info("Configured Requests to ignore: {}", ignoreAuditRequests);
         }
 
-        // check if some categories are invalid
-        for (String event : disabledRestCategories) {
-        	try {
-        		AuditMessage.Category.valueOf(event.toUpperCase());
-        	} catch(Exception iae) {
-        		log.error("Unkown category {}, please check opendistro_security.audit.config.disabled_categories settings", event);
-        	}
-		}
-
-        // check if some categories are invalid
-        for (String event : disabledTransportCategories) {
-            try {
-                AuditMessage.Category.valueOf(event.toUpperCase());
-            } catch(Exception iae) {
-                log.error("Unkown category {}, please check opendistro_security.audit.config.disabled_categories settings", event);
-            }
-        }
-
         this.excludeSensitiveHeaders = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_EXCLUDE_SENSITIVE_HEADERS, true);
+    }
+
+    private static List<String> getConfigList(final Settings settings,
+                                              final String key,
+                                              final List<String> defaultList) {
+        List<String> list = settings.getAsList(key, defaultList);
+        if (list.size() == 1 && "NONE".equals(list.get(0))) {
+            return Collections.emptyList();
+        }
+        return list;
     }
 
     @Override
     public void logFailedLogin(String effectiveUser, boolean securityadmin, String initiatingUser, TransportRequest request, Task task) {
         final String action = null;
 
-        if(!checkTransportFilter(Category.FAILED_LOGIN, action, effectiveUser, request)) {
+        if(!checkTransportFilter(AuditCategory.FAILED_LOGIN, action, effectiveUser, request)) {
             return;
         }
 
         final TransportAddress remoteAddress = getRemoteAddress();
-        final List<AuditMessage> msgs = RequestResolver.resolve(Category.FAILED_LOGIN, getOrigin(), action, null, effectiveUser, securityadmin, initiatingUser, remoteAddress, request, getThreadContextHeaders(), task, resolver, clusterService, settings, logRequestBody, resolveIndices, resolveBulkRequests, opendistrosecurityIndex, excludeSensitiveHeaders, null);
+        final List<AuditMessage> msgs = RequestResolver.resolve(AuditCategory.FAILED_LOGIN, getOrigin(), action, null, effectiveUser, securityadmin, initiatingUser, remoteAddress, request, getThreadContextHeaders(), task, resolver, clusterService, settings, logRequestBody, resolveIndices, resolveBulkRequests, opendistrosecurityIndex, excludeSensitiveHeaders, null);
 
         for(AuditMessage msg: msgs) {
             save(msg);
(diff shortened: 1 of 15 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
```diff
@@ -0,0 +1,44 @@
...
+    public static EnumSet<AuditCategory> parse(final List<String> categories) {
+        EnumSet<AuditCategory> set = EnumSet.noneOf(AuditCategory.class);
+        if (categories == null)
+            return set;
+
+        categories
+                .stream()
+                .filter(Objects::nonNull)
+                .map(category -> category.toUpperCase())
+                .forEach(category -> {
```
vrozov: nit: use `stream` API:
```
        return categories
                .stream()
                .filter(Objects::nonNull)
                .map(String::toUpperCase)
                .map(category -> {
                    try {
                        return AuditCategory.valueOf(category);
                    } catch (IllegalArgumentException ex) {
                        log.error("Invalid category type: {} Exception: {}", category, ex.getMessage());
                        return null;
                    }
                })
                .filter(Objects::nonNull)
                .collect(Collectors.toCollection(() -> set));
```

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
```diff
@@ -0,0 +1,44 @@
...
+                    try {
+                        set.add(AuditCategory.valueOf(category));
+                    } catch (IllegalArgumentException ex) {
+                        log.error("Invalid category type: {} Exception: {}", category, ex.getMessage());
```
vrozov: I am not sure that it is sufficient to log an error message and continue. It is possible that a category has a typo that causes parsing exception and it may go unnoticed (lost in log files).

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
```diff
@@ -0,0 +1,44 @@
...
+    AUTHENTICATED,
+    COMPLIANCE_DOC_READ,
+    COMPLIANCE_DOC_WRITE,
+    COMPLIANCE_EXTERNAL_CONFIG,
+    COMPLIANCE_INTERNAL_CONFIG_READ,
+    COMPLIANCE_INTERNAL_CONFIG_WRITE;
+
+    private static Logger log = LogManager.getLogger(AuditCategory.class);
+
+    public static EnumSet<AuditCategory> parse(final List<String> categories) {
```
vrozov: Consider using `Iterable` or `Collection` instead of a `List`. There is no need to know `size()`, so `Iterable` (preferred) or `Collection` should be sufficient.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
sujithvm: I ve considered using this but i felt it was not efficient to loop through the list twice again to filter and convert to set. I would prefer more traditional approach for doing this (removing stream api itself). What do you think?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
sujithvm: I believe if we can recover from some error, then complain and recover. I hope error logs are not that casually ignored.  
In this case, if incorrect category is provided, simply complain and ignore it.

sujithvm:
AuditCategory is a type of audit log and AuditMessage represents the actual message content that is logged. It does not make sense for category to be a subclass of message as we might want to do different operations based on category. Ideally would want something like this class { category, content } to actually represent what the an audit log means.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
vrozov: There will be no additional iterations over the stream (the only one is when it is consumed in the `collect()`), so it would be as efficient as `forEach()`.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
vrozov: Imagine that the `AuditCategory` is constructed using a call to REST API from `curl` and somebody mistypes a category name. The return code of REST API would be 200 and the user who called REST API would assume that it succeeded and even not bother to look at the error log.

vrozov:
My understanding is that `AuditCategory` specifies what categories of `AuditMessage` should be enabled/disabled.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
sujithvm: That should directly return a bad request if you are trying to make a log message for an incorrect category type. 
We can try to convert to Enum and then throw error in the validator of that rest api handler.  Not sure if we should handle it here.

sujithvm:
AuditMessage stores only the content of message. 
Next PR is to move AuditMessage into a builder pattern what can build a message for a given category.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
sujithvm: Yes. I thought of ignoring the second loop. 
I have updated to use stream api

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
vrozov: In the example, I refer to constructing `EnumSet<AuditCategory>` from a string passed as REST API parameter, not to constructing `AuditMessage` with an invalid `AuditCategory`.

sujithvm: Alright. Do you want to throw exception here instead ? Or 
Use something like:

```
        return categories
                .stream()
                .filter(Objects::nonNull)
                .map(String::toUpperCase)
                .map(AuditCategory::valueOf)
                .collect(Collectors.toCollection(() -> set));
```
And let java throw the IllegalArgumentException ?

sujithvm: This way, it will throw exception and prevent the service from starting up if elasticsearch.yml is configured incorrectly and can also assert for exceptions in REST api validators and throw bad request appropriately.

vrozov: It is OK to rely on `valueOf()` to throw an `IllegalArgumentException` or to convert it to a checked exception. I am slightly more inclined to `IllegalArgumentException`.

sujithvm: AuditCategory::valueOf will throw IllegalArgumentException if cannot be converted to enum. Am I missing something?

sujithvm: @vrozov The current behavior is to actually post error and ignore invalid enum and have some tests to do so.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -129,27 +128,17 @@ protected AbstractAuditLog(Settings settings, final ThreadPool threadPool, final
         restAuditingEnabled = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_REST, true);
         transportAuditingEnabled = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_TRANSPORT, true);
 
-        disabledRestCategories = new ArrayList<>(settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_REST_CATEGORIES, defaultDisabledCategories).stream()
-                .map(c->c.toUpperCase()).collect(Collectors.toList()));
+        final List<String> defaultDisabledCategories = Arrays.asList(AuditCategory.AUTHENTICATED.toString(), AuditCategory.GRANTED_PRIVILEGES.toString());
```
vrozov: Consider making it `private static` in `AuditCategory`.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -184,44 +173,34 @@ protected AbstractAuditLog(Settings settings, final ThreadPool threadPool, final
...
+    }
 
-        // check if some categories are invalid
-        for (String event : disabledTransportCategories) {
-            try {
-                AuditMessage.Category.valueOf(event.toUpperCase());
-            } catch(Exception iae) {
-                log.error("Unkown category {}, please check opendistro_security.audit.config.disabled_categories settings", event);
-            }
+    private static List<String> getConfigList(final Settings settings,
```
vrozov: Consider moving to `AuditCategory`.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -184,44 +173,34 @@ protected AbstractAuditLog(Settings settings, final ThreadPool threadPool, final
...
+                                              final String key,
+                                              final List<String> defaultList) {
+        List<String> list = new ArrayList<>(settings.getAsList(key, defaultList));
+        if (list.size() == 1 && "NONE".equals(list.get(0))) {
+            list.clear();
```
vrozov: No need to create new `ArrayList`. Use original return from `settings.getAsList(key, defaultList)` and `Collections.emptyList()` instead of clearing a new one.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: This is generic reading of config values and feel it should not be part of AuditCategory. Will be moving this to AuditConfig class

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
vrozov: > AuditCategory::valueOf will throw IllegalArgumentException if cannot be converted to enum. Am I missing something?

No. I was referring to the same.

> The current behavior is to actually post error and ignore invalid enum and have some tests to do so

IMO, there is a problem with an approach that tries to correct an invalid input. At a minimum, `AuditCategory` should allow a caller to get the list of categories that were not parsed or throw an exception.

sujithvm: Can we have this when we need it?

vrozov: What is your concern? Changing tests?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
vrozov: `AuditConfig` or `AuditCategory`, it does not belong to `AbstractAuditLog`, IMO.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
sujithvm: I meant to say, lets do it when we have a use-case ready. For sure, this will come up in future and let's target it that time. Maybe we will have more requirements.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: I plan to refactor these audit configurations into a separate AuditConfig class to represent all audit log configuration (basically abstracting it out of AbstractAuditLog). 
Let me create a milestone in Github for having better clarity of changes I want to make.

vrozov: As this is a new function, it will be better to put it to `AuditCategory` compared to `AbstractAuditLog`. If necessary, please add TODO comment.

sujithvm: getConfigList is something that will be reused by other configs like ignoredAuditUsers, ignoredComplianceUsersForRead, ignoredComplianceUsersForWrite. And they all should go into an AuditConfig class.

sujithvm: This is going to be fixed with #304 306

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
sujithvm: Could you explain the rationale/use-case for including this now? I think we are not on the same page. Where and what do you want do with string values that cannot be parsed ?

vrozov: The current behavior is to ignore an invalid value and continue. The invalid value may come from a configuration file (the only way supported now) or from an API call with a request to update/change configuration (in the future). In both cases, likely a user (system admin) made a mistake and simply mistyped **valid**  audit category name that the user (system admin) wants to be in effect. How quickly the user (system admin) may realize that the system ignored or partially ignored the input is questionable.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
vrozov: Can you introduce `AuditConfig` class in this PR with a single method and extend it in #306?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
vrozov: Note that the original code provided a better hint in the logging message of what setting has an unknown/invalid category. It may be good to keep it the same way should you decide to keep the existing behavior.  It is not necessary to log the `InvalidArgumentException` message.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: Why add here and not in #306 ?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
sujithvm: I will revert to throwing exception and will update tests to check for exception.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
vrozov: It is already used in this PR.

sujithvm: Sorry I m confused. What is already used in this PR?

vrozov: `getConfigList()` method

sujithvm: Are you saying to include all AuditConfig changes in this PR itself ?

sujithvm: #306 should now have only #323 and #312 changes if that what you are asking for

vrozov: No, only `getConfigList()`. Note that I have not looked at the other PR yet.

sujithvm: What is the value in adding just that in this ? Am I missing something?

vrozov: It will finalize this change and will not require more changes in the follow-up PRs.

sujithvm: I don't think we should assume any change is finalized and will not undergo more changes in the future PRs.
This function is just pulled out of existing code in AbstractAuditLog class into a method. 
I don't think it's a good idea to add incomplete AuditConfig just for not reviewing this change again as the new diff will have changes. 

I would recommend reviewing #306 instead which is a more complete solution of what you are looking for AuditConfig + AuditCategory.

vrozov: Once `getConfigList()` is moved to `AuditConfig` what other changes do you plan to apply to it?

sujithvm: It may not have any new change but I m not seeing any value in adding incomplete AuditConfig class here. 
I feel I m missing your point. Are you saying getConfigList does not need to reviewed again?

vrozov: >It may not have any new change but I m not seeing any value in adding incomplete AuditConfig class here.

`AuditConfig` is likely to evolve over time, so it will never be complete. 

>I feel I m missing your point. Are you saying getConfigList does not need to reviewed again?

It may require a review in case the follow-up PRs need to introduce other changes to `getConfigList()` that are not necessary for this PR, but it will be easy to review. In addition, it will also simplify effort if somebody needs to look at git history for the function.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -129,27 +130,15 @@ protected AbstractAuditLog(Settings settings, final ThreadPool threadPool, final
...
-                .map(c->c.toUpperCase()).collect(Collectors.toList()));
-
-        if(disabledRestCategories.size() == 1 && "NONE".equals(disabledRestCategories.get(0))) {
-            disabledRestCategories.clear();
-        }
-
-        if (disabledRestCategories.size() > 0) {
-            log.info("Configured categories on rest layer to ignore: {}", disabledRestCategories);
-        }
+        disabledRestCategories = AuditCategory.parse(getConfigList(
```
vrozov: My understanding is that not all `AuditCategory` applies to Rest or Transport categories. It may be good to log a warning if an admin tries to disable auditing for a category that does not apply. Note that this case is different from an invalid category.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -184,44 +173,34 @@ protected AbstractAuditLog(Settings settings, final ThreadPool threadPool, final
...
-            try {
-                AuditMessage.Category.valueOf(event.toUpperCase());
-            } catch(Exception iae) {
-                log.error("Unkown category {}, please check opendistro_security.audit.config.disabled_categories settings", event);
-            }
+    private static List<String> getConfigList(final Settings settings,
+                                              final String key,
+                                              final List<String> defaultList) {
+        List<String> list = settings.getAsList(key, defaultList);
+        if (list.size() == 1 && "NONE".equals(list.get(0))) {
```
vrozov: I'd recommend start deprecating `"NONE"` as a setting and log a warning. It is easy to use "none" instead of "NONE" especially that category names are case-insensitive.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: Since you are reviewing #306  we can close this PR

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: Maybe raise another issue for this?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: Then it should complain and throw exception. It is documented to use NONE https://opendistro.github.io/for-elasticsearch-docs/docs/security-audit-logs/#exclude-categories

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
vrozov: Please don't close this PR. It is almost ready to go.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
vrozov: OK, I'll file an issue and see what others think.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: #306 should have the change you were looking with getConfigList()

vrozov: I'd prefer not to squash commits in #306, so #306 won't address my concern.

sujithvm: So confused. What is your concern ? and why not squash commits?

sujithvm: Doesn't #306 address your concern with `getConfigList()` ?

vrozov: I'd prefer to keep `AuditCategory` changes separate from other changes introduced in #306.

sujithvm: It is the same commit that is cherry-picked 0cb732f . So changes are as it is

vrozov: Without squashing all commits together #306 does not address my concern regarding `getConfigList()` and I'd prefer not to squash commits to keep `AuditCategory` changes separate from other changes introduced in #306.

sujithvm: You concern of getConfigList() is resolved with #306 .
Why do you prefer to keep this change separate. How does it matter ?

sujithvm: If you are worried about any git history and having an atomic merge step (auditconfig with getConfigList), then going forward with #306 is the right way.

vrozov: With 2 separate commits, it will be much easier to understand what was changed and why and how changes are related for anyone who is not involved in the current PR in the future.

vrozov: I think you also agree that `getConfigList()` does not belong to `AbstractAuditLog` and it should be part of `AuditConfig` class. Should this method previously exist as part of `AbstractAuditLog` it would be OK to move it in subsequent PR. As it is a newly introduced method, I would prefer to put it in the right place when it is introduced.

sujithvm: I agree getConfigList() belongs to AuditConfig class and thats exactly what #306 does. 
At the same time, I dont believe this must exist in some incomplete class that may be introduced in a future PR.

Squashing the commits of #306 would look like everything was included in a single PR/commit keeping git history clean. 
Well, since this PR is not merged yet, this is not the right place when it was introduced and the right place would be #306

vrozov: Squashing all commits in #306 would create a single commit with a large number of changes that logically can be split into two commits. This will not only mess up the git history but more importantly, it will create a huge commit that will be harder to understand.

sujithvm:
I hope commit 31e72eafc6ba66b08d2e31cc53fd20ae6b010617 solves your concern of adding new method getConfigList . Reverted back to using what AbstractAuditLog originally had for parsing it.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -129,25 +130,27 @@ protected AbstractAuditLog(Settings settings, final ThreadPool threadPool, final
         restAuditingEnabled = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_REST, true);
         transportAuditingEnabled = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_TRANSPORT, true);
 
-        disabledRestCategories = new ArrayList<>(settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_REST_CATEGORIES, defaultDisabledCategories).stream()
-                .map(c->c.toUpperCase()).collect(Collectors.toList()));
+        final List<String> disabledRestCategoriesList = settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_REST_CATEGORIES, DEFAULT_DISABLED_CATEGORIES);
 
-        if(disabledRestCategories.size() == 1 && "NONE".equals(disabledRestCategories.get(0))) {
-            disabledRestCategories.clear();
+        if (disabledRestCategoriesList.size() == 1 && "NONE".equalsIgnoreCase(disabledRestCategoriesList.get(0))) {
```
vrozov: nit: as you are constructing an empty `EnumSet` directly here, add `disabledRestCategoriesList.isEmpty()` condition.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
```diff
@@ -0,0 +1,34 @@
...
+    AUTHENTICATED,
+    COMPLIANCE_DOC_READ,
+    COMPLIANCE_DOC_WRITE,
+    COMPLIANCE_EXTERNAL_CONFIG,
+    COMPLIANCE_INTERNAL_CONFIG_READ,
+    COMPLIANCE_INTERNAL_CONFIG_WRITE;
+
+    public static EnumSet<AuditCategory> parse(final Collection<String> categories) {
+        EnumSet<AuditCategory> set = EnumSet.noneOf(AuditCategory.class);
+        if (categories == null)
```
vrozov: Is it necessary to allow `null`? Should it check for an empty collection and throw `NullPointerException`?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: Do you mean `disabledRestCategoriesList.isEmpty() || disabledRestCategoriesList.size() == 1 && "NONE".equalsIgnoreCase(disabledRestCategoriesList.get(0))` ? 

Is it not redundant? as it will go to else and build an empty set.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
sujithvm: I don't think so. As far as this parse is concerned, absence of values in input (whether null || empty) should mean no output categories.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
vrozov: It is not redundant, but, yes, the end result is the same. Question is if you short-circuit call to `parse()` when `disabledRestCategoriesList` is implicitly empty, why is it not short-circuited when it is explicitly empty?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
vrozov: While you may handle `null` the same way as an empty collection, the `parse()` function itself follows "prefer empty collection over null" and converts `null` to an empty set.

sujithvm: I didnt quite follow you. 
If we have an input list of nulls, since we are filtering non-nulls in the stream we would return empty set. 
This is what I meant by absence of values and we are return an empty set. [null, null...] = null => return empty set.

vrozov: What if the collection `categories` itself is `null`? When I referred to "check for an empty collection" I meant instead of `if (categories == null)`, check `if (categories.isEmpty())`.

You brought a good question, how `null` should be handled inside the collection. IMO, that is a programming mistake to pass `null` inside the categories collection.

sujithvm: Are you saying to short-circuit `if (categories == null || categories.isEmpty())`  ?
I mean to say, absence of values whether categories is null or empty or list of only null strings should all mean the same thing -> empty category set.

sujithvm: @vrozov If you could tell exactly the change you are looking for, i could can do that and move on. I don't want to be blocked with such issues.

vrozov: My suggestion is not to allow `null` at all, whether the collection itself is `null` or it contains `null`. Change `if (categores == null)` to `if (categories.isEmpty())` and remove the first `filter`.

PR #306, opened 2020-03-20
AuditLog: Refactor to audit config

### LINKED ISSUE #312: Refactor AuditConfig
Changes:
- Refactor audit log config settings used in AbstractAuditLog into a config class. 
- Refactor usage of AuditConfig used in other places

sujithvm:
Fixes #312 

No new functionality added.


src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,195 @@
...
+    private final boolean excludeSensitiveHeaders;
+    private final List<String> ignoredAuditUsers;
+    private final List<String> ignoredComplianceUsersForRead;
+    private final List<String> ignoredComplianceUsersForWrite;
+    private final List<String> ignoreAuditRequests;
+    private final EnumSet<AuditCategory> disabledRestCategories;
+    private final EnumSet<AuditCategory> disabledTransportCategories;
+    private final String opendistrosecurityIndex;
+
+    public AuditConfig(final boolean restAuditingEnabled,
```
vrozov: I'd recommend using Builder pattern here as `AuditConfig` constructor takes a large number of arguments with the same type.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -118,77 +100,12 @@
...
-
-    private static List<String> getConfigList(final Settings settings,
-                                              final String key,
-                                              final List<String> defaultList) {
-        List<String> list = settings.getAsList(key, defaultList);
-        if (list.size() == 1 && "NONE".equals(list.get(0))) {
-            return Collections.emptyList();
-        }
-        return list;
+        this.log.info("Configured audit settings: {}", auditConfig.toString());
```
vrozov: do not use `this`. Usually, `log` is a class variable, not an instance variable.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,195 @@
...
+                ", ignoredAuditUsers=" + ignoredAuditUsers +
+                ", ignoredComplianceUsersForRead=" + ignoredComplianceUsersForRead +
+                ", ignoredComplianceUsersForWrite=" + ignoredComplianceUsersForWrite +
+                ", ignoreAuditRequests=" + ignoreAuditRequests +
+                ", disabledRestCategories=" + disabledRestCategories +
+                ", disabledTransportCategories=" + disabledTransportCategories +
+                ", opendistrosecurityIndex='" + opendistrosecurityIndex + '\'' +
+                '}';
+    }
+}
```
vrozov: Add new line.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -81,32 +80,15 @@
...
-    private final List<String> ignoreAuditRequests;
-    private final EnumSet<AuditCategory> disabledRestCategories;
-    private final EnumSet<AuditCategory> disabledTransportCategories;
-    private final List<String> defaultIgnoredUsers = Arrays.asList("kibanaserver");
-    private final boolean excludeSensitiveHeaders;
-
-    private final String opendistrosecurityIndex;
-    private static final List<String> writeClasses = new ArrayList<>();
 
+    private final AuditConfig auditConfig;
```
vrozov: I guess it should be `protected` to be consistent with other member variables.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -737,6 +654,7 @@ private boolean checkTransportFilter(final AuditCategory category, final String
             return false;
         }
 
+        final Collection<String>ignoredAuditUsers = auditConfig.getIgnoredAuditUsers();
         if (ignoredAuditUsers.size() > 0 && WildcardMatcher.matchAny(ignoredAuditUsers, effectiveUser)) {
```
vrozov: IMO, `isEmpty()` will be better

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: Will make logger a class static

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,195 @@
...
+            Arrays.asList(AuditCategory.AUTHENTICATED.toString(),
+                    AuditCategory.GRANTED_PRIVILEGES.toString());
+
+    private final boolean restAuditingEnabled;
+    private final boolean transportAuditingEnabled;
+    private final boolean resolveBulkRequests;
+    private final boolean logRequestBody;
+    private final boolean resolveIndices;
+    private final boolean excludeSensitiveHeaders;
+    private final List<String> ignoredAuditUsers;
```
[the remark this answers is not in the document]
sujithvm: Ignore users accepts pattern which may cause an issue with ordering and pattern matching.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: Since log is used in subclasses. Will not make it static as it needs reference to class.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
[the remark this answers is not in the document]
vrozov: If an order is important (I don't think it is, as it is not "match first" or "match last"), use `Set` that preserves order (`LinkedHashSet`) as the implementation.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
sujithvm: I did think about this but I don't see any places where config params will be set independently. Maybe when we need it, we can have it

vrozov: If you don't want to introduce Builder now (I am OK with that), make the constructor `private`.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
vrozov: Please don't use `this`, so that if later `log` is changed to a class variable, it won't be necessary to introduce change here.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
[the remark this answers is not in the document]
sujithvm: This would propagate passing a higher level structure LinkedHashSet. We can't simply use Set type because it ll be important for consumers to know the underlying implementation that the set is ordered.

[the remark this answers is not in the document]
sujithvm: I looked into the current usage and don't think ordering matters so I ll move to Set with HashSet implementation

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -118,81 +99,12 @@
...
-        final List<String> disabledRestCategoriesList = settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_REST_CATEGORIES, DEFAULT_DISABLED_CATEGORIES);
-
-        if (disabledRestCategoriesList.isEmpty() || (disabledRestCategoriesList.size() == 1 && "NONE".equalsIgnoreCase(disabledRestCategoriesList.get(0)))) {
-            disabledRestCategories = EnumSet.noneOf(AuditCategory.class);
-        } else {
-            disabledRestCategories = AuditCategory.parse(disabledRestCategoriesList);
-        }
-
-        if (!disabledRestCategories.isEmpty()) {
-            log.info("Configured categories on rest layer to ignore: {}", disabledRestCategories);
```
sujithvm: Logging entire configuration when configured. Hence removed these info logs.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -80,33 +79,15 @@
...
 
 public abstract class AbstractAuditLog implements AuditLog {
+    protected final Logger log = LogManager.getLogger(this.getClass());
 
-    private static final List<String> DEFAULT_DISABLED_CATEGORIES = Arrays.asList(AuditCategory.AUTHENTICATED.toString(), AuditCategory.GRANTED_PRIVILEGES.toString());
+    private final ThreadPool threadPool;
+    private final IndexNameExpressionResolver resolver;
+    private final ClusterService clusterService;
+    private final Settings settings;
+    private final AuditConfig auditConfig;
```
vrozov: If `auditConfig` is `final` and immutable (all fields are also final) how it can be reloaded dynamically?

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfigTest.java
```diff
@@ -0,0 +1,85 @@
...
+public class AuditConfigTest {
+
+    @Test
+    public void testDefault() {
+        // arrange
+        final Settings settings = Settings.builder().build();
+        final Set<String> defaultIgnoredUser = Collections.singleton("kibanaserver");
+        final EnumSet<AuditCategory> defaultDisabledCategories = EnumSet.of(AUTHENTICATED, GRANTED_PRIVILEGES);
+        // act
+        final AuditConfig auditConfig = AuditConfig.getConfig(settings);
```
vrozov: Use `EMPTY_SETTINGS`.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,190 @@
...
+
+        final EnumSet<AuditCategory> disabledTransportCategories;
+        final List<String> disabledTransportCategoriesList = settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_TRANSPORT_CATEGORIES, DEFAULT_DISABLED_CATEGORIES);
+        if (disabledTransportCategoriesList.isEmpty() || (disabledTransportCategoriesList.size() == 1 && "NONE".equalsIgnoreCase(disabledTransportCategoriesList.get(0)))) {
+            disabledTransportCategories = EnumSet.noneOf(AuditCategory.class);
+        } else {
+            disabledTransportCategories = AuditCategory.parse(disabledTransportCategoriesList);
+        }
+
+        final List<String> ignoredAuditUsers = new ArrayList<>(settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_IGNORE_USERS, DEFAULT_IGNORED_USERS));
```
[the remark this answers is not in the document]
sujithvm: Won't this have a side-effect clearing the function unless getAsList returns a new list always?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -118,81 +99,12 @@
...
-            log.info("Configured Users to ignore for write compliance events: {}", ignoredComplianceUsersForWrite);
-        }
-
-        ignoreAuditRequests = settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_IGNORE_REQUESTS, Collections.emptyList());
-        if (ignoreAuditRequests.size() > 0) {
-            log.info("Configured Requests to ignore: {}", ignoreAuditRequests);
-        }
-
-        this.excludeSensitiveHeaders = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_EXCLUDE_SENSITIVE_HEADERS, true);
+        log.info("Configured audit settings: {}", auditConfig.toString());
```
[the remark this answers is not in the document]
sujithvm: What would you suggest?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: Yes this needs to be non-final when reloaded dynamically. I thought of doing it in #319 when the change is actually getting merged. Till then we don't need it. I feel having rigidity helps in making refactors less error-prone. Thats why I put this as final. What do you think?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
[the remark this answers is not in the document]
vrozov: There is no need to clear the original collection, use empty collection.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,190 @@
...
+        final boolean resolveBulkRequests = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_RESOLVE_BULK_REQUESTS, false);
+        final boolean logRequestBody = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_LOG_REQUEST_BODY, true);
+        final boolean resolveIndices = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_RESOLVE_INDICES, true);
+        final boolean excludeSensitiveHeaders = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_EXCLUDE_SENSITIVE_HEADERS, true);
+        final String opendistrosecurityIndex = settings.get(ConfigConstants.OPENDISTRO_SECURITY_CONFIG_INDEX_NAME, ConfigConstants.OPENDISTRO_SECURITY_DEFAULT_CONFIG_INDEX);
+
+        final EnumSet<AuditCategory> disabledRestCategories;
+        final List<String> disabledRestCategoriesList = settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_REST_CATEGORIES, DEFAULT_DISABLED_CATEGORIES);
+        if (disabledRestCategoriesList.isEmpty() || (disabledRestCategoriesList.size() == 1 && "NONE".equalsIgnoreCase(disabledRestCategoriesList.get(0)))) {
```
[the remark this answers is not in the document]
sujithvm: I would appreciate if you could help design this helper function. Previously I had 

```
private static List<String> getSettingAsList(final Settings settings, final String key, final List<String> defaultList) {
        final List<String> list = settings.getAsList(key, defaultList);
        if (list.size() == 1 && "NONE".equals(list.get(0))) {
        }
        return list;
    }
```

Looks like other params consider only "NONE" as implicitly empty (none, None are not considered). Do you want to retain this behavior or clear it in #325

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,190 @@
...
+    private final EnumSet<AuditCategory> disabledTransportCategories;
+    private final String opendistrosecurityIndex;
+
+    private AuditConfig(final boolean isRestApiAuditEnabled,
+                        final boolean isTransportAuditEnabled,
+                        final boolean resolveBulkRequests,
+                        final boolean logRequestBody,
+                        final boolean resolveIndices,
+                        final boolean excludeSensitiveHeaders,
+                        final Set<String> ignoredAuditUsers,
```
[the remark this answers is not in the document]
sujithvm: I feel whoever is calling the constructor should deal with that. In this case, getConfig(settings). What if "NONE" is one of the users? It would be wrong to handle this in constructor. What are your thoughts?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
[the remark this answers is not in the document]
vrozov: I guess it will be good to log (one settings per log entry) only those settings that are different from the default, especially for boolean flags.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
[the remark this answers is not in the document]
vrozov: OK, let's assume that "NONE" is a valid input in the constructor and it means an empty list only in settings.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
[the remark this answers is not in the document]
vrozov: One way is to pass a flag whether "NONE" is case sensitive or not. Assuming that "NONE" will be deprecated soon, I suggest retaining the current behavior.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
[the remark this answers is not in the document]
sujithvm: Original logging was done only for lists and not for boolean. So I thought, it ll be better to log all. 
What do you think of prettifying toString() to something more descriptive ?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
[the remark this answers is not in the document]
sujithvm: Alright!
Sadly, I feel case in-sensitive handling of none was accidental in categories because of making it uppercase for enums and now have to deal with complexity.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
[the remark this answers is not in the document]
vrozov: It is an option but it will most likely mess up Intellij debugger. I'd recommend avoiding `toString()`.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,206 @@
...
+                ignoredComplianceUsersForWrite,
+                ignoreAuditRequests,
+                disabledRestCategories,
+                disabledTransportCategories,
+                opendistrosecurityIndex);
+    }
+
+    private static List<String> getSettingAsList(final Settings settings, final String key, final List<String> defaultList, final boolean ignoreCaseForNone) {
+        final List<String> list = settings.getAsList(key, defaultList);
+        if (list.size() == 1) {
```
vrozov: nit: consider 
```
if (list.size() == 1 && "NONE".equals(ignoreCaseForNone? list.get(0).toUpperCase() : list.get(0)))
```

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,206 @@
...
+        this.ignoredAuditUsers = ignoredAuditUsers;
+        this.ignoredComplianceUsersForRead = ignoredComplianceUsersForRead;
+        this.ignoredComplianceUsersForWrite = ignoredComplianceUsersForWrite;
+        this.ignoreAuditRequests = ignoredAuditRequests;
+        this.disabledRestCategories = disabledRestCategories;
+        this.disabledTransportCategories = disabledTransportCategories;
+        this.opendistrosecurityIndex = opendistrosecurityIndex;
+    }
+
+    public static AuditConfig getConfig(Settings settings) {
```
vrozov: Consider renaming `getConfig()` to `from()`.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -118,81 +99,18 @@
...
-
-        ignoreAuditRequests = settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_IGNORE_REQUESTS, Collections.emptyList());
-        if (ignoreAuditRequests.size() > 0) {
-            log.info("Configured Requests to ignore: {}", ignoreAuditRequests);
-        }
-
-        this.excludeSensitiveHeaders = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_EXCLUDE_SENSITIVE_HEADERS, true);
+        this.auditConfig = AuditConfig.getConfig(settings);
+
+        log.info("Configured audit setting for auditing on rest layer : {}", auditConfig.isRestApiAuditEnabled());
```
vrozov: Consider
```
log.info("Auditing on REST API is {}", auditConfig.isRestApiAuditEnabled()? "enabled" : "disabled");
```

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -118,81 +99,18 @@
...
+        log.info("Configured audit setting for auditing on transport layer : {}", auditConfig.isTransportApiAuditEnabled());
+        log.info("Configured audit setting to log request body : {}", auditConfig.shouldLogRequestBody());
+        log.info("Configured audit setting to resolve bulk requests : {}", auditConfig.shouldResolveBulkRequests());
+        log.info("Configured audit setting to resolve indices : {}", auditConfig.shouldResolveIndices());
+        log.info("Configured audit setting to exclude sensitive headers : {}", auditConfig.shouldExcludeSensitiveHeaders());
+        log.info("All configured audit settings : {}", auditConfig.toString());
```
vrozov: remove this line

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,206 @@
...
+    public EnumSet<AuditCategory> getDisabledTransportCategories() {
+        return disabledTransportCategories;
+    }
+
+    public String getOpendistrosecurityIndex() {
+        return opendistrosecurityIndex;
+    }
+
+    @Override
+    public String toString() {
```
vrozov: Instead of `toString()` I recommend delegating logging to `AuditConfig` class itself: `public void log(Logger log)`.

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfigTest.java
```diff
@@ -0,0 +1,104 @@
...
+        assertEquals(auditConfig.getIgnoredAuditRequests(), Collections.singleton("test-request"));
+        assertEquals(auditConfig.getIgnoredComplianceUsersForRead(), ImmutableSet.of("test-user-1", "test-user-2"));
+        assertEquals(auditConfig.getIgnoredComplianceUsersForWrite(), ImmutableSet.of("test-user-3", "test-user-4"));
+        assertEquals(auditConfig.getDisabledRestCategories(), EnumSet.of(BAD_HEADERS, SSL_EXCEPTION));
+        assertEquals(auditConfig.getDisabledTransportCategories(), EnumSet.of(FAILED_LOGIN, MISSING_PRIVILEGES));
+        assertEquals("test-index", auditConfig.getOpendistrosecurityIndex());
+    }
+
+    @Test
+    public void testNone() {
```
vrozov: Please also add test for empty.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: I would have personally preferred this to see all configured settings. 
Why are you suggesting to remove this ?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
sujithvm: Maybe make a new issue to add this pattern to other classes as well? 

I feel the correct way of doing this is make a new interface Loggable for log and then implement it. Also should be able to configure the log level.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
vrozov: Is it redundant?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
vrozov: I don't see a use case for `Loggable` interface. I think that INFO is the most appropriate level for logging AuditConfig. For now, I would not extend the pattern to other classes.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -118,81 +99,18 @@
...
-        if(ignoredComplianceUsersForRead.size() == 1 && "NONE".equals(ignoredComplianceUsersForRead.get(0))) {
-            ignoredComplianceUsersForRead.clear();
-        }
-
-        if (ignoredComplianceUsersForRead.size() > 0) {
-            log.info("Configured Users to ignore for read compliance events: {}", ignoredComplianceUsersForRead);
-        }
+        log.info("Auditing on REST API is {}", auditConfig.isRestApiAuditEnabled() ? "enabled" : "disabled");
+        log.info("Auditing on Transport API is {}", auditConfig.isTransportApiAuditEnabled() ? "enabled" : "disabled");
+        log.info("Configured audit setting to log request body : {}", auditConfig.shouldLogRequestBody());
```
vrozov: Please apply the same pattern: "Auditing of request body is {}"...

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -118,81 +99,18 @@
...
-        ignoredComplianceUsersForRead = new ArrayList<>(settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_READ_IGNORE_USERS, defaultIgnoredUsers));
-
-        if(ignoredComplianceUsersForRead.size() == 1 && "NONE".equals(ignoredComplianceUsersForRead.get(0))) {
-            ignoredComplianceUsersForRead.clear();
-        }
-
-        if (ignoredComplianceUsersForRead.size() > 0) {
-            log.info("Configured Users to ignore for read compliance events: {}", ignoredComplianceUsersForRead);
-        }
+        log.info("Auditing on REST API is {}", auditConfig.isRestApiAuditEnabled() ? "enabled" : "disabled");
```
vrozov: Consider adding excluded REST API audit categories here in case `auditConfig.isRestApiAuditEnabled()` is `true`. The same for Transport API. I don't see them being logged.

sujithvm: This was supposed to be done via bulk logging of toString()

sujithvm: I will log every param. 
Even is rest or transport api is disabled some categories are excluded. https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java#L718

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -80,33 +79,15 @@
 import com.google.common.io.BaseEncoding;
 
 public abstract class AbstractAuditLog implements AuditLog {
+    protected final Logger logger = LogManager.getLogger(this.getClass());
```
vrozov: Sorry if I missed that rename in the prior reviews. Can you please rename it back to `log` as other classes mostly use `log` for the `Logger`.

sujithvm: `logger` seems more appropriate than `log`

vrozov: I don't have a strong preference of `log` over `logger`, just want to be consistent. For me, both `log` and `logger` are fine.

As the majority of other classes use `log` and the change is not part of `AuditConfig` refactoring, let's keep it as `log` in this PR/commit.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,214 @@
...
+        logger.info("Auditing of request body is {}.", logRequestBody ? "enabled" : "disabled");
+        logger.info("Bulk requests resolution is {} during request auditing.", resolveBulkRequests ? "enabled" : "disabled");
+        logger.info("Index resolution is {} during request auditing.", resolveIndices ? "enabled" : "disabled");
+        logger.info("Sensitive headers auditing is {}.", excludeSensitiveHeaders ? "enabled" : "disabled");
+        logger.info("Auditing requests from {} users is disabled.", ignoredAuditUsers);
+        logger.info("Compliance read operation requests auditing from {} users is disabled.", ignoredComplianceUsersForRead);
+        logger.info("Compliance write operation requests auditing from {} users is disabled.", ignoredComplianceUsersForWrite);
+        logger.info("{} are excluded from REST API auditing.", disabledRestCategories);
+        logger.info("{} are excluded from Transport API auditing.", disabledTransportCategories);
+        logger.info("Open distro auditing uses {} index(alias?) to write auditing events.", opendistrosecurityIndex);
```
vrozov: My bad for not being clear. I was not sure if `opendistrosecurityIndex` can be an alias instead of an index, so I put "alias?". If it can be an alias, please use "index or alias", otherwise just "index".

sujithvm: It uses that index to monitor change events for internal configuration.

sujithvm:
@vrozov Apologies I had push new changes. Opendistro index should not be part of AuditConfig and cannot be reloaded/changed. It is an internal index that is maintained by the plugin and as far as auditing is concerned, it should only monitor write/read events done to that. Reverting that back to what was originially there.

vrozov:
I don't see why the index can't be changed. It is configurable in the settings and is not hardcoded. Even if changing the index won't be allowed, what is wrong with getting it from `AuditConfig`? I assume that `AuditConfig` will support changing one or more parameters at a time and `AuditConfig` will encapsulate the logic of which parameters support dynamic reloading and which one not.

sujithvm:
It is currently configured in elasticsearch.yml and changing it might warrant a cluster restart. 
I removed it from AuditConfig simply because it is not an audit configuration parameter. 
I feel encapsulating unrelated parameters and defining whats reloadable and whats not as part of audit configuration will unnecessarily make it more complex. 
What are your thoughts?

vrozov:
By looking at the code it is hard to justify moving index out of `AuditConfig`. Once the functionality evolves and it is clear that the index needs to be handled differently, we can move it out of `AuditConfig`. For now, I'd suggest keeping it as part of `AuditConfig`.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,204 @@
...
+        logger.info("Auditing on REST API is {}.", isRestApiAuditEnabled ? "enabled" : "disabled");
+        logger.info("Auditing on Transport API is {}.", isTransportApiAuditEnabled ? "enabled" : "disabled");
+        logger.info("Auditing of request body is {}.", logRequestBody ? "enabled" : "disabled");
+        logger.info("Bulk requests resolution is {} during request auditing.", resolveBulkRequests ? "enabled" : "disabled");
+        logger.info("Index resolution is {} during request auditing.", resolveIndices ? "enabled" : "disabled");
+        logger.info("Sensitive headers auditing is {}.", excludeSensitiveHeaders ? "enabled" : "disabled");
+        logger.info("Auditing requests from {} users is disabled.", ignoredAuditUsers);
+        logger.info("Compliance read operation requests auditing from {} users is disabled.", ignoredComplianceUsersForRead);
+        logger.info("Compliance write operation requests auditing from {} users is disabled.", ignoredComplianceUsersForWrite);
+        logger.info("{} are excluded from REST API auditing.", disabledRestCategories);
```
vrozov: Assuming that you will make another round of changes, can you move this to line 175?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,204 @@
...
+        logger.info("{} are excluded from Transport API auditing.", disabledTransportCategories);
```
vrozov: and this one to line 176.

sujithvm:
Opendistro security index is an index that the plugin uses to store all the plugin data. Please check its usage. It should not be part of audit config. I don't see a reason why this should be moved into audit config and moved out later.

vrozov:
I consider it to be an index where audit logging will store audit events. Currently, there is a single index where audit events are stored and where security plugin stores other information, but can't that change in the future? Or my understanding of that setting is wrong?

sujithvm:
OPENDISTRO_SECURITY_CONFIG_INDEX_NAME (opendistro_security.config_index_name, default `.opendistro_security`) is where plugin data will be stored. 

OPENDISTRO_SECURITY_AUDIT_ES_INDEX(index, default `security-auditlog-'YYYY.MM.dd'`) is where audit logs will be changed if destination is set to internal elasticsearch.

vrozov:
Should OPENDISTRO_SECURITY_AUDIT_ES_INDEX be part of `AuditConfig`?

sujithvm:
It wont be part of AuditConfig. It may be part of AuditDestinationConfig (or something like that) if we are doing it in the future. It is not necessary to have elasticsearch as destination. Audit logs can be written to file or some cloud logging streams.

vrozov:
I think that multiple audit config classes are not necessary and are not desirable. It will be good to have a single class that handles all settings related to audit. Some of them may be static and some may be dynamic.

sujithvm:
There are two parts to consider.

1) What to audit and settings around that (AuditConfig + ComplianceConfig which will be merged with AuditConfig in #313 ) 

2) Where to push the audit messages and settings around (Destinations - file writer settings, cloud watch settings, kafka settings, ES settings)   

It may actually be meaningful to keep them separate. 
In this case this index where messages are stored come under second category hence not interested in storing that in 1st audit config category

vrozov:
`AuditConfig` name suggests that it is not only a configuration for "what" to audit, but also "where". Please rename so it is clear that it only covers "what" part. I'd consider making `AuditConfig` a top level class with a static class that is responsible for "what" as an inner class. Later, more static classes can be added to it.

sujithvm:
Having static inner classes is reasonable

sujithvm:
@vrozov Do you have anymore suggested changes here?

vrozov:
Please see my comments regarding changing the order of logging, otherwise introducing an inner class to be responsible for "what" settings is the only change.

sujithvm:
@vrozov Order of logging is already changed. Is something missing?
What is the outer and inner class name structure your prefer?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        public static Filter from(Settings settings) {
+            final boolean isRestApiAuditEnabled = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_REST, true);
+            final boolean isTransportAuditEnabled = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_TRANSPORT, true);
+            final boolean resolveBulkRequests = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_RESOLVE_BULK_REQUESTS, false);
+            final boolean logRequestBody = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_LOG_REQUEST_BODY, true);
+            final boolean resolveIndices = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_RESOLVE_INDICES, true);
+            final boolean excludeSensitiveHeaders = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_EXCLUDE_SENSITIVE_HEADERS, true);
+            final String opendistrosecurityIndex = settings.get(ConfigConstants.OPENDISTRO_SECURITY_CONFIG_INDEX_NAME, ConfigConstants.OPENDISTRO_SECURITY_DEFAULT_CONFIG_INDEX);
```
vrozov, line 68: `opendistrosecurityIndex` is not used.

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfigFilterTest.java
```diff
@@ -0,0 +1,126 @@
...
+import com.amazon.opendistroforelasticsearch.security.support.ConfigConstants;
+import com.google.common.collect.ImmutableSet;
+import org.elasticsearch.common.settings.Settings;
+import org.junit.Test;
+
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.Set;
+
+import static com.amazon.opendistroforelasticsearch.security.auditlog.impl.AuditCategory.*;
```
debjanibnrj, line 13: wildcard imports are generally considered bad practice.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,205 @@
...
+            this.ignoredComplianceUsersForWrite = ignoredComplianceUsersForWrite;
+            this.ignoreAuditRequests = ignoredAuditRequests;
+            this.disabledRestCategories = disabledRestCategories;
+            this.disabledTransportCategories = disabledTransportCategories;
+        }
+
+        public static Filter from(Settings settings) {
+            final boolean isRestApiAuditEnabled = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_REST, true);
+            final boolean isTransportAuditEnabled = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_TRANSPORT, true);
+            final boolean resolveBulkRequests = settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_RESOLVE_BULK_REQUESTS, false);
```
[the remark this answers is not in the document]
sujithvm: Retained the existing default. 
Resolving bulk requests results in additional computation hence user must intentionally set to true.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -80,33 +79,16 @@
 import com.google.common.io.BaseEncoding;
 
 public abstract class AbstractAuditLog implements AuditLog {
-
-    private static final List<String> DEFAULT_DISABLED_CATEGORIES = Arrays.asList(AuditCategory.AUTHENTICATED.toString(), AuditCategory.GRANTED_PRIVILEGES.toString());
-
     protected final Logger log = LogManager.getLogger(this.getClass());
```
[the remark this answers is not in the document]
sujithvm: I had renamed this to logger. @vrozov wanted it renamed back to log

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfigFilterTest.java
sujithvm: I assumed it is generally fine to have wildcard imports for constants/values. 
Removing this would add 10 other import lines.

debjanibnrj: Seeing as your using most audit categories for the tests this is a fair point. For `import static org.junit.Assert.*;` it may be better to specify the specific imports.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,205 @@
...
+                    excludeSensitiveHeaders,
+                    ignoredAuditUsers,
+                    ignoredComplianceUsersForRead,
+                    ignoredComplianceUsersForWrite,
+                    ignoreAuditRequests,
+                    disabledRestCategories,
+                    disabledTransportCategories);
+        }
+
+        private static List<String> getSettingAsList(final Settings settings, final String key, final List<String> defaultList, final boolean ignoreCaseForNone) {
```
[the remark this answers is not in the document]
sujithvm: Sounds good!

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfigFilterTest.java
sujithvm: Should probably write a checkstyle rule for this

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,205 @@
...
+import org.apache.logging.log4j.Logger;
+import org.elasticsearch.common.settings.Settings;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Set;
+
+public class AuditConfig {
```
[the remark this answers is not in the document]
vrozov: Add `private` constructor for `AuditConfig` to avoid class misuse. It is a container for other classes inside `AuditConfig`.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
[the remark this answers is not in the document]
vrozov: @debjanibnrj It is more consistent with how `Logger` is named in other places and avoids unnecessary diffs.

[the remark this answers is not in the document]
sujithvm: @vrozov Fair point with reducing diffs but `auditConfigFilter.log(log)` would be confusing to anyone especially because this is around audit "logging"

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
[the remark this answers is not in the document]
vrozov: Converting the output to a set has an additional cost when it is passed to `AuditCategory.parse()` as it will be constructing a different set and checking for uniqueness again (not a big deal).

sujithvm:
@debjanibnrj Could you please review this again. I missed adding private constructor

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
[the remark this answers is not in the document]
vrozov: IMO, it is a problem with method name, not `Logger` name. Rename `AuditConfig.Filter.log()`.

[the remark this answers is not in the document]
sujithvm: It is not a static function if thats what you mean.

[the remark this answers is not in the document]
vrozov: I mean that the minor readability problem is caused by `AuditConfig.Filter` class method name `log()`. It is not important whether it is static or not.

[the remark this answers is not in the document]
sujithvm: If you want it renamed to something specific I will rename as you had previously suggested to use log() otherwise let's leave this. I m fine with this.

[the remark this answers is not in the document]
vrozov: I am fine the current function name or with a different function name if you want to add readability. I strongly prefer `log` not to be renamed to `logger`.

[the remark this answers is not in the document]
sujithvm: @vrozov Can you approve this PR in that case ?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,268 @@
...
+                    ignoredComplianceUsersForRead,
+                    ignoredComplianceUsersForWrite,
+                    ignoreAuditRequests,
+                    disabledRestCategories,
+                    disabledTransportCategories);
+        }
+
+        private static Set<String> getSettingAsSet(final Settings settings, final String key, final List<String> defaultList, final boolean ignoreCaseForNone) {
+            final List<String> list = settings.getAsList(key, defaultList);
+            if (list.size() == 1 && "NONE".equals(ignoreCaseForNone? list.get(0).toUpperCase() : list.get(0))) {
```
vrozov, line 134: In case conversion to `Set` is done inside this function, it is better to add `isEmpty()` check here. I am not 100% convinced that it is necessary to convert to `Set` from `List` here as it is not always necessary and in the future there may be different requirements for case-sensitivity of the returned `Set`.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -0,0 +1,268 @@
...
+                    excludeSensitiveHeaders,
+                    ignoredAuditUsers,
+                    ignoredComplianceUsersForRead,
+                    ignoredComplianceUsersForWrite,
+                    ignoreAuditRequests,
+                    disabledRestCategories,
+                    disabledTransportCategories);
+        }
+
+        private static Set<String> getSettingAsSet(final Settings settings, final String key, final List<String> defaultList, final boolean ignoreCaseForNone) {
```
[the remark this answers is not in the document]
sujithvm: Will move it when needed as this has more restricted scope.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
sujithvm: Will change it when needed in the future.

PR #309, opened 2020-03-23
[Optimization] Memoize results of resolveIndexPatterns for Bulk requests

DmitryOlshansky:
Introduce per request resolve mask, this allows us to bypass the whole resolveIndexPatterns+builder.add combo
if multiple writes in one request go to the same alias/index. This is the common case in bulk indexing.

*Issue #, if available:*

*Description of changes:*

Split off from a bigger PR focused on solving perf problem with indexing to long chians of indices with a write alias to one of them.

https://github.com/opendistro-for-elasticsearch/security/pull/258

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
@@ -159,7 +159,7 @@ public final class IndexResolverReplacer implements DCFListener {
         return false;
     }
 
-    private Resolved resolveIndexPatterns(final IndicesOptions indicesOptions, final boolean isSearchOrFieldsCapabilities, final String... requestedPatterns0) {
+    private Resolved resolveIndexPatterns(final IndicesOptions indicesOptions, final boolean enableCrossClusterResolution, final String... requestedPatterns0) {
 
         if(log.isTraceEnabled()) {
             log.trace("resolve requestedPatterns: "+ Arrays.toString(requestedPatterns0));
@@ -177,7 +177,7 @@ public final class IndexResolverReplacer implements DCFListener {
 
         final RemoteClusterService remoteClusterService = OpenDistroSecurityPlugin.GuiceHolder.getRemoteClusterService();
 
-        if(remoteClusterService.isCrossClusterSearchEnabled() && isSearchOrFieldsCapabilities) {
+        if(remoteClusterService.isCrossClusterSearchEnabled() && enableCrossClusterResolution) {
             remoteIndices = new HashSet<>();
             final Map<String, OriginalIndices> remoteClusterIndices = OpenDistroSecurityPlugin.GuiceHolder.getRemoteClusterService()
                     .groupIndices(indicesOptions, requestedPatterns0, idx -> resolver.hasIndexOrAlias(idx, clusterService.state()));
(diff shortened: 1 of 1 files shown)


DmitryOlshansky:
spinoff of https://github.com/opendistro-for-elasticsearch/security/pull/258

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -161,7 +159,7 @@ private static final boolean isLocalAll(final String... requestedPatterns) {
         return false;
     }
 
-    private Resolved resolveIndexPatterns(final IndicesOptions indicesOptions, final Object request, final String... requestedPatterns0) {
+    private Resolved resolveIndexPatterns(final IndicesOptions indicesOptions, final boolean isSearchOrFieldsCapabilities, final String... requestedPatterns0) {
```
dinusX: The only comment that is related to new variable name. Not very intuitive.
I was thinking initially to renamed it to "isSearchOrFieldsCapabilitiesRequest", but Vlad suggested a better one "enableCrossClusterResolution", since this variable is basically indicating if we need to do cross-cluster resolution based on type of the request.

DmitryOlshansky:
> Looking at the code, it seems that even for simple _bulk and _msearch requests with a high number of operations on the same index, this can provide a performance improvement. Unfortunately, it's not clear how much. Probably we'll need to run some performance tests with high _bulk ingestion to see the difference.

Aside from synthetic stuff I described you can always just run esrally from pip.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
DmitryOlshansky: Done, sounds  better for sure

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -316,28 +313,59 @@ public boolean replace(final TransportRequest request, boolean retainMode, Strin
...
-            public String[] provide(String[] original, Object localRequest, boolean supportsReplace) {
-                final IndicesOptions indicesOptions = indicesOptionsFrom(localRequest);
-                final Resolved iResolved = resolveIndexPatterns(indicesOptions, localRequest, original);
+        // set of previously resolved index requests to avoid resolving
+        // the same index more than once while processing bulk requests
+        final Set<IndexResolveKey> alreadyResolved = new HashSet<>();
+
+        getOrReplaceAllIndices(request, (original, localRequest, supportsReplace) -> {
+            final IndicesOptions indicesOptions = indicesOptionsFrom(localRequest);
+            final boolean isSearchOrFieldCapabilities = localRequest instanceof FieldCapabilitiesRequest || localRequest instanceof SearchRequest;
```
vrozov: Either inline or rename to `enableCrossClusterResolution`.

DmitryOlshansky: Yeah, missed that one - fixed!

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -316,28 +313,59 @@ public boolean replace(final TransportRequest request, boolean retainMode, Strin
         }, false);
     }
 
+    private static final class IndexResolveKey {
+        private final IndicesOptions opts;
+        private final boolean enableCrossClusterResolution;
+        private final String[] original;
+        public IndexResolveKey(IndicesOptions opts, boolean isSearchOrFieldCapabilities, String[] original) {
```
vrozov: one more occurrence :)

DmitryOlshansky: renaming is hard ;)
Done

PR #333, opened 2020-03-30
Fix updating password hash without replacing the entry

sujithvm:
Bug:
- Updating user hash was replacing the entire entry causing it to lose the roles attached to it directly from internal users api.
- Not returning bad request when both password and hash are not provided or empty

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AccountApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AccountApiAction.java
@@ -199,26 +199,27 @@ public class AccountApiAction extends AbstractApiAction {
         final SecurityJsonNode securityJsonNode = new SecurityJsonNode(content);
         final String currentPassword = content.get("current_password").asText();
         final Hashed internalUserEntry = (Hashed) internalUser.getCEntry(username);
-        final String hash = internalUserEntry.getHash();
+        final String currentHash = internalUserEntry.getHash();
 
-        if (hash == null || !OpenBSDBCrypt.checkPassword(hash, currentPassword.toCharArray())) {
+        if (currentHash == null || !OpenBSDBCrypt.checkPassword(currentHash, currentPassword.toCharArray())) {
             badRequestResponse(channel, "Could not validate your current password.");
             return;
         }
 
         // if password is set, it takes precedence over hash
-        final String providedPassword = securityJsonNode.get("password").asString();
-        final String providedHash = securityJsonNode.get("hash").asString();
-        if (Strings.isNullOrEmpty(providedPassword) && Strings.isNullOrEmpty(providedHash)) {
+        final String password = securityJsonNode.get("password").asString();
+        final String hash;
+        if (Strings.isNullOrEmpty(password)) {
+            hash = securityJsonNode.get("hash").asString();
+        } else {
+            hash = hash(password.toCharArray());
+        }
+        if (Strings.isNullOrEmpty(hash)) {
             badRequestResponse(channel, "Both provided password and hash cannot be null/empty.");
             return;
         }
 
-        final String newHash = Strings.isNullOrEmpty(providedPassword)
-                ? providedHash
-                : hash(providedPassword.toCharArray());
-
-        internalUserEntry.setHash(newHash);
+        internalUserEntry.setHash(hash);
 
         saveAnUpdateConfigs(client, request, CType.INTERNALUSERS, internalUser, new OnSucessActionListener<IndexResponse>(channel) {
             @Override
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AccountApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AccountApiAction.java
@@ -35,6 +35,7 @@ import org.bouncycastle.crypto.generators.OpenBSDBCrypt;
 import org.elasticsearch.action.index.IndexResponse;
 import org.elasticsearch.client.Client;
 import org.elasticsearch.cluster.service.ClusterService;
+import org.elasticsearch.common.Strings;
 import org.elasticsearch.common.bytes.BytesReference;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.common.transport.TransportAddress;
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AccountApiAction.java
```diff
@@ -197,32 +198,27 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
 
         final SecurityJsonNode securityJsonNode = new SecurityJsonNode(content);
         final String currentPassword = content.get("current_password").asText();
-        final String hash = ((Hashed) internalUser.getCEntry(username)).getHash();
+        final Hashed internalUserEntry = (Hashed) internalUser.getCEntry(username);
+        final String hash = internalUserEntry.getHash();
```
vrozov: rename `hash` to `currentHash` to be consistent with `currentPassword`

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AccountApiAction.java
```diff
@@ -197,32 +198,27 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
...
-        final String plainTextPassword = securityJsonNode.get("password").asString();
-        final String origHash = securityJsonNode.get("hash").asString();
-        if (plainTextPassword != null && plainTextPassword.length() > 0) {
-            contentAsNode.remove("password");
-            contentAsNode.put("hash", hash(plainTextPassword.toCharArray()));
-        } else if (origHash != null && origHash.length() > 0) {
-            contentAsNode.remove("password");
-        } else if (plainTextPassword != null && plainTextPassword.isEmpty() && origHash == null) {
-            contentAsNode.remove("password");
+        final String providedPassword = securityJsonNode.get("password").asString();
```
vrozov:
```
        // if password is set, it takes precedence over hash
        final String password = securityJsonNode.get("password").asString();
        final String hash;
        if (Strings.isNullOrEmpty(password)) {
            hash = securityJsonNode.get("hash").asString();
        } else {
            hash = hash(password.toCharArray());
        }
        if (Strings.isNullOrEmpty(hash)) {
            badRequestResponse(channel, "Both provided password and hash cannot be null/empty.");
            return;
        }

        internalUserEntry.setHash(hash);
```

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AccountApiTest.java
```diff
@@ -94,6 +96,16 @@ public void testPutAccount() throws Exception {
         response = rh.executePutRequest(ENDPOINT, payload, encodeBasicHeader(testUser, testPass));
         assertEquals(HttpStatus.SC_BAD_REQUEST, response.getStatusCode());
 
+        // test - bad request as hash/password is missing
+        payload = "{\"current_password\":\"" + testPass + "\"}";
+        response = rh.executePutRequest(ENDPOINT, payload, encodeBasicHeader(testUser, testPass));
+        assertEquals(HttpStatus.SC_BAD_REQUEST, response.getStatusCode());
+
+        // test - bad request as hash/password is empty
```
vrozov: test with empty hash?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AccountValidator.java
```diff
@@ -27,5 +27,7 @@ public AccountValidator(RestRequest request, BytesReference ref, Settings esSett
         super(request, ref, esSettings, param);
         allowedKeys.put("current_password", DataType.STRING);
         mandatoryKeys.add("current_password");
+        mandatoryOrKeys.add("hash");
```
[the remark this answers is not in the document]
sujithvm: I did consider this. Looks like InternalUsersValidator also inherits from the same and for that it is optional as some of the internal user apis are designed that way. Putting it there would make it mandatory for internal user apis. 
Hence I put it in AccountValidator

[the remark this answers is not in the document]
vrozov: I would prefer to remove the check especially that it is present in `AccountApiAction` with probably a better error message if the validation does not pass.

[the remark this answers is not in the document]
sujithvm: This seems to consistent with the other APIs. If the mandatory keys are absent, then it throws a bad request with invalid configuration error.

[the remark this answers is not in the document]
vrozov: Why is it necessary to validate it twice? If you want to rely on `AccountValidator`, remove the check from `AccountApiAction` (line 217). I'd prefer the check to be done in `AccountApiAction`.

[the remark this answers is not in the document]
sujithvm: If you pass 
{
"password": "",
"hash": ""
}
then the validator check wont fail as the params exist. 
But it fails in the account api action because they are empty/null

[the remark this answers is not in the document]
vrozov: OK, so `AccountValidator` can't even invalidate all cases.

hardik-k-shah: Does this mean user always need to pass key "hash": "" even if user just want to pass string password or vice-versa ?

vrozov: No, only hash or password (or any other key present in `mandatoryOrKeys` should be present. I think that value of `mandatoryOrKey` is minimal as parent or child class can add another value to it and the validation won't work anymore. I'd suggest to remove it.

sujithvm: That is correct. One of the mandatory keys must be provided (hash or password) and other can be absent. Other APIs like role_mapping do the same. 

I agree that it is not very flexible but works for this use-case. 
I can remove it but code might fail at a much later stage only while comparing for hash and password.

vrozov: I suggest to remove the code and deprecate `mandatoryOrKeys`. It is easy to misuse and break validation accidentally. As there is already code that checks that either "password" or "hash" is present, this code is not necessary.

sujithvm: This should probably be a separate issue and propose a better validation structure to use in place of this. 

I do understand that this is not flexible when it comes to `mandatory or keys` in multiple groups eg: (f1 or f2) and (f3 or f4) but at the same time, this is not enforced. 

But before we deprecate and remove this, we should have something that will 
- define the schema of api params instead of digging through the action code.
- fail at early stage (at the beginning of processing request)
- have more rigid bad request validation done here instead of in action code which is more likely to be missed (error-prone). 

I strongly feel we deprecate and remove this only when we have better structure in place as this does it basic job (not saying it does the above perfectly).
 
If you insist I will remove this to unblock the PR but would like to hear your thoughts.

vrozov: I don't suggest to deprecate `mandatoryOrKeys` in this PR. My proposal is to avoid introducing a new dependency on it. As there is only one more `RolesMappingValidator` validator that depends on `mandatoryOrKeys`, there is not much need to make it consistent with other validators.

Later we can decide whether it is worth extending `mandatoryOrKeys` or not. IMO, two cases where it may be useful do not provide enough justification to make it more generic especially that there is still logic in `AccountApiAction` to deal with `password` and `hash`.

I do not insist, but strongly recommend to remove the dependency on `mandatoryOrKeys` from the `AccountValidator`.

sujithvm: I absolutely don't mind removing it from the PR but I would like to know the following which is not clear to me:
- why are we deprecating something that works?
- what if the new improved solution is based out of this ? 
- what is the pain point of this in the use cases we have?

vrozov: >why are we deprecating something that works?

While it works, it is fragile and is not flexible.

>what if the new improved solution is based out of this?

It depends on the effort and the value the new improved solution provides. If the current functionality can be easily extended (I doubt), we may decide to extend it instead of deprecating

>what is the pain point of this in the use cases we have

A duplicate check that has to be introduced anyway

hardik-k-shah:
Thanks @sujithvm for fixing this and @vrozov for reviewing it. 
I am happy to see healthy discussion on GitHub directly. So, community gets benefit as well.

PR #334, opened 2020-03-31
Upgrade checkout action to v2 to fix "fatal: reference is not a tree" issue

vrozov:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
         java-version: 11.0.x
 
     - name: Checkout security
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
 
     - name: Checkstyle
       run: mvn checkstyle:checkstyle
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -17,7 +17,7 @@ jobs:
         java-version: 11.0.x
 
     - name: Checkout security
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
 
     - name: Build
       run: |


.github/workflows/ci.yml
```diff
@@ -12,24 +12,39 @@ on:
...
+
     - name: Checkout security
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
 
     - name: Checkstyle
       run: mvn checkstyle:checkstyle
-      
+
+    - name: prepare openssl
```
sujithvm: I think we are moving away from openssl and using java ssl. 
@debjanibnrj what is the openssl status?

debjanibnrj: only for 7.4+, other branches still support it

sujithvm: Okay. So we don't need this change for master and only for older branches. Is that correct ?

debjanibnrj: Yes, any branch before 7.4 will still need is. 7.4+ will default to using JDK SSL

.github/workflows/ci.yml
```diff
@@ -12,24 +12,39 @@ on:
...
+      run: |
+        mkdir -p tmp/build-openssl && cd tmp/build-openssl
+        curl -OL https://ftp.openssl.org/source/openssl-1.0.1u.tar.gz
```
sujithvm: Also the workflow runs on Ubuntu image. Can we not directly use openssl and set OPENDISTRO_SECURITY_TEST_OPENSSL_OPT=true for the test ?

vrozov: Does it come with a compatible openssl version? My understanding is that netty version used by the plugin requires openssl 1.0.x.

.github/workflows/ci.yml
vrozov: According to the release notes openssl is not supported for Java 12+ and CI uses Java 11 with OPENDISTRO_SECURITY_TEST_OPENSSL_OPT set to `true`. @dinusX  @hardik-k-shah what is your take here?

hardik-k-shah: Currently, OpenSSL is not working with 7.4+. We did not get chance to debug that yet. 
To keep master branch working, I guess we need to disable OpenSSL, till we find a fix

vrozov: Looks like I miss something here. On the master branch, PR CI workflow uses opendistroforelasticsearch/security-maven:v1 that sets OPENDISTRO_SECURITY_TEST_OPENSSL_OPT  to true and it (CI workflow) succeeds.

.github/workflows/ci.yml
vrozov: Github image does not come with openssl preinstalled: https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md

.github/workflows/ci.yml
debjanibnrj: Did a quick scan through our codebase and it looks like `OPENDISTRO_SECURITY_TEST_OPENSSL_OPT` is only being used as part of this test - https://github.com/opendistro-for-elasticsearch/security/blob/cbcc6245bd923c545fc7e52dabe1bed267ddbf6c/src/test/java/com/amazon/opendistroforelasticsearch/security/ssl/OpenSSLTest.java#L58.


On the master branch running - 
`mvn -Dtest=OpenSSLTest#testEnsureOpenSSLAvailability test` passes as the netty library is still returning true for the statement - https://github.com/opendistro-for-elasticsearch/security/blob/cbcc6245bd923c545fc7e52dabe1bed267ddbf6c/src/test/java/com/amazon/opendistroforelasticsearch/security/ssl/OpenSSLTest.java#L61.

However for 7.4+ versions of ES security node to node communication **with Openssl** was failing (due to unknown reasons) which resulted in having to disable OpenSsl for 7.4+ versions. [PR](https://github.com/opendistro-for-elasticsearch/security/pull/224).

vrozov: @debjanibnrj Thank you for the explanation. Are there plans to investigate and enable openssl back later or is it permanently not supported? Is the issue with openssl specific to Java 12+ as the release notes suggest or it is common across all Java versions?

All, will it be OK to run tests without openssl installed and configured and with OPENDISTRO_SECURITY_TEST_OPENSSL_OPT not set on master and branches that support ES 7.4+?

debjanibnrj: We would like to get back to supporting OpenSSL. We have not invested enough time in discovering the issue with node to node communication. Will open an issue to investigate. If we see that it is not do-able we will have to take a stronger call.

Based on what I'm seeing here - https://github.com/opendistro-for-elasticsearch/security/blob/opendistro-1.4/src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/OpenDistroSecuritySSLPlugin.java#L85
OpenSSL should be supported for Java versions below 12.

You can remove the OPENDISTRO_SECURITY_TEST_OPENSSL_OPT from 7.4+ branches, I don't think it will affect the tests, but would reverify by executing a test run for the same.

PR #336, opened 2020-03-31
Security 1.1.0.6 changes 

sujithvm:
*Issue #, if available:*

*Description of changes:*


vrozov:
Please squash the first and the last commit together and split the second commit into CI/CD workflow and version upgrade. The version upgrade commits should be the last one (so it can be tagged).

PR #347, opened 2020-04-01
Increase http client timeout from 5 to 10 seconds to avoid unit test failures

vrozov:
*Description of changes:*

Increased REST request timeout in http client from 5 to 10 seconds to avoid unit test failures


vrozov:
The change is not required with the backport from #351

PR #362, opened 2020-04-07
Implement APIs and datamodel to configure nodes_dn dynamically.

krishnaggk:

**Summary of changes**:

  operations (roles, users etc)

**Example**:

```
GET /_opendistro/_security/api/nodesdn
{"cluster1":{"nodes_dn": ["CN=cluster1.example.com"]},"cluster2":{"nodes_dn": ["CN=cluster2.example.com"]}}

GET /_opendistro/_security/api/nodesdn/cluster1
{"cluster1":{"nodes_dn": ["CN=cluster1.example.com"]}}

PUT /_opendistro/_security/api/nodesdn/cluster3 -d'{"nodes_dn": ["CN=cluster3.example.com"]}'

PATCH /_opendistro/_security/api/nodesdn/cluster3 \
-d'{"op":"add", "path": "/nodesdn/-", "value": "CN=cluster-3b.example.com"}'

GET /_opendistro/_security/api/nodesdn/cluster3
{"cluster3":{"nodes_dn": ["CN=cluster3.example.com", "CN=cluster-3b.example.com"]}}

DELETE /_opendistro/_security/api/nodesdn/cluster2


```

**Testing**:

- Added tests for CRUD operations, enabling/disabling based on YML config
- Added cross-cluster specific test scenarios
- Since the intent is to keep it backward compatible, the "nodesdn" document isn't populated in the index in all tests except the specific tests testing the API. This ensures current internalusers, roles etc are not impacted by changes in repository layer.


src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigFactory.java
```diff
@@ -320,5 +328,39 @@ public String getHash(String user) {
             return Collections.emptyList();
         }
     }
+
+    private static class NodesDnModelV7 extends NodesDnModel {
```
vrozov, line 332: What is the difference between `NodesDnModelV6` and `NodesDnModelV7`?

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/impl/v6/NodesDnV6.java
```diff
@@ -0,0 +1,30 @@
+package com.amazon.opendistroforelasticsearch.security.securityconf.impl.v6;
```
vrozov, line 1: What is the difference between `NodesDnV6` and `NodesDnV7`?

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigFactory.java
```diff
@@ -230,7 +238,7 @@ public void registerDCFListener(DCFListener listener) {
     }
     
     public static interface DCFListener {
-        void onChanged(ConfigModel cm, DynamicConfigModel dcm, InternalUsersModel ium);
+        void onChanged(ConfigModel cm, DynamicConfigModel dcm, InternalUsersModel ium, NodesDnModel nm);
```
vrozov, line 241: I don't think that changing the interface is well justified. Only `DefaultInterClusterRequestEvaluator` is interested in `NodesDnModel`, all other classes that implement `DCFListener` actually do not use `NodesDnModel` on the callback event.

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/impl/v6/NodesDnV6.java
krishnaggk: Though no difference in structure, the classes serve as a way to identify the config version. V7 versions when persisted to index includes meta fields while V6 versions don't.

I'll take another look to see if we can make it work with single class.

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigFactory.java
krishnaggk: The behavior in opendistroforelasticsearch/security for older versions (0.9 and 0.10) had support for subscribing and listening to changes for specific type.

However that was removed in later versions where irrespective of type that changed, listener receives everything - https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java#L294-L298

I'll give this another thought, but do you have any suggestion?

vrozov: Please see #364

krishnaggk: > Please see #364

I left a comment on yours. May be Hardik has some context on the behavior change. But I an adapt to your change quickly once it is pushed

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
```diff
@@ -0,0 +1,150 @@
...
+        this.staticNodesDnFromEsYml = settings.getAsList(ConfigConstants.OPENDISTRO_SECURITY_NODES_DN, Collections.emptyList());
+    }
+
+    @Override
+    protected void registerHandlers(RestController controller, Settings settings) {
+        if (settings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_NODES_DN_DYNAMIC_CONFIG_ENABLED, false)) {
+            controller.registerHandler(Method.GET, "/_opendistro/_security/api/nodesdn/{name}", this);
+            controller.registerHandler(Method.GET, "/_opendistro/_security/api/nodesdn/", this);
+            controller.registerHandler(Method.DELETE, "/_opendistro/_security/api/nodesdn/{name}", this);
+            controller.registerHandler(Method.PUT, "/_opendistro/_security/api/nodesdn/{name}", this);
```
debjanibnrj, line 63: for PUT, PATCH and DELETE are we allowing users to modify nodesdn for the existing cluster? is there a way in the current design to distinguish between "cross_cluster" node dn and ones for the existing cluster?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/NodesDnValidator.java
```diff
@@ -0,0 +1,16 @@
+package com.amazon.opendistroforelasticsearch.security.dlic.rest.validation;
+
+import org.elasticsearch.common.bytes.BytesReference;
+import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.rest.RestRequest;
+
+public class NodesDnValidator extends AbstractConfigurationValidator {
+
+    public NodesDnValidator(final RestRequest request, boolean isSuperAdmin, final BytesReference ref, final Settings esSettings, Object... param) {
+        super(request, ref, esSettings, param);
```
debjanibnrj, line 10: For PUT requests where API can enter free-form value for nodesdn you may want to validate the actual DN string. Eg: "CN=OpenDistro,OU=Server CA,O=OpenDistro,C=US" is acceptable but "xyz" is not.

src/main/java/com/amazon/opendistroforelasticsearch/security/transport/DefaultInterClusterRequestEvaluator.java
```diff
@@ -47,15 +56,34 @@
...
 
-public final class DefaultInterClusterRequestEvaluator implements InterClusterRequestEvaluator {
+public final class DefaultInterClusterRequestEvaluator implements InterClusterRequestEvaluator, DCFListener {
 
     private final Logger log = LogManager.getLogger(this.getClass());
     private final String certOid;
-    private final List<String> nodesDn;
+    private final List<String> staticNodesDnFromEsYml;
+    private boolean dynamicNodesDnConfigEnabled;
+    private volatile Map<String, List<String>> dynamicNodesDn;
```
hardik-k-shah, line 65: We are already listening changes and updating it
do we still need to make it volatile?

src/main/java/com/amazon/opendistroforelasticsearch/security/transport/DefaultInterClusterRequestEvaluator.java
```diff
@@ -47,15 +56,34 @@
...
+        this.dynamicNodesDn = Collections.emptyMap();
+    }
+
+    public void subscribeForChanges(DynamicConfigFactory dynamicConfigFactory) {
+        if (this.dynamicNodesDnConfigEnabled) {
+            dynamicConfigFactory.registerDCFListener(this);
+        }
+    }
+
+    private List<String> getNodesDnToEvaluate() {
```
hardik-k-shah, line 80: what will happen during the node/cluster boot up time when .security_index is not available? dynamicNodesDn will be initialized with empty list?

src/test/java/com/amazon/opendistroforelasticsearch/security/ccstest/CrossClusterSearchTests.java
```diff
@@ -827,4 +879,113 @@ public void testCcsAggregationsDnfof() throws Exception {
...
+        assertThat(ccs.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertThat(ccs.getBody(), not(containsString("security_exception")));
+        assertThat(ccs.getBody(), containsString("\"timed_out\" : false"));
+        assertThat(ccs.getBody(), not(containsString("crl1")));
+        assertThat(ccs.getBody(), containsString("crl2"));
+        assertThat(ccs.getBody(), containsString("cross_cluster_two:twitter"));
+    }
+
+    @Test
+    public void testCcsWithDiffCertsWithNodesDnDynamicallyAdded() throws Exception {
```
hardik-k-shah, line 968: should we also add -ve test? where we do not add new cert dynamically and statically?

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigFactory.java
krishnaggk: > +1 about difference question,
> 
> and as this is new configuration, do we really need V6 as well? Are we supporting this for 6x and needed for 6.8 to 7.x migration?

Yes, will be supporting 6.7+

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
krishnaggk: Yes, the GET has a show_all which shows the existing cluster nodesdn from YML with special key (STATIC_ES_YML_NODES_DN)

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/NodesDnValidator.java
krishnaggk: Good point, but I noticed there is no validation while loading values from ES YML. Let me look this up and accommodate the validation.

src/test/java/com/amazon/opendistroforelasticsearch/security/ccstest/CrossClusterSearchTests.java
krishnaggk: There is, testCcsWithDiffCertsWithNoNodesDnUpdate()

src/main/java/com/amazon/opendistroforelasticsearch/security/transport/DefaultInterClusterRequestEvaluator.java
krishnaggk: > what will happen during the node/cluster boot up time when .security_index is not available? dynamicNodesDn will be initialized with empty list?

Correct. For boot up it uses the value from YML alone.

src/main/java/com/amazon/opendistroforelasticsearch/security/transport/DefaultInterClusterRequestEvaluator.java
krishnaggk: > We are already listening changes and updating it
> do we still need to make it volatile?

Yes, it is mostly to ensure the writes are quickly picked up in the subsequent reads. Otherwise further reads could end up reading from cache and miss the update.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/NodesDnValidator.java
```diff
@@ -0,0 +1,16 @@
+package com.amazon.opendistroforelasticsearch.security.dlic.rest.validation;
+
+import org.elasticsearch.common.bytes.BytesReference;
+import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.rest.RestRequest;
+
+public class NodesDnValidator extends AbstractConfigurationValidator {
+
+    public NodesDnValidator(final RestRequest request, boolean isSuperAdmin, final BytesReference ref, final Settings esSettings, Object... param) {
```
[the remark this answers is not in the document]
krishnaggk: > where are we using isSuperAdmin boolean?

Will remove, not using this anymore.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
```diff
@@ -0,0 +1,150 @@
...
+        if (STATIC_ES_YML_NODES_DN.equals(name)) {
+            return false;
+        }
+        return super.isReservedAndAccessible(existingConfiguration, name);
+    }
+
+    @Override
+    protected void handleGet(final RestChannel channel, RestRequest request, Client client, final JsonNode content) throws IOException {
+        final String resourcename = request.param("name");
+
```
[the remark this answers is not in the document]
krishnaggk: > super admin check will be evaluated through handleApiRequest() ?

Correct. handleApiRequest is invoked first which further dispatches to handleGet, handlePut etc

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
```diff
@@ -0,0 +1,150 @@
...
+    protected void consumeParameters(final RestRequest request) {
+        request.param("name");
+        request.param("show_all");
+    }
+
+    @Override
+    protected boolean isReservedAndAccessible(SecurityDynamicConfiguration<?> existingConfiguration, String name) {
+        if (STATIC_ES_YML_NODES_DN.equals(name)) {
+            return false;
+        }
```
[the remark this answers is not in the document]
krishnaggk: > couldn't get this condition here.
> Reserved is a type of configuration (roles, users, mapping etc) which can be only updated by super admin.
> why are we checking ES_YML_NODES_DN here?

It is an amalgamation of super-admin and **accessible**. Since this is static coming from YML, the intention is to not allow updates to this key.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -192,6 +192,11 @@ protected void handlePut(final RestChannel channel, final RestRequest request, f
 
 		final SecurityDynamicConfiguration<?> existingConfiguration = load(getConfigName(), false);
 
+		if (existingConfiguration.getSeqNo() < 0) {
+		    forbidden(channel, "Security index need to be updated to support '" + getConfigName().toLCString() + "'. Use OpenDistroSecurityAdmin to populate.");
```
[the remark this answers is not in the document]
krishnaggk: > couldn't understand this case.
> When this (existingConfiguration.getSeqNo() < 0) can happen?
> 
> And are you suggesting Security Admin tool to initialize the index?

Correct. The need comes from nodesdn usecase. This is being introduced in older versions and hence it should be backward compatible, which means in the cases where the index is already created, the APIs will not work out of the box and will need security admin to create the document in index.

Since the need is there, I added this generic check to provide better error message rather than ending with 5xx.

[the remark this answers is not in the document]
krishnaggk: > When this (existingConfiguration.getSeqNo() < 0) can happen?

This will happen when the index doesn't exist. ConfigurationLoader handles empty scenario and returns an object with negative seq number.

src/main/java/com/amazon/opendistroforelasticsearch/security/transport/DefaultInterClusterRequestEvaluator.java
vrozov: Consider using `AtomicReference` instead of `volatile`: `    private final AtomicReference<Map<String, List<String>>> dynamicNodesDn = new AtomicReference<>(Collections.emptyMap());`

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
```diff
@@ -0,0 +1,150 @@
...
+
+        configuration.removeOthers(resourcename);
+        successResponse(channel, configuration);
+    }
+
+    private void putStaticEntry(SecurityDynamicConfiguration<?> configuration) {
+        if (NodesDnV7.class.equals(configuration.getImplementingClass())) {
+            NodesDnV7 nodesDnV7 = new NodesDnV7();
+            nodesDnV7.setNodesDn(staticNodesDnFromEsYml);
```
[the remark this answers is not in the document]
krishnaggk: Primary DN will continue to be only in elasticsearch.yml.

If you notice, the static entry only shows up with explicit "show_all" flag. It is never propagated to index and the SecurityAdmin will also not use this flag while backing up or backfilling.

src/main/java/com/amazon/opendistroforelasticsearch/security/transport/DefaultInterClusterRequestEvaluator.java
krishnaggk: > Consider using `AtomicReference` instead of `volatile`: `private final AtomicReference<Map<String, List<String>>> dynamicNodesDn = new AtomicReference<>(Collections.emptyMap());`

Since I don't need CAS semantics, it seems like using volatile direclty is simpler. Do you see any other benefits of using AtomicReference in this specific scenario?

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigFactory.java
krishnaggk: Merged the classes.

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/impl/v6/NodesDnV6.java
krishnaggk: Merged the classes.

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigFactory.java
krishnaggk: Adapted to pub-sub changes in master.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/NodesDnValidator.java
krishnaggk: I took another look and realized, we support both regex and wildcard, similar to how the YML value is consumed. So we won't be able to validate.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
krishnaggk: Please reopen conv if this doesn't answer your question.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/NodesDnValidator.java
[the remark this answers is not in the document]
krishnaggk: Done.

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/Migration.java
```diff
@@ -96,7 +96,20 @@
         }
         return c7;
     }
-    
+
+    public static SecurityDynamicConfiguration<NodesDnV7> migrateNodesDn(SecurityDynamicConfiguration<NodesDnV6> n6) {
```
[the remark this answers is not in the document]
krishnaggk: I did merge the classes. However this migration section remains since the version is part of the configuration in v7 as Meta fields.

**V7**

```yml
_meta:
  type: "nodesdn"
  config_version: 2
cluster1:
  nodes_dn:
    - "CN=abc.com"
```

**V6**
```yml
cluster1:
  nodes_dn:
    - "CN=abc.com"
```

src/main/java/com/amazon/opendistroforelasticsearch/security/transport/DefaultInterClusterRequestEvaluator.java
krishnaggk: >     1. Can you please make sure if nodes_dn API calls are being audit logged properly?
>     2. Under which event it is being audit logged? Do we need to create separate audit log events?

The current audit coverage covers [limited REST API access audit logging](https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java#L368-L375)

Since the Admin access is checked after the section in the above link, they are not logged currently. However this behavior is consistent with other security API such as SSL certs APIs that were recently added. Since these are all admin API, do you think we can add them collectively?

(It looks like security index updates aren't logged, or that would have had the required info).

>     3. Can we verify 6.8 to 7.1 migration?
>     4. Can we test if security admin tools are working as expected ?

Good point. This was already being tested in SecurityAdminMigrationTests, but added a non-empty file to improve coverage.

```bash
$ pwd
~/code/native_es_stuff/fork/security/data/utest_n1_ffork_1_t17447483324660_migration

$ cat nodes_dn.yml
---
cluster1:
  nodes_dn:
  - "cn=popeye"

$ cat v7/nodes_dn.yml
---
_meta:
  type: "nodesdn"
  config_version: 2
cluster1:
  nodes_dn:
  - "cn=popeye"
  ```

>     5. Do we need any code change for certificate oid?

Nope. CertOid is fall back logic if there is no match for nodes_dn. It is singular value and hence makes sense only in YML.

File: src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java (hunk 4 above)
Comment (debjanibnrj) [reply src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java:62]: thanks for the explanation

**Comment by vrozov** [comment]
Please update the PR to exclude all files that do not require changes (for example unused imports).

**Comment by krishnaggk** [comment]
> Please update the PR to exclude all files that do not require changes (for example unused imports).

Good point. Done.

File: src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/NodesDnValidator.java [hunk 15]
```diff
@@ -0,0 +1,16 @@
...
+
+public class NodesDnValidator extends AbstractConfigurationValidator {
+
+    public NodesDnValidator(final RestRequest request, final BytesReference ref, final Settings esSettings, Object... param) {
+        super(request, ref, esSettings, param);
+        this.payloadMandatory = true;
+
+        allowedKeys.put("nodes_dn", DataType.ARRAY);
+        mandatoryKeys.add("nodes_dn");
+    }
```
sujithvm, line 15: Do we need some validation here of the input as we accepting directly List<String> ?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
```diff
@@ -0,0 +1,145 @@
...
+        super.handleApiRequest(channel, request, client);
+    }
+
+    protected void consumeParameters(final RestRequest request) {
+        request.param("name");
+        request.param("show_all");
+    }
+
+    @Override
+    protected boolean isReservedAndAccessible(SecurityDynamicConfiguration<?> existingConfiguration, String name) {
```
sujithvm, line 83: I am slightly confused here. isReservedAndAccessible checks if class implements StaticDefinable interface which NodeDn does not. 
Did you mean to override isReserved() here instead ? 

Since we are calling super.isReservedAndAccessible(), this function will always return false ?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
```diff
@@ -0,0 +1,145 @@
...
+        if (STATIC_ES_YML_NODES_DN.equals(name)) {
+            return false;
```
sujithvm, line 85: Is this gonna handle a PUT request of this form ? `PUT /_opendistro/_security/api/nodesdn/STATIC_ES_YML_NODES_DN`  ? as thats what is passed as name.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
```diff
@@ -0,0 +1,145 @@
...
+            controller.registerHandler(Method.DELETE, "/_opendistro/_security/api/nodesdn/{name}", this);
+            controller.registerHandler(Method.PUT, "/_opendistro/_security/api/nodesdn/{name}", this);
+            controller.registerHandler(Method.PATCH, "/_opendistro/_security/api/nodesdn/", this);
+            controller.registerHandler(Method.PATCH, "/_opendistro/_security/api/nodesdn/{name}", this);
+        }
+    }
+
+    @Override
+    protected void handleApiRequest(RestChannel channel, RestRequest request, Client client) throws IOException {
+        if (!isSuperAdmin()) {
```
sujithvm, line 70: Can the API be blocked from RestApiPrivilegeEvaluator instead of doing this here? 
https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/RestApiPrivilegesEvaluator.java#L213

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
```diff
@@ -144,14 +138,16 @@ public void run() {
...
+                                        ConfigHelper.uploadFile(client, cd+"roles_mapping.yml", opendistrosecurityIndex, CType.ROLESMAPPING, DEFAULT_CONFIG_VERSION);
+                                        ConfigHelper.uploadFile(client, cd+"internal_users.yml", opendistrosecurityIndex, CType.INTERNALUSERS, DEFAULT_CONFIG_VERSION);
+                                        ConfigHelper.uploadFile(client, cd+"action_groups.yml", opendistrosecurityIndex, CType.ACTIONGROUPS, DEFAULT_CONFIG_VERSION);
+                                        if(DEFAULT_CONFIG_VERSION == 2) {
+                                            ConfigHelper.uploadFile(client, cd+"tenants.yml", opendistrosecurityIndex, CType.TENANTS, DEFAULT_CONFIG_VERSION);
                                         }
+                                        final boolean populateEmptyIfFileMissing = true;
+                                        ConfigHelper.uploadFile(client, cd+"nodes_dn.yml", opendistrosecurityIndex, CType.NODESDN, DEFAULT_CONFIG_VERSION, populateEmptyIfFileMissing);
```
sujithvm, line 150: Do we reach this stage when node is re-started or opendistro index is already initialized ? I assumed it would throw exception from line #136 ? ResourceAlreadyExistsException

src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigHelper.java
```diff
@@ -70,11 +78,36 @@ public static void uploadFile(Client tc, String filepath, String index, CType cT
                 throw new Exception("   FAIL: Configuration for '" + cType.toLCString()
                         + "' failed for unknown reasons. Pls. consult logfile of elasticsearch");
             }
-        } catch (Exception e) {
-            throw e;
         }
     }
-    
+
+    public static Reader createFileOrStringReader(CType cType, int configVersion, String filepath, boolean populateEmptyIfFileMissing) throws Exception {
```
sujithvm, line 84: Do we need this populateEmptyIfFileMissing logic? Can we add the file and assume the file to exists in the builds like we do for other configs in securityconfig dir ?

src/test/java/com/amazon/opendistroforelasticsearch/security/SecurityAdminMigrationTests.java
```diff
@@ -21,6 +21,9 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import com.amazon.opendistroforelasticsearch.security.securityconf.impl.CType;
```
sujithvm, line 24: Do need any of these imports? Just seeing the diff

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/NodesDnValidator.java
krishnaggk: The nodes_dn accepts regex and wildcard similar to how the YML config is consumed. Hence chose not to validate.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
krishnaggk: > I am slightly confused here. isReservedAndAccessible checks if class implements StaticDefinable interface which NodeDn does not.
> Did you mean to override isReserved() here instead ?

Reserved means, it is only updateable by Admin. In this case, the STATIC_ES_YML_NODES_DN key shouldn't be updateable by Admin too.

> Since we are calling super.isReservedAndAccessible(), this function will always return false ?
Nope. On the contrary, it returns `true` all the time

```java
	protected boolean isReservedAndAccessible(final SecurityDynamicConfiguration<?> existingConfiguration,
											  String name) {
		if( isReserved(existingConfiguration, name) && !isSuperAdmin()) {
			return false;
		}
		return true;
	}

```

Since `isReserved` returns false, it defaults to returning true.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
krishnaggk: correct.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
krishnaggk: It won't reach here if index is already created. However it reaches here on the first time bootstrap of the cluster, where after index creation it populates the documents.

src/test/java/com/amazon/opendistroforelasticsearch/security/SecurityAdminMigrationTests.java
krishnaggk: Good point - I'll do another scan to rid off these stray imports.

src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigHelper.java
krishnaggk: The intention is to keep it backward compatible so that if customers perform minor version upgrade, they are not impacted by the document not being present. That said, there are following paths where in I can eliminate missing file handling in couple of places.

1. ConfigurationRepository.uploadFile() - This path is invoked only during index bootstrap which means folks performing minor version upgrade would not hit this path as the index already exists. Missing file handling can be removed from here.
2. OpenDistroSecurityAdmin.migrate(),upload() - These paths provide a way to populate missing documents and thus having missing file handling will be useful.

Now I can remove path 1, but it could help remote cases of migration without using SecurityAdmin tool and ensure the index is always populated. Let me know if you strongly feel the need to remove and I can remove it.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
krishnaggk: I tried that approach - however that handles RBAC which doesn't support special casing for admin DN (AdminDN is required since this is sensitive setting parallel to cert hot-reload)

sujithvm: AdminDN is already injected in the constructor and user can be determined from the thread pool context. How differently are they handled in the api action ?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/NodesDnApiAction.java
sujithvm: Yes. Sorry I did mean that function will never return false. 

If we override isReserved() and put the check `STATIC_ES_YML_NODES_DN.equals(name)`  in the function, won't we have a same behaviour?  Because what we are trying to override here is the reservedness of a resource.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/NodesDnValidator.java
sujithvm: I m just worried about how bad requests are handled. Will passing any random string be accepted like `Random= testing`?

krishnaggk:
> Thanks for making these changes!
> 
>     1. Can you please make sure if nodes_dn API calls are being audit logged properly?
>     2. Under which event it is being audit logged? Do we need to create separate audit log events?

Added a test to validate audit logging. Here is an example result

```json
{
  "sg" : "utest_n1_ffork_1_t53705646691331",
  "audit_node_name" : "node_utest_n1_ffork_1_t53705646691331_num2",
  "audit_category" : "COMPLIANCE_INTERNAL_CONFIG_READ",
  "audit_request_origin" : "REST",
  "audit_request_body" : "{\"nodesdn\":\"{\\\"cluster1\\\":{\\\"nodes_dn\\\":[\\\"cn=popeye\\\"]},\\\"_meta\\\":{\\\"type\\\":\\\"nodesdn\\\",\\\"config_version\\\":2}}\"}",
  "audit_node_id" : "f8PPfgNWTsa5loNP5fsAAg",
  "@timestamp" : "2020-04-14T07:55:29.247+00:00",
  "audit_format_version" : 4,
  "audit_trace_doc_id" : "nodesdn",
  "audit_node_host_address" : "127.0.0.1",
  "audit_request_effective_user" : "CN=kirk,OU=client,O=client,L=Test,C=DE",
  "audit_trace_indices" : [
	".opendistro_security"
  ],
  "audit_trace_resolved_indices" : [
	".opendistro_security"
  ],
  "audit_node_host_name" : "127.0.0.1"
}
{
  "audit_compliance_operation" : "UPDATE",
  "sg" : "utest_n1_ffork_1_t53705646691331",
  "audit_node_name" : "node_utest_n1_ffork_1_t53705646691331_num1",
  "audit_category" : "COMPLIANCE_INTERNAL_CONFIG_WRITE",
  "audit_request_origin" : "REST",
  "audit_compliance_doc_version" : 2,
  "audit_request_body" : "{\"cluster1\":{\"nodes_dn\":[\"cn=popeye\"]},\"_meta\":{\"type\":\"nodesdn\",\"config_version\":2}}",
  "audit_node_id" : "e4wV4PTdSjixuQuN7AXn8A",
  "@timestamp" : "2020-04-14T07:55:29.311+00:00",
  "audit_compliance_diff_is_noop" : true,
  "audit_format_version" : 4,
  "audit_request_remote_address" : "127.0.0.1",
  "audit_trace_doc_id" : "nodesdn",
  "audit_node_host_address" : "127.0.0.1",
  "audit_request_effective_user" : "CN=kirk,OU=client,O=client,L=Test,C=DE",
  "audit_trace_shard_id" : 3,
  "audit_trace_indices" : [
	".opendistro_security"
  ],
  "audit_trace_resolved_indices" : [
	".opendistro_security"
  ],
  "audit_node_host_name" : "127.0.0.1"
}
```


>     3. Can we verify 6.8 to 7.1 migration?
As we discussed offline, 7.1 code has special API for migration. I'll ensure nodes_dn migration works in that path too. Current code ensures migration via security-admin tool is tested.

>     4. Can we test if security admin tools are working as expected ?
Yes, they are part of existing UT.

>     5. Do we need any code change for certificate oid?
As discsued offline, no. OID is more appropriate for singe vaue and hence YML is better location.

krishnaggk:
> Thanks for making these changes!
> 
>     1. Can you please make sure if nodes_dn API calls are being audit logged properly?
>     2. Under which event it is being audit logged? Do we need to create separate audit log events?

Added a test to validate audit logging. Here is an example result

```json
{
  "sg" : "utest_n1_ffork_1_t53705646691331",
  "audit_node_name" : "node_utest_n1_ffork_1_t53705646691331_num2",
  "audit_category" : "COMPLIANCE_INTERNAL_CONFIG_READ",
  "audit_request_origin" : "REST",
  "audit_request_body" : "{\"nodesdn\":\"{\\\"cluster1\\\":{\\\"nodes_dn\\\":[\\\"cn=popeye\\\"]},\\\"_meta\\\":{\\\"type\\\":\\\"nodesdn\\\",\\\"config_version\\\":2}}\"}",
  "audit_node_id" : "f8PPfgNWTsa5loNP5fsAAg",
  "@timestamp" : "2020-04-14T07:55:29.247+00:00",
  "audit_format_version" : 4,
  "audit_trace_doc_id" : "nodesdn",
  "audit_node_host_address" : "127.0.0.1",
  "audit_request_effective_user" : "CN=kirk,OU=client,O=client,L=Test,C=DE",
  "audit_trace_indices" : [
	".opendistro_security"
  ],
  "audit_trace_resolved_indices" : [
	".opendistro_security"
  ],
  "audit_node_host_name" : "127.0.0.1"
}
{
  "audit_compliance_operation" : "UPDATE",
  "sg" : "utest_n1_ffork_1_t53705646691331",
  "audit_node_name" : "node_utest_n1_ffork_1_t53705646691331_num1",
  "audit_category" : "COMPLIANCE_INTERNAL_CONFIG_WRITE",
  "audit_request_origin" : "REST",
  "audit_compliance_doc_version" : 2,
  "audit_request_body" : "{\"cluster1\":{\"nodes_dn\":[\"cn=popeye\"]},\"_meta\":{\"type\":\"nodesdn\",\"config_version\":2}}",
  "audit_node_id" : "e4wV4PTdSjixuQuN7AXn8A",
  "@timestamp" : "2020-04-14T07:55:29.311+00:00",
  "audit_compliance_diff_is_noop" : true,
  "audit_format_version" : 4,
  "audit_request_remote_address" : "127.0.0.1",
  "audit_trace_doc_id" : "nodesdn",
  "audit_node_host_address" : "127.0.0.1",
  "audit_request_effective_user" : "CN=kirk,OU=client,O=client,L=Test,C=DE",
  "audit_trace_shard_id" : 3,
  "audit_trace_indices" : [
	".opendistro_security"
  ],
  "audit_trace_resolved_indices" : [
	".opendistro_security"
  ],
  "audit_node_host_name" : "127.0.0.1"
}
```


>     3. Can we verify 6.8 to 7.1 migration?
As we discussed offline, 7.1 code has special API for migration. I'll ensure nodes_dn migration works in that path too. Current code ensures migration via security-admin tool is tested.

>     4. Can we test if security admin tools are working as expected ?
Yes, they are part of existing UT.

>     5. Do we need any code change for certificate oid?
As discsued offline, no. OID is more appropriate for singe vaue and hence YML is better location.

PR #364, opened 2020-04-08
Migrate dynamic config from listener pattern to publish/subscribe event bus

vrozov:
*Issue #, if available: n/a

*Description of changes: publish/subscribe event bus is more flexible and will require less changes in the future. Please see [Event Bus](https://github.com/google/guava/wiki/EventBusExplained)

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/auth/BackendRegistry.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/auth/BackendRegistry.java
@@ -65,23 +65,22 @@ import com.amazon.opendistroforelasticsearch.security.auth.blocking.ClientBlockR
 import com.amazon.opendistroforelasticsearch.security.auth.internal.NoOpAuthenticationBackend;
 import com.amazon.opendistroforelasticsearch.security.configuration.AdminDNs;
 import com.amazon.opendistroforelasticsearch.security.http.XFFResolver;
-import com.amazon.opendistroforelasticsearch.security.securityconf.ConfigModel;
-import com.amazon.opendistroforelasticsearch.security.securityconf.DynamicConfigFactory.DCFListener;
 import com.amazon.opendistroforelasticsearch.security.securityconf.DynamicConfigModel;
-import com.amazon.opendistroforelasticsearch.security.securityconf.InternalUsersModel;
 import com.amazon.opendistroforelasticsearch.security.ssl.util.Utils;
 import com.amazon.opendistroforelasticsearch.security.support.ConfigConstants;
 import com.amazon.opendistroforelasticsearch.security.support.HTTPHelper;
 import com.amazon.opendistroforelasticsearch.security.user.AuthCredentials;
 import com.amazon.opendistroforelasticsearch.security.user.User;
+
 import com.google.common.base.Strings;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.RemovalListener;
 import com.google.common.cache.RemovalNotification;
 import com.google.common.collect.Multimap;
+import com.google.common.eventbus.Subscribe;
 
-public class BackendRegistry implements DCFListener {
+public class BackendRegistry {
 
     protected final Logger log = LogManager.getLogger(this.getClass());
     private SortedSet<AuthDomain> restAuthDomains;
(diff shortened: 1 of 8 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigFactory.java
```diff
@@ -225,12 +223,12 @@ public final boolean isInitialized() {
         return initialized.get();
     }
     
-    public void registerDCFListener(DCFListener listener) {
-        listeners.add(listener);
+    public void registerDCFListener(Object listener) {
```
debjanibnrj, line 226: Since we're removing DCFListener maybe a more generic name like registerListener?

krishnaggk:
> *Issue #, if available: n/a
> 
> *Description of changes: publish/subscribe event bus is more flexible and will require less changes in the future. Please see [Event Bus](https://github.com/google/guava/wiki/EventBusExplained)
> 
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

It looks cleaner, one question that bothers me is why it moved from specific listener to listening to everything from 0.10 (ES 6.x) to 1.0 (ES 7.x). But since the tests have passed may be there is nothing to be concerned about.

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigFactory.java
vrozov: I think it would be OK to keep the original name. It signifies the event bus the listener subscribes to. Note that in the past it was also not necessary to have `DCF` in the `registerDCFListener()` name, as it was clear that the registration is done for `DCFListener`.

vrozov:
> I see we're removing DCFListener from most places. Do other classes use it? Should we remove the DCFListener class as well?

The `DCFListener` interface is removed as it is not in use by any other classes.

>Other thing I noticed is that we don't have any unit tests for these classes, maybe something to consider adding to our issues.

The functionality is tested as part of `SecurityAdminMigrationTests` and few other tests.

vrozov:
>It looks cleaner, one question that bothers me is why it moved from specific listener to listening to everything from 0.10 (ES 6.x) to 1.0 (ES 7.x). But since the tests have passed may be there is nothing to be concerned about.

It's hard to say why it moved in that direction. My guess is that the design goal was to avoid code duplication and unify it in `DynamicConfigFactory` rather than listening to all types of events.

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigFactory.java
```diff
@@ -81,7 +82,7 @@ private void loadStaticConfig() throws IOException {
     protected final Logger log = LogManager.getLogger(this.getClass());
     private final ConfigurationRepository cr;
     private final AtomicBoolean initialized = new AtomicBoolean();
-    private final List<DCFListener> listeners = new ArrayList<>();
+    private final EventBus eventBus = new EventBus("DynamicConfig");
```
sujithvm, line 85: Wondering if this should be injected from the constructor as some other classes can post on the same event bus.

src/main/java/com/amazon/opendistroforelasticsearch/security/auth/BackendRegistry.java
```diff
@@ -207,8 +206,8 @@ public void invalidateCache() {
         transportImpersonationCache.invalidateAll();
     }
 
-    @Override
-    public void onChanged(ConfigModel cm, DynamicConfigModel dcm, InternalUsersModel ium) {
+    @Subscribe
+    public void onDynamicConfigModelChanged(DynamicConfigModel dcm) {
```
sujithvm, line 210: How does this know which event bus to subscribe to? 
What happens if I make another event bus and post DynamicConfigModel ? Does this automatically subscribe for both?

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigFactory.java
vrozov: Currently `eventBus` is not shared across `DynamicConfigFactory` class instances similar to how `listeners` was not shared. Till there is a use case where `eventBus` needs to be shared, I'd prefer to keep it encapsulated inside `DynamicConfigFactory`.

src/main/java/com/amazon/opendistroforelasticsearch/security/auth/BackendRegistry.java
vrozov: It is necessary to register with the event bus. Please see `DynamicConfigFactory.registerDCFListener()`.

Issue #313, opened 2020-03-24
Refactor ComplianceConfig


sujithvm:
- Refactor the usage of ComplianceConfig to generate config from settings

This will help in hot-reloading of compliance config params 

Dependent on #312

PR #368, opened 2020-04-09
Refactor compliance config

sujithvm:
Issue #313 

Changes:

- constructor based instantiation 

- added tests 

- moved externalConfigLogged and environment out of compliance config as they are storing specific startup state and should not be part of compliance config 

- moved out functionality to check if index is immutable using index resolver (removed index resolver dependency from config)

Similar PR: https://github.com/opendistro-for-elasticsearch/security/pull/306

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
@@ -188,7 +188,6 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
     private final boolean sslOnly;
     private final List<String> demoCertHashes = new ArrayList<String>(3);
     private volatile OpenDistroSecurityFilter odsf;
-    private volatile ComplianceConfig complianceConfig;
     private volatile Environment environment;
     private volatile IndexResolverReplacer irr;
     private AtomicBoolean externalConfigLogged = new AtomicBoolean();
@@ -238,7 +237,6 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
             this.advancedModulesEnabled = false;
             this.sslOnly = false;
             this.sslCertReloadEnabled = false;
-            complianceConfig = null;
             log.warn("Open Distro Security plugin installed but disabled. This can expose your configuration (including passwords) to the public.");
             return;
         }
@@ -250,7 +248,6 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
             this.dlsFlsConstructor = null;
             this.advancedModulesEnabled = false;
             this.sslCertReloadEnabled = false;
-            complianceConfig = null;
             log.warn("Open Distro Security plugin run in ssl only mode. No authentication or authorization is performed");
             return;
         }
@@ -489,14 +486,13 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
     }
 
     private CheckedFunction<DirectoryReader, DirectoryReader, IOException>  loadFlsDlsIndexSearcherWrapper(final IndexService indexService,
-            final ComplianceIndexingOperationListener ciol, final ComplianceConfig complianceConfig) {
+            final ComplianceIndexingOperationListener ciol) {
         try {
             CheckedFunction<DirectoryReader, DirectoryReader, IOException>  flsdlsWrapper = (CheckedFunction<DirectoryReader, DirectoryReader, IOException> ) dlsFlsConstructor
                     .newInstance(indexService, settings, Objects.requireNonNull(adminDns),
                             Objects.requireNonNull(cs),
                             Objects.requireNonNull(auditLog),
                             Objects.requireNonNull(ciol),
-                            Objects.requireNonNull(complianceConfig),
                             Objects.requireNonNull(evaluator));
             if(log.isDebugEnabled()) {
                 log.debug("FLS/DLS enabled for index {}", indexService.index().getName());
@@ -512,6 +508,7 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
         //called for every index!
 
         if (!disabled && !client && !sslOnly) {
+            final ComplianceConfig complianceConfig = auditLog.getCurrentComplianceConfig();
             log.debug("Handle complianceConfig="+complianceConfig+"/dlsFlsAvailable: "+dlsFlsAvailable+"/auditLog="+auditLog.getClass()+" for onIndexModule() of index "+indexModule.getIndex().getName());
             if (dlsFlsAvailable) {
 
@@ -520,13 +517,13 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
                 assert complianceConfig!=null:"compliance config must not be null here";
                 
                 if(complianceConfig.writeHistoryEnabledForIndex(indexModule.getIndex().getName())) {
-                    ciol = ReflectionHelper.instantiateComplianceListener(complianceConfig, Objects.requireNonNull(auditLog));
+                    ciol = ReflectionHelper.instantiateComplianceListener(Objects.requireNonNull(auditLog));
                     indexModule.addIndexOperationListener(ciol);
                 } else {
                     ciol = new ComplianceIndexingOperationListener();
                 }
 
-                indexModule.setReaderWrapper(indexService -> loadFlsDlsIndexSearcherWrapper(indexService, ciol, complianceConfig));
+                indexModule.setReaderWrapper(indexService -> loadFlsDlsIndexSearcherWrapper(indexService, ciol));
                 indexModule.forceQueryCacheProvider((indexSettings,nodeCache)->new QueryCache() {
 
                     @Override
@@ -755,7 +752,7 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
         final IndexNameExpressionResolver resolver = new IndexNameExpressionResolver();
         irr = new IndexResolverReplacer(resolver, clusterService, cih);
         auditLog = ReflectionHelper.instantiateAuditLog(settings, configPath, localClient, threadPool, resolver, clusterService);
-        complianceConfig = (dlsFlsAvailable && (auditLog.getClass() != NullAuditLog.class)) ? ComplianceConfig.from(settings) : null;
+        final ComplianceConfig complianceConfig = (dlsFlsAvailable && (auditLog.getClass() != NullAuditLog.class)) ? ComplianceConfig.from(settings) : null;
         log.debug("Compliance config is "+complianceConfig+" because of dlsFlsAvailable: "+dlsFlsAvailable+" and auditLog="+auditLog.getClass());
         if (complianceConfig != null)
             complianceConfig.log(log);
@@ -778,10 +775,8 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
 
         adminDns = new AdminDNs(settings);
         
-        cr = (ConfigurationRepository) ConfigurationRepository.create(settings, this.configPath, threadPool, localClient, clusterService, auditLog,
-                complianceConfig);
+        cr = ConfigurationRepository.create(settings, this.configPath, threadPool, localClient, clusterService, auditLog);
 
-        //cr.subscribeOnLicenseChange(complianceConfig); TODO : Remove this line post compilation
         final XFFResolver xffResolver = new XFFResolver(threadPool);
         backendRegistry = new BackendRegistry(settings, adminDns, xffResolver, auditLog, threadPool);
         
@@ -802,7 +797,7 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
         
         cr.setDynamicConfigFactory(dcf);
         
-        odsf = new OpenDistroSecurityFilter(evaluator, adminDns, dlsFlsValve, auditLog, threadPool, cs, complianceConfig, compatConfig, irr);
+        odsf = new OpenDistroSecurityFilter(evaluator, adminDns, dlsFlsValve, auditLog, threadPool, cs, compatConfig, irr);
 
 
         final String principalExtractorClass = settings.get(SSLConfigConstants.OPENDISTRO_SECURITY_SSL_TRANSPORT_PRINCIPAL_EXTRACTOR_CLASS, null);
@@ -1016,6 +1011,7 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
         }
         final Set<ModuleInfo> securityModules = ReflectionHelper.getModulesLoaded();
         log.info("{} Open Distro Security modules loaded so far: {}", securityModules.size(), securityModules);
+        final ComplianceConfig complianceConfig = auditLog.getCurrentComplianceConfig();
         if(complianceConfig != null && complianceConfig.isEnabled() && complianceConfig.shouldLogExternalConfig() && !externalConfigLogged.getAndSet(true)) {
             log.info("logging external config");
             auditLog.logExternalConfig(settings, environment);
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
@@ -41,6 +41,7 @@ import java.security.MessageDigest;
 import java.security.PrivilegedAction;
 import java.security.Security;
 import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
@@ -190,7 +191,7 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
     private volatile ComplianceConfig complianceConfig;
     private volatile Environment environment;
     private volatile IndexResolverReplacer irr;
-    private volatile boolean externalConfigLogged = false;
+    private AtomicBoolean externalConfigLogged = new AtomicBoolean();
     private volatile NamedXContentRegistry namedXContentRegistry = null;
     private volatile DlsFlsRequestValve dlsFlsValve = null;
 
@@ -1015,10 +1016,9 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
         }
         final Set<ModuleInfo> securityModules = ReflectionHelper.getModulesLoaded();
         log.info("{} Open Distro Security modules loaded so far: {}", securityModules.size(), securityModules);
-        if(complianceConfig != null && complianceConfig.isEnabled() && complianceConfig.shouldLogExternalConfig() && !externalConfigLogged) {
+        if(complianceConfig != null && complianceConfig.isEnabled() && complianceConfig.shouldLogExternalConfig() && !externalConfigLogged.getAndSet(true)) {
             log.info("logging external config");
             auditLog.logExternalConfig(settings, environment);
-            externalConfigLogged = true;
         }
     }
 
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
@@ -1011,10 +1011,12 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
         }
         final Set<ModuleInfo> securityModules = ReflectionHelper.getModulesLoaded();
         log.info("{} Open Distro Security modules loaded so far: {}", securityModules.size(), securityModules);
-        final ComplianceConfig complianceConfig = auditLog.getCurrentComplianceConfig();
-        if(complianceConfig != null && complianceConfig.isEnabled() && complianceConfig.shouldLogExternalConfig() && !externalConfigLogged.getAndSet(true)) {
-            log.info("logging external config");
-            auditLog.logExternalConfig(settings, environment);
+        if (auditLog != null) {
+            final ComplianceConfig complianceConfig = auditLog.getCurrentComplianceConfig();
+            if(complianceConfig != null && complianceConfig.isEnabled() && complianceConfig.shouldLogExternalConfig() && !externalConfigLogged.getAndSet(true)) {
+                log.info("logging external config");
+                auditLog.logExternalConfig(settings, environment);
+            }
         }
     }
 
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
@@ -751,12 +751,7 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
 
         final IndexNameExpressionResolver resolver = new IndexNameExpressionResolver();
         irr = new IndexResolverReplacer(resolver, clusterService, cih);
-        auditLog = ReflectionHelper.instantiateAuditLog(settings, configPath, localClient, threadPool, resolver, clusterService);
-        final ComplianceConfig complianceConfig = (dlsFlsAvailable && (auditLog.getClass() != NullAuditLog.class)) ? ComplianceConfig.from(settings) : null;
-        log.debug("Compliance config is "+complianceConfig+" because of dlsFlsAvailable: "+dlsFlsAvailable+" and auditLog="+auditLog.getClass());
-        if (complianceConfig != null)
-            complianceConfig.log(log);
-        auditLog.setComplianceConfig(complianceConfig);
+        auditLog = ReflectionHelper.instantiateAuditLog(settings, configPath, localClient, threadPool, resolver, clusterService, dlsFlsAvailable);
         
         sslExceptionHandler = new AuditLogSslExceptionHandler(auditLog);
 
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
@@ -188,7 +188,9 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
     private final List<String> demoCertHashes = new ArrayList<String>(3);
     private volatile OpenDistroSecurityFilter odsf;
     private volatile ComplianceConfig complianceConfig;
+    private volatile Environment environment;
     private volatile IndexResolverReplacer irr;
+    private volatile boolean externalConfigLogged = false;
     private volatile NamedXContentRegistry namedXContentRegistry = null;
     private volatile DlsFlsRequestValve dlsFlsValve = null;
 
@@ -737,6 +739,7 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
         this.threadPool = threadPool;
         this.cs = clusterService;
         this.localClient = localClient;
+        this.environment = environment;
 
         final List<Object> components = new ArrayList<Object>();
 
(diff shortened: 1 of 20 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -41,209 +56,302 @@
...
+    }
 
-	public boolean isEnabled() {
-        return this.enabled;
+    /**
+     * Checks if compliance is enabled
+     * @return true/false
+     */
+    public boolean isEnabled() {
+        return true;
```
opendistroforelasticsearch-security, line 228: I see you have tried to replace the values with class member variables before this line. 
Why opposite here?

palashhedau, line 228: I see you have changed values with the class member variables. Why opposite here?

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -30,6 +30,21 @@
 
 package com.amazon.opendistroforelasticsearch.security.compliance;
 
+import com.amazon.opendistroforelasticsearch.security.support.ConfigConstants;
```
vrozov, line 33: Please avoid reordering of imports.

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
sujithvm: I will revert this back to class member variable. 

I actually wanted to remove this isEnabled() itself because its always set to true. I believe this is something we inherited from search-guard where compliance was offered as a separate feature that could be disabled. 
Other places check for this explicitly hence I retained for some future use-cases.

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
sujithvm: Done!

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -330,12 +335,27 @@ private boolean checkImmutableIndices(Object request, ActionListener listener) {
         }
         
         if(request instanceof IndexRequest) {
-            if(complianceConfig != null && complianceConfig.isIndexImmutable(request)) {
+            if(complianceConfig != null && isIndexImmutable(request)) {
```
vrozov, line 338: move check `complianceConfig != null ` to `isIndexImmutable()`.

sujithvm: Done!

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
```diff
@@ -1010,10 +1015,10 @@ public void onNodeStarted() {
...
         final Set<ModuleInfo> securityModules = ReflectionHelper.getModulesLoaded();
         log.info("{} Open Distro Security modules loaded so far: {}", securityModules.size(), securityModules);
-        if(complianceConfig != null && complianceConfig.isEnabled() && complianceConfig.isLogExternalConfig() && !complianceConfig.isExternalConfigLogged()) {
-        	log.info("logging external config");
-        	auditLog.logExternalConfig(complianceConfig.getSettings(), complianceConfig.getEnvironment());
-            complianceConfig.setExternalConfigLogged(true);
+        if(complianceConfig != null && complianceConfig.isEnabled() && complianceConfig.shouldLogExternalConfig() && !externalConfigLogged) {
+            log.info("logging external config");
+            auditLog.logExternalConfig(settings, environment);
+            externalConfigLogged = true;
```
vrozov, line 1021: Possible race condition, two threads may race for `externalConfigLogged`.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -311,7 +316,7 @@ private boolean checkImmutableIndices(Object request, ActionListener listener) {
                 || request instanceof IndicesAliasesRequest //TODO only remove index
                 ) {
             
-            if(complianceConfig != null && complianceConfig.isIndexImmutable(request)) {
+            if (isIndexImmutable(request)) {
```
vrozov, line 319: nit: move to enclosing `if`

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -330,12 +335,27 @@ private boolean checkImmutableIndices(Object request, ActionListener listener) {
         }
         
         if(request instanceof IndexRequest) {
-            if(complianceConfig != null && complianceConfig.isIndexImmutable(request)) {
+            if (isIndexImmutable(request)) {
```
vrozov, line 338: nit: move to enclosing `if`

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -330,12 +335,27 @@ private boolean checkImmutableIndices(Object request, ActionListener listener) {
...
                 ((IndexRequest) request).opType(OpType.CREATE);
             }
         }
         
         return false;
     }
 
-}
\ No newline at end of file
+    private boolean isIndexImmutable(Object request) {
```
vrozov, line 346: nit: consider `isRequestIndexImmutable`

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -330,12 +335,27 @@ private boolean checkImmutableIndices(Object request, ActionListener listener) {
...
+        if (complianceConfig == null || !complianceConfig.isEnabled()) {
```
vrozov, line 347: Actually this check is redundant. There is the same condition on line 198 before `checkImmutableIndices()` call.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -330,12 +335,27 @@ private boolean checkImmutableIndices(Object request, ActionListener listener) {
...
+            return false;
+        }
+
+        final Set<String> immutableIndicesPatterns = complianceConfig.getImmutableIndicesPatterns();
+        if (immutableIndicesPatterns.isEmpty()) {
```
vrozov, line 352: Should this be part of the check on line 198? You can check that `immutableIndicesPatterns` is not empty once for `BulkShardRequest` and assume that it is not empty here or pass it to `checkImmutableIndices()`.

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
sujithvm: externalConfigLogged is set t volatile. This is existing behavior as well. It is just pulled out of compliance config

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
sujithvm: Retained the existing function name. Wanted to reduce diffs

vrozov: This is a new function, the diff is already there.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
sujithvm: Well, I think this should be retained here while calling this function

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
sujithvm: Retained existing code.

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
vrozov: `volatile` does not prevent race condition, it only provides memory barrier. It will be OK to file that as an issue and address in a separate PR.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
vrozov: Why? It is a private method and there is no need to duplicate the check.

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
sujithvm: Yes it would be great to address it separately.
Are you saying onNodeStarted will called by multiple threads ? Need to investigate the race condition

vrozov: It is a *possible* race condition. As plugin has no control when and how `onNodeStarted()` is called, it will be safer to avoid the race condition, especially that it will not cost anything. Just replace `volatile` with `AtomicBoolean` and use `getAndSet()`.

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -45,205 +45,315 @@
...
+        this.logReadMetadataOnly = logReadMetadataOnly;
+        this.logWriteMetadataOnly = logWriteMetadataOnly;
+        this.logDiffsForWrite = logDiffsForWrite;
+        this.watchedWriteIndices = watchedWriteIndices;
+        this.immutableIndicesPatterns = immutableIndicesPatterns;
+        this.opendistrosecurityIndex = opendistrosecurityIndex;
 
-        if(saltAsString.equals(ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_SALT_DEFAULT)) {
-            log.warn("If you plan to use field masking pls configure "+ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_SALT+" to be a random string of 16 chars length identical on all nodes");
+        final byte[] saltAsBytes = saltAsString.getBytes(StandardCharsets.UTF_8);
```
vrozov, line 110: Consider:
```
        if (saltAsString.equals(ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_SALT_DEFAULT)) {
            log.warn("If you plan to use field masking pls configure compliance salt {} to be a random string of 16 chars length identical on all nodes", saltAsString);
        }
        try {
            ByteBuffer byteBuffer = StandardCharsets.UTF_8.encode(saltAsString);
            byteBuffer.get(salt16);
            if (byteBuffer.remaining() > 0) {
                log.warn("Provided compliance salt {} is greater than 16 bytes. Only the first 16 bytes are used for salting", saltAsString);
            }
        } catch (BufferUnderflowException e) {
            throw new ElasticsearchException("Provided compliance salt " + saltAsString + " must at least contain 16 bytes", e);
        }
```
Please add test cases for salt.

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -45,205 +45,315 @@
...
-        
-        salt16 = Arrays.copyOf(saltAsBytes, 16);
-        this.opendistrosecurityIndex = settings.get(ConfigConstants.OPENDISTRO_SECURITY_CONFIG_INDEX_NAME, ConfigConstants.OPENDISTRO_SECURITY_DEFAULT_CONFIG_INDEX);
-        
+        this.salt16 = Arrays.copyOf(saltAsBytes, SALT_SIZE);
+
         //opendistro_security.compliance.pii_fields:
         //  - indexpattern,fieldpattern,fieldpattern,....
-        for(String watchedReadField: watchedReadFields) {
+        for (String watchedReadField : watchedReadFields) {
```
vrozov, line 124: Consider:
```
        this.readEnabledFields = watchedReadFields.stream()
                .map(watchedReadField -> watchedReadField.split(","))
                .filter(split -> split.length != 0)
                .collect(Collectors.toMap(
                        split -> split[0],
                        split -> split.length == 1 ?
                                Collections.singleton("*") : Arrays.stream(split).skip(1).collect(Collectors.toSet())
                ));
```

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -45,205 +45,315 @@
...
+    private static final int CACHE_SIZE = 1000;
+    private static final String INTERNAL_ELASTICSEARCH = "internal_elasticsearch";
+
     private final boolean logExternalConfig;
     private final boolean logInternalConfig;
-    private final LoadingCache<String, Set<String>> cache;
+    private final boolean logReadMetadataOnly;
+    private final boolean logWriteMetadataOnly;
+    private final boolean logDiffsForWrite;
+    private final List<String> watchedWriteIndices;
```
vrozov, line 77: Should it be `Set<String> watchedWriteIndicesPatterns` to be consistent with `Set<String> immutableIndicesPatterns`?

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -45,205 +45,315 @@
...
     private final Set<String> immutableIndicesPatterns;
-    private final byte[] salt16;
     private final String opendistrosecurityIndex;
-    private final IndexResolverReplacer irr;
-    private final Environment environment;
-    private final AuditLog auditLog;
+
+    private final Map<String, Set<String>> readEnabledFields = new HashMap<>(100);
+    private final LoadingCache<String, Set<String>> cache;
```
vrozov, line 82: nit: please change `cache` to a more descriptive name.

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -45,205 +45,315 @@
...
+    public Set<String> getImmutableIndicesPatterns() {
+        return immutableIndicesPatterns;
+    }
+
+    /**
+     * Get the salt in bytes for filed anonymization
+     * @return salt in bytes
+     */
+    public byte[] getSalt16() {
+        return salt16.clone();
```
vrozov, line 267: I don't think that copy/clone is necessary, but if it is, I would rather use `Arrays.copyOf()`.

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -45,205 +45,315 @@
...
+        if (auditLogPattern != null) {
+            if (index.equalsIgnoreCase(getExpandedIndexName(auditLogPattern, null))) {
+                return Collections.emptySet();
             }
         }
 
         final Set<String> tmp = new HashSet<String>(100);
-        for(String indexPattern: readEnabledFields.keySet()) {
-            if(indexPattern != null && !indexPattern.isEmpty() && WildcardMatcher.match(indexPattern, index)) {
+        for (String indexPattern : readEnabledFields.keySet()) {
```
vrozov, line 291: Consider:
```
        return readEnabledFields.entrySet().stream()
                .filter(entry -> WildcardMatcher.match(entry.getKey(), index))
                .flatMap(entry -> entry.getValue().stream())
                .collect(Collectors.toSet());
```
There is no point of adding `null` or `empty` `indexPattern` to `readEnabledFields`, add it as a filter to the constructor.

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
sujithvm: The array can be modified by some other classes right? Requesting salt will give the modified array.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -87,6 +87,7 @@
     private final Settings settings;
     private final AuditConfig.Filter auditConfigFilter;
     private final String opendistrosecurityIndex;
+    protected ComplianceConfig complianceConfig;
```
vrozov, line 90: `private volatile` or `private final AtomicReference`

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/AuditLog.java
```diff
@@ -69,13 +69,15 @@
...
-    void logDocumentWritten(ShardId shardId, GetResult originalIndex, Index currentIndex, IndexResult result, ComplianceConfig complianceConfig);
+    void logDocumentRead(String index, String id, ShardId shardId, Map<String, String> fieldNameValues);
+    void logDocumentWritten(ShardId shardId, GetResult originalIndex, Index currentIndex, IndexResult result);
     void logDocumentDeleted(ShardId shardId, Delete delete, DeleteResult result);
     void logExternalConfig(Settings settings, Environment environment);
     
     // compliance config
     void setComplianceConfig(ComplianceConfig complianceConfig);
+
+    ComplianceConfig getCurrentComplianceConfig();
```
vrozov, line 80: `getCurrentComplianceConfig` -> `getComplianceConfig`

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: I thought of doing the volatile change when subscribe is added. 
But I ll add it here itself

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/AuditLog.java
sujithvm: I thought the name getCurrentComplianceConfig will be more descriptive that config fetched is the latest/current config

sujithvm: Let me know if you would like to use `getComplianceConfig` itself or some other name.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/AuditLog.java
```diff
@@ -69,13 +69,15 @@
...
 
-    void logDocumentRead(String index, String id, ShardId shardId, Map<String, String> fieldNameValues, ComplianceConfig complianceConfig);
-    void logDocumentWritten(ShardId shardId, GetResult originalIndex, Index currentIndex, IndexResult result, ComplianceConfig complianceConfig);
+    void logDocumentRead(String index, String id, ShardId shardId, Map<String, String> fieldNameValues);
+    void logDocumentWritten(ShardId shardId, GetResult originalIndex, Index currentIndex, IndexResult result);
     void logDocumentDeleted(ShardId shardId, Delete delete, DeleteResult result);
     void logExternalConfig(Settings settings, Environment environment);
     
     // compliance config
     void setComplianceConfig(ComplianceConfig complianceConfig);
```
vrozov, line 78: Is it necessary to have `setComplianceConfig()` method? Can it be instantiated inside `AbstractAuditLog` constructor?

sujithvm: I want to retain this method as subscribing for changes should be as simple as adding @Subscribe to the function for changes.

vrozov: `@Subscribe` should not be on the interface method. `NullAuditLog` should not subscribe to the event bus.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/AuditLog.java
vrozov: `getCurrentComplienceConfig()` is confusing, IMO. It is just a getter method, it can only get the current value, there is no any other values to get.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
vrozov: If you remove setter (`setComplienceConfig()`), it can wait till subscribe is added.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/AuditLog.java
sujithvm: Sorry I meant the function in AuditLogImpl and not this interface method

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
sujithvm: Alright. I will remove setter `setComplienceConfig`  and re-add when needed

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -87,6 +87,7 @@
     private final Settings settings;
     private final AuditConfig.Filter auditConfigFilter;
     private final String opendistrosecurityIndex;
+    protected volatile ComplianceConfig complianceConfig;
```
vrozov, line 90: It should be `private`. If necessary, there is public getter.

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -31,292 +31,354 @@
...
+        // if open distro index (internal index) check if internal config logging is enabled
+        if (opendistrosecurityIndex.equals(index)) {
             return logInternalConfig;
         }
-        
         try {
-            return !cache.get(index).isEmpty();
+            return !readEnabledFieldsCache.get(index).isEmpty();
         } catch (ExecutionException e) {
             log.error(e);
```
vrozov, line 354: Consider:
```
            log.warn("Failed to get index {} fields enabled for read from cache. Bypassing cache.", index, e);
            return getFieldsForIndex(index).isEmpty();
```

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -31,292 +31,354 @@
...
-            final Set<String> fields = cache.get(index);
-            if(fields.isEmpty()) {
+            final Set<String> fields = readEnabledFieldsCache.get(index);
+            if (fields.isEmpty()) {
                 return false;
             }
-
             return WildcardMatcher.matchAny(fields, field);
         } catch (ExecutionException e) {
             log.error(e);
```
vrozov, line 380: Consider:
```
            log.warn("Failed to get index {} fields enabled for read from cache. Bypassing cache.", index, e);
            fields = getFieldsForIndex(index);
```

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AuditMessageRouter.java
```diff
@@ -65,22 +63,18 @@ public AuditMessageRouter(final Settings settings, final Client clientProvider,
...
-	public void setComplianceConfig(ComplianceConfig complianceConfig) {
-		this.complianceConfig = complianceConfig;
-	}
-
 	public boolean isEnabled() {
 		return this.enabled;
 	}
 
-	public final void route(final AuditMessage msg) {
+	public final void route(final AuditMessage msg, final boolean complianceDisabled) {
```
vrozov, line 70: `complianceDisabled` -> `isComplienceConfigEnabled` (true -> enabled)

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AuditMessageRouter.java
```diff
@@ -65,22 +63,18 @@ public AuditMessageRouter(final Settings settings, final Client clientProvider,
...
 
-	public void setComplianceConfig(ComplianceConfig complianceConfig) {
-		this.complianceConfig = complianceConfig;
-	}
-
 	public boolean isEnabled() {
 		return this.enabled;
 	}
 
-	public final void route(final AuditMessage msg) {
```
vrozov, line 76: keep `public final void route(final AuditMessage msg)` and use the most common default to call `route(msg, ...)`

sujithvm: What do you mean by most common default ?

vrozov:
```
route(msg, true)
```

sujithvm: Are you saying to have two functions
1) route(msg, true)
2) route(msg) calling the above ? 

route(msg) would probably be useful only for tests (which I have updated already). Dont want to add more code to main just for tests.

vrozov: It is OK to provide a default so that test don't need to be changed.

sujithvm: The tests are already updated

vrozov: Please consider the following approach:
1. revert changes to the `route()` method (it should only take `AuditMessage msg` as an argument)
2. introduce a `boolean` instance variable `areRoutesEnabled` (`false` by default).
3. in `route()` instead of  `isComplianceConfigEnabled` use `areRoutesEnabled`.
4. rename `setupRoutes()` to `enableRoutes()` and make it public
5. do not call `enableRoutes()` from `AuditMessageRouter` constructor, instead, call it from `AuditLogImpl` if it is `enabled` and `complianceConfig` is also enabled

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
vrozov: There are no other classes that modify the array. In this particular case, if you track how salt is used, it will be copied twice.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AuditMessageRouter.java
sujithvm: Could you explain the value in doing this and insisting on removing the current change ? 

IMO, I see no issue in route(msg, isComplianceConfigEnabled) as it determines what route to take from the audit log based on current compliance configuration for hot reload. 

Also would appreciate if you could send the changes in a PR against this branch so that we can quickly move forward :)

vrozov: The value of the proposal is that
1. `areRoutesEnabled` is the state of `AuditMessageRouter` instance that is simply a delegate of `AuditLogImpl`, not a parameter for the `route()` method (that should only take message to route). Such state will be shared by many calls to the `route()` until there is a state transition.
2. It is not necessary to initialize routes if routes are not enabled.

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
sujithvm: If you look at MaskedField where you are passing salt16, you are passing the salt to 3rd lib function like this https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/MaskedField.java#L176 

You never know if other libs modify the array or not. Is it safer to return a copy and I would prefer doing the same.

Moreover copying salt is current functionality and can changes around this if needed be done separately?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AuditMessageRouter.java
vrozov: Please see https://github.com/sujithvm/security/pull/57

sujithvm: I would prefer AuditMessageRouter improvements to be done separately.
Would appreciate if you could share your changes :) Dont want to trigger another set of changes around this.

vrozov: The improvements to `AuditMessageRouter` are minimal, but it is necessary to keep original `route()` to take message as the only parameter.

PR #377, opened 2020-04-14
[opendistro-0.9 backport] Implement APIs and datamodel to configure nodes_dn dynamically.

krishnaggk:
This is a backport of pull request #362

**Summary of changes**:

  operations (roles, users etc)

---


src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -346,7 +346,7 @@ public ConfigUpdatingActionListener(Client client, ActionListener<Response> dele
 		@Override
 		public void onResponse(Response response) {
 
-			final ConfigUpdateRequest cur = new ConfigUpdateRequest(new String[] { "config", "roles", "rolesmapping", "internalusers", "actiongroups" });
+			final ConfigUpdateRequest cur = new ConfigUpdateRequest(ConfigConstants.ALL_CONFIG_NAMES.toArray(new String[0]));
```
sujithvm, line 349: nit: Is it possible to make another constant variable for this? 
`ConfigConstants.ALL_CONFIG_NAMES.toArray(new String[0])` as it seems to be used quite a lot.

PR #396, opened 2020-04-21
[opendistro-1.4 backport] Implement APIs and datamodel to configure nodes_dn dynamically.

krishnaggk:
This is a port of pull request #362 (cherry-picked from master) and corresponding dependent commits.

**Summary of changes**:
[opendistro-1.4 backport] Implement APIs and datamodel to configure nodes_dn dynamically.

  operations (roles, users etc)

---


krishnaggk:
Closing this - will create a new one without the experimentation stuff.

PR #402, opened 2020-04-22
Implement put if absent behavior for security config

sujithvm:
Use case: If you want to include new config file (say audit.yml for audit logging configuration) to a cluster (which already contains opendistro index), then it loads new config file to an existing index.

Current behavior:

- If opendistro index does not exist (run CreateIndexRequest)
  - Create a new index 
  - Load the security config files
- if opendistro index exists (run CreateIndexRequest)
  - throw exception and skips trying to load any security config files

Proposed behavior:
- check if opendistro index exists (run GetIndexRequest)
- if does not exist, create index using CreateIndexRequest
- try to load any security config into the index using IndexRequest with optype set to Create for put if absent behavior
- if doc already exists (throw VersionConflictEngineException) do nothing as config is already loaded.

This behavior is good if we are trying to populate new config into the index instead of skipping attempt to load them. 

Put if absent behavior: 
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#docs-index-api-query-params

Error is thrown when doc with same id and index is provided when opType is set to CREATE
https://www.elastic.co/guide/en/elasticsearch/client/java-rest/master/java-rest-high-document-index.html#java-rest-high-document-index-response

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigHelper.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigHelper.java
@@ -39,6 +39,7 @@ import java.io.StringReader;
 import com.amazon.opendistroforelasticsearch.security.securityconf.impl.Meta;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.elasticsearch.action.DocWriteRequest.OpType;
 import org.elasticsearch.action.index.IndexRequest;
 import org.elasticsearch.action.support.WriteRequest.RefreshPolicy;
 import org.elasticsearch.client.Client;
@@ -52,6 +53,7 @@ import org.elasticsearch.common.xcontent.XContentType;
 import com.amazon.opendistroforelasticsearch.security.DefaultObjectMapper;
 import com.amazon.opendistroforelasticsearch.security.securityconf.impl.CType;
 import com.amazon.opendistroforelasticsearch.security.securityconf.impl.SecurityDynamicConfiguration;
+import org.elasticsearch.index.engine.VersionConflictEngineException;
 
 import static org.elasticsearch.common.xcontent.DeprecationHandler.THROW_UNSUPPORTED_OPERATION;
 
@@ -64,22 +66,30 @@ public class ConfigHelper {
     }
 
     public static void uploadFile(Client tc, String filepath, String index, CType cType, int configVersion, boolean populateEmptyIfFileMissing) throws Exception {
-        LOGGER.info("Will update '" + cType + "' with " + filepath + " and populate it with empty doc if file missing and populateEmptyIfFileMissing=" + populateEmptyIfFileMissing);
+        final String configType = cType.toLCString();
+        LOGGER.info("Will update '" + configType + "' with " + filepath + " and populate it with empty doc if file missing and populateEmptyIfFileMissing=" + populateEmptyIfFileMissing);
 
         if (!populateEmptyIfFileMissing) {
             ConfigHelper.fromYamlFile(filepath, cType, configVersion, 0, 0);
         }
-        
+
         try (Reader reader = createFileOrStringReader(cType, configVersion, filepath, populateEmptyIfFileMissing)) {
 
-            final String res = tc
-                    .index(new IndexRequest(index).type(configVersion==1?"security":"_doc").id(cType.toLCString()).setRefreshPolicy(RefreshPolicy.IMMEDIATE)
-                            .source(cType.toLCString(), readXContent(reader, XContentType.YAML))).actionGet().getId();
+            final IndexRequest indexRequest = new IndexRequest(index)
+                    .type(configVersion == 1 ? "security" : "_doc")
+                    .id(configType)
+                    .opType(OpType.CREATE)
+                    .setRefreshPolicy(RefreshPolicy.IMMEDIATE)
+                    .source(configType, readXContent(reader, XContentType.YAML));
+            final String res = tc.index(indexRequest).actionGet().getId();
 
-            if (!cType.toLCString().equals(res)) {
-                throw new Exception("   FAIL: Configuration for '" + cType.toLCString()
+            if (!configType.equals(res)) {
+                throw new Exception("   FAIL: Configuration for '" + configType
                         + "' failed for unknown reasons. Pls. consult logfile of elasticsearch");
             }
+            LOGGER.info("Doc with id '{}' and version {} is updated in {} index.", configType, configVersion, index);
+        } catch (VersionConflictEngineException versionConflictEngineException) {
+            LOGGER.info("Index {} already contains doc with id {}, skipping update.", index, configType);
         }
     }
 


src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
```diff
@@ -130,11 +131,21 @@ public void run() {
...
+                                        LOGGER.info("Security index {} exists.", opendistrosecurityIndex);
+                                    } else {
+                                        LOGGER.info("Security index {} does not exist. Attempting to create new index.", opendistrosecurityIndex);
+                                        existsOrCreated = client.admin().indices().create(new CreateIndexRequest(opendistrosecurityIndex)
+                                                .settings(indexSettings))
+                                                .actionGet().isAcknowledged();
+                                        LOGGER.info("Index {} created?: {}", opendistrosecurityIndex, existsOrCreated);
+                                    }
+
+                                    if (existsOrCreated) {
```
dinusX, line 148: This code will be executed every time you restart an instance, which means that it will always override the existing config with default values, which is not what we want.  
Why not use instead some API to update the config values ?

sujithvm: Yes, this will be executed on every restart. 

It will attempt to create config document with put-if-absent behavior by setting OpType to CREATE in ConfigHelper. 

If config exists, elastic throws a version conflict exception as specified here https://www.elastic.co/guide/en/elasticsearch/client/java-rest/master/java-rest-high-document-index.html#java-rest-high-document-index-response so plugin will simply skip config documents that already exist.

dinusX: How this will help you to load "audit.yml" ? Won't this simply throw an exception during first attempt to put the "config.yml" file ?

sujithvm: Say the existing cluster has documents roles, rolemapping, internalusers ... already loaded from their yml files when the cluster first started. 

Now you want to include audit configuration from audit.yml file. 

With this change, you include the audit.yml file and restart the cluster. 
Updates will fail for roles, rolemapping, internalusers .. (throwing exception) but will pass for audit as the document does not exist in the index yet (Put-If-Absent) behavior.

sujithvm:
@hardik-k-shah I did look into writing tests for this, but found it very hard. Here is my finding:

- A functional test is more suitable than unit test for this functionality. 
- But it is hard to write such a functional test which will `start cluster -> load some config files -> stop cluster -> add new config file -> start cluster again`. 
- The issue is that after `load some config files` say roles.yml, rolemapping.yml but not load internalusers.yml the plugin starts complaining that data is not found for internalusers and file not found if not present in security config directory.  
- I cannot add a new file like audit.yml because its supporting implementation class Audit.java (v6,v7...) and functionality does not exist in the plugin yet and again plugin complains. 
- There are no existing tests around this in the codebase as well. 

I will try to create mocked unit tests around the exists or created logic. 

Any pointers how to proceed ?

sujithvm:
For testing:
- installed opendistro security plugin and started the node
- uninstalled the plugin and install plugin from #409 with audit.yml and started the node

Screen Shot 2020-04-26 at 12 39 41 PM

sujithvm:
> 1) Somebody removed "yml" file. (let's assume rollback in case of audit). Will this impact in any way the boot-up code?
- Works: If rollbacked to a version which does not have yml file changes ie: `ConfigHelper.uploadFile(client, cd+"audit.yml", opendistrosecurityIndex, CType.AUDIT, DEFAULT_CONFIG_VERSION);` as this throws FileNotFoundException
- **Note: With this change, yml files must always be present in securityconfig. They cannot be deleted even after bootstrapping is over. Rolling back to a version which does not originially contain the yml file is fine.**


> 2) Start 10 nodes(or any high number of nodes) at the same time. Test this for 2 use-cases: the very first time the cluster boots (no indices present), 

Setup: 
```
NUM_NODES=10

# extract package for 10 nodes 
for i in $(seq 1 $NUM_NODES) ; do tar xf elasticsearch-oss-7.6.1-darwin-x86_64.tar.gz && mv elasticsearch-7.6.1 elasticsearch-node-$i ;  done

# install plugin (put if absent changes)
for i in $(seq 1 $NUM_NODES); do cd elasticsearch-node-$i ; bin/elasticsearch-plugin install -b file:///Users/svadakke/Projects/OpenDistro/security/target/releases/opendistro_security-1.6.0.0.zip ; sh plugins/opendistro_security/tools/install_demo_configuration.sh -y -i -c ; cd .. ;  done

# Start the nodes
for i in $(seq 1 $NUM_NODES); do cd elasticsearch-node-$i ; (bin/elasticsearch -E node.name=node-$i -E cluster.initial_master_nodes=node-1 -E node.max_local_storage_nodes=$NUM_NODES &)  ; cd .. ; done
```
**Result:**
Works: 
- Started 10 nodes
- Node initially complain data not found eg:  `[2020-04-29T18:28:53,355][WARN ][c.a.o.s.c.ConfigurationLoaderSecurity7] [node-10] No data for tenants while retrieving configuration for [INTERNALUSERS, ACTIONGROUPS, CONFIG, ROLES, ROLESMAPPING, TENANTS, NODESDN]  (index=.opendistro_security and type=_doc)` which is expected. 
- Once primary shard is allocated the document is indexed and written. 
eg: `[2020-04-29T18:28:57,318][INFO ][c.a.o.s.s.ConfigHelper   ] [node-1] Doc 'TENANTS' with version 2 is updated.` 


> 2) Start 10 nodes(or any high number of nodes) at the same time. Test this for 2 use-cases: 
during restart (creation of a new document ex: audit.yml). We don't want to have any issues during concurrent execution of that code.

Setup (similar to previous - but removing and installing plugin with audit.yml changes)
```
for i in $(seq 1 $NUM_NODES); do cd elasticsearch-node-$i ; bin/elasticsearch-plugin remove opendistro_security ; bin/elasticsearch-plugin install -b file:///Users/svadakke/Projects/OpenDistro/security/target/releases/opendistro_security-1.6.0.0.zip ; sh plugins/opendistro_security/tools/install_demo_configuration.sh -y -i -c ; cd .. ;  done
```

**Result:**
Works:
- Single update seen.
`[2020-04-29T18:56:57,567][INFO ][c.a.o.s.s.ConfigHelper   ] [node-4] Doc 'AUDIT' with version 2 is updated.`
- Other nodes 
`[2020-04-29T18:56:57,543][INFO ][c.a.o.s.s.ConfigHelper   ] [node-8] Doc 'AUDIT' with version 2 or higher already exists. Doing nothing.
][c.a.o.s.c.ConfigurationRepository] [node-3] Default config applied`

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
```diff
@@ -130,11 +131,21 @@ public void run() {
                                     indexSettings.put("index.number_of_shards", 1);
                                     indexSettings.put("index.auto_expand_replicas", "0-all");
 
-                                    boolean ok = client.admin().indices().create(new CreateIndexRequest(opendistrosecurityIndex)
-                                            .settings(indexSettings))
-                                            .actionGet().isAcknowledged();
-                                    LOGGER.info("Index {} created?: {}", opendistrosecurityIndex, ok);
-                                    if(ok) {
+                                    boolean existsOrCreated = client.admin().indices()
```
vrozov, line 134: What is the benefit of checking if the index already exists? Now there will be possibly two requests instead of one when one seems to provide exactly the same functionality. The `create` request should not be much more heavyweight compared to the `does exist` request to try to avoid it.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
```diff
@@ -147,7 +158,7 @@ public void run() {
                                         ConfigHelper.uploadFile(client, cd+"nodes_dn.yml", opendistrosecurityIndex, CType.NODESDN, DEFAULT_CONFIG_VERSION, populateEmptyIfFileMissing);
                                         LOGGER.info("Default config applied");
                                     } else {
-                                        LOGGER.error("Can not create {} index", opendistrosecurityIndex);
+                                        LOGGER.error("Can not use or create {} index", opendistrosecurityIndex);
```
vrozov, line 161: Is there a way to get an error message with the details of why the index can't be created? There is no yet attempt to "use" the index, is the error message should be "Failed to create {} index"?

src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigHelper.java
```diff
@@ -71,13 +73,16 @@ public static void uploadFile(Client tc, String filepath, String index, CType cT
...
+                    .index(new IndexRequest(index).type(configVersion==1?"security":"_doc").id(cType.toLCString()).opType(DocWriteRequest.OpType.CREATE).setRefreshPolicy(RefreshPolicy.IMMEDIATE)
                             .source(cType.toLCString(), readXContent(reader, XContentType.YAML))).actionGet().getId();
 
             if (!cType.toLCString().equals(res)) {
                 throw new Exception("   FAIL: Configuration for '" + cType.toLCString()
                         + "' failed for unknown reasons. Pls. consult logfile of elasticsearch");
             }
+            LOGGER.info("Doc '{}' with version {} is updated.", cType, configVersion);
+        } catch (VersionConflictEngineException versionConflictEngineException) {
+            LOGGER.warn("Doc with id {} in index {} already exists. Doing nothing.", cType.toLCString(), index);
```
vrozov, line 85: IMO, the warning level is not appropriate as it is not a misconfiguration or something went wrong. On a multi-node cluster, only one node will update the index, all other nodes are supposed to fail and would log the warning message.

sujithvm: What do you think is the appropriate level ? 
IMO the node tried to do something and failure is expected so put as warning. These logs are on the node level.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
sujithvm: Do you mean to always try creating and catch the failure as index exists case? 
Only the exists condition is going to be executed after the index is created. 
IMO why do something that we know node will always cry about and will throw an error ?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
sujithvm: Well I just wanted to mean that "not ok" means, it could not attempt to use the index to update the documents. 
I don't mind changing the grammar. Let me know

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
vrozov: Is there already a check that the index does not exist on line 235 before `installDefaultConfig` is set to `true`? Is this additional check prevents race condition between nodes or it is a different check?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
vrozov: Do you mean to say "Request to create index {} is not acknowledged. The index may not be usable to load dynamic configuration."?

src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigHelper.java
vrozov: It depends on who is the target of the message. If it is intended for ES admin, it should be INFO (but admin should be able to do something, like delete index or upload it manually) or DEBUG, if it is intended for plugin developers.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
sujithvm: ES resolves concurrency for document updates amongst the nodes via document versioning.
So that looks like some different check (maybe threads within the same node).

vrozov: Based on the comments the code on line 235 checks if the index exists (but does it using different API). What is the difference between using `clusterService.state().metaData().hasConcreteIndex(opendistrosecurityIndex)` and `client.admin().indices().exists(new IndicesExistsRequest(opendistrosecurityIndex))`? As an attempt to create the index would be done only if `installDefaultConfig` is `true`, so `clusterService.state().metaData().hasConcreteIndex(opendistrosecurityIndex)` must be `false`, what is the purpose of `client.admin().indices().exists(new IndicesExistsRequest(opendistrosecurityIndex))` check?

sujithvm: IndicesExistsRequest uses admin client to check if index exists. Maybe at some point the index was not concrete index and that api was used or need admin client 🤔 I did try with `clusterService.state().metaData().hasConcreteIndex(opendistrosecurityIndex)` and it returns false

vrozov: I still don't understand the difference between using admin client and cluster service state metadata (Yes, I understand that the former one is a network call while the last one uses in-memory state. But should both be consistent and if not, why one will be different than the other?)

In addition, `exists()` and `IndicesExistsRequest` are removed from the latest ES. IMO, we should try to avoid introducing a dependency on it now.

Please check other ES plugins to see how they check whether or not an index exists and create it.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
```diff
@@ -136,27 +133,30 @@ public void run() {
...
+                                    final String cd = lookupDir != null ? (lookupDir + "/") : new Environment(settings, configPath).pluginsFile().toAbsolutePath().toString() + "/opendistro_security/securityconfig/";
+
+                                    ConfigHelper.uploadFile(client, cd + "config.yml", opendistrosecurityIndex, CType.CONFIG, DEFAULT_CONFIG_VERSION);
+                                    ConfigHelper.uploadFile(client, cd + "roles.yml", opendistrosecurityIndex, CType.ROLES, DEFAULT_CONFIG_VERSION);
+                                    ConfigHelper.uploadFile(client, cd + "roles_mapping.yml", opendistrosecurityIndex, CType.ROLESMAPPING, DEFAULT_CONFIG_VERSION);
+                                    ConfigHelper.uploadFile(client, cd + "internal_users.yml", opendistrosecurityIndex, CType.INTERNALUSERS, DEFAULT_CONFIG_VERSION);
+                                    ConfigHelper.uploadFile(client, cd + "action_groups.yml", opendistrosecurityIndex, CType.ACTIONGROUPS, DEFAULT_CONFIG_VERSION);
+                                    ConfigHelper.uploadFile(client, cd + "tenants.yml", opendistrosecurityIndex, CType.TENANTS, DEFAULT_CONFIG_VERSION);
+
```
hardik-k-shah, line 152: Do we need to add audit.yml here? 
or it is cover in separate PR?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
```diff
@@ -213,19 +213,27 @@ public void run() {
 
     }
 
-    public void initOnNodeStart() {
+    private boolean indexExists(String index) {
+        GetIndexResponse getIndexResponse = client.admin().indices()
```
hardik-k-shah, line 217: why are we moving to client.admin() from clusterService.state().metaData().hasConcreteIndex(opendistrosecurityIndex)

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
sujithvm: Yes. It is going to be covered in #409

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
sujithvm: clusterService.state() maintains a cached copy of the cluster state. 

To reproduce the behavior: 
- install plugin 
- start elasticsearch
- restart elasticsearch
- restart elasticsearch

When it starts up, the cluster state is not updated correctly because of which plugin thinks the index does not exist. 
You would see this printed everytime coming from [here](https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java#L223)
```
[2020-06-18T11:19:36,804][INFO ][c.a.o.s.c.ConfigurationRepository] [smoketestnode] .opendistro_security index does not exist yet, so we create a default config
```

Doing the GetIndexResponse would query the ES cluster and get the status whether the index exists or not.

sujithvm: Why we dont see the impact is because this log level is debug [here](https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java#L159) 

If you change this to error or info, you would that its throwing error every time in opendistro.

vrozov:
Please split this PR into 2:
- Fix for the index exists check. I am not even sure that such check is necessary. Consider "create index if absent" approach  similar to "create document (config) if absent"
- Create config if absent

sujithvm:
Changes:
- removed the index exists check
- add an exception handler to identify index exists. Index will be created if absent
- Put document configuration if absent

src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigHelper.java
```diff
@@ -64,22 +66,26 @@ public static void uploadFile(Client tc, String filepath, String index, CType cT
...
 
-            if (!cType.toLCString().equals(res)) {
-                throw new Exception("   FAIL: Configuration for '" + cType.toLCString()
+            if (!configType.equals(res)) {
+                throw new Exception("   FAIL: Configuration for '" + configType
                         + "' failed for unknown reasons. Pls. consult logfile of elasticsearch");
             }
+            LOGGER.info("Doc with id '{}' and version {} is updated in {} index.", configType, configVersion, index);
+        } catch (VersionConflictEngineException versionConflictEngineException) {
+            LOGGER.info("Doc with id '{}' already exists in {} index. Doing nothing.", configType, index);
```
vrozov, line 88: nit: "Index {} already contains doc with id {}, skipping update."

sujithvm:
Will use this PR for ConfigHelper changes and will raise a separate PR for ConfigurationRepository.

PR #409, opened 2020-04-23
Hot reloading audit configuration

sujithvm:
**Hot-reloading for audit configuration:**

1) Users can hot reload audit configuration (audit config filter + compliance config)
2) Initial configuration can be specified in securityconfig/audit.yml file 
```
_meta:
  type: "audit"
  config_version: 2

config:
  # turn on-off audit logging
  enabled: true
  audit:
    # rest
    enable_rest: true
    disabled_rest_categories:
      - AUTHENTICATED
      - GRANTED_PRIVILEGES

    # transport
    enable_transport: true
    disabled_transport_categories:
      - AUTHENTICATED
      - GRANTED_PRIVILEGES

    # ignore
    ignore_users:
      - kibanaserver
    ignore_requests: []

    # verbose attributes
    resolve_bulk_requests: false
    log_request_body: true
    resolve_indices: true
    exclude_sensitive_headers: true

  compliance:
    # turn on-off compliance
    enabled: true

    # configs
    internal_config: true
    external_config: true

    # compliance read
    read_metadata_only: true
    read_watched_fields: []
    read_ignore_users: []

    # compliance write
    write_metadata_only: true
    write_log_diffs: false
    write_watched_indices: []
    write_ignore_users: []

```
3) Users can use REST API for updating the config. Blocking updates to certain fields is not implemented yet in this PR
```
GET _opendistro/_security/api/audit/
PATCH _opendistro/_security/api/audit/
PUT _opendistro/_security/api/audit/config
```

Class components for reviewing:
1) AuditConfig (underlying configuration class which is serialized / deserialized when written to or read from config). This can be V6, V7 (note: its same for audit)
2) AuditModel and AuditModelImpl (interfaces with Audit class). Mainly used for interfacing v6, v7 .. versions
3) AuditApiAction (rest handler for audit configuration)
4) AuditValidator (basic validator for rest api)
5) AuditLog and AuditLogImpl (subscribes to and listens for audit config changes). All other components fetch config from AuditLog
6) AuditMigrator (packaged tool similar to securityAdmin to migrate elasticsearch.yml settings to audit.yml)

Other main updates:
- Updated all tests 
- Updated the supporting classes ie: SecurityAdmin.


src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
```diff
@@ -188,9 +189,7 @@
     private final boolean sslOnly;
     private final List<String> demoCertHashes = new ArrayList<String>(3);
     private volatile OpenDistroSecurityFilter odsf;
-    private volatile Environment environment;
     private volatile IndexResolverReplacer irr;
-    private AtomicBoolean externalConfigLogged = new AtomicBoolean();
```
sujithvm, line 193: Moving environment and externalConfigLogged to AuditLogImpl to consolidate audit logging changes. 
When node starts, configuration may not be loaded yet to log external configuration and this OpenDistroSecurityPlugin must be subscribed as well. To avoid doing that, moved the functionality to AuditLogImpl

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
```diff
@@ -514,14 +513,8 @@ public void onIndexModule(IndexModule indexModule) {
 
                 final ComplianceIndexingOperationListener ciol;
 
-                assert complianceConfig!=null:"compliance config must not be null here";
-                
-                if(complianceConfig.writeHistoryEnabledForIndex(indexModule.getIndex().getName())) {
-                    ciol = ReflectionHelper.instantiateComplianceListener(Objects.requireNonNull(auditLog));
-                    indexModule.addIndexOperationListener(ciol);
-                } else {
-                    ciol = new ComplianceIndexingOperationListener();
```
sujithvm, line 523: complianceConfig may be null when IndexModule is called for .opendistro_security index. 
ComplianceIndexingOperationListenerImpl checks for null hence removing this logic.

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
```diff
@@ -896,22 +892,6 @@ public Settings additionalSettings() {
             settings.add(Setting.groupSetting(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_ENDPOINTS + ".",  Property.NodeScope));
             settings.add(Setting.intSetting(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_THREADPOOL_SIZE, 10, Property.NodeScope, Property.Filtered));
             settings.add(Setting.intSetting(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_THREADPOOL_MAX_QUEUE_LEN, 100*1000, Property.NodeScope, Property.Filtered));
-            settings.add(Setting.boolSetting(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_LOG_REQUEST_BODY, true, Property.NodeScope, Property.Filtered));
-            settings.add(Setting.boolSetting(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_RESOLVE_INDICES, true, Property.NodeScope, Property.Filtered));
```
sujithvm, line 900: Removing registration of audit logging params. These must be migrated to new audit.yml. 
Cluster startup will fail if they are provided. This is desired as part of deprecating the old way of specifying config in elasticsearch.yml.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditLogImpl.java
```diff
@@ -94,126 +97,155 @@ public void run() {
 
 	}
 
+	@Subscribe
+    public void onAuditModelChanged(final AuditModel auditModel) {
```
sujithvm, line 101: Subscribe for changes

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -206,6 +204,36 @@ public static ComplianceConfig from(Settings settings) {
...
+     * index - the index used for storing audit logs to avoid monitoring it.
+     * @param auditModel audit model
+     * @param settings settings
+     * @return ComplianceConfig
+     */
+    public static ComplianceConfig from(AuditModel auditModel, Settings settings) {
+        final String opendistrosecurityIndex = settings.get(ConfigConstants.OPENDISTRO_SECURITY_CONFIG_INDEX_NAME, ConfigConstants.OPENDISTRO_SECURITY_DEFAULT_CONFIG_INDEX);
+        final String type = settings.get(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_TYPE_DEFAULT, null);
+        final String index = settings.get(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX + ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ES_INDEX, "'security-auditlog-'YYYY.MM.dd");
+
```
sujithvm, line 221: These 3 params 
- opendistrosecurityIndex
- type
- index
 are fed from elasticsearch.yml. 
They are not reloadable but they are needed to determine if internal index is being updated or the current index is storing audit logs if destination is set to internal elasticsearch (ie: dont audit log an index used for storing audit logs) .

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiAction.java
```diff
@@ -0,0 +1,156 @@
...
+ *     "write_log_diffs" : false,
+ *     "write_watched_indices" : [ ],
+ *     "write_ignore_users" : [ ],
+ *     "salt" : "aaaabbbbccccdddd"
+ *   }
+ *
+ * PATCH _opendistro/_security/api/audit/config
+ * [{"op": "replace", "path": "/config/enable_rest", "value": "true"}]
+ */
+public class AuditApiAction extends PatchableResourceApiAction {
```
sujithvm, line 99: Rest handler for audit configuration

securityconfig/audit.yml
```diff
@@ -0,0 +1,49 @@
+_meta:
+  type: "audit"
+  config_version: 2
```
vrozov, line 3: Is there `config_version` 1?

legacy/securityconfig_v6/audit.yml
```diff
@@ -0,0 +1,45 @@
+config:
```
vrozov, line 1: Is there a plan to backport the functionality to v6?

securityconfig/audit.yml
```diff
@@ -0,0 +1,49 @@
+_meta:
```
vrozov, line 1: What is the migration path from prior versions?

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/impl/AuditModel.java
```diff
@@ -0,0 +1,52 @@
+package com.amazon.opendistroforelasticsearch.security.securityconf.impl;
```
vrozov, line 1: Please clarify the object model. I don't see why it is necessary to introduce `AuditModel` interface with `AuditModelImpl` implementation that delegates everything to `Audit`.

sujithvm: I have followed the pattern for InternalUsersModel, ConfigModel ..

The underlying implementation class Audit becomes version specific. There can be AuditV6, AuditV7, AuditV8 ... 
The Model class abstracts that. However for ES 6xx and 7xx we are going to use the same underlying class Audit

If the underlying class / document structure is going to be changed in future say something changes in ES 8 or we add new structure, this becomes handy to abstract that out.

legacy/securityconfig_v6/audit.yml
sujithvm: Yes. This will be backported to ES 6xx as well

securityconfig/audit.yml
sujithvm: We currently have migration path via 6.xx -> 7.1 
There is migrate api in opendistro-1.1 branch https://github.com/opendistro-for-elasticsearch/security/blob/opendistro-1.1/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/MigrateApiAction.java 

Security Admin can also be used to migrate the document. 
However the only migration needed will be add _meta field for Audit same as for NodesDn

securityconfig/audit.yml
sujithvm: 7xx have followed config_version 2 for all internal documents in security config. 
6xx have used 1 (or none)

securityconfig/audit.yml
vrozov: What is the migration path for the plugin itself? How prior installations that do not have `audit.yml` will be upgraded to installations that require `audit.yml`?

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/impl/AuditModel.java
vrozov: I don't think this is necessary. When needed the class hierarchy can be changed, especially that 6.x model should be the same as 7.x model and it is not clear whether 8.x model will be different.

securityconfig/audit.yml
sujithvm: I have separated that out in this PR https://github.com/opendistro-for-elasticsearch/security/pull/402

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/impl/AuditModel.java
sujithvm: I agree it is not necessary at the moment but implemented it for consistency with other security configs and also to make it easier if the doc structure were to change. Even the recent NodeDn reload was done in a similar way https://github.com/krishnaggk/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigFactory.java#L330 

Let me know if you would like me to remove it.

vrozov: Yes, I'd like to remove it. Please see https://github.com/vrozov/security/tree/auditlog

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AbstractConfigurationValidator.java
```diff
@@ -282,9 +282,15 @@ private void addErrorMessage(final XContentBuilder builder, final String message
...
-        PAYLOAD_MANDATORY("Request body required for this action."), SECURITY_NOT_INITIALIZED("Security index not initialized");
+        NONE("ok"),
+        INVALID_CONFIGURATION("Invalid configuration"),
+        INVALID_PASSWORD("Invalid password"),
+        WRONG_DATATYPE("Wrong datatype"),
+        BODY_NOT_PARSEABLE("Could not parse content of request."),
+        PAYLOAD_NOT_ALLOWED("Request body not allowed for this action."),
+        PAYLOAD_MANDATORY("Request body required for this action."),
+        SECURITY_NOT_INITIALIZED("Security index not initialized"),
+        INVALID_SALT("Salt must be atleast 16 chars long. If greater, only the first 16 bytes will used for salting.");
```
vrozov, line 293: typo: "at least". Will it be better to use "Salt must be 16 bytes long". Note that 16 chars != 16 bytes.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AuditValidator.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    public boolean validate() {
+        if (!super.validate()) {
+            return false;
+        }
+
+        if ((request.method() == RestRequest.Method.PUT || request.method() == RestRequest.Method.PATCH)
+                && this.content != null
+                && this.content.length() > 1) {
```
vrozov, line 77: content length 1 is probably not valid, but content length 2 is probably not valid either, should it be `this.content.length() > 0` to distinguish from not empty content?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AuditValidator.java
```diff
@@ -0,0 +1,123 @@
...
+                    }
+                    // validate salt
+                    if (!validateSalt(contentAsMap)) {
+                        this.errorType = ErrorType.INVALID_SALT;
+                        return false;
+                    }
+                }
+            } catch (NotXContentException e) {
+                //this.content is not valid json/yaml
+                log.error("Invalid xContent: " + e, e);
```
vrozov, line 94: Use smart logging. Do you mean to log invalid content or exception? Exception will be logged anyway.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AuditValidator.java
```diff
@@ -0,0 +1,123 @@
...
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    private boolean validateAuditCategories(final Map<String, Object> contentAsMap) {
+        for (String key: ImmutableList.of(Audit.Key.DISABLED_REST_CATEGORIES, Audit.Key.DISABLED_TRANSPORT_CATEGORIES)) {
```
vrozov, line 103: nit: consider using Java streams API.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AuditValidator.java
```diff
@@ -0,0 +1,123 @@
...
+            try {
+                if (contentAsMap.containsKey(key)) {
+                    AuditCategory.parse((Collection<String>) contentAsMap.get(key));
```
vrozov, line 106: Avoid double `get()`

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AuditValidator.java
```diff
@@ -0,0 +1,123 @@
...
+                log.error("Could not parse body of key {}", key);
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean validateSalt(final Map<String, Object> contentAsMap) {
+        if (contentAsMap.containsKey(Audit.Key.SALT)) {
+            final String salt = (String) contentAsMap.get(Audit.Key.SALT);
```
vrozov, line 118: Avoid double `get()`.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AbstractConfigurationValidator.java
sujithvm: We are using UTF-8 encoding for converting to bytes right so each char should be 1 byte ? 
Also this is something that will be returned back to user so thought that user will understand "chars" usage better than "bytes"

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AuditValidator.java
sujithvm: Yes.` log.error("Invalid content passed in the request", e);` should be sufficient.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AbstractConfigurationValidator.java
vrozov: Only 127 chars can be converted to a single byte using UTF-8. I guess it is not valid to assume that only ASCII will be used for the salt.

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
vrozov: Other than being consistent with `roles.yml`, `internal_users.yml`, what is the benefit of deprecating the old way of specifying audit config settings?

sujithvm: Audit configuration will always be loaded from index. User might be confused why the settings are different/ignored from elasticsearch.yml it is starting up with.

vrozov: Does the same apply to audit.yml? Users might be confused that the actual settings are different/ignored. IMO, it will be better to gradually deprecate those settings.

sujithvm: I feel the users will be assume the same behavior as others roles.yml, internal_users.yml ...

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
```diff
@@ -514,14 +511,8 @@ public void onIndexModule(IndexModule indexModule) {
...
                 final ComplianceIndexingOperationListener ciol;
 
-                assert complianceConfig!=null:"compliance config must not be null here";
-                
-                if(complianceConfig.writeHistoryEnabledForIndex(indexModule.getIndex().getName())) {
-                    ciol = ReflectionHelper.instantiateComplianceListener(Objects.requireNonNull(auditLog));
-                    indexModule.addIndexOperationListener(ciol);
-                } else {
-                    ciol = new ComplianceIndexingOperationListener();
-                }
```
sujithvm, line 524: Covered in #491

legacy/securityconfig_v6/audit.yml
```diff
@@ -0,0 +1,46 @@
...
```
vrozov, line 1: auduit.yml.example?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/routing/AuditMessageRouter.java
```diff
@@ -41,138 +45,148 @@
...
+            }
+        } else {
+            storagePool.submit(msg, sink);
+            if (log.isTraceEnabled()) {
+                log.trace("will store on sink {} asynchronously", sink.getClass().getSimpleName());
+            }
+        }
+    }
+
+}
```
vrozov, line 192: Add new line

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -388,4 +381,4 @@ public boolean readHistoryEnabledForField(String index, String field) {
         }
         return matcher.test(field);
     }
-}
+}
```
vrozov, line 384: undo the change

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/TracingTests.java
```diff
@@ -49,15 +50,21 @@ public void testHTTPTrace() throws Exception {
...
-                .put(ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_READ_WATCHED_FIELDS, "*")
-                .put(ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_WRITE_WATCHED_INDICES, "*")
-                .put("opendistro_security.audit.resolve_bulk_requests", true)
                 .put("opendistro_security.audit.config.log4j.logger_name", "opendistro_security_action_trace")
                 .put("opendistro_security.audit.config.log4j.level", "TRACE")
                 .build();
 
         setup(Settings.EMPTY, new DynamicSecurityConfig(), settings, true, ClusterConfiguration.DEFAULT);
 
+        Settings auditSettings = Settings.builder()
```
vrozov, line 59: Is this change necessary?

tools/audit_config_migrater.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
vrozov, line 1: file mask should be the same as other .sh files.

src/test/resources/nodes_dn.yml
```diff
@@ -0,0 +1,4 @@
+---
```
vrozov, line 1: Is this change necessary?

src/test/java/com/amazon/opendistroforelasticsearch/security/test/DynamicSecurityConfig.java
```diff
@@ -50,6 +50,7 @@
     private String securityInternalUsers = "internal_users.yml";
     private String securityActionGroups = "action_groups.yml";
     private String securityNodesDn = "nodes_dn.yml";
+    private String securityAudit = "audit.yml";
```
vrozov, line 53: nit: static

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/AbstractAuditlogiUnitTest.java
```diff
@@ -105,9 +116,22 @@ protected boolean validateJson(final String json) {
...
+
+    protected void updateAuditConfig(final Settings settings) throws Exception {
+        updateAuditConfig(AuditTestUtils.createAuditPayload(settings));
+    }
+
+    protected void updateAuditConfig(final String payload) throws Exception {
+        final boolean sendAdminCertificate = rh.sendAdminCertificate;
+        final String keystore = rh.keystore;
+        rh.sendAdminCertificate = true;
+        rh.keystore = "auditlog/kirk-keystore.jks";
```
vrozov, line 131: Why is it OK to use this file?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -214,31 +397,368 @@ public void log(Logger logger) {
...
+            ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_WRITE_WATCHED_INDICES
+    );
+
+    public static Set<String> getDeprecatedKeys(final Settings settings) {
+        return AuditConfig.DEPRECATED_KEYS
+                .stream()
+                .filter(settings::hasValue)
+                .collect(Collectors.toSet());
+    }
+}
```
vrozov, line 764: nit: add new line

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/AbstractAuditlogiUnitTest.java
sujithvm: This is an admin cert used only in the tests to update the configuration.

src/test/java/com/amazon/opendistroforelasticsearch/security/test/DynamicSecurityConfig.java
sujithvm: Class usage is nit static and this member is modified by setSecurityAudit() function defined below

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -162,38 +158,35 @@ public void log(Logger logger) {
...
-                logExternalConfig,
-                logInternalConfig,
-                logReadMetadataOnly,
-                watchedReadFields,
-                ignoredComplianceUsersForRead,
-                logWriteMetadataOnly,
-                logDiffsForWrite,
-                watchedWriteIndices,
-                ignoredComplianceUsersForWrite,
+                configCompliance.isComplianceEnabled(),
```
vrozov, line 180: This constructor should take `configCompliance` and store it as a member variable instead of copying each individual value.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -214,31 +397,368 @@ public void log(Logger logger) {
...
+        public void setWriteIgnoreUsers(Set<String> writeIgnoreUsers) {
+            if (writeIgnoreUsers != null) {
+                this.writeIgnoreUsers = writeIgnoreUsers;
+            } else {
+                this.readIgnoreUsers = Collections.emptySet();
+            }
+        }
+    }
+
+    public static class Key {
```
vrozov, line 623: IMO, it will be better to move those strings to `Filter` or `Compliance` classes instead of sharing them in a single class. It is OK to do this in a follow up PR.

src/test/resources/nodes_dn.yml
sujithvm: There are tests in src/test/java/com/amazon/opendistroforelasticsearch/security/SecurityAdminTests.java referencing securityconfig in main package instead of tests resources.

node_dn file was missing in the test resources hence added it.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationLoaderSecurity7.java
```diff
@@ -84,6 +87,14 @@
         log.debug("Index is: {}", securityIndex);
     }
 
+    /**
+     * Checks if audit config doc is present in security index
+     * @return true/false
+     */
+    boolean isAuditConfigDocPresentInIndex() {
+        return isAuditConfigDocPresentInIndex.get();
```
vrozov, line 95: How is it enforced that `isAuditConfigDocPresentInIndex` is initialized before `isAuditConfigDocPresentInIndex()` is called?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
```diff
@@ -195,6 +204,21 @@ public void run() {
                         }
                     }
 
+                    final Set<String> deprecatedAuditKeysInSettings = AuditConfig.getDeprecatedKeys(settings);
+                    if (!deprecatedAuditKeysInSettings.isEmpty()) {
+                        LOGGER.warn("Following keys {} are deprecated in elasticsearch settings. They will be removed in plugin v2.0.0.0", deprecatedAuditKeysInSettings);
+                    }
+                    final boolean isAuditConfigDocPresentInIndex = cl.isAuditConfigDocPresentInIndex();
+                    if (isAuditConfigDocPresentInIndex) {
+                        if (!deprecatedAuditKeysInSettings.isEmpty()) {
```
vrozov, line 212: What is the final agreement on the case where audit entries are present both in the index and elasticsearch.yml?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
```diff
@@ -377,4 +405,4 @@ private static String formatDate(long date) {
     public static int getDefaultConfigVersion() {
         return ConfigurationRepository.DEFAULT_CONFIG_VERSION;
     }
-}
+}
```
vrozov: nit: add new line

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/AbstractAuditlogiUnitTest.java
vrozov: It comes from "auditlog", what if another cert is used in a different unit test? If it is always the same, why it comes from "auditlog"?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiAction.java
```diff
@@ -0,0 +1,185 @@
...
+        if (!cl.isAuditHotReloadingEnabled()) {
+            notImplemented(channel, request.method());
+            return;
+        }
+        super.handleApiRequest(channel, request, client);
+    }
+
+    @Override
+    protected void handlePut(final RestChannel channel, final RestRequest request, final Client client, final JsonNode content) throws IOException {
+        if (!"config".equals(request.param("name"))) {
```
vrozov, line 149: Use `RESOURCE_NAME`

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/ValidateApiAction.java
```diff
@@ -96,13 +98,17 @@ protected void handleGet(RestChannel channel, RestRequest request, Client client
             final SecurityDynamicConfiguration<InternalUserV6> internalUsersV6 = (SecurityDynamicConfiguration<InternalUserV6>) load(CType.INTERNALUSERS, true, acceptInvalid);
             final SecurityDynamicConfiguration<RoleV6> rolesV6 = (SecurityDynamicConfiguration<RoleV6>) load(CType.ROLES, true, acceptInvalid);
             final SecurityDynamicConfiguration<RoleMappingsV6> rolesmappingV6 = (SecurityDynamicConfiguration<RoleMappingsV6>) load(CType.ROLESMAPPING, true, acceptInvalid);
+            final SecurityDynamicConfiguration<NodesDn> nodesDnV6 = (SecurityDynamicConfiguration<NodesDn>) load(CType.NODESDN, true);
```
vrozov, line 101: Can you move this to a separate PR, I am not sure why `NodesDn` were not included before.

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/compliance/ComplianceAuditlogTest.java
```diff
@@ -86,6 +93,36 @@ public void testSourceFilter() throws Exception {
...
     }
 
+    @Test
+    public void testComplianceEnable() throws Exception {
+        Settings additionalSettings = Settings.builder()
+                .put("opendistro_security.audit.type", TestAuditlogImpl.class.getName())
+                .build();
+
+        setup(additionalSettings);
+        rh.sendAdminCertificate = true;
```
vrozov, line 103: Order in which tests are executed by JUnit is not defined, this may be the first one to be executed and it will set `sendAdminCertificate` to `true`.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationLoaderSecurity7.java
sujithvm: As part of bootstrapping the configs are reloaded https://github.com/sujithvm/security/blob/auditlog/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java#L192

This would ensure whether doc is present in the index or not.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
sujithvm: Warn the user that settings are present at both places but continue using from index. Do not fail startup.  
Index takes priority over elasticsearch.yml.

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/TracingTests.java
```diff
@@ -332,11 +333,17 @@ public void uncaughtException(Thread t, Throwable e) {
     @Test
     public void testAdvancedMapping() throws Exception {
         Settings settings = Settings.builder()
-                .put(ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_READ_WATCHED_FIELDS, "*")
```
vrozov, line 335: Is the change necessary? Can't it rely on `settings` to configure audit in the test not directly related to it?

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiActionTest.java
```diff
@@ -0,0 +1,285 @@
...
+    public void testApi() throws Exception {
+        setupWithRestRoles();
+        rh.keystore = "restapi/kirk-keystore.jks";
+
+        // admin cred with roles in test yml files
+        final Header adminCredsHeader = encodeBasicHeader("sarek", "sarek");
+        // non-admin
+        final Header nonAdminCredsHeader = encodeBasicHeader("random", "random");
+
+        {
```
vrozov, line 90: Why java block (`{ ... }`) is necessary?

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiActionTest.java
```diff
@@ -0,0 +1,285 @@
...
+            // No creds, no admin certificate - UNAUTHORIZED
+            rh.sendAdminCertificate = false;
+            testActions(HttpStatus.SC_UNAUTHORIZED);
```
vrozov, line 93: nit: take desired `sendAdminCertificate` as a parameter for `testActions`.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
vrozov: OK, just want to be sure that it is the final agreement and it is documented

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationLoaderSecurity7.java
```diff
@@ -101,6 +112,11 @@ public void success(SecurityDynamicConfiguration<?> dConf) {
                 if(log.isDebugEnabled()) {
                     log.debug("Received config for {} (of {}) with current latch value={}", dConf.getCType().toLCString(), Arrays.toString(events), latch.getCount());
                 }
+                // Audit configuration doc is available in the index.
+                // Configuration can be hot-reloaded.
+                if (dConf.getCType() == CType.AUDIT) {
+                    isAuditConfigDocPresentInIndex.set(true);
```
vrozov, line 118: `isAuditConfigDocPresentInIndex` must be set *before* `latch` is updated to avoid possibility of the waiting thread using not updated state.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
sujithvm: Done!

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/AbstractAuditlogiUnitTest.java
sujithvm: Current usage of tests & associated test resources seem to be done this way. Tests in auditlog package are referencing the test resources (certs & config) present in the associated resources/auditlog directory.

securityconfig/audit.yml.example
```diff
@@ -0,0 +1,52 @@
...
+    # enable/disable compliance
+    enabled: true
+
+    # configs
+    internal_config: true
+    external_config: false
+
+    # compliance read
+    read_metadata_only: true
+    read_watched_fields: {}
```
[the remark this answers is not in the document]
sujithvm: read_watched_fields is going to be a map because fields are involved. 
```
read_watched_fields : {
  "indexName1": ["field1", "field2"],
  "indexPattern": ["fields"]
}
```
whereas write_watched_indices is just list of indexes. 
```
write_watched_indices: ["indexName1", "indexPattern"]
```

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/AuditLog.java
```diff
@@ -72,10 +73,12 @@
     void logDocumentRead(String index, String id, ShardId shardId, Map<String, String> fieldNameValues);
     void logDocumentWritten(ShardId shardId, GetResult originalIndex, Index currentIndex, IndexResult result);
     void logDocumentDeleted(ShardId shardId, Delete delete, DeleteResult result);
-    void logExternalConfig(Settings settings, Environment environment);
```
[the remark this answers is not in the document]
sujithvm: https://github.com/opendistro-for-elasticsearch/security/pull/409/files#diff-5f0cc3e225606da6800e9edc758caf36R125 posts the external config. This is done within the implementation and hence removed it from the interface.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -1,167 +1,363 @@
...
+        @JsonProperty(value = Key.DISABLED_REST_CATEGORIES)
+        private EnumSet<AuditCategory> disabledRestCategories = DEFAULT_DISABLED_CATEGORIES;
+        @JsonProperty(value = Key.ENABLE_TRANSPORT)
+        private boolean isTransportApiAuditEnabled = true;
+        @JsonProperty(value = Key.DISABLED_TRANSPORT_CATEGORIES)
+        private EnumSet<AuditCategory> disabledTransportCategories = DEFAULT_DISABLED_CATEGORIES;
+        @JsonProperty(value = Key.RESOLVE_BULK_REQUESTS)
+        private boolean resolveBulkRequests = false;
+        @JsonProperty(value = Key.LOG_REQUEST_BODY)
+        private boolean logRequestBody = true;
```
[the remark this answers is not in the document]
sujithvm: Current setting was true incase it was missing in elasticsearch.yml . Can change this if needed.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -214,31 +397,368 @@ public void log(Logger logger) {
...
-                    ", ignoredAuditUsers=" + ignoredAuditUsersMatcher +
-                    ", ignoreAuditRequests=" + ignoredAuditRequestsMatcher +
-                    '}';
+            logger.info("Auditing requests from {} users is disabled.", ignoredAuditUsers);
+        }
+    }
+
+    public static class Compliance {
+        @JsonProperty(value = Key.ENABLED)
+        private boolean complianceEnabled = true;
```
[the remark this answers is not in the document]
sujithvm: Current behavior is setting it to true always https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java#L84 . Will review the defaults once more

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -214,31 +397,368 @@ public void log(Logger logger) {
...
+        @JsonProperty(value = Key.INTERNAL_CONFIG_ENABLED)
+        private boolean internalConfigEnabled = true;
```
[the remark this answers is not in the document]
sujithvm: Yes internal_config_enalbed logs changes to security index. Since audit logging is also stored in the index, changes will be logged. However if disabled then config changes are not recorded. Will whitelist to record audit config changes always.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -214,31 +397,368 @@ public void log(Logger logger) {
...
+        }
+
+        /**
+         * Set users to be ignored for auditing doc read requests
+         * @param readIgnoreUsers users
+         */
+        @JsonSetter(value = Key.READ_IGNORE_USERS, nulls = Nulls.AS_EMPTY)
+        public void setReadIgnoreUsers(Set<String> readIgnoreUsers) {
+            if (readIgnoreUsers != null) {
+                this.readIgnoreUsers = readIgnoreUsers;
```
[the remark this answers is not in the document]
sujithvm: Patterns are supported. Will rename the variable

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -214,31 +397,368 @@ public void log(Logger logger) {
...
+        }
+
+        /**
+         * Set indices to watch for doc write requests
+         * @param writeWatchedIndices indices
+         */
+        @JsonSetter(value = Key.WRITE_WATCHED_INDICES, nulls = Nulls.AS_EMPTY)
+        public void setWriteWatchedIndices(List<String> writeWatchedIndices) {
+            if (writeWatchedIndices != null) {
+                this.writeWatchedIndices = writeWatchedIndices;
```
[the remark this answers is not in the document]
sujithvm: Index patterns are supported. May need to rename the variable

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/AbstractAuditlogiUnitTest.java
vrozov: OK, but similar issue applies to `AuditTestUtils.updateAuditConfig()` that is used by `AbstractRestApiUnitTest`.

securityconfig/audit.yml.example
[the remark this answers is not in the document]
hardik-k-shah: Thanks. It is best to add sample/ comment in example YML file

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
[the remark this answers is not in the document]
hardik-k-shah: Will whitelist to record audit config changes always. --> why do we need to do whitelist? We should not treat auditlogging config specially. It should be same as other user/action/roles/role-mapping, cert-Dn etc.. 

It should be enabled (Default: True) and customer should be able to disable if they want.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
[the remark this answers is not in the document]
hardik-k-shah: renaming is good. But I would also add comments/ sample in audit_logging sample YML file.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiAction.java
```diff
@@ -0,0 +1,185 @@
...
+    @Override
+    public List<Route> routes() {
+        return routes;
+    }
+
+    @Override
+    protected void handleApiRequest(final RestChannel channel, final RestRequest request, final Client client) throws IOException {
+        // if audit config doc is not available in security index,
+        // disable audit APIs
+        if (!cl.isAuditHotReloadingEnabled()) {
```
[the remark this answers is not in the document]
sujithvm: Yes. We blocking all APIs including GET. Thought that there is no usage for GET as well if the hot-reloading is disabled. However this can be changed. Do you want to return the elasticsearch.yml settings ?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditLogImpl.java
```diff
@@ -40,31 +41,43 @@
...
+						final Client clientProvider,
+						final ThreadPool threadPool,
+						final IndexNameExpressionResolver resolver,
+						final ClusterService clusterService,
+						final Environment environment,
+						final boolean dlsFlsAvailable) {
+		super(settings, threadPool, resolver, clusterService, environment);
+		this.settings = settings;
+		this.dlsFlsAvailable = dlsFlsAvailable;
+		if (!dlsFlsAvailable) {
```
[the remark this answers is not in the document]
sujithvm: https://github.com/opendistro-for-elasticsearch/security/issues/547

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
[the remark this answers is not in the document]
sujithvm: Resolved with #542

legacy/securityconfig_v6/audit.yml.example
```diff
@@ -1,46 +1,77 @@
...
+
+    # Log only diffs for document updates
     write_log_diffs: false
+
+    # List of indices to watch for write events. Wildcard patterns are supported
+    # write_watched_indices: ["twitter", "logs-*"]
     write_watched_indices: []
-    write_ignore_users: []
+
+    # List of users to ignore for write events. Wildcard patterns are supported. Eg:
```
hardik-k-shah, line 75: Add Example or remove Eg:

legacy/securityconfig_v6/audit.yml.example
```diff
@@ -1,46 +1,77 @@
...
+    # Map of indexes and fields to monitor for read events. Wildcard patterns are supported for both index names and fields. Eg:
+    # read_watched_fields: {
+    #   "twitter": ["message"]
+    #   "logs-*": ["id", "attr*"]
+    # }
     read_watched_fields: {}
-    read_ignore_users: []
 
-    # compliance write
+    # List of users to ignore for read events
```
hardik-k-shah, line 61: wildcard pattern?

securityconfig/audit.yml.example
```diff
@@ -3,50 +3,79 @@ _meta:
...
     write_metadata_only: true
+
+    # Log only diffs for document updates
     write_log_diffs: false
+
+    # List of indices to watch for write events. Wildcard patterns are supported
+    # write_watched_indices: ["twitter", "logs-*"]
     write_watched_indices: []
+
+    # List of users to ignore for write events. Wildcard patterns are supported. Eg:
```
hardik-k-shah, line 79: same as above

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -122,6 +122,8 @@ protected void onComplianceConfigChanged(ComplianceConfig complianceConfig) {
         this.complianceConfig = complianceConfig;
         enableRoutes();
         this.complianceConfig.log(log);
+        // External config is audit logged only once per node start and config from index is not available at that time.
+        // The audit event will be created only if enabled and not already logged.
```
hardik-k-shah, line 126: And externalConfig parameter can be hot-reloadable which will impact functionality of "logExternalConfig()"

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -17,6 +18,9 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+
+import static com.amazon.opendistroforelasticsearch.security.DefaultObjectMapper.getOrDefault;
```
vrozov, line 23: Double import. Please check merge.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -340,4 +344,36 @@ public String toString() {
...
+            ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_WRITE_WATCHED_INDICES
+    );
+
+    public static Set<String> getDeprecatedKeys(final Settings settings) {
+        return AuditConfig.DEPRECATED_KEYS
+                .stream()
+                .filter(settings::hasValue)
+                .collect(Collectors.toSet());
+    }
+}
```
vrozov, line 379: nit: add new line.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -105,7 +109,7 @@ public static AuditConfig from(final Settings settings) {
      * Audit logger will use these settings to determine what audit logs are to be generated.
      */
     public static class Filter {
-        private static final Filter DEFAULT = Filter.from(Settings.EMPTY);
+        public static final Filter DEFAULT = Filter.from(Settings.EMPTY);
```
vrozov, line 112: it can be kept private

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
sujithvm: Duplicate import removed

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
sujithvm: Made it visible for testing as one of the tests is using it

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -161,6 +164,24 @@ static Filter from(Map<String, Object> properties) {
             final Set<String> ignoredAuditUsers = ImmutableSet.copyOf(getOrDefault(properties, "ignore_users", DEFAULT_IGNORED_USERS));
             final Set<String> ignoreAuditRequests = ImmutableSet.copyOf(getOrDefault(properties, "ignore_requests", Collections.emptyList()));
 
+            boolean invalid = properties.keySet()
```
vrozov, line 167: This can be simplified to `ImmutableSet.of("enable_rest", ...).containsAll(properties.keySet())` with `ImmutableSet.of(...)` being `private static Set<String>`

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfigFilterTest.java
```diff
@@ -111,4 +111,4 @@ public void testEmpty() {
         assertTrue(auditConfigFilter.getDisabledRestCategories().isEmpty());
         assertTrue(auditConfigFilter.getDisabledTransportCategories().isEmpty());
     }
-}
+}
```
vrozov, line 114: nit: undo this change

PR #410, opened 2020-04-24
Do not shutdown logger context during Node close

vrozov:
*Issue #, if available: Avoid changing logger configuration during tests

*Description of changes:

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/security/test/helper/cluster/ClusterHelper.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/security/test/helper/cluster/ClusterHelper.java
@@ -44,8 +44,6 @@ import java.util.stream.Collectors;
 import org.apache.commons.io.FileUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.logging.log4j.core.LoggerContext;
-import org.apache.logging.log4j.core.config.Configurator;
 import org.elasticsearch.ElasticsearchTimeoutException;
 import org.elasticsearch.action.admin.cluster.health.ClusterHealthResponse;
 import org.elasticsearch.action.admin.cluster.node.info.NodeInfo;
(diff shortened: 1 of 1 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/security/test/helper/cluster/ClusterHelper.java
```diff
@@ -244,8 +242,6 @@ public final void stopCluster() throws Exception {
 
     private static void closeNode(Node node) {
         try {
-            LoggerContext context = (LoggerContext) LogManager.getContext(false);
-            Configurator.shutdown(context);
```
debjanibnrj, line 248: any reason why we are not shutting down this logger context?

vrozov: The `LoggerContext` is not cluster specific and it was not created during cluster creation. It must not be shutdown when a node is closed and/or when cluster is stopped. Shutting down `LoggerContext` may cause a change in `Logger` levels and missing log messages that may help with troubleshooting unit tests.

dinusX: This code is used only by tests . Maybe some tests need this for correct functionality or more verbosity ? 

What is the reason for removing this code ? Are there some logs or output missing ? 

Do you have any example of the difference in output "before/after" to understand how this change helps ?

vrozov: The code only affects the amount of logging and I would not expect tests to depend on disabling or shutting down logging that makes it less verbose. In addition, the order of tests execution is not defined and one unit test may not rely on another unit test to shutdown the logging context.

The easiest way to see the difference is to change `cas` logger level from `error` to `trace` in `log4j2-test.properties` and run `SSLTest` or `OpenSSLTest` (or any other test suite that creates two or more clusters). After the subject code is called, there would be no entries from the `cas` logger logged.

PR #420, opened 2020-04-24
Fix for OpenSSLTest.testNodeClientSSL failure

vrozov:
*Issue #, if available:*

*Description of changes:*
- Initialize "path.data", "path.logs" and "path.home" for every ES Node started during tests
- Change "path.data", "path.logs" and "path.home" to use "target/data"


src/test/java/com/amazon/opendistroforelasticsearch/security/ssl/SSLTest.java
```diff
@@ -605,19 +605,23 @@ public void testNodeClientSSL() throws Exception {
...
 
         final Settings tcSettings = Settings.builder().put("cluster.name", clusterInfo.clustername).put("path.home", ".")
                 .put("node.name", "client_node_" + new Random().nextInt())
                 .put("node.data", false)
                 .put("node.master", false)
                 .put("node.ingest", false)
+                .put("path.data", "./target/data/"+clusterInfo.clustername+"/ssl/data")
+                .put("path.logs", "./target/data/"+clusterInfo.clustername+"/ssl/logs")
+                .put("path.home", "./target")
+                .put("discovery.initial_state_timeout","8s")
```
debjanibnrj, line 619: nit: this setting seems to be repeated on line 620.

vrozov:
> OpenSSLTest.testNodeClientSSL test was failing with MasterNotFound exception. How did changing the path.data, path.logs and path.home resolve the test issue?

`MasterNotDiscoveredException` was caused by the Node being rejected from joining the cluster as its UUID would not match the cluster UUID that it tried to join. To avoid it, it is necessary to start the Node with no UUID assigned. It is usually the case during the tests unless "path.data" points to an existing directory that was previously used by another Node. Note that `ClusterHelper` sets those properties for Nodes it creates and uses unique `clusterName` to avoid the issue. The fix sets "path.data" ("path.logs" and "path.home") for other Nodes created during the tests.

I also changed "path.data", "path.logs" and "path.home" to root at "target/data", so that "mvn clean" will delete any leftover directories that were not deleted during tests (due to exceptions).

debjanibnrj:
> MasterNotDiscoveredException was caused by the Node being rejected from joining the cluster as its UUID would not match the cluster UUID that it tried to join. To avoid it, it is necessary to start the Node with no UUID assigned. It is usually the case during the tests unless "path.data" points to an existing directory that was previously used by another Node. Note that ClusterHelper sets those properties for Nodes it creates and uses unique clusterName to avoid the issue. The fix sets "path.data" ("path.logs" and "path.home") for other Nodes created during the tests.
> I also changed "path.data", "path.logs" and "path.home" to root at "target/data", so that "mvn clean" will delete any leftover directories that were not deleted during tests (due to exceptions).

Thanks for the detailed explanation.

src/test/java/com/amazon/opendistroforelasticsearch/security/SlowIntegrationTests.java
```diff
@@ -75,7 +75,9 @@ public void testNodeClientAllowedWithServerCertificate() throws Exception {
                 .put("node.data", false)
                 .put("node.master", false)
                 .put("node.ingest", false)
-                .put("path.home", "/tmp")
+                .put("path.data", "./target/data/" + clusterInfo.clustername + "/cert/data")
```
dinusX, line 78: The changes look good, but I was wandering if using prefix "./target" is a good idea ? 

I saw that this folder is used for the output jars during package build.    
Maybe instead we should create some temporary folders or use different prefix "test-target" (or anything that contains test) .   
Do we have other tests that also use "./target" ?

vrozov: "target" is the default location that maven uses for builds and temporary files including compiler output, unit test (surefire) reports, jars, and more. As jars are created from "target/classes" or "target/test-classes", there will be no conflict and no unnecessary files will be included into the target jar files.

The goal of the change is to use "mvn clean" to delete any leftover directories/files that were not cleaned/deleted during the tests.

PR #422, opened 2020-04-24
Enable limited OpenSSL support

vrozov:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/DefaultOpenDistroSecurityKeyStore.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/DefaultOpenDistroSecurityKeyStore.java
@@ -57,12 +57,12 @@ import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLException;
 import javax.net.ssl.SSLParameters;
 
+import io.netty.util.internal.PlatformDependent;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import org.elasticsearch.ElasticsearchException;
 import org.elasticsearch.ElasticsearchSecurityException;
-import org.apache.lucene.util.Constants;
 import org.elasticsearch.SpecialPermission;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.env.Environment;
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/DefaultOpenDistroSecurityKeyStore.java
```diff
@@ -133,14 +133,11 @@ public DefaultOpenDistroSecurityKeyStore(final Settings settings, final Path con
                 .getAsBoolean(SSLConfigConstants.OPENDISTRO_SECURITY_SSL_TRANSPORT_ENABLE_OPENSSL_IF_AVAILABLE, true);
 
         if(!OpenDistroSecuritySSLPlugin.OPENSSL_SUPPORTED && OpenSsl.isAvailable() && (settings.getAsBoolean(SSLConfigConstants.OPENDISTRO_SECURITY_SSL_HTTP_ENABLE_OPENSSL_IF_AVAILABLE, true) || settings.getAsBoolean(SSLConfigConstants.OPENDISTRO_SECURITY_SSL_TRANSPORT_ENABLE_OPENSSL_IF_AVAILABLE, true) )) {
-            String text = "Support for OpenSSL has been removed from Open Distro Security since Elasticsearch 7.4.0. Use JDK SSL instead\n";
-            if(Constants.JRE_IS_MINIMUM_JAVA11) {
-                text += "Since you are running Java "+Constants.JAVA_VERSION+" you should not experience any performance impact but maybe not all your ciphers are supported. If you experience problems upgrade to Java 11+";
+            if (PlatformDependent.javaVersion() < 12) {
+                log.warn("Support for OpenSSL with Java 11 or prior versions require using Netty allocator. Set 'es.unsafe.use_netty_default_allocator' system property to true");
```
debjanibnrj, line 137: What does the 'use_netty_default_allocator' property do and why is it necessary for supporting node to node communication with OpenSSL?

debjanibnrj:
Please elaborate on how you tested these changes.

src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/DefaultOpenDistroSecurityKeyStore.java
vrozov: The `es.unsafe.use_netty_default_allocator` System property is used by `NettyAllocator` to determine which `ByteBufAllocator` to use. With this property set to `true`, ES will use native Netty allocator, otherwise, it will use its own `NettyAllocator.NoDirectBuffers` allocator that does not support direct buffers. As openssl is supported through netty and uses tcnative that needs direct buffers, it is necessary to configure ES to use Netty allocator if openssl is required/preferred.

vrozov:
Tested using `OpenSSLTest`. Please let me know if this is not sufficient.

debjanibnrj:
> Tested using OpenSSLTest. Please let me know if this is not sufficient.

Would be useful to manually test out the changes as well. This can be done by spinning up cluster of nodes with elasticsearch oss (in this case for es 7.6) with the security plugin built with these changes. As far as I remember when we removed support for OpenSSL, the tests were still passing but cross-cluster communication for Java version <12 were failing.

vrozov:
There is a difference in the test environment introduced in #339 that explains why tests were passing while inter-cluster communication was failing for Java 11 or prior versions. In the past, tests were running on the docker image without openssl installed and tests that exercise openssl functionality were disabled on `OpenSsl.isAvailable()` condition. Those tests are now enabled and they won't pass without changes in this PR.

debjanibnrj:
> There is a difference in the test environment introduced in #339 that explains why tests were passing while inter-cluster communication was failing for Java 11 or prior versions. In the past, tests were running on the docker image without openssl installed and tests that exercise openssl functionality were disabled on `OpenSsl.isAvailable()` condition. Those tests are now enabled and they won't pass without changes in this PR.

Thanks for the explanation.

PR #373, opened 2020-04-10
Remove invalid option.

fbarbeira:
`opendistro_security.audit.enable_request_details` is not valid anymore.

I tried with ODFE 1.6.0 and ES gives the following error:

```
[2020-04-10T16:52:22,592][ERROR][o.e.b.Bootstrap          ] [data01] Exception
java.lang.IllegalArgumentException: unknown setting [opendistro_security.audit.enable_request_details] did you mean any of [opendistro_security.audit.enable_rest, opendistro_security.audit.enable_transport, opendistro_security.audit.log_request_body]?
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #452, opened 2020-05-02
Version bump 1.7.0.0

sujithvm:
Supporting ES: 7.6.1 (same as plugin version 1.6.0.0)

Changes:

- Implemented APIs and datamodel to configure nodes_dn dynamically #445 (backported from master #362)
- Performance improvement by memorizing results of resolveIndexPatterns for Bulk requests (backported from master #309)
- Performance improvement by implementing faster version of implies type perm #302
- Enabled limited OpenSSL support
- Changed file permissions for securityconfig and tools #387
- Fixed bug which caused user to lose roles on account password update #333
- Refactored to use Greenrobot EventBus #445 (backported from master #370)
- Refactored Resolved class, dropped unused fields and simplified logic (backported from master #310)
- Refactored audit logging related classes #445 (backported from master #303, #445, #306, #373, #368)

### DIFF
--- a/opendistro-elasticsearch-security.release-notes
+++ b/opendistro-elasticsearch-security.release-notes
@@ -1,6 +1,6 @@
 ## 2020-05-01 Version 1.7.0.0
 
-- Support for Elasticsearch 7.6.1
+- Supported Elasticsearch version 7.6.1 (same as Open Distro for Elasticsearch Security plugin version 1.6.0.0)
 - Implemented APIs and datamodel to configure nodes_dn dynamically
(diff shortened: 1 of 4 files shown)


opendistro-elasticsearch-security.release-notes
```diff
@@ -1,3 +1,16 @@
+## 2020-05-01 Version 1.7.0.0
+
+- Support for Elasticsearch 7.6.1
```
vrozov, line 3: Please mention that there is no change since 1.6.0.0

PR #461, opened 2020-05-14
Fix routes in master

sujithvm:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/MigrateApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/MigrateApiAction.java
@@ -17,6 +17,8 @@ package com.amazon.opendistroforelasticsearch.security.dlic.rest.api;
 
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.Collections;
+import java.util.List;
 
 import org.elasticsearch.Version;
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/MigrateApiAction.java
```diff
@@ -76,8 +81,8 @@ public MigrateApiAction(final Settings settings, final Path configPath, final Re
     }
 
     @Override
-    protected void registerHandlers(RestController controller, Settings settings) {
-        controller.registerHandler(Method.POST, "/_opendistro/_security/api/migrate", this);
+    public List<Route> routes() {
```
debjanibnrj, line 84: for my understanding why are we using routes for these APIs and not using registerHandlers?

sujithvm: ES has changed the API in 7.7 to use routes() to register all the REST handlers. 

Here is the change by ES: https://github.com/elastic/elasticsearch/pull/51950

debjanibnrj: Thanks for clarifying.

PR #464, opened 2020-05-18
Adds workflow to push jar to maven repo

gaiksaya:
*Description of changes:* This workflow will push the jar to maven staging repo which will further will be used for release process.

[Tested](https://github.com/gaiksaya/security/runs/686501761?check_suite_focus=true)

### DIFF
--- a/.github/workflows/push-jar-maven.yml
+++ b/.github/workflows/push-jar-maven.yml
@@ -1,0 +1,44 @@
+name: Upload Jar to Maven
+
+on:
+  push:
+    tags: 
+      - v*
+jobs:
+  upload-security-jar:
+    runs-on: [ubuntu-16.04]
+
+    name: Upload Jar to Maven
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+        
+      - name: Configure AWS CLI
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+
+  # Since CD workflow uses this version of java
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11.0.x
+
+      - name: Uploading Jar to Maven
+        env:
+          passphrase: ${{ secrets.PASSPHRASE }}
+        run: |
+          cd ..
+          aws s3 cp s3://opendistro-docs/github-actions/pgp-public-key .
+          aws s3 cp s3://opendistro-docs/github-actions/pgp-private-key .
+          
+          gpg --import pgp-public-key
+          gpg --allow-secret-key-import --import pgp-private-key
+
+          aws s3 cp s3://opendistro-docs/github-actions/settings.xml /home/runner/.m2/
+          
+          
+          cd security
+          mvn -Dgpg.passphrase=$passphrase clean deploy -DskipTests -DstagingDescription="security plugin maven release"   -P release -DaltDeploymentRepository=sonatype-nexus-staging::default::https://oss.sonatype.org/service/local/staging/deploy/maven2


.github/workflows/push-jar-maven.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    name: Upload Jar to Maven
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+        
+      - name: Configure AWS CLI
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
hardik-k-shah, line 19: Curious to know: from where this access key/ secret key come?

.github/workflows/push-jar-maven.yml
```diff
@@ -0,0 +1,44 @@
...
+          passphrase: ${{ secrets.PASSPHRASE }}
+        run: |
+          cd ..
+          aws s3 cp s3://opendistro-docs/github-actions/pgp-public-key .
+          aws s3 cp s3://opendistro-docs/github-actions/pgp-private-key .
+          
+          gpg --import pgp-public-key
+          gpg --allow-secret-key-import --import pgp-private-key
+
+          aws s3 cp s3://opendistro-docs/github-actions/settings.xml /home/runner/.m2/
```
hardik-k-shah, line 40: what is there in this "s3://opendistro-docs/github-actions/settings.xml"

.github/workflows/push-jar-maven.yml
gaiksaya: It is for the s3 bucket (account: openes-infra@amazon.com) that is used for staging and release of ODFE

.github/workflows/push-jar-maven.yml
gaiksaya: > what is there in this "s3://opendistro-docs/github-actions/settings.xml"

Staging repo details and authentication for the same

.github/workflows/push-jar-maven.yml
```diff
@@ -0,0 +1,44 @@
+name: Upload Jar to Maven
```
vrozov, line 1: What is the reason to have it as a separate workflow and not as part of CD.yml?

gaiksaya: > What is the reason to have it as a separate workflow and not as part of CD.yml?

It can be but one issue I faced was with the OS that it runs on. gpg signing has some [issues](https://github.com/gaiksaya/security/runs/686751935?check_suite_focus=true) with ubuntu-latest where as it works fine with ubuntu-16.04

vrozov: What is the downside of changing CD to run on the version of ubuntu where gpg works? Can you confirm with github or ubuntu communities why gpg does not work on the latest.

.github/workflows/push-jar-maven.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11.0.x
+
+      - name: Uploading Jar to Maven
+        env:
+          passphrase: ${{ secrets.PASSPHRASE }}
+        run: |
+          cd ..
```
vrozov, line 33: `cd ..`?

.github/workflows/push-jar-maven.yml
```diff
@@ -0,0 +1,44 @@
...
+          aws s3 cp s3://opendistro-docs/github-actions/pgp-public-key .
+          aws s3 cp s3://opendistro-docs/github-actions/pgp-private-key .
+          
+          gpg --import pgp-public-key
```
vrozov, line 37: Does `gpg` need both public and private keys?

.github/workflows/push-jar-maven.yml
vrozov: `/home/runner/.m2` is the directory documented or github is free to change it anytime?

sujithvm, line 40: Any reason why we want to copy the credentials from S3 instead of adding them to github secrets and reading from there ? 

https://help.github.com/en/actions/language-and-framework-guides/publishing-java-packages-with-maven

sujithvm: Please check #481 (uses CD pipeline and sensitive data is added to github secrets)

PR #468, opened 2020-05-20
Refactor Base64Helper class

vrozov:
*Description of changes:*

- introduced SafeObjectOutputStream to catch unauthorized serializations
- introduced Base64HelperTest

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/support/Base64Helper.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/support/Base64Helper.java
@@ -14,7 +14,7 @@
  */
 
 /*
- * Portions Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Portions Copyright 2019-2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/support/Base64Helper.java
```diff
@@ -38,79 +44,123 @@
...
+        User.class,
+        SourceFieldsContext.class,
+        LdapUser.class,
+        SearchEntry.class,
+        LdapEntry.class,
+        AbstractLdapBean.class,
+        LdapAttribute.class
+    );
+
+    private static final List<Class<?>> SAFE_ASSIGNABLE_FROM_CLASSES = ImmutableList.of(
```
debjanibnrj, line 86: Should this be a Set as well?

src/main/java/com/amazon/opendistroforelasticsearch/security/support/Base64Helper.java
```diff
@@ -38,79 +44,123 @@
...
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
 
 import org.elasticsearch.ElasticsearchException;
+import org.elasticsearch.common.Strings;
 
-import com.amazon.opendistroforelasticsearch.security.resolver.IndexResolverReplacer;
```
sujithvm, line 52: Is `IndexResolverReplacer.Resolved` not needed or missed ?

vrozov: It is not needed.

src/main/java/com/amazon/opendistroforelasticsearch/security/support/Base64Helper.java
vrozov: It can be `List` or `Set` as there is no lookup by key support required and elements needs to be unique. As it is statically populated and we can guarantee uniqueness, `List` or even `Collection` is sufficient.

PR #475, opened 2020-05-27
Update compliance config usage

sujithvm:
- moved compliance ignore read and write users into compliance config as it belongs there

- moved environment external logging into audit log and store state to check its logged only once per node start. When hot-reloaded, compliance config wont be available in index yet at the time of node start

- use ComplianceIndexingOperationListenerImpl as compliance config can be changed dynamically. This is called by index and at the time of creation setting the object

- use <string, set<string>> type for read enabled fields and move creation to the source generating the config

- add null checks for compliance config as it can be null

- updated reflection helper to construct object by constructor instead of reflection

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
@@ -49,6 +49,7 @@ import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import com.google.common.collect.ImmutableSet;
 import org.apache.lucene.index.DirectoryReader;
 import com.amazon.opendistroforelasticsearch.security.ssl.rest.OpenDistroSecuritySSLReloadCertsAction;
 import com.amazon.opendistroforelasticsearch.security.ssl.rest.OpenDistroSecuritySSLCertsInfoAction;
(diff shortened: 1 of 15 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
```diff
@@ -508,20 +508,14 @@ public void onIndexModule(IndexModule indexModule) {
...
 
-                assert complianceConfig!=null:"compliance config must not be null here";
-                
-                if(complianceConfig.writeHistoryEnabledForIndex(indexModule.getIndex().getName())) {
-                    ciol = ReflectionHelper.instantiateComplianceListener(Objects.requireNonNull(auditLog));
-                    indexModule.addIndexOperationListener(ciol);
-                } else {
-                    ciol = new ComplianceIndexingOperationListener();
-                }
+                ciol = ReflectionHelper.instantiateComplianceListener(indexName, Objects.requireNonNull(auditLog));
```
debjanibnrj, line 517: nit: final ComplianceIndexingOperationListener ciol = ReflectionHelper.instantiateComplianceListener(indexName, Objects.requireNonNull(auditLog));

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -17,14 +17,15 @@
  */
 public class AuditConfig {
 
+    public static final List<String> DEFAULT_IGNORED_USERS = Collections.singletonList("kibanaserver");
```
debjanibnrj, line 20: Is it possible for DEFAULT_IGNORED_USERS to be different for AuditConfig and ComplianceConfig?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -425,15 +429,15 @@ public void logDocumentRead(String index, String id, ShardId shardId, Map<String
     @Override
     public void logDocumentWritten(ShardId shardId, GetResult originalResult, Index currentIndex, IndexResult result) {
 
-        if(complianceConfig == null || !complianceConfig.writeHistoryEnabledForIndex(shardId.getIndexName())) {
+        if(complianceConfig == null || !complianceConfig.isEnabled() || !complianceConfig.writeHistoryEnabledForIndex(shardId.getIndexName())) {
```
debjanibnrj, line 432: Saw that this was used twice. May want to make this into a common utility function.

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
```diff
@@ -121,49 +128,51 @@ public Index preIndex(final ShardId shardId, final Index index) {
...
+                Objects.requireNonNull(is);
+
+                if (result.getFailure() != null || index.origin() != org.elasticsearch.index.engine.Engine.Operation.Origin.PRIMARY) {
+                    return;
+                }
 
-            if(is.getShardOrNull(shardId.getId()) == null) {
-                return;
-            }
+                if (is.getShardOrNull(shardId.getId()) == null) {
```
debjanibnrj, line 150: You can club this with the if statement on the line above https://github.com/opendistro-for-elasticsearch/security/pull/475/files#diff-5c8a4dc5ef259d5edbdab7104f8ff0f2R146

vrozov:
My understanding is that some changes in the PR are independent from each other. If it is the case, please split this PR into multiple PRs with logically complete set of changes.

sujithvm:
Well, the connection with all the changes is around how compliance config is used. Want to avoid the overhead of creating and backporting tiny few lines PRs. If you insist I will split them.

vrozov:
It will be easier to review, understand and approve individual changes.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
sujithvm: It is the same as kibanaserver is the only special user whose requests we want to ignore. This change has been moved to #484

sujithvm:
This PR is being covered in smaller PRs #483 #491 #484

PR #481, opened 2020-06-01
Update CD workflow to publish artifacts to maven central

sujithvm:
- Upload to Maven Central as part of CD pipeline.
- Alternative approach to #464 by adding the sensitive data to github secrets 
SONATYPE_USER (nexus username), 
SONATYPE_PASSWORD (nexus password), 
PGP_PRIVATE_KEY (private key for signing)
PGP_PASSPHRASE (passphrase)


vrozov:
Is it necessary to have `settings.xml`? What is the reason we can't rely on [Publishing Java packages with Maven](https://help.github.com/en/actions/language-and-framework-guides/publishing-java-packages-with-maven) action?

sujithvm:
setup-java also creates a settings.xml file and repository must be configured. settings.xml is already present in the project hence reusing that.

vrozov:
IMO, it will be better to align with github workflow, especially that 
- `settings.xml` are usually maven specific (maven repo, user or machine) and should not be part of the source control.
- existing `settings.xml` don't seem to provide any additional functionality that can't be achieved with the standard github action.

I'd suggest to remove `settings.xml` and move settings to `pom.xml` if necessary.

.github/workflows/cd.yml
```diff
@@ -53,6 +56,14 @@ jobs:
         aws s3 cp artifacts/*.rpm $s3_path/rpms/opendistro-security/
         aws cloudfront create-invalidation --distribution-id ${{ secrets.DISTRIBUTION_ID }} --paths '/downloads/*'
 
+    - name: Upload Artifacts to Maven Central
+      env:
+        SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
+        SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+      run: |
+        cat <(echo -e "${{ secrets.PGP_PRIVATE_KEY }}") | gpg --batch --import
```
vrozov, line 64:
```suggestion
       gpg --batch --import --no-tty <(echo -e "${{ secrets.PGP_PRIVATE_KEY }}") 
```

.github/workflows/cd.yml
```diff
@@ -53,6 +56,14 @@ jobs:
...
+        mvn -B clean deploy -Padvanced -Prelease -DskipTests -Dgpg.passphrase=${{ secrets.GPG_PASSPHRASE }}
```
vrozov, line 65: GPG_PASSPHRASE -> PGP_PASSPHRASE or PGP_PRIVATE_KEY -> GPG_PRIVATE_KEY

pom.xml
```diff
@@ -58,6 +58,14 @@
         </developer>
     </developers>
 
+    <distributionManagement>
```
vrozov, line 61: Consider moving to `release` profile.

.github/workflows/cd.yml
sujithvm: @gaiksaya Could you confirm this? Looks like we have GPG_PASSPHRASE and PGP_PRIVATE_KEY

gaiksaya: Yes that's right @sujithvm along with pgp public key

vrozov: It should be consistent, either PGP or GPG. Public key is not a secret.

sujithvm: Are you saying the key and passphrase itself or the names here must be consistent? I don't know how both were generated. 

I think we just private key to sign the artifact and public key seems to be already distributed here http://pool.sks-keyservers.net/pks/lookup?search=OpenDistroForElasticsearch&fingerprint=on&op=index

vrozov: Actually both, passphrase should be for the generated private key and names needs to be consistent as they refer to a related settings.

Public key does not need to be added to the secrets.

sujithvm: Not referencing public key here so it is not added to secrets. 

Looking at this https://blog.sonatype.com/2010/01/how-to-generate-pgp-signatures-with-maven/ it looks fine to have pgp signing key with gpg passphrase and names need not be consistent.

The passphrase seems to be generated for the private key only.

@vrozov Am I missing something?

vrozov: > Not referencing public key here so it is not added to secrets.

Correct, I replied to @gaiksaya in response to 
> Yes that's right @sujithvm along with pgp public key

> Looking at this https://blog.sonatype.com/2010/01/how-to-generate-pgp-signatures-with-maven/ it looks fine to have pgp signing key with gpg passphrase and names need not be consistent.

There is only (pgp/gpg) signing key that is protected by (pgp/gpg) password. If you refer to it as a pgp key, should not it be pgp password/passphrase? Note that gpg is one of implementations of pgp spec, so it will be better either to use implementation (gpg) or spec (pgp) to refer to a (private) key and it's passphrase.

sujithvm: Are you asking to rename the secret `secrets.GPG_PASSPHRASE` to `secrets.PGP_PASSPHRASE ` ? And have `-Dgpg.passphrase=${{ secrets.PGP_PASSPHRASE }}` ?

vrozov: Yes :)

sujithvm: 😄 I was assuming there was some mistake in the way the signatures were made making them incompatible

Issue #325, opened 2020-03-25
AuditLog: deprecate "NONE" in favor of "~" or "null"


vrozov:

dbjnbnrj:
Mentioned in the earlier PRs as well. Prefer null.

vrozov:
https://symfony.com/doc/current/components/yaml/yaml_format.html#nulls

nateynateynate:
Looks like the merges have completed the requirements for this one. I'm going to close it out. Please feel free to reopen if this is in error.

PR #484, opened 2020-06-02
Move compliance ignore users from audit config to compliance config

sujithvm:
- Moving compliance ignore users from audit config to compliance config as thats the correct place for this setting to exist. 
- Migrated all the tests

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
@@ -20,6 +20,8 @@ import java.util.Set;
  */
 public class AuditConfig {
 
+    public static final List<String> DEFAULT_IGNORED_USERS = Collections.singletonList("kibanaserver");
+
     private AuditConfig() { }
 
     /**
@@ -27,7 +29,6 @@ public class AuditConfig {
      * Audit logger will use these settings to determine what audit logs are to be generated.
      */
     public static class Filter {
-        private static final List<String> DEFAULT_IGNORED_USERS = Collections.singletonList("kibanaserver");
         private static final List<String> DEFAULT_DISABLED_CATEGORIES =
                 Arrays.asList(AuditCategory.AUTHENTICATED.toString(),
                         AuditCategory.GRANTED_PRIVILEGES.toString());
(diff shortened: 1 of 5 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -286,10 +245,16 @@ public String toString() {
...
                     ", excludeSensitiveHeaders=" + excludeSensitiveHeaders +
                     ", ignoredAuditUsers=" + ignoredAuditUsersMatcher +
-                    ", ignoredComplianceUsersForRead=" + ignoredComplianceUsersForReadMatcher +
-                    ", ignoredComplianceUsersForWrite=" + ignoredComplianceUsersForWriteMatcher +
                     ", ignoreAuditRequests=" + ignoredAuditRequestsMatcher +
                     '}';
         }
     }
+
+    public static Set<String> getSettingAsSet(final Settings settings, final String key, final List<String> defaultList, final boolean ignoreCaseForNone) {
```
debjanibnrj, line 253: Is it possible to move this to a common util class, since this seems to be used by both AuditConfig and ComplianceConfig?

vrozov: Should `ComplianceConfig` be moved to `AuditConfig`? Looks like `Filter` and `Compliance` are both related to `Audit`.

sujithvm: There will be a `Compliance` to represent storage in index and generate ComplianceConfig out of it. Dependency on settings is stopping from directly using it unless there is some workaround. 
This is where I want to get to https://github.com/opendistro-for-elasticsearch/security/blob/5a81a60cc05d37d92a965e8cb03a5214c10e03c1/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java 

Hence thought of doing it separate PRs.

vrozov: Consider cutting a separate (preliminary) PR to move `ComplienceConfig` to `AuditConfig` if that will help to address @debjanibnrj comment.

sujithvm: Referencing @debjanibnrj comment I don't think there needs to be a separate util class as it can be just a util function within AuditConfig and the function itself might disappear with  #325

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -39,8 +40,6 @@ private AuditConfig() { }
         private final boolean resolveIndices;
         private final boolean excludeSensitiveHeaders;
         private final WildcardMatcher ignoredAuditUsersMatcher;
-        private final WildcardMatcher ignoredComplianceUsersForReadMatcher;
```
vrozov, line 42: Please update `Filter` constructor and remove unused parameters.

PR #486, opened 2020-06-02
Preventing users from mapping to hidden/reserved opendistro_security_…

debjanibnrj:
…roles

*Issue #, if available:* 

*Description of changes:*
Currently an authorized user can create a new user and attach a reserved role using a PUT request to the /_opendistro/_security/api/user/(new_username_here) endpoint. 

Eg: 
```
PUT _opendistro/_security/api/internalusers/kirkuser
{
  "password": "kirkpass",
  "backend_roles": ["captains", "starfleet"],
  "opendistro_security_roles": ["restricted_role"]
}
```

This change will forbid users from associating user with restricted` opendistro_security_roles`

*Tests:*
Updated unit tests

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
@@ -115,20 +115,19 @@ public class InternalUsersApiAction extends PatchableResourceApiAction {
         final ObjectNode contentAsNode = (ObjectNode) content;
         final SecurityJsonNode securityJsonNode = new SecurityJsonNode(contentAsNode);
 
-
-        // Don't allow user to add reserved or hidden role
-        final SecurityDynamicConfiguration<?> rolesConfiguration = load(CType.ROLES, false);
+        // Don't allow user to add hidden, reserved or non-existent role
         final List<String> opendistroSecurityRoles = securityJsonNode.get("opendistro_security_roles").asList();
         if (opendistroSecurityRoles != null) {
+            final SecurityDynamicConfiguration<?> rolesConfiguration = load(CType.ROLES, false);
             for (final String role: opendistroSecurityRoles) {
 
-                if (isHidden(rolesConfiguration, role)) {
+                if (!rolesConfiguration.exists(role) || isHidden(rolesConfiguration, role)) {
                     notFound(channel, "Role '"+role+"' is not found.");
                     return;
                 }
 
                 if (isReserved(rolesConfiguration, role)) {
-                    forbidden(channel, "Role '" + role + "' is read-only.");
+                    forbidden(channel, "Role '" + role + "' is reserved.");
                     return;
                 }
             }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
@@ -99,21 +99,39 @@ public class InternalUsersApiAction extends PatchableResourceApiAction {
         // TODO it might be sensible to consolidate this with the overridden method in
         // order to minimize duplicated logic
 
-        final SecurityDynamicConfiguration<?> configuration = load(getConfigName(), false);
+        final SecurityDynamicConfiguration<?> internalUsersConfiguration = load(getConfigName(), false);
 
-        if (isHidden(configuration, username)) {
+        if (isHidden(internalUsersConfiguration, username)) {
             forbidden(channel, "Resource '" + username + "' is not available.");
             return;
         }
 
         // check if resource is writeable
-        if (!isReservedAndAccessible(configuration, username)) {
+        if (!isReservedAndAccessible(internalUsersConfiguration, username)) {
             forbidden(channel, "Resource '" + username + "' is read-only.");
             return;
         }
 
         final ObjectNode contentAsNode = (ObjectNode) content;
         final SecurityJsonNode securityJsonNode = new SecurityJsonNode(contentAsNode);
+
+
+        // Don't allow user to add reserved or hidden role
+        final SecurityDynamicConfiguration<?> rolesConfiguration = load(CType.ROLES, false);
+        final List<String> opendistroSecurityRoles = securityJsonNode.get("opendistro_security_roles").asList();
+        if (opendistroSecurityRoles != null) {
+            for (final String role: opendistroSecurityRoles) {
+                if (isHidden(rolesConfiguration, role)) {
+                    forbidden(channel, "Role '"+role+"' is not available.");
+                    return;
+                }
+
+                if (isReserved(rolesConfiguration, role)) {
+                    forbidden(channel, "Cannot associate user with reserved role '"+role+"'");
+                    return;
+                }
+            }
+        }
 
         // if password is set, it takes precedence over hash
         final String plainTextPassword = securityJsonNode.get("password").asString();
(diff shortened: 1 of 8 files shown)


vrozov:
Was it allowed before? What if a user with a reserved role was created using prior versions, should it be allowed? Should we at least log a warning message if such user(s) exists?

debjanibnrj:
Hi @vrozov,

Was it allowed before? 

Mapping users to hidden/readonly roles is not supported in ES7.0 via the `rolesmapping` API and throws an exception like - 

What if a user with a reserved role was created using prior versions, should it be allowed? Should we at least log a warning message if such user(s) exists?

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/DynamicConfigFactory.java
```diff
@@ -189,13 +189,14 @@ public void onChange(Map<CType, SecurityDynamicConfiguration<?>> typeToConfig) {
 
                 log.debug("Static tenants loaded ({})", staticTenants.getCEntries().size());
 
-                log.debug("Static configuration loaded (total roles: {}/total action groups: {}/total tenants: {})", roles.getCEntries().size(), actionGroups.getCEntries().size(), tenants.getCEntries().size());
+                log.debug("Static configuration loaded (total roles: {}/total action groups: {}/total tenants: {})", roles.getCEntries().size(),
```
vrozov, line 192: nit: move `roles.getCEntries().size()` to the next line.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -99,22 +99,41 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
...
+        if (opendistroSecurityRoles != null) {
+            for (final String role: opendistroSecurityRoles) {
+
+                if (isHidden(rolesConfiguration, role)) {
+                    notFound(channel, "Role '"+role+"' is not found.");
+                    return;
+                }
+
+                if (isReserved(rolesConfiguration, role)) {
+                    forbidden(channel, "Role '" + role + "' is read-only.");
```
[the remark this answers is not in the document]
debjanibnrj: Had named this read-only to match other isReserved checks w- https://github.com/opendistro-for-elasticsearch/security/blob/c7b28ad6f0a472113404135858e9852949971621/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/PatchableResourceApiAction.java#L114-L116. 

In this case I think it is fine to name it reserved

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -99,22 +99,41 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
...
 
+
+        // Don't allow user to add reserved or hidden role
+        final SecurityDynamicConfiguration<?> rolesConfiguration = load(CType.ROLES, false);
+        final List<String> opendistroSecurityRoles = securityJsonNode.get("opendistro_security_roles").asList();
+        if (opendistroSecurityRoles != null) {
+            for (final String role: opendistroSecurityRoles) {
+
+                if (isHidden(rolesConfiguration, role)) {
+                    notFound(channel, "Role '"+role+"' is not found.");
```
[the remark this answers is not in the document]
debjanibnrj: We currently prevent users from modifying hidden roles - 
```
PUT  _opendistro/_security/api/roles/ES_hidden
{
  "status": "FORBIDDEN",
  "message": "Resource 'ES_hidden' is not available."
}
```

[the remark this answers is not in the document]
debjanibnrj: We can make this clearer for the user by mentioning the role is hidden but am unsure if that is a security concern. Right now we default to saying `Role is not found` in any case where the role is hidden - 

https://github.com/opendistro-for-elasticsearch/security/blob/ed6f901838e80e14116057bbffde02a3cf89e27e/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java#L151-L154

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -99,22 +99,44 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
...
+        final List<String> opendistroSecurityRoles = securityJsonNode.get("opendistro_security_roles").asList();
+        if (opendistroSecurityRoles != null) {
+            final SecurityDynamicConfiguration<?> rolesConfiguration = load(CType.ROLES, false);
+            for (final String role: opendistroSecurityRoles) {
+
+                if (rolesConfiguration.getCEntry(role) == null) {
+                    notFound(channel, "Role '"+role+"' does not exist.");
+                }
+
+                if (isHidden(rolesConfiguration, role)) {
```
vrozov, line 128: move `if (rolesConfiguration.getCEntry(role) == null)` to here:
```suggestion
                if (rolesConfiguration.exists(role) || isHidden(rolesConfiguration, role)) {
```

src/test/resources/roles.yml
```diff
@@ -1031,3 +1031,54 @@ role_foo_all:
...
+  index_permissions:
+    - index_patterns:
+        - "twitter"
+        - "analytics"
+      dls: null
+      fls: null
+      masked_fields: null
+      allowed_actions:
+        - "*"
+  tenant_permissions: []
```
vrozov, line 1084: add new line

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
debjanibnrj: I am slightly conflicted about treating hidden and non-existent roles the same way. 
For non-existent roles the customer can resolve the issue by creating the new role, for hidden roles this will not be possible. If we return the same error message for both do you think it will cause confusion for the customer?

vrozov: Hidden resource(role) means that for all practical purposes it can't be distinguished from not existing resource(role). Check, for example, how patch request for hidden resource/role is handled. The only exception where it is not possible to avoid exposing hidden resources is create(put) request.

thenom:
Hi @vrozov and @debjanibnrj 

Sorry if this is a stupid question, i am from an xpack licensed background and just trying to convert over.

Can you please clarify what exactly is and isnt supposed to be allowed as i cant find a definition of this or the JSON properties meanings in the documentation.  There seems to be a lot of terminology in here that matches\overlaps the config and some that is possibly missing or doesnt exist.

It seems if you read through this PR that a user should NOT be able to attach a reserved\restricted\hidden\read only role.  If you look at the Kibana listing for roles it seems that a role is classed as 'reserved' if either reserved, static and i assume hidden. Kibana lists alerting_view_alerts as a reserved role but you state above that the rolesmapping will forbid mapping a user to a 'reserved' role but this doesnt seem to be the case (OD 1.8.0):

```
PUT _opendistro/_security/api/rolesmapping/alerting_view_alerts
{
  "backend_roles": [],
  "users": [
    "aduser"
  ]
}

{
  "status" : "OK",
  "message" : "'alerting_view_alerts' updated."
}
```

```
GET _opendistro/_security/api/roles/alerting_view_alerts
{
  "alerting_view_alerts" : {
    "reserved" : true,
    "hidden" : false,
...
    "static" : false
  }
}
```

Can you clarify the terminology used and how this applies to the JSON config?

Thanks

PR #495, opened 2020-06-05
Renaming metadata_body to metadata_content

debjanibnrj:
*Issue #, if available:*

*Description of changes:*
Based on @dinusX  comment -
It seems that SAML already has a couple of settings that are similar to “metadata_body” where they keep the body of a file.
However all these settings end with "_content" suffix.
https://opendistro.github.io/for-elasticsearch-docs/docs/security-configuration/saml/

We should rename “meatadata_body” to “metadata_content” to keep setting names consistent, while it’s not too late.


src/main/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticator.java
```diff
@@ -288,15 +288,15 @@ private MetadataResolver createMetadataResolver(final Settings settings, final P
 
         final String idpMetadataUrl = settings.get("idp.metadata_url");
         final String idpMetadataFile = settings.get("idp.metadata_file");
-        final String idpMetadataBody = settings.get("idp.metadata_body");
+        final String idpMetadataBody = settings.get("idp.metadata_content");
```
vrozov, line 291: Please introduce `public static final String`.

debjanibnrj: These values are being set in a private method. This will need us to define it at class level and initialize the values in the constructor. Any reason for preferring this?

vrozov: It is a standard practice to define constant string as a class level variable.

Issue #376, opened 2020-04-13
Client log spam: "WARNING: Compact JWS does not have 3 parts"


DandyDeveloper:
This is similar to the issue reported here: https://github.com/opendistro-for-elasticsearch/security/issues/176

However, I am not impacted too hard by this problem I would just like to know where / how we can prevent this warning from happening.

All the client nodes are receiving events from a fluentd instance and are reporting this when using a hard-coded user in a OIDC setup environment: 

```
WARNING: Compact JWS does not have 3 parts
Apr 12, 2020 9:00:17 PM org.apache.cxf.rs.security.jose.jws.JwsCompactConsumer <init>
WARNING: Compact JWS does not have 3 parts
Apr 12, 2020 9:00:17 PM org.apache.cxf.rs.security.jose.jws.JwsCompactConsumer <init>
WARNING: Compact JWS does not have 3 parts
```

Since upgrading to 1.4.0 we've got an additional message which shows the class handling the token.. However, this is just adding to the spam. 

Does anyone know why this occurs?

DandyDeveloper:
Looks like it's what you'd expect: 

https://github.com/apache/cxf/blob/master/rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jws/JwsCompactConsumer.java#L52

The problem is clear but not the root cause... I'm guessing the tokens generated for users when OIDC is enabled is different from normal logins?

DandyDeveloper:
@hardik-k-shah @shivdoshi 

Can I get some suggestions on this?

wyllys66:
I see the same issue using Keycloak as the OpenId provider. Using opendistro version 1.7.0

dinusX:
Hi,

According to the JWT documentation the token should have 3 parts: header, body and signature.
I'm suspecting that maybe your JWT tokens don't have signature.

Could you take one of the tokens and try to crack it using the following website: https://jwt.io/
This should give you more data about the missing part.

Also, check in your application if there is any setting to enable signing of the jwt tokens.

DandyDeveloper:
@dinusX Which JWT? It looks like there are session tokens generated through the OIDC proivder login. 

Where is the JWT stored for a session auth? I looked at my login chain and I see no JWT generated at any point. Is this a server side request that we don't see?

DandyDeveloper:
@dinusX Any news on this? I'm still a bit confused on which JWT you're referring to here?

[...]

PR #501, opened 2020-06-11
Avoid using Basic Authorization header as JWT token

vrozov:
*Issue #, if available:* #376

*Description of changes:*
When Authorization header includes "Basic" it can't be JWT token, so we don't need to try to parse it as JWT.

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
+++ b/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
@@ -137,7 +137,7 @@ public abstract class AbstractHTTPJwtAuthenticator implements HTTPAuthenticator 
 
     protected String getJwtTokenString(RestRequest request) {
         String jwtToken = request.header(jwtHeaderName);
-        if (isDefaultAuthHeader && BASIC.matcher(jwtToken).matches()) {
+        if (isDefaultAuthHeader && jwtToken != null && BASIC.matcher(jwtToken).matches()) {
             jwtToken = null;
         }
 
--- a/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
+++ b/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
@@ -20,9 +20,11 @@ import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Collection;
 import java.util.Map.Entry;
+import java.util.regex.Pattern;
 
 import org.apache.cxf.rs.security.jose.jwt.JwtClaims;
 import org.apache.cxf.rs.security.jose.jwt.JwtToken;
+import org.apache.http.HttpHeaders;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.elasticsearch.ElasticsearchSecurityException;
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
```diff
@@ -51,9 +53,12 @@
 
     protected final Logger log = LogManager.getLogger(this.getClass());
 
+    private static final Pattern BASIC = Pattern.compile("^\\s*Basic\\s.*", Pattern.CASE_INSENSITIVE);
```
dinusX, line 56: Instead of skipping basic Auth headers, can we skip any non "bearer" headers ?

vrozov: It sounds that "Bearer" scheme is optional. Please see `HTTPJwtAuthenticatorTest.testNonBearer()` that test that `HTTPJwtAuthenticator` can accept JWT in `Authorization` header without `Bearer`.

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
```diff
@@ -46,17 +48,20 @@
     private final static Logger log = LogManager.getLogger(AbstractHTTPJwtAuthenticator.class);
 
     private static final String BEARER = "bearer ";
+    private static final Pattern BASIC = Pattern.compile("^\\s*Basic\\s.*", Pattern.CASE_INSENSITIVE);
```
dinusX, line 51: According to the specifications the token will always start with "Basic ". I don't think we should allow spaces before.
https://tools.ietf.org/html/rfc2617#page-5

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
dinusX: I see.

Would be great if you can add a unit-test for this use-case.
In addition to this, there might be other "Authorization" headers that needs to be checked similarly as "Basic".

I would prefer also to discuss this internally if we should push this change, because usually this issue is happening when users configure incorrectly the order of authentication configurations.

vrozov: > In addition to this, there might be other "Authorization" headers that needs to be checked similarly as "Basic".

Can be addressed in a follow up PR if necessary.

> I would prefer also to discuss this internally if we should push this change, because usually this issue is happening when users configure incorrectly the order of authentication configurations.

Actually it is the correct order. Putting Basic authenticator in front will cause `No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'` warning message even when a valid JWT token is present and inability to login with JWT. Please see [HTTPHelper](https://github.com/opendistro-for-elasticsearch/security/blob/6fe4e0ffca29af61afa8492d7afd5846a3e6753a/src/main/java/com/amazon/opendistroforelasticsearch/security/support/HTTPHelper.java#L48)

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
vrozov: As we don't need to ensure that it is a valid "Basic" authentication scheme here, but only need to filter out invalid JWT tokens to avoid JWT validation warnings, it is OK to allow spaces before "Basic" to be consistent with `trim()` used in [HttpHelper.java](https://github.com/opendistro-for-elasticsearch/security/blob/6fe4e0ffca29af61afa8492d7afd5846a3e6753a/src/main/java/com/amazon/opendistroforelasticsearch/security/support/HTTPHelper.java#L48)

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
Gauravbtc: @dinusX how we can set JWT request header token in request payload

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
sujithvm, line 51: Factor out BASIC pattern as it is re-used ?

sujithvm: Out of scope for PR but any reason why HTTPJwtAuthenticator 
 does not inherit from AbstractHTTPJwtAuthenticator ?

vrozov: I don't see any good reason, will be good to cleanup this in a follow up PR.

dinusX:
For some reason, I’m not seeing your comment in the PR. Replying over email.

You have 2 options to set JWT token.
- First option is using the “Authorization” header. “Authorization: Bearer <jwt_token>”
- Second option is to use a custom header, which you can configure using “jwt_header” setting. (ex: “jwt_header”: “MyCustomHeader”. Request: “MyCustomHeader: Bearer <jwt_token”.
- And the third option is to use an url parameter that you’ll have to send as part of url request path.  You can configure it using “jwt_url_parameter” setting. (ex: GET /_cat/indices?jwt_token=<token_value> )

For more details see:
https://opendistro.github.io/for-elasticsearch-docs/docs/security-configuration/configuration/

PR #506, opened 2020-06-15
Refactor salt from compliance config into Salt class

sujithvm:
Changes:
- Removing salt from ComplianceConfig as it is not reloaded as part of it. Enabling/Disabling compliance for audit logging should not change the field anonymization behavior. 
- Refactored the salt usage into a separate FLS configuration class (inner class of DlsFlsConfig which could store other DlsFls configurations)
- Moved all the associated tests. 
- OpenDistroSecurityFlsDlsIndexSearcherWrapper was called per index, so the config had to be injected down from top level into DlsFlsFilterLeafReader (where its being used) to avoid being created everytime

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/DlsFlsFilterLeafReader.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/DlsFlsFilterLeafReader.java
@@ -106,6 +106,7 @@ class DlsFlsFilterLeafReader extends FilterLeafReader {
     private final MaskedFieldsMap maskedFieldsMap;
     private final ShardId shardId;
     private final boolean maskFields;
+    private final Salt salt;
 
     private DlsGetEvaluator dge = null;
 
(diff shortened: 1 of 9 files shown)


vrozov:
IMO, it is an overkill to introduce 2 new classes to hold a single byte array.

sujithvm:
Could not find another place to put this setting and byte array is subjected to modification so didn't want inject the array directly. 
However I will remove the inner class for Fls if we don't have a lot of configuration around it. 
Let me know if theres another place you would consider putting this setting.

sujithvm:
Refactored to a single Salt class

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/DlsFlsFilterLeafReader.java
```diff
@@ -106,23 +106,25 @@
...
-        maskFields = (auditlog.getComplianceConfig().isEnabled() && maskedFields != null && maskedFields.size() > 0);
+        maskFields = (maskedFields != null && maskedFields.size() > 0);
 
         this.indexService = indexService;
         this.threadContext = threadContext;
         this.clusterService = clusterService;
         this.auditlog = auditlog;
-        this.maskedFieldsMap = MaskedFieldsMap.extractMaskedFields(maskFields, maskedFields, auditlog.getComplianceConfig().getSalt16());
+        this.salt = salt;
+        this.maskedFieldsMap = MaskedFieldsMap.extractMaskedFields(maskFields, maskedFields, this.salt.getSalt16());
```
vrozov, line 127: nit: pass `Salt` to `MaskedFieldsMap.extractMaskedFields` and add `private final Salt salt` to `MaskedFieldsMap`.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/Salt.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Portions Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
vrozov, line 2: I don't think it is necessary to limit copyright for this file

src/test/java/com/amazon/opendistroforelasticsearch/security/configuration/SaltTest.java
```diff
@@ -0,0 +1,76 @@
+package com.amazon.opendistroforelasticsearch.security.configuration;
```
vrozov, line 1: Add copyright

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/DlsFlsFilterLeafReader.java
```diff
@@ -106,23 +106,25 @@
...
 
     DlsFlsFilterLeafReader(final LeafReader delegate, final Set<String> includesExcludes,
                            final Query dlsQuery, final IndexService indexService, final ThreadContext threadContext,
                            final ClusterService clusterService,
-                           final AuditLog auditlog, final Set<String> maskedFields, final ShardId shardId) {
+                           final AuditLog auditlog, final Set<String> maskedFields, final ShardId shardId, final Salt salt) {
         super(delegate);
 
-        maskFields = (auditlog.getComplianceConfig().isEnabled() && maskedFields != null && maskedFields.size() > 0);
+        maskFields = (maskedFields != null && maskedFields.size() > 0);
```
vrozov, line 120: Is the similar check on line 1099 necessary? If it is removed here, should it be also removed in `getRuntimeMaskedFieldInfo()`?

PR #507, opened 2020-06-15
Support ES 7.7.1

vrozov:


sujithvm:
Do we need to bump the plugin version also as a different ES version is being supported? 1.8.1.0

PR #513, opened 2020-06-19
Allow superadmin to update/delete hidden resources required for tenan…

palashhedau:
*Description of changes:*

1. Allow superAdmin to update/delete hidden resources required for tenantinfo check. This can update the hidden roles which will be used by tenantinfo check.
2. Allow superAdmin to see hidden resources 

The user required to start the kibana process will have kibana_server roles which is currently hidden. The kibana_server will give privileges to properly start kibana and its component. So updating hidden role is required to add the user to the hidden role.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
@@ -272,7 +272,9 @@ public abstract class AbstractApiAction extends BaseRestHandler {
 	}
 
 	protected void filter(SecurityDynamicConfiguration<?> builder) {
-		builder.removeHidden();
+		if (!isSuperAdmin()){
+			builder.removeHidden();
+		}
 		builder.set_meta(null);
 	}
 
@@ -513,8 +515,7 @@ public abstract class AbstractApiAction extends BaseRestHandler {
 	}
 
 	protected final boolean isHidden(SecurityDynamicConfiguration<?> configuration, String resourceName) {
-		final Object o = configuration.getCEntry(resourceName);
-		return o != null && o instanceof Hideable && ((Hideable) o).isHidden();
+		return configuration.isHidden(resourceName) && !isSuperAdmin();
 	}
 
 	protected final boolean isStatic(SecurityDynamicConfiguration<?> configuration, String resourceName) {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
@@ -121,7 +121,7 @@ public class InternalUsersApiAction extends PatchableResourceApiAction {
             final SecurityDynamicConfiguration<?> rolesConfiguration = load(CType.ROLES, false);
             for (final String role: opendistroSecurityRoles) {
 
-                if (!rolesConfiguration.exists(role) || isHidden(rolesConfiguration, role)) {
+                if (!rolesConfiguration.exists(role) || rolesConfiguration.isHidden(role)) {
                     notFound(channel, "Role '"+role+"' is not found.");
                     return;
                 }
--- a/src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/ActionGroupsApiTest.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/ActionGroupsApiTest.java
@@ -173,20 +173,22 @@ public class ActionGroupsApiTest extends AbstractRestApiUnitTest {
 		Assert.assertEquals(HttpStatus.SC_CREATED, response.getStatusCode());
         Assert.assertFalse(response.getBody().contains("Resource 'GET_UT' is read-only."));
 
-        // -- GET_UT hidden resource, must be 404
+        // -- GET_UT hidden resource, must be 404 but super admin can find it
         rh.sendAdminCertificate = true;
         response = rh.executeGetRequest("/_opendistro/_security/api/actiongroups/INTERNAL", new Header[0]);
-        Assert.assertEquals(HttpStatus.SC_NOT_FOUND, response.getStatusCode());
+        Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
+        Assert.assertTrue(response.getBody().contains("\"hidden\":true"));
 
 		// -- DELETE hidden resource, must be 404
         rh.sendAdminCertificate = true;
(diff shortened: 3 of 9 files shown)


sujithvm:
Could you add more information why tenant info check needs super-admin to update hidden resources ?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -555,4 +557,12 @@ protected boolean isReservedAndAccessible(final SecurityDynamicConfiguration<?>
 		return true;
 	}
 
+	protected boolean isHiddenAndAccessible(final SecurityDynamicConfiguration<?> existingConfiguration,
```
vrozov, line 560: Is it necessary to introduce this method? Should `isHidden()` be modified to exclude super admin?

palashhedau: I was thinking about modifying but there are some occurrences in the code where they require isHidden and hence can not modify.
Also it is consistent to the isReserved, but thats a second excuse.

vrozov: The PR makes "hidden" resource to be "visible" to super admin, should not the other occurrences of `isHidden()` be replaced too?

palashhedau: Most of them are replaced.
The only one which is not replaced is the GET request where we have to specify if it is a hidden resource is not in the response 

https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AccountApiAction.java#L140

vrozov: IMO, it will be better to change one place instead of changing multiple locations. Introduce `isHidden(String resourceName)` into `SecurityDynamicConfiguration` and call it directly from GET.

palashhedau: I can remove the isHidden function alltogether itself from AbstractApiAction as it might look like some code duplication. 
isHidden will point to SecurityDynamicConfiguration and wherever supderAdmin check is required will be added like 
`isHidden && superAdmin()`

Or do you want the way you said. IsHidden at both the location, only single GET request points to `SecurityDynamicConfiguration`

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -121,7 +121,7 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
             final SecurityDynamicConfiguration<?> rolesConfiguration = load(CType.ROLES, false);
             for (final String role: opendistroSecurityRoles) {
 
-                if (!rolesConfiguration.exists(role) || isHidden(rolesConfiguration, role)) {
+                if (!rolesConfiguration.exists(role) || rolesConfiguration.isHidden(role)) {
```
vrozov, line 124: It will be good to understand the desired behavior here. If super admin is able to see a role, should she be able to assign that role to a user? @debjanibnrj

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/ActionGroupsApiTest.java
```diff
@@ -174,17 +174,20 @@ public void testActionGroupsApi() throws Exception {
         Assert.assertFalse(response.getBody().contains("Resource 'GET_UT' is read-only."));
 
         // -- GET_UT hidden resource, must be 404
-        rh.sendAdminCertificate = true;
+        rh.sendAdminCertificate = false;
```
vrozov, line 177: Please add a super admin test case that checks that the resource is visible and marked as hidden.

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/RolesApiTest.java
```diff
@@ -144,7 +144,7 @@ public void testRolesApi() throws Exception {
 
         // hidden role
         response = rh.executeGetRequest("/_opendistro/_security/api/roles/opendistro_security_internal", new Header[0]);
-        Assert.assertEquals(HttpStatus.SC_NOT_FOUND, response.getStatusCode());
+        Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
```
vrozov, line 147: For "OK" response, should it also check the content of the response?

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/ActionGroupsApiTest.java
palashhedau: I didnt wanted it it be false as I have separate test cases for it.

I should have reverted it, but I guess, i missed out. Reverting now

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/RolesApiTest.java
palashhedau: we can check. I will add a few

palashhedau: We can add. Adding a few now

palashhedau: added

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/RolesApiTest.java
```diff
@@ -144,7 +144,8 @@ public void testRolesApi() throws Exception {
 
         // hidden role
         response = rh.executeGetRequest("/_opendistro/_security/api/roles/opendistro_security_internal", new Header[0]);
-        Assert.assertEquals(HttpStatus.SC_NOT_FOUND, response.getStatusCode());
+        Assert.assertTrue(response.getBody().contains("\"hidden\":true"));
+        Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
```
vrozov, line 148: nit: it is better to check for the status code first. When test does not fail, it does not matter, but may help with debugging if test fails. The same applies to all other places where response returns success code.

palashhedau: done

PR #516, opened 2020-06-23
Upgrade ES to 7.8.0

vrozov:
*Issue #, if available:*

*Description of changes:*
Upgrade ES version to 7.8.0 and plugin version to 1.9.0

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
@@ -64,6 +64,8 @@ import com.google.common.collect.ListMultimap;
 import com.google.common.collect.MultimapBuilder.SetMultimapBuilder;
 import com.google.common.collect.SetMultimap;
 
+import static org.elasticsearch.cluster.metadata.IndexAbstraction.Type.ALIAS;
+
 public class ConfigModelV7 extends ConfigModel {
 
     protected final Logger log = LogManager.getLogger(this.getClass());
@@ -556,7 +558,7 @@ public class ConfigModelV7 extends ConfigModel {
 
                         //#557
                         //final String[] allIndices = resolver.concreteIndexNames(cs.state(), IndicesOptions.lenientExpandOpen(), "*");
-                        final String[] allIndices = cs.state().metaData().getConcreteAllOpenIndices();
+                        final String[] allIndices = cs.state().metadata().getConcreteAllOpenIndices();
                         Arrays.stream(allIndices).filter(permitted).forEach(res::add);
                     }
                     retVal.addAll(res);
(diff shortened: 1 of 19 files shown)


debjanibnrj:
Following tests seem to have failed in the CI build - 

ProtectedIndicesTests.testAccessSnapshot
BasicAuditlogTest.testAuthenticated
SnapshotRestoreTests.testSnapshot
SnapshotRestoreTests.testSnapshotRestore
SnapshotRestoreTests.testSnapshotEnableSecurityIndexRestore
SnapshotRestoreTests.testSnapshotCheckWritePrivileges
IndexIntegrationTests.testIndices
HttpIntegrationTests.testHTTPBasic
HttpIntegrationTests.testBulk

src/test/resources/roles_bulk.yml
```diff
@@ -17,6 +17,6 @@ bulk:
     allowed_actions:
     - "indices:data/write/index"
     - "indices:admin/mapping/put"
-    - "indices:admin/create"
+    - "indices:admin/auto_create"
```
dinusX, line 20: Should you have here both ? "create" and "auto_create" ?

vrozov: bulk requires only "auto_create". I can add "create", but it won't be used.

dinusX: Looking at the ES PR, I see that they added "auto_create" everywhere where "create" was present, so ideally we should have both, just in case. 

In addition to these changes, could you review all the places were we use "indices:admin/create" ?
Example:
https://github.com/opendistro-for-elasticsearch/security/blob/cbcc6245bd923c545fc7e52dabe1bed267ddbf6c/src/main/resources/static_config/static_roles.yml

https://github.com/opendistro-for-elasticsearch/security/blob/a3d7a56e7c21ff011dad66a75389d157732f43fc/src/main/resources/static_config/static_action_groups.yml

https://github.com/opendistro-for-elasticsearch/security/blob/3aef3a3143f30fb95ac499742b80e2de75a355d1/src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigConstants.java

vrozov: We can follow the same approach and replace "auto_create" with "create" while checking if user is allowed to perform "auto_create" action.

Based on testing restoring snapshot does not require permission for "auto_create".

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/ConfigModelV7.java
```diff
@@ -743,8 +745,10 @@ public String getUnresolvedIndexPattern(User user) {
             WildcardMatcher matcher = WildcardMatcher.from(unresolved);
             String[] resolved = null;
             if (!(matcher instanceof WildcardMatcher.Exact)) {
-                final String[] aliasesForPermittedPattern = cs.state().getMetaData().getAliasAndIndexLookup().entrySet().stream()
-                        .filter(e -> e.getValue().isAlias()).filter(e -> matcher.test(e.getKey())).map(e -> e.getKey())
+                final String[] aliasesForPermittedPattern = cs.state().getMetadata().getIndicesLookup().entrySet().stream()
+                        .filter(e -> e.getValue().getType() == ALIAS)
```
debjanibnrj, line 749: nit: is ALIAS a string? `.equals` comparison may be better here.

vrozov: `ALIAS` is enum.

PR #517, opened 2020-06-23
Remove the faulty index exists check and have more predictable behavior

sujithvm:
*Issue #, if available:*

*Description of changes:*

- `clusterService.state().metaData().hasConcreteIndex(opendistrosecurityIndex)` checks if index exists with a cached copy and it incorrectly determines index is absent especially when the cluster bootstraps. 
- Removed the above check as its incorrectly functioning.
- added a try, catch around index creation to capture index/resource exists exception
- Updated log statements 
- Same behavior as we currently have

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
@@ -45,10 +45,12 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import com.google.common.collect.ImmutableMap;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.elasticsearch.ElasticsearchException;
 import org.elasticsearch.ExceptionsHelper;
+import org.elasticsearch.ResourceAlreadyExistsException;
 import org.elasticsearch.action.admin.cluster.health.ClusterHealthRequest;
 import org.elasticsearch.action.admin.cluster.health.ClusterHealthResponse;
 import org.elasticsearch.action.admin.indices.create.CreateIndexRequest;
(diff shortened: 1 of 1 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/ConfigurationRepository.java
```diff
@@ -126,30 +128,36 @@ public void run() {
...
-                                    if(ok) {
-                                        ConfigHelper.uploadFile(client, cd+"config.yml", opendistrosecurityIndex, CType.CONFIG, DEFAULT_CONFIG_VERSION);
-                                        ConfigHelper.uploadFile(client, cd+"roles.yml", opendistrosecurityIndex, CType.ROLES, DEFAULT_CONFIG_VERSION);
-                                        ConfigHelper.uploadFile(client, cd+"roles_mapping.yml", opendistrosecurityIndex, CType.ROLESMAPPING, DEFAULT_CONFIG_VERSION);
-                                        ConfigHelper.uploadFile(client, cd+"internal_users.yml", opendistrosecurityIndex, CType.INTERNALUSERS, DEFAULT_CONFIG_VERSION);
-                                        ConfigHelper.uploadFile(client, cd+"action_groups.yml", opendistrosecurityIndex, CType.ACTIONGROUPS, DEFAULT_CONFIG_VERSION);
-                                        if(DEFAULT_CONFIG_VERSION == 2) {
-                                            ConfigHelper.uploadFile(client, cd+"tenants.yml", opendistrosecurityIndex, CType.TENANTS, DEFAULT_CONFIG_VERSION);
+
+                                    try {
```
vrozov, line 132: Is this `try` necessary`?

sujithvm: I have put this try because it enables you to perform load other configs (audit.yml) or perform activities within the stored context block.

vrozov: nit: introduce a new method that creates index and returns true if index was created and false otherwise.

PR #542, opened 2020-07-01
Use Jackson to serialize and de-serialize audit configuration

sujithvm:
*Issue #, if available:*

*Description of changes:*

- Provide serialization & deserialization for audit configuration using Jackson. This is a preparatory PR for storing configuration in the security index.
- Inject settings present using Jackson which is used by compliance config
- All the defaults are based off what is currently implemented by reading off elasticsearch settings. 
- Added tests

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
@@ -36,13 +36,13 @@ import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.util.Map;
 
-import com.fasterxml.jackson.databind.InjectableValues;
 import org.elasticsearch.SpecialPermission;
 
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.InjectableValues;
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
@@ -34,6 +34,7 @@ import java.io.IOException;
 import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
+import java.util.Map;
 
 import com.fasterxml.jackson.databind.InjectableValues;
 import org.elasticsearch.SpecialPermission;
@@ -66,6 +67,16 @@ public class DefaultObjectMapper {
         objectMapper.setInjectableValues(injectableValues);
         YAML_MAPPER.setInjectableValues(injectableValues);
         defaulOmittingObjectMapper.setInjectableValues(injectableValues);
+    }
+
+    public static boolean getOrDefault(Map<String, Object> properties, String key, boolean defaultValue) {
+        Boolean value = (Boolean)properties.get(key);
+        return value != null ? value.booleanValue() : defaultValue;
+    }
+
+    public static <T> T getOrDefault(Map<String, Object> properties, String key, T defaultValue) {
+        T value = (T)properties.get(key);
+        return value != null ? value : defaultValue;
     }
 
     public static <T> T readTree(JsonNode node, Class<T> clazz) throws IOException {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
@@ -35,6 +35,7 @@ import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 
+import com.fasterxml.jackson.databind.InjectableValues;
 import org.elasticsearch.SpecialPermission;
 
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
@@ -59,6 +60,12 @@ public class DefaultObjectMapper {
         defaulOmittingObjectMapper.setSerializationInclusion(Include.NON_DEFAULT);
         defaulOmittingObjectMapper.enable(JsonParser.Feature.STRICT_DUPLICATE_DETECTION);
         YAML_MAPPER.enable(JsonParser.Feature.STRICT_DUPLICATE_DETECTION);
+    }
+
+    public static void inject(final InjectableValues.Std injectableValues) {
+        objectMapper.setInjectableValues(injectableValues);
+        YAML_MAPPER.setInjectableValues(injectableValues);
+        defaulOmittingObjectMapper.setInjectableValues(injectableValues);
     }
 
     public static <T> T readTree(JsonNode node, Class<T> clazz) throws IOException {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
@@ -19,6 +19,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import static com.amazon.opendistroforelasticsearch.security.DefaultObjectMapper.getOrDefault;
+
 /**
  * Class represents configuration for audit logging.
  * Expected class structure
@@ -60,8 +62,11 @@ import java.util.Set;
  */
 public class AuditConfig {
 
-    public static final Set<String> DEFAULT_IGNORED_USERS_SET = Collections.singleton("kibanaserver");
-    public static final List<String> DEFAULT_IGNORED_USERS_LIST = ImmutableList.copyOf(DEFAULT_IGNORED_USERS_SET);
+    public static final List<String> DEFAULT_IGNORED_USERS = Collections.singletonList("kibanaserver");
+
+    private AuditConfig() {
+        this(true, null, null);
+    }
 
     @JsonProperty("enabled")
     private final boolean auditLogEnabled;
@@ -80,12 +85,6 @@ public class AuditConfig {
 
     public ComplianceConfig getCompliance() {
         return compliance;
-    }
-
-    private AuditConfig() {
-        this.auditLogEnabled = true;
-        this.filter = Filter.DEFAULT;
-        this.compliance = ComplianceConfig.DEFAULT;
     }
 
     @VisibleForTesting
@@ -110,9 +109,7 @@ public class AuditConfig {
         private static final Filter DEFAULT = Filter.from(Settings.EMPTY);
 
         private final boolean isRestApiAuditEnabled;
-        private final EnumSet<AuditCategory> disabledRestCategories;
         private final boolean isTransportApiAuditEnabled;
-        private final EnumSet<AuditCategory> disabledTransportCategories;
         private final boolean resolveBulkRequests;
         private final boolean logRequestBody;
         private final boolean resolveIndices;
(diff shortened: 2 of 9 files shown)


vrozov:
Please see https://github.com/sujithvm/security/pull/192

sujithvm:
Index in JsonProperty is mainly used for ordering in tests and does not require the json to be any order.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -1,45 +1,130 @@
...
-    private AuditConfig() { }
+    private AuditConfig() {
+        this(true, null, null);
+    }
+
+    @JsonProperty("enabled")
+    private final boolean auditLogEnabled;
+    @JsonProperty("audit")
+    private final Filter filter;
+    @JsonProperty("compliance")
```
debjanibnrj, line 75: don't think you need @JsonProperty here since the variable name is compliance

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -55,12 +140,42 @@ private Filter(final boolean isRestApiAuditEnabled,
...
+        @JsonCreator
+        @VisibleForTesting
+        static Filter from(Map<String, Object> properties) {
+            final boolean isRestApiAuditEnabled = getOrDefault(properties,"enable_rest", true);
+            final boolean isTransportAuditEnabled = getOrDefault(properties,"enable_transport", true);
+            final boolean resolveBulkRequests = getOrDefault(properties, "resolve_bulk_requests", false);
+            final boolean logRequestBody = getOrDefault(properties, "log_request_body", true);
+            final boolean resolveIndices = getOrDefault(properties, "resolve_indices", true);
+            final boolean excludeSensitiveHeaders = getOrDefault(properties, "exclude_sensitive_headers", true);
+            final EnumSet<AuditCategory> disabledRestCategories = AuditCategory.parse(getOrDefault(properties,"disabled_rest_categories", ConfigConstants.OPENDISTRO_SECURITY_AUDIT_DISABLED_CATEGORIES_DEFAULT));
```
debjanibnrj, line 160: Think @vrozov had recently modified code for this here - https://github.com/opendistro-for-elasticsearch/security/pull/543/files#diff-c7afdd983cdd9cca05fa869112ff95b9R76-R77

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -53,25 +57,40 @@
...
  */
+@JsonAutoDetect(getterVisibility = JsonAutoDetect.Visibility.NONE)
 public class ComplianceConfig {
 
     private static final Logger log = LogManager.getLogger(ComplianceConfig.class);
+    public static final ComplianceConfig DEFAULT = ComplianceConfig.from(Settings.EMPTY);
     private static final int CACHE_SIZE = 1000;
     private static final String INTERNAL_ELASTICSEARCH = "internal_elasticsearch";
 
     private final boolean logExternalConfig;
```
debjanibnrj, line 79: Shouldn't this be @JsonProperty(value="external_config")? Any reason why it is not mentioned here and added here - https://github.com/opendistro-for-elasticsearch/security/pull/542/files#diff-0e04f575b10578faf6c21e321ea023b3R293 ?

Same for following lines - https://github.com/opendistro-for-elasticsearch/security/pull/542/files#diff-0e04f575b10578faf6c21e321ea023b3R80-R82?

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
```diff
@@ -34,7 +34,9 @@
 import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
+import java.util.Map;
 
+import com.fasterxml.jackson.databind.InjectableValues;
```
palashhedau, line 39: nit: can align imports with same parents

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
```diff
@@ -61,6 +63,22 @@
...
     }
 
+    public static void inject(final InjectableValues.Std injectableValues) {
+        objectMapper.setInjectableValues(injectableValues);
+        YAML_MAPPER.setInjectableValues(injectableValues);
+        defaulOmittingObjectMapper.setInjectableValues(injectableValues);
+    }
+
+    public static boolean getOrDefault(Map<String, Object> properties, String key, boolean defaultValue) {
+        Boolean value = (Boolean)properties.get(key);
```
palashhedau, line 73: can we directly cast as boolean?

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
```diff
@@ -105,17 +126,19 @@ private ComplianceConfig(
...
-                        split -> WildcardMatcher.from(split[0]),
-                        split -> split.length == 1 ?
-                                Collections.singleton("*") : Arrays.stream(split).skip(1).collect(ImmutableSet.toImmutableSet())
-                ));
+        this.watchedReadFields = watchedReadFields;
+        this.ignoredComplianceUsersForRead = ignoredComplianceUsersForRead;
+        this.watchedWriteIndicesPatterns = watchedWriteIndicesPatterns;
+        this.ignoredComplianceUsersForWrite = ignoredComplianceUsersForWrite;
+
+        this.readEnabledFields = watchedReadFields.entrySet().stream()
```
debjanibnrj, line 134: what is the difference between this and https://github.com/opendistro-for-elasticsearch/security/pull/542/files#diff-0e04f575b10578faf6c21e321ea023b3R255-R262? Why are we handling them differently?

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
sujithvm: It is of type Object in properties

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfigSerializeTest.java
```diff
@@ -0,0 +1,247 @@
...
+        objectMapper.setInjectableValues(iv);
+    }
+
+    @Test
+    public void testDefaultSerialize() throws JsonProcessingException {
+        // arrange
+        final AuditConfig audit = new AuditConfig(true, null, null);
+        // act
+        final String json = objectMapper.writeValueAsString(audit);
+        assertEquals("{" +
```
debjanibnrj, line 45: I remember using the minidev JSONObject object to create JSON objects for testing in unit tests here - https://github.com/opendistro-for-elasticsearch/security/blob/master/src/test/java/com/amazon/opendistroforelasticsearch/security/ssl/OpenDistroSecuritySSLReloadCertsActionTests.java#L67-L70. 

It helped make the tests little more readable for me. Just a suggestion in case your dealing with a lot of JSON object creation via testing

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
sujithvm: It is the same. Look below. 
This function is used by jackson and function at the bottom for elasticsearch yml settings

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
sujithvm: The other place converts a comma separated list provided from elasticsearch settings yml to Map.
Eg: 
`twitter,id,message` is converted to 
```
{ 
   "twitter*": ["id", "message"]
}
```

And this converts map to WildcardMatcher patterns.
```
{
   <twitter wilcard pattern>: ["id", "message"]
}
```

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
sujithvm: Because other properties are used internally within the class only. They dont have associated getter. 
Jackon prefers getter binding and if not available uses property binding.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
vrozov: By default only public properties are auto detected by Jackson (if I remember correctly).

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
vrozov: No, it will cause `NullPointerException` if the property is missing in the `properties`. Should `properties` always contain `key` property, it would be possible to cast directly to `boolean` (unboxing).

src/main/java/com/amazon/opendistroforelasticsearch/security/compliance/ComplianceConfig.java
vrozov: To be more exact, it is not Jackson that prefers getter. By annotating getters, we tell Jackson to use getters during serialization and use `@JsonCreator` during de-serialization. For those properties that don't have getters, there is no other options.

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfigSerializeTest.java
sujithvm: @debjanibnrj Looks like minidev is some internal dependency. I used XContentBuilder and Jackson to assert all jsons and also removed the index field making it cleaner.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
sujithvm: It seems to work because the property/getter with visibility off and variable name matches.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -114,9 +113,9 @@ public static AuditConfig from(final Settings settings) {
         private final boolean logRequestBody;
         private final boolean resolveIndices;
         private final boolean excludeSensitiveHeaders;
-        @JsonProperty(value = "ignore_users", index = 8)
+        @JsonProperty(value = "ignore_users")
```
vrozov, line 116:
```suggestion
        @JsonProperty("ignore_users")
```

src/test/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfigSerializeTest.java
```diff
@@ -0,0 +1,334 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.InjectableValues;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import org.elasticsearch.common.Strings;
+import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.common.xcontent.XContentBuilder;
+import org.elasticsearch.common.xcontent.XContentFactory;
```
vrozov, line 16: Any reason to use `XContentBuilder` instead of `ObjectMapper`?

sujithvm: It was cleaner to use XContentBuilder to build a json object compared to ObjectMapper. With object mapper, had to make separate child nodes and add them to main json object nodes whereas xcontentbuilder has a more fluid API. 
Main intent was clarity in creating json object for test and there seems to be many ways to do it.

vrozov: Object built with XContentBuilder needs to be converted to String and parsed using ObjectMapper anyway. By using ObjectMapper we can bypass that step.

sujithvm: Created issue #656

PR #543, opened 2020-07-03
AuditConfig and AuditCategory minor refactoring

vrozov:

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigConstants.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigConstants.java
@@ -30,11 +30,18 @@
 
 package com.amazon.opendistroforelasticsearch.security.support;
 
+import com.amazon.opendistroforelasticsearch.security.auditlog.impl.AuditCategory;
+
+import org.elasticsearch.common.settings.Settings;
+
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 
 public class ConfigConstants {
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigConstants.java
```diff
@@ -247,4 +256,12 @@
     public static final List<String> OPENDISTRO_SECURITY_PROTECTED_INDICES_DEFAULT = Collections.emptyList();
     public static final String OPENDISTRO_SECURITY_PROTECTED_INDICES_ROLES_KEY = "opendistro_security.protected_indices.roles";
     public static final List<String> OPENDISTRO_SECURITY_PROTECTED_INDICES_ROLES_DEFAULT = Collections.emptyList();
+
+    public static Set<String> getSettingAsSet(final Settings settings, final String key, final List<String> defaultList, final boolean ignoreCaseForNone) {
```
sujithvm, line 260: Would be nice if the util function is moved out of ConfigConstants (has only constants). 
This is used only in AuditCategory mainly because of "NONE" dependency which should be removed by #325 . Consider moving it there or address this change with 325 ?

vrozov: It is a shared utility function and I did not find a better place to put it without introducing a new class.

debjanibnrj: Out of curiousity, what are the drawbacks of introducing a new Util class for this function?

vrozov: Every java class requires it's metadata to be loaded into JVM and consumes memory. Not a big deal, but it may be better to avoid introducing a class with a single static method.

debjanibnrj: Thanks for the explanation

PR #546, opened 2020-07-06
Plugin Security - Initial changes for roles injection

skkosuri-amzn:
*Issue #, if available:*
To support below use cases: 
https://github.com/opendistro-for-elasticsearch/alerting/issues/6
https://github.com/opendistro-for-elasticsearch/alerting/issues/215
Similar use cases are present for other plugins.

*Description of changes:*
1. Added a way to inject roles using thread-context. This is for the plugin background jobs which doesn't have user context.
2. Added new API `/rolesinfo`, REST and transport layer call. This call is used to get user and roles details of the logged on user. I couldn't use `/authinfo` call, as it doesn't have a transport layer call.
3. To support `/rolesinfo` transport layer call, had to moved the request, response, action classes to commons library. (opendistro-commons). That means security plugin will have dependency on this library.
4. Unit tests are not added to this first version. If this approach looks good, will go ahead and tests.


plugin-descriptor.properties
```diff
@@ -22,4 +22,4 @@ java.version=1.8
 # elasticsearch release. This version is checked when the plugin
 # is loaded so Elasticsearch will refuse to start in the presence of
 # plugins with the incorrect elasticsearch.version.
-elasticsearch.version=7.7.1
+elasticsearch.version=7.7.0
```
skkosuri-amzn, line 25: Version change is only for testing on 7.7.0

pom.xml
```diff
@@ -68,7 +68,7 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <maven.compiler.release>8</maven.compiler.release>
 
-        <elasticsearch.version>7.7.1</elasticsearch.version>
+        <elasticsearch.version>7.7.0</elasticsearch.version>
```
skkosuri-amzn, line 71: Version change is only for testing on 7.7.0

skkosuri-amzn:
Creating PR's against master.

PR #548, opened 2020-07-08
Support "true" and "false" String to boolean conversion in DefaultObjectMapper.getOrDefault()

vrozov:

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
@@ -46,6 +46,8 @@ import com.fasterxml.jackson.databind.InjectableValues;
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
```diff
@@ -69,9 +71,25 @@ public static void inject(final InjectableValues.Std injectableValues) {
...
-        Boolean value = (Boolean)properties.get(key);
-        return value != null ? value.booleanValue() : defaultValue;
+    public static boolean getOrDefault(Map<String, Object> properties, String key, boolean defaultValue) throws JsonProcessingException {
+        Object value = properties.get(key);
+        if (value == null) {
+            return defaultValue;
+        } else if (value instanceof Boolean) {
+            return (boolean)value;
+        } else if (value instanceof String) {
+            String text = ((String)value).trim();
```
debjanibnrj, line 81:
```suggestion
            String text = ((String)value).trim().toLower();
            if ("true".equals(text)) {
```

vrozov: The only lower case "true" and capitalized "True" are supported ("TRUE" and "tRue" are not valid) the same way how Jackson default deserializer works.

PR #550, opened 2020-07-09
Add new permissions for Kibana 7.8

trexx:
*Description of changes:*

New indices introduced in Kibana 7.8 are missing permissions in the default configuration causing exceptions to be logged.

### DIFF
--- a/src/main/resources/static_config/static_roles.yml
+++ b/src/main/resources/static_config/static_roles.yml
@@ -87,6 +87,14 @@ kibana_server:
   - "indices:data/read/scroll*"
   index_permissions:
   - index_patterns:
+    - ".apm-custom-link"
+    allowed_actions:
+    - "indices_all"
+  - index_patterns:
+    - ".apm-agent-configuration"
+    allowed_actions:
+    - "indices_all"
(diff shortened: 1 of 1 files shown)


trexx:
Hi @vrozov indeed this change is for the proprietary Elastic stack.

The OpenDistro FAQ states:

> Can I use Open Distro for Elasticsearch plugins with the proprietary Elastic Stack?
> 
> Our plugins are tested to work with Open Distro for Elasticsearch and the OSS-only downloads from Elastic. They have not been tested with the default download from Elastic, which includes proprietary software, but we welcome the community to test and enhance individual plugins for compatibility with the Elastic Stack.

Is this not the case any more or have I misunderstood the FAQ?

vrozov:
> Is this not the case any more or have I misunderstood the FAQ?

@trexx You may still use the security plugin in a way that does not contradict Apache license and other licenses required to use the proprietary Elastic Search. It does not mean that ODFE release will ship with the default configuration for the proprietary Elastic Search and/or plugin(s). Feel free to modify your configuration in any way that does not violate the licenses.

PR #559, opened 2020-07-13
Implement ability to configure readonly fields for audit configuration

sujithvm:
*Issue #, if available:*

*Description of changes:*

- Ability to configure readonly configs. This is provided in elasticsearch.yml. Customer must provide Jackson based path (if incorrect will fail on startup)
- Audit GET request returns "_readonly" field containing the configured readonly properties.
- PUT/PATCH requests 409 (conflict) if readonly params are updated in the request. Performs Jackson JSON checking to verify if fields are updated. 
- Super-admin can update/bypass the readonly fields
- Changed the underlying type of disabled categories from EnumSet to ImmutableSet so that ordering of arrays for JSON is maintained for comparison of fields during readonly check. 


src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -383,4 +383,16 @@ public String toString() {
...
+            "/enable",
+            "/audit/enable_rest", "/audit/disabled_rest_categories",
+            "/audit/enable_transport", "/audit/disabled_transport_categories",
+            "/audit/resolve_bulk_requests", "/audit/log_request_body", "/audit/resolve_indices", "/audit/exclude_sensitive_headers",
+            "/audit/ignore_users", "/audit/ignore_requests",
+            "/compliance/enabled",
+            "/compliance/external_config", "/compliance/internal_config",
+            "/compliance/read_metadata_only", "/compliance/read_watched_fields", "/compliance/read_ignore_users",
+            "/compliance/write_metadata_only", "/compliance/write_log_diffs", "/compliance/write_watched_indices", "/compliance/write_ignore_users"
+    );
```
dinusX, line 397: Can we generate this list dynamically instead of hard-coding it?  
This is not flexible long-term in case if we add new configs.

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
```diff
@@ -881,8 +881,12 @@ public Settings additionalSettings() {
             settings.add(Setting.listSetting(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_IGNORE_REQUESTS, Collections.emptyList(), Function.identity(), Property.NodeScope)); //not filtered here
             settings.add(Setting.boolSetting(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_RESOLVE_BULK_REQUESTS, false, Property.NodeScope, Property.Filtered));
             settings.add(Setting.boolSetting(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_EXCLUDE_SENSITIVE_HEADERS, true, Property.NodeScope, Property.Filtered));
-    
-            
+            settings.add(Setting.listSetting(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_READONLY_FIELDS, Collections.emptyList(), Function.identity(), (list) -> {
+                if (!AuditConfig.FIELD_PATHS.containsAll(list)) {
+                    throw new IllegalArgumentException("Invalid resource paths for read-only configs passed.");
+                }
```
dinusX, line 887: Should we fail or ignore with a message in case the path does not exist ?

In the future if you remove one of the existing paths, it might cause ES to not start anymore if this setting was not updated.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiAction.java
```diff
@@ -153,6 +160,21 @@ protected void handlePut(final RestChannel channel, final RestRequest request, f
...
+        final SecurityDynamicConfiguration<?> configuration = load(getConfigName(), true);
+        filter(configuration);
+
+        final String resourcename = getResourceName();
+        if (!configuration.exists(resourcename)) {
+            notFound(channel, "Resource '" + resourcename + "' not found.");
+            return;
+        }
+
+        configuration.putCObject(READONLY_FIELD, readonlyFields);
```
dinusX, line 174: Can this by any chance override the existing list from the configuration?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/PatchableResourceApiAction.java
```diff
@@ -143,6 +143,11 @@ private void handleSinglePatch(RestChannel channel, RestRequest request, Client
             }
         }
 
+        if (isReadonlyFieldUpdated(existingResourceAsJsonNode, patchedResourceAsJsonNode)) {
+            request.params().clear();
+            conflict(channel, "Attempted to update readonly property.");
```
dinusX, line 148: Let's use "read-only" (with dash) in all the messages for consistency.

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiActionTest.java
```diff
@@ -52,6 +69,177 @@ public void testInvalidPath() throws Exception {
...
+        assertEquals(expectedStatus, response.getStatusCode());
+    }
+
+    private void testReadonlyBoolean(final ObjectNode json, final String config, final String resource) throws Exception {
+        final String resourcePath = "/config" + config + "/" + resource;
+        ((ObjectNode)json.at(config)).put(resource, true);
+        testPutRequest(json, HttpStatus.SC_OK, true);
+        ((ObjectNode)json.at(config)).put(resource, false);
+        testPutRequest(json, HttpStatus.SC_CONFLICT, false, adminCredsHeader);
+        testBooleanPatch(resourcePath, false, HttpStatus.SC_CONFLICT, adminCredsHeader);
```
dinusX, line 159: Should we similarly test with admin certificate in the additional Patch methods ?

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiActionTest.java
```diff
@@ -176,13 +359,21 @@ private void testPatchAction(final int expectedStatus, final Header... headers)
...
+    private void testBooleanPatch(final String patchResource, final boolean value, final int expected, final Header... headers) throws Exception {
+        RestHelper.HttpResponse response = rh.executePatchRequest(ENDPOINT, "[{\"op\": \"add\",\"path\": \"" + patchResource + "\",\"value\": " + value + "}]", headers);
+        assertEquals(expected, response.getStatusCode());
+        if (expected == HttpStatus.SC_OK) {
+            assertTrue(readTree(rh.executeGetRequest(ENDPOINT, headers).getBody()).at(patchResource).asBoolean());
+        }
+    }
+
     private void testBoolean(final String patchResource, final int expected, final Header... headers) throws Exception {
         // make true
```
dinusX, line 378: You can rewrite this method to reuse your new method "testBooleanPatch"

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -383,4 +383,16 @@ public String toString() {
                 .filter(settings::hasValue)
                 .collect(Collectors.toSet());
     }
+
+    public static final ImmutableSet<String> FIELD_PATHS = ImmutableSet.of(
+            "/enable",
+            "/audit/enable_rest", "/audit/disabled_rest_categories",
```
dinusX, line 389: Should this be "enable_rest" or "enable_rest_categories" ?

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiActionTest.java
sujithvm: I have added on line 146 to test all actions as super-admin which includes Patch methods

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiAction.java
sujithvm: No. This will not override existing list in configuration. This is not stored in the index. Just added in the GET response. `load(getConfigName(), true)` will return a new object with configuration from index.

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
sujithvm: I think it is better to fail.
The fields are updated to dynamically fetch all valid fields now. If customer has configured something wrong, then customers may ignore message and wonder why it is not working correctly. 

If some path is completely removed in the future, then it will force user to migrate which is desirable IMO. 
Any thoughts ?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
sujithvm: It is `disabled_rest_categories` as it blacklists categories ignored. This field is now dynamically fetched via Jackson properties of that class.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
sujithvm: Done. Using Jackson to find all the properties of a given class.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -385,4 +384,11 @@ public String toString() {
                 .filter(settings::hasValue)
                 .collect(Collectors.toSet());
     }
+
+    public static final Set<String> FIELD_PATHS = Stream.of(
```
vrozov, line 388: Use `Sets.union()`.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiAction.java
vrozov: Can it be part of the index? Why is it part of elasticsearch.yml?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditCategory.java
```diff
@@ -22,19 +23,18 @@
...
         if (categories.isEmpty())
-            return set;
+            return Collections.emptySet();
 
         return categories
                 .stream()
                 .map(String::toUpperCase)
                 .map(AuditCategory::valueOf)
-                .collect(Collectors.toCollection(() -> set));
+                .collect(ImmutableSet.toImmutableSet());
```
vrozov, line 34: Which order is important here and why? Should not natural order be sufficient?

sujithvm: If the entries are passed as array, then the enum set reorders based on the order of elements defined within EnumSet. 
For eg: `["AUTHENTICATED", "GRANTED_PRIVILEGES"]` gets reordered as `["GRANTED_PRIVILEGES", "AUTHENTICATED"]` as `GRANTED_PRIVILEGES comes` before `AUTHENTICATED `in EnumSet ordering. 

This breaks if the same put request is called twice because to compare two JSON arrays are equal, the order must be same too. Also in the UI, the elements gets reordered on refetching. To avoid this, using ImmutableSet so that ordering remains the same as the way user enters them.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiAction.java
sujithvm: Mainly because of the following reasons:
- Configs need same types for keys. Cannot have different implementing classes for keys in a configuration. For eg: internalusers all the keys are of same implementing type.
- Admins also cannot change the readonly properties. Super-admin can bypass them. Not likely these settings are going to be changed often once deployed. 
- Considered storing in "_meta" of the document but we currently do expose APIs to update them. If at all user, user wants to change the readonly fields, then using super-admin we must replace the entire document after fetching the current configuration instead of updating in elasticsearch.yml (and doing a rolling restart). 

Let me know your thoughts

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiActionTest.java
```diff
@@ -52,6 +81,177 @@ public void testInvalidPath() throws Exception {
...
+        List<String> actual = Streams.stream(readTree(response.getBody()).at("/config/audit/disabled_rest_categories").iterator())
+                .map(JsonNode::textValue)
+                .collect(Collectors.toList());
+        assertEquals(testCategories, actual);
+    }
+
+    @Test
+    public void testReadonlyApi() throws Exception {
+        final List<String> readonlyFields = ImmutableList.of("/audit/enable_rest", "/audit/disabled_rest_categories", "/audit/ignore_requests", "/compliance/read_watched_fields");
+        updateStaticResourceReadonly(readonlyFields);
```
sujithvm, line 105: Had to resort to updating the test file in test resources as the resource file could not be injected into AuditApiAction for testing. 
In tearDown clearing it. 

Attempted to change via Reflection to inject another value for the `STATIC_RESOURCE` but making tests to be unstable

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -201,6 +202,11 @@ protected void handlePut(final RestChannel channel, final RestRequest request, f
 			return;
 		}
 
+		if (isReadonlyFieldUpdated(Utils.convertJsonToJackson(existingConfiguration, false).get(getResourceName()), content)) {
```
vrozov, line 205: Delegate conversion from JSON to `isReadonlyFieldUpdated()`, otherwise that conversion impacts all API actions even one that blindly return `false`.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/config/AuditConfig.java
```diff
@@ -385,4 +383,10 @@ public String toString() {
                 .filter(settings::hasValue)
                 .collect(Collectors.toSet());
     }
+
+    public static final Set<String> FIELD_PATHS = Sets.union(
+            DefaultObjectMapper.getFields(AuditConfig.class, "/"),
+            Sets.union(
+                    DefaultObjectMapper.getFields(AuditConfig.Filter.class, "/audit/"),
```
vrozov, line 390: reuse `AuditConfig.Filter.FIELDS` and `ComplianceConfig.FIELDS`.

sujithvm: A prefix path needs to be added to represent the resource path hence did not reuse

sujithvm: Done

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiAction.java
sujithvm: Storing it in static_audit.yml resource file

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AuditApiAction.java
```diff
@@ -126,6 +139,16 @@ public AuditApiAction(final Settings settings,
         super(settings, configPath, controller, client, adminDNs, cl, cs, principalExtractor, privilegesEvaluator, threadPool, auditLog);
         this.privilegesEvaluator = privilegesEvaluator;
         this.threadContext = threadPool.getThreadContext();
+        try {
+            this.readonlyFields = DefaultObjectMapper.YAML_MAPPER
+                    .readValue(this.getClass().getResourceAsStream(STATIC_RESOURCE), new TypeReference<Map<String, List<String>>>() {})
+                    .get(READONLY_FIELD);
+            if (!AuditConfig.FIELD_PATHS.containsAll(this.readonlyFields)) {
+                throw new IllegalArgumentException("Invalid read-only field paths provided in file " + STATIC_RESOURCE);
```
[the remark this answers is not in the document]
sujithvm: What would you suggest? Is it fine to throw the same `StaticResourceException` ?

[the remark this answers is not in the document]
sujithvm: Throwing StaticResourceException

PR #587, opened 2020-07-27
Support REST API for logging audit messages

vrozov:
*Issue #, if available:*

*Description of changes:*


sujithvm:
Looks good if we are choosing this approach. Just a couple of comments: 

1) I wonder if it will be better to assume the consumer will create the full audit message and in the API action just validate and filter for ignore user/request/disabled rest categories. 
This way we won't have to replace properties in the current thread and make new request/channel. 

Only rationale is that, these requests have not reached elasticsearch so we don't have to force it to look like it's a request it has received.

Any thoughts? 

2) We can add POST handler to the existing AuditApiAction?

PR #609, opened 2020-07-31
Refactor opendistro_security_action_trace logger

vrozov:
*Issue #, if available:*

*Description of changes:* Encapsulate logic of `opendistro_security_action_trace` Logger

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
@@ -52,6 +52,9 @@ import com.amazon.opendistroforelasticsearch.security.configuration.OpenDistroSe
 import com.amazon.opendistroforelasticsearch.security.configuration.Salt;
 import com.amazon.opendistroforelasticsearch.security.ssl.rest.OpenDistroSecuritySSLReloadCertsAction;
 import com.amazon.opendistroforelasticsearch.security.ssl.rest.OpenDistroSecuritySSLCertsInfoAction;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.apache.lucene.search.QueryCachingPolicy;
 import org.apache.lucene.search.Weight;
 import org.bouncycastle.jce.provider.BouncyCastleProvider;
@@ -166,6 +169,8 @@ import com.google.common.collect.Lists;
 public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin implements ClusterPlugin, MapperPlugin {
 
     private static final String KEYWORD = ".keyword";
+    private static final Logger actionTrace = LogManager.getLogger("opendistro_security_action_trace");
+
     private final boolean dlsFlsAvailable;
     private boolean sslCertReloadEnabled;
     private volatile OpenDistroSecurityRestFilter securityRestHandler;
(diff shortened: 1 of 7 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -526,8 +527,8 @@ public String kibanaOpendistroRole() {
             additionalPermissionsRequired.addAll(ConfigConstants.OPENDISTRO_SECURITY_SNAPSHOT_RESTORE_NEEDED_WRITE_PRIVILEGES);
         }
 
-        if(actionTrace.isTraceEnabled() && additionalPermissionsRequired.size() > 1) {
-            actionTrace.trace(("Additional permissions required: "+additionalPermissionsRequired));
+        if (additionalPermissionsRequired.size() > 1) {
```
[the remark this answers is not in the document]
vrozov: No, that will break encapsulation and will not give any performance benefit.

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
```diff
@@ -189,6 +194,18 @@
     private volatile DlsFlsRequestValve dlsFlsValve = null;
     private volatile Salt salt;
 
+    public static boolean isActionTraceEnabled() {
```
sujithvm, line 197: Why not move to a separate class to fully encapsulate the logic ?

src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluator.java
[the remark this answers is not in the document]
sujithvm: Just trying to understand this. 
Is it not possible to check via the static method import you have done for other classes ? 
This is executed for every request computing the logging string will add to performance degradation ?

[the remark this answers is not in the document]
vrozov: I changed logging to smart logging format that avoids formatting log output unless the corresponding level (in this case `trace`) is enabled. Adding call to `isActionTraceEnabled()` will actually cause performance degradation when `trace` is enabled (due to extra call) and will not improve performance when it is disabled.

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
vrozov: The functionality is too small to justify for a separate class, IMO.

sujithvm: I agree these 3 functions may be small but I dont think OpenDistroSecurityPlugin is the right place for this.

vrozov: Please suggest which class will be a better place to move those methods (new class is not an option).

sujithvm: Curious as to why new class is not an option?

vrozov: - The number of new lines of code that will be necessary to introduce with a new class (class declaration and private constructor) would be comparable to number of lines introduced in these 3 methods.
- It helps to avoid using more memory in JVM metaspace (not a big deal)

sujithvm: IMO different functionality should warrant a separate class even if it is small and should not be comparing the number of lines changed. 
Isn't the goal of this PR to encapsulate the logic for action tracer ? 

Just curious to know how much more memory we are talking about

vrozov: The functionality is embedded into the `Logger actionTrace` member variable, there is no need to have a separate class for it.

The primary goal of the PR is to avoid excessive calls to `LogManager.getLogger()` in various places and especially every time `OpenDistroSecurityRequestHandler` instance is created.

> Just curious to know how much more memory we are talking about

Not much, but we can leave this memory to other classes where a separate class is better justified.

PR #616, opened 2020-08-03
Remove unnecessary check of remote address for null

vrozov:
*Description of changes:* There is no need to check remote TransportAddress for null before calling `PrivilegesEvaluator.mapRoles(...)`

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/OpenDistroSecurityIndexSearcherWrapper.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/OpenDistroSecurityIndexSearcherWrapper.java
@@ -122,11 +122,8 @@ public class OpenDistroSecurityIndexSearcherWrapper implements CheckedFunction<D
     }
 
     protected final boolean isPermittedOnIndex() {
-        final TransportAddress caller = (TransportAddress) this.threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
-        final User user = (User) threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
-        if (caller == null || user == null) {
-            return false;
-        }
+        final User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+        final TransportAddress caller = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
         final Set<String> securityRoles = evaluator.mapRoles(user, caller);
         if (allowedRolesMatcher.matchAny(securityRoles)) {
             return true;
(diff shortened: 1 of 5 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/OpenDistroSecurityIndexSearcherWrapper.java
```diff
@@ -122,11 +122,8 @@ protected final boolean isBlockedIndexRequest() {
     }
 
     protected final boolean isPermittedOnIndex() {
-        final TransportAddress caller = (TransportAddress) this.threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
-        final User user = (User) threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
-        if (caller == null || user == null) {
-            return false;
-        }
```
dinusX, line 129: Can you explain why removing this is OK ?

I don't see any checks for null in "mapRoles" method.
https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluator.java#L436

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/OpenDistroSecurityIndexSearcherWrapper.java
```diff
@@ -122,11 +122,8 @@ protected final boolean isBlockedIndexRequest() {
     }
 
     protected final boolean isPermittedOnIndex() {
-        final TransportAddress caller = (TransportAddress) this.threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
-        final User user = (User) threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
```
dinusX, line 126: Will this automatic object mapping work on some older versions of Java ?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/OpenDistroSecurityIndexSearcherWrapper.java
vrozov: - There are few other places where `mapRoles()` method is called without the check for the remote address to be not null.
- The `mapRoles()` call eventually translated to the call to `Set<String> map(final User user, final TransportAddress caller)` (either in `ConfigModelV6` or `ConfigModelV7`) that correctly handles cases when `caller` is `null`

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/OpenDistroSecurityIndexSearcherWrapper.java
vrozov: Yes, the type inference is available since 1.8 and it is not new for the project, there are few other places where it is already used.

Issue #617, opened 2020-08-03
SAML auth not working in 1.9.0.1


shawnz:
Hi there, after upgrading to 1.9.0.1, SAML auth stopped working for me. Downgrading to 1.9.0.0 fixed the issue. Here is the message I got:

```
[2020-08-03T14:29:10,029][INFO ][o.o.s.m.r.i.AbstractReloadingMetadataResolver] [node-1] Metadata Resolver SamlHTTPMetadataResolver com.amazon.dlic.auth.http.saml.HTTPSamlAuthenticator_1: New metadata successfully loaded for 'https://...'
[2020-08-03T14:29:10,053][INFO ][o.o.s.m.r.i.AbstractReloadingMetadataResolver] [node-1] Metadata Resolver SamlHTTPMetadataResolver com.amazon.dlic.auth.http.saml.HTTPSamlAuthenticator_1: Next refresh cycle for metadata provider 'https://...' will occur on '2020-08-03T21:29:09.760Z' ('2020-08-03T17:29:09.760-04:00' local time)
[2020-08-03T14:29:10,139][WARN ][c.a.o.s.s.ReflectionHelper] [node-1] Unable to enable 'com.amazon.dlic.auth.http.saml.HTTPSamlAuthenticator' due to java.lang.reflect.InvocationTargetException
[2020-08-03T14:29:10,140][ERROR][c.a.o.s.s.DynamicConfigModelV7] [node-1] Unable to initialize auth domain saml_auth_domain=AuthcDomain [http_enabled=true, transport_enabled=false, order=1, http_authenticator=HttpAuthenticator [challenge=true, type=saml, config={idp={metadata_url=https://..., entity_id=https://...}, sp={entity_id=https://...}, kibana_url=https://..., roles_key=..., exchange_key=...}], authentication_backend=AuthcBackend [type=noop, config={}], description=Authenticate via SAML] due to ElasticsearchException[java.lang.reflect.InvocationTargetException]; nested: InvocationTargetException; nested: ExceptionInInitializerError; nested: AccessControlException[access denied ("java.util.PropertyPermission" "org.apache.xml.security.ignoreLineBreaks" "write")];
org.elasticsearch.ElasticsearchException: java.lang.reflect.InvocationTargetException
        at com.amazon.opendistroforelasticsearch.security.support.ReflectionHelper.instantiateAAA(ReflectionHelper.java:244) ~[opendistro_security-1.9.0.1.jar:1.9.0.1]
[...]

PR #626, opened 2020-08-04
Fix AccessControlException during HTTPSamlAuthenticator initialization.

vrozov:
*Issue #, if available:* #617

*Description of changes:* call to `com.onelogin.saml2.settings.SettingsBuilder.build()` requires "org.apache.xml.security.ignoreLineBreaks" property write permission in OneLogin SAML Java toolkit since version 2.4.0

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/http/saml/Saml2SettingsProvider.java
+++ b/src/main/java/com/amazon/dlic/auth/http/saml/Saml2SettingsProvider.java
@@ -15,7 +15,9 @@
 
 package com.amazon.dlic.auth.http.saml;
 
+import java.security.AccessController;
 import java.security.PrivateKey;
+import java.security.PrivilegedAction;
 import java.util.AbstractMap;
 import java.util.Collection;
 import java.util.HashMap;
@@ -26,6 +28,8 @@ import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+
+import org.elasticsearch.SpecialPermission;
 import org.elasticsearch.common.settings.Settings;
 import org.joda.time.DateTime;
 import org.opensaml.core.criterion.EntityIdCriterion;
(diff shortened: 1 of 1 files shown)


debjanibnrj:
Could you mention how you tested these changes?

vrozov:
> Could you mention how you tested these changes?

The changes were tested manually by configuring security plugin with SAML authentication.

src/main/java/com/amazon/dlic/auth/http/saml/Saml2SettingsProvider.java
```diff
@@ -95,8 +99,8 @@ Saml2Settings get() throws SamlConfigException {
             // TODO allow overriding of IdP metadata?
             settingsBuilder.fromValues(configProperties);
             settingsBuilder.fromValues(new SamlSettingsMap(this.esSettings));
-
-            return settingsBuilder.build();
+            SpecialPermission.check();
+            return AccessController.doPrivileged((PrivilegedAction<Saml2Settings>) () -> settingsBuilder.build());
```
dinusX, line 103: This call will wrap all exceptions into PriviledgedActionException. Could you check if this method throws any exception and write code to propagate it correctly.

Example: https://github.com/opendistro-for-elasticsearch/security/blob/d62ac5334609ea92b913caa5e73917efb8b29043/src/main/java/com/amazon/dlic/auth/ldap2/LDAPAuthenticationBackend2.java#L102

vrozov: It only applies to checked exceptions, as `com.onelogin.saml2.settings.SettingsBuilder.build()` does not throw checked exceptions, the unwrapping similar to what is done in LDAPAuthenticationBackend2.java is not necessary.

dinusX: Answer to my question: https://docs.oracle.com/javase/8/docs/technotes/guides/security/doprivileged.html#handling_exceptions

PR #635, opened 2020-08-05
Opendistro 0.9

palashhedau:
Backport changes

Allowing SuperAdmin to delete hidden resources, so that tenantinfo api can have additional check for usertype with hidden resource

Tenantinfo api will check role of the user from security config, and if matched will allow to work with this api

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
@@ -145,7 +145,7 @@ public abstract class AbstractApiAction extends BaseRestHandler {
 
 		final Tuple<Long, Settings> existingAsSettings = loadAsSettings(getConfigName(), false);
 
-		if (isHidden(existingAsSettings.v2(), name)) {
+		if (!isHiddenAndAccessible(existingAsSettings.v2(), name)) {
 			notFound(channel, getResourceName() + " " + name + " not found.");
 			return;
 		}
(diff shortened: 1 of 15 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -581,4 +583,11 @@ protected boolean isReadOnlyAndAccessible(Settings settings, String name) {
 		return true;
 	}
 
+	protected boolean isHiddenAndAccessible(Settings settings, String name) {
```
dinusX, line 586: Something does not look correct here.
Method is "isHiddenAndAccessible" but it returns "false" when object is hidden and not super admin. 
This is  too confusing.

Can we name it instead  "isAccessible" or convert it to opposite "isInaccessible" or any other name that is less confusing.

src/main/java/com/amazon/opendistroforelasticsearch/security/rest/TenantInfoAction.java
```diff
@@ -127,7 +136,46 @@ public void accept(RestChannel channel) throws Exception {
...
+            String kibanaOpendistroRole = evaluator.kibanaOpendistroRole();
+            if (Strings.isNullOrEmpty(kibanaOpendistroRole)) {
+                return false;
+            }
+
+            Map<String, Object> kibanaOpendistroRoleEntry = (Map<String, Object>) rolesMappingConfiguration.getOrDefault(kibanaOpendistroRole, null);
+
+            if (kibanaOpendistroRoleEntry != null) {
+                ArrayList<String> kibanaOpendistroRoleUsers = (ArrayList<String>) kibanaOpendistroRoleEntry.get("users");
+                return kibanaOpendistroRoleUsers != null && kibanaOpendistroRoleUsers.contains(user.getName());
```
dinusX, line 167: I just now realized that this check is done only by username. The role mapping can be  based on backend role as well. Is there any other way to make this check based on both criterias.

PR #637, opened 2020-08-07
Clean up rest and transport header filtering

sujithvm:
*Issue #, if available:*

*Description of changes:*

- Cleanup header filtering logic. Creating two maps in one flow. 
- Input can be immutable map hence making a map clone (retaining behavior)
- Using wildcard pattern
- Added tests 

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
@@ -26,9 +26,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
-import java.util.stream.Collectors;
 
 import com.amazon.opendistroforelasticsearch.security.auditlog.config.AuditConfig;
+import com.amazon.opendistroforelasticsearch.security.support.WildcardMatcher;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.http.client.utils.URIBuilder;
 import org.elasticsearch.ExceptionsHelper;
@@ -54,7 +54,7 @@ import com.amazon.opendistroforelasticsearch.security.dlic.rest.support.Utils;
 public final class AuditMessage {
 
     //clustername and cluster uuid
-    private static final String AUTHORIZATION_HEADER = "Authorization";
+    private static final WildcardMatcher AUTHORIZATION_HEADER = WildcardMatcher.from("Authorization", false);
     public static final String FORMAT_VERSION = "audit_format_version";
     public static final String CATEGORY = "audit_category";
     public static final String REQUEST_EFFECTIVE_USER = "audit_request_effective_user";
@@ -307,15 +307,11 @@ public final class AuditMessage {
 
     public void addRestHeaders(Map<String,List<String>> headers, boolean excludeSensitiveHeaders) {
         if(headers != null && !headers.isEmpty()) {
-            if(excludeSensitiveHeaders) {
-                final Map<String, List<String>> headersClone = new HashMap<String, List<String>>(headers)
-                        .entrySet().stream()
-                        .filter(map -> !map.getKey().equalsIgnoreCase(AUTHORIZATION_HEADER))
-                        .collect(Collectors.toMap(p -> p.getKey(), p -> p.getValue()));
-                auditInfo.put(REST_REQUEST_HEADERS, headersClone);
-            } else {
-                auditInfo.put(REST_REQUEST_HEADERS, new HashMap<String, List<String>>(headers));
+            final Map<String, List<String>> headersClone = new HashMap<>(headers);
+            if (excludeSensitiveHeaders) {
+                headersClone.keySet().removeIf(AUTHORIZATION_HEADER);
             }
+            auditInfo.put(REST_REQUEST_HEADERS, headersClone);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
```diff
@@ -307,15 +307,11 @@ public void addRestParams(Map<String,String> params) {
...
         if(headers != null && !headers.isEmpty()) {
-            if(excludeSensitiveHeaders) {
-                final Map<String, List<String>> headersClone = new HashMap<String, List<String>>(headers)
-                        .entrySet().stream()
-                        .filter(map -> !map.getKey().equalsIgnoreCase(AUTHORIZATION_HEADER))
-                        .collect(Collectors.toMap(p -> p.getKey(), p -> p.getValue()));
-                auditInfo.put(REST_REQUEST_HEADERS, headersClone);
-            } else {
-                auditInfo.put(REST_REQUEST_HEADERS, new HashMap<String, List<String>>(headers));
+            final Map<String, List<String>> headersClone = new HashMap<>(headers);
```
vrozov, line 310: -Is it necessary to create the clone even if `excludeSensitiveHeaders` is `false`?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
```diff
@@ -307,15 +307,11 @@ public void addRestParams(Map<String,String> params) {
...
+            if (excludeSensitiveHeaders) {
+                headersClone.keySet().removeIf(AUTHORIZATION_HEADER);
```
vrozov, line 312:
```
headers = Maps.filterEntries(headers, header -> !AUTHORIZATION_HEADER.test(header.getKey()));
```

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
```diff
@@ -339,15 +335,11 @@ void addRestRequestInfo(final RestRequest request, final AuditConfig.Filter filt
...
         if(headers != null && !headers.isEmpty()) {
-            if(excludeSensitiveHeaders) {
-                final Map<String,String> headersClone = new HashMap<String,String>(headers)
-                        .entrySet().stream()
-                        .filter(map -> !map.getKey().equalsIgnoreCase(AUTHORIZATION_HEADER))
-                        .collect(Collectors.toMap(p -> p.getKey(), p -> p.getValue()));
-                auditInfo.put(TRANSPORT_REQUEST_HEADERS, headersClone);
-            } else {
-                auditInfo.put(TRANSPORT_REQUEST_HEADERS, new HashMap<String,String>(headers));
+            final Map<String, String> headersClone = new HashMap<>(headers);
```
vrozov, line 338: Avoid code duplication. Merge `addTransportHeaders` and `addRestHeaders` to a single method with one more parameter.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
sujithvm: Input can be unmodifiable map. IMO its better to have consistency in what the method is setting for that attribute. It would assign unmodifiable map if excludeSensitiveHeaders is false or modifiable map if true. Retaining current behavior of setting modifiable map in either case.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
sujithvm: What is the advantage of doing this ?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
sujithvm: Both are accepting different map types.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
vrozov: Why does it matter? Where and why it can be modified?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
vrozov: There is no copy/clone, it is a read only filtered view on the original map.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
vrozov: Accept `Map<String, ?>`, value type actually does not matter.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
sujithvm: As i ve mentioned in the PR body, retaining the current behavior as it was specifically doing that. 
This may help in building a redactor module.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
sujithvm: Not in favor of merging `addTransportHeaders` and `addRestHeaders` into one function with accepting generics as value types will be permissive for consumers. 
Maybe the validation and filtering can be done as a separate function with generics. 
Intention of this PR was not to do a lot of refactor (check PR body). Maybe raise a separate issue?

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
vrozov: I don't see why is it necessary to keep the original behavior.  It should be redactor module responsibility to create a copy or filter original `Map` if necessary.

src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditMessage.java
vrozov: Yes, please open a follow up PR.
- avoid code duplication. IMO, it is OK to accept generics as AuditMessage does not care in what format Headers are provided.
- header filtering logic should avoid creating a new map and then removing elements from them.

sujithvm: Create #657

Issue #324, opened 2020-03-25
AuditLog: log a warning message if list of disabled categories contains a category that is not applicable


vrozov:

nateynateynate:
Hygiene. Milestone closed.

PR #638, opened 2020-08-07
Fail on invalid rest and transport categories

sujithvm:
*Issue #, if available:*

*Description of changes:*

- Addresses #324 to fail on invalid categories in the request.
- Not all the audit categories are valid for REST transport layer
- Fails with 400 bad request if invalid categories are passed

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AuditValidator.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AuditValidator.java
@@ -2,11 +2,35 @@ package com.amazon.opendistroforelasticsearch.security.dlic.rest.validation;
 
 import com.amazon.opendistroforelasticsearch.security.DefaultObjectMapper;
 import com.amazon.opendistroforelasticsearch.security.auditlog.config.AuditConfig;
+import com.amazon.opendistroforelasticsearch.security.auditlog.impl.AuditCategory;
+import com.google.common.collect.ImmutableSet;
 import org.elasticsearch.common.bytes.BytesReference;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.rest.RestRequest;
 
+import java.util.Set;
+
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AuditValidator.java
```diff
@@ -2,12 +2,36 @@
...
+import com.google.common.collect.ImmutableSet;
 import org.elasticsearch.common.bytes.BytesReference;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.rest.RestRequest;
 
+import java.util.Set;
+
 public class AuditValidator extends AbstractConfigurationValidator {
 
+    private static final Set<AuditCategory> REST_CATEGORIES = ImmutableSet.of(
```
debjanibnrj, line 15: nit: maybe name this DISABLED_REST_CATEGORIES so its more obvious.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/AuditValidator.java
```diff
@@ -2,12 +2,36 @@
...
+import com.google.common.collect.ImmutableSet;
 import org.elasticsearch.common.bytes.BytesReference;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.rest.RestRequest;
 
+import java.util.Set;
+
 public class AuditValidator extends AbstractConfigurationValidator {
 
+    private static final Set<AuditCategory> DISABLED_REST_CATEGORIES = ImmutableSet.of(
```
hardik-k-shah, line 15: why these categories are disable at rest layer? 
Aren't these all valid categories?

sujithvm: This is a validation check. They are valid categories for "disabled_rest_categories" field. Similarly for transport

PR #660, opened 2020-08-24
Using onelogin util helper instead of DocumentBuilderFactory

debjanibnrj:
… convertToString util helper

*Issue #, if available:*

*Description of changes:* Replaced DocumentBuilderFactory with onelogin util to prevent XXE attacks. Onelogin util has checks for XXE attacks. See - https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/util/Util.java#L335-L409

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticator.java
+++ b/src/main/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticator.java
@@ -90,7 +90,6 @@ public class HTTPSamlAuthenticator implements HTTPAuthenticator, Destroyable {
     private AuthTokenProcessorHandler authTokenProcessorHandler;
     private HTTPJwtAuthenticator httpJwtAuthenticator;
     private Settings jwtSettings;
-    private static final DocumentBuilderFactory documentBuilderFactory = getDocumentBuildFactory();
 
     private static int resolverIdCounter = 0;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticator.java
```diff
@@ -426,17 +425,9 @@ private String getSamlRequestQueryStringSignature(String samlRequestQueryString)
...
-        documentBuilderFactory.setNamespaceAware(true);
-        return documentBuilderFactory;
-    }
-
     private static Element getMetadataDOM(final String xmlString) throws IOException, SAXException, ParserConfigurationException {
-        DocumentBuilder builder = null;
         try {
-            builder = documentBuilderFactory.newDocumentBuilder();
-            Document doc = builder.parse(new InputSource(new StringReader(xmlString)));
+            Document doc = Util.convertStringToDocument(xmlString);
```
dinusX, line 430: Should we use "Util.loadXML(...)" method instead ?
https://github.com/onelogin/java-saml/blob/master/core/src/main/java/com/onelogin/saml2/util/Util.java#L162

debjanibnrj: Yes, looks like Util.loadXML method is better in this case.

PR #663, opened 2020-08-26
Close AuditLog while closing OpenDistroSecurityPlugin and unregister shutdown hook when closing AuditLogImpl.

vrozov:
*Description of changes:* In unit tests relying on shutdown hook leads to the memory leak as more and more hooks are added every time `AuditLogImpl` is constructed.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditLogImpl.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditLogImpl.java
@@ -49,6 +49,7 @@ public final class AuditLogImpl extends AbstractAuditLog {
 	private final boolean dlsFlsAvailable;
 	private final boolean messageRouterEnabled;
 	private volatile boolean enabled;
+	private final Thread shutdownHook;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/auditlog/impl/AuditLogImpl.java
```diff
@@ -85,23 +86,11 @@ public AuditLogImpl(final Settings settings,
...
-							close();
-						} catch (final IOException e) {
-							log.warn("Exception while shutting down message router", e);
-						}
-					}
-				});
-				log.debug("Shutdown Hook registered");
-				return null;
-			}
+		shutdownHook = new Thread(() -> messageRouter.close());
```
dinusX, line 89: Shouldn't this be ```this.close()``` instead of ```messageRouter.close()``` ?

vrozov: We can't call `this.close()` from the shutdown hook as it calls `removeShutdownHook()` while shutdown is in process.

PR #698, opened 2020-09-04
Add support for nested roles in OIDC config (e.g. keycloak roles (realm_access.roles))

volkflo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/security/issues/476
https://github.com/opendistro-for-elasticsearch/security/issues/551

*Description of changes:*

This should fix the problem of loading keycloak roles with OIDC.

After this PR the following setting is supported for nested JWT (`roles_key: realm_access.roles`):

```yaml
authc:
    openid_auth_domain:
        http_enabled: true
        transport_enabled: true
        order: 0
        http_authenticator:
        type: openid
        challenge: false
        config:
            subject_key: preferred_username
            roles_key: realm_access.roles
            openid_connect_url: https://my-keycloak-server.com:8443/auth/realms/demo/.well-known/openid-configuration
        authentication_backend:
        type: noop
```

This is a draft PR to get your opinion because I am new to opendistro I just wrote the code and the tests.
I still need to figure out how to run and how to debug the project. 

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
+++ b/src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
@@ -231,14 +231,13 @@ public class HTTPJwtAuthenticator implements HTTPAuthenticator {
 
     @SuppressWarnings("unchecked")
     protected String[] extractRoles(final Claims claims) {
-    	// no roles key specified
+        String[] roles = new String[0];
+        // no roles key specified
     	if(rolesKey == null) {
-    		return new String[0];
+    		return roles;
     	}
 
-        String[] roles = new String[0];
     	Object rolesObject;
-
     	if (rolesKey.contains(".")){
     	    // try to get nested roles e.g. keycloak roles
             rolesObject = getNestedRoles(claims);
@@ -247,6 +246,7 @@ public class HTTPJwtAuthenticator implements HTTPAuthenticator {
     	    rolesObject = claims.get(rolesKey, Object.class);
             if(rolesObject == null) {
                 log.warn("Failed to get roles from JWT claims with roles_key '{}'. Check if this key is correct and available in the JWT payload.", rolesKey);
+                return roles;
             } else {
                 roles = String.valueOf(rolesObject).split(",");
             }
--- a/src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
+++ b/src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
@@ -24,7 +24,9 @@ import java.security.PrivilegedAction;
 import java.security.PublicKey;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.X509EncodedKeySpec;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.regex.Pattern;
 
@@ -175,7 +177,7 @@ public class HTTPJwtAuthenticator implements HTTPAuthenticator {
             	return null;
             }
 
-            final String[] roles = extractRoles(claims, request);
+            final String[] roles = extractRoles(claims);
 
             final AuthCredentials ac = new AuthCredentials(subject, roles).markComplete();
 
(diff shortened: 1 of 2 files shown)


bekihm (dismissed):

bekihm:
@opendistroforelasticsearch-security Can you review and merge this PR please :-)

anirudha:
any update on this, when will this merge ?

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
```diff
@@ -256,6 +266,31 @@ protected String extractSubject(final Claims claims, final RestRequest request)
...
+    /**
+     * Returns the roles from a nested claim
+     * @param claims the claims
+     * @return the roles as ArrayList
+     */
+    private Object getNestedRoles(Claims claims) {
+        String[] keys = rolesKey.split("\\.");
+        Object value = null;
+
+        for (String key: keys) {
```
palashhedau, line 278: Please provide an example here as a comment, what this logic do exactly.
Provide some JSON and exaplanation.

palashhedau: Basically wanted to see what is the format of values in this case? Is it always map? What if is something different in case of other IDPs?

volkflo: The old functionality is untouched see (https://github.com/opendistro-for-elasticsearch/security/blob/35209bb9012c4b5d51bd4cea87046000b67c72b7/src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java#L208-L264)

Basically now you can use:
For keycloaks representation `roles_key: realm_access.roles` for 
```json
"realm_access": {
    "roles": [
      "elasticadmin",
      "offline_access",
      "devportaluser",
      "uma_authorization",
      "apiadmin",
      "apiuser"
    ]
  }
```
As it was before `roles_key: roles`
```json
"roles": [
    "elasticadmin",
    "offline_access",
    "devportaluser",
    "uma_authorization",
    "apiadmin",
    "apiuser"
  ]
```
You can find the full keycloak token here (note that it contain a workaround with a custom mapper to make it useable with opendistro): https://gist.github.com/volkflo/8ce36aa6861596c10e1092f1b39603c6

palashhedau (approved):

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
```diff
@@ -228,19 +230,27 @@ protected String extractSubject(final Claims claims, final RestRequest request)
...
-    	final Object rolesObject = claims.get(rolesKey, Object.class);
-    	if(rolesObject == null) {
-    		log.warn("Failed to get roles from JWT claims with roles_key '{}'. Check if this key is correct and available in the JWT payload.", rolesKey);
-    		return new String[0];
+    		return roles;
     	}
 
-    	String[] roles = String.valueOf(rolesObject).split(",");
+    	Object rolesObject;
+    	if (rolesKey.contains(".")){
```
hardik-k-shah, line 241: Question for my understanding: 
Can someone have "." in their roleskey  but not having actual nested roles?  
 Is there any other IdP who allows "." in their rolekey name without actually having nested roles?

src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java
```diff
@@ -228,6 +228,41 @@ public void testRoles() throws Exception {
         Assert.assertEquals(2, creds.getBackendRoles().size());
     }
 
+    @Test
+    public void testNestedRoles() throws Exception {
+        class NextedRoles implements Serializable {
```
cliu123, line 233: Please fix the typo `NestedRoles `

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
volkflo: I am not sure about other IdPs (we only use keycloak).
The keycloak default uses the "." for nested if you provide a custom claim. The default one is always nested.
image

volkflo: The main reason for this PR is that all examples in the [documentation ](https://opendistro.github.io/for-elasticsearch-docs/docs/security/configuration/openid-connect/#configure-openid-connect-integration) use keycloak.
But the examples will not work without custom claims that map the roles into a new claim.

src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java
volkflo: Fixed :)

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
hardik-k-shah: Based on your screenshot information, I think this IdP also supports using '.' literally into key name by escaping. That means key names can have '.' in their name but it will not be nested if '.' was escaped using '\'. 

In this case to avoid breaking customers who are using '.' literally in their key name, I think we should introduce some flag in configuration for nested claims. (like is_roles_key_nested: true  | false with default set to false). 

For ex.,

authc:
    openid_auth_domain:
        http_enabled: true
        transport_enabled: true
        order: 0
        http_authenticator:
        type: openid
        challenge: false
        config:
            subject_key: preferred_username
            roles_key: realm_access.roles
            is_roles_key_nested: true 
            openid_connect_url: https://my-keycloak-server.com:8443/auth/realms/demo/.well-known/openid-configuration
        authentication_backend:
        type: noop


thoughts?

palashhedau: Nested Roles can be used like this in the configuration 

A JSON path expression to the payload that stores the user’s roles, for example $["foo"]["user"]["roles"] where foo is the claim name in the token 

In your case, did you try adding secuirty configuration as ["realm_access"]["roles"]?

volkflo: I think roles_path is a good idea. However, I am not sure if I will be able to implement this and create the corresponding tests in the near future (due to limited time).

palashhedau: I can take this up and add it to my tasks.

hardik-k-shah: Created issue from 

> Just realised we dont have this functionality
> 
> We should introduced roles_path
> 
> `
> 
> ```java
> if (jsonRolesPath != null) {
>                 Object roles = JsonPath.using(jsonPathConfig).parse(claims.asMap()).read(jsonRolesPath);
>                 return splitRoles(roles);
>         
> ```
> 
> This way customer using the "." in their key in payload will not be affected

Created issue using this to track separately: https://github.com/opendistro-for-elasticsearch/security/issues/1003  

We can close this PR, if we are agree to this.  Thanks all.

PR #715, opened 2020-09-11
Validate resource names for config resources

sujithvm:
*Issue #, if available:*

*Description of changes:*

- Currently it is possible to create faulty resource names that may be used in API paths. For eg: `%3chtml%3e%3cscript%3etest123%3c%2fscript%3e%3c%2fhtml%3e`. This causes the UI also to render as HTML and API path cannot be recognized correctly (some actions like DELETE  won't work)
- Limit resource name creation to only alphanumeric characters, underscores or hyphens. 
- Avoid characters that may have special meaning in API paths or HTML

Screen Shot 2020-09-11 at 3 00 41 AM

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
@@ -31,6 +31,7 @@ import org.elasticsearch.action.support.WriteRequest.RefreshPolicy;
 import org.elasticsearch.client.Client;
 import org.elasticsearch.client.node.NodeClient;
 import org.elasticsearch.cluster.service.ClusterService;
+import org.elasticsearch.common.Strings;
 import org.elasticsearch.common.bytes.BytesReference;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.common.util.concurrent.ThreadContext.StoredContext;
@@ -177,8 +178,7 @@ public abstract class AbstractApiAction extends BaseRestHandler {
 
 		final String name = request.param("name");
 
-		if (name == null || name.length() == 0) {
-			badRequestResponse(channel, "No " + getResourceName() + " specified.");
+		if (isResourceNameInvalid(channel, name)) {
 			return;
 		}
 
@@ -612,4 +612,16 @@ public abstract class AbstractApiAction extends BaseRestHandler {
 		return true;
 	}
 
+	boolean isResourceNameInvalid(final RestChannel channel, final String resourceName) {
+		if (Strings.isNullOrEmpty(resourceName)) {
+			badRequestResponse(channel, "No " + getResourceName() + " specified.");
+			return true;
+		}
+
+		if (!ConfigConstants.OPENDISTRO_SECURITY_RESTAPI_RESOURCE_NAME_REGEX.matcher(resourceName).matches()) {
+			badRequestResponse(channel, "Resource name must contain only alphanumeric characters, underscores or hyphens.");
+			return true;
+		}
+		return false;
+	}
 }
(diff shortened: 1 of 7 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -177,8 +178,7 @@ protected void handlePut(final RestChannel channel, final RestRequest request, f
 
 		final String name = request.param("name");
 
-		if (name == null || name.length() == 0) {
-			badRequestResponse(channel, "No " + getResourceName() + " specified.");
+		if (isResourceNameInvalid(channel, name)) {
```
debjanibnrj, line 181: what happens to existing users that are using special characters in their name?

vrozov:
> Currently it is possible to create faulty resource names that may be used in API paths. For eg: %3chtml%3e%3cscript%3etest123%3c%2fscript%3e%3c%2fhtml%3e. This causes the UI also to render as HTML and API path

It is UI issue, not a backend. UI needs properly escape special (for UI rendering) characters. This includes both internal and external users and roles.

src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigConstants.java
```diff
@@ -236,6 +237,7 @@
     public static final String OPENDISTRO_SECURITY_RESTAPI_ENDPOINTS_DISABLED = "opendistro_security.restapi.endpoints_disabled";
     public static final String OPENDISTRO_SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX = "opendistro_security.restapi.password_validation_regex";
     public static final String OPENDISTRO_SECURITY_RESTAPI_PASSWORD_VALIDATION_ERROR_MESSAGE = "opendistro_security.restapi.password_validation_error_message";
+    public static final Pattern OPENDISTRO_SECURITY_RESTAPI_RESOURCE_NAME_REGEX = Pattern.compile("^[a-zA-Z0-9_-]+$"); // alphanumeric characters, underscores and hyphens
```
vrozov, line 240: This restriction is too strong and it will prevent resources with "." or any non ASCII letters.

sujithvm: Why is it desirable to create resources with non ASCII characters for well defined security configurations like internalusers, roles, action groups ... ? 

Even characters like "." needs to be URL encoded which makes it less readable anyways.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
sujithvm: Users can still GET and DELETE resources. The restriction is on PUT. They will have to DELETE and PUT resource again with validated names if they want to modify the same resource 

There are characters which are blocked in the UI already https://github.com/opendistro-for-elasticsearch/security-kibana-plugin/blob/master/public/apps/configuration/sections/internalusers/controller.js#L118 . Same restriction is not applicable in the REST API

vrozov: It is UI that should follow REST API restrictions, not reverse.

src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigConstants.java
vrozov: User names must support all languages (not only English) and characters like . and @ are quite common in user names as well.

sujithvm: Fair point. Will consult for recommendation and close the issue for now

PR #716, opened 2020-09-11
Change status code for hidden resources to NOT_FOUND and read-only sources to METHOD_NOT_ALLOWED

sujithvm:
*Issue #, if available:*

*Description of changes:*

- Some places return NOT_FOUND (404) for hidden resource and other places return FORBIDDEN (403). Consolidate them to return NOT_FOUND for resources that not available as user may have authorization to perform the request and FORBIDDEN is not appropriate.

- Return METHOD_NOT_ALLOWED(405) when read-only resources are updated as user may have authorization to perform the request and FORBIDDEN is not appropriate

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
@@ -146,13 +146,7 @@ public abstract class AbstractApiAction extends BaseRestHandler {
 
 		final SecurityDynamicConfiguration<?> existingConfiguration = load(getConfigName(), false);
 
-		if (isHidden(existingConfiguration, name)) {
-			notFound(channel, getResourceName() + " " + name + " not found.");
-			return;
-		}
-
-		if (isReadOnly(existingConfiguration, name)) {
-			forbidden(channel, "Resource '"+ name +"' is read-only.");
+		if (isWriteable(channel, existingConfiguration, name)) {
 			return;
 		}
 
@@ -189,13 +183,7 @@ public abstract class AbstractApiAction extends BaseRestHandler {
 		    return;
 		}
 
-		if (isHidden(existingConfiguration, name)) {
-			forbidden(channel, "Resource '"+ name +"' is not available.");
-			return;
-		}
-
-		if (isReadOnly(existingConfiguration, name)) {
-			forbidden(channel, "Resource '"+ name +"' is read-only.");
+		if (isWriteable(channel, existingConfiguration, name)) {
 			return;
 		}
 
@@ -529,6 +517,10 @@ public abstract class AbstractApiAction extends BaseRestHandler {
 				"Method " + method.name() + " not supported for this action.");
 	}
 
+	void methodNotAllowed(RestChannel channel, String message) {
+        response(channel, RestStatus.METHOD_NOT_ALLOWED, message);
+    }
+
 	protected final boolean isReserved(SecurityDynamicConfiguration<?> configuration, String resourceName) {
 		if(isStatic(configuration, resourceName)) { //static is also always reserved
 			return true;
@@ -606,10 +598,23 @@ public abstract class AbstractApiAction extends BaseRestHandler {
 		}
 
 		if (isReadOnly(rolesMappingConfiguration, role)) {
-			forbidden(channel, "Role '" + role + "' has read-only role-mapping.");
+			methodNotAllowed(channel, "Role '" + role + "' has read-only role-mapping.");
 			return false;
 		}
 		return true;
 	}
 
+	boolean isWriteable(final RestChannel channel, final SecurityDynamicConfiguration<?> configuration, final String name) {
+		if (isHidden(configuration, name)) {
+			notFound(channel, "Resource '"+ name +"' is not available.");
+			return true;
+		}
+
+		if (isReadOnly(configuration, name)) {
+            methodNotAllowed(channel, "Resource '"+ name +"' is read-only.");
+			return true;
+		}
+
+		return false;
+	}
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
@@ -146,7 +146,7 @@ public abstract class AbstractApiAction extends BaseRestHandler {
 
 		final SecurityDynamicConfiguration<?> existingConfiguration = load(getConfigName(), false);
 
-		if (isWriteable(channel, existingConfiguration, name)) {
+		if (!isWriteable(channel, existingConfiguration, name)) {
 			return;
 		}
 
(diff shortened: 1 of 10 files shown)


debjanibnrj (approved):

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -606,10 +594,23 @@ protected boolean isValidRolesMapping(final RestChannel channel, final String ro
...
 
 		if (isReadOnly(rolesMappingConfiguration, role)) {
-			forbidden(channel, "Role '" + role + "' has read-only role-mapping.");
+			conflict(channel, "Role '" + role + "' has read-only role-mapping.");
 			return false;
 		}
 		return true;
 	}
 
+	boolean isResourceWriteProtected(final RestChannel channel, final SecurityDynamicConfiguration<?> configuration, final String name) {
```
vrozov, line 603: rename to `isWriteable()`

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -606,10 +594,23 @@ protected boolean isValidRolesMapping(final RestChannel channel, final String ro
 		}
 
 		if (isReadOnly(rolesMappingConfiguration, role)) {
-			forbidden(channel, "Role '" + role + "' has read-only role-mapping.");
+			conflict(channel, "Role '" + role + "' has read-only role-mapping.");
```
vrozov, line 597: I don't think that `409` is the correct status code here. Note that  according to spec `Conflict` "is used in situations where the user might be able to resolve the conflict and resubmit the request" and `Forbidden` "might be forbidden for reasons unrelated to the credentials."

sujithvm: As per [wiki](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors) and resources 405 is recommended for PUT on a read-only resource so will change to that.

vrozov: Please check ES behavior. It uses 403 for read-only indices. It will be better to keep the current status code as both 403 and 405 are acceptable.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -606,10 +598,23 @@ protected boolean isValidRolesMapping(final RestChannel channel, final String ro
...
+			notFound(channel, "Resource '"+ name +"' is not available.");
+			return true;
+		}
+
+		if (isReadOnly(configuration, name)) {
+            methodNotAllowed(channel, "Resource '"+ name +"' is read-only.");
+			return true;
+		}
+
+		return false;
```
vrozov, line 618: should return `true` when the resource is writable (not hidden or read-only).

sujithvm: Will revert back from `isWriteable`. Did not want to unnecessarily negate the value during the check. It was initially `isResourceWriteProtected`

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
sujithvm: This would mean the opposite of the behavior

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
sujithvm: I see elasticsearch returning 403 for AuthorizationException but 405/409 seems to commonly used for read-only resources so its okay to change to 405

sujithvm: User has authorization to perform the request but the resource is read-only. Returning 405 would differentiate the behavior

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
vrozov: `isResourceWriteProtected` implies that resource is *not* writeable. It is better to avoid such implied negation.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
vrozov: I refer to ES behavior when `index.blocks.write` is set to `true`:
```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "cluster_block_exception",
        "reason" : "index [readonly] blocked by: [FORBIDDEN/8/index write (api)];"
      }
    ],
    "type" : "cluster_block_exception",
    "reason" : "index [readonly] blocked by: [FORBIDDEN/8/index write (api)];"
  },
  "status" : 403
}
```

sujithvm: Slightly torn on this. Why is it necessary to follow ES while most common practice is to return 405 ?

vrozov: AFAIK, there is no common practice and both 403 and 405 are used. Security plugin is part of ES deployment, so being consistent with ES is more important (and also avoids breaking REST API backward compatibility with prior security plugin versions).

sujithvm: Is returning 403 for readonly resource an ES practice only? Don't see it being commonly used

sujithvm: Returning 403 for a read only resource when user has correct authn & authz to perform the request feels wrong hence still preferring to return 405

vrozov: > Is returning 403 for readonly resource an ES practice only? Don't see it being commonly used

No, it is not.

> Returning 403 for a read only resource when user has correct authn & authz to perform the request feels wrong hence still preferring to return 405

Specification explicitly defines that 403 may be returned "for reasons unrelated to the credentials."

I don't see a need to change the behavior of the API from 403 to 405 unless there is a specific ask from a security plugin users.

sujithvm: Returning appropriate status code will differentiate behavior when user has correct authorization but modifying read only resource and there I see a need. 
I shall open up an issue for this and based on user discussion we can make/retain the change.

sujithvm: https://github.com/opendistro-for-elasticsearch/security/issues/719

sujithvm:
Will reopen based on https://github.com/opendistro-for-elasticsearch/security/issues/719

PR #755, opened 2020-10-01
Create release-drafter.yml

cliu123:
*Issue #, if available:*

*Description of changes:*
In order to avoid going through all the merged PRs and generate release notes manually at the time of release, release-drafter is introduced to keep track of the changes and generate release notes automatically as and when each PR gets merged into master branch

### DIFF
--- a/release-drafter.yml
+++ b/release-drafter.yml
@@ -1,8 +1,8 @@
 name: Release-drafter
 
 on:
-#   push:
-#     branches: [ master ]
+  push:
+    branches: [ master ]
   pull_request:
     branches: [ master ]
 
@@ -12,8 +12,5 @@ jobs:
     steps:
       # Drafts the next Release notes as Pull Requests are merged into "master"
       - uses: release-drafter/release-drafter@v5
-        # with:
-          # (Optional) specify config name to use, relative to .github/. Default: release-drafter.yml
-          # config-name: my-config.yml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/release-drafter.yml
+++ b/release-drafter.yml
@@ -1,0 +1,19 @@
+name: Release-drafter
+
+on:
+#   push:
+#     branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+   update_release_draft:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


vrozov:
Please provide an example of the test run.

sujithvm:
- Can we add template also to categorize into various categories Enhancements, Bug fixes, maintenance ? We already have labels for them 

- We create draft release as part of CD workflow https://github.com/opendistro-for-elasticsearch/security/blob/master/.github/workflows/cd.yml#L67 Is it possible to fetch release drafter output url and then upload artifacts to that draft ?

release-drafter.yml
```diff
@@ -0,0 +1,19 @@
+name: Release-drafter
+
+on:
+#   push:
+#     branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
sujithvm, line 7: Should this be push on master branch which would run after a pull request is squashed and merged?

release-drafter.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+   update_release_draft:
+    runs-on: ubuntu-latest
+    steps:
+      # Drafts the next Release notes as Pull Requests are merged into "master"
+      - uses: release-drafter/release-drafter@v5
+        # with:
+          # (Optional) specify config name to use, relative to .github/. Default: release-drafter.yml
+          # config-name: my-config.yml
```
sujithvm, line 17: Can we remove these comments ?

release-drafter.yml
cliu123: Resolved.

release-drafter.yml
cliu123: Resolved.

release-drafter.yml
sujithvm: Are both the trigger needed ?

cliu123: I believe so. When a PR gets checked in on master branch, the change should also be reflected in release notes.

sujithvm: Can you validate this behavior? AFAIK pull_request will trigger when pull request is created against the master branch. We want the release notes to be updated only after the pull request is merged to master (which results in a push on master branch). 

Check release drafter: https://github.com/release-drafter/release-drafter#usage

PR #756, opened 2020-10-02
Redact BCrypt security config internal hashes from audit logs

sujithvm:
*Issue #, if available:*

*Description of changes:*
- BCrypt hashes of internal users may be considered sensitive
- If auditing security index and doc is internalusers then redact all BCrypt hashes with `__HASH__` 
- Added separate methods for adding request body for security index configs as check is already done and not penalizing non security config logs 
- Changed type of convertStructuredMapToBytes to `Map<String, ?>` as used by Elasticsearch
- Updated tests 


hsiang9431-amzn (approved):
Add context:
This issue is seen if the following settings is disabled:
`Security/Audit logs/Compliance settings/Read metadata`
For reproducing the bug, default configuration has to be changed

Bug fix confirmed in this PR, approving it

debjanibnrj (approved):

Issue #739, opened 2020-09-29
fix build package system to remove "advanced"


hardik-k-shah:
mvn clean package -Padvanced 

"advanced" argument is no longer needed post merging of repositories.

sujithvm:
- Merge advanced profile executions with maven assembly goals https://github.com/opendistro-for-elasticsearch/security/blob/master/pom.xml#L832 
- Remove checksum-maven-plugin 
- Remove -Padvanced from README.md, ci.yml, cd.yml (workflows) and any other reference to it

PR #757, opened 2020-10-02
Remove checksum-maven-plugin && Remove -Padvanced

cliu123:
*Issue #, if available:*
#739 
*Description of changes:*
1. Remove checksum-maven-plugin from pom.xml
2. Remove -Padvanced from ci.yml, cd.yml and ReadMe.md.
3. Remove the advanced profile and move it’s executions into the regular build package goal for the maven assembly plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -868,36 +868,6 @@
                             </execution>
                         </executions>
                     </plugin>
-                    <plugin>
-                        <groupId>com.floragunn</groupId>
-                        <artifactId>checksum-maven-plugin</artifactId>
-                        <version>1.7.1</version>
-                        <executions>
-                            <execution>
-                                <goals>
-                                    <goal>files</goal>
-                                </goals>
-                                <phase>package</phase>
-                            </execution>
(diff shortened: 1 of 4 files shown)


vrozov:
Please remove `advanced` profile from pom.xml. Note that it is necessary to generate plugin .zip file.

pom.xml
```diff
@@ -829,114 +829,52 @@
     </build>
     <profiles>
         <profile>
-            <id>advanced</id>
```
sujithvm, line 832: We cannot just delete the advanced profile. We need to move the executions of maven assembly plugin

pom.xml
```diff
@@ -829,114 +829,52 @@
...
+            <id>veracode</id>
```
sujithvm, line 832: Out of scope for this PR but `veracode` profile can be removed too

pom.xml
vrozov: There should be also a way to suppress creation of .zip file (just generate .jar).

PR #760, opened 2020-10-09
Fix build package system to remove advanced

cliu123:
*Issue #, if available:*
#739 
*Description of changes:*
1. Remove checksum-maven-plugin from pom.xml
2. Remove -Padvanced from ci.yml, cd.yml and ReadMe.md.
3. Remove the advanced profile and move it’s executions into the regular build package goal for the maven assembly plugin.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #762, opened 2020-10-10
Fix build package system to remove advanced

cliu123:
*Issue #, if available:*
#739

*Description of changes:*
1. Remove checksum-maven-plugin from pom.xml
2. Remove -Padvanced from ci.yml, cd.yml and ReadMe.md.
3. Remove the advanced profile and move it’s executions into the regular build package goal for the maven assembly plugin.


sujithvm:
You can reuse pull requests (#757, #760) so that review comments are not lost. Following up from #757 can you check if `.zip` artifacts are being deployed in release?

pom.xml
```diff
@@ -525,6 +525,42 @@
 
     <build>
         <plugins>
+            <plugin>
+                <artifactId>maven-assembly-plugin</artifactId>
```
sujithvm, line 529: Can you test disabling `attach` configuration in `maven-assembly-plugin` to see if the resulting zips are ignored in maven release?

cliu123:
> You can reuse pull requests (#757, #760) so that review comments are not lost. Following up from #757 can you check if `.zip` artifacts are being deployed in release?

Those PRs were directly created for ODFE repo. To be compliant with the best practice better, I re-did the code change in my own fork and created this PR.

cliu123:
> You can reuse pull requests (#757, #760) so that review comments are not lost. Following up from #757 can you check if `.zip` artifacts are being deployed in release?

Yes, the opendistro_security-<version_id>.zip and opendistro_security-<version_id>-securityadmin-standalone.zip are deployed in /security/taregt/release.

vrozov:
"mvn package" should behave the same as before and not create .zip file. There is no need to create .zip file every time .jar is created. .zip file is only required for initial plugin installation or during ci/cd builds.

pom.xml
```diff
@@ -525,6 +525,44 @@
...
+                        </configuration>
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                    </execution>
+                    <execution>
+                        <id>securityadmin</id>
+                        <phase>package</phase>
+                        <configuration>
+                            <attach>false</attach>
```
cliu123, line 552: @sujithvm Here is to disable attach.

sujithvm:
> "mvn package" should behave the same as before and not create .zip file. There is no need to create .zip file every time .jar is created. .zip file is only required for initial plugin installation or during ci/cd builds.

IMO it is okay to update the default package goal. The default packaging expected by Elasticsearch is zip (jars + plugin descriptor + plugin security policy + securityconfig files + ...) and not just the core jar alone. This is done by other opendistro plugins as well. While pushing to maven central, the assembly is not attached hence only the jar files will be uploaded.

pom.xml
```diff
@@ -525,6 +525,44 @@
...
         <plugins>
+            <plugin>
+                <artifactId>maven-assembly-plugin</artifactId>
+                <version>3.1.1</version>
+                <executions>
+                    <execution>
+                        <id>plugin</id>
+                        <phase>package</phase>
+                        <configuration>
+                            <attach>false</attach>
```
cliu123, line 536: @sujithvm Here is to disable attach.

vrozov:
> > "mvn package" should behave the same as before and not create .zip file. There is no need to create .zip file every time .jar is created. .zip file is only required for initial plugin installation or during ci/cd builds.
> 
> IMO it is okay to update the default package goal. The default packaging expected by Elasticsearch is zip (jars + plugin descriptor + plugin security policy + securityconfig files + ...) and not just the core jar alone. This is done by other opendistro plugins as well. While pushing to maven central, the assembly is not attached hence only the jar files will be uploaded.

By default `mvn package` should not produce .zip file. It is intended for everyday developer usage and it is not necessary to build .zip file in such cases. There must be an opt-in flag that CI/CD and developers can use to produce plugin .zip file similar to how advanced profile was used before. I would even consider moving .zip file creation to the gradle build. Note that in a longer run, we should use only one build system (either maven or gradle).

sujithvm:
> > > "mvn package" should behave the same as before and not create .zip file. There is no need to create .zip file every time .jar is created. .zip file is only required for initial plugin installation or during ci/cd builds.
> > 
> > 
> > IMO it is okay to update the default package goal. The default packaging expected by Elasticsearch is zip (jars + plugin descriptor + plugin security policy + securityconfig files + ...) and not just the core jar alone. This is done by other opendistro plugins as well. While pushing to maven central, the assembly is not attached hence only the jar files will be uploaded.
> 

> By default `mvn package` should not produce .zip file.  

Could you point me to where this is documented. Configuration in maven assembly plugin is exactly provided to update such build configurations. Default "package"-ing ES plugin is zip and makes sense to update `package` goal.  

> It is intended for everyday developer usage and it is not necessary to build .zip file in such cases. 

Well, my everyday developer usage has been to create .zip file and do `elasticsearch-plugin install <zip file>`. 
How do developers use jar directly without any of the plugin descriptor, plugin security policy, config files, tools  ... 

> I would even consider moving .zip file creation to the gradle build.

Doing `./gradlew build` is similar to `mvn package`. Are you saying we need build flavors in gradle also instead of default gradle build?

vrozov:
It is documented in maven goal. Currently I can build just .jar file by using package goal for the default profile. We don't need two profiles, but we need ability to build only .jar file or build both .jar and .zip. For every day development, I don't need .zip to be built every time I use maven to build .jar. It is not necessary to go through full cycle of uninstalling and installing plugin, so it is sufficient to build .jar and replace single .jar in in the plugin installation location.

gradle builds .rpm and .deb packages, it is not the same as the `package` goal in maven that usually produces just .jar file. Note that separation of .rpm and .deb from .zip is artificial, it is just that gradle has better support for creating .rpm and .deb packages, but it can easily package .zip as well.

cliu123:
> "mvn package" should behave the same as before and not create .zip file. There is no need to create .zip file every time .jar is created. .zip file is only required for initial plugin installation or during ci/cd builds.

The scope of this PR is to remove the advanced profile. In order to aviod keep this PR stale, I can raise another issue to implement a flat that allow developers to choose to create .zip+jar or jar only. Can we handle it this way?

cliu123:
> "mvn package" should behave the same as before and not create .zip file. There is no need to create .zip file every time .jar is created. .zip file is only required for initial plugin installation or during ci/cd builds.

@vrozov The scope of this PR is to remove the advanced profile. In order to aviod keep this PR stale, I can raise another issue to implement a flag that allows developers to choose to create .zip+jar or jar only. Can we handle it this way?

vrozov:
The scope of the PR is to remove advanced profile and keep existing functionality. Currently, there is an option to build jar file only by using default profile.

cliu123:
> The scope of the PR is to remove advanced profile and keep existing functionality. Currently, there is an option to build jar file only by using default profile.
I don't think that I got the purpose/context behind this change then. I was told that we were trying to remove advanced profile and remains the same build configuration as if advanced profile is activated. However, it sounds like we still want the flexibility to choose to create jar + zip files or to creat only jar. Given that advanced profile has already provided developers with the flexibility and it's been working perfectly for us. Why would we remove it then? Could you please help me understand why we need to remove advanced profile?

cliu123:
> The scope of the PR is to remove advanced profile and keep existing functionality. Currently, there is an option to build jar file only by using default profile.

I don't think that I got the purpose/context behind this change then. I was told that we were trying to remove advanced profile and remains the same build configuration as if advanced profile is activated. However, it sounds like we still want the flexibility to choose to create jar + zip files or to creat only jar. Given that advanced profile has already provided developers with the flexibility and it's been working perfectly for us. Why would we remove it then? Could you please help me understand why we need to remove advanced profile?

vrozov:
advanced profile was used for historical reasons to allow extra dependencies. This is no longer needed. It is still necessary to have an option to build only .jar file to reduce the build time that doubles when .zip file is created.

sujithvm:
@vrozov I don't think this PR breaks the functionality of building jar only. Can do the following
- **Build zip + jar :** `mvn clean package -DskipTests`
- **Build jar only :** `mvn clean package -DskipTests -Dassembly.skipAssembly`

vrozov:
Please make `-Dassembly.skipAssembly` default.

sujithvm:
I still feel we should update the default goal to remain consistent with standard packaging practices for Elasticsearch plugins (as followed by other Opendistro plugins as well). Users can opt-out (instead of opt-in) for zip. Even if you see upstream Elasticsearch, just the assemble task would build for all packaging distributions. For initial development you need to build zip and most users building project here may not be active developers. This PR does not break the ability to build jar alone (also no change in build time) for whatever developer use case of replacing jar which I would not recommend as Elasticsearch provides "elasticsearch-plugin" tool to install/remove. 

@cliu123 what are your thoughts?

vrozov:
I would strongly prefer to keep the existing behavior where it is necessary to opt-in for advanced profile and creating .zip file. 

Not active developers on the project are likely to use an existing artifact, active developers (the primary consumers of maven build) are more likely to build .jar file only.

cliu123:
> I still feel we should update the default goal to remain consistent with standard packaging practices for Elasticsearch plugins (as followed by other Opendistro plugins as well). Users can opt-out (instead of opt-in) for zip. Even if you see upstream Elasticsearch, just the assemble task would build for all packaging distributions. For initial development you need to build zip and most users building project here may not be active developers. This PR does not break the ability to build jar alone (also no change in build time) for whatever developer use case of replacing jar which I would not recommend as Elasticsearch provides "elasticsearch-plugin" tool to install/remove.
> 
> @cliu123 what are your thoughts?

I actually agree with this. Now that developers have the options to opt-out for zip just by typing "-Dassembly.skipAssembly" in the command, we've already provide the options. And this behavior is consistent with what other ODFE plugins do and even upstream Elasticsearch.

cliu123:
> I would strongly prefer to keep the existing behavior where it is necessary to opt-in for advanced profile and creating .zip file.
> 
> Not active developers on the project are likely to use an existing artifact, active developers (the primary consumers of maven build) are more likely to build .jar file only.

I don't believe there's a way to set the default value of "-Dassembly.skipAssembly" to true unless creating a profile. I can note the usage of "-Dassembly.skipAssembly" in Readme file.

vrozov:
There is no need to introduce profile to set `assembly.skipAssembly` to `true`: add `assembly.skipAssembly` to properties with the value set to `true`. When it is necessary to build .zip, invoke maven with `-Dassembly.skipAssembly=false`.

README.md
```diff
@@ -80,7 +80,7 @@ mvn clean test
 * Build artifacts (zip, deb, rpm)
 
 ```
-mvn clean package -Padvanced -DskipTests
+mvn clean package -DskipTests
```
vrozov, line 83: `-Dassembly.skipAssembly=false`?

cliu123: Done.

cliu123:
Closing this PR since the decision has been changed.

PR #765, opened 2020-10-13
When replacing .kibana index with multi-tenant index, create index with alias if one already does not exist

vrozov:
*Description of changes:* Create explicit `CreateIndexRequest` if index or alias does not exist instead of relying on auto index create ES functionality.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
@@ -16,6 +16,7 @@
 package com.amazon.opendistroforelasticsearch.security.configuration;
 
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -23,6 +24,7 @@ import org.elasticsearch.ElasticsearchException;
 import org.elasticsearch.action.ActionRequest;
 import org.elasticsearch.action.DocWriteRequest;
 import org.elasticsearch.action.IndicesRequest.Replaceable;
+import org.elasticsearch.action.admin.indices.alias.Alias;
 import org.elasticsearch.action.admin.indices.create.CreateIndexRequest;
 import org.elasticsearch.action.admin.indices.mapping.get.GetFieldMappingsIndexRequest;
 import org.elasticsearch.action.admin.indices.mapping.get.GetFieldMappingsRequest;
@@ -40,6 +42,8 @@ import org.elasticsearch.action.termvectors.MultiTermVectorsRequest;
 import org.elasticsearch.action.termvectors.TermVectorsRequest;
 import org.elasticsearch.action.update.UpdateRequest;
 import org.elasticsearch.client.Client;
+import org.elasticsearch.cluster.metadata.IndexAbstraction;
+import org.elasticsearch.cluster.metadata.IndexMetadata;
 import org.elasticsearch.cluster.metadata.IndexNameExpressionResolver;
 import org.elasticsearch.cluster.service.ClusterService;
 import org.elasticsearch.threadpool.ThreadPool;
@@ -49,10 +53,15 @@ import com.amazon.opendistroforelasticsearch.security.resolver.IndexResolverRepl
 import com.amazon.opendistroforelasticsearch.security.securityconf.DynamicConfigModel;
 import com.amazon.opendistroforelasticsearch.security.user.User;
 
+import com.google.common.collect.ImmutableMap;
+
 public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
 
     private static final String USER_TENANT = "__user__";
     private static final String EMPTY_STRING = "";
+    private static final Map<String, Object> KIBANA_INDEX_SETTINGS = ImmutableMap.of(
+            IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1
+    );
 
     protected final Logger log = LogManager.getLogger(this.getClass());
 
@@ -88,13 +97,13 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
      *
      */
     @Override
-    public Boolean replaceKibanaIndex(final ActionRequest request, final String action, final User user, final DynamicConfigModel config,
+    public ReplaceResult replaceKibanaIndex(final ActionRequest request, final String action, final User user, final DynamicConfigModel config,
                                       final Resolved requestedResolved, final Map<String, Boolean> tenants) {
 
         final boolean enabled = config.isKibanaMultitenancyEnabled();//config.dynamic.kibana.multitenancy_enabled;
 
         if (!enabled) {
-            return null;
+            return CONTINUE_EVALUATION_REPLACE_RESULT;
         }
 
         //next two lines needs to be retrieved from configuration
@@ -116,10 +125,10 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
             }
 
             if (kibanaIndexOnly && !isTenantAllowed(request, action, user, tenants, "global_tenant")) {
-                return Boolean.TRUE;
+                return ACCESS_DENIED_REPLACE_RESULT;
             }
 
-            return null;
+            return CONTINUE_EVALUATION_REPLACE_RESULT;
         }
 
         if (USER_TENANT.equals(requestedTenant)) {
@@ -132,13 +141,13 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
         }
 
         //request not made by the kibana server and user index is the only index/alias involved
-        if (!user.getName().equals(kibanaserverUsername) && requestedResolved.getAllIndices().size() == 1
-                && requestedResolved.getAllIndices().contains(toUserIndexName(kibanaIndexName, requestedTenant))) {
-
-            if (isTenantAllowed(request, action, user, tenants, requestedTenant)) {
-                return Boolean.FALSE;
+        if (!user.getName().equals(kibanaserverUsername)) {
+            final Set<String> indices = requestedResolved.getAllIndices();
+            final String tenantIndexName = toUserIndexName(kibanaIndexName, requestedTenant);
+            if (indices.size() == 1 && indices.iterator().next().startsWith(tenantIndexName) &&
+                    isTenantAllowed(request, action, user, tenants, requestedTenant)) {
+                    return ACCESS_GRANTED_REPLACE_RESULT;
             }
-
         }
 
         //intercept when requests are not made by the kibana server and if the kibana index/alias (.kibana) is the only index/alias involved
@@ -150,15 +159,15 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
             }
 
             if (!isTenantAllowed(request, action, user, tenants, requestedTenant)) {
-                return Boolean.TRUE;
+                return ACCESS_DENIED_REPLACE_RESULT;
             }
 
             // TODO handle user tenant in that way that this tenant cannot be specified as
             // regular tenant
             // to avoid security issue
 
-            replaceIndex(request, kibanaIndexName, toUserIndexName(kibanaIndexName, requestedTenant), action);
-            return Boolean.FALSE;
+            final String tenantIndexName = toUserIndexName(kibanaIndexName, requestedTenant);
+            return newAccessGrantedReplaceResult(replaceIndex(request, kibanaIndexName, tenantIndexName, action));
 
         } else if (!user.getName().equals(kibanaserverUsername)) {
 
@@ -170,18 +179,30 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
 
         }
 
+        return CONTINUE_EVALUATION_REPLACE_RESULT;
+    }
+
+    private CreateIndexRequest checkAndCreateRequestIfAbsent(final String name) {
+        final IndexAbstraction indexAbstraction = clusterService.state().getMetadata().getIndicesLookup().get(name);
+        if (indexAbstraction == null) {
+            return new CreateIndexRequest(name + "_1")
+                    .settings(KIBANA_INDEX_SETTINGS)
+                    .alias(new Alias(name));
+        }
+        log.debug("{} {} already exists", indexAbstraction.getType(), indexAbstraction.getName());
         return null;
     }
 
-    private void replaceIndex(final ActionRequest request, final String oldIndexName, final String newIndexName, final String action) {
+    private CreateIndexRequest replaceIndex(final ActionRequest request, final String oldIndexName, final String newIndexName, final String action) {
         boolean kibOk = false;
+        CreateIndexRequest createIndexRequest = null;
 
         if (log.isDebugEnabled()) {
             log.debug("{} index will be replaced with {} in this {} request", oldIndexName, newIndexName, request.getClass().getName());
         }
 
         if (request instanceof GetFieldMappingsIndexRequest || request instanceof GetFieldMappingsRequest) {
-            return;
+            return createIndexRequest;
         }
 
         //handle msearch and mget
@@ -192,7 +213,7 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
 
         // CreateIndexRequest
         if (request instanceof CreateIndexRequest) {
-            ((CreateIndexRequest) request).index(newIndexName);
+            ((CreateIndexRequest) request).index(newIndexName + "_1").alias(new Alias(newIndexName));
             kibOk = true;
         } else if (request instanceof BulkRequest) {
 
@@ -203,6 +224,7 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
                 }
 
                 if (ar instanceof IndexRequest) {
+                    createIndexRequest = checkAndCreateRequestIfAbsent(newIndexName);
                     ((IndexRequest) ar).index(newIndexName);
                 }
 
@@ -240,6 +262,7 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
             ((UpdateRequest) request).index(newIndexName);
             kibOk = true;
         } else if (request instanceof IndexRequest) {
+            createIndexRequest = checkAndCreateRequestIfAbsent(newIndexName);
             ((IndexRequest) request).index(newIndexName);
             kibOk = true;
         } else if (request instanceof DeleteRequest) {
@@ -265,6 +288,7 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
         if (!kibOk) {
             log.warn("Dont know what to do (2) with {}", request.getClass());
         }
+        return createIndexRequest;
     }
 
     private String toUserIndexName(final String originalKibanaIndex, final String tenant) {
@@ -276,8 +300,12 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
         return originalKibanaIndex + "_" + tenant.hashCode() + "_" + tenant.toLowerCase().replaceAll("[^a-z0-9]+", EMPTY_STRING);
     }
 
-    private boolean resolveToKibanaIndexOrAlias(final Resolved requestedResolved, final String kibanaIndexName) {
-        return (requestedResolved.getAllIndices().size() == 1 && requestedResolved.getAllIndices().contains(kibanaIndexName))
-                || (requestedResolved.getAliases().size() == 1 && requestedResolved.getAliases().contains(kibanaIndexName));
+    private static boolean resolveToKibanaIndexOrAlias(final Resolved requestedResolved, final String kibanaIndexName) {
+        final Set<String> allIndices = requestedResolved.getAllIndices();
+        if (allIndices.size() == 1 && allIndices.iterator().next().equals(kibanaIndexName)) {
+            return true;
+        }
+        final Set<String> aliases = requestedResolved.getAliases();
+        return (aliases.size() == 1 && aliases.iterator().next().equals(kibanaIndexName));
     }
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
@@ -61,7 +61,7 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
 
     private static final String USER_TENANT = "__user__";
     private static final String EMPTY_STRING = "";
-    private static final String SUFFIX = "_1";
+    private static final String KIBANA_INDEX_SUFFIX = "_1";
     private static final Map<String, Object> KIBANA_INDEX_SETTINGS = ImmutableMap.of(
             IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1,
             IndexMetadata.SETTING_AUTO_EXPAND_REPLICAS, "0-1"
@@ -188,7 +188,7 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
 
     private CreateIndexRequest newCreateIndexRequestIfAbsent(final String name) {
         final Map<String, IndexAbstraction> indicesLookup = clusterService.state().getMetadata().getIndicesLookup();
-        final String concreteName = name.concat(SUFFIX);
+        final String concreteName = name.concat(KIBANA_INDEX_SUFFIX);
         if (Arrays.stream(new String[]{name, concreteName})
                 .map(s -> indicesLookup.get(s))
                 .filter(Objects::nonNull)
@@ -220,7 +220,7 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
         // CreateIndexRequest
         if (request instanceof CreateIndexRequest) {
             // use new name for alias and suffixed index name
-            ((CreateIndexRequest) request).index(newIndexName.concat(SUFFIX)).alias(new Alias(newIndexName));
+            ((CreateIndexRequest) request).index(newIndexName.concat(KIBANA_INDEX_SUFFIX)).alias(new Alias(newIndexName));
             kibOk = true;
         } else if (request instanceof BulkRequest) {
 
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
@@ -15,7 +15,9 @@
 
 package com.amazon.opendistroforelasticsearch.security.configuration;
 
+import java.util.Arrays;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import org.apache.logging.log4j.LogManager;
@@ -59,8 +61,10 @@ public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
 
     private static final String USER_TENANT = "__user__";
     private static final String EMPTY_STRING = "";
+    private static final String SUFFIX = "_1";
     private static final Map<String, Object> KIBANA_INDEX_SETTINGS = ImmutableMap.of(
-            IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1
+            IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1,
+            IndexMetadata.SETTING_AUTO_EXPAND_REPLICAS, "0-1"
     );
 
     protected final Logger log = LogManager.getLogger(this.getClass());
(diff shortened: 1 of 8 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -192,7 +230,7 @@ private void replaceIndex(final ActionRequest request, final String oldIndexName
 
         // CreateIndexRequest
         if (request instanceof CreateIndexRequest) {
-            ((CreateIndexRequest) request).index(newIndexName);
+            ((CreateIndexRequest) request).index(newIndexName + "_0").alias(new Alias(newIndexName));
```
cliu123, line 233: Just out of curiosity, does CreateIndexReqeust come to here at all? I tried doing same thing, but it didn't work.

vrozov: `CreateIndexRequest ` may be submitted using REST API, it won't be submitted as part of regular Kibana workflow. The change is required for the consistent behavior in index re-write.

cliu123: That makes sense. Thanks for explanatino!

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -203,6 +241,7 @@ private void replaceIndex(final ActionRequest request, final String oldIndexName
                 }
 
                 if (ar instanceof IndexRequest) {
+                    createIndexAndAliasIfAbsent(newIndexName);
```
cliu123, line 244: This looks interesting. Before this change, does the logic not even check whether the index exists before starting indexing? Looks like a bug to me if there's really no this check. Have we faced any problem because of this at all?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -49,10 +54,16 @@
...
 
+import com.google.common.collect.ImmutableMap;
+
 public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
 
     private static final String USER_TENANT = "__user__";
     private static final String EMPTY_STRING = "";
+    private static final Map<String, Object> KIBANA_INDEX_SETTINGS = ImmutableMap.of(
+            IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1,
+            IndexMetadata.SETTING_AUTO_EXPAND_REPLICAS, "0-all"
```
sujithvm, line 65: Is there need to set this as auto expand? This index is created for every user and may flood with replicas based on the number of nodes and users in the system.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -173,6 +184,33 @@ public Boolean replaceKibanaIndex(final ActionRequest request, final String acti
         return null;
     }
 
+    private boolean createIndexAndAliasIfAbsent(final String name) {
+        final IndexAbstraction indexAbstraction = clusterService.state().getMetadata().getIndicesLookup().get(name);
+        if (indexAbstraction != null) {
+            log.debug("{} {} already exists", indexAbstraction.getType(), indexAbstraction.getName());
+            return false;
+        }
+        final CreateIndexRequest createIndexRequest = new CreateIndexRequest(name + "_0")
```
sujithvm, line 193: Why `_0` suffix?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -173,6 +184,33 @@ public Boolean replaceKibanaIndex(final ActionRequest request, final String acti
...
+            if (acknowledged) {
+                log.info("Created new index {} with aliases {}", createIndexRequest.index(), createIndexRequest.aliases());
+                return true;
+            }
+        } catch (ResourceAlreadyExistsException resourceAlreadyExistsException) {
+            log.warn("Index {} or aliases {} already exist", createIndexRequest.index(), createIndexRequest.aliases());
+            return false;
+        }
+        log.warn("Request to create new index {} with aliases {} was not acknowledged", createIndexRequest.index(), createIndexRequest.aliases());
+        throw new ElasticsearchException("Request to create new index {} with aliases {} was not acknowledged", createIndexRequest.index(), createIndexRequest.aliases());
```
sujithvm, line 211: nit: avoid duplication and factor out error message.

sujithvm (commented):
Could you update the PR description with fix and logic especially around alias? Trying to understand how creating an alias x with index x_0 is helping

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
vrozov: Likely not required, I copied settings from `.kibana` index.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
vrozov: ES implicitly (auto) creates index if it does not exist.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -283,7 +293,30 @@ public int order() {
                 if(!dlsFlsValve.invoke(request, listener, pres.getAllowedFlsFields(), pres.getMaskedFields(), pres.getQueries())) {
                     return;
                 }
-                chain.proceed(task, action, request, listener);
+                final CreateIndexRequest createIndexRequest = pres.getRequest();
```
palashhedau, line 296: Will this always be CreateIndexRequest when reached here? What about bulk request?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -170,18 +179,30 @@ public Boolean replaceKibanaIndex(final ActionRequest request, final String acti
...
+                    .alias(new Alias(name));
+        }
+        log.debug("{} {} already exists", indexAbstraction.getType(), indexAbstraction.getName());
         return null;
     }
 
-    private void replaceIndex(final ActionRequest request, final String oldIndexName, final String newIndexName, final String action) {
+    private CreateIndexRequest replaceIndex(final ActionRequest request, final String oldIndexName, final String newIndexName, final String action) {
         boolean kibOk = false;
+        CreateIndexRequest createIndexRequest = null;
```
palashhedau, line 198: All the request will be treated as CreateIndexRequest?

src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluatorResponse.java
```diff
@@ -41,6 +43,7 @@
     Map<String,Set<String>> maskedFields;
     Map<String,Set<String>> queries;
     PrivilegesEvaluatorResponseState state = PrivilegesEvaluatorResponseState.PENDING;
+    CreateIndexRequest request;
```
palashhedau, line 46: Who sets this?

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
vrozov: It will be always `CreateIndexRequest`

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
vrozov: `CreateIndexRequest` replaces auto create index request.

src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluatorResponse.java
vrozov: `checkAndCreateRequestIfAbsent()`

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
palashhedau: Just wondering as I saw couple of bulk request reached here. What happens to those? Not sure if you noticed the bulk request

vrozov: It is not the original request.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -203,6 +224,7 @@ private void replaceIndex(final ActionRequest request, final String oldIndexName
                 }
 
                 if (ar instanceof IndexRequest) {
+                    createIndexRequest = checkAndCreateRequestIfAbsent(newIndexName);
```
dinusX, line 227: Not sure how often there might be bulk requests, but do you think that it's worth to add some optimization here, and generate the object only once for the first Index Request (in case if there are N write requests to the same kibana+tenant index).

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -203,6 +224,7 @@ private void replaceIndex(final ActionRequest request, final String oldIndexName
...
                     ((IndexRequest) ar).index(newIndexName);
                 }
```
dinusX, line 230: Was not able to add comment to the below code (see below 3 lines).

Should we create the kibana+tenant index for UpdateRequest as well?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -170,18 +179,30 @@ public Boolean replaceKibanaIndex(final ActionRequest request, final String acti
 
         }
 
+        return CONTINUE_EVALUATION_REPLACE_RESULT;
+    }
+
+    private CreateIndexRequest checkAndCreateRequestIfAbsent(final String name) {
+        final IndexAbstraction indexAbstraction = clusterService.state().getMetadata().getIndicesLookup().get(name);
+        if (indexAbstraction == null) {
+            return new CreateIndexRequest(name + "_1")
```
sujithvm, line 188: Can any suffix like "security" be used? My understanding is that migrated index will the number suffix added.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -170,18 +179,30 @@ public Boolean replaceKibanaIndex(final ActionRequest request, final String acti
 
         }
 
+        return CONTINUE_EVALUATION_REPLACE_RESULT;
+    }
+
+    private CreateIndexRequest checkAndCreateRequestIfAbsent(final String name) {
```
sujithvm, line 185: Can we rename this to just "createIndexRequestIfAbsent" ? not clear from the name what the check is

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
sujithvm: Can the suffix logic be consolidated into a separate function to avoid missing any logic in the future. I see "_1" used in the subsequent create index logic too.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
sujithvm, line 227: Why are these requests done differently ? 

`createIndexRequest = checkAndCreateRequestIfAbsent(newIndexName);` 
and
`((CreateIndexRequest) request).index(newIndexName + "_1").alias(new Alias(newIndexName));`

src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -338,19 +337,19 @@ public PrivilegesEvaluatorResponse evaluate(final User user, String action0, fin
...
-
-            if (replaceResult == Boolean.FALSE) {
-                presponse.allowed = true;
+            if (!replaceResult.continueEvaluation) {
+                if (replaceResult.accessDenied) {
+                    auditLog.logMissingPrivileges(action0, request, task);
+                } else {
+                    presponse.allowed = true;
+                    presponse.request = replaceResult.createIndexRequest;
+                }
```
sujithvm, line 352: Is it possible to factor out to separate function for replacing kibana index to avoid duplication ?

sujithvm:
Can we also add tests around this ?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -49,10 +53,15 @@
...
 import com.amazon.opendistroforelasticsearch.security.user.User;
 
+import com.google.common.collect.ImmutableMap;
+
 public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
 
     private static final String USER_TENANT = "__user__";
     private static final String EMPTY_STRING = "";
+    private static final Map<String, Object> KIBANA_INDEX_SETTINGS = ImmutableMap.of(
+            IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1
```
nibix, line 63: Is this a normal setting for the Kibana index?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
nibix, line 185: This (in combination with the code executing the CreateIndexRequest inside OpenDistroSecurityFilter) might suffer from concurrency problems if several index requests are done in quick succession. Possibly one should give ResourceAlreadyExistsException a special treatment when handling errors.

Also, is it safe to have a CreateIndexRequest directly followed by an IndexRequest?

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
nibix, line 296: Style: This is somewhat a breach of separation of concerns. This might lead to confusion (as illustrated by the existing comments) and hard to understand logic. One should try to concentrate the logic in PrivilegesInterceptorImpl. I can think of two alternatives:

- Provide a generic callback instead of the CreateIndexRequest
- Dynamically extend the ActionFilterChain. This would however probably require extension of a couple of method interfaces.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -92,6 +98,8 @@
 
 public class OpenDistroSecurityFilter implements ActionFilter {
 
+    private static final String CREATE_INDEX_ERROR_MSG = "Request to create new index {} with aliases {} was not acknowledged";
```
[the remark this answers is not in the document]
nibix: I'd like to add one important detail here, also in : `isAcknowledged() == false` does not mean that the index creation failed. If it failed, we'll enter the `onFailed` callback below. `isAcknowledged() == false` just means that acknowledgement from the shards timed out. It is still possible that the index will be created. If I am not mistaken, the default timeout is 30 seconds. Thus, this probably only occurs in high-load situations.

Considering that we immediately after this call might want to write documents using the alias name, it seems to be correct to abort the operation. However, the error message should probably reflect this.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -283,7 +293,30 @@ public int order() {
...
-                chain.proceed(task, action, request, listener);
+                final CreateIndexRequest createIndexRequest = pres.getRequest();
+                if (createIndexRequest == null) {
+                    chain.proceed(task, action, request, listener);
+                } else {
+                    client.admin().indices().create(createIndexRequest, new ActionListener<CreateIndexResponse>() {
+                        @Override
+                        public void onResponse(CreateIndexResponse createIndexResponse) {
+                            log.debug("CreateIndexRequest {}, CreateIndexResponse {}", createIndexRequest, createIndexResponse);
+                            if (createIndexResponse.isAcknowledged()) {
```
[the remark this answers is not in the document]
vrozov: Assuming that such requests are originated from user interaction with the Kibana UI, I don't expect multiple concurrent requests from the same user. In a rear case when it happens, retrying operation on Kibana UI will allow user to proceed.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
vrozov: My understanding of UpdateRequest is that it tries to update an existing document. It means that both document and index should already exist.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
vrozov: The suffix is consistent with the suffix used by Kibana. Kibana creates `.kibana_1` index with `.kibana` alias. I'd prefer to follow the same convention.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
vrozov: One creates new `CreateIndexRequest`, another modifies an existing `CreateIndexRequest`. For `CreateIndexRequest` we don't need to create another request.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
vrozov: In the current implementation `OpenDistroSecurityFilter` is responsible for executing the `chain` and pre-processing (submitting)`CreateIndexRequest` while `PrivilegesInterceptorImpl` is responsible for creating `CreateIndexRequest` instance if one is necessary.

I don't think that introducing callback will add more clarity, unless I miss something in your proposal. The same for `ActionFilterChain`, it is possible to return new instance of `ActionFilterChain` instead of `CreateIndexRequest` that will either call `client.admin().indices().create()` or `chain.proceed()` directly. This will require passing `chain` all the way to `PrivilegesInterceptorImpl` and propagating new `ActionFilterChain` back to `OpenDistroSecurityFilter`.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
vrozov: Yes, this is the default setting for `.kibana_1` index.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
vrozov: The code on line 227 and 265 are the same:
```
createIndexRequest = checkAndCreateRequestIfAbsent(newIndexName);
```

sujithvm: I meant to ask about 216 but I see its a CreateIndexRequest

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
sujithvm: I was wondering if it safe to interfere with what Kibana does. My understanding is that we just shouldn't have an index terminating with a long number. As long as that is met, Kibana index migration would work correctly.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -132,13 +141,13 @@ public Boolean replaceKibanaIndex(final ActionRequest request, final String acti
...
         //request not made by the kibana server and user index is the only index/alias involved
-        if (!user.getName().equals(kibanaserverUsername) && requestedResolved.getAllIndices().size() == 1
-                && requestedResolved.getAllIndices().contains(toUserIndexName(kibanaIndexName, requestedTenant))) {
-
-            if (isTenantAllowed(request, action, user, tenants, requestedTenant)) {
-                return Boolean.FALSE;
+        if (!user.getName().equals(kibanaserverUsername)) {
+            final Set<String> indices = requestedResolved.getAllIndices();
+            final String tenantIndexName = toUserIndexName(kibanaIndexName, requestedTenant);
+            if (indices.size() == 1 && indices.iterator().next().startsWith(tenantIndexName) &&
```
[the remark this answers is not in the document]
vrozov: Discussed with @dinusX offline. Option to use alias is not available and will require modification beyond the scope of the current PR. Using two equals will lead to inconsistent behavior after upgrade and kibana tenant index migration. Using `startsWith()` indeed will allow user to create `.kibana_<hash>_<user>*` indices similar how user could read/write/delete documents in `.kibana_<hash>_user` index without permissions being explicitly granted.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
vrozov: We do not interfere with Kibana, we simply follow the same convention as Kibana.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
nibix: So far, `OpenDistroSecurityFilter` was a very generic module and not concerned with multi-tenancy at all. All multi-tenancy logic was concentrated in `PrivilegesInterceptorImpl`. This separation would be no longer valid any more.

A callback would bring more clarity by still being able to keep the logic in `PrivilegesInterceptorImpl`. 

As I wrote, this is just a style issue. I can understand that callbacks would also complicate the code. When keeping the code in `OpenDistroSecurityFilter`, however, I'd recommend to clearly separate this code from the generic request handling and mark this as multi-tenancy handling.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
vrozov: I do not expect multiple concurrent requests that interact with the same multi-tenant index as such requests are originated from Kibana UI and it should mean that the same user is used by multiple sessions. I am not even sure that Kibana supports such use case. Anyway, I added handling for `ResourceAlreadyExistsException` as a safety net.

The existing implementation relies on ES to auto create index on `IndexRequest` and the difference with the new approach is that it is security plugin that submits CreateIndexRequest instead of ES submitting similar request as part of auto create.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
vrozov: All logic related to multi-tenancy and whether or not it is necessary to create multi-tenancy index and details of such request is encapsulated in the `PrivilegesInterceptorImpl`. `OpenDistroSecurityFilter` is only responsible for submitting `CreateIndexRequest` and has no knowledge about multi-tenancy.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -49,10 +55,17 @@
...
+import com.google.common.collect.ImmutableMap;
+
 public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
 
     private static final String USER_TENANT = "__user__";
     private static final String EMPTY_STRING = "";
+    private static final String SUFFIX = "_1";
+    private static final Map<String, Object> KIBANA_INDEX_SETTINGS = ImmutableMap.of(
+            IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1,
+            IndexMetadata.SETTING_AUTO_EXPAND_REPLICAS, "0-1"
```
dinusX, line 67: Does the original ".kibana_1" index have a similar setting ? If not, should we use the same settings as original kibana index ?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -49,10 +55,17 @@
 import com.amazon.opendistroforelasticsearch.security.securityconf.DynamicConfigModel;
 import com.amazon.opendistroforelasticsearch.security.user.User;
 
+import com.google.common.collect.ImmutableMap;
+
 public class PrivilegesInterceptorImpl extends PrivilegesInterceptor {
 
     private static final String USER_TENANT = "__user__";
     private static final String EMPTY_STRING = "";
+    private static final String SUFFIX = "_1";
```
dinusX, line 64: Can we rename this to something more clear, like "KIBANA_INDEX_SUFFIX"

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -138,6 +149,10 @@ public int order() {
             apply0(task, action, request, listener, chain);
         }
     }
+
+    private static Set<String> alias2Name(Set<Alias> aliases) {
+        return aliases.stream().map(a -> a.name()).collect(ImmutableSet.toImmutableSet());
```
dinusX, line 154: Should we add a null check for "aliases" ?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/PrivilegesInterceptorImpl.java
vrozov: ".kibana_1" is created with the same settings, number of shards 1 and auto expand replicas "0-1".

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
vrozov: It assumes not `null` `aliases` that is always true for `CreateIndexResponse.aliases()`.

dinusX (approved):

src/test/java/com/amazon/opendistroforelasticsearch/security/multitenancy/test/MultitenancyTests.java
```diff
@@ -243,11 +249,12 @@ public void testMtMulti() throws Exception {
             Map indexSettings = new HashMap();
             indexSettings.put("number_of_shards", 1);
             indexSettings.put("number_of_replicas", 0);
-            tc.admin().indices().create(new CreateIndexRequest(".kibana_92668751_admin")
-                .settings(indexSettings))
+            tc.admin().indices().create(new CreateIndexRequest(kibanaIndex)
+                .settings(indexSettings)
+                .alias(new Alias(".kibana_92668751_admin")))
```
sujithvm, line 254: Can we add a test for kibana tenant index ending with a large number ?

sujithvm: Or is it taken care of by Kibana?

sujithvm (approved):

PR #767, opened 2020-10-13
Change CD workflow to use new staging bucket for artifacts

camerski:
*Description of changes:*
The infrastructure team is separating the production and staging locations into different AWS accounts. Plugins need to modify their workflows to publish to the new locations.

This PR changes the CD workflow to add a build number and write the zip, deb, and rpm plugin artifacts to staging.artifacts.opendistroforelasticsearch.amazon.com. The write to S3 currently fails (see https://github.com/camerski/security/actions/runs/305148456) because the secrets have not been updated; the secrets will be updated at the same time this PR is merged.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -50,11 +50,25 @@ jobs:
 
     - name: Upload Artifacts to S3
       run: |
-        s3_path=s3://artifacts.opendistroforelasticsearch.amazon.com/downloads
-        aws s3 cp artifacts/*.zip $s3_path/elasticsearch-plugins/opendistro-security/
-        aws s3 cp artifacts/*.deb $s3_path/debs/opendistro-security/
-        aws s3 cp artifacts/*.rpm $s3_path/rpms/opendistro-security/
-        aws cloudfront create-invalidation --distribution-id ${{ secrets.DISTRIBUTION_ID }} --paths '/downloads/*'
+        zip=`ls artifacts/*.zip`
+        rpm=`ls artifacts/*.rpm`
+        deb=`ls artifacts/*.deb`
+
+        # Inject the build number before the suffix
+        zip_outfile=`basename ${zip%.zip}-build-${GITHUB_RUN_NUMBER}.zip`
+        rpm_outfile=`basename ${rpm%.rpm}-build-${GITHUB_RUN_NUMBER}.rpm`
+        deb_outfile=`basename ${deb%.deb}-build-${GITHUB_RUN_NUMBER}.deb`
+
+        s3_prefix="s3://staging.artifacts.opendistroforelasticsearch.amazon.com/snapshot/elasticsearch-plugins/security/"
+
+        echo "Copying ${zip} to ${s3_prefix}${zip_outfile}"
+        aws s3 cp --quiet $zip ${s3_prefix}${zip_outfile}
+
+        echo "Copying ${rpm} to ${s3_prefix}${rpm_outfile}"
+        aws s3 cp --quiet $rpm ${s3_prefix}${rpm_outfile}
+
+        echo "Copying ${deb} to ${s3_prefix}${deb_outfile}"
+        aws s3 cp --quiet $deb ${s3_prefix}${deb_outfile}
 
     - name: Upload Artifacts to Maven Central
       env:


hardik-k-shah:
Does this mean that each plugin will only upload artifacts to staging S3 and infra team will take over post that ( publishing artifacts and distribution)?

camerski:
> Does this mean that each plugin will only upload artifacts to staging S3 and infra team will take over post that ( publishing artifacts and distribution)?

For the zip, rpm, and deb artifacts that are bundled with ODFE, yes. You will still publish to Github or Maven using your existing process.

.github/workflows/cd.yml
```diff
@@ -50,11 +50,25 @@ jobs:
...
-        aws s3 cp artifacts/*.zip $s3_path/elasticsearch-plugins/opendistro-security/
-        aws s3 cp artifacts/*.deb $s3_path/debs/opendistro-security/
-        aws s3 cp artifacts/*.rpm $s3_path/rpms/opendistro-security/
-        aws cloudfront create-invalidation --distribution-id ${{ secrets.DISTRIBUTION_ID }} --paths '/downloads/*'
+        zip=`ls artifacts/*.zip`
+        rpm=`ls artifacts/*.rpm`
+        deb=`ls artifacts/*.deb`
+
+        # Inject the build number before the suffix
+        zip_outfile=`basename ${zip%.zip}-build-${GITHUB_RUN_NUMBER}.zip`
```
vrozov, line 58: Will it be better to modify `pom.xml` and `build.gradle` to produce output files in the required format by passing `${GITHUB_RUN_NUMBER}`?

camerski: I would not modify those, because it would also change the names of the artifacts that go into Sonatype and the Github release. I think those artifacts do not want the build number.

hardik-k-shah (approved):

sujithvm:
@camerski Making this change would result in plugin being more dependent on the infra team for release and cannot make a standalone release when the plugin warrants one (including a patch release) not tied to ODFE distribution. Is this part of long term goal? 
What is the process to promote from staging?

camerski:
> @camerski Making this change would result in plugin being more dependent on the infra team for release and cannot make a standalone release when the plugin warrants one (including a patch release) not tied to ODFE distribution. Is this part of long term goal?
> What is the process to promote from staging?

Good question. I'll explain why we're making the change first.

Currently, it is possible for plugin artifacts to be overwritten during an ODFE release cycle, which means that we cannot guarantee that ODFE ships with exactly the artifacts that were tested. It also means we cannot reliably reproduce an ODFE build, because the artifacts may have been overwritten since the last build. The current staging location is publicly available and artifacts pushed to staging have not always been fully tested, so there is a risk that we have publicly-available artifacts with regressions.

Moving to a standalone staging location with build numbers allows ODFE to provide stronger guarantees about what we're releasing in the official ODFE distribution, and makes it possible for us to raise the quality bar on artifacts that are made available to the public.

Unfortunately, this does somewhat tie plugin releases to the ODFE distribution in the short term. For in-cycle releases, the infra team will promote from staging to production when the ODFE release is ready. For out-of-cycle releases the infra team can manually promote the new artifacts into production. Not ideal, I know, but we should be able to provide one-business-day turnaround, and eventually we may provide self-service tools for automating this.

You have read access to the staging bucket so it doesn't block you from also publishing the artifacts to your own distribution channel.

Long term, we plan to move to a model where each plugin owns their own staging / production locations and simply sends the infrastructure team a notification when a new plugin version is available. This is more like a Linux distribution model: When ODFE does a release the infra team just picks the plugin versions they want to include, and the plugin authors are free to use their own versioning system, publication workflows, and artifact locations.

Feel free to set up a call with me if you want to discuss in person.

.github/workflows/cd.yml
vrozov: Instead of github build number, can we use commit sha? This will allow us to build the same artifact if necessary.

camerski: That's an option, but it makes it harder for us to know what the latest build is. The GITHUB_RUN_NUMBER increases monotonically, which makes figuring out what to promote to production easier. If you feel strongly then I'm happy to change it - the release process doesn't rely on human-readable names (or, at least, won't rely on those names soon).

vrozov: This will give us an option to trigger CD workflow without impacting infra build for example when we tag release candidate commit with the final release tag.

camerski: The GITHUB_RUN_NUMBER also ensures that you will not impact the infra build.

camerski: In the interests of getting this unblocked, would you be OK with shipping this change as-is, and if we need to come back and change it to a hash then we can do that after the next release? This isn't a one-way door.

vrozov: It is not so much important for me whether it is commit hash or github build number as long as I can re-run CD workflow on the same commit without impacting/confusing infra build. In case github build number will be incremented each time CD build runs, will not that cause confusion whether it is the same artifact as already published or a new artifact?

camerski: That's OK, it will not cause any confusion. Once we select, say, build-15 as the version that goes into the next ODFE then that is locked in. If build-16 is identical then that's OK - we don't care. Once the release train has departed then there are no more changes :)

vrozov (approved):

vrozov:
@camerski The test CD workflow failed. Please take a look.

camerski:
> @camerski The test CD workflow failed. Please take a look.

Yes, that's to be expected. As noted in the PR message, the github secrets have not been updated yet. I should have made it clearer that this PR should not be merged yet - can you please revert the change, and I'll re-push it when we're ready to update the secrets.

PR #768, opened 2020-10-13
Demo config for plugins security and default alerting roles

skkosuri-amzn:
*Issue #, if available:*
 For https://github.com/opendistro-for-elasticsearch/alerting/issues/6

*Description of changes:*
1. Add alerting plugin config and alerts indices to `system_indices`
2. Add default alerting roles

### DIFF
--- a/securityconfig/roles.yml
+++ b/securityconfig/roles.yml
@@ -29,11 +29,11 @@ alerting_ack_alerts:
 alerting_full_access:
   reserved: true
   cluster_permissions:
-    - "cluster_monitor"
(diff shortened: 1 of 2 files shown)


sujithvm (approved):
Just to note that these roles.yml are applied only during bootstrapping of the cluster. Any cluster already having security index and migrating may not have these roles and hope its absence is handled in the alerting plugin. Since it is a reserved role, only superadmins can modify it. Maybe good to document it for customers

vrozov (approved):
Please provide a write up for release notes.

PR #769, opened 2020-10-14
Create release drafter

cliu123:
*Issue #, if available:*

*Description of changes:*
1. In order to save the effort and time going through all the merged PRs and generate release notes manually at the time of release, this PR is to introduce Release Drafter Github action into the workflow to keep track of all merged PRs and generate release notes.
2. The action will be triggered together with CI and CD, but the release note will be only generated upon merging PRs to master branch.
3. An example of the usage is [here](https://github.com/opendistro-for-elasticsearch/alerting/releases)


pom.xml
```diff
@@ -525,6 +525,44 @@
 
     <build>
         <plugins>
+            <plugin>
```
sujithvm, line 528: Remove changes not related to this PR

.github/release-drafter-config.yml
```diff
@@ -0,0 +1,23 @@
...
+    label: "breaking change"
+
+  - title: "Features"
+    label: "feature"
+
+  - title: "Enhancements"
+    label: "enhancements"
+
+  - title: "Bug Fixes"
+    labels: "bug fix"
```
sujithvm, line 23: Can you verify these are the same labels we have?

.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,18 @@
+name: Release-drafter
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
```
sujithvm, line 7: Can you confirm we need this trigger on pull_request? I don't the requirement on the official docs or general usage as well.

.github/release-drafter-config.yml
```diff
@@ -0,0 +1,23 @@
+# Configuration for Release Drafter: https://github.com/toolmantim/release-drafter
+template: |
```
sujithvm, line 2: Please refer to Alerting plugin https://github.com/opendistro-for-elasticsearch/alerting/blob/master/.github/draft-release-notes-config.yml so that we can have consistency across the release notes.

.github/release-drafter-config.yml
```diff
@@ -0,0 +1,23 @@
...
+  <!-- Optional: add a release summary here -->
+  $CHANGES
+
+name-template: $NEXT_MINOR_VERSION
+tag-template: $NEXT_MINOR_VERSION
+# Uses a more common 2-digit versioning in Jenkins plugins. Can be replaced by semver: $MAJOR.$MINOR.$PATCH
+version-template: $MAJOR.$MINOR
```
sujithvm, line 9: We don't follow this versioning style. We have 4 parts to semantic versioning. 
Also remove any unnecessary comments

.github/workflows/release-drafter.yml
cliu123: We normally don't do push on master branch, but I added this just in case anybody having authorization to push on master and really doing it.

vrozov: We do push on master branch.

.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  update_release_draft:
+    runs-on: ubuntu-latest
+    steps:
+      # Drafts the next Release notes as Pull Requests are merged into "master"
+      - uses: release-drafter/release-drafter@v5
+        with:
+          config-name: release-drafter-config.yml
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
vrozov, line 18: add \<eol>

.github/workflows/release-drafter.yml
vrozov: Do we need it on every push or pull? Don't we need this only on the release or release candidate tag?

cliu123: > Do we need it on every push or pull? Don't we need this only on the release or release candidate tag?

@vrozov Line #6 and #7 just intends to log the changes pushed on master. If we don't have to track those changes, I'll remove this.

.github/release-drafter-config.yml
```diff
@@ -0,0 +1,21 @@
...
+sort-by: merged_at
+sort-direction: ascending
+
+# Organizing the tagged PRs into unified ODFE categories
+categories:
+  - title: "Enhancements"
+    labels: "enhancement"
+
+  - title: "Bug fixes"
+    labels: "bug fix"
```
vrozov, line 18: Is this a newly added label? I see two labels `bug` and `bug fix`.

.github/workflows/release-drafter.yml
vrozov: This is github workflow/action definition file and it defines when action is triggered. I don't think it is necessary to trigger release draft action on every PR or push to the master. IMO, that should be merged into CD workflow.

.github/release-drafter-config.yml
cliu123: Yes, I added "bug fix" label in order to correctly categorize "Bug fixes" PRs in release notes.

.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,18 @@
...
+    branches: [master]
+
+jobs:
+  update_release_draft:
+    runs-on: ubuntu-latest
+    steps:
+      # Drafts the next Release notes as Pull Requests are merged into "master"
+      - uses: release-drafter/release-drafter@v5
+        with:
+          config-name: release-drafter-config.yml
```
vrozov, line 16: rename to `release-notes-drafter-config.yml`

.github/workflows/release-drafter.yml
cliu123: It'll be triggered on the PRs only when they get merged to master branch, not all PRs. I thought that release notes intend to track all changes merged/pushed to master branch(Please correct me if I'm wrong), which is why the action gets triggered up on PR merge and commit push on master branch.

.github/release-drafter-config.yml
cliu123: Done

.github/release-drafter-config.yml
cliu123: Done.

.github/workflows/release-drafter.yml
cliu123: Done.

.github/workflows/release-drafter.yml
cliu123: Done.

.github/release-drafter-config.yml
vrozov: How is it different from the existing `bug` label? Why do we need a new label?

.github/workflows/release-drafter.yml
vrozov: I do see it running on your PR that suggests that it will be triggered on every PR and every PR merge. Unless I miss something, I don't see why this is necessary and why it can't run only as part of CD.

cliu123: There was a miscommunication. I should say that the release note gets updated only when PRs get merged to master branch or commits get pushed on master branch. The action runs on other PRs, but it stops nearly immediately without updating release notes at all.

.github/release-drafter-config.yml
cliu123: The bug label is to categorize issues. But this 'bug fix' label is for PRs.

vrozov: Why the same label can't be used for both? IMO, it is very confusing and hard to remember which one should be used for issues and which one for PRs.

.github/workflows/release-drafter.yml
vrozov: It is not necessary to update release notes each time PR is merged. It is also not necessary to run the action on PR update (whether it exits without updating release notes or not).

.github/release-drafter-config.yml
cliu123: "bug" is described as "Somthing isn't working" on the label itself. But "bug fix" in intuitively express that it is a kind of fixes, which it doesn't seem to be confusing at all. And typing in the "labels" search bar will auto-completes the labels, which we can easily choose 'bug' or 'bug fixes'. However, if you strongly insist that there's no need for a new label, I can change it.

.github/workflows/release-drafter.yml
cliu123: I've removed the trigger for PR updates. But I believe that the action should be triggerd when PRs get merged to master. Otherwise, it wouldn't update release notes with the merge PR.

.github/release-drafter-config.yml
vrozov: I don't think such distinction is necessary. `bug` on the PR can't mean that PR introduces a bug, it already means that PR fixes a bug. I also checked few other projects (guava and elasticsearch) and I don't think that they do such distinction. Not as part of this PR, but it will be good to adopt a better labelling scheme (for example similar to guava) and change `bug` to `defect`.

cliu123: That's fine. I've removed 'bug fix' label and updated the PR to use 'bug' label to recognize the 'Bug fixes' PRs.

.github/release-notes-drafter-config.yml
```diff
@@ -0,0 +1,21 @@
+# Configuration for Release Drafter: https://github.com/toolmantim/release-drafter
+template: |
+  Compatible with Elasticsearch (set version here).
+  $CHANGES
+
+# Setting the formatting and sorting for the release notes body
+name-template: Version (set version here)
+change-template: '* $TITLE (#$NUMBER)'
```
vrozov, line 8:
```suggestion
change-template: '* $TITLE [#$NUMBER](https://github.com/opendistro-for-elasticsearch/security/pull/$NUMBER)'
```

dinusX (approved):

.github/workflows/release-drafter.yml
sujithvm: `pull_request` event is triggered when pull request is created against a branch and not when it is merged. You will have to specifically check for pull request `closed` and `merged` events but not needed here

sujithvm (approved):

PR #770, opened 2020-10-14
TenantInfo API should also check backend roles apart from just users …

palashhedau:
TenantInfo API should also check backend roles apart from just users in rolesmapping

Unit test added 

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/rest/TenantInfoAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/rest/TenantInfoAction.java
@@ -37,6 +37,9 @@ import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
 import java.util.SortedMap;
+
+import com.amazon.opendistroforelasticsearch.security.securityconf.impl.v6.RoleMappingsV6;
+import com.amazon.opendistroforelasticsearch.security.securityconf.impl.v7.RoleMappingsV7;
 import com.google.common.base.Strings;
 
 import com.amazon.opendistroforelasticsearch.security.configuration.ConfigurationRepository;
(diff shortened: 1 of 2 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/TenantInfoActionTest.java
```diff
@@ -57,7 +61,15 @@ public void testTenantInfoAPI() throws Exception {
...
 
-        response = rh.executePutRequest("/_opendistro/_security/api/rolesmapping/opendistro_security_internal", payload, new Header[0]);
+        response = rh.executePutRequest("/_opendistro/_security/api/rolesmapping/opendistro_security_internal", userPayload, new Header[0]);
+        Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
+
+        rh.sendAdminCertificate = false;
+        response = rh.executeGetRequest("_opendistro/_security/tenantinfo");
+        Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
+
+        rh.sendAdminCertificate = true;
```
cliu123, line 71: Why use admin certificate for users but not for backend role?

src/main/java/com/amazon/opendistroforelasticsearch/security/rest/TenantInfoAction.java
```diff
@@ -168,7 +171,18 @@ private boolean isAuthorized() {
...
+                if(roleMapping.getUsers().contains(user.getName())) {
+                    return true;
+                }
+                List<String> backendRoles = null;
+                if (roleMapping instanceof RoleMappingsV6) {
+                    backendRoles = ((RoleMappingsV6) roleMapping).getBackendroles();
+                } else if (roleMapping instanceof RoleMappingsV7) {
+                    backendRoles = ((RoleMappingsV7) roleMapping).getBackend_roles();
+                }
+                return backendRoles != null && backendRoles.contains(user.getName());
```
debjanibnrj, line 184: Won't this be backendRoles.contains(kibanaOpendistroRole) ?

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/TenantInfoActionTest.java
palashhedau: It depend upon the test we are performing.
If we want to send credentials with request, we use it, else no

PR #778, opened 2020-10-20
Adding support for SSL dual mode (#712)

sachetalva:
* Adding SSL dual mode support for SSLOnly

(cherry picked from commit 27f33fb)

Co-authored-by: Vengadanathan Srinivasan <vengadan@amazon.com>

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/security/test/SingleClusterTest.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/security/test/SingleClusterTest.java
@@ -42,6 +42,9 @@ import com.amazon.opendistroforelasticsearch.security.test.helper.rest.RestHelpe
 
 public abstract class SingleClusterTest extends AbstractSecurityUnitTest {
 
+    private static final int DEFAULT_MASTER_NODE_NUM = 3;
+    private static final int DEFAULT_FIRST_DATA_NODE_NUM = 2;
+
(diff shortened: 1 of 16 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/security/test/SingleClusterTest.java
```diff
@@ -42,8 +42,8 @@
 
 public abstract class SingleClusterTest extends AbstractSecurityUnitTest {
 
-    private static final int DEFAULT_MASTER_NODE_NUM = 3;
```
hardik-k-shah: why this change is specific to 0.9 branch? 

I see in other versions it is 
private static final int DEFAULT_MASTER_NODE_NUM = 3;
private static final int DEFAULT_FIRST_DATA_NODE_NUM = 2;

sachetalva: This numbering is with respect to DEFAULT_MASTER_WITHOUT_SECURITY_PLUGIN and DEFAULT_ONE_DATA_NODE_WITHOUT_SECURITY_PLUGIN defined in ClusterConfiguration.
These enums define a 3 node cluster with 1 node not having the security plugin.
in ES 6.x cluster the node numbering seems to be in a reverse order compared to ES 7.x cluster, which why had to do this change for ES 6.x plugins

hardik-k-shah (approved):

PR #796, opened 2020-10-24
Fix the way to store empty password in ES6

cliu123:
*Issue #, if available:*

*Description of changes:*
The empty password stored in ES6 causes upgrade fail. This PR is to fix the storage machnism of empty password in ES6.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
@@ -122,9 +122,10 @@ public class InternalUsersApiAction extends PatchableResourceApiAction {
 
         // if password is set, it takes precedence over hash
         String plainTextPassword = additionalSettingsBuilder.get("password");
-        if (plainTextPassword != null && plainTextPassword.length() > 0) {
+        if (plainTextPassword != null) {
             additionalSettingsBuilder.remove("password");
-            additionalSettingsBuilder.put("hash", hash(plainTextPassword.toCharArray()));
+            if (plainTextPassword.length() > 0)
+                additionalSettingsBuilder.put("hash", hash(plainTextPassword.toCharArray()));
         }
 
         // check if user exists


vrozov (changes requested):
In case empty password is not allowed, why not to explicitly deny it instead of relying on the presence of the `hash` in the request? What is the use case for allowing empty `password` and empty or not empty `hash`?

The PR does not address migration issue for existing clusters where empty password may be already present.

cliu123:
> In case empty password is not allowed, why not to explicitly deny it instead of relying on the presence of the `hash` in the request? What is the use case for allowing empty `password` and empty or not empty `hash`?
> 
> The PR does not address migration issue for existing clusters where empty password may be already present.

"Creates or replaces the specified user. You must specify either password (plain text) or hash (the hashed user password). If you specify password, the security plugin automatically hashes the password before storing it.". 
Quote from the documentation of Security plugin: https://opendistro.github.io/for-elasticsearch-docs/docs/security/access-control/api/#create-user

vrozov:
If password is specified, the hash of the specified password should be stored. Empty password is a *specified* password. It is supposed to be either hashed and not ignored or not allowed. Based on how it was handled previously, it looks like empty passwords are not allowed.

cliu123:
> If password is specified, the hash of the specified password should be stored. Empty password is a _specified_ password. It is supposed to be either hashed and not ignored or not allowed. Based on how it was handled previously, it looks like empty passwords are not allowed.

But based on my testing, internal users with empty password get created successfully in both ES6 and ES7, as long as hash is not empty. In terms of a new internal user creation, as long as one of password and hash is specified and not empty, the new internal user gets created successfully. But password will take precedence if both of them are not empty. That's true in both documentation, the code and my testing results.

vrozov:
What is the purpose and use case for specifying empty password when it is replaced with hash no matter what? When password is not specified, hash is used, when password is specified, password (whether empty or not) should be used. If empty password is not allowed, it should not be allowed whether or not hash is specified.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -122,9 +122,13 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
 
         // if password is set, it takes precedence over hash
         String plainTextPassword = additionalSettingsBuilder.get("password");
-        if (plainTextPassword != null && plainTextPassword.length() > 0) {
-            additionalSettingsBuilder.remove("password");
-            additionalSettingsBuilder.put("hash", hash(plainTextPassword.toCharArray()));
+        if (plainTextPassword != null) {
+            if (plainTextPassword.length() == 0)
+                additionalSettingsBuilder.remove("password");
```
palashhedau, line 127: This statement is present in both "if" and "else".
We can remove from "if else" and just put before "if else"

palashhedau (commented):
Add a unit test. There might already be there, but just add some assert or something

vrozov:
Please start with changing the behavior on master if it applies to 7.x.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -122,9 +122,10 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
 
         // if password is set, it takes precedence over hash
         String plainTextPassword = additionalSettingsBuilder.get("password");
-        if (plainTextPassword != null && plainTextPassword.length() > 0) {
+        if (plainTextPassword != null) {
             additionalSettingsBuilder.remove("password");
-            additionalSettingsBuilder.put("hash", hash(plainTextPassword.toCharArray()));
+            if (plainTextPassword.length() > 0)
+                additionalSettingsBuilder.put("hash", hash(plainTextPassword.toCharArray()));
```
sujithvm, line 128: If empty password is not allowed then we should fail the request instead of ignoring it. Empty password if set may be considered as valid password and hash can be generated for empty string.

PR #798, opened 2020-10-26
Add user & roles to the thread context 

skkosuri-amzn:
*Issue #, if available:*
*Description of changes:*
 - Add user & roles info as a string to the thread context.
 - Format of the string:  username|backend_role1,backend_role2|odferole1,odferole2
-  This string is accessable only in transport layer, as the odferoles are resolved only in transport layer.
-  key : `_opendistro_security_user_roles_string`


sujithvm (dismissed):

hardik-k-shah:
Why do we need this change? What is the usage?

skkosuri-amzn:
> Why do we need this change? What is the usage?

Hi @hardik-k-shah ,  This is for the plugins to get opendistro roles from thread context, instead of making a rest call to `/authinfo`.

hardik-k-shah:
> > Why do we need this change? What is the usage?
> 
> Hi @hardik-k-shah , This is for the plugins to get opendistro roles from thread context, instead of making a rest call to `/authinfo`.

Thanks. Based on my understanding thread context already have details about user and role.

hardik-k-shah:
> > > Why do we need this change? What is the usage?
> > 
> > 
> > Hi @hardik-k-shah , This is for the plugins to get opendistro roles from thread context, instead of making a rest call to `/authinfo`.
> 
> Thanks. Based on my understanding thread context already have details about user and role.

Just followed up with @vrozov offline.  "This detail is available as User object in thread context and plugin needs to depend on Security jar to read from object. 
So, to avoid dependency on the plugin jar  we are also storing as string format."

hardik-k-shah (dismissed):

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -291,7 +292,11 @@ public int order() {
             if (log.isDebugEnabled()) {
                 log.debug(pres);
             }
-            
+
+            if(threadContext.getTransient(OPENDISTRO_SECURITY_USER_ROLES_STRING) == null) {
```
vrozov, line 296: nit: populate only if there is a call to `chain`. If access is denied, it is necessary to populate `OPENDISTRO_SECURITY_USER_ROLES_STRING`?

src/main/java/com/amazon/opendistroforelasticsearch/security/user/User.java
```diff
@@ -263,4 +263,8 @@ public final void addOpenDistroSecurityRoles(final Collection<String> securityRo
     public final Set<String> getOpenDistroSecurityRoles() {
         return this.openDistroSecurityRoles == null ? Collections.emptySet() : Collections.unmodifiableSet(this.openDistroSecurityRoles);
     }
+
+    public final String getUserRolesString() {
+        return name + "|" + String.join(",", getRoles()) + "|" + String.join(",", getOpenDistroSecurityRoles());
```
vrozov, line 268: Can roles or odfe roles contain `,`?

skkosuri-amzn: As far as I know, it can't contain `,`. Pipe separated string format is used at other places including : UserInjector, RolesInjector.

src/main/java/com/amazon/opendistroforelasticsearch/security/support/ConfigConstants.java
```diff
@@ -99,6 +99,8 @@
     public static final String OPENDISTRO_SECURITY_USER = OPENDISTRO_SECURITY_CONFIG_PREFIX+"user";
     public static final String OPENDISTRO_SECURITY_USER_HEADER = OPENDISTRO_SECURITY_CONFIG_PREFIX+"user_header";
 
+    public static final String OPENDISTRO_SECURITY_USER_ROLES_STRING = OPENDISTRO_SECURITY_CONFIG_PREFIX + "user_roles_string";
```
vrozov, line 102: For consistency with `injected_user` (that is also a String) please rename to remove `STRING`: 'OPENDISTRO_SECURITY_USER_ROLES_STRING` ->`OPENDISTRO_SECURITY_USER_AND_ROLES`, `"user_roles_string"` -> "user&roles" or "user_and_roles".

skkosuri-amzn: Renamed. thanks.

hardik-k-shah (approved):

sujithvm (approved):

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
vrozov: Does this require order of plugins loading to be enforced? If yes, how that will be done?

vrozov: Discussed with @skkosuri-amzn, the ordering needs to be researched. Can be done as a follow up PR.

vrozov (approved):

PR #809, opened 2020-10-27
Fix empty password issue in upgrade from 6x to 7x

cliu123:
*Issue #, if available:*

*Description of changes:*
Unrecognized field "password" exception gets thrown out during migrating from 6.x to 7.x. 
This PR is to fix the logic handling empty password in migration.


src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
```diff
@@ -214,6 +217,19 @@ public T run() throws Exception {
             return AccessController.doPrivileged(new PrivilegedExceptionAction<T>() {
                 @Override
                 public T run() throws Exception {
+                    if (jt.getBindings().getTypeParameters().get(0).getRawClass() == InternalUserV6.class) {
```
palashhedau, line 220: Do we want to check if this is for the migration from 6.x to 7.x or we are doing this even if is called when invoked from 6.x to 6.x_higher?

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
```diff
@@ -214,6 +217,19 @@ public T run() throws Exception {
...
+                        JsonNode root = readTree(string);
+                        Iterator<JsonNode> internalUsers = root.elements();
+                        while (internalUsers.hasNext()) {
+                            ObjectNode internalUser = (ObjectNode) internalUsers.next();
+                            JsonNode plainTextPassword = internalUser.get("password");
+                            if (plainTextPassword == null)
+                                continue;
+                            if (plainTextPassword.asText().length() == 0)
```
palashhedau, line 228: Add some comments around this change as why are we doing this

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
```diff
@@ -214,6 +217,19 @@ public T run() throws Exception {
             return AccessController.doPrivileged(new PrivilegedExceptionAction<T>() {
                 @Override
                 public T run() throws Exception {
+                    if (jt.getBindings().getTypeParameters().get(0).getRawClass() == InternalUserV6.class) {
+                        JsonNode root = readTree(string);
```
palashhedau, line 221: Add some unit tests around this. This will help regain code coverage.

vrozov:
It should be sufficient to introduce `setPassword(String password)` setter to `InternalUserV6`. Optionally setter can assert that `password` is an empty string.

cliu123:
> It should be sufficient to introduce `setPassword(String password)` setter to `InternalUserV6`. Optionally setter can assert that `password` is an empty string.

But Optional.isEmpty() asserts Optional.value==null instead of empty string. Using optional setter could require more changes in other places?

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
cliu123: Yes, we want to do this check for both 6.x and 7.x. Preventing empty password is the expected behavior for both 6.x and 7.x. This PR is to fix the empty password issue in migration for existing domains that have had empty passwords. We're preventing empty password in Internal User creation in another PR.

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
sujithvm, line 228: You can directly remove any password if exists? We should not be storing any plaintext password. Only hash should be stored

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
```diff
@@ -214,6 +217,19 @@ public T run() throws Exception {
...
+                        Iterator<JsonNode> internalUsers = root.elements();
+                        while (internalUsers.hasNext()) {
+                            ObjectNode internalUser = (ObjectNode) internalUsers.next();
+                            JsonNode plainTextPassword = internalUser.get("password");
+                            if (plainTextPassword == null)
+                                continue;
+                            if (plainTextPassword.asText().length() == 0)
+                                internalUser.remove("password");
+                        }
+                        return objectMapper.readValue(writeValueAsString(root, false), jt);
```
sujithvm, line 231: Can you confirm if we are using default object mapper and not non-validating object mapper ?

sujithvm:
Wondering if any change can be directly done in https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/Migration.java#L160 ?

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
cliu123: Yes, empty passwords will be removed from here to prevent migration issue caused by empty password for existing domains that have already had internal users with empty passwords.

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
cliu123: Migration hits DefaultObjectMapper.java based on my debugging. It calls com.fasterxml.jackson.databind.ObjectMapper.java

vrozov:
> > It should be sufficient to introduce `setPassword(String password)` setter to `InternalUserV6`. Optionally setter can assert that `password` is an empty string.
> 
> But Optional.isEmpty() asserts Optional.value==null instead of empty string. Introducing optional setter could require more changes?

It is not optional setter, setter will be called only when "password" field is present in serialized json. I don't think that any other changes will be necessary. When I say "optionally", I mean that it is up to you to decide whether it is necessary to check that "password" is empty (it can't be anything else).

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
sujithvm: I mean to say, we don't have to explicitly check for empty password. It should not exist at all and even if it exists, it will be empty.

cliu123:
> > > It should be sufficient to introduce `setPassword(String password)` setter to `InternalUserV6`. Optionally setter can assert that `password` is an empty string.
> > 
> > 
> > But Optional.isEmpty() asserts Optional.value==null instead of empty string. Introducing optional setter could require more changes?
> 
> It is not optional setter, setter will be called only when "password" field is present in serialized json. I don't think that any other changes will be necessary. When I say "optionally", I mean that it is up to you to decide whether it is necessary to check that "password" is empty (it can't be anything else).

Yeah, this is a much simpler solution and it perfectly keeps the consistency with the way how other fields get populated in Internal User object. Thanks for pointing this out, Vlad!

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/impl/v6/InternalUserV6.java
```diff
@@ -79,6 +79,7 @@ public String getHash() {
         public void setHash(String hash) {
             this.hash = hash;
         }
+        public void setPassword(String password){}
```
vrozov, line 82: Please add comments explaining why the empty setter for password is necessary. Follow code style.

src/main/java/com/amazon/opendistroforelasticsearch/security/DefaultObjectMapper.java
cliu123: I agree. Thanks for clarifying this!

Issue #56, opened 2019-05-30
User impersonation doesn't work in default Kubernetes + Ingress environment


dpotapov:
The header `opendistro_security_impersonate_as` contains underscore characters, which are banned by default:
http://nginx.org/en/docs/http/ngx_http_core_module.html#underscores_in_headers

Ingress can be configured with this server-level option:
https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#enable-underscores-in-headers

However not everyone have possibility to change it. Thus would be good to have impersonation header with dashes at least...

nateynateynate:
Hi @dpotapov! Thanks for the issue. It's been a while since submission - I'd love to find out if this is still causing you any pain. Do you mind trying this out on the latest version of OpenSearch to see if those underscored headers are still being blocked? 

Thanks! 

Nate

dpotapov:
hi, I'm no longer working on that project. Feel free to close the issue since no one else hit into this.

PR #818, opened 2020-10-30
Reject empty password in internal user creation

cliu123:
*Issue #, if available:* 

*Description of changes:*
Given that Kibana and ES don't allow login with empty password, it'd make sense to deny empty password in internal user creation.
This PR is to add the validation to deny empty password in internal user creation.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
@@ -17,6 +17,7 @@ package com.amazon.opendistroforelasticsearch.security.dlic.rest.validation;
 
 import com.amazon.opendistroforelasticsearch.security.ssl.util.Utils;
 import com.amazon.opendistroforelasticsearch.security.support.ConfigConstants;
+import org.elasticsearch.common.Strings;
 import org.elasticsearch.common.bytes.BytesReference;
 import org.elasticsearch.common.compress.NotXContentException;
 import org.elasticsearch.common.settings.Settings;
@@ -53,39 +54,44 @@ public class CredentialsValidator extends AbstractConfigurationValidator {
         final String regex = this.esSettings.get(ConfigConstants.OPENDISTRO_SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX, null);
 
         if ((request.method() == RestRequest.Method.PUT || request.method() == RestRequest.Method.PATCH)
-                && regex != null
-                && !regex.isEmpty()
                 && this.content != null
                 && this.content.length() > 1) {
             try {
                 final Map<String, Object> contentAsMap = XContentHelper.convertToMap(this.content, false, XContentType.JSON).v2();
-                if (contentAsMap != null && contentAsMap.containsKey("password")) {
-                    final String password = (String) contentAsMap.get("password");
-
-                    // Password can be null/empty for an existing user. Regex will validate password if present
-                    if (password != null && !password.isEmpty() && !regex.isEmpty() && !Pattern.compile("^"+regex+"$").matcher(password).matches()) {
-                        if(log.isDebugEnabled()) {
-                            log.debug("Regex does not match password");
-                        }
+                String password = (String) contentAsMap.get("password");
+                if (password != null) {
+                    // Password is not allowed to be empty if present.
+                    if (password.isEmpty()) {
                         this.errorType = ErrorType.INVALID_PASSWORD;
                         return false;
                     }
 
-                    final String username = Utils.coalesce(request.param("name"), hasParams() ? (String) param[0] : null);
-
-                    if (username == null || username.isEmpty()) {
-                        if (log.isDebugEnabled()) {
-                            log.debug("Unable to validate username because no user is given");
+                    if (!Strings.isNullOrEmpty(regex)) {
+                        // Password can be null for an existing user. Regex will validate password if present
+                        if (!Pattern.compile("^"+regex+"$").matcher(password).matches()) {
+                            if(log.isDebugEnabled()) {
+                                log.debug("Regex does not match password");
+                            }
+                            this.errorType = ErrorType.INVALID_PASSWORD;
+                            return false;
                         }
-                        return false;
-                    }
 
-                    if (username.toLowerCase().equals(password.toLowerCase())) {
-                        if (log.isDebugEnabled()) {
-                            log.debug("Username must not match password");
+                        final String username = Utils.coalesce(request.param("name"), hasParams() ? (String) param[0] : null);
+
+                        if (username == null || username.isEmpty()) {
+                            if (log.isDebugEnabled()) {
+                                log.debug("Unable to validate username because no user is given");
+                            }
+                            return false;
                         }
-                        this.errorType = ErrorType.INVALID_PASSWORD;
-                        return false;
+
+                        if (username.toLowerCase().equals(password.toLowerCase())) {
+                            if (log.isDebugEnabled()) {
+                                log.debug("Username must not match password");
+                            }
+                            this.errorType = ErrorType.INVALID_PASSWORD;
+                            return false;
+                        }
                     }
                 }
             } catch (NotXContentException e) {
--- a/src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
@@ -410,6 +410,9 @@ public class UserApiTest extends AbstractRestApiUnitTest {
         addUserWithPassword("ok2", "$aA123456789", HttpStatus.SC_CREATED);
         addUserWithPassword("ok3", "$Aa123456789", HttpStatus.SC_CREATED);
         addUserWithPassword("ok4", "$1aAAAAAAAAA", HttpStatus.SC_CREATED);
+        addUserWithPassword("empty_password_no_hash", "", HttpStatus.SC_BAD_REQUEST);
+        addUserWithPasswordAndHash("empty_password", "", "$%^123", HttpStatus.SC_BAD_REQUEST);
+        addUserWithPasswordAndHash("null_password", null, "$%^123", HttpStatus.SC_BAD_REQUEST);
 
         response = rh.executePatchRequest("/_opendistro/_security/api/internalusers", "[{ \"op\": \"add\", \"path\": \"/ok4\", \"value\": {\"password\": \"bla\", \"backend_roles\": [\"vulcan\"] } }]", new Header[0]);
         Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, response.getStatusCode());
@@ -448,6 +451,9 @@ public class UserApiTest extends AbstractRestApiUnitTest {
         Assert.assertTrue(response.getBody().contains("xxx"));
 
         response = rh.executePutRequest("/_opendistro/_security/api/internalusers/ok1", "{\"backend_roles\":[\"my-backend-role\"],\"attributes\":{},\"password\":\"\"}", new Header[0]);
+        Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, response.getStatusCode());
+
+        response = rh.executePutRequest("/_opendistro/_security/api/internalusers/ok1", "{\"backend_roles\":[\"my-backend-role\"],\"attributes\":{},\"password\":\"Admin_123\"}", new Header[0]);
         Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
 
         response = rh.executePutRequest("/_opendistro/_security/api/internalusers/ok1", "{\"backend_roles\":[\"my-backend-role\"],\"attributes\":{}}", new Header[0]);
@@ -502,6 +508,9 @@ public class UserApiTest extends AbstractRestApiUnitTest {
                 HttpStatus.SC_CREATED);
 
         response = rh.executePutRequest("/_opendistro/_security/api/internalusers/user1", "{\"hash\":\"$2a$12$n5nubfWATfQjSYHiWtUyeOxMIxFInUHOAx8VMmGmxFNPGpaBmeB.m\",\"password\":\"\",\"backend_roles\":[\"admin\",\"rolea\"]}");
+        Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, response.getStatusCode());
+
+        response = rh.executePutRequest("/_opendistro/_security/api/internalusers/user1", "{\"hash\":\"$2a$12$n5nubfWATfQjSYHiWtUyeOxMIxFInUHOAx8VMmGmxFNPGpaBmeB.m\",\"password\":\"Admin_123\",\"backend_roles\":[\"admin\",\"rolea\"]}");
         Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
 
         response = rh.executeGetRequest("/_opendistro/_security/api/internalusers/user1");
@@ -511,6 +520,9 @@ public class UserApiTest extends AbstractRestApiUnitTest {
                 HttpStatus.SC_CREATED);
 
         response = rh.executePutRequest("/_opendistro/_security/api/internalusers/user2", "{\"password\":\"\",\"backend_roles\":[\"admin\",\"rolex\"]}");
+        Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, response.getStatusCode());
+
+        response = rh.executePutRequest("/_opendistro/_security/api/internalusers/user2", "{\"password\":\"Admin_123\",\"backend_roles\":[\"admin\",\"rolex\"]}");
         Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
 
         response = rh.executeGetRequest("/_opendistro/_security/api/internalusers/user2");
@@ -532,8 +544,12 @@ public class UserApiTest extends AbstractRestApiUnitTest {
         response = rh.executeDeleteRequest("/_opendistro/_security/api/internalusers/sarek" , new Header[0]);
         Assert.assertEquals(HttpStatus.SC_FORBIDDEN, response.getStatusCode());
 
+        // Patch read only users
+        response = rh.executePatchRequest("/_opendistro/_security/api/internalusers/sarek", "[{ \"op\": \"add\", \"path\": \"/sarek/description\", \"value\": \"foo\" }]", new Header[0]);
+        Assert.assertEquals(HttpStatus.SC_FORBIDDEN, response.getStatusCode());
+
         // Put read only users
-        response = rh.executePutRequest("/_opendistro/_security/api/internalusers/sarek", "[{ \"op\": \"add\", \"path\": \"/sarek/description\", \"value\": \"foo\" }]", new Header[0]);
+        response = rh.executePutRequest("/_opendistro/_security/api/internalusers/sarek", "{ \"opendistro_security_roles\": [\"opendistro_security_reserved\"]}", new Header[0]);
         Assert.assertEquals(HttpStatus.SC_FORBIDDEN, response.getStatusCode());
 
         // Patch single read only user
@@ -552,8 +568,12 @@ public class UserApiTest extends AbstractRestApiUnitTest {
         response = rh.executeDeleteRequest("/_opendistro/_security/api/internalusers/hide" , new Header[0]);
         Assert.assertEquals(HttpStatus.SC_NOT_FOUND, response.getStatusCode());
 
+        // Patch hidden users
+        response = rh.executePatchRequest("/_opendistro/_security/api/internalusers/hide", "[{ \"op\": \"add\", \"path\": \"/sarek/description\", \"value\": \"foo\" }]", new Header[0]);
+        Assert.assertEquals(HttpStatus.SC_NOT_FOUND, response.getStatusCode());
+
         // Put hidden users
-        response = rh.executePutRequest("/_opendistro/_security/api/internalusers/hide", "[{ \"op\": \"add\", \"path\": \"/sarek/description\", \"value\": \"foo\" }]", new Header[0]);
+        response = rh.executePutRequest("/_opendistro/_security/api/internalusers/hide", "{ \"opendistro_security_roles\": [\"opendistro_security_reserved\"]}", new Header[0]);
         Assert.assertEquals(HttpStatus.SC_NOT_FOUND, response.getStatusCode());
 
         // Put reserved role is forbidden for non-superadmin
--- a/src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractRestApiUnitTest.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractRestApiUnitTest.java
@@ -181,6 +181,15 @@ public abstract class AbstractRestApiUnitTest extends SingleClusterTest {
 		rh.sendAdminCertificate = sendAdminCertificate;
 	}
 
+	protected void addUserWithPasswordAndHash(String username, String password, String hash, int status) throws Exception {
+		boolean sendAdminCertificate = rh.sendAdminCertificate;
+		rh.sendAdminCertificate = true;
+		HttpResponse response = rh.executePutRequest("/_opendistro/_security/api/internalusers/" + username, "{\"hash\": \"" + hash + "\", \"password\": \"" + password + "\"}",
+				new Header[0]);
+		Assert.assertEquals(status, response.getStatusCode());
+		rh.sendAdminCertificate = sendAdminCertificate;
+	}
+
 	protected void checkGeneralAccess(int status, String username, String password) throws Exception {
 		boolean sendAdminCertificate = rh.sendAdminCertificate;
 		rh.sendAdminCertificate = false;


sujithvm:
Any reason why we want to allow empty password? Empty password may be considered as security vulnerability and most UI/front end systems block passing empty passwords

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -116,8 +116,14 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
             }
         }
 
-        // if password is set, it takes precedence over hash
+        // Reject empty password
         final String plainTextPassword = securityJsonNode.get("password").asString();
+        if (plainTextPassword != null && plainTextPassword.isEmpty()){
+            badRequestResponse(channel, "Please specify a non-empty 'password' when creating a new internal user.");
```
sujithvm, line 122: Can you reject in CredentialsValidator which is used by the account API for self-change password as well ?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -116,8 +116,14 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
...
+            return;
+        }
+
+        // if password is set, it takes precedence over hash
```
sujithvm, line 126: Can you remove other checks for `isEmpty()` ?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
cliu123: I tried that. But CredentialsValidator validates regex, and it respondes status code 201: {"status":"error","reason":"Password does not match minimum criteria"}. I believe what we want for empty password is status code 400: {"status":"BAD_REQUEST","message":"Please specify a non-empty 'password' when creating a new internal user."}.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
cliu123: Done.

sujithvm: Can you simplify the checks? we only need to check for password is not null and replace hash so that it takes precedence

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
sujithvm: It returns 201 with that error message? This should not happen. 400 must be returned. Can you verify again

cliu123: Moved the validation to CreadentialValidator.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
```diff
@@ -53,39 +53,45 @@ public boolean validate() {
...
+                        // Regex will validate password if present
+                        if (password != null && !Pattern.compile("^"+regex+"$").matcher(password).matches()) {
+                            if(log.isDebugEnabled()) {
+                                log.debug("Regex does not match password");
+                            }
+                            this.errorType = ErrorType.INVALID_PASSWORD;
+                            return false;
+                        }
+
+                        final String username = Utils.coalesce(request.param("name"), hasParams() ? (String) param[0] : null);
```
sujithvm, line 79: Fine with keeping current behavior but username validation does not depend on regex

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -448,7 +449,7 @@ public void testPasswordRules() throws Exception {
         Assert.assertTrue(response.getBody().contains("xxx"));
 
         response = rh.executePutRequest("/_opendistro/_security/api/internalusers/ok1", "{\"backend_roles\":[\"my-backend-role\"],\"attributes\":{},\"password\":\"\"}", new Header[0]);
-        Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
+        Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, response.getStatusCode());
```
sujithvm, line 452: Do include test for empty password but fix request instead of response for the tests as they are asserting updating other attributes

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -534,7 +535,7 @@ public void testUserApiForNonSuperAdmin() throws Exception {
 
         // Put read only users
         response = rh.executePutRequest("/_opendistro/_security/api/internalusers/sarek", "[{ \"op\": \"add\", \"path\": \"/sarek/description\", \"value\": \"foo\" }]", new Header[0]);
-        Assert.assertEquals(HttpStatus.SC_FORBIDDEN, response.getStatusCode());
+        Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, response.getStatusCode());
```
sujithvm, line 538: Why is this bad request ? it is updating read only user and password field is not present

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -554,7 +555,7 @@ public void testUserApiForNonSuperAdmin() throws Exception {
 
         // Put hidden users
         response = rh.executePutRequest("/_opendistro/_security/api/internalusers/hide", "[{ \"op\": \"add\", \"path\": \"/sarek/description\", \"value\": \"foo\" }]", new Header[0]);
-        Assert.assertEquals(HttpStatus.SC_NOT_FOUND, response.getStatusCode());
+        Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, response.getStatusCode());
```
sujithvm, line 558: Why is this bad request ? it is updating reserved user and password field is not present

sujithvm:
Please check if security kibana plugin needs to be updated to not provide empty password in any request. Admin user should be able to update attributes for other users without having to supply any password.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
```diff
@@ -53,39 +53,45 @@ public boolean validate() {
...
                 if (contentAsMap != null && contentAsMap.containsKey("password")) {
                     final String password = (String) contentAsMap.get("password");
 
-                    // Password can be null/empty for an existing user. Regex will validate password if present
-                    if (password != null && !password.isEmpty() && !regex.isEmpty() && !Pattern.compile("^"+regex+"$").matcher(password).matches()) {
-                        if(log.isDebugEnabled()) {
-                            log.debug("Regex does not match password");
-                        }
+                    // Password is not allowed to be empty if present.
+                    if (password != null && password.isEmpty()) {
```
sujithvm, line 64: nit: Use `isNullOrEmpty`

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
cliu123: Moving Line #56 - 57 down [here](https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java#L56) caused NotXContentException on Line #6. The method validate() still returns false for this use case, but bad request instead of forbidden.

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
cliu123: Ditto

cliu123:
> Please check if security kibana plugin needs to be updated to not provide empty password in any request. Admin user should be able to update attributes for other users without having to supply any password.

That's a good point! I'll check with Kibana team.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
cliu123: `isNullOrEmpty` is equivalent to `password != null || password.isEmpty()`, but I thought what we want to is `password != null && password.isEmpty()`.

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
cliu123: Fixed the response for expected behavior change. And keep the original test case below.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/InternalUsersApiAction.java
cliu123: Done.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
sujithvm: I was referring to regex null or empty but looks like that line of code got removed in this code revision

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
```diff
@@ -62,8 +62,14 @@ public boolean validate() {
                 if (contentAsMap != null && contentAsMap.containsKey("password")) {
                     final String password = (String) contentAsMap.get("password");
 
+                    // Password is not allowed to be empty if present.
+                    if (password != null && password.isEmpty()) {
```
sujithvm, line 66: Is the empty password checked only if regex is configured ? I see lines L56-57 inserted back in.
```
&& regex != null
&& !regex.isEmpty()	           
```

cliu123: Yes, [XContentHelper.convertToMap()](https://github.com/opendistro-for-elasticsearch/security/blob/017ae5f4ec975c3324934ab58df5305297979ba7/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java#L61) has dependencies on `regex != null && !regex.isEmpty()`. Without Line 56-57, it throws out [`NotXContentException`](https://github.com/opendistro-for-elasticsearch/security/blob/017ae5f4ec975c3324934ab58df5305297979ba7/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java#L100) and makes the `validate()` return a false and respond bad request.

sujithvm: Could you explain the dependency of `XContentHelper.convertToMap()` on regex and why it is throwing `NotXContentException ` ? 

We would still allow users to use empty passwords unless users specify regex. Why do we want this behavior instead of rejecting empty passwords for all scenarios.

cliu123: > Could you explain the dependency of `XContentHelper.convertToMap()` on regex and why it is throwing `NotXContentException ` ?
> 
> We would still allow users to use empty passwords unless users specify regex. Why do we want this behavior instead of rejecting empty passwords for all scenarios.

There is a easy way to verify the dependencies. If you comment out the line 56 - 57 and run UserApiTest.testUserApiForNonSuperAdmin(), [this test](https://github.com/opendistro-for-elasticsearch/security/blob/4ed9f558b93082b44e80d5991d9b67c253e9f7f2/src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java#L537) will get a bad response instead of a forbidden. If you check the log, you'll see the `NotXContentException`.

sujithvm: Are you saying the test code needs to be fixed? If yes, please fix as needed. 
`XContentHelper.convertToMap()` is an elasticsearch function. I don't see how it can be dependent on the regex configuration of the security plugin

cliu123: It's not dependent on the regex configuation. `XContentHelper.convertToMap()` cannot map XContent correctly when `XContent` content contains a `null` field. In this case, without `regex != null`, the regex can be null, which makes the `XContentHelper.convertToMap()` throw out [`NotXContentException`](https://github.com/elastic/elasticsearch/blob/724b244d7a72e18b9bd184f4bcb9cfc1f4967a16/server/src/main/java/org/elasticsearch/common/compress/CompressorFactory.java#L49)

sujithvm: Regex is not part of content hence it cannot interfere with `XContentHelper.convertToMap()`

sujithvm: It is converting array PATCH array body with PUT request in the test `testUserApiForNonSuperAdmin` and is failing to convert array into map. Need to fix the tests

cliu123: Ok. That's the culprit. I fixed it.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
cliu123: > nit: Use `isNullOrEmpty`

Done

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
cliu123: Fixed the test.

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
cliu123: Fixed the test.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
cliu123: I see.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
```diff
@@ -53,39 +54,45 @@ public boolean validate() {
...
+                            }
+                            return false;
                         }
-                        return false;
-                    }
 
-                    if (username.toLowerCase().equals(password.toLowerCase())) {
-                        if (log.isDebugEnabled()) {
-                            log.debug("Username must not match password");
+                        if (username.toLowerCase().equals(password.toLowerCase())) {
```
sujithvm, line 89: Can password be null here? 
Maybe check for password is not null directly in the contentAsMap as these conditions are based on non-null password entry validation

sujithvm: Also not dependent on regex but it is inside `Strings.isNullOrEmpty(regex)` block? I understand the current behavior validates password against username only if regex is defined. Just looking for improvements here. Any thoughts?

sujithvm (dismissed):

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
cliu123: Yes, I thought about that too. In theory, password has nothing to do with regex. I feel like there're some points/use cases to validate password against username no matter regex is null/empty or not and refactoring could be needed here. It's worth to discuss with the team and to get more inputs.

sujithvm: Can we address in this PR itself as change is happening around it

cliu123: I'll create another PR to make this change given that this PR has been sitting there for more than 3 weeks and it has met the goal. Let's stick with the original scope of PRs.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
```diff
@@ -53,39 +54,45 @@ public boolean validate() {
...
                 if (contentAsMap != null && contentAsMap.containsKey("password")) {
                     final String password = (String) contentAsMap.get("password");
 
-                    // Password can be null/empty for an existing user. Regex will validate password if present
-                    if (password != null && !password.isEmpty() && !regex.isEmpty() && !Pattern.compile("^"+regex+"$").matcher(password).matches()) {
-                        if(log.isDebugEnabled()) {
-                            log.debug("Regex does not match password");
-                        }
+                    // Password is not allowed to be empty if present.
+                    if (password != null && password.isEmpty()) {
```
vrozov, line 65: Under what condition `password` will be `null` here? It is executed with `contentAsMap.containsKey("password")` condition.

cliu123: In theory, contentAsMap.containsKey() prevents null password. But before checking isEmpty(), it's good to ensure the object "password" exists, isn't it?

vrozov: I don't see why two checks are necessary. The check `if (contentAsMap != null && contentAsMap.containsKey("password"))` looks redundant to me.

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -410,6 +410,7 @@ public void testPasswordRules() throws Exception {
         addUserWithPassword("ok2", "$aA123456789", HttpStatus.SC_CREATED);
         addUserWithPassword("ok3", "$Aa123456789", HttpStatus.SC_CREATED);
         addUserWithPassword("ok4", "$1aAAAAAAAAA", HttpStatus.SC_CREATED);
+        addUserWithPasswordAndHash("empty_password", "", "$%^123", HttpStatus.SC_BAD_REQUEST);
```
vrozov, line 413: Is there a test that adds user with empty password and no hash?

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -448,7 +449,10 @@ public void testPasswordRules() throws Exception {
         Assert.assertTrue(response.getBody().contains("xxx"));
 
         response = rh.executePutRequest("/_opendistro/_security/api/internalusers/ok1", "{\"backend_roles\":[\"my-backend-role\"],\"attributes\":{},\"password\":\"\"}", new Header[0]);
-        Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
+        Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, response.getStatusCode());
+
+        response = rh.executePutRequest("/_opendistro/_security/api/internalusers/non-empty_password", "{\"backend_roles\":[\"my-backend-role\"],\"attributes\":{},\"password\":\"Admin_123\"}", new Header[0]);
```
vrozov, line 454: The request creates new user. Please change it to update existing user "ok1".

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
vrozov: Will be also good to test request to create user with `"password": null`

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
cliu123: If `contentAsMap` is `null` or `password` field doesn't exist, there's no need to validate password at all.

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
cliu123: > Is there a test that adds user with empty password and no hash?

Line 451-452

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
cliu123: This intends to test user creation. Line 451-452 is to test updating existing user.

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
cliu123: Done.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
vrozov: How `contentAsMap` can be `null`? Does not check `password != null` supersede `contentAsMap.containsKey("password")`?

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
vrozov: User creation is tested using `addUserWithPassword()`, how this test is different?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
cliu123: I removed `password != null`.

src/test/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/UserApiTest.java
cliu123: I change it to update the existing user "ok1".

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
vrozov: Will it be better to remove `contentAsMap != null && contentAsMap.containsKey("password")`?

cliu123: If removing `contentAsMap != null && contentAsMap.containsKey("password")`, we'd need to keep `if (password != null && password.isEmpty()) {` at [Line 65](https://github.com/opendistro-for-elasticsearch/security/blob/1bbf8e2f6d81c5fd7f28a861ba2d56785c630405/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java#L65) and add `password != null` [Line 89](https://github.com/opendistro-for-elasticsearch/security/blob/1bbf8e2f6d81c5fd7f28a861ba2d56785c630405/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java#L89). Will this look better? If so, I'll update it.

vrozov: Replace `contentAsMap != null && contentAsMap.containsKey("password")` with 
```
String password = contentAsMap.get("password");
if (password != null) {
```

cliu123: Done.

vrozov (approved):

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/validation/CredentialsValidator.java
```diff
@@ -53,39 +54,44 @@ public boolean validate() {
...
                 final Map<String, Object> contentAsMap = XContentHelper.convertToMap(this.content, false, XContentType.JSON).v2();
-                if (contentAsMap != null && contentAsMap.containsKey("password")) {
-                    final String password = (String) contentAsMap.get("password");
-
-                    // Password can be null/empty for an existing user. Regex will validate password if present
-                    if (password != null && !password.isEmpty() && !regex.isEmpty() && !Pattern.compile("^"+regex+"$").matcher(password).matches()) {
-                        if(log.isDebugEnabled()) {
-                            log.debug("Regex does not match password");
-                        }
+                String password = (String) contentAsMap.get("password");
```
sujithvm, line 61: nit: use final same as original `final String password = (String) contentAsMap.get("password");`

sujithvm: Can `contentAsMap` be null? I see the null check is removed

cliu123: `final Map<String, Object> contentAsMap = XContentHelper.convertToMap(this.content, false, XContentType.JSON).v2();` prevents null `contentAsMap`.

sujithvm: Does `prevents null` mean it throws null pointer exception or returns empty map?

sujithvm (approved):

PR #836, opened 2020-11-10
Security configuration for reporting and notification plugins

akbhatta:
*Issue #, if available:*

*Description of changes:*
Added system index configuration for reporting and notification plugin
Added default roles configuration required for reporting functionality
Updated demo config and example with above system index configurations

### DIFF
--- a/securityconfig/elasticsearch.yml.example
+++ b/securityconfig/elasticsearch.yml.example
@@ -222,4 +222,4 @@ opendistro_security.audit.type: internal_elasticsearch
 # Enable system indices
 # opendistro_security.system_indices.enabled: true
 # Specify a list of indices to mark as system. These indices will only be visible / mutable by members of the above setting, in addition to needing permission to the index via a normal role.
-# opendistro_security.system_indices.indices: ['.opendistro-alerting-config', '.opendistro-ism-*']
+# opendistro_security.system_indices.indices: ['.opendistro-alerting-config', '.opendistro-ism-*', '.opendistro-reports-*', '.opendistro-notifications-*']
--- a/securityconfig/roles.yml
+++ b/securityconfig/roles.yml
@@ -61,3 +61,35 @@ anomaly_full_access:
         - 'indices_monitor'
         - 'indices:admin/aliases/get'
         - 'indices:admin/mappings/get'
+
+# Allows users to read and download Reports
+reports_instances_read_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/opendistro/reports/instance/list'
+    - 'cluster:admin/opendistro/reports/instance/get'
+    - 'cluster:admin/opendistro/reports/menu/download'
+
+# Allows users to read and download Reports and Report-definitions
+reports_read_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/opendistro/reports/definition/get'
+    - 'cluster:admin/opendistro/reports/definition/list'
+    - 'cluster:admin/opendistro/reports/instance/list'
+    - 'cluster:admin/opendistro/reports/instance/get'
+    - 'cluster:admin/opendistro/reports/menu/download'
+
+# Allows users to all Reports functionality
+reports_full_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/opendistro/reports/definition/create'
+    - 'cluster:admin/opendistro/reports/definition/update'
+    - 'cluster:admin/opendistro/reports/definition/on_demand'
+    - 'cluster:admin/opendistro/reports/definition/delete'
+    - 'cluster:admin/opendistro/reports/definition/get'
+    - 'cluster:admin/opendistro/reports/definition/list'
+    - 'cluster:admin/opendistro/reports/instance/list'
+    - 'cluster:admin/opendistro/reports/instance/get'
+    - 'cluster:admin/opendistro/reports/menu/download'
--- a/tools/install_demo_configuration.sh
+++ b/tools/install_demo_configuration.sh
@@ -373,7 +373,7 @@ echo "opendistro_security.enable_snapshot_restore_privilege: true" | $SUDO_CMD t
 echo "opendistro_security.check_snapshot_restore_write_privileges: true" | $SUDO_CMD tee -a "$ES_CONF_FILE" > /dev/null
 echo 'opendistro_security.restapi.roles_enabled: ["all_access", "security_rest_api_access"]' | $SUDO_CMD tee -a "$ES_CONF_FILE" > /dev/null
 echo 'opendistro_security.system_indices.enabled: true' | $SUDO_CMD tee -a "$ES_CONF_FILE" > /dev/null
-echo 'opendistro_security.system_indices.indices: [".opendistro-alerting-config", ".opendistro-alerting-alert*", ".opendistro-anomaly-results*", ".opendistro-anomaly-detector*", ".opendistro-anomaly-checkpoints", ".opendistro-anomaly-detection-state"]' | $SUDO_CMD tee -a "$ES_CONF_FILE" > /dev/null
+echo 'opendistro_security.system_indices.indices: [".opendistro-alerting-config", ".opendistro-alerting-alert*", ".opendistro-anomaly-results*", ".opendistro-anomaly-detector*", ".opendistro-anomaly-checkpoints", ".opendistro-anomaly-detection-state", ".opendistro-reports-*", ".opendistro-notifications-*"]' | $SUDO_CMD tee -a "$ES_CONF_FILE" > /dev/null
 
 #cluster.routing.allocation.disk.threshold_enabled
 if $SUDO_CMD grep --quiet -i "^cluster.routing.allocation.disk.threshold_enabled" "$ES_CONF_FILE"; then


src/main/resources/static_config/static_roles.yml
```diff
@@ -85,6 +85,8 @@ kibana_server:
   - "cluster:admin/xpack/monitoring*"
   - "indices:admin/template*"
   - "indices:data/read/scroll*"
+  - 'cluster:admin/opendistro/reports/*'
```
hardik-k-shah, line 88: Why kibana_server role needs this permission?

akbhatta: Kibana reporting plugin runs a background thread to poll job from ES reporting plugin. The Kibana executes the job in background and sends notification. It also updates status of the reports.

hardik-k-shah: And I believe we are planning to use kibanaserver user (which is mapped to kibana_server role) ? 

What exactly these two permissions are ?
'cluster:admin/opendistro/reports/*'
'cluster:admin/opendistro/notifications/*'

is there any other option to avoid using kibanaserver user?

hardik-k-shah: Also please keep formatting same and use double quote instead single.

securityconfig/elasticsearch.yml.example
```diff
@@ -215,11 +215,11 @@ opendistro_security.audit.type: internal_elasticsearch
 # Specify a list of roles a user must be member of to touch any protected index.
 # opendistro_security.protected_indices.roles: ['all_access']
 # Specify a list of indices to mark as protected. These indices will only be visible / mutable by members of the above setting, in addition to needing permission to the index via a normal role.
-# opendistro_security.protected_indices.indices: ['.opendistro-alerting-config', '.opendistro-ism-*']
+# opendistro_security.protected_indices.indices: ['.opendistro-alerting-config', '.opendistro-ism-*', '.opendistro-reports-*', '.opendistro-notifications-*']
```
skkosuri-amzn, line 218: This is example, but curious why you want to add reports to protected_indices?

securityconfig/roles.yml
```diff
@@ -61,3 +61,35 @@ anomaly_full_access:
         - 'indices_monitor'
         - 'indices:admin/aliases/get'
         - 'indices:admin/mappings/get'
+
+# Allows users to read and download Reports
+reports_instances_read_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/opendistro/reports/instance/get_all'
+    - 'cluster:admin/opendistro/reports/instance/get'
```
skkosuri-amzn, line 70: What the difference between`get` and `get_all` ? Why we have to two API?

src/main/resources/static_config/static_roles.yml
cliu123: > And I believe we are planning to use kibanaserver user (which is mapped to kibana_server role) for running those background job ?
> 
> What exactly these two permissions are ? Can you describe more in details?
> 'cluster:admin/opendistro/reports/_' 'cluster:admin/opendistro/notifications/_'
> 
> is there any other option to avoid using kibanaserver user?

Would it be better to add a new user and grant the permission to it?

akbhatta: Yes. kibanaserver is used for running background job.
cluster:admin/opendistro/notifications/* --> Send mail operation using notification plugin(currently there is only 1 operation).
cluster:admin/opendistro/reports/ --> all operations of reports plugin (currently we are using get_all, get, update, poll operations)

We are open to suggestion on using any system user for background operation. We cannot use the user role as until we pick the job we don't know which user is that job for.

Will update quite ' --> " and move before indices:admin

securityconfig/roles.yml
```diff
@@ -61,3 +61,35 @@ anomaly_full_access:
...
+    - 'cluster:admin/opendistro/reports/menu/download'
+
+# Allows users to read and download Reports and Report-definitions
+reports_read_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/opendistro/reports/definition/get'
+    - 'cluster:admin/opendistro/reports/definition/get_all'
+    - 'cluster:admin/opendistro/reports/instance/get_all'
+    - 'cluster:admin/opendistro/reports/instance/get'
```
skkosuri-amzn, line 80: What the difference between `get` and `get_all` ? Why we have to two API?

akbhatta: yes they are 2 different APIs. thinking again may be it makes sense to rename get_all to list

securityconfig/roles.yml
akbhatta: yes they are 2 different APIs. thinking again may be it makes sense to rename get_all to list

src/main/resources/static_config/static_roles.yml
akbhatta: is it preferred to explicitly add all used operations or use "*"?

securityconfig/roles.yml
skkosuri-amzn: Standard way is have action as `cluster:admin/opendistro/reports/instance/get` and API like `GET reports\{id}` gets single report `GET \reports` get all reports. You also can support filters as required.  This is how we do on Alerting: https://github.com/opendistro-for-elasticsearch/alerting/blob/master/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestGetDestinationsAction.kt

securityconfig/elasticsearch.yml.example
akbhatta: If system index is not enabled then could it be configured as protected?

securityconfig/roles.yml
akbhatta: How to support response uniformity with pagination support? Initially we had the same way but then changed to 2 separate APIs.

src/main/resources/static_config/static_roles.yml
hardik-k-shah: I would really prefer to use different user for background jobs instead piggybacking on kibanaserver user. 
You can have that user in reporting/notification configuration and if needed we can also add it into kibana config  

kibanaserver is just for kibana to run healthcheck and other calls for kibana UI. And hence kept it with very minimal set of permissions. 
Additionally, due to security threat associated with kibanaserver user, we want to keep permission very minimal to avoid any kibana user to gain benefit of kibanaserver user permission.  

Are reporting and notification plugins read 'kibanaserver' user and credential from kibana.yml and use it for background job? 
if yes, then can each plugin have their own user with its own default role with limited set of permission?  And can this user be defined in plugin configuration file ?  (I would prefer it into plugin yml file instead kibana.yml if possible)

securityconfig/elasticsearch.yml.example
hardik-k-shah: are these indices created only for plugin to consume (like alerting and ISM indices )? 
does any of the users including admin need direct access to this indices using ES APIs?

akbhatta: are these indices created only for plugin to consume (like alerting and ISM indices )? --> YES
does any of the users including admin need direct access to this indices using ES APIs? --> NO. users should not have direct access

hardik-k-shah: then these indices should be part of system indices and not protected indices. 

protected indices will be deprecated once all plugin move their indices to system indices.

skkosuri-amzn: Yes, protected_indices will be deprecated.

akbhatta: will remove.

src/main/resources/static_config/static_roles.yml
akbhatta: Are reporting and notification plugins read 'kibanaserver' user and credential from kibana.yml and use it for background job? --> NO. I believe kibana is already running with this user and no need to store/manage credential. Please let me know if my understanding is wrong.

src/main/resources/static_config/static_roles.yml
```diff
@@ -83,6 +83,10 @@ kibana_server:
   - "cluster_monitor"
   - "cluster_composite_ops"
   - "cluster:admin/xpack/monitoring*"
+  - "cluster:admin/opendistro/reports/instance/poll"
```
hardik-k-shah, line 86: Can you describe each of these permissions?

akbhatta: - "cluster:admin/opendistro/reports/instance/poll" --> Get scheduled report instance information (metadata from configuration index).
  - "cluster:admin/opendistro/reports/instance/get" --> get information of specific report instance
  - "cluster:admin/opendistro/reports/instance/update_status" --> change status of report instance from scheduled to success/failure
  - "cluster:admin/opendistro/notifications/send" --> send notification mail to email ids configured in report instance

securityconfig/roles.yml
```diff
@@ -61,3 +61,35 @@ anomaly_full_access:
         - 'indices_monitor'
         - 'indices:admin/aliases/get'
         - 'indices:admin/mappings/get'
+
+# Allows users to read and download Reports
+reports_instances_read_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/opendistro/reports/instance/list'
```
skkosuri-amzn, line 69: Whats the difference between instance `list` and `get`?

securityconfig/roles.yml
```diff
@@ -61,3 +61,35 @@ anomaly_full_access:
...
+    - 'cluster:admin/opendistro/reports/instance/get'
+    - 'cluster:admin/opendistro/reports/menu/download'
+
+# Allows users to read and download Reports and Report-definitions
+reports_read_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/opendistro/reports/definition/get'
+    - 'cluster:admin/opendistro/reports/definition/list'
```
skkosuri-amzn, line 78: similar,  definition `get` vs `list`?

securityconfig/roles.yml
```diff
@@ -61,3 +61,35 @@ anomaly_full_access:
...
+    - 'cluster:admin/opendistro/reports/instance/list'
+    - 'cluster:admin/opendistro/reports/instance/get'
+    - 'cluster:admin/opendistro/reports/menu/download'
+
+# Allows users to all Reports functionality
+reports_full_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/opendistro/reports/definition/create'
+    - 'cluster:admin/opendistro/reports/definition/update'
```
skkosuri-amzn, line 88: Curious, why did you separate `create` and `update`?

securityconfig/roles.yml
akbhatta: get -> get one item
list -> get all paginated response for items

securityconfig/roles.yml
akbhatta: similar

securityconfig/roles.yml
akbhatta: validation for create and update are different

securityconfig/roles.yml
skkosuri-amzn: example: https://restfulapi.net/http-methods/#get

securityconfig/roles.yml
skkosuri-amzn: reference : https://restfulapi.net/http-methods/#put

sujithvm (approved):

hardik-k-shah (approved):

Issue #718, opened 2020-09-19
Returns security exception when an invalid index name is used.


skkosuri-amzn:
When an index with invalid name is mentioned, instead of returning `invalid_index_name_exception` security plugin throws security_exception.
Example:
`curl -XGET 'https://localhost:9200/_##pdt_data/_search' -u admin:admin --insecure` Note the invalid chars in the index name.
Response:
`
{"error":{"root_cause":[{"type":"security_exception","reason":"Unexpected exception indices:admin/get"}],"type":"security_exception","reason":"Unexpected exception indices:admin/get"},"status":500}`

Its works fine: if we mention a missing index, it correctly throws: index_not_found_exception.

PR #865, opened 2020-12-02
Catch and respond invalid_index_name_exception when an index with invalid name is mentioned

cliu123:
*Issue #, if available:*
#718 

*Description of changes:*

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
@@ -355,6 +355,13 @@ public class OpenDistroSecurityFilter implements ActionFilter {
                 listener.onFailure(new ElasticsearchSecurityException(err, RestStatus.FORBIDDEN));
                 return;
             }
+        } catch (ElasticsearchException e) {
+            if (task != null) {
+                log.debug("Failed to apply filter. Task id: {} ({}). Action: {}", task.getId(), task.getDescription(), action, e);
+            } else {
+                log.debug("Failed to apply filter. Action: {}", action, e);
+            }
+            listener.onFailure(e);
         } catch (Throwable e) {
             log.error("Unexpected exception "+e, e);
             listener.onFailure(new ElasticsearchSecurityException("Unexpected exception " + action, RestStatus.INTERNAL_SERVER_ERROR));
--- a/src/test/java/com/amazon/opendistroforelasticsearch/security/IndexIntegrationTests.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/security/IndexIntegrationTests.java
@@ -30,6 +30,7 @@
 
 package com.amazon.opendistroforelasticsearch.security;
 
+import java.net.URLEncoder;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.TimeZone;
@@ -455,6 +456,17 @@ public class IndexIntegrationTests extends SingleClusterTest {
         Assert.assertEquals(HttpStatus.SC_OK, rh.executeGetRequest("*kibana/_search?pretty", encodeBasicHeader("aliastest", "nagilum")).getStatusCode());
         Assert.assertEquals(HttpStatus.SC_OK, rh.executeGetRequest(".ki*ana/_search?pretty", encodeBasicHeader("aliastest", "nagilum")).getStatusCode());
         Assert.assertEquals(HttpStatus.SC_OK, rh.executeGetRequest(".kibana/_search?pretty", encodeBasicHeader("aliastest", "nagilum")).getStatusCode());
+    }
+
+    @Test
+    public void testIndexResolveInvalidIndexName() throws Exception {
+        setup();
+        final RestHelper rh = nonSslRestHelper();
+
+        // invalid_index_name_exception should be thrown and responded when invalid index name is mentioned in requests.
+        HttpResponse res = rh.executeGetRequest(URLEncoder.encode("_##pdt_data/_search", "UTF-8"), encodeBasicHeader("ccsresolv", "nagilum"));
+        Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, res.getStatusCode());
+        Assert.assertTrue(res.getBody().contains("invalid_index_name_exception"));
     }
     
     @Test


src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/DlsFlsEvaluator.java
```diff
@@ -67,7 +68,7 @@ public DlsFlsEvaluator(Settings settings, ThreadPool threadPool) {
     }
 
     public PrivilegesEvaluatorResponse evaluate(final ActionRequest request, final ClusterService clusterService, final IndexNameExpressionResolver resolver, final Resolved requestedResolved, final User user,
-                                                final SecurityRoles securityRoles, final PrivilegesEvaluatorResponse presponse) {
+                                                final SecurityRoles securityRoles, final PrivilegesEvaluatorResponse presponse) throws InvalidIndexNameException {
```
vrozov, line 71: Is it necessary to add `InvalidIndexNameException` to `throws`?

cliu123: Removed the `throws` and caught the `RuntimeException` in `OpenDistroSecurityFilter`.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -355,7 +356,11 @@ public void onFailure(Exception e) {
                 listener.onFailure(new ElasticsearchSecurityException(err, RestStatus.FORBIDDEN));
                 return;
             }
-        } catch (Throwable e) {
+        }
```
vrozov, line 359: nit: please keep original code style: `} catch () {` on a single line.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -355,7 +356,11 @@ public void onFailure(Exception e) {
...
+        catch (InvalidIndexNameException err) {
```
vrozov, line 360: should it catch `ElasticsearchException` instead?

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -355,7 +356,11 @@ public void onFailure(Exception e) {
...
+            listener.onFailure(err);
```
vrozov, line 361: add debug logging of the exception.

src/test/java/com/amazon/opendistroforelasticsearch/security/IndexIntegrationTests.java
```diff
@@ -473,7 +473,13 @@ public void testCCSIndexResolve() throws Exception {
 
         res = rh.executeGetRequest("/*:.abc-6,.abc-6/_search", encodeBasicHeader("ccsresolv", "nagilum"));
         Assert.assertEquals(HttpStatus.SC_OK, res.getStatusCode());
+
         //TODO: Change for 25.0 to be forbidden (possible bug in ES regarding ccs wildcard)
+
+        // invalid_index_name_exception should be thrown and responded when invalid index name is mentioned in requests.
+        res = rh.executeGetRequest("_pdt_data/_search", encodeBasicHeader("ccsresolv", "nagilum"));
```
sujithvm, line 480: Can you add a test for the same index name in ticket `_##pdt_data` ?

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
cliu123: Done.

cliu123: Done.

src/test/java/com/amazon/opendistroforelasticsearch/security/IndexIntegrationTests.java
cliu123: I've tried that, but the `#` is identified as a fragment and caused  `java.lang.IllegalArgumentException: Illegal character in fragment at index`.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
cliu123: Done.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
cliu123: Done.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -355,6 +356,9 @@ public void onFailure(Exception e) {
                 listener.onFailure(new ElasticsearchSecurityException(err, RestStatus.FORBIDDEN));
                 return;
             }
+        } catch (ElasticsearchException err) {
+            log.debug(err);
```
vrozov, line 360: Will this log exception? Please provide sample output.

cliu123: Changed it to log exception. The output is: `Unexpected exception [_] InvalidIndexNameException[Invalid index name [_], must not start with '_'.]`

src/test/java/com/amazon/opendistroforelasticsearch/security/IndexIntegrationTests.java
sujithvm: Was this not the original intention of the ticket? `_pdt_data` is missing index which is already correctly throwing index not found exception?

cliu123: I agree that this it exactly the index name mentioned in the ticket. But `##` causes `java.net.URISyntaxException`at `at java.base/java.net.URI.create(URI.java:906)`. I hope [this post](https://stackoverflow.com/questions/20941947/java-net-uri-illegal-character-in-fragment-at-index-xxx-due-to-character) helps explain the reason. 
I've changed the index name in the test and added a `#`.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -77,6 +77,7 @@
 import org.elasticsearch.common.util.concurrent.ThreadContext.StoredContext;
 import org.elasticsearch.index.reindex.DeleteByQueryRequest;
 import org.elasticsearch.index.reindex.UpdateByQueryRequest;
+import org.elasticsearch.indices.InvalidIndexNameException;
```
vrozov, line 80: Is this still necessary?

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -355,6 +356,9 @@ public void onFailure(Exception e) {
                 listener.onFailure(new ElasticsearchSecurityException(err, RestStatus.FORBIDDEN));
                 return;
             }
+        } catch (ElasticsearchException err) {
+            log.error("Unexpected exception "+ err, err);
```
vrozov, line 360: This is an anti-pattern. There is no point to log exception message twice.

src/test/java/com/amazon/opendistroforelasticsearch/security/IndexIntegrationTests.java
vrozov: Please check ES behavior without security plugin being installed.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
cliu123: I followed the existing code style of [this line](https://github.com/opendistro-for-elasticsearch/security/blob/f3e474c1783e7afc592dcab4f6b72806abc47281/src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java#L370). Does this look better now?

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
cliu123: I followed the existing code style of [this line](https://github.com/opendistro-for-elasticsearch/security/blob/f3e474c1783e7afc592dcab4f6b72806abc47281/src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java#L370) Does this look better now?

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
cliu123: Removed.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
vrozov: When logging an exception, it is always better to have stack trace logged as well. It is also necessary to avoid premature string concatenation and rely on place holder instead. I don't think that "Unexpected" is the right message here. It is expected for the wrong parameters provided (wrong index name). I also think that logging it at error level is not necessary and debug (or warning) should be sufficient.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -355,6 +355,9 @@ public void onFailure(Exception e) {
                 listener.onFailure(new ElasticsearchSecurityException(err, RestStatus.FORBIDDEN));
                 return;
             }
+        } catch (ElasticsearchException err) {
```
vrozov, line 358: nit: `err` -> `e`

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
cliu123: Changed the exception logs.

src/test/java/com/amazon/opendistroforelasticsearch/security/IndexIntegrationTests.java
cliu123: Without security plugin being installed, ES behaves in the way that we expect after this fix. It responds a invalid_index_name_exception to `curl -XGET 'http://localhost:9200/_##pdt_data/_search'`.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
cliu123: Done

src/test/java/com/amazon/opendistroforelasticsearch/security/IndexIntegrationTests.java
sujithvm: Is it possible to add this in the tests?

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -355,6 +355,9 @@ public void onFailure(Exception e) {
                 listener.onFailure(new ElasticsearchSecurityException(err, RestStatus.FORBIDDEN));
                 return;
             }
+        } catch (ElasticsearchException e) {
+            log.debug("Failed to apply filter. Task id: {} ({}). Action: {}", task.getId(), task.getDescription(), action, e);
```
vrozov, line 359: May `task` be `null`?

cliu123: `task` is not going to be `null` here. Otherwise, other usages of `task` prior to this line in this method will simply complain null pointer exception.

src/test/java/com/amazon/opendistroforelasticsearch/security/IndexIntegrationTests.java
cliu123: @sujithvm As I mentioned in the previous comments, using `##` fails the unit test due to `java.lang.IllegalArgumentException: Illegal character in fragment at index`, which is not related to ES or security plugin.

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
sujithvm: Where previously are we accessing a member of task? null check should be handled in the functions it is passed to

src/test/java/com/amazon/opendistroforelasticsearch/security/IndexIntegrationTests.java
cliu123: Applied url encoding to the test. Thanks @sujithvm!

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
cliu123: A lot of usages of task in this function, such as [here](https://github.com/opendistro-for-elasticsearch/security/blob/f382a820ad8512c878a1eaa85bf604e2d4aa4b7d/src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java#L319)

sujithvm: We are not accessing a member of the object there. Elasticsearch proceed function will handle the null task. Here we are directly accessing a class member `task.getId()`, `task.getDescription()`

cliu123: Yes, this is indeed the only place where the members of task are accesses. I've added the null check.

vrozov (dismissed):

src/test/java/com/amazon/opendistroforelasticsearch/security/IndexIntegrationTests.java
```diff
@@ -473,7 +474,13 @@ public void testCCSIndexResolve() throws Exception {
 
         res = rh.executeGetRequest("/*:.abc-6,.abc-6/_search", encodeBasicHeader("ccsresolv", "nagilum"));
         Assert.assertEquals(HttpStatus.SC_OK, res.getStatusCode());
+
         //TODO: Change for 25.0 to be forbidden (possible bug in ES regarding ccs wildcard)
+
+        // invalid_index_name_exception should be thrown and responded when invalid index name is mentioned in requests.
+        res = rh.executeGetRequest(URLEncoder.encode("_##pdt_data/_search", "UTF-8"), encodeBasicHeader("ccsresolv", "nagilum"));
+        Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, res.getStatusCode());
+        Assert.assertTrue(res.getBody().contains("invalid_index_name_exception"));
```
sujithvm, line 483: Is it possible to create separate test as it is not related to cross cluster search?

cliu123: Done

sujithvm (dismissed):

src/main/java/com/amazon/opendistroforelasticsearch/security/filter/OpenDistroSecurityFilter.java
```diff
@@ -355,6 +355,14 @@ public void onFailure(Exception e) {
                 listener.onFailure(new ElasticsearchSecurityException(err, RestStatus.FORBIDDEN));
                 return;
             }
+        } catch (ElasticsearchException e) {
+            if (task != null) {
```
vrozov, line 359: nit: please follow code style of the file. Even though we don't enforce code style across files in the project, it will be better to keep it the same at least within file/function. In this method `} else {` should be on the same line as `}`.

cliu123: Done.

vrozov (approved):

sujithvm (approved):

PR #866, opened 2020-12-02
Use reflection to get reduceOrder, termBytes and format due to java.lang.IllegalAccessError

vrozov:
`StringTermsGetter` can not access fields of `StringTerms` and `StringTerms.Bucket` as it is loaded by a different class loader.


src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/DlsFlsValveImpl.java
```diff
@@ -253,4 +260,58 @@ public void accept(StringTerms.Bucket bucket) {
...
         }
     }
+
+    private static class StringTermsGetter {
+        private static final Field REDUCE_ORDER = getField(InternalTerms.class, "reduceOrder");
+        private static final Field TERM_BYTES = getField(StringTerms.Bucket.class, "termBytes");
+        private static final Field FORMAT = getField(InternalTerms.Bucket.class, "format");
+
+        private StringTermsGetter() {
+        }
```
palashhedau, line 270: nit: single line {}?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/DlsFlsValveImpl.java
```diff
@@ -253,4 +260,58 @@ public void accept(StringTerms.Bucket bucket) {
...
+            return AccessController.doPrivileged((PrivilegedAction<Field>) () -> getFieldPrivileged(cls, name));
+        }
+
+        private static <T, C> T getFieldValue(Field field, C c) {
+            try {
+                return (T)field.get(c);
+            } catch (IllegalArgumentException | IllegalAccessException e) {
+                log.error("Exception while getting value {} of class {}", field.getName(), c.getClass().getSimpleName(), e);
+                if (e instanceof RuntimeException) {
+                    throw (RuntimeException) e;
```
palashhedau, line 298: Cant it be handled the same way as "throw new RuntimeException(e)" ?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/DlsFlsValveImpl.java
vrozov: I don't think it is a common style to have single line `{}`.

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/DlsFlsValveImpl.java
vrozov: There is no need to wrap `RuntimeException` in another `RuntimeException`

palashhedau (approved):

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/DlsFlsValveImpl.java
```diff
@@ -253,4 +260,58 @@ public void accept(StringTerms.Bucket bucket) {
             this.bucket = bucket;
         }
     }
+
+    private static class StringTermsGetter {
+        private static final Field REDUCE_ORDER = getField(InternalTerms.class, "reduceOrder");
```
debjanibnrj, line 265: Curious why we are using the InternalTerms.class instead of the previous StringTerms.class

debjanibnrj: Nvm, I see that StringTerms is a child class of InternalTerms.

debjanibnrj (approved):

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/DlsFlsValveImpl.java
vrozov: For `getDeclaredField()` to find the field it needs to be called for the class where the field is declared. In case of `reduceOrder` it is declared in `InternalTerms.class` and is accessible from subclasses (`StringTerms.class`)

PR #876, opened 2020-12-04
Fix AuthCredentials equality

### LINKED ISSUE #873: Caching authenticated users broken with anonymous auth
The backend registry tries to cache the authenticated `User` based on the `AuthCredentials`. When using the `HTTPClientCertAuthenticator`, the `AuthCredentials.password` is null. This causes the equality check to fail for equal `AuthCredentials.equals` even though it should not, because the function returns false if the object has a null password field: https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/user/AuthCredentials.java#L183.

This results in _many_ calls to our authentication backend, which is unfortunately a blocking call to LDAP and synchronized on the cache.

nicot:
Fixes #873.

Description of changes:
This PR fixes the `AuthCredentials` class to correctly consider two `AuthCredentials` equal if they have the same username and both instances have null passwords. Before this PR, if both instances had null passwords, they would be considered not equal, causing them to never be cached. This fixes caching for backends using anonymous authentication.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/user/AuthCredentials.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/user/AuthCredentials.java
@@ -37,6 +37,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import org.elasticsearch.ElasticsearchSecurityException;
@@ -175,19 +176,17 @@ public final class AuthCredentials {
     public boolean equals(Object obj) {
         if (this == obj)
             return true;
-        if (obj == null)
+        if (obj == null || getClass() != obj.getClass()) {
             return false;
-        if (getClass() != obj.getClass())
-            return false;
+        }
+
         AuthCredentials other = (AuthCredentials) obj;
-        if (internalPasswordHash == null || other.internalPasswordHash == null || !MessageDigest.isEqual(internalPasswordHash, other.internalPasswordHash))
-            return false;
-        if (username == null) {
-            if (other.username != null)
-                return false;
-        } else if (!username.equals(other.username))
-            return false;
-        return true;
+        return Objects.equals(username, other.username)
+            && Arrays.equals(password, other.password)
+            && Objects.equals(nativeCredentials, other.nativeCredentials)
+            && Objects.equals(backendRoles, other.backendRoles)
+            && MessageDigest.isEqual(internalPasswordHash, other.internalPasswordHash)
+            && Objects.equals(attributes, other.attributes);
     }
 
     @Override
--- a/src/test/java/com/amazon/opendistroforelasticsearch/security/cache/CachingTest.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/security/cache/CachingTest.java
@@ -58,7 +58,7 @@ public class CachingTest extends SingleClusterTest{
 
         Assert.assertEquals(3, DummyHTTPAuthenticator.getCount());
         Assert.assertEquals(1, DummyAuthorizer.getCount());
-        Assert.assertEquals(3, DummyAuthenticationBackend.getAuthCount());
+        Assert.assertEquals(1, DummyAuthenticationBackend.getAuthCount());
         Assert.assertEquals(0, DummyAuthenticationBackend.getExistsCount());
     }
 
@@ -103,7 +103,7 @@ public class CachingTest extends SingleClusterTest{
 
         Assert.assertEquals(4, DummyHTTPAuthenticator.getCount());
         Assert.assertEquals(3, DummyAuthorizer.getCount());
-        Assert.assertEquals(4, DummyAuthenticationBackend.getAuthCount());
+        Assert.assertEquals(1, DummyAuthenticationBackend.getAuthCount());
         Assert.assertEquals(2, DummyAuthenticationBackend.getExistsCount());
 
     }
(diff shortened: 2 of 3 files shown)


hardik-k-shah:
@nicot  Thanks for your contribution. Can you please verify why CI/CD is failing and fix tests if needed?

nicot:
I've updated the tests to reflect the updated logic. This new caching seems to respect the comments on AuthenticationBackend better. I _think_ that the new counts in the test are correct, and the old ones were a workaround, but I'm not sure, so it would be great to confirm my understanding here.

src/main/java/com/amazon/opendistroforelasticsearch/security/user/AuthCredentials.java
```diff
@@ -175,19 +176,16 @@ public int hashCode() {
...
-            if (other.username != null)
-                return false;
-        } else if (!username.equals(other.username))
-            return false;
-        return true;
+        return Objects.equals(username, other.username)
+            && Objects.equals(nativeCredentials, other.nativeCredentials)
+            && Objects.equals(backendRoles, other.backendRoles)
+            && MessageDigest.isEqual(internalPasswordHash, other.internalPasswordHash)
+            && Objects.equals(attributes, other.attributes);
```
cliu123, line 188: Add `&& Objects.equals(password, other.password)` here?

src/test/java/com/amazon/opendistroforelasticsearch/security/user/AuthCredentialsTests.java
```diff
@@ -0,0 +1,19 @@
+package com.amazon.opendistroforelasticsearch.security.user;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class AuthCredentialsTests {
+  @Test
+  public void testEquality() {
```
cliu123, line 8: Should we add 2 test cases?
1. `object.internalPasswordHash == null` && `other. internalPasswordHash == null` ==> equals() returns true
2. `object.internalPasswordHash == null` && `other. internalPasswordHash == <anything other than null>` returns false.

src/test/java/com/amazon/opendistroforelasticsearch/security/cache/CachingTest.java
```diff
@@ -58,7 +58,7 @@ public void testRestCaching() throws Exception {
 
         Assert.assertEquals(3, DummyHTTPAuthenticator.getCount());
         Assert.assertEquals(1, DummyAuthorizer.getCount());
-        Assert.assertEquals(3, DummyAuthenticationBackend.getAuthCount());
+        Assert.assertEquals(1, DummyAuthenticationBackend.getAuthCount());
```
cliu123, line 61: Why is this change necessary? It doesn't look related to your change.

src/test/java/com/amazon/opendistroforelasticsearch/security/cache/CachingTest.java
```diff
@@ -103,7 +103,7 @@ public void testRestCachingWithImpersonation() throws Exception {
 
         Assert.assertEquals(4, DummyHTTPAuthenticator.getCount());
         Assert.assertEquals(3, DummyAuthorizer.getCount());
-        Assert.assertEquals(4, DummyAuthenticationBackend.getAuthCount());
+        Assert.assertEquals(1, DummyAuthenticationBackend.getAuthCount());
```
cliu123, line 106: Why is this change necessary? It doesn't look related to your change.

src/test/java/com/amazon/opendistroforelasticsearch/security/cache/CachingTest.java
nicot: My change fixes caching auth credentials, so the authentication backend is hit less. That's how I found this bug in the first place.

src/test/java/com/amazon/opendistroforelasticsearch/security/cache/CachingTest.java
nicot: Same here, caching auth credentials was broken before because equality wasn't working on the key. Now that it is fixed the calls to the authentication backend are properly cached and the backend is hit less.

src/main/java/com/amazon/opendistroforelasticsearch/security/user/AuthCredentials.java
nicot: The password is always filled with nulls and then set to null in the constructor, and the internalPasswordHash is used instead. We can check it here though if you prefer.

cliu123: Is it possible to be non-null values in any scenarios? If so, I'd add a check for it.

nicot: It is not possible, but I can still add the check, is that what you'd prefer?

cliu123: Yes, let's add the check. Thanks!

cliu123:
Meanwhile, could you kindly rebase/squash your commits and only keep a single commit with the message same as your PR title? That would help us generate release notes up releases. Thanks!

cliu123 (approved):

cliu123:
@nicot  Thanks for contributing!

src/test/java/com/amazon/opendistroforelasticsearch/security/user/AuthCredentialsTests.java
cliu123: @nicot Can you add these 2 test cases?

nicot: A user password will be null if we set the `backendRoles`.

> object.internalPasswordHash == null && other.internalPasswordHash == null ==> equals() returns true

This is the test case on line 9 — the users have backendRoles, so the password is null for both.


> object.internalPasswordHash == null && other.internalPasswordHash == <non-null value> ==> equals() returns false.

This is the test case on line 28 — one user has backendRoles and one has a password, and they are not equal.

palashhedau (approved):

PR #886, opened 2020-12-10
Optimize creating new collection objects in IndexResolverReplacer

sujithvm:
*Issue #, if available:*

*Description of changes:*

- Currently whenever we access [Resolved class members](https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java#L383), we return a new unmodified list creating new objects. Some of these class functions are accessed a number of times. 

- Interchanging list and array in couple of places for certain functions. 

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
@@ -118,13 +118,13 @@ public class IndexResolverReplacer {
         this.clusterInfoHolder = clusterInfoHolder;
     }
 
-    private static boolean isAllWithNoRemote(final String... requestedPatterns) {
+    private static boolean isAllWithNoRemote(final String[] requestedPatterns) {
         final List<String> patterns = requestedPatterns == null ? null : Arrays.asList(requestedPatterns);
         if (IndexNameExpressionResolver.isAllIndices(patterns)) {
             return true;
         }
         if (patterns.size() == 1) {
-            final String pattern = patterns.iterator().next();
+            final String pattern = requestedPatterns[0];
             return pattern == null || pattern.equals("*");
         }
         return false;
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
@@ -100,14 +100,12 @@ import com.amazon.opendistroforelasticsearch.security.securityconf.DynamicConfig
 import com.amazon.opendistroforelasticsearch.security.support.SnapshotRestoreHelper;
 import com.amazon.opendistroforelasticsearch.security.support.WildcardMatcher;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 
 import static org.elasticsearch.cluster.metadata.IndexAbstraction.Type.ALIAS;
 
 public class IndexResolverReplacer {
 
-    private static final Set<String> NULL_SET = new HashSet<>(Collections.singleton(null));
     private final Logger log = LogManager.getLogger(this.getClass());
     private final IndexNameExpressionResolver resolver;
     private final ClusterService clusterService;
@@ -120,42 +118,26 @@ public class IndexResolverReplacer {
         this.clusterInfoHolder = clusterInfoHolder;
     }
 
-    private static final boolean isAllWithNoRemote(final String... requestedPatterns) {
-
-        final List<String> patterns = requestedPatterns==null?null:Arrays.asList(requestedPatterns);
-
-        if(IndexNameExpressionResolver.isAllIndices(patterns)) {
+    private static boolean isAllWithNoRemote(final String... requestedPatterns) {
+        final List<String> patterns = requestedPatterns == null ? null : Arrays.asList(requestedPatterns);
+        if (IndexNameExpressionResolver.isAllIndices(patterns)) {
             return true;
         }
-
-        if(patterns.size() == 1 && patterns.contains("*")) {
-            return true;
+        if (patterns.size() == 1) {
+            final String pattern = patterns.iterator().next();
+            return pattern == null || pattern.equals("*");
         }
-
-        if(new HashSet<String>(patterns).equals(NULL_SET)) {
-            return true;
-        }
-
         return false;
     }
 
-    private static final boolean isLocalAll(final String... requestedPatterns) {
-
-        final List<String> patterns = requestedPatterns==null?null:Arrays.asList(requestedPatterns);
-
-        if(IndexNameExpressionResolver.isAllIndices(patterns)) {
+    private static boolean isLocalAll(final Collection<String> patterns) {
+        if (IndexNameExpressionResolver.isAllIndices(patterns)) {
             return true;
         }
-
-        if(patterns.contains("_all")) {
+        if (patterns.contains("_all")) {
             return true;
         }
-
-        if(new HashSet<String>(patterns).equals(NULL_SET)) {
-            return true;
-        }
-
-        return false;
+        return patterns.size() == 1 && patterns.iterator().next() == null;
     }
 
     private Resolved resolveIndexPatterns(final IndicesOptions indicesOptions, final boolean enableCrossClusterResolution, final String... requestedPatterns0) {
@@ -173,6 +155,7 @@ public class IndexResolverReplacer {
 
         Set<String> remoteIndices;
         final List<String> localRequestedPatterns = new ArrayList<>(Arrays.asList(requestedPatterns0));
+        final boolean isLocalAll = isLocalAll(localRequestedPatterns);
 
         final RemoteClusterService remoteClusterService = OpenDistroSecurityPlugin.GuiceHolder.getRemoteClusterService();
 
@@ -205,10 +188,10 @@ public class IndexResolverReplacer {
             remoteIndices = Collections.emptySet();
         }
 
-        final Set<String> matchingAliases;
-        Set<String> matchingAllIndices;
+        final Collection<String> matchingAliases;
+        Collection<String> matchingAllIndices;
 
-        if (isLocalAll(requestedPatterns0)) {
+        if (isLocalAll) {
             if (log.isTraceEnabled()) {
                 log.trace(Arrays.toString(requestedPatterns0) + " is an LOCAL ALL pattern");
             }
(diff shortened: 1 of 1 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -120,42 +119,26 @@ public IndexResolverReplacer(IndexNameExpressionResolver resolver, ClusterServic
...
-        }
-
-        if(patterns.size() == 1 && patterns.contains("*")) {
+    private static boolean isAllWithNoRemote(final String... requestedPatterns) {
+        final List<String> patterns = requestedPatterns == null ? null : Arrays.asList(requestedPatterns);
+        if (IndexNameExpressionResolver.isAllIndices(patterns)) {
             return true;
         }
-
-        if(new HashSet<String>(patterns).equals(NULL_SET)) {
```
sujithvm, line 135: Removing creating new hash set with single element null value

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -120,42 +119,26 @@ public IndexResolverReplacer(IndexNameExpressionResolver resolver, ClusterServic
...
+        if (IndexNameExpressionResolver.isAllIndices(patterns)) {
             return true;
         }
-
-        if(patterns.contains("_all")) {
+        if (patterns.contains("_all")) {
             return true;
         }
-
-        if(new HashSet<String>(patterns).equals(NULL_SET)) {
```
sujithvm, line 154: Same as above

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -351,53 +333,61 @@ public Resolved resolveRequest(final Object request) {
...
-                return true;
-            }
+        public Resolved(final StreamInput in) throws IOException {
+            this(in.readList(StreamInput::readString),
+                    in.readList(StreamInput::readString),
+                    in.readList(StreamInput::readString),
+                    in.readList(StreamInput::readString));
+        }
 
-            return aliases.contains("*") && allIndices.contains("*") && types.contains("*");
```
sujithvm, line 380: Removed `types.contains("*")` check as it always is true as types is defined as `ImmutableSet.of("*");`

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -351,53 +333,61 @@ public Resolved resolveRequest(final Object request) {
...
+            return isLocalAll;
         }
 
         public Set<String> getAliases() {
-            return Collections.unmodifiableSet(aliases);
+            return aliases;
         }
 
         public Set<String> getAllIndices() {
-            return Collections.unmodifiableSet(allIndices);
```
sujithvm, line 388: This is excessively used and is returning new set everytime

sujithvm: https://github.com/opendistro-for-elasticsearch/security/search?q=getAllIndices

src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -298,13 +298,6 @@ public PrivilegesEvaluatorResponse evaluate(final User user, String action0, fin
                             && !requestedResolved.getAllIndices().isEmpty()
                             ) {
 
-                        if(requestedResolved.getAllIndices().isEmpty()) {
```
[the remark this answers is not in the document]
vrozov: Should the condition be outside of the `if` statement? I am not so much concerned with removing the redundant check here as that it was intended to be in a different place.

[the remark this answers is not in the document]
sujithvm: I didn't intend to make any functional changes in this PR. Removed as the condition will never be executed. I can remove this diff and raise a separate PR if that helps. 

Slightly confused if condition must be removed here or move this block afterwards. 
```
if (dnfofEnabled
    && (action0.startsWith("indices:data/read/"))	                            
    && !requestedResolved.getAllIndices().isEmpty()	                           
```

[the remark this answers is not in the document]
vrozov: Let's do this change in a separate PR.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -120,42 +119,26 @@ public IndexResolverReplacer(IndexNameExpressionResolver resolver, ClusterServic
...
+        final List<String> patterns = requestedPatterns == null ? null : Arrays.asList(requestedPatterns);
+        if (IndexNameExpressionResolver.isAllIndices(patterns)) {
             return true;
         }
-
-        if(new HashSet<String>(patterns).equals(NULL_SET)) {
-            return true;
+        if (patterns.size() == 1) {
+            final String pattern = patterns.iterator().next();
+            return pattern == null || pattern.equals("*");
```
vrozov, line 129: When (use case) `pattern` will be `null`?

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -172,7 +155,8 @@ private Resolved resolveIndexPatterns(final IndicesOptions indicesOptions, final
         }
 
         Set<String> remoteIndices;
-        final List<String> localRequestedPatterns = new ArrayList<>(Arrays.asList(requestedPatterns0));
+        final List<String> localRequestedPatterns = Lists.newArrayList(requestedPatterns0);
```
vrozov, line 158: `Lists.newArrayList()` is likely to be deprecated in favor of the original code.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
sujithvm: Not sure of the use case / scenario where it may be null but looks like the check was specifically put in place hence retained it. Can any changes around this be done in separate PR ?

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -120,42 +118,26 @@ public IndexResolverReplacer(IndexNameExpressionResolver resolver, ClusterServic
...
-    private static final boolean isAllWithNoRemote(final String... requestedPatterns) {
-
-        final List<String> patterns = requestedPatterns==null?null:Arrays.asList(requestedPatterns);
-
-        if(IndexNameExpressionResolver.isAllIndices(patterns)) {
-            return true;
-        }
-
-        if(patterns.size() == 1 && patterns.contains("*")) {
+    private static boolean isAllWithNoRemote(final String... requestedPatterns) {
```
vrozov, line 121:
```suggestion
    private static boolean isAllWithNoRemote(final String[] requestedPatterns) {
```

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -120,42 +118,26 @@ public IndexResolverReplacer(IndexNameExpressionResolver resolver, ClusterServic
...
+        final List<String> patterns = requestedPatterns == null ? null : Arrays.asList(requestedPatterns);
+        if (IndexNameExpressionResolver.isAllIndices(patterns)) {
             return true;
         }
-
-        if(new HashSet<String>(patterns).equals(NULL_SET)) {
-            return true;
+        if (patterns.size() == 1) {
+            final String pattern = patterns.iterator().next();
```
vrozov, line 127:
```suggestion
            final String pattern = requestedPatterns[0];
```

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
vrozov: It is strange to get an array with a single element that is `null` instead of an empty array. As we are touching code here, I'd like to understand what is the use case before introducing changes.

sujithvm: I don't want to introduce any logical changes here. The logic before and after the code change is the same. 

There may be other scenarios with other requests too and this check seems to be specifically put in place hence don't want to remove it 

Here is a use-case
- SingleShardRequest [indices](https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java#L705) where index can be [nullable](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/action/support/single/shard/SingleShardRequest.java#L81) 

- Checking request [here](https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java#L271) which can have a single null element in array.

vrozov: The code for `SingleShardRequest` looks buggy to me. If `SingleShardRequest` index is not `null`, it will add it twice to the array, if it is `null`, it creates an array with single element that is `null` instead of an empty array or `null`.

I don't suggest to fix `SingleShardRequest` code in this PR, but it will be better to fix `SingleShardRequest` first in a separate PR.

sujithvm: Already raised this PR https://github.com/opendistro-for-elasticsearch/security/pull/896

vrozov: Likely that will be indication of another bug similar to one fixed in #896.

sujithvm:
Closing PR as intent/scope is changed and has dependency on #896 . Will remove diff around `isAllWithNoRemote`, `isLocalAll`.

PR #889, opened 2020-12-11
 Throwing ElasticsearchException if subject key is not found

debjanibnrj:
*Issue #, if available:* N/A

*Description of changes:*
This PR contains 2 changes

* Throwing ElasticsearchException if subject == null. If subject is null that means we have not been able to find the username from the JWT token. Throwing an exception will make this more explicit

*  In HTTPSamlAuthenticator am using SAML roles_key and subject_key as the JWT roles_key and subject_key. This will ensure that while logging/retrieving subject_key and roles_key we are using the SAML key for SAML auth instead of the JWT key. 

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
+++ b/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
@@ -120,7 +120,7 @@ public abstract class AbstractHTTPJwtAuthenticator implements HTTPAuthenticator 
 
         if (subject == null) {
             log.error("No subject found in JWT token");
-            return null;
+            throw new ElasticsearchSecurityException("No subject found in JWT token", RestStatus.FORBIDDEN);
         }
 
         final String[] roles = extractRoles(claims);
--- a/src/test/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticatorTest.java
+++ b/src/test/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticatorTest.java
@@ -37,10 +37,12 @@ import javax.net.ssl.KeyManagerFactory;
 import com.amazon.opendistroforelasticsearch.security.DefaultObjectMapper;
 import org.apache.cxf.rs.security.jose.jws.JwsJwtCompactConsumer;
 import org.apache.cxf.rs.security.jose.jwt.JwtToken;
+import org.elasticsearch.ElasticsearchException;
 import org.elasticsearch.common.bytes.BytesArray;
 import org.elasticsearch.common.bytes.BytesReference;
 import org.elasticsearch.common.io.stream.BytesStreamOutput;
 import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.common.util.concurrent.ThreadContext;
 import org.elasticsearch.common.xcontent.XContentBuilder;
 import org.elasticsearch.common.xcontent.XContentType;
 import org.elasticsearch.rest.RestChannel;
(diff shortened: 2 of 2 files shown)


src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
```diff
@@ -120,7 +120,7 @@ private AuthCredentials extractCredentials0(final RestRequest request) throws El
 
         if (subject == null) {
             log.error("No subject found in JWT token");
-            return null;
+            throw new ElasticsearchSecurityException("No subject found in JWT token", RestStatus.FORBIDDEN);
```
cliu123, line 123: Do we need tests for this change?

src/main/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticator.java
```diff
@@ -339,11 +339,13 @@ private Settings createJwtAuthenticatorSettings(Settings settings) {
         settingsBuilder.put(jwtSettings);
 
         if (jwtSettings.get("roles_key") == null && settings.get("roles_key") != null) {
```
palashhedau, line 341: What happen when roles_key is null? I know its not a part of this CR, so just wanted to check?

debjanibnrj: if roles_key is null we default to using "roles"

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
debjanibnrj: Added a unit test called "wrongSubjectKeyTest" below

src/main/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticator.java
```diff
@@ -339,11 +339,13 @@ private Settings createJwtAuthenticatorSettings(Settings settings) {
...
-            settingsBuilder.put("roles_key", "roles");
+            settingsBuilder.put("roles_key", settings.get("roles_key"));
```
palashhedau, line 342: What was the issue here again? 
Looks like this settings are used to extract details from a JWT token. 

What was the behavior before and what it is now?

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
cliu123: Should we verify that the error gets thrown out in the test?

src/main/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticator.java
palashhedau: so basically roles_key is clear, if nothing is specified, we don't use roles from the token/SAMLResponse.

src/main/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticator.java
debjanibnrj: By default when we're using SAML we are settings both the JWT roles_key and subject_key and the SAML roles_key and subject_key. JWT roles_key and subject_key are not used while trying to decode the SAML Response. 

One issue is when we are logging SAML errors we are using JWT roles_key and subject_key which is causing confusion. That is why I am explicitly setting the JWT roles_key value to the SAML roles_key for all Saml related calls . (https://github.com/opendistro-for-elasticsearch/security/blob/master/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java)

palashhedau (dismissed):

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
vrozov, line 123: Other `HTTPAuthenticator` do not raise `ElasticsearchSecurityException` when they can't validate credentials. Is it possible that authentication header should be processed by another `HTTPAuthenticator`?

vrozov:
The changes seems to be unrelated to each other. Please split them into 2 PR.

PR #895, opened 2020-12-14
Optimize debug log enable check

sujithvm:

 
Refactoring
 
 2. __Github Issue # or roaddmap entry, if available:__
N/A

- `log.isDebugEnabled()` and `log.isTraceEnabled()` is called multiple times in the same request path or method.
-  Each of the log function internally checks if it is enabled or not but enabling debug/trace logs is generally a transient change for debugging.
- Removed some debug logs which are in addition to `log.warn` as warn messages will always be shown in addition to debug messages
- Fixed string formatting of log messages 

- Reused computed variables within same method context

 5. __What is the old behavior before changes and new behavior after changes?__ (_Please add any example/logs/screenshot if avaialble_)
- No changes in behavior. 

 6. __Testing done:__ (_Please provide details of testing done: Unit testing, integration testing and manaual testing_) 
 No functional changes. Existing automated tests are sufficient
 
 
 7. __TO-DOs, if any:__ (_Please describe pending items and provide github issues# for each of them)
N/A

No

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/ldap/backend/LDAPAuthorizationBackend.java
+++ b/src/main/java/com/amazon/dlic/auth/ldap/backend/LDAPAuthorizationBackend.java
@@ -239,7 +239,7 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
                         }
                     });
                 } catch (PrivilegedActionException e) {
-                    log.warn("Unable to restore classloader because of " + e.getException(), e.getException());
+                    log.warn("Unable to restore classloader because of ", e);
                 }
             }
         }
@@ -257,9 +257,11 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
         Connection connection = null;
         Exception lastException = null;
 
+        final boolean isDebugEnabled = log.isDebugEnabled();
+        final boolean isTraceEnabled = log.isTraceEnabled();
         for (String ldapHost : ldapHosts) {
 
-            if (log.isTraceEnabled()) {
+            if (isTraceEnabled) {
                 log.trace("Connect to {}", ldapHost);
             }
 
@@ -278,7 +280,7 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
                 final ConnectionConfig config = new ConnectionConfig();
                 config.setLdapUrl("ldap" + (enableSSL ? "s" : "") + "://" + split[0] + ":" + port);
 
-                if (log.isTraceEnabled()) {
+                if (isTraceEnabled) {
                     log.trace("Connect to {}", config.getLdapUrl());
                 }
 
@@ -287,7 +289,7 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
                 final String bindDn = settings.get(ConfigConstants.LDAP_BIND_DN, null);
                 final String password = settings.get(ConfigConstants.LDAP_PASSWORD, null);
 
-                if (log.isDebugEnabled()) {
+                if (isDebugEnabled) {
                     log.debug("bindDn {}, password {}", bindDn,
                             password != null && password.length() > 0 ? "****" : "<not set>");
                 }
@@ -299,7 +301,7 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
                 final boolean enableClientAuth = settings.getAsBoolean(ConfigConstants.LDAPS_ENABLE_SSL_CLIENT_AUTH,
                         ConfigConstants.LDAPS_ENABLE_SSL_CLIENT_AUTH_DEFAULT);
 
-                if (log.isDebugEnabled()) {
+                if (isDebugEnabled) {
                     if (enableClientAuth && bindDn == null) {
                         log.debug("Will perform External SASL bind because client cert authentication is enabled");
                     } else if (bindDn == null) {
@@ -340,9 +342,6 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
             } catch (final Exception e) {
                 lastException = e;
                 log.warn("Unable to connect to ldapserver {} due to {}. Try next.", ldapHost, e.toString());
-                if (log.isDebugEnabled()) {
-                    log.debug("Unable to connect to ldapserver due to ", e);
-                }
                 Utils.unbindAndCloseSilently(connection);
                 if (needRestore) {
                     restoreClassLoader0(cl);
@@ -369,7 +368,7 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
 
         final Connection delegate = connection;
 
-        if(log.isDebugEnabled()) {
+        if (isDebugEnabled) {
             log.debug("Opened a connection, total count is now {}", CONNECTION_COUNTER.incrementAndGet());
         }
 
@@ -377,7 +376,7 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
 
             @Override
             public Response<Void> reopen(BindRequest request) throws LdapException {
-                if(log.isDebugEnabled()) {
+                if (isDebugEnabled) {
                     log.debug("Reopened a connection");
                 }
                 return delegate.reopen(request);
@@ -385,7 +384,7 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
 
             @Override
             public Response<Void> reopen() throws LdapException {
-                if(log.isDebugEnabled()) {
+                if (isDebugEnabled) {
                     log.debug("Reopened a connection");
                 }
                 return delegate.reopen();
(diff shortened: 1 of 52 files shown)


hardik-k-shah:
We are still checking log.isDebugEnabled once per class. 
Looking for more idea: Is it possible to move it at request level? (once per http or transport request)

sujithvm:
> We are still checking log.isDebugEnabled once per class.
> Looking for more idea: Is it possible to move it at request level? (once per http or transport request)

Main request path is reusing the single computed value as much as possible.
- [OpenDistroSecurityFilter.java ](https://github.com/opendistro-for-elasticsearch/security/pull/895/files#diff-e722c17228b4f01d30b92b69eda762f4248ff4aa54312fcd9d793725bb2044baR298)
  - [PrivilegesEvaluator.java](https://github.com/opendistro-for-elasticsearch/security/pull/895/files#diff-1183213ba048a27e13919954f03c732134daa40fb9c619cda944127431e2fe24R190)
    - [dlsFlsEvaluator](https://github.com/opendistro-for-elasticsearch/security/pull/895/files#diff-1183213ba048a27e13919954f03c732134daa40fb9c619cda944127431e2fe24R232)
    - [securityIndexAccessEvaluator](https://github.com/opendistro-for-elasticsearch/security/pull/895/files#diff-1183213ba048a27e13919954f03c732134daa40fb9c619cda944127431e2fe24R242)
    - [protectedIndexAccessEvaluator](https://github.com/opendistro-for-elasticsearch/security/pull/895/files#diff-1183213ba048a27e13919954f03c732134daa40fb9c619cda944127431e2fe24R247)
    - [PrivilegesInterceptor](https://github.com/opendistro-for-elasticsearch/security/pull/895/files#diff-1183213ba048a27e13919954f03c732134daa40fb9c619cda944127431e2fe24R356)
    - [checkFilteredAliases](https://github.com/opendistro-for-elasticsearch/security/pull/895/files#diff-1183213ba048a27e13919954f03c732134daa40fb9c619cda944127431e2fe24R446)
    - [evaluateAdditionalIndexPermissions](https://github.com/opendistro-for-elasticsearch/security/pull/895/files#diff-1183213ba048a27e13919954f03c732134daa40fb9c619cda944127431e2fe24R499)

- [OpenDistroSecurityRequestHandler.java
](https://github.com/opendistro-for-elasticsearch/security/pull/895/files#diff-5e3ce20033d2d66d179192b90f659338227530a2fe1e98f0f4a6b5b31d3d54a4R246)
  - [BackendRegistry.java](https://github.com/opendistro-for-elasticsearch/security/pull/895/files#diff-7e318f2097c8025fa2814351ef3a46fa1cc288bb6d319621f47941ff367f8644R231) 
    - [authenticate](https://github.com/opendistro-for-elasticsearch/security/pull/895/files#diff-7e318f2097c8025fa2814351ef3a46fa1cc288bb6d319621f47941ff367f8644R231) 
    - [authenticate](https://github.com/opendistro-for-elasticsearch/security/pull/895/files#diff-7e318f2097c8025fa2814351ef3a46fa1cc288bb6d319621f47941ff367f8644R354)

vikansal:
It looks like the debug flag is now passed using stack. While that is not bad, if the purpose was to trace the actions done for the User's request, then the request can have a flag to mark if it is to be debug logged. Therefor having 'debug' a property of Request might be an option. It will still allow new request objects to pickup a debug value if the level configuration changed and was picked up dynamically.

sujithvm:
> It looks like the debug flag is now passed using stack. While that is not bad, if the purpose was to trace the actions done for the User's request, then the request can have a flag to mark if it is to be debug logged. Therefor having 'debug' a property of Request might be an option. It will still allow new request objects to pickup a debug value if the level configuration changed and was picked up dynamically.

Is there some way you would suggest? Looked into following but I fear it would be worse than what we have
- Updating thread context. For each request thread, update thread context with debug/trace enabled values and pass it around. 
- Creating a wrapper around request with (request + debug/trace properties) but involves a lot of wrapping and unwrapping.

vrozov:
> > It looks like the debug flag is now passed using stack. While that is not bad, if the purpose was to trace the actions done for the User's request, then the request can have a flag to mark if it is to be debug logged. Therefor having 'debug' a property of Request might be an option. It will still allow new request objects to pickup a debug value if the level configuration changed and was picked up dynamically.
> 
> Is there some way you would suggest? Looked into following but I fear it would be worse than what we have
> 
> * Updating thread context. For each request thread, update thread context with debug/trace enabled values and pass it around.
> * Creating a wrapper around request with (request + debug/trace properties) but involves a lot of wrapping and unwrapping.

@sujithvm I don't think @vikansal proposes to introduce log level scoped to a user request or specific user. That will require a separate PR. His question is related to passing debug flag on the stack. As long as it is limited to private methods, the scope is still at class level, not request/user level.

sujithvm:
> @sujithvm I don't think @vikansal proposes to introduce log level scoped to a user request or specific user. That will require a separate PR. His question is related to passing debug flag on the stack. As long as it is limited to private methods, the scope is still at class level, not request/user level.

If I understood correctly, proposal was to add an attribute to each request whether it is debuggable and may require a change in upstream Elasticsearch to an attribute

vrozov:
This is only necessary if the original intention of the PR was to introduce request level logging. If the intention is to optimize debug level check in the scope of a class or a method, it is not necessary. The former actually requires a separate PR.

I would recommend to avoid passing debug level flag between methods. Even though it is fine for private methods, it introduces more confusion compared to performance benefits it will provide.

vrozov:
Please see https://logging.apache.org/log4j/log4j-2.2/performance.html. It suggests that the cost of calling `isDebugEnabled()` is negligible unless it is an extremely hot code.

sujithvm:
> This is only necessary if the original intention of the PR was to introduce request level logging

Intention is only to optimize debug level check being evaluated multiple times in the same path flow. 

> It suggests that the cost of calling isDebugEnabled() is negligible unless it is an extremely hot code.

This mainly optimizes the request path flow which can be considered hot code and the same debug/trace evaluation is done ~20-30 times

src/main/java/com/amazon/dlic/auth/ldap/backend/LDAPAuthorizationBackend.java
```diff
@@ -257,9 +257,11 @@ private static Connection getConnection0(final Settings settings, final Path con
         Connection connection = null;
         Exception lastException = null;
 
+        final boolean isDebugEnabled = log.isDebugEnabled();
```
debjanibnrj, line 260: can we make these class level fields, then we can utilize them for the subclasses as well

src/main/java/com/amazon/dlic/auth/ldap2/LDAPAuthorizationBackend2.java
```diff
@@ -408,12 +409,13 @@ private void fillRoles0(final User user, final AuthCredentials optionalAuthCreds
 
     protected Set<LdapName> resolveNestedRoles(final LdapName roleDn, final Connection ldapConnection,
                                                String userRoleName, int depth, final boolean rolesearchEnabled,
-                                               Set<Map.Entry<String, Settings>> roleSearchBaseSettingsSet)
+                                               Set<Map.Entry<String, Settings>> roleSearchBaseSettingsSet,
+                                               final boolean isDebugEnabled, final boolean isTraceEnabled)
```
debjanibnrj, line 413: anyway to avoid passing these in args and declare them as class variables.

src/main/java/com/amazon/dlic/auth/ldap2/LDAPUserSearcher.java
```diff
@@ -101,7 +101,7 @@ private LdapEntry existsSearchingUntilFirstHit(Connection ldapConnection, String
                     SearchScope.SUBTREE);
 
             if (log.isDebugEnabled()) {
```
debjanibnrj, line 103: do we need isDebugEnabled variable here?

src/main/java/com/amazon/dlic/util/SettingsBasedSSLConfigurator.java
```diff
@@ -414,7 +414,7 @@ public SSLConfig(SSLContext sslContext, String[] supportedProtocols, String[] su
             this.effectiveKeyAlias = effectiveKeyAlias;
 
             if (log.isDebugEnabled()) {
```
debjanibnrj, line 416: do we need isDebugEnabled variable here?

sujithvm: This is not reused or loop in same method scope

debjanibnrj (approved):

vrozov (approved):

PR #896, opened 2020-12-14
Remove list creation for single shard request

sujithvm:

 
 Refactoring
 
 2. __Github Issue # or roaddmap entry, if available:__
N/A

- Avoid creating list and converting to array.
- SingleShardRequest has only one index/pattern in request https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/action/support/single/shard/SingleShardRequest.java#L97

- Avoid creating list to save on object creation and performance

 5. __What is the old behavior before changes and new behavior after changes?__ (_Please add any example/logs/screenshot if avaialble_)
- No change in behavior

 6. __Testing done:__ (_Please provide details of testing done: Unit testing, integration testing and manaual testing_) 
 Existing automated tests. No change in behavior expected.

 
 7. __TO-DOs, if any:__ (_Please describe pending items and provide github issues# for each of them)
N/A

N/A

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
@@ -691,13 +691,13 @@ public class IndexResolverReplacer {
             }
             ((UpdateRequest) request).index(newIndices.length!=1?null:newIndices[0]);
         } else if (request instanceof SingleShardRequest) {
-            final SingleShardRequest<?> gr = (SingleShardRequest<?>) request;
-            final String index = gr.index();
-            String[] newIndices = provider.provide(index == null ? null : new String[]{index}, request, true);
-            if (!checkIndices(request, newIndices, true, allowEmptyIndices)) {
+            final SingleShardRequest<?> singleShardRequest = (SingleShardRequest<?>) request;
+            final String index = singleShardRequest.index();
+            String[] indices = provider.provide(index == null ? null : new String[]{index}, request, true);
+            if (!checkIndices(request, indices, true, allowEmptyIndices)) {
                 return false;
             }
-            ((SingleShardRequest) request).index(newIndices.length!=1?null:newIndices[0]);
+            ((SingleShardRequest) request).index(indices.length!=1?null:indices[0]);
         } else if (request instanceof FieldCapabilitiesIndexRequest) {
             // FieldCapabilitiesIndexRequest does not support replacing the indexes.
             // However, the indexes are always determined by FieldCapabilitiesRequest which will be reduced below
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
@@ -692,20 +692,7 @@ public class IndexResolverReplacer {
             ((UpdateRequest) request).index(newIndices.length!=1?null:newIndices[0]);
         } else if (request instanceof SingleShardRequest) {
             final SingleShardRequest<?> gr = (SingleShardRequest<?>) request;
-            final String[] indices = gr.indices();
-            final String index = gr.index();
-
-            final List<String> indicesL = new ArrayList<String>();
-
-            if (index != null) {
-                indicesL.add(index);
-            }
-
-            if (indices != null && indices.length > 0) {
-                indicesL.addAll(Arrays.asList(indices));
-            }
-
-            String[] newIndices = provider.provide(indicesL.toArray(new String[0]), request, true);
+            String[] newIndices = provider.provide(gr.indices(), request, true);
             if (!checkIndices(request, newIndices, true, allowEmptyIndices)) {
                 return false;
             }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
@@ -692,7 +692,8 @@ public class IndexResolverReplacer {
             ((UpdateRequest) request).index(newIndices.length!=1?null:newIndices[0]);
         } else if (request instanceof SingleShardRequest) {
             final SingleShardRequest<?> gr = (SingleShardRequest<?>) request;
-            String[] newIndices = provider.provide(gr.indices(), request, true);
+            final String index = gr.index();
+            String[] newIndices = provider.provide(index == null ? null : new String[]{index}, request, true);
             if (!checkIndices(request, newIndices, true, allowEmptyIndices)) {
                 return false;
             }


hardik-k-shah (approved):

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -692,20 +692,7 @@ private boolean getOrReplaceAllIndices(final Object request, final IndicesProvid
...
-            final List<String> indicesL = new ArrayList<String>();
-
-            if (index != null) {
-                indicesL.add(index);
-            }
-
-            if (indices != null && indices.length > 0) {
-                indicesL.addAll(Arrays.asList(indices));
-            }
-
```
cliu123, line 707: Out of curiosity, why the old logic adds both `index` and `indices` to `indecL`? But why the new logic ignores `index`?

sujithvm: Check the link to [SingleShardRequest.java](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/action/support/single/shard/SingleShardRequest.java#L97) shared in the PR description. It will be the same behavior. Don't know why old code was written that way hence removing it

vrozov:
It looks like a bug fix to me, not a refactoring.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -692,20 +692,7 @@ private boolean getOrReplaceAllIndices(final Object request, final IndicesProvid
...
-            String[] newIndices = provider.provide(indicesL.toArray(new String[0]), request, true);
+            String[] newIndices = provider.provide(gr.indices(), request, true);
```
vrozov, line 695: In case of `null` index, passing array with single element that is `null` does not sound right to me. Can't you pass `null` instead?

sujithvm: I don't see a bug here as the code does the same thing in an undesirable way hence chose it to be refactoring. 
Single element `null` is handled by the inner code. There may be other requests suffering from same issue but will track it in separate issue.

vrozov: In case index is not `null`, how many elements will be in the `indicesL`?

sujithvm: Passing `null` instead of array with single element `null` may not have the same [behavior](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java#L656). Hence don't want to change it.
If taking advantage of knowledge of underlying implementation, `indicesL` will be of size `1`.

vrozov: I don't understand reference to the [behavior](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java#L656). The function is called from `isAllWithNoRemote()` and `isLocalAll()` (did I miss other places?). Do you mean the the result of those 2 methods will be different if we pass `null` or an array with single `null` element? If not, what is the other side effect of passing null?

sujithvm: This is with reference to your suggestion in #886 removing the null check. 
I don't know the side effects of passing null or changing behavior hence uncomfortable making the change.

vrozov: It goes in circle. In #886 the check for an array with single `null` element can't be removed as it is created and passed here.

I am -0 on the PR. I don't block it, as it is an improvement compared to the original code, but also don't approve it as I would prefer to see why `null` can't be passed instead of an array with single `null` element.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -692,20 +692,8 @@ private boolean getOrReplaceAllIndices(final Object request, final IndicesProvid
...
-            if (index != null) {
-                indicesL.add(index);
-            }
-
-            if (indices != null && indices.length > 0) {
-                indicesL.addAll(Arrays.asList(indices));
-            }
-
-            String[] newIndices = provider.provide(indicesL.toArray(new String[0]), request, true);
+            String[] newIndices = provider.provide(index == null ? null : new String[]{index}, request, true);
```
vrozov, line 696: +1

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -692,20 +692,8 @@ private boolean getOrReplaceAllIndices(final Object request, final IndicesProvid
             ((UpdateRequest) request).index(newIndices.length!=1?null:newIndices[0]);
         } else if (request instanceof SingleShardRequest) {
             final SingleShardRequest<?> gr = (SingleShardRequest<?>) request;
```
vrozov, line 694: nit: `gr` -> `singleShardRequest`. Please also use `singleShardRequest` on line 700

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -692,20 +692,8 @@ private boolean getOrReplaceAllIndices(final Object request, final IndicesProvid
...
-
-            if (index != null) {
-                indicesL.add(index);
-            }
-
-            if (indices != null && indices.length > 0) {
-                indicesL.addAll(Arrays.asList(indices));
-            }
-
-            String[] newIndices = provider.provide(indicesL.toArray(new String[0]), request, true);
```
vrozov, line 708: nit: `newIndices ` -> `indices`

vrozov (approved):
Will be good if you can address code style issues as well.

src/main/java/com/amazon/opendistroforelasticsearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -691,25 +691,13 @@ private boolean getOrReplaceAllIndices(final Object request, final IndicesProvid
...
-            if (!checkIndices(request, newIndices, true, allowEmptyIndices)) {
+            final SingleShardRequest<?> singleShardRequest = (SingleShardRequest<?>) request;
+            final String index = singleShardRequest.index();
+            String[] indices = provider.provide(index == null ? null : new String[]{index}, request, true);
+            if (!checkIndices(request, indices, true, allowEmptyIndices)) {
                 return false;
             }
-            ((SingleShardRequest) request).index(newIndices.length!=1?null:newIndices[0]);
+            ((SingleShardRequest) request).index(indices.length!=1?null:indices[0]);
```
vrozov, line 700: nit: sorry if it was not clear from my previous comment. The cast is not necessary. We can use `singleShardRequest` instead.

sujithvm:
Closing PR and will raise new PR to include this change and optimizations in functions `isAllWithNoRemote`, `isLocalAll` from PR #886

PR #946, opened 2021-01-03
Support for ES system index

vrozov:
1. __Category:__ Enhancement/Maintenance
1. __Github Issue # or road-map entry, if available:__
1. __Description of changes:__ Implement `SystemIndexPlugin` interface
1. __Why these changes are required?__ Comply with ES system indices
1. __What is the old behavior before changes and new behavior after changes?__ (_Please add any example/logs/screen-shot if available_)
1. __Testing done:__ UT
1. __TO-DOs, if any:__ (_Please describe pending items and provide Github issues# for each of them_)
1. __Is it backport from master branch?__ (_If yes, please add backport PR # and commits #_)

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
@@ -96,6 +96,7 @@ import org.elasticsearch.index.Index;
 import org.elasticsearch.index.IndexModule;
 import org.elasticsearch.index.cache.query.QueryCache;
 import org.elasticsearch.index.shard.SearchOperationListener;
+import org.elasticsearch.indices.SystemIndexDescriptor;
 import org.elasticsearch.indices.breaker.CircuitBreakerService;
 import org.elasticsearch.plugins.ClusterPlugin;
 import org.elasticsearch.plugins.MapperPlugin;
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
```diff
@@ -1088,7 +1089,14 @@ public void onNodeStarted() {
             }
         };
     }
-    
+
+    @Override
+    public Collection<SystemIndexDescriptor> getSystemIndexDescriptors(Settings settings) {
```
dinusX, line 1094: Shouldn't this be part of the OpenDistroSecuritySSLPlugin.java class ?

Could there be any issues during deployments when switching type to SecurityIndexPlugin ?

dinusX:
Unrelated, we might want to consider making the security index "hidden" which will exclude it from wildcard queries.

See: https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
vrozov: `OpenDistroSecuritySSLPlugin.java` has default implementation of `getSystemIndexDescriptors()`, only `OpenDistroSecurityPlugin` overrides `getSystemIndexDescriptors()` to include `.opendistro_security` index.

I don't see any issue with deployments while enabling SecurityIndexPlugin, it is not a persisted settings in the metadata, so scope of the change is a single node as far as I can see.

vrozov:
> Unrelated, we might want to consider making the security index "hidden" which will exclude it from wildcard queries.
> 
> See: https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html

Likely needs a confirmation, my understanding is that system index is already hidden (excluded from wildcard queries). If not, this needs to be handled as a separate PR.

src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
dinusX: After some discussion, it seems that OpenDistroSecuritySSLPlugin class is used only to enable encryption-in-transit which does not use .opendistro_security index. 

Keeping the method here seems the right approach.

dinusX (approved):

cliu123 (approved):

Issue #879, opened 2020-12-08
IllegalArgumentException: No enum constant CType.AUDIT on mixed version cluster


vrozov:
```
2020-11-16T00:20:13,206][DEBUG][c.a.o.s.a.c.TransportConfigUpdateAction] [] #[org.elasticsearch.transport.RemoteTransportException,java.lang.IllegalArgumentException]#failed to execute on node []
org.elasticsearch.transport.RemoteTransportException: [][:9300][cluster:admin/opendistro_security/config/update[n]]
Caused by: java.lang.IllegalArgumentException: No enum constant com.amazon.opendistroforelasticsearch.security.securityconf.impl.CType.AUDIT
at java.lang.Enum.valueOf(Enum.java:240) ~[?:?]
at com.amazon.opendistroforelasticsearch.security.securityconf.impl.CType.valueOf(CType.java:23) ~[?:?]
at com.amazon.opendistroforelasticsearch.security.securityconf.impl.CType.fromString(CType.java:46) ~[?:?]
at com.amazon.opendistroforelasticsearch.security.securityconf.impl.CType.lambda$fromStringValues$1(CType.java:58) ~[?:?]
at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195) ~[?:?]
at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948) ~[?:?]
at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484) ~[?:?]
at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474) ~[?:?]
at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913) ~[?:?]
at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578) ~[?:?]
at com.amazon.opendistroforelasticsearch.security.securityconf.impl.CType.fromStringValues(CType.java:58) ~[?:?]
at com.amazon.opendistroforelasticsearch.security.action.configupdate.TransportConfigUpdateAction.nodeOperation(TransportConfigUpdateAction.java:116) ~[?:?]
at com.amazon.opendistroforelasticsearch.security.action.configupdate.TransportConfigUpdateAction.nodeOperation(TransportConfigUpdateAction.java:59) ~[?:?]
[...]

PR #953, opened 2021-01-08
ConfigUpdateRequest should include only updated CType

vrozov:
1. __Category:__ Bug fix
2. __Github Issue # or road-map entry, if available:__ #879
3. __Description of changes:__ include only modified config CType into `ConfigUpdateRequest`
4. __Why these changes are required?__ it is not necessary to reload configs not modified as part of PATCH request
5. __What is the old behavior before changes and new behavior after changes?__ all config CTypes are reloaded (even ones that were not modified/updated)
6. __Testing done:__ UT
7. __TO-DOs, if any:__ 

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Objects;
 
 import com.amazon.opendistroforelasticsearch.security.DefaultObjectMapper;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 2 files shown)


cliu123:
How are dependent CTypes handled in this implementation?

vrozov:
> How are dependent CTypes handled in this implementation?

The dependencies are handled by `DynamicConfigFactory.onChange()` that loads all `CType` configurations from `ConfigurationRepository` cache. `ConfigurationRepository` cache is updated in `reloadConfiguration()` call.

cliu123 (dismissed):

debjanibnrj (approved):

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -310,27 +311,27 @@ protected void saveAnUpdateConfigs(final Client client, final RestRequest reques
...
+		private final String[] cTypes;
 		private final Client client;
 		private final ActionListener<Response> delegate;
 
-		public ConfigUpdatingActionListener(Client client, ActionListener<Response> delegate) {
-			super();
-			this.client = client;
-			this.delegate = delegate;
+		public ConfigUpdatingActionListener(String[] cTypes, Client client, ActionListener<Response> delegate) {
+			this.cTypes = Objects.requireNonNull(cTypes, "cTypes must not be null");
```
dinusX, line 326: Should we validate that the list of strings contains valid cTypes ? What happens if it will get some invalid CType ?

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/MigrateApiAction.java
```diff
@@ -119,15 +122,26 @@ protected void handlePost(RestChannel channel, RestRequest request, Client clien
         final SecurityDynamicConfiguration<WhitelistingSettings> whitelistingSettingV6 = (SecurityDynamicConfiguration<WhitelistingSettings>) load(CType.WHITELIST, true);
         final SecurityDynamicConfiguration<AuditConfig> auditConfigV6 = (SecurityDynamicConfiguration<AuditConfig>) load(CType.AUDIT, true);
 
+        final ImmutableMap.Builder<String, SecurityDynamicConfiguration<?>> builder = ImmutableMap.builder();
+
         final SecurityDynamicConfiguration<ActionGroupsV7> actionGroupsV7 = Migration.migrateActionGroups(actionGroupsV6);
+        builder.put(CType.ACTIONGROUPS.toLCString(), actionGroupsV7);
```
dinusX, line 128: Nit: An alternative approach would be storing objects in a set.
The SecurityDynamicConfiguration<> objects have a method called getCType() that will give you CType of an object.

src/main/java/com/amazon/opendistroforelasticsearch/security/dlic/rest/api/AbstractApiAction.java
vrozov: The validation is done on the target node. Note that even if we validate that `cType` is valid on the source node, it may still be invalid on the target node.

dinusX (approved):

cliu123 (approved):

PR #991, opened 2021-01-22
Allowing dual mode to be enabled in full mode

jasonkim-datadog:
Passed all existing unit tests, seeing if we can get this to work in CI after this.


jasonkim-datadog:
Wrong ES fork

PR #389, opened 2020-04-20
Changing file permissions for securityconfig and tools

debjanibnrj:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Backport of PR https://github.com/opendistro-for-elasticsearch/security/pull/388

PR #1117, opened 2021-04-08
Fix index creation

hsiang9431-amzn:

 
 Bug fix
 

https://github.com/opendistro-for-elasticsearch/security/issues/893
https://github.com/opendistro-for-elasticsearch/security-kibana-plugin/issues/598

 Fix requests hitting `_resolve` endpoint failed with `403 forbidden` error due to accessing indices without proper permission. By re-configuring the logic in code, requests that were to return `403` will now success with `200` and not permitted indices will be filtered from its result.

Additional code added to unit tests and configuration ensures such behavior change does not compromise code integrity.

Customers reported being unable to create index in Kibana due to previously mentioned `403 forbidden` error. Such error is effecting customer production thus needs ti be fixed.

Old behavior: `_resolve` endpoint is not properly handled in code, thus causing it to return `403` status code when ever indices that are not allowed access to the user are contained in the response. This error causes Kibana failed to allow user to create indices that they should otherwise be allowed.

New behavior: `_resolve` api is properly handled. If previously mentioned invalid access is encountered, security plugin automatically remove invalid indices accesses in the response and return with `200 ok` status.

Unit test regarding such feature is updated to accommodate with such change.
Manual tests on Kibana web UI is conducted to ensure creation of desired index pattern is allowed.
 
 
Update multi-cluster integratoin tests as `_resolve` api introduced cross-cluster support for getting different index patterns

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluator.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluator.java
@@ -55,6 +55,7 @@ import org.elasticsearch.action.admin.indices.delete.DeleteIndexAction;
 import org.elasticsearch.action.admin.indices.mapping.get.GetFieldMappingsRequest;
 import org.elasticsearch.action.admin.indices.mapping.put.AutoPutMappingAction;
 import org.elasticsearch.action.admin.indices.mapping.put.PutMappingAction;
+import org.elasticsearch.action.admin.indices.resolve.ResolveIndexAction;
 import org.elasticsearch.action.bulk.BulkAction;
 import org.elasticsearch.action.bulk.BulkItemRequest;
 import org.elasticsearch.action.bulk.BulkRequest;
@@ -386,10 +387,11 @@ public class PrivilegesEvaluator {
             }
         }
 
-        if (dnfofEnabled
+        if (action0.equals("indices:admin/resolve/index")
+           ||   (dnfofEnabled
                 && (action0.startsWith("indices:data/read/")
                 || action0.startsWith("indices:admin/mappings/fields/get")
-                || action0.equals("indices:admin/shards/search_shards"))) {
+                || action0.equals("indices:admin/shards/search_shards")))) {
 
             if(requestedResolved.getAllIndices().isEmpty()) {
                 presponse.missingPrivileges.clear();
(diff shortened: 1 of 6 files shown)


cliu123:
Could you please close the [PR1100](https://github.com/opendistro-for-elasticsearch/security/pull/1100) if it's no longer needed?

hsiang9431-amzn:
> Could you please close the [PR1100](https://github.com/opendistro-for-elasticsearch/security/pull/1100) if it's no longer needed?

sure, closed it

vrozov (changes requested):
I don't see why we are not taking the approach similar to what was done before in `TermsAggregationEvaluator`. The `dnfofEnabled` flag should be limited to the handling requests when `do_not_fail_on_forbidden` setting is set to `true` while newly introduced `ResolveEvaluator` should be completely independent from the `dnfofEnabled` flag and handle just the Resolve request independently from `do_not_fail_on_forbidden` setting.

vrozov:
The PR title is quite confusing, it suggests that there is a problem with create index requests, not with handling resolve requests.

src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -353,6 +354,35 @@ public PrivilegesEvaluatorResponse evaluate(final User user, String action0, fin
         final Set<String> allIndexPermsRequired = evaluateAdditionalIndexPermissions(request, action0);
         final String[] allIndexPermsRequiredA = allIndexPermsRequired.toArray(new String[0]);
 
+        // handle resolve requests
+        if (action0.equals("indices:admin/resolve/index")) {
```
sujithvm, line 358: Can we use ResolveIndexAction.NAME ?

src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -353,6 +354,35 @@ public PrivilegesEvaluatorResponse evaluate(final User user, String action0, fin
...
+
+            if(requestedResolved.getAllIndices().isEmpty()) {
```
sujithvm, line 360: Is it possible to short circuit the condition for `do_not_fail_on_forbidden` line `#389` and reuse the same logic ?

src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluator.java
vrozov: Why don't we use request type instead of request name.

src/main/java/com/amazon/opendistroforelasticsearch/security/privileges/PrivilegesEvaluator.java
vrozov: @sujithvm do you mean refactor code under `do_not_fail_on_forbidden` condition? I am not sure what do you mean by short circuit?

sujithvm: By [short circuit](https://en.wikipedia.org/wiki/Short-circuit_evaluation) here I mean, add the `action0.equals("indices:admin/resolve/index")` to the if statement as it currently executes only if `do_not_fail_on_forbidden` is enabled. 

Refactoring the code to reuse it again would be fine as well

vrozov: I don't think that combining `do_not_fail_on_forbidden` with other conditions is the right approach.

PR #1119, opened 2021-04-13
Introducing passive_intertransport_auth to facilitate communication b…

dhiAmzn:
…etween nodes with adv sec enabled and nodes without adv sec enabled.

Changes to ssl_dual_mode_enabled to support the same.

 1. __Category:__ New feature
 

 
 3. __Description of changes:__ Introducing opendistro security dynamic config feature flag "passive_intertransport_auth".
 Additionally, making changes to ssl_dual_node behaviour.
The changes allow transport communication between nodes with advanced security enabled and nodes without advanced security enabled (ssl only or plugin disabled).
The feature flag should be used transiently to facilitate migration of cluster to using advanced security.

Currently, migrating to advanced security requires a full cluster restart as any other kind of migration process causes transport communication to break between old and newer nodes. These changes will allow a cluster to be migrated to use advanced security without any downtime, by allowing old and new nodes to successfully communicate at the transport layer.
NOTE: There is an existing feature `disable_intertransport_auth` that if enabled will cause transport level authorization to be skipped. However, the side affect of this is that any non-super admin user will be able to update opendistro_security during migration. 
`passive_intertransport_auth` does not skip authorization, instead it injects a default user, which you can map to the permissions you want to give during this transient phase of migration.

Old Behaviour: Transport communication from old node to a new node fails at the new nodes with the error: `No user found for...`
New Behaviour: Transport communication from old node to a new node succeeds. When a user is not found and this feature flag is set to true, we inject a default user. Note that this user should have all index and cluster permissions.

 Extensive UTs to cover all cases.

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
@@ -191,6 +191,7 @@ public final class OpenDistroSecurityPlugin extends OpenDistroSecuritySSLPlugin 
     private volatile ClusterService cs;
     private volatile AuditLog auditLog;
     private volatile BackendRegistry backendRegistry;
+    private volatile CompatConfig compatConfig;
     private volatile SslExceptionHandler sslExceptionHandler;
     private volatile Client localClient;
     private final boolean disabled;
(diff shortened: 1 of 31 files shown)


dhiAmzn:
Resolving in favour of https://github.com/opensearch-project/security/pull/1142

PR #1156, opened 2021-05-03
Introducing passive_intertransport_auth to facilitate communication b…

dhiAmzn:
…etween nodes with adv sec enabled and nodes without adv sec enabled.

Changes to ssl_dual_mode_enabled to support the same.

Signed-off-by: Dhiresh Jain <dhireshj@amazon.com>

__NOTE__: This commit has been moved from opendistro repo to opensearch. The manual test was done using docker in opendistro. Docker testing is currently unavailable in opensearch so did not perform a manual test here yet.
 
 1. __Category:__ New feature
 

 
 3. __Description of changes:__ Introducing opendistro security dynamic config feature flag "passive_intertransport_auth".
 Additionally, making changes to ssl_dual_node behaviour.
The changes allow transport communication between nodes with advanced security enabled and nodes without advanced security enabled (ssl only or plugin disabled).
The feature flag should be used transiently to facilitate migration of cluster to using advanced security.

Currently, migrating to advanced security requires a full cluster restart as any other kind of migration process causes transport communication to break between old and newer nodes. These changes will allow a cluster to be migrated to use advanced security without any downtime, by allowing old and new nodes to successfully communicate at the transport layer.
NOTE: There is an existing feature `disable_intertransport_auth` that if enabled will cause transport level authorization to be skipped. However, the side affect of this is that any non-super admin user will be able to update opendistro_security during migration. 
`passive_intertransport_auth` does not skip authorization, instead it injects a default user, which you can map to the permissions you want to give during this transient phase of migration.

Old Behaviour: Transport communication from old node to a new node fails at the new nodes with the error: `No user found for...`
New Behaviour: Transport communication from old node to a new node succeeds. When a user is not found and this feature flag is set to true, we inject a default user. Note that this user should have all index and cluster permissions.

 Extensive UTs to cover all cases.

```
__With passive_intertransport_auth disabled:__
dhireshj@f45c899db61b ~/w/d/dhi_security (transport_passive_mode)> curl "https://localhost:9200/_cat/nodes" -k
192.168.240.3 18 22 44 2.01 1.88 1.04 dimr * elasticsearch1
192.168.240.2 55 20 49 2.01 1.88 1.04 dir  - elasticsearch2
dhireshj@f45c899db61b ~/w/d/dhi_security (transport_passive_mode)> curl -X PUT "https://localhost:9200/test5?pretty" -H 'Content-Type: application/json' -d'
                                                                   {
                                                                       "settings": {
                                                                           "index": {
                                                                             "number_of_shards": 1,
                                                                             "auto_expand_replicas": "0-all",
                                                                             "routing.allocation.exclude._name": "elasticsearch1"
                                                                           }
                                                                       }
                                                                   }' -k
{
  "acknowledged" : true,
  "shards_acknowledged" : true,
  "index" : "test5"
}
dhireshj@f45c899db61b ~/w/d/dhi_security (transport_passive_mode)> curl -XPOST "https://localhost:9200/test5/_doc/1" -k -H 'Content-Type: application/json' -d'{}' -k
{"error":{"root_cause":[{"type":"security_exception","reason":"No user found for indices:admin/mapping/auto_put"}],"type":"security_exception","reason":"No user found for indices:admin/mapping/auto_put"},"status":500}⏎


__With passive_intertransport_auth enabled:__

dhireshj@f45c899db61b ~/w/d/dhi_security (transport_passive_mode) [60]> curl "https://localhost:9200/_cat/nodes" -k
192.168.240.3 17 21 48 2.84 1.97 0.95 dimr * elasticsearch1
192.168.240.2 26 21 49 2.84 1.97 0.95 dir  - elasticsearch2
dhireshj@f45c899db61b ~/w/d/dhi_security (transport_passive_mode)> curl -X PUT "https://localhost:9200/test4?pretty" -H 'Content-Type: application/json' -d'
                                                                   {
                                                                       "settings": {
                                                                           "index": {
                                                                             "number_of_shards": 1,
                                                                             "auto_expand_replicas": "0-all",
                                                                             "routing.allocation.exclude._name": "elasticsearch1"
                                                                           }
                                                                       }
                                                                   }' -k
{
  "acknowledged" : true,
  "shards_acknowledged" : true,
  "index" : "test4"
}
dhireshj@f45c899db61b ~/w/d/dhi_security (transport_passive_mode) [52]> curl -XPOST "https://localhost:9200/test4/_doc/1" -k -H 'Content-Type: application/json' -d'{}' -k
{"_index":"test4","_type":"_doc","_id":"1","_version":1,"result":"created","_shards":{"total":1,"successful":1,"failed":0},"_seq_no":0,"_primary_term":1}⏎
```

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/src/main/java/org/opensearch/security/configuration/CompatConfig.java
+++ b/src/main/java/org/opensearch/security/configuration/CompatConfig.java
@@ -30,7 +30,7 @@
 
 package org.opensearch.security.configuration;
 
-import com.amazon.opendistroforelasticsearch.security.setting.OpensearchDynamicSetting;
+import org.opensearch.security.setting.OpensearchDynamicSetting;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.common.settings.Settings;
--- a/src/main/java/org/opensearch/security/configuration/CompatConfig.java
+++ b/src/main/java/org/opensearch/security/configuration/CompatConfig.java
@@ -30,6 +30,7 @@
 
 package org.opensearch.security.configuration;
 
+import com.amazon.opendistroforelasticsearch.security.setting.OpensearchDynamicSetting;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.common.settings.Settings;
@@ -39,15 +40,19 @@ import org.greenrobot.eventbus.Subscribe;
 import org.opensearch.security.securityconf.DynamicConfigModel;
 import org.opensearch.security.support.ConfigConstants;
 
+import static org.opensearch.security.support.ConfigConstants.SECURITY_UNSUPPORTED_PASSIVE_INTERTRANSPORT_AUTH_INITIALLY;
+
 public class CompatConfig {
 
     private final Logger log = LogManager.getLogger(getClass());
     private final Settings staticSettings;
     private DynamicConfigModel dcm;
+    private final OpensearchDynamicSetting<Boolean> transportPassiveAuthSetting;
 
-    public CompatConfig(final Environment environment) {
+    public CompatConfig(final Environment environment, final OpensearchDynamicSetting<Boolean> transportPassiveAuthSetting) {
         super();
-        this.staticSettings = environment.settings(); 
+        this.staticSettings = environment.settings();
+        this.transportPassiveAuthSetting = transportPassiveAuthSetting;
     }
     
     @Subscribe
@@ -105,23 +110,10 @@ public class CompatConfig {
      * Returns true if passive transport auth is enabled
      */
     public boolean transportInterClusterPassiveAuthEnabled() {
-        final boolean interClusterAuthInitiallyPassive = staticSettings.getAsBoolean(ConfigConstants.SECURITY_UNSUPPORTED_PASSIVE_INTERTRANSPORT_AUTH_INITIALLY, false);
-
-        if(interClusterAuthInitiallyPassive) {
-            if(dcm == null) {
-                if(log.isTraceEnabled()) {
-                    log.trace("dynamicSecurityConfig is null, initially static interClusterAuthPassive");
-                }
-                return false;
-            } else {
-                final boolean interClusterAuthDynamicallyPassive = dcm.isInterTransportAuthPassive();
-                if(log.isTraceEnabled()) {
-                    log.trace("opendistro_security.dynamic.passive_intertransport_auth {}", interClusterAuthDynamicallyPassive);
-                }
-                return interClusterAuthDynamicallyPassive;
-            }
-        } else {
-            return false;
+        final boolean interClusterAuthInitiallyPassive = transportPassiveAuthSetting.getDynamicSettingValue();
+        if(log.isTraceEnabled()) {
+            log.trace("{} {}", SECURITY_UNSUPPORTED_PASSIVE_INTERTRANSPORT_AUTH_INITIALLY, interClusterAuthInitiallyPassive);
         }
+        return interClusterAuthInitiallyPassive;
     }
 }
--- a/src/main/java/org/opensearch/security/configuration/CompatConfig.java
+++ b/src/main/java/org/opensearch/security/configuration/CompatConfig.java
@@ -100,4 +100,28 @@ public class CompatConfig {
             return true;
         }
     }
+
+    /**
+     * Returns true if passive transport auth is enabled
+     */
+    public boolean transportInterClusterPassiveAuthEnabled() {
+        final boolean interClusterAuthInitiallyPassive = staticSettings.getAsBoolean(ConfigConstants.SECURITY_UNSUPPORTED_PASSIVE_INTERTRANSPORT_AUTH_INITIALLY, false);
+
+            if(interClusterAuthInitiallyPassive) {
+                if(dcm == null) {
+                    if(log.isTraceEnabled()) {
+                        log.trace("dynamicSecurityConfig is null, initially static interClusterAuthPassive");
+                    }
+                    return false;
+                } else {
+                    final boolean interClusterAuthDynamicallyPassive = dcm.isInterTransportAuthPassive();
+                    if(log.isTraceEnabled()) {
+                        log.trace("opendistro_security.dynamic.passive_intertransport_auth {}", interClusterAuthDynamicallyPassive);
+                    }
+                    return interClusterAuthDynamicallyPassive;
+                }
+            } else {
+                return false;
+            }
+    }
 }
--- a/src/main/java/org/opensearch/security/configuration/CompatConfig.java
+++ b/src/main/java/org/opensearch/security/configuration/CompatConfig.java
@@ -107,21 +107,21 @@ public class CompatConfig {
     public boolean transportInterClusterPassiveAuthEnabled() {
         final boolean interClusterAuthInitiallyPassive = staticSettings.getAsBoolean(ConfigConstants.SECURITY_UNSUPPORTED_PASSIVE_INTERTRANSPORT_AUTH_INITIALLY, false);
 
-            if(interClusterAuthInitiallyPassive) {
-                if(dcm == null) {
-                    if(log.isTraceEnabled()) {
-                        log.trace("dynamicSecurityConfig is null, initially static interClusterAuthPassive");
-                    }
-                    return false;
-                } else {
-                    final boolean interClusterAuthDynamicallyPassive = dcm.isInterTransportAuthPassive();
-                    if(log.isTraceEnabled()) {
-                        log.trace("opendistro_security.dynamic.passive_intertransport_auth {}", interClusterAuthDynamicallyPassive);
-                    }
-                    return interClusterAuthDynamicallyPassive;
+        if(interClusterAuthInitiallyPassive) {
+            if(dcm == null) {
+                if(log.isTraceEnabled()) {
+                    log.trace("dynamicSecurityConfig is null, initially static interClusterAuthPassive");
                 }
-            } else {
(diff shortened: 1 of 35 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java
```diff
@@ -191,6 +191,7 @@
     private volatile ClusterService cs;
     private volatile AuditLog auditLog;
     private volatile BackendRegistry backendRegistry;
+    private volatile CompatConfig compatConfig;
```
vrozov, line 194: Why it is necessary to make `compatConfig` member variable?

dhiAmzn: Will revert. I initially required it because it was being used in multiple places, but not required anymore.

src/test/java/com/amazon/opendistroforelasticsearch/security/AdvancedSecurityMigrationTests.java
```diff
@@ -0,0 +1,284 @@
+package com.amazon.opendistroforelasticsearch.security;
```
vengadanathan-s, line 1: nit: does license info needed to be added at top ?

src/main/java/com/amazon/opendistroforelasticsearch/security/configuration/CompatConfig.java
```diff
@@ -100,4 +100,28 @@ public boolean transportInterClusterAuthEnabled() {
...
     }
+
+    /**
+     * Returns true if passive transport auth is enabled
+     */
+    public boolean transportInterClusterPassiveAuthEnabled() {
+        final boolean interClusterAuthInitiallyPassive = staticSettings.getAsBoolean(ConfigConstants.OPENDISTRO_SECURITY_PASSIVE_INTERTRANSPORT_AUTH_INITIALLY, false);
+
+            if(interClusterAuthInitiallyPassive) {
+                if(dcm == null) {
```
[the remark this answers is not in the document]
dhiAmzn: `dcm` is null when the security index has not been initialised.
We are returning false as it is the default value, and the reason we are returning the default value here is because other feature flags are also doing the same and I wanted the behaviour to be consistent.
Adding to that, with the current flow, the only side effect it causes when it is null is that the exception thrown is not very clear of what is actually happening i.e. it will throw a "No user found for" instead of something like "Security index not initialised"

If we want to instead throw a runtime exception here, then the concerns I have are:
1. What about the other places where we return a default value? Do we make changes there too to have a consistent experience?
2. What if a caller wants to handle it differently instead of throwing an exception. I can't think of a case yet but want to have that discussion. On similar lines, what if the caller wants to throw a different exception or the corresponding exception message? For both these cases, if we agree caller needs to have more freedom on how to treat the null value, perhaps we can return a NULL value instead so caller can decide.

@vrozov - Would like your input here.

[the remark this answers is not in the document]
vrozov: I am OK with the current approach as it is consistent with the prior behavior. If it needs to be changed, we can do it in a separate PR for all cases.

src/main/java/com/amazon/opendistroforelasticsearch/security/user/User.java
```diff
@@ -55,6 +55,7 @@
 public class User implements Serializable, Writeable, CustomAttributesAware {
 
     public static final User ANONYMOUS = new User("opendistro_security_anonymous", Lists.newArrayList("opendistro_security_anonymous_backendrole"), null);
+    public static final User DEFAULT_TRANSPORT_USER = new User("opendistro_security_default_transport_user", Lists.newArrayList("opendistro_security_default_transport_backendrole"), null);
```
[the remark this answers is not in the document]
vrozov: Use regular (not javadoc style) comment. Security plugin is not a java library and does not expose Java API for other projects/plugins to use.

vengadanathan-s (dismissed):

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/impl/v7/ConfigV7.java
```diff
@@ -112,6 +113,7 @@ public String toString() {
         public String filtered_alias_mode = "warn";
         public boolean disable_rest_auth;
         public boolean disable_intertransport_auth;
+        public boolean passive_intertransport_auth;
```
vrozov, line 116: What if this flag is enabled on a domain with prior versions of the security plugin installed? Won't it cause backward compatibility issue and failures due to jackson unknown property exception?

src/main/java/org/opensearch/security/configuration/CompatConfig.java
```diff
@@ -100,4 +105,15 @@ public boolean transportInterClusterAuthEnabled() {
             return true;
         }
     }
+
+    /**
+     * Returns true if passive transport auth is enabled
+     */
+    public boolean transportInterClusterPassiveAuthEnabled() {
+        final boolean interClusterAuthInitiallyPassive = transportPassiveAuthSetting.getDynamicSettingValue();
```
vengadanathan-s, line 113: How are we honoring the YML setting value that is set or are we ignoring it ?

dhiAmzn: Setting it in TransportPassiveAuthSetting constructor

src/main/java/org/opensearch/security/setting/TransportPassiveAuthSetting.java
```diff
@@ -0,0 +1,36 @@
...
+package org.opensearch.security.setting;
+
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
+
+public class TransportPassiveAuthSetting extends OpensearchDynamicSetting<Boolean> {
+
+    public TransportPassiveAuthSetting(final Settings settings) {
+        super(Setting.boolSetting(
```
vengadanathan-s, line 25: Nit: can this be set as variables and passed to improve readability, i almost missed the second arg passed.

vengadanathan-s (approved):

src/main/java/org/opensearch/security/setting/OpensearchDynamicSetting.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public T getDynamicSettingValue() {
+        return dynamicSettingValue;
+    }
+
+    public Setting<T> getDynamicSetting() {
+        return dynamicSetting;
+    }
+
+}
```
vrozov, line 64: nit: fix new line

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
```diff
@@ -55,6 +55,7 @@
 import org.opensearch.transport.TransportRequest;
 import org.opensearch.transport.TransportRequestHandler;
 
+
```
vrozov, line 58: undo

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
```diff
@@ -63,6 +64,10 @@
 import org.opensearch.security.support.ConfigConstants;
 import org.opensearch.security.support.HeaderHelper;
 import org.opensearch.security.user.User;
+import org.opensearch.security.ssl.transport.SecuritySSLRequestHandler;
+import org.opensearch.security.ssl.transport.SSLConfig;
+
+
```
vrozov, line 70: remove extra empty line

src/main/java/com/amazon/opendistroforelasticsearch/security/securityconf/impl/v7/ConfigV7.java
dhiAmzn: Current design does not support backward compatibility when introducing new feature flags. Discussed offline, and for this change moving the flag from opendistro config to es setting

dhiAmzn: The backward incompatibility occurs when the following happens:
Old node does not have the code checked in, and a request to update this feature flag is sent to this old node. This will cause cluster to fail all requests and requires manual intervention.

src/main/java/org/opensearch/security/setting/TransportPassiveAuthSetting.java
dhiAmzn: Makes sense, will change it

vengadanathan-s (dismissed):

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
```diff
@@ -278,6 +278,18 @@ public Settings get(int i) {
         };
     }
 
+    protected NodeSettingsSupplier genericMinimumSecuritySettings(List<Settings> others, List<Boolean> sslOnly) {
+
+        return i -> {
```
vrozov, line 283: this code is hard to understand without comments. Should it check size of `sslOnly` and `others` instead?

dhiAmzn: Can check for overflow and underflow explicitly and then set to default value. Thoughts?

vrozov (approved):

vengadanathan-s (approved):

vrozov:
@vengadanathan-s when merging PR please make sure to delete all unnecessary details from the commit message.

PR #1176, opened 2021-05-21
Settings migration to Opensearch

andy840314:

 
 Refactoring
 
 
Several files to be modified

`OpenSearchSecurityPlugin` (https://github.com/opensearch-project/security/blob/f698fe6c2168da3765f9aa151774da64ec024412/src/main/java/com/amazon/opendistroforelasticsearch/security/OpenDistroSecurityPlugin.java#L858), `OpenSearchSecuritySSLPlugin` (https://github.com/opensearch-project/security/blob/f698fe6c2168da3765f9aa151774da64ec024412/src/main/java/com/amazon/opendistroforelasticsearch/security/ssl/OpenDistroSecuritySSLPlugin.java#L318):
Change `getSettings()` to add settings from `SecuritySettings`/`SSLSecuritySettings` instead of add settings directly. e.g. 
`settings.add(Setting.boolSetting(ConfigConstants.SECURITY_SSL_ONLY, ...));` → `settings.add(SecuritySettings.SECURITY_SSL_ONLY)`;

All files that uses `ConfigConstants` and `SSLConfigConstants`:
Remove prefix `OPENDISTRO_`

All files that contains ` opendistro_security` string for settings for several test files):
Change `opendistro_security` to `plugins.security`

*6* files to be added

`ConfigConstants.java`: 
Copy and modify file `LegacyOpenDistroConfigConstants.java`. Change any string that contains `opendistro_security` to `plugins.security`,  remove prefix `OPENDISTRO_`  for variable names.

`SecuritySettings.java`:
New class for settings. Use `LegacyOpenDistroSecuritySettings` as fallbackSetting  to ensure backward compatibility for opendistro

`LegacyOpenDistroSecuritySettings.java`
New class for legacy settings. Define the default value for settings here.

`SSLConfigConstants.java`:   
Copy and modify file `LegacyOpenDistroSSLConfigConstants.java`. Change any string that contains `opendistro_security` to `plugins.security`,  remove prefix `OPENDISTRO_`  for variable names.

`SSLSecuritySettings.java`:
New class for SSL settings. Use `LegacyOpenDistroSSLSecuritySettings` as fallbackSetting  to ensure backward compatibility for opendistro

`LegacyOpenDistroSSLSecuritySettings.java`:
New class for legacy SSL settings,  Define the default value for SSL settings here.

*2* files name changed

`ConfigConstants.java` →   `LegacyOpenDistroConfigConstants.java`
`SSLConfigConstants.java` → `LegacyOpenDistroSSLConfigConstants.java`

1 test file added

`SecuritySettingTests.java`

 Settings rename and backward compatibility for OpenDistro

 Unit tests
 
 
    indicated in the file; or

    in the file; or

    it.


.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
     - name: Build OpenSearch
       working-directory: ./OpenSearch
-      run: ./gradlew publishToMavenLocal -Dbuild.snapshot=false
+      run: ./gradlew publishToMavenLocal -Dbuild.version_qualifier=beta1 -Dbuild.snapshot=false
```
dblock, line 48: `-rc1`

cliu123: @dblock Both `rc1` and removing `-Dbuild.version_qualifier` fails security plugin build, which is this [issue](https://github.com/opensearch-project/OpenSearch/issues/762) for.

vrozov: The problem was introduced in https://github.com/opensearch-project/security/pull/1174 that removes qualifier but does not update pom dependency. I wonder how that PR was tested and approved.

dblock: This works here because you have hard-coded beta1 in the plugin for the OpenSearch dependency. See https://github.com/opensearch-project/OpenSearch/issues/762#issuecomment-846291955. 

Change this to be rc1, and update your code to say rc1 instead beta1. 

These qualifiers don't do anything in the build as long as they are consistent so that the dependency on OpenSearch can be resolved, but in order for the plugin to start the version of the engine and the plugin will have to match, and the next version is rc1.

vrozov: > This works here because you have hardcoded beta1 in pom.xml for the OpenSearch dependency. See opensearch-project/OpenSearch#762 (comment).

@dblock I guess you refer to CI, but it works only because maven artifacts are cached: https://github.com/opensearch-project/security/blob/5a81a424f4dc0dd1e72df9d8c27710e9dbcdec1b/.github/workflows/ci.yml#L32 Without caching maven artifacts or on a fresh box with no prior build of beta1 artifacts, the build would fail.

> These qualifiers don't do anything in the build as long as they are consistent so that the dependency on OpenSearch can be resolved, but in order for the plugin to start the version of the engine and the plugin will have to match, and the next version is rc1.

If `rc1` does not correspond to a commit/tag in Opensearch, why is it used? If plugins are supposed to be built against the latest commit of 1.x branch, it should use `SNAPSHOT` (this is the meaning of `SNAPSHOT`), not an `rc1` tag/qualifier.

dblock: Yes re: cached maven artifacts, I didn't see that, should probably be removed now

You're right on rc1. If we had Maven we would be able to say exactly what you're building against, which could be an rc1 eventually. Right now if you called everything -SNAPSHOT you wouldn't have a way to rename that to rc1 because it would need a change in code. I think we need to fix https://github.com/opensearch-project/OpenSearch/issues/763 first, then we can undo this craziness. It's just a workaround to produce something called `-rc1` everywhere.

vrozov: > Yes re: cached maven artifacts, I didn't see that, should probably be removed now

Caching maven artifacts should not be removed, it works as designed to speed up security plugin build and avoid loading artifacts from maven on each build. What should be removed is using CI to build and publish Opensearch artifacts to local maven repo.

dblock: You can be right, or you can prevent a similar problem from happening in the short term. Your call.

vrozov: Similar problems should be prevented by testing and PR review, not by removing what worked before and is not broken. Even if caching is removed, what prevents CI of using wrong branch/tag and naming it `beta1` or `rc1`?

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -46,229 +46,229 @@
...
+    public static final String SECURITY_PROTECTED_INDICES_ENABLED_KEY = "plugins.security.protected_indices.enabled";
+    public static final Boolean SECURITY_PROTECTED_INDICES_ENABLED_DEFAULT = false;
+    public static final String SECURITY_PROTECTED_INDICES_KEY = "plugins.security.protected_indices.indices";
+    public static final List<String> SECURITY_PROTECTED_INDICES_DEFAULT = Collections.emptyList();
+    public static final String SECURITY_PROTECTED_INDICES_ROLES_KEY = "plugins.security.protected_indices.roles";
+    public static final List<String> SECURITY_PROTECTED_INDICES_ROLES_DEFAULT = Collections.emptyList();
 
     // Roles injection for plugins
-    public static final String OPENDISTRO_SECURITY_INJECTED_ROLES = "opendistro_security_injected_roles";
+    public static final String SECURITY_INJECTED_ROLES = "opendistro_security_injected_roles";
```
skkosuri-amzn, line 265: Any particular reason this is not changed ?  "opendistro_security_injected_roles"

andy840314: This is not a setting i believe. The scope for this PR is to rename the settings. Correct me if i am wrong.

src/test/java/org/opensearch/security/SecuritySettingsTests.java
```diff
@@ -0,0 +1,454 @@
...
+import org.opensearch.security.support.SecuritySettings;
+
+import java.util.Map;
+
+public class SecuritySettingsTests {
+    
+    @Test
+    public void testLegacyOpenDistroSettingsFallback() {
+        Assert.assertEquals(
+                SecuritySettings.SECURITY_ADVANCED_MODULES_ENABLED.get(Settings.EMPTY),
```
vrozov, line 30: What is the purpose of the test? While getting setting from `Settings.EMPTY` even two unrelated settings will produce the same result.

andy840314: This is to test if the default value of `SecuritySettings` has fallback to the default value of `LegacyOpenDistroSecuritySettings`.

vrozov:
Was this PR tested for backward compatibility during rolling upgrade from ODFE distribution?

andy840314:
> backward compatibility during rolling upgrade from ODFE distribution?

Can you specify more about this?

dblock:
> Was this PR tested for backward compatibility during rolling upgrade from ODFE distribution?

Unlikely because https://github.com/opensearch-project/OpenSearch/issues/669 prevents this altogether rn.

src/test/resources/auditlog/endpoints/sink/configuration_all_variants.yml
```diff
@@ -1,47 +1,48 @@
...
-        type: log4j
-        config:
-          log4j.logger_name: loggername
-          log4j.level: invalid
+plugins:
+  security:
+    audit:
+      type: debug
+      endpoints:
+        eNDpoint1:
```
hsiang9431-amzn, line 6: is this capitol case correct?

src/test/resources/auditlog/endpoints/sink/configuration_all_variants.yml
```diff
@@ -1,47 +1,48 @@
...
+          type: internal_opensearch
+          config:
+            key: value
+            wedontneed: anyconfigforinternal
+        endpoint2:
+          type: external_opensearch
+        endPOINT3:
```
hsiang9431-amzn, line 13: is this capitol case correct?

it may not be case sensitive though

src/test/resources/auditlog/endpoints/sink/configuration_no_default.yml
```diff
@@ -1,32 +1,33 @@
...
-      endpoint9:
-        type: external_opensearch
-        config:
-          endpoints: stringhere
\ No newline at end of file
+plugins:
+  security:
+    audit:
+      endpoints:
+        eNDpoint1:
```
hsiang9431-amzn, line 5: upper case here?

src/test/resources/auditlog/endpoints/sink/configuration_no_default.yml
```diff
@@ -1,32 +1,33 @@
...
+          type: internal_opensearch
+          config:
+            key: value
+            wedontneed: anyconfigforinternal
+        endpoint2:
+          type: external_opensearch
+        endPOINT3:
```
hsiang9431-amzn, line 12: upper case here?

src/test/resources/auditlog/endpoints/sink/configuration_all_variants.yml
andy840314: This is as same as the original code. The only part changed is 
```
opendistro_security:
``` 
to 
```
plugins:
  security:
```

src/test/resources/auditlog/endpoints/sink/configuration_all_variants.yml
andy840314: See comment above.

src/test/resources/auditlog/endpoints/sink/configuration_no_default.yml
andy840314: See comment above.

src/test/resources/auditlog/endpoints/sink/configuration_no_default.yml
andy840314: See comment above.

vrozov:
> > Was this PR tested for backward compatibility during rolling upgrade from ODFE distribution?
> 
> Unlikely because [opensearch-project/OpenSearch#669](https://github.com/opensearch-project/OpenSearch/issues/669) prevents this altogether rn.

I highly doubt that even after fixing [opensearch-project/OpenSearch#669](https://github.com/opensearch-project/OpenSearch/issues/669), it will work in clusters where security plugin is installed and enabled. The PR introduces backward incompatible changes.

andy840314:
> I highly doubt that even after fixing opensearch-project/OpenSearch#669, it will work in clusters where security plugin is installed and enabled. The PR introduces backward incompatible changes.

Can you point out the backward incompatible changes? This code is compatible for the old `opendistro_security.` settings.

saratvemulapalli:
> > I highly doubt that even after fixing [opensearch-project/OpenSearch#669](https://github.com/opensearch-project/OpenSearch/issues/669), it will work in clusters where security plugin is installed and enabled. The PR introduces backward incompatible changes.
> 
> Can you point out the backward incompatible changes? This code is compatible for the old `opendistro_security.` settings.

+1 @vrozov if there are more changes to be taken care of could you list them down on an issue?
Also curious what they are, and if they are specific to security plugin or all the plugins...

vrozov:
> > > I highly doubt that even after fixing [opensearch-project/OpenSearch#669](https://github.com/opensearch-project/OpenSearch/issues/669), it will work in clusters where security plugin is installed and enabled. The PR introduces backward incompatible changes.
> > 
> > 
> > Can you point out the backward incompatible changes? This code is compatible for the old `opendistro_security.` settings.
> 
> +1 @vrozov if there are more changes to be taken care of could you list them down on an issue?
> Also curious what they are, and if they are specific to security plugin or all the plugins...

@andy840314 The change impacts not only settings, but also inter node communication (for example  headers). I am not in position to point to all possible backward incompatible changes, so was my request to extensively test those changes.

@saratvemulapalli those are specific to the security plugin, plus it may impact few other plugins that pass info to the security plugin using thread context.

andy840314:
> @andy840314 The change impacts not only settings, but also inter node communication (for example headers). I am not in position to point to all possible backward incompatible changes, so was my request to extensively test those changes.

Actually, only the variable name for those constant are changed. 
e.g. `OPENDISTRO_SECURITY_USER_HEADER` -> `SECURITY_USER_HEADER`
The constant string in the variable is not modified so it should be fine.

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
```diff
@@ -376,25 +376,25 @@ public String toString() {
...
-            ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_INTERNAL_CONFIG_ENABLED,
-            ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_EXTERNAL_CONFIG_ENABLED,
-            ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_READ_METADATA_ONLY,
-            ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_READ_IGNORE_USERS,
-            ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_READ_WATCHED_FIELDS,
-            ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_WRITE_METADATA_ONLY,
-            ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_WRITE_LOG_DIFFS,
-            ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_WRITE_IGNORE_USERS,
-            ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_WRITE_WATCHED_INDICES
+            ConfigConstants.SECURITY_AUDIT_ENABLE_REST,
```
vrozov, line 379: This introduces new key and immediately deprecates it. All those keys are already deprecated and should not be migrated to new keys.

andy840314:
These changes will split into different PR as discussed.

PR #1179, opened 2021-05-22
Add build.version_qualifier and make security plugin compatible with OpenSearch 1.0.0-rc1

cliu123:

 
 Maintenance
 
 
- Build OpenSearch 1.0.0-rc1 on CI.
- Make security plugin compatible with OpenSearch 1.0.0-rc1.

OpenSearch release 1.0.0-rc1

No changes in behavior

 UT
 
 
    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
 
     - name: Build OpenSearch
       working-directory: ./OpenSearch
-      run: ./gradlew publishToMavenLocal -Dbuild.snapshot=false
+      run: ./gradlew publishToMavenLocal -Dbuild.version_qualifier=rc1 -Dbuild.snapshot=false
 
     - name: Checkstyle
       run: mvn -B checkstyle:checkstyle
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
 
     - name: Build OpenSearch
       working-directory: ./OpenSearch
-      run: ./gradlew publishToMavenLocal -Dbuild.version_qualifier=snapshot -Dbuild.snapshot=false
+      run: ./gradlew publishToMavenLocal -Dbuild.version_qualifier=rc1 -Dbuild.snapshot=false
 
     - name: Checkstyle
       run: mvn -B checkstyle:checkstyle
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
 
     - name: Build OpenSearch
       working-directory: ./OpenSearch
-      run: ./gradlew publishToMavenLocal -Dbuild.version_qualifier=rc1 -Dbuild.snapshot=false
+      run: ./gradlew publishToMavenLocal -Dbuild.version_qualifier=snapshot -Dbuild.snapshot=false
 
     - name: Checkstyle
       run: mvn -B checkstyle:checkstyle
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,7 @@ plugins {
 buildDir = 'gradle-build'
 
 ext {
-    opensearchVersion = '1.0.0.0-rc1'
+    opensearchVersion = '1.0.0.0-beta1'
     isSnapshot = "true" == System.getProperty("build.snapshot", "true")
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,7 @@ plugins {
 buildDir = 'gradle-build'
 
 ext {
-    opensearchVersion = '1.0.0.0-rc1'
+    opensearchVersion = '1.0.0.0-snapshot'
     isSnapshot = "true" == System.getProperty("build.snapshot", "true")
 }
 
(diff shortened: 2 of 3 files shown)


dblock (approved):

andy840314 (dismissed):

.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
     - name: Build OpenSearch
       working-directory: ./OpenSearch
-      run: ./gradlew publishToMavenLocal -Dbuild.snapshot=false
+      run: ./gradlew publishToMavenLocal -Dbuild.version_qualifier=rc1 -Dbuild.snapshot=false
```
vrozov, line 48: What makes this `rc1`? The checkout references `1.x` branch that is a moving target.

build.gradle
```diff
@@ -47,6 +47,7 @@ group = "org.opensearch"
 version = "${opensearchVersion}" + (isSnapshot ? "-SNAPSHOT" : "")
 
 if (!project.hasProperty("archivePath")) {
+
```
vrozov, line 50: undo

vrozov (changes requested):

.github/workflows/ci.yml
cliu123: Yes, CI checkout `1.x` branch of OpenSearch, but the version of OpenSearch core is `1.0.0-rc1` for RC, so plugins CI also builds `opensearch-1.0.0-rc1` and plugins build on it. This is an agreement between plugins and OpenSearch core, for example [AD](https://github.com/opensearch-project/anomaly-detection/blob/0ec9ef3dafdc61d97e929d8f0fe6644f4e372b93/.github/workflows/CI.yml#L39) and [Index Management](https://github.com/opensearch-project/index-management/blob/72018d48588e51ef1beb728d8f88f0245492704d/.github/workflows/test-and-build-workflow.yml#L36) that have already made the changes.

build.gradle
cliu123: Done.

.github/workflows/ci.yml
vrozov: @cliu123 Please see my comment https://github.com/opensearch-project/security/pull/1176#discussion_r637427860 regarding difference between `SNAPSHOT` and other tags. Either CI should use https://github.com/opensearch-project/OpenSearch/releases/tag/1.0.0-rc1 or it should use `SNAPSHOT`. Mixing both is confusing.

cliu123: As @dblock mentioned on [#1176](https://github.com/opensearch-project/security/pull/1176#discussion_r637427860): "in order for the plugin to start the version of the engine and the plugin will have to match, and the next version is rc1."
I don't really see any potential risks to build OpenSearch with the qualifier `rc1` that is consistent with the versions of both OpenSearch core and plugins. It would help to keep consistent across all plugins unless there are risks to do so.

vrozov: `rc1` must correspond to the actual rc1 candidate of OpenSearch, not just to a random/latest commit on the 1.x branch.

vrozov (changes requested):

.github/workflows/ci.yml
dblock: @vrozov You're right, what should really happen is that we take an XYZ build of OpenSearch, test it, and label it as `rc1` _after_ we're done. However today this is proving itself to be difficult because versions must match across OpenSearch and plugins for everything to start together, including the `rc1` qualifier, if any. So it's a chicken and egg problem. I described what we _should_ be doing in https://github.com/opensearch-project/OpenSearch/issues/763, please add your comments.

For this PR I suggest letting it slide to make rc1 because I don't think we can fix 763 in time. But for rc2 or afterwards we should, I'll followup on getting that assigned.

vrozov: @dblock I don't see a chicken and egg problem here and there was no such problem with the ODFE releases: ES releases first, plugins use released ES artifacts at build, test and runtime. The same process should be followed with Opensearch: [`rc1`](https://github.com/opensearch-project/OpenSearch/releases/tag/1.0.0-rc1) is tagged  (best will be to publish artifacts) and plugins should use the tag. Once `rc2` is available on Opensearch, plugins should update to that tag. Using latest commit on `1.x` branch, but pretending that it is `rc1` makes security plugin build unstable and more importantly testing of security plugin would be done against latest commit, not `rc1`, so it can't guarantee that plugin is compatible with the actual `rc1` Opensearch.

dblock: I agree.

dblock: @MMakaria: this is an ask to effectively change the release train sequence, I had brought it up before

dblock (approved):

.github/workflows/ci.yml
cliu123: Changing that would make security plugin built on different version/commits of OpenSearch core from other plugins. Let's have a decision before changing this. Whatever the decision is going to be, all plugins will need to take the same action or to stick with building plugins on top of `1.x` branch of OpenSearch core.

dblock:
If OpenSearch was built with -rc1 and this is built with -snapshot I think the plugin will fail to start as the versions must match exactly. See https://github.com/opensearch-project/OpenSearch/issues/763.

dblock (approved):
OpenSearch is building with `rc1`, so this needs to be `rc1`.

camerski (approved):

PR #1195, opened 2021-05-27
Extended role injection support for cross cluster requests

saikaranam-amazon:

 
- Enhancement
 
 
N/A

- Extended role injection support for cross cluster requests.
- Added audit logs for role injection

- Extends role injection for cross cluster requests

 - Added Integration tests and tested locally
 
 
N/A

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/src/main/java/org/opensearch/security/auth/RolesInjector.java
+++ b/src/main/java/org/opensearch/security/auth/RolesInjector.java
@@ -15,6 +15,7 @@
 
 package org.opensearch.security.auth;
 
+import org.opensearch.security.auditlog.AuditLog;
 import org.opensearch.security.support.ConfigConstants;
 import org.opensearch.security.user.User;
 import com.google.common.collect.ImmutableSet;
@@ -22,6 +23,8 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.tasks.Task;
+import org.opensearch.transport.TransportRequest;
 
 import java.util.Set;
 
@@ -35,12 +38,13 @@ import java.util.Set;
  */
 final public class RolesInjector {
     protected final Logger log = LogManager.getLogger(RolesInjector.class);
+    private final AuditLog auditLog;
 
-    public RolesInjector() {
-        //empty
+    public RolesInjector(AuditLog auditLog) {
+        this.auditLog = auditLog;
     }
 
-    public Set<String> injectUserAndRoles(final ThreadContext ctx) {
+    public Set<String> injectUserAndRoles(TransportRequest transportRequest, String action, Task task, final ThreadContext ctx) {
         final String injectedUserAndRoles = ctx.getTransient(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROLES);
         if (injectedUserAndRoles == null) {
             return null;
@@ -67,15 +71,17 @@ final public class RolesInjector {
         Set<String> roles = ImmutableSet.copyOf(strs[1].split(","));
 
         if(user != null && roles != null) {
-            addUser(user, ctx);
+            addUser(user, transportRequest, action, task, ctx);
         }
         return roles;
     }
 
-    private void addUser(final User user, final ThreadContext threadContext) {
+    private void addUser(final User user, final TransportRequest transportRequest,
+                         final String action, final Task task, final ThreadContext threadContext) {
         if(threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER) != null)
             return;
 
         threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_USER, user);
+        auditLog.logSucceededLogin(user.getName(), false, null, transportRequest, action, task);
     }
 }
--- a/src/main/java/org/opensearch/security/filter/SecurityFilter.java
+++ b/src/main/java/org/opensearch/security/filter/SecurityFilter.java
@@ -131,7 +131,7 @@ public class SecurityFilter implements ActionFilter {
         this.compatConfig = compatConfig;
         this.indexResolverReplacer = indexResolverReplacer;
         this.immutableIndicesMatcher = WildcardMatcher.from(settings.getAsList(ConfigConstants.SECURITY_COMPLIANCE_IMMUTABLE_INDICES, Collections.emptyList()));
-        this.rolesInjector = new RolesInjector();
+        this.rolesInjector = new RolesInjector(auditLog);
         this.backendRegistry = backendRegistry;
         log.info("{} indices are made immutable.", immutableIndicesMatcher);
     }
@@ -171,7 +171,7 @@ public class SecurityFilter implements ActionFilter {
             if (complianceConfig != null && complianceConfig.isEnabled()) {
                 attachSourceFieldContext(request);
             }
-            final Set<String> injectedRoles = rolesInjector.injectUserAndRoles(threadContext);
+            final Set<String> injectedRoles = rolesInjector.injectUserAndRoles(request, action, task, threadContext);
             boolean enforcePrivilegesEvaluation = false;
             User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
             if(user == null && (user = backendRegistry.authenticate(request, null, task, action)) != null) {
--- a/src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
+++ b/src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
@@ -148,10 +148,16 @@ public class SecurityRequestHandler<T extends TransportRequest> extends Security
             //bypass non-netty requests
             if(channelType.equals("direct")) {
                 final String userHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER);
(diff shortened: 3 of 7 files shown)


src/main/java/org/opensearch/security/auth/UserInjector.java
```diff
@@ -57,14 +59,13 @@
     private final ThreadPool threadPool;
     private final AuditLog auditLog;
     private final XFFResolver xffResolver;
-    private final Boolean injectUserEnabled;
+    private final Boolean injectRESTUserEnabled;
```
vrozov, line 62: I don't see a need to restrict this just to REST API. User injection is either globally enabled or globally disabled on the cluster.

saikaranam-amazon: We don't want to explicitly block on the user injection support (as with role injection) for the transport requests. Note that, this setting was enabled previously for only the REST requests and we added support for transport request afterwards.

With other plugins depending on user injection and using this settings for the transport requests , this should be enabled by default.

vrozov: It may be enabled by default, but cluster admin must have an option not to trust user injection from other plugins/nodes. I don't see a need to have a separate settings for REST and transport that would allow to disable one type of injection and enable another.

saikaranam-amazon: I will move the setting to be enabled by default and move the everything under this setting. Let me know, If this addresses your concern.

vrozov: The default should be the most secure option. On a preconfigured cluster that has multiple plugins installed, the settings can be explicitly enabled.

saikaranam-amazon:
Ping - @vrozov 
Addressed your comments. Let me know, If anything needs an update.

src/main/java/org/opensearch/security/auth/RolesInjector.java
```diff
@@ -67,15 +71,17 @@ public RolesInjector() {
...
     }
 
-    private void addUser(final User user, final ThreadContext threadContext) {
+    private void addUser(final User user, final TransportRequest transportRequest,
+                         final String action, final Task task, final ThreadContext threadContext) {
         if(threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER) != null)
             return;
 
         threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_USER, user);
+        auditLog.logSucceededLogin(user.getName(), true, null, transportRequest, action, task);
```
vrozov, line 85: Should `securityadmin` be set to `false`? AFAIK, only for certificate user it should be true.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -348,9 +348,13 @@ public void onFailure(Exception e) {
                 }
             } else {
                 auditLog.logMissingPrivileges(action, request, task);
-                String err = (injectedRoles == null) ?
-                        String.format("no permissions for %s and %s", pres.getMissingPrivileges(), user) :
-                        String.format("no permissions for %s and associated roles %s", pres.getMissingPrivileges(), injectedRoles);
+                String err = String.format("no permissions for %s and %s", pres.getMissingPrivileges(), user);
+                if(!pres.getMissingSecurityRoles().isEmpty()) {
+                    err = String.format("no permissions for user %s to assume roles %s", user, pres.getMissingSecurityRoles());
```
vrozov, line 353: should be consistent with other error messages "no permissions for %s ... "

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -348,9 +348,13 @@ public void onFailure(Exception e) {
...
+                }
```
vrozov, line 354:
```suggestion
                } else {
                    err = (injectedRoles == null) ? ...
```

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -206,14 +208,26 @@ public PrivilegesEvaluatorResponse evaluate(final User user, String action0, fin
...
 
+        final PrivilegesEvaluatorResponse presponse = new PrivilegesEvaluatorResponse();
+
         final TransportAddress caller = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
-        final Set<String> mappedRoles = (injectedRoles == null) ? mapRoles(user, caller) : injectedRoles;
+        Set<String> mappedRoles = (injectedRoles == null) ? mapRoles(user, caller) : injectedRoles;
+        final String assumeRolesString = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_ASSUME_ROLES);
+        if(assumeRolesString != null) {
+            HashSet<String> assumeRoleSet = new HashSet<>(Arrays.asList(assumeRolesString.split(",")));
+            if(!mappedRoles.containsAll(assumeRoleSet)) {
```
vrozov, line 218: I don't follow that logic. What is the significance of comparing role names? What if I have two roles with the same name but with different permissions? What if two roles have different names but the same permissions? Please open new issue and describe use case and how that use case will be solved.

saikaranam-amazon: Added here: https://github.com/opensearch-project/security/issues/1337

vrozov: What this check enforce and why is it needed?

saikaranam-amazon: The mappedRoles contains all the roles associated with a particular user. The `assumeRoleSet ` transients populated by plugins can force the evaluation to restrict on `assumeRolesSet`. With this:
 - The user is confirmed to be associated with these roles(assumeRolesSet) and
 -  If the permissions evaluation succeeds, the assume roles have the required permissions to invoke transport actions

If either assumeRolesSet(from the transients) is not associated with user or if there are these roles associated with the user but doesn't have permissions to invoke transport actions, then security exception is thrown.

vrozov: I see, so this change is not directly related to the role injection? If yes, please split PR into two or more (if there are more independent changes) separate PRs.

saikaranam-amazon: Done: https://github.com/opensearch-project/security/pull/1367

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
```diff
@@ -142,10 +142,14 @@ protected void messageReceivedDecorate(final T request, final TransportRequestHa
...
                 final String userHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER);
+                final String injectedRolesHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROLES_HEADER);
                 final String injectedUserHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER_HEADER);
 
                 if(Strings.isNullOrEmpty(userHeader)) {
-                    if(!Strings.isNullOrEmpty(injectedUserHeader)) {
+                    if(!Strings.isNullOrEmpty(injectedRolesHeader)) {
+                        getThreadContext().putTransient(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROLES, injectedRolesHeader);
+                    }
+                    else if(!Strings.isNullOrEmpty(injectedUserHeader)) {
```
vengadanathan-s, line 152: When both user and role are injected what is the behaviour ? should we fail ?

vengadanathan-s (approved):

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
saikaranam-amazon: Currently, user injection is controlled via configuration flag (false by default) and we are taking precedence on role injection.

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
```diff
@@ -142,10 +142,14 @@ protected void messageReceivedDecorate(final T request, final TransportRequestHa
             //bypass non-netty requests
             if(channelType.equals("direct")) {
                 final String userHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER);
+                final String injectedRolesHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROLES_HEADER);
                 final String injectedUserHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER_HEADER);
 
                 if(Strings.isNullOrEmpty(userHeader)) {
-                    if(!Strings.isNullOrEmpty(injectedUserHeader)) {
```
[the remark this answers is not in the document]
saikaranam-amazon: We don't have user injection directly enabled for the plugins and all the current plugins under OpenSearch will be using role injection and there shouldn't be any clash between these two headers

saikaranam-amazon:
> possible to update/add test cases for cross cluster request for role injection?

Added tests for Cross cluster role injection

hardik-k-shah (approved):

cliu123 (approved):

PR #1158, opened 2021-05-04
Return HTTP 409 if get parallel put request

andy840314:
##  opendistro-for-elasticsearch/security pull request intake form
_Please provide as much details as possible to get feedback/acceptance on your PR quickly_

 
 1. __Category:__ (_Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation_)
 Bug fix
 
 
 2. __Github Issue # or road-map entry, if available:__
#1095 

 3. __Description of changes:__

 4. __Why these changes are required?__ 

 5. __What is the old behavior before changes and new behavior after changes?__ (_Please add any example/logs/screen-shot if available_)

Send parallel put/delete requests

before: 
returns
`{"status":"INTERNAL_SERVER_ERROR","message":"Error [internalusers]: version conflict, ......`

after: 
returns
`{"status":"CONFLICT","message":"[internalusers]: version conflict, ......`

 6. __Testing done:__ (_Please provide details of testing done: Unit testing, integration testing and manual testing_)
 
Unit test

Manual testing with Curl
 
```
PUT _opendistro/_security/api/internalusers/test1
{
  "password": "test1"
}
PUT _opendistro/_security/api/internalusers/test1
{
  "password": "test1"
}
```
returns
```
HTTP/1.1 409 Conflict
content-type: application/json; charset=UTF-8
content-length: 160

{"status":"CONFLICT","message":"[internalusers]: version conflict, required seqNo [29], primary term [6]. current document has seqNo [30] and primary term [7]"}
HTTP/1.1 200 OK
content-type: application/json; charset=UTF-8
content-length: 44

{"status":"OK","message":"'test1' updated."}
```
 
 7. __TO-DOs, if any:__ (_Please describe pending items and provide Github issues# for each of them_)

 8. __Is it backport from main branch?__ (_If yes, please add backport PR # and commits #_)

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
[...]

PR #1331, opened 2021-07-20
Fix race condition on async test for PR #1158

andy840314:

 
 Test fix
 
 
    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
+++ b/src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
@@ -93,7 +93,6 @@ public class UserApiTest extends AbstractRestApiUnitTest {
                     created = true;
                     break;
                 case HttpStatus.SC_OK:
-                    Assert.assertTrue(created);
                     break;
                 default:
                     Assert.assertEquals(HttpStatus.SC_CONFLICT, sc);
--- a/src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
+++ b/src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
@@ -87,11 +87,17 @@ public class UserApiTest extends AbstractRestApiUnitTest {
         boolean created = false;
         for (HttpResponse response : responses) {
             int sc = response.getStatusCode();
-            if (sc == HttpStatus.SC_CREATED) {
-                Assert.assertFalse(created);
-                created = true;
-            } else {
-                Assert.assertTrue("Status code should be 409 or 200", sc == HttpStatus.SC_CONFLICT || sc == HttpStatus.SC_CREATED);
+            switch (sc) {
+                case HttpStatus.SC_CREATED:
+                    Assert.assertFalse(created);
+                    created = true;
+                    break;
+                case HttpStatus.SC_OK:
+                    Assert.assertTrue(created);
+                    break;
+                default:
+                    Assert.assertEquals(HttpStatus.SC_CONFLICT, sc);
+                    break;
             }
         }
         deleteUser("test1");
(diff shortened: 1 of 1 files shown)


src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -84,15 +84,9 @@ public void testParallelPutRequests() throws Exception {
...
-        boolean created = false;
         for (HttpResponse response : responses) {
             int sc = response.getStatusCode();
-            if (sc == HttpStatus.SC_CREATED) {
-                Assert.assertFalse(created);
-                created = true;
-            } else {
-                Assert.assertEquals(HttpStatus.SC_CONFLICT, sc);
-            }
+            Assert.assertTrue("Status code should be 409 or 200", sc == HttpStatus.SC_CONFLICT || sc == HttpStatus.SC_CREATED);
```
vrozov, line 89: The original logic must stay. The return code is SC_CREATED once, otherwise it is either SC_CONFLICT or SC_OK.

andy840314: Reverted.

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -91,7 +91,7 @@ public void testParallelPutRequests() throws Exception {
                 Assert.assertFalse(created);
                 created = true;
             } else {
-                Assert.assertEquals(HttpStatus.SC_CONFLICT, sc);
+                Assert.assertTrue("Status code should be 409 or 200", sc == HttpStatus.SC_CONFLICT || sc == HttpStatus.SC_CREATED);
```
vrozov, line 94: This is not a correct condition (should be `SC_OK` instead of `SC_CREATED`). IMO, it will be better not to mix 409 and 200 codes together, consider using `switch(sc)` with `SC_CREATED`, `SC_OK` and `default` cases.

andy840314: Yes, `SC_OK` should be seen only after `SC_CREATED`. `SC_CONFLICT` will be in the `default` case. Fixed.

vrozov: I was not referring to the order of `SC_OK`, especially that order in which requests are processed is not necessarily the same as `requests` order.

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -87,11 +87,17 @@ public void testParallelPutRequests() throws Exception {
...
-                created = true;
-            } else {
-                Assert.assertEquals(HttpStatus.SC_CONFLICT, sc);
+            switch (sc) {
+                case HttpStatus.SC_CREATED:
+                    Assert.assertFalse(created);
+                    created = true;
+                    break;
+                case HttpStatus.SC_OK:
+                    Assert.assertTrue(created);
```
vrozov, line 96: you can't assert that

andy840314: ok, its because async request won't necessarily return in order right? Then what should I put in this case? It doesn't make sense to put `Assert.assertEquals(HttpStatus.SC_OK, sc);` here since the condition is already `case HttpStatus.SC_OK:`. Or just delete this line?

vrozov: It is because thread pool does not provide a guarantee on task order execution and there is no guarantee of what request is processed first. There is no synchronization/order being enforced. Just delete the assert.

vrozov (approved):

hsiang9431-amzn (approved):

Issue #1403, opened 2021-08-13
[BUG] Snapshot build doesn't produce a snapshot build or an artifact named with -SNAPSHOT


dblock:
**Describe the bug**

Coming from https://github.com/opensearch-project/opensearch-build/pull/189, attempting to build a snapshot version of this plugin:

```
mvn -B clean package -Padvanced -DskipTests -Dopensearch.version=1.0.0-SNAPSHOT -Dbuild.snapshot=true
```

produces `target/releases/opensearch-security-1.0.0.0.zip`, I expected it to produce `opensearch-security-1.0.0.0-SNAPSHOT.zip` like other plugins.

or for it to get renamed in output from assembly step?

```
./gradlew build buildDeb buildRpm --no-daemon -ParchivePath=target/releases/opensearch-security-1.0.0.0.zip -Dbuild.snapshot=true
```

**Expected behavior**

Produce `opensearch-security-1.0.0.0-SNAPSHOT.zip` and any other artifact to include `-SNAPSHOT` as well as internal version to include SNAPSHOT so that we don't accidentally promote snapshot builds to release.

vrozov:
@dblock `-Dbuild.snapshot=true` is a gradle option, not a maven option. Passing it to maven does not make maven to build snapshot.

dblock:
> @dblock `-Dbuild.snapshot=true` is a gradle option, not a maven option. Passing it to maven does not make maven to build snapshot.

How do I build a snapshot version of the security plugin?

dblock:
With changes in https://github.com/opensearch-project/security/pull/1409, building a `-SNAPSHOT` version against OpenSearch 1.1. works like so:

```
mvn -B clean package -Padvanced -DskipTests -Dopensearch.version=1.1.0-SNAPSHOT
artifact_zip=$(ls $(pwd)/target/releases/opensearch-security-*.zip | grep -v admin-standalone)
./gradlew assemble --no-daemon -ParchivePath=$artifact_zip -Dopensearch.version=1.1.0-SNAPSHOT
```

dblock:
#1409 was closed in favor of https://github.com/opensearch-project/opensearch-build/pull/356

Issue #1400, opened 2021-08-11
[BUG] Plugin does not build


smortex:
**Describe the bug**
In order to attempt to fix a bug, I would like to rebuild the java part of the security plugin and drop it in place of the one bundled in the opensearch-1.0.0 archive.

**To Reproduce**
1. Clone this repo
2. Run `mvn clean test` as advised in the README.

**Expected behavior**
mvn find all dependencies and builds the plugin

**Actual result**

Some dependencies cannot be found:

```sh-session
romain@optiplex-3050 ~/Projects/security % mvn clean test
[...]
[INFO] -----------------< org.opensearch:opensearch-security >-----------------
[INFO] Building OpenSearch Security 1.0.0.0
[INFO] --------------------------------[ jar ]---------------------------------
[WARNING] The POM for org.opensearch.plugin:transport-netty4-client:jar:1.0.0 is missing, no dependency information available
[WARNING] The POM for org.opensearch:opensearch:jar:1.0.0 is missing, no dependency information available
[WARNING] The POM for org.opensearch.client:opensearch-rest-high-level-client:jar:1.0.0 is missing, no dependency information available
[WARNING] The POM for org.opensearch.plugin:reindex-client:jar:1.0.0 is missing, no dependency information available
[WARNING] The POM for org.opensearch:opensearch-ssl-config:jar:1.0.0 is missing, no dependency information available
[WARNING] The POM for org.opensearch.plugin:percolator-client:jar:1.0.0 is missing, no dependency information available
[WARNING] The POM for org.opensearch.plugin:lang-mustache-client:jar:1.0.0 is missing, no dependency information available
[WARNING] The POM for org.opensearch.plugin:parent-join-client:jar:1.0.0 is missing, no dependency information available
[WARNING] The POM for org.opensearch.plugin:aggs-matrix-stats-client:jar:1.0.0 is missing, no dependency information available
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  2.144 s
[INFO] Finished at: 2021-08-10T19:12:39-10:00
[INFO] ------------------------------------------------------------------------
[...]

PR #1335, opened 2021-07-23
Upgrade OpenSearch version to 1.1.0

cliu123:
##  opensearch-security pull request intake form
_Please provide as much details as possible to get feedback/acceptance on your PR quickly_

 
 1. __Category:__ (_Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation_)
 Maintenance
 
 
 2. __Github Issue # or road-map entry, if available:__

 3. __Description of changes:__

 4. __Why these changes are required?__ 

 5. __What is the old behavior before changes and new behavior after changes?__ (_Please add any example/logs/screen-shot if available_)

 6. __Testing done:__ (_Please provide details of testing done: Unit testing, integration testing and manual testing_)
 

- Built security plugin with OpenSearch 1.1.0
- Installed security plugin on OpenSearch 1.1.0
- [CD run](https://github.com/cliu123/security/runs/3406281257?check_suite_focus=true)

 
 
 7. __TO-DOs, if any:__ (_Please describe pending items and provide Github issues# for each of them_)

 8. __Is it backport from main branch?__ (_If yes, please add backport PR # and commits #_)

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1409, opened 2021-08-17
Add support for -Dbuild.snapshot=true/false and default to building snapshot.

dblock:
Signed-off-by: dblock <dblock@amazon.com>

 Maintenance
 
https://github.com/opensearch-project/security/issues/1403

- Default to building a -SNAPSHOT version of this plugin.
- Adds support for passing `-Dbuild.snapshot=false` to not build a snapshot. 

The downside is that the pom.xml is incorrect for -SNAPSHOT build (contains non-snapshot version because pom.xml doesn't know anything about profiles).

Building the end-to-end bundle needs the artifacts to be able to build snapshot vs. not snapshot.

The build could only produce a snapshot version vs. non-snapshot.

#### Snapshot (default)

```
mvn -B clean package -Padvanced -DskipTests

[INFO] -----------------< org.opensearch:opensearch-security >-----------------
[INFO] Building OpenSearch Security 1.1.0.0-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------

...

[INFO] --- maven-jar-plugin:3.1.1:jar (default-jar) @ opensearch-security ---
[INFO] Building jar: /home/ubuntu/source/opensearch-project/security/dblock-security/target/opensearch-security-1.1.0.0-SNAPSHOT.jar
[INFO]
[INFO] --- git-commit-id-plugin:2.2.6:validateRevision (validate-the-git-infos) @ opensearch-security ---
[INFO]
[INFO] --- maven-jar-plugin:3.1.1:test-jar (default) @ opensearch-security ---
[INFO] Building jar: /home/ubuntu/source/opensearch-project/security/dblock-security/target/opensearch-security-1.1.0.0-SNAPSHOT-tests.jar
[INFO]
[INFO] --- maven-assembly-plugin:3.1.1:single (plugin) @ opensearch-security ---
[INFO] Reading assembly descriptor: /home/ubuntu/source/opensearch-project/security/dblock-security/src/main/assemblies/plugin.xml
[INFO] Building zip: /home/ubuntu/source/opensearch-project/security/dblock-security/target/releases/opensearch-security-1.1.0.0-SNAPSHOT.zip
[INFO] 
[INFO] --- maven-assembly-plugin:3.1.1:single (securityadmin) @ opensearch-security ---
[INFO] Reading assembly descriptor: /home/ubuntu/source/opensearch-project/security/dblock-security/src/main/assemblies/securityadmin-standalone.xml
[WARNING] The following patterns were never triggered in this artifact exclusion filter:
o  'org.opensearch:dlic*'

[INFO] Building zip: /home/ubuntu/source/opensearch-project/security/dblock-security/target/releases/opensearch-security-1.1.0.0-SNAPSHOT-securityadmin-standalone.zip
[WARNING] The following patterns were never triggered in this artifact exclusion filter:
o  'org.opensearch:dlic*'

[INFO] Building tar: /home/ubuntu/source/opensearch-project/security/dblock-security/target/releases/opensearch-security-1.1.0.0-SNAPSHOT-securityadmin-standalone.tar.gz
[INFO] 
[INFO] --- checksum-maven-plugin:1.7.1:files (default) @ opensearch-security ---
[INFO] opensearch-security-1.1.0.0-SNAPSHOT.zip - SHA-512 : e9f9434e058ddfeb3be9468889afcc46cf031c9591b79317fd0cb60cc16e6a5eda5951bfb86a5a1c1551af2b29b1e37f7bff1976c918521bdd987f888bbe73ab
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------

$ unzip -p target/opensearch-security-1.1.0.0-SNAPSHOT.jar META-INF/MANIFEST.MF
Manifest-Version: 1.0
Created-By: Apache Maven 3.6.3
Build-Jdk: 14.0.2
Implementation-Title: OpenSearch Security
Implementation-Version: 1.1.0.0-SNAPSHOT <-------------------------------------------
Implementation-Vendor-Id: org.opensearch
Implementation-URL: https://github.com/opensearch-project/security
Build-Time: 2021-08-30T22:05:33Z
Built-By: Amazon OpenDistro ElasticSearch Security Parent
git-sha1: 7d3518f77f7e614e7bec8523b2c867b34f1b67d6

```

#### Not Snapshot

```
mvn -B clean package -Padvanced -DskipTests -Dbuild.snapshot=false

[INFO] -----------------< org.opensearch:opensearch-security >-----------------
[INFO] Building OpenSearch Security 1.1.0.0
[INFO] --------------------------------[ jar ]---------------------------------

...

[INFO] --- maven-jar-plugin:3.1.1:test-jar (default) @ opensearch-security ---
[INFO] Building jar: /home/ubuntu/source/opensearch-project/security/dblock-security/target/opensearch-security-1.1.0.0-tests.jar
[INFO] 
[INFO] --- maven-assembly-plugin:3.1.1:single (plugin) @ opensearch-security ---
[INFO] Reading assembly descriptor: /home/ubuntu/source/opensearch-project/security/dblock-security/src/main/assemblies/plugin.xml
[INFO] Building zip: /home/ubuntu/source/opensearch-project/security/dblock-security/target/releases/opensearch-security-1.1.0.0.zip
[INFO] 
[INFO] --- maven-assembly-plugin:3.1.1:single (securityadmin) @ opensearch-security ---
[INFO] Reading assembly descriptor: /home/ubuntu/source/opensearch-project/security/dblock-security/src/main/assemblies/securityadmin-standalone.xml
[WARNING] The following patterns were never triggered in this artifact exclusion filter:
o  'org.opensearch:dlic*'

[INFO] Building zip: /home/ubuntu/source/opensearch-project/security/dblock-security/target/releases/opensearch-security-1.1.0.0-securityadmin-standalone.zip
[WARNING] The following patterns were never triggered in this artifact exclusion filter:
o  'org.opensearch:dlic*'

[INFO] Building tar: /home/ubuntu/source/opensearch-project/security/dblock-security/target/releases/opensearch-security-1.1.0.0-securityadmin-standalone.tar.gz
[INFO] 
[INFO] --- checksum-maven-plugin:1.7.1:files (default) @ opensearch-security ---
[INFO] opensearch-security-1.1.0.0.zip - SHA-512 : 57fa1109843795d12f0c56b51e3352b1e50aa9bdaf4dbf05bb17517019c46f6f95774f0e3d020a422a9b7692dc2e4cb9e19d561101dbe0a383f226701dc4ada0
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------

$ unzip -p target/opensearch-security-1.1.0.0.jar META-INF/MANIFEST.MF
Manifest-Version: 1.0
Created-By: Apache Maven 3.6.3
Build-Jdk: 14.0.2
Implementation-Title: OpenSearch Security
Implementation-Version: 1.1.0.0 <------------------------------------------
Implementation-Vendor-Id: org.opensearch
Implementation-URL: https://github.com/opensearch-project/security
Build-Time: 2021-08-30T22:02:25Z
Built-By: Amazon OpenDistro ElasticSearch Security Parent
git-sha1: 7d3518f77f7e614e7bec8523b2c867b34f1b67d6

```

In a previous iteration of this PR, installed `opensearch-security` plugin on OpenSearch 1.1.0, and started the cluster.
 
Use the profile to depend on OpenSearch 1.1.0-SNAPSHOT as well.

No

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
 
     - name: Build OpenSearch
       working-directory: ./OpenSearch
-      run: ./gradlew publishToMavenLocal -Dbuild.snapshot=false
+      run: ./gradlew publishToMavenLocal
 
     - name: Checkstyle
       run: mvn -B checkstyle:checkstyle
--- a/plugin-descriptor.properties
+++ b/plugin-descriptor.properties
@@ -3,7 +3,7 @@
 description=Provide access control related features for OpenSearch 1.0.0
 #
 # 'version': plugin's version
-version=1.1.0.0-SNAPSHOT
+version=1.1.0.0
 #
 # 'name': the plugin name
 name=opensearch-security
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@
     <groupId>org.opensearch</groupId>
     <artifactId>opensearch-security</artifactId>
     <packaging>jar</packaging>
-    <version>1.1.0.0-SNAPSHOT</version>
+    <version>${version}</version>
     <name>OpenSearch Security</name>
     <description>OpenSearch Security</description>
     <url>https://github.com/opensearch-project/security</url>
@@ -69,6 +69,7 @@
         <maven.compiler.release>8</maven.compiler.release>
 
         <opensearch.version>1.1.0</opensearch.version>
+        <version>1.1.0.0</version>
 
         <!-- deps -->
         <netty-native.version>2.0.25.Final</netty-native.version>
@@ -831,6 +832,18 @@
         </resources>
     </build>
     <profiles>
+        <profile>
+            <activation>
+                <property>
+                    <name>build.snapshot</name>
+                    <value>!false</value>
+                </property>
+            </activation>
+            <properties>
+                <version>1.1.0.0-SNAPSHOT</version>
+                <opensearch.version>1.1.0.0-SNAPSHOT</opensearch.version>
+            </properties>
+        </profile>
         <profile>
             <id>advanced</id>
             <build>
--- a/build.gradle
+++ b/build.gradle
@@ -38,6 +38,7 @@ buildDir = 'gradle-build'
 
 ext {
     securityPluginVersion = '1.1.0.0'
+    opensearchVersion = System.getProperty("opensearch.version", "1.1.0-SNAPSHOT")
     isSnapshot = "true" == System.getProperty("build.snapshot", "true")
 }
 
@@ -67,7 +68,7 @@ ospackage {
     user 'root'
     permissionGroup 'root'
 
-    requires('opensearch-oss', "1.1.0", EQUAL)
+    requires('opensearch-oss', opensearchVersion - '-SNAPSHOT', EQUAL)
     packager = 'Amazon'
     vendor = 'Amazon'
     os = 'LINUX'


vrozov (changes requested):
Building SNAPSHOT must be a separate PR from upgrade to OpenSearch 1.1, especially that there is already open PR #1335.

dblock:
I'll be happy to rebase this on top of https://github.com/opensearch-project/security/pull/1335 when it's merged.

dblock:
There's probably some more that can be done here to make `plugin-descriptor.properties` auto-populate these versions.

cliu123:
@dblock With [this PR](https://github.com/opensearch-project/security/pull/1335) merged, the `main` branch now support OpenSearch `1.1`. Please rebase this branch against `main` branch.

dblock:
This is ready to review @vrozov @cliu123.

plugin-descriptor.properties
```diff
@@ -3,7 +3,7 @@
 description=Provide access control related features for OpenSearch 1.0.0
 #
 # 'version': plugin's version
-version=1.0.1.0
+version=1.1.0.0
```
cliu123, line 6: This is security plugin version, not the OpenSearch version. If the goal of this PR is to add `-SNAPSHOT` suffix in the OpenSearch version, this line doesn't need to be changed.

pom.xml
```diff
@@ -68,7 +68,8 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <maven.compiler.release>8</maven.compiler.release>
 
-        <opensearch.version>1.1.0</opensearch.version>
+        <opensearch.version>1.1.0-SNAPSHOT</opensearch.version>
+        <version>1.1.0.0</version>
```
cliu123, line 72: This is referenced on the line 37 that is security plugin version. Same as [this commend](https://github.com/opensearch-project/security/pull/1409/files#r695406712), doesn't seem to need a change at here.

pom.xml
```diff
@@ -103,7 +104,6 @@
         <url>https://github.com/opensearch-project/security</url>
         <connection>scm:git:git@github.com:opensearch-project/security.git</connection>
         <developerConnection>scm:git:git@github.com:opensearch-project/security.git</developerConnection>
-        <tag>1.0.1.0</tag>
```
cliu123, line 106: Why are we removing the tag here?

pom.xml
```diff
@@ -831,6 +831,17 @@
...
     <profiles>
+        <profile>
+            <activation>
+                <property>
+                    <name>build.snapshot</name>
+                    <value>!false</value>
+                </property>
+            </activation>
+            <properties>
+                <version>1.1.0.0-SNAPSHOT</version>
```
cliu123, line 842: Looks like this PR is bumping security version from `1.0.1.0` to `1.1.0.0`. If that's the case, is that the case?

plugin-descriptor.properties
dblock: The plugin version needs to be incremented because [we're building OpenSearch 1.1](https://github.com/opensearch-project/opensearch-build/blob/main/manifests/opensearch-1.1.0.yml). Every other plugin says it's version 1.1.0.0-SNAPSHOT.

pom.xml
dblock: Are you saying I need to make two separate PRs, one for the version bump and one for building snapshots by default?

pom.xml
dblock: What is the purpose of it here? I believe it's unused.

pom.xml
dblock: Yes, this is the case.

cliu123 (dismissed):

pom.xml
vrozov: `<url>`, `<connection>` and `<developerConnection>` are not used either. It is there mostly for historical reasons. I don't see a good reason to remove `<tag>` only.

pom.xml
vrozov: Must be consistent with `plugin-descriptor.properties`, I don't see how this can be done using profiles.

dblock:
@vrozov 

We failed at having a scalable mechanism to distribute maven credentials across 15 plugins, so whereas OpenSearch bundle build 1.0 picked up output from security plugin CD, it will not for 1.1. Bundle will build from the top, including opensearch-min and security plugin, from source. Given that, this change ensures that security CD doesn't build an artifact that can be accidentally promoted to a release, i.e. always depend on a snapshot version of OpenSearch and produce a snapshot version of security plugin. The change in profiles does that, defaults to snapshot build unless you explicitly specify `-Dbuild.snapshot=false`, this is net new.

The plugin built in bundle is version 1.1, not 1.0.1. Unlike in ODFE or 1.0, plugins are now aligning with OpenSearch: when development starts on version X in a branch 'X', the version in that branch is incremented to X.  

Given that, what do you need me to change in this PR to get this in?

vrozov:
@dblock 
- Security plugin CD pipeline does not publish anything to maven, it creates github draft release.
- Plugin version matches OpenSearch version in the first 3 digits. The last digit can be independently changed by the plugin.
- Changing dependency on OpenSearch from `1.1.0` to `1.1.0-SNAPSHOT` is a self contained change similar to how version was changed from `1.0.0` to `1.1.0` in #1335. Prerequisite of building OpenSearch locally or in CI/CD introduces lots of confusion and is error prone (#1400).
- I don't quite understand how profile can be used to switch between `SNAPSHOT` and release. That will allow to use any *random* commit to build release version that is not supposed to be the release version. Additionally, maven profile has no impact on `plugin-descriptor.properties`, so profile solution is not complete.

dblock:
> @dblock
> 
> * Security plugin CD pipeline does not publish anything to maven, it creates github draft release.

I was trying to explain how we ended up here. The CD pipeline output is not used by anyone other than dev at this point, and thus should be a snapshot build that depends on OpenSearch snapshot. 

> * Plugin version matches OpenSearch version in the first 3 digits. The last digit can be independently changed by the plugin.

Yes.

> * Changing dependency on OpenSearch from `1.1.0` to `1.1.0-SNAPSHOT` is a self contained change similar to how version was changed from `1.0.0` to `1.1.0` in [Upgrade OpenSearch version to 1.1.0 #1335](https://github.com/opensearch-project/security/pull/1335). Prerequisite of building OpenSearch locally or in CI/CD introduces lots of confusion and is error prone ([[BUG] Plugin does not build #1400](https://github.com/opensearch-project/security/issues/1400)).

I am not trying to change dependency from OpenSearch 1.1.0 to 1.1.0-SNAPSHOT. I am trying to make security plugin build against 1.1.0-SNAPSHOT by default, and 1.1.0 if `-Dbuild.snapshot=false` is specified. I can split these commits if it helps, but it will hide this fact. Just tell me what to do.

The pre-requisite of building OpenSearch locally will go away with https://github.com/opensearch-project/opensearch-build/issues/20.

> * I don't quite understand how profile can be used to switch between `SNAPSHOT` and release. That will allow to use any _random_ commit to build release version that is not supposed to be the release version. Additionally, maven profile has no impact on `plugin-descriptor.properties`, so profile solution is not complete.

You have two pipelines: 
- CI in this repo, needs to build snapshot
- Bundle pipeline, needs to build release

Plugins have had `-Dbuild.snapshot=true/false` forever, except security plugin. How do you propose we accomplish this for the security plugin?

I will deal with `plugin-descriptor.properties` when we settle simpler question.

vrozov:
> Plugins have had -Dbuild.snapshot=true/false forever, except security plugin. How do you propose we accomplish this for the security plugin?

Other plugins had this option as they utilize OpenSearch gradle extension. Security plugin is not built with gradle, it is built with maven and never supported `-Dbuild.snapshot`. I still don't see how that can be done in maven due to dependency issue and impact on `plugin-descriptor.properties`.

vrozov:
> I was trying to explain how we ended up here. The CD pipeline output is not used by anyone other than dev at this point, and thus should be a snapshot build that depends on OpenSearch snapshot.

While CD pipeline is no longer used to publish to maven and to S3 to produce distribution bundle, it is used to create **release** draft on github. There is no usage of CD pipeline outside of release build. For PR and builds on `main` CI is used.

dblock:
Extracted version increment into https://github.com/opensearch-project/security/pull/1429.

.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
     - name: Build OpenSearch
       working-directory: ./OpenSearch
-      run: ./gradlew publishToMavenLocal -Dbuild.snapshot=false
+      run: ./gradlew publishToMavenLocal
```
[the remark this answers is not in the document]
dblock: It does both at build time.

plugin-descriptor.properties
dblock: This properties file by default has a number without snapshot. Then, with a profile, we change it to be `-SNAPSHOT` unless `-Dbuild.snapshot=false`. You can't have the logical condition the other way around if you want the default to be `-Dbuild.snapshot=false`.

pom.xml
dblock: It had to move to properties or you cannot set it via a profile.

pom.xml
dblock: Update: by default the version comes from plugin.properties and is set to 1.1.0.0. This sets the version to 1.1.0.0-SNAPSHOT unless you specify `-Dbuild.snapshot=false`. Meaning it sets it by default. This effectively enables building both, snapshot or not snapshot, with snapshot by default.

If you put 1.1.0.0-SNAPSHOT in the properties file, then this profile condition would have to be inverted. But then the default can't be snapshot=true.

dblock:
A hacky-er alternative, https://github.com/opensearch-project/opensearch-build/pull/356.

vrozov:
Maven does not support using the same `pom.xml` to describe artifacts with two different versions (for example `1.1.0.0` and `1.1.0.0-SNAPSHOT`). The proposed solution only works to generate .jar and .zip artifacts during the build. AFAIK, such `pom.xml` with dual versions will be invalid should it be installed into local and/or remote maven repository.

@dblock I would vote for the other solution that you proposed https://github.com/opensearch-project/opensearch-build/pull/356

dblock:
I agree, closing in favor of https://github.com/opensearch-project/opensearch-build/pull/356.

PR #1430, opened 2021-08-31
Build against OpenSearch 1.1.1-SNAPSHOT.

dblock:
Signed-off-by: dblock <dblock@dblock.org>

Maintenace
 
 
Build against OpenSearch 1.1.0-SNAPSHOT.

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
 
     - name: Build OpenSearch
       working-directory: ./OpenSearch
-      run: ./gradlew publishToMavenLocal -Dbuild.snapshot=false
+      run: ./gradlew publishToMavenLocal
 
     - name: Checkstyle
       run: mvn -B checkstyle:checkstyle
--- a/plugin-descriptor.properties
+++ b/plugin-descriptor.properties
@@ -22,4 +22,4 @@ java.version=1.8
 # opensearch release. This version is checked when the plugin
 # is loaded so OpenSearch will refuse to start in the presence of
 # plugins with the incorrect opensearch.version.
-opensearch.version=1.1.0
+opensearch.version=1.1.0-SNAPSHOT
--- a/plugin-descriptor.properties
+++ b/plugin-descriptor.properties
@@ -17,9 +17,11 @@ classname=org.opensearch.security.OpenSearchSecurityPlugin
 # separated by "."'s and may have leading zeros
 java.version=1.8
 #
-# 'opensearch.version' version of elasticsearch compiled against
+# 'opensearch.version' version of OpenSearch compiled against
+# Does not include -SNAPSHOT or the plugin will not install,
+# see https://github.com/opensearch-project/opensearch-build/issues/359.
 # You will have to release a new version of the plugin for each new
-# opensearch release. This version is checked when the plugin
+# OpenSearch release. This version is checked when the plugin
 # is loaded so OpenSearch will refuse to start in the presence of
 # plugins with the incorrect opensearch.version.
-opensearch.version=1.1.0-SNAPSHOT
+opensearch.version=1.1.0
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,7 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <maven.compiler.release>8</maven.compiler.release>
 
-        <opensearch.version>1.1.0</opensearch.version>
+        <opensearch.version>1.1.0-SNAPSHOT</opensearch.version>
 
         <!-- deps -->
         <netty-native.version>2.0.25.Final</netty-native.version>


.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
     - name: Build OpenSearch
       working-directory: ./OpenSearch
-      run: ./gradlew publishToMavenLocal -Dbuild.snapshot=false
```
cliu123, line 48: Would you please also change in [CD](https://github.com/opensearch-project/security/blob/main/.github/workflows/cd.yml#L38)?

plugin-descriptor.properties
```diff
@@ -22,4 +22,4 @@ java.version=1.8
 # opensearch release. This version is checked when the plugin
 # is loaded so OpenSearch will refuse to start in the presence of
 # plugins with the incorrect opensearch.version.
-opensearch.version=1.1.0
```
cliu123, line 25: This cannot be changed. It needs to be remain `1.1.0`. Otherwise, security plugin cannot be installed on OpenSearch. This is a lesson learnt in the beta release.

> Exception in thread "main" java.lang.IllegalArgumentException: illegal version format - snapshot labels are not supported
	at org.opensearch.Version.fromStringSlow(Version.java:176)
	at org.opensearch.Version.fromString(Version.java:171)
	at org.opensearch.plugins.PluginInfo.readFromProperties(PluginInfo.java:197)
	at org.opensearch.plugins.InstallPluginCommand.loadPluginInfo(InstallPluginCommand.java:797)
	at org.opensearch.plugins.InstallPluginCommand.installPlugin(InstallPluginCommand.java:856)
	at org.opensearch.plugins.InstallPluginCommand.execute(InstallPluginCommand.java:263)
	at org.opensearch.plugins.InstallPluginCommand.execute(InstallPluginCommand.java:237)
	at org.opensearch.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:99)
	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:140)
	at org.opensearch.cli.MultiCommand.execute(MultiCommand.java:104)
	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:140)
	at org.opensearch.cli.Command.main(Command.java:103)
	at org.opensearch.plugins.PluginCli.main(PluginCli.java:60)

.github/workflows/ci.yml
vrozov: CD should not be changed. CD only runs against release version. This is `SNAPSHOT` version.

cliu123: In this case, the changes in this PR will be reverted upon release? Otherwise, CD would fail.

plugin-descriptor.properties
vrozov: @cliu123 Have you tested that against `SNAPSHOT` build of OpenSearch?

cliu123: Tested that against `opensearch-1.1.0-SNAPSHOT` to reproduce the above error.

.github/workflows/ci.yml
vrozov: It is expected CD to fail if it is accidentally runs against `SNAPSHOT`.

plugin-descriptor.properties
vrozov: @dblock @nknize Is it correct that `SNAPSHOT` version of OpenSearch expects plugin to use not `SNAPSHOT` version in the plugin descriptor specification?

dblock: 🤷 we should fix this if it's the case

dblock: I am not sure what the original intent of all of this was. I'd want to be able to install whatever on a snapshot build of OpenSearch.

dblock: Even if this is the case, I think we should change this as all the other plugins build -SNAPSHOT against -SNAPSHOT. If we cannot install those plugins it's a good thing, they would only be used for building and running tests?

dblock: I opened https://github.com/opensearch-project/opensearch-build/issues/359 to discuss.

I think you want a dependency on -SNAPSHOT anyway for CI, to recognize that the OpenSearch dependency is a moving target. But it seems problematic for developers because they cannot install the snapshot build in OpenSearch.

vrozov: This is not a build issue. The Opensearch library dependency is changed to `1.1.0.0-SNAPSHOT`. The question is only for plugin descriptor where Opensearch `1.1.0.0-SNAPSHOT` (not the build) expects `1.1.0.0` version.

cliu123: The `opensearch.version` looks nothing but a property in [plugin descriptor](https://www.elastic.co/guide/en/elasticsearch/plugins/current/plugin-authors.html#_mandatory_elements_for_plugins). `1.1.0.0` without `-SNAPSHOT` suffix should not cause problems.

vrozov: Question here is if plugin version is `1.1.0.0-SNAPSHOT` and it is compiled and tested against `1.1.0-SNAPSHOT` OpenSearch, is it valid to specify `1.1.0` in the plugin descriptor. That will allow to install `SNAPSHOT` version of plugin on the *release* version of OpenSearch.

dblock: The comment in https://github.com/opensearch-project/OpenSearch/blob/19e54d662488a183c1388e66cfe7463a24c6aa7c/buildSrc/src/main/resources/plugin-descriptor.properties#L48 is pretty clear that it wants the version of OpenSearch compiled against. So I think if we want to be able to install snapshot builds on OpenSearch-SNAPSHOT it needs to be supported in that code.

dblock: Can we please have the discussion in https://github.com/opensearch-project/opensearch-build/issues/359 first of what we want for all plugins?

dblock: I checked other plugins and they write 1.1.0 via `opensearchplugin` gradle extension, including when the depenency is -SNAPSHOT.

vrozov (approved):

cliu123 (approved):

PR #1601, opened 2022-02-09
Updates the test retry-count to give flaky tests more chances to pass

DarshitChanpura:
### Description
This PR changes the number of retries maven should do if a test fails. It is done so in order to give flaky tests more runs to pass, and hence to improves chances of CI pipeline to not fail because of these flaky tests.
* Refactoring
* Allows flaky tests to have better chance of passing
* The retry-count changes from `3` to `5`

### Check List

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -632,8 +632,8 @@
 
                 <configuration>
                     <argLine>-Xmx3072m</argLine>
-                    <rerunFailingTestsCount>3</rerunFailingTestsCount>
-                    <forkCount>3</forkCount>
+                    <rerunFailingTestsCount>5</rerunFailingTestsCount>
+                    <forkCount>5</forkCount>
                     <reuseForks>true</reuseForks>
                     <!--
                     <parallel>methods</parallel>


pom.xml
```diff
@@ -632,8 +632,8 @@
 
                 <configuration>
                     <argLine>-Xmx3072m</argLine>
-                    <rerunFailingTestsCount>3</rerunFailingTestsCount>
-                    <forkCount>3</forkCount>
+                    <rerunFailingTestsCount>5</rerunFailingTestsCount>
+                    <forkCount>5</forkCount>
```
davidlago, line 636: Is the `forkCount` change needed? would we have enough memory with the 3Gb allocated in L653 if we now have 5 threads running tests? I think we'd be ok just changing the `rerunFailingTestsCount` attribute to `5`.

DarshitChanpura: okay... i wasn't sure about this change... but i will change forkCount back to 3

peternied (approved):

PR #1604, opened 2022-02-11
Adding auto delete workflow for backport branches

VachaShah:
Signed-off-by: Vacha Shah <vachshah@amazon.com>

### Description
Workflow to automatically delete the branches created by the [backport action ](https://github.com/opensearch-project/k-NN/blob/main/.github/workflows/backport.yml)of type `backport/*`.

This PR needs to be backported to release branches so that the auto delete runs on backport PRs created for those branches.

* Category: Enhancement
In order to automatically delete branches created by the backport workflow.
Branches have to be manually deleted after auto backport PR is merged without this change.

### Issues Resolved

### Testing
This works in OpenSearch repo.

### Check List

### DIFF
--- a/.github/workflows/delete_backport_branch.yml
+++ b/.github/workflows/delete_backport_branch.yml
@@ -1,0 +1,15 @@
+name: Delete merged branch of the backport PRs
+on: 
+  pull_request:
+    types:
+      - closed
+  
+jobs:
+  delete-branch:
+    runs-on: ubuntu-latest
+    if: startsWith(github.event.pull_request.head.ref,'backport/')
+    steps:
+      - name: Delete merged branch
+        uses: SvanBoxel/delete-merged-branch@b77e873cee00b09f55cc553bd24aae5f8dfc9157
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


davidlago (dismissed):

.github/workflows/delete_backport_branch.yml
```diff
@@ -0,0 +1,15 @@
...
+    types:
+      - closed
+  
+jobs:
+  delete-branch:
+    runs-on: ubuntu-latest
+    if: startsWith(github.event.pull_request.head.ref,'backport/')
+    steps:
+      - name: Delete merged branch
+        uses: SvanBoxel/delete-merged-branch@main
```
peternied, line 13: We should not use branch references for security reasons, this should be a commit id or tag to be sure only the approved version of the plugin is running

peternied (changes requested):

davidlago (approved):

peternied (approved):
Thanks!

VachaShah:
@davidlago @peternied This PR needs to be backported to the release branches so that this workflow can run when PRs are backported to release branches. Which release branches would those be? I can add appropriate labels on this PR so it can be auto-backported.

peternied:
@davidlago you know better than I at this point, but I promise I'll pay attention

davidlago:
Good catch! We only have `main` which is `1.3.0-SNAPSHOT` but we do have branch `1.2` whose head is currently `1.2.4-SNAPSHOT` so we should indeed backport to that. @peternied could you also please check whether we need a similar change for the dashboards security plugin?

peternied:
https://github.com/opensearch-project/security-dashboards-plugin/pull/901 is tracking this change getting merged.

PR #1623, opened 2022-02-17
Create the plugin-descriptor programmatically

peternied:
### Description
After migrating from maven to gradle, the build script for security in OpenSearch-build was broken. We could not use the default script and had to make modifications, see https://github.com/opensearch-project/opensearch-build/pull/1631

### Testing
Built locally and confirmed the contents of the programmatically generated file matched the intended behavoir that was previously done by overrides, see https://github.com/opensearch-project/opensearch-build/blob/main/scripts/components/security/build.sh#L70-L71

### Check List


build.gradle
```diff
@@ -307,7 +307,24 @@ task bundleSecurityAdminStandaloneTarGz(dependsOn: jar, type: Tar) {
     }
 }
 
+task createPluginDescriptor() {
+    List<String> descriptorProperties = [
+        "description=Provide access control related features for OpenSearch",
+        "version=${version}",
+        "name=opensearch-security",
+        "classname=org.opensearch.security.OpenSearchSecurityPlugin",
+        "java.version=1.8",
```
DarshitChanpura, line 316: should we feed java version as a variable instead of hardcoding?

davidlago (approved):

PR #1628, opened 2022-02-18
Remove standalone admin tools zip file

peternied:
### Description
There is a copy of the admin tools that are produced on build but are
not consumed or distributed as part of this project.  Until we have
documentation and/or tooling that uses this we should remove it from the
project.

By removing this file the security plugin to remove its custom build logic from OpenSearch-build, [see]( https://github.com/opensearch-project/opensearch-build/issues/1631#issuecomment-1042471356).

### Testing
Searched through opensearch.org, this repository, and the security-dashboards-plugin repo for references to these stand alone tools.

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -273,39 +273,6 @@ task bundle(dependsOn: jar, type: Zip) {
     }
 }
 
-task bundleSecurityAdminStandalone(dependsOn: jar, type: Zip) {
-    archiveClassifier = 'securityadmin-standalone'
-    from(configurations.runtimeClasspath) {
-        into 'deps/'
-    }
-    from(project.jar) {
-        into 'deps/'
-    }
-    from('tools') {
-        into 'tools/'
-    }
-    from('securityconfig') {
-        into 'deps/securityconfig'
-    }
-}
-task bundleSecurityAdminStandaloneTarGz(dependsOn: jar, type: Tar) {
-    archiveClassifier = 'securityadmin-standalone'
-    archiveExtension = 'tar.gz'
-    compression = Compression.GZIP
-    from(configurations.runtimeClasspath) {
-        into 'deps/'
-    }
-    from(project.jar) {
-        into 'deps/'
-    }
-    from('tools') {
-        into 'tools/'
-    }
-    from('securityconfig') {
-        into 'deps/securityconfig'
-    }
-}
-
 task createPluginDescriptor() {
     List<String> descriptorProperties = [
         "description=Provide access control related features for OpenSearch",
@@ -324,8 +291,6 @@ bundle.doLast() {
 
 tasks.assemble.dependsOn(bundle)
 tasks.bundle.dependsOn(createPluginDescriptor)
-tasks.assemble.dependsOn(bundleSecurityAdminStandalone)
-tasks.assemble.dependsOn(bundleSecurityAdminStandaloneTarGz)
 
 clean {
     delete 'data/'
--- a/build.gradle
+++ b/build.gradle
@@ -273,6 +273,39 @@ task bundle(dependsOn: jar, type: Zip) {
     }
 }
 
+task bundleSecurityAdminStandalone(dependsOn: jar, type: Zip) {
+    archiveClassifier = 'securityadmin-standalone'
+    from(configurations.runtimeClasspath) {
+        into 'deps/'
+    }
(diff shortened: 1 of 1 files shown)


DarshitChanpura (approved):

peternied:
Note; lets hold off merging this until the 23rd is can we can dig up any additional references.

davidlago:
So this one does not present a maintenance overhead (as it is just an additional gradle action) but it can come in handy if users want to be able to run admin commands from a host where the security plugin is located. I propose we keep it around for now until we have a better alternative (or a more formal RFC/soliciting comments/deprecation notice etc to give users a chance to raise their hands if this removal causes pain for them).

build.gradle
```diff
@@ -324,8 +291,6 @@ bundle.doLast() {
 
 tasks.assemble.dependsOn(bundle)
 tasks.bundle.dependsOn(createPluginDescriptor)
-tasks.assemble.dependsOn(bundleSecurityAdminStandalone)
```
peternied, line 327: > So this one does not present a maintenance overhead (as it is just an additional gradle action)
@davidlago 

Due to these lines, they are built by default.  Would you prefer I leave the tasks, but remove these lines so you have to opt in with `./gradlew bundleSecurityAdminStandaloneTarGz`.  Note; this feature has no documentation in this codebase which is why I figured it would make more sense to remove it entirely.

davidlago: I think that's a good compromise. Let's move to non-default first to put the potential users "on notice". If they try to run it and it is not built by default they have an easy fix, and we can think of removing in 2.0. Do you have any ideas for how to do the deprecation notice @peternied ?

davidlago: Note: this should give us an opportunity to engage with people using this, if anyone, to make sure we understand the use case and perhaps have an argument against deprecation/removal.

peternied: > Do you have any ideas for how to do the deprecation notice?

Typically for a breaking change we have the weekly meeting that is a great place to communicate them.  However, for this change it is murkier since we hardly have supported this tool.  I've created a [post in the OpenSearch forums](https://discuss.opendistrocommunity.dev/t/security-admin-standalone-should-there-be-more-investment-in-this-tool/8748) as a starting point.  @stockholmux Did you have additional thoughts on how we might want to spread the word or gather input?

peternied: I've seen no traction over the past few days, unless we have other thoughts I'd like to disable building by default and we can add it back if there is concern.

peternied:
@opensearch-project/security I'd like to merge this PR as the stand alone tools can be built on their own even after this change goes in

cliu123:
> @opensearch-project/security I'd like to merge this PR as the stand alone tools can be built on their own even after this change goes in

Sorry, I lost track of this PR a bit. @peternied @davidlago What was the plan for this?

peternied:
> > @opensearch-project/security I'd like to merge this PR as the stand alone tools can be built on their own even after this change goes in
> 
> 
> 
> Sorry, I lost track of this PR a bit. @peternied @davidlago What was the plan for this?

My plan as documented in this PR is to disable building the stand alone changes by default, I've started a forum thread and this PR with no comment.  If we find that developers are impacted it's trivial to revert, what do you think?

cliu123:
> > > @opensearch-project/security I'd like to merge this PR as the stand alone tools can be built on their own even after this change goes in
> > 
> > 
> > Sorry, I lost track of this PR a bit. @peternied @davidlago What was the plan for this?
> 
> My plan as documented in this PR is to disable building the stand alone changes by default, I've started a forum thread and this PR with no comment. If we find that developers are impacted it's trivial to revert, what do you think?

I like the idea of keeping the distribution as clean as possible. As @davidlago mentioned, would it mind adding a deprecation notice on this tool in 1.3 and remove this tool in 2.0 as a breaking change if no objections from the community?

peternied:
> Would it mind adding a deprecation notice on this tool in 1.3 and remove this tool in 2.0 as a breaking change if no objections from the community?

Lets work on the notice separately from this PR, we switched from mvn to gradle without considering this a breaking change - unless that change was itself an oversight.

Otherwise lets close this PR, this has been a lot of back and forth with little gain.

davidlago (approved):

cliu123 (approved):

PR #1632, opened 2022-02-19
Include cause in unexpected exceptions

peternied:
### Description
Include cause in unexpected exceptions

Also added `info` level logging to see these messages in a PR
And removed artifacts step as they were never used and the martix of jdk
release would override them (saves couple minutes from check runtime)

### Testing
Added an exception and verified it appeared in the output

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/filter/SecurityFilter.java
+++ b/src/main/java/org/opensearch/security/filter/SecurityFilter.java
@@ -373,7 +373,7 @@ public class SecurityFilter implements ActionFilter {
             listener.onFailure(e);
         } catch (Throwable e) {
             log.error("Unexpected exception "+e, e);
-            listener.onFailure(new OpenSearchSecurityException("Unexpected exception " + action, RestStatus.INTERNAL_SERVER_ERROR));
+            listener.onFailure(new OpenSearchSecurityException("Unexpected exception " + action, RestStatus.INTERNAL_SERVER_ERROR, e));
         }
     }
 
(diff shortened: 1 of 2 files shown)


cliu123:
The cause(error message) would be exposed to client. Is it sufficient to log the cause as a part of stacktrace in logs?

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -373,7 +373,7 @@ public void onFailure(Exception e) {
             listener.onFailure(e);
         } catch (Throwable e) {
             log.error("Unexpected exception "+e, e);
-            listener.onFailure(new OpenSearchSecurityException("Unexpected exception " + action, RestStatus.INTERNAL_SERVER_ERROR));
+            listener.onFailure(new OpenSearchSecurityException("Unexpected exception " + action, RestStatus.INTERNAL_SERVER_ERROR, e));
```
cliu123, line 376: The cause(error message) would be exposed to client. Is it sufficient to log the cause as a part of stacktrace in logs?

peternied: That is a good call out, I'll create an issue test for that so we don't do this by accident.

Do you have a recommendation on how to make sure in the tests that we are getting the log output on line 375, it seems like in practice this isn't a real issue, but for our test cases we aren't seeing the expected results?

davidlago (approved):

peternied:
Being that this PR was created to expose the exception message, and we shouldn't expose these messages in this way I have created https://github.com/opensearch-project/security/issues/1644 to track the root cause (tests aren't collecting/outputing log.error from the cluster under test).  Closing out this PR.

Issue #1653, opened 2022-02-22
[BUG] Security plugin transport layer(multi-node cluster) does not work due to missing required reflective access (JDK17)


cliu123:
**Describe the bug**
Security plugin 1.3.0.0 does not work on OpenSearch 1.3.0 multi-node clusters without required reflective access.
This issue was originated from: https://github.com/opensearch-project/security/issues/1619

**To Reproduce**
1. Create a OpenSearch 1.3.0 cluster with at least 2 nodes.
2. Install security plugin 1.3.0.0(security plugin repo [main branch](https://github.com/opensearch-project/security)) on each of the nodes.
3. Send a basic request(such as `_cat/nodes` or `_cat/indices`) to the OpenSearch cluster.
4. See error:
```
% curl -k -XGET -u admin:admin https://localhost:9200/_cat/indices
{"error":{"root_cause":[{"type":"security_exception","reason":"Unexpected exception indices:monitor/stats"}],"type":"security_exception","reason":"Unexpected exception indices:monitor/stats"},"status":500}% 
```
Stacktrace in logs:
```
org.opensearch.OpenSearchSecurityException: Unexpected exception cluster:monitor/nodes/info
         at org.opensearch.security.filter.SecurityFilter.apply0(SecurityFilter.java:376) [opensearch-security-1.3.0.0.jar:1.3.0.0]
         at org.opensearch.security.filter.SecurityFilter.apply(SecurityFilter.java:154) [opensearch-security-1.3.0.0.jar:1.3.0.0]
         at org.opensearch.action.support.TransportAction$RequestFilterChain.proceed(TransportAction.java:192) [opensearch-1.3.0.jar:1.3.0]
         at org.opensearch.performanceanalyzer.action.PerformanceAnalyzerActionFilter.apply(PerformanceAnalyzerActionFilter.java:99) [opensearch-performance-analyzer-1.3.0.0.jar:1.3.0.0]
         at org.opensearch.action.support.TransportAction$RequestFilterChain.proceed(TransportAction.java:192) [opensearch-1.3.0.jar:1.3.0]
         at org.opensearch.action.support.TransportAction.execute(TransportAction.java:169) [opensearch-1.3.0.jar:1.3.0]
         at org.opensearch.action.support.TransportAction.execute(TransportAction.java:97) [opensearch-1.3.0.jar:1.3.0]
[...]

PR #1668, opened 2022-03-07
Remove java17 from 1.3 build matrix

DarshitChanpura:
### Description
Removes Java-17 support from CI matrix as the expected behavior is that 1.3 will fail for Java-17. Refer this discussion more details: https://github.com/opensearch-project/job-scheduler/pull/130

### Check List

### DIFF
--- a/src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java
+++ b/src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java
@@ -30,7 +30,6 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
 
-import org.junit.Before;
 import org.opensearch.common.settings.Settings;
 
 import org.apache.http.HttpHeaders;
@@ -42,24 +41,12 @@ import org.mockito.internal.util.reflection.FieldSetter;
 import org.opensearch.security.user.AuthCredentials;
 import org.opensearch.security.util.FakeRestRequest;
 import com.google.common.io.BaseEncoding;
-
-import static org.junit.Assume.assumeFalse;
-
 public class HTTPJwtAuthenticatorTest {
 
     final static byte[] secretKey = new byte[1024];
 
     static {
         new SecureRandom().nextBytes(secretKey);
-    }
-
-    /*
-    This test fails during Java 17 build due to a known bug: https://bugs.openjdk.java.net/browse/JDK-8251547
-    TODO: This method should be removed once a fix is implemented
-    */
-    @Before
-    public void isJavaVersionBelow17(){
-        assumeFalse(System.getProperty("java.version").startsWith("17"));
     }
 
     @Test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        jdk: [8, 11, 14, 17]
+        jdk: [8, 11, 14]
 
     steps:
 


cliu123:
Please rebase.

cliu123:
Are these needed to be removed?
https://github.com/opensearch-project/security/blob/77d8861fbd704685765ddbf8688c33cab5a1ae32/src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java#L57-L63

peternied (commented):
I don't think we need this change.  We can support additional versions of the JDK without impacting the distribution.

cliu123:
> I don't think we need this change. We can support additional versions of the JDK without impacting the distribution.

JDK 17 breaks security plugin 1.3: https://github.com/opensearch-project/security/issues/1653

DarshitChanpura:
> Are these needed to be removed?
> 
> https://github.com/opensearch-project/security/blob/77d8861fbd704685765ddbf8688c33cab5a1ae32/src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java#L57-L63

We can remove this once we fix the Java-17 bug... but we can leave it as is right now

peternied (commented):
> JDK 17 breaks security plugin 1.3: https://github.com/opensearch-project/security/issues/1653

Yes there is a bug, you cannot run the distribution on JDK17, and distribution 1.3 will not support JDK17. 

The the security plugin build will need to support JDK17 for us to fix that issue.  No reason to disable it to re-enable it slightly later.  We aren't 'fixing' the problem by disabling the CI build in this PR.

DarshitChanpura:
> > JDK 17 breaks security plugin 1.3: #1653
> 
> Yes there is a bug, you cannot run the distribution on JDK17, and distribution 1.3 will not support JDK17.
> 
> The the security plugin build will need to support JDK17 for us to fix that issue. No reason to disable it to re-enable it slightly later. We aren't 'fixing' the problem by disabling the CI build in this PR.

Agreed...we can close this PR as this change won't is not affecting anything outside our CI pipeline

cliu123:
>The the security plugin build will need to support JDK17 for us to fix that issue. No reason to disable it to re-enable it slightly later. We aren't 'fixing' the problem by disabling the CI build in this PR.

@peternied @DarshitChanpura This PR is disabling JDK 17 only on 1.3 instead of `main`. We are not going to re-enable JDK 17 on 1.3. It has been decided that 1.3 will not support JDK 17:
https://github.com/opensearch-project/job-scheduler/pull/130
https://github.com/opensearch-project/opensearch-plugins/issues/64

JDK 17 will not be disabled on `main`, and the issues blocking security plugin to support JDK 17 will be fixed in 2.0(`main`). It would not make sense to leave JDK 17 that is not being supported on 1.3 branch CI. Please correct me if there are other reasons to keep JDK 17 but not to support it.

cliu123:
> We can remove this once we fix the Java-17 bug... but we can leave it as is right now

@DarshitChanpura IMHO, if 1.3 will not support JDK 17, those changes should be reverted. User should see that 1.3 does not support JDK 17, which is explicitly documented, so it will be expected to fail those tests if users run those tests with JDK 17. The issues blocking JDK 17 support will only be fixed in 2.0 instead of 1.3.

DarshitChanpura:
> > We can remove this once we fix the Java-17 bug... but we can leave it as is right now
> 
> @DarshitChanpura IMHO, if 1.3 will not support JDK 17, those changes should be reverted. User is expected to see that 1.3 does not support JDK 17, which is explicitly documented, so it will be expected to fail those tests if users run those tests with JDK 17. The issues blocking JDK 17 support will only be fixed in 2.0 instead of 1.3.

Okay...I'm gonna revert the test related change too...and we can decide whether we want to move forward with this merge or not.

peternied (approved):

cliu123 (approved):

src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java
```diff
@@ -42,9 +41,6 @@
 import org.opensearch.security.user.AuthCredentials;
 import org.opensearch.security.util.FakeRestRequest;
 import com.google.common.io.BaseEncoding;
-
```
cliu123, line 45: nit: Leave an empty row after imports.

PR #1685, opened 2022-03-16
Introduce BWC tests in security plugin

### LINKED ISSUE #1417: Add backwards compatibility tests for automation
Add backwards compatibility tests for the plugin to automate end-to-end upgrade paths for faster releases. Developer documentation for implementing bwc tests and hooking them to CI: https://github.com/opensearch-project/opensearch-plugins/blob/main/TESTING.md#backwards-compatibility-testing

cliu123:

### Description
* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation) Enhancement
* Why these changes are required? Automate backward compatibility tests

### Issues Resolved
#1417

### Testing
BWC tests run in CI.

### Check List


peternied (changes requested):
Can we generate the cert files at runtime instead of including them as binaries?

cliu123:
> Can we generate the cert files at runtime instead of including them as binaries?

@peternied Great question! There's no way to generate these certs in runtime. They must be provided as test resources. Same case in other plugins, for example [index-management-plugin](https://github.com/opensearch-project/index-management/tree/main/src/test/resources/security). And these are just demo certs, so they are safe to provided as test resources. There're other similar certs provided as test resource in security plugin, for example, [these](https://github.com/opensearch-project/security/tree/main/src/test/resources/ssl) and [these ](https://github.com/opensearch-project/security/tree/main/src/test/resources/sslConfigurator/pem). BTW, there's no need to block this PR with this comment. PRs won't be merged untill all comments are resolved.

peternied:
I don't mean the OpenSearch service runtime, but during a setup phase before the bwc test are actually run? The heart of my goal is avoiding following the anti-pattern of checking in certificates

cliu123:
> I don't mean the OpenSearch service runtime, but during a setup phase before the bwc test are actually run? The heart of my goal is avoiding following the anti-pattern of checking in certificates

Yes, that's not ideal, but this is the best we can do since there's no way to generate those demo certs in runtime or setup phase. But this doesn't introduce any security vulnerability because those are just demo certs like the certs that have already exist in security plugin and other plugins such as Index Management Plugin as I mentioned above.

peternied:
> no way to generate those demo certs in runtime or setup phase

Why not, could you explain why this is not possible?

cliu123:
> > no way to generate those demo certs in runtime or setup phase
> 
> Why not, could you explain why this is not possible?

The content of the certs are all hardcoded. I don't see a way to generate those certs in runtime or prepare steps. If there's a way, please post the guide.

peternied:
We should be able to use command line tools such as openssl, [how to](https://emcorrales.com/blog/how-to-use-a-self-signed-certificate-with-gradle),  I am happy to make a PR onto this change to keep the security posture of this new code high.

cliu123:
> We should be able to use command line tools such as openssl, [how to](https://emcorrales.com/blog/how-to-use-a-self-signed-certificate-with-gradle), I am happy to make a PR onto this change to keep the security posture of this new code high.

Cool! That is something unrelated to BWC tests. Let's include those changes in a separate PR. We can either merge this PR first if we haven't had a clear timeline on when the separate PR will be ready, or I can rebase this branch after you merge the separate PR into `main` branch. But please eliminate all the hardcoded certs in the security plugin repo in the separate PR.

davidlago:
Agree with @cliu123 as this removal of hardcoded certs is probably best as a follower PR/issue. For example, if we decide not to merge this one _and_ we don't work on the removal of hardcoded certs, 2.0 will still have them. I suggest we go ahead and merge this one and then track the other work perhaps linking it to the overarching work of improving the getting started experience that @setiah is leading.

davidlago (approved):

bwc-test/build.gradle
```diff
@@ -0,0 +1,274 @@
...
+
+ext {
+    projectSubstitutions = [:]
+    licenseFile = rootProject.file('LICENSE.TXT')
+    noticeFile = rootProject.file('NOTICE')
+}
+
+buildscript {
+    ext {
+        opensearch_version = System.getProperty("opensearch.version", "1.3.0-SNAPSHOT")
```
DarshitChanpura, line 50: will  this change with version change?

DarshitChanpura (approved):

bwc-test/build.gradle
cliu123: Agree. This will need to be updated when upgrading [OpenSearch dependency version](https://github.com/opensearch-project/security/blob/main/build.gradle#L59). It'd be nice to explore a more dynamic way to avoid this hardcoded version.

peternied:
@opensearch-project/security Thanks for the feedback, we can work on improving the security posture separately with this change in place.

PR #1708, opened 2022-03-24
Convert Plugin install to only build once

peternied:
### Description
When running in the OpenSearch distribution the bundled jdk version is
used, this has a more dramatic impact on the capabilities and
limitations of plugins.  Removing the extra versions as it was providing
a false sense of test coverage.  These test confirm basic installation
and service startup - no depth of coverage that would be exposed by
runtime versions.

### Check List


DarshitChanpura (commented):
Out of curiosity, what would this change achieve in a larger picture?

peternied:
> Out of curiosity, what would this change achieve in a larger picture?

This lowers the number of resources used by GitHub Actions when we add a pull request, it also changes the look of the check to look like: `Plugin Install / Plugin Install (pull_request) Successful in 2m`, instead of having several calls with JDK versions in them

davidlago (approved):

DarshitChanpura (approved):

PR #1660, opened 2022-03-02
tests: Replace TransportClient by NodeClient

rs-eliatra:
### Description
[Describe what this change achieves]
* Category = Maintenance
* Why these changes are required?
Check description in: https://github.com/opensearch-project/security/issues/1578

### Issues Resolved
https://github.com/opensearch-project/security/issues/1578

### Testing
Testing done: 
- unit testing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1715, opened 2022-03-29
Speeding up tests

peternied:
### Description
Sleep statements should rarely be used, instead apis should allow for
blocking/non-blocking calling patterns.  Attempting to remove and clean
up these statements to speed up PR workflows.

Github Action workflows for this repo were 75+ minutes, with this change 40 minutes. **47% reduction in time.**

### Check List

### DIFF
--- a/src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
+++ b/src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
@@ -192,18 +192,7 @@ public abstract class AbstractSecurityUnitTest {
             Assert.assertEquals(clusterInfo.numNodes, cur.getNodes().size());
 
             SearchResponse sr = tc.search(new SearchRequest(".opendistro_security")).actionGet();
-            //Assert.assertEquals(5L, sr.getHits().getTotalHits());
-
             sr = tc.search(new SearchRequest(".opendistro_security")).actionGet();
-            //Assert.assertEquals(5L, sr.getHits().getTotalHits());
-
-            String type=securityConfig.getType();
-            try {
-                Thread.sleep(10000);
-            } catch (InterruptedException e) {
-                e.printStackTrace();
-            }
-
         }
     }
 


DarshitChanpura (approved):

cliu123:
The sleep was introduced by [the removal of TransportClient PR](https://github.com/opensearch-project/security/pull/1660). The intent is to make sure all configurations are loaded into security index, which used to be done as following. Sleep might not be a good option and doesn't really ensure that. Is there a better/faster way to ensure the existense of the configurations? 
```
            Assert.assertTrue(tc.get(new GetRequest(".opendistro_security", type, "config")).actionGet().isExists());
            Assert.assertTrue(tc.get(new GetRequest(".opendistro_security", type,"internalusers")).actionGet().isExists());
            Assert.assertTrue(tc.get(new GetRequest(".opendistro_security", type,"roles")).actionGet().isExists());
            Assert.assertTrue(tc.get(new GetRequest(".opendistro_security", type,"rolesmapping")).actionGet().isExists());
            Assert.assertTrue(tc.get(new GetRequest(".opendistro_security", type,"actiongroups")).actionGet().isExists());
            Assert.assertFalse(tc.get(new GetRequest(".opendistro_security", type,"rolesmapping_xcvdnghtu165759i99465")).actionGet().isExists());
            Assert.assertTrue(tc.get(new GetRequest(".opendistro_security", type,"config")).actionGet().isExists());
            if (indexRequests.stream().anyMatch(i -> CType.NODESDN.toLCString().equals(i.id()))) {
                Assert.assertTrue(tc.get(new GetRequest(".opendistro_security", type,"nodesdn")).actionGet().isExists());
```

peternied:
First when I noticed and removed this sleep I had the same concern about stability.  In the test setup OpenSearch nodes are created on a couple of threads, these communicate via http.  From what I've seen when our tests cases are running, we are making calls via http which is out of process communication.

When moving across the process boundary layer to the network the jvm stops processing work on the test thread and can resume work on any other busy node threads.  This boundary layer means a busy cpu operation is very likely to complete on the nodes within that same process before the network connection of the test's inbound request is processed.

This is not reliable protection on a production system; however, this seems to be working well with our JVM selections and our different environments.  We might notice issues in the future - I imagine certain classes of tests might need additional checks, but we can deal with them as they present themselves.

>  Is there a better/faster way to ensure the existence of the configurations?

In OpenSearch there is a [task api](https://opensearch.org/docs/latest/opensearch/rest-api/tasks/) that can be used to track work that is in progress or complete.  If we needed additional assurance that the configuration was complete, I would advocate for a way to get the task id associated with this so we can use `GET _tasks/<task_id>?wait_for_completion=true` to ensure that work has finished making sure the system is stable.

Alternatively we could add into those nodes that would give us an in-process signaling channel such as `Object.wait(...)` with a plugin that is designed to make this information available.

PR #1716, opened 2022-03-30
Fix data-stream name resolution for wild-cards

sandeshkr419:
Signed-off-by: Sandesh Kumar <sandeshkr419@gmail.com>

### Description
[Describe what this change achieves]
* Category: Bug fix
* Security plugin is not able to invalidate non-permitted data-stream (get/delete/stats) requests sent using wild-card expressions. 
This is because the list of 'allIndices' in 'ResolvedIndicesProvider' class does not has the resolved names of data streams. If the 'allIndices' variable when resolved, is empty -> leads the authorization to succeed as there are no eligible index patterns to block the request. 
In this change, we add the the names of resolved data streams to 'allIndices' so data stream names also can get resolved. 

* What is the old behavior before changes and new behavior after changes?

Suppose the indices are as follows in a OS cluster.

```
kusandes@a483e789b681 ~ % curl -XGET 'https://localhost:9200/_cat/indices?v' --insecure -u 'admin:admin'
health status index                        uuid                   pri rep docs.count docs.deleted store.size pri.store.size
kusandes@a483e789b681 ~ % curl -XGET 'https://localhost:9200/_cat/indices?v' --insecure -u 'admin:admin'
health status index                        uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   .ds-logs-nginx1-000001       3V1VOJ4NQ0KEqtrJ51el8g   1   1          0            0       208b           208b
yellow open   logs2                        AwHqtjeITBGumadYv5GNuQ   1   1          0            0       208b           208b
yellow open   .ds-logs-nginx11-000001      lnMe0uUPToKbQKJCrMUoSw   1   1          0            0       208b           208b
yellow open   logs1                        oR5D0gWFT_u5lbpvSi-SOA   1   1          0            0       208b           208b
yellow open   security-auditlog-2022.03.28 YknIcv-GQLGtdocWUA6-UA   1   1         31            0     94.8kb         94.8kb
yellow open   .ds-logs-nginx3-000001       E2fpfxHSSMarY84WMIjhhQ   1   1          0            0       208b           208b
green  open   .opendistro_security         h6L4aW-QR5KMrCui6BL12g   1   0          9            0     87.9kb         87.9kb
```

Assume the user ```sandesh1``` to have the following data-streams related permissions:
```
kusandes@a483e789b681 ~ % curl -XPUT https://localhost:9200/_plugins/_security/api/roles/ds_all -u 'admin:admin' --insecure -H 'Content-Type: application/json' -d '
[...]

PR #1719, opened 2022-03-30
Revert "Fix data-stream name resolution for wild-cards"

peternied:
Reverts opensearch-project/security#1716


peternied:
This change should not be released with the 1.3.1 release today

PR #1735, opened 2022-04-01
Introduce dfm_empty_overrides_all setting to enable role without dls/fls to override roles with dls/fls

### LINKED ISSUE #1572: Multiple roles and DLS/FLS: role with no DLS/FLS should "win"
**Is your feature request related to a problem? Please describe.**
situation:
* there are two roles: role A with DLS, role B without DLS, both giving access to an index X
* role A is granted to all users (intent: grant it to all "real" users so that they see only the data they need)
* role B is granted to a technical user to give him additional rights (e.g. data import user which needs to be able to see all documents)
* because role A is granted to all users it is also given to the technical user

due to the user having role B he should be able to see _all_ documents in index X, even though he also has role A.

**Describe the solution you'd like**
there should be an option to change the behaviour so that the role with no DLS wins.

Search Guard has this feature using a config-option (system-wide setting: `dfm_empty_overrides_all`):
* [release notes](https://docs.search-guard.com/latest/changelog-searchguard-7x-42_0_0#document--and-field-level-security)
* [documentation for DLS](https://docs.search-guard.com/latest/document-level-security#multiple-roles-and-document-level-security)

**Describe alternatives you've considered**
* granting the role(s) with DLS only to users with specific roles: e.g. in our setup there's only one user in the whole system which shouldn't have this role. there's no way (e.g. backend role in the JWT) which could be used to identify the other users
* granting the role(s) with DLS to all users except users of a specific group: it's not possible to define a role mapping as "everything except"
also, both options would be cumbersome to manage the more roles exist and there's a risk that then a user might not be granted a restricting role and can see everything by accident.

**Additional context**
n/a

cliu123:
…fls to override roles with dls/fls

### Description
* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation) New feature

### Issues Resolved
#1572

### Testing
UTs

### Check List


rursprung:
thanks a lot for this!

@cliu123 maybe you could use `--author` on `git commit` to ensure that the author (@jochenkressin) is properly reflected?

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
```diff
@@ -366,19 +348,29 @@ public EvaluatedDlsFlsConfig getDlsFls(User user, IndexNameExpressionResolver re
             Map<String, Set<String>> dlsQueriesByIndex = new HashMap<String, Set<String>>();            
             Map<String, Set<String>> flsFields = new HashMap<String, Set<String>>();
             Map<String, Set<String>> maskedFieldsMap = new HashMap<String, Set<String>>();
-                        
+
+            // we capture all concrete indices that do not have any
+            // DLS/FLS/Masked Fields restrictions. If the dmf_empty_overwrites_all
```
rursprung, line 353: nit-picking:
```suggestion
            // DLS/FLS/Masked Fields restrictions. If the dfm_empty_overwrites_all
```

i have to say: this line finally explained to me what "dfm" seems to stand for: DLS/FLS/Masked 😄. maybe it'd make sense to document the meaning of this abbreviation somewhere?

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
```diff
@@ -366,19 +348,29 @@ public EvaluatedDlsFlsConfig getDlsFls(User user, IndexNameExpressionResolver re
...
 
 					if (dls != null && dls.length() > 0) {
 
 						for (String concreteIndex : concreteIndices) {
 							dlsQueriesByIndex.computeIfAbsent(concreteIndex, (key) -> new HashSet<String>()).add(dls);
 						}
-					}
+					} else if (dfmEmptyOverwritesAll) {
+                        noDlsConcreteIndices.addAll(concreteIndices);
+                    }
```
rursprung, line 373: indentation seems off here

src/test/java/org/opensearch/security/dlic/dlsfls/DfmOverwritesAllTest.java
```diff
@@ -0,0 +1,227 @@
...
+        Assert.assertFalse(response.getBody().contains("value-3-4"));
+        Assert.assertFalse(response.getBody().contains("value-4-4"));
+
+        // field2 - masked
+        Assert.assertFalse(response.getBody().contains("value-2-1"));
+        Assert.assertFalse(response.getBody().contains("value-2-2"));
+        Assert.assertFalse(response.getBody().contains("value-2-3"));
+        Assert.assertFalse(response.getBody().contains("value-2-4"));
+
+        // field2 - check also some masked vallues
```
rursprung, line 130: nit-picking:
```suggestion
        // field2 - check also some masked values
```

src/test/java/org/opensearch/security/dlic/dlsfls/DfmOverwritesAllTest.java
```diff
@@ -0,0 +1,227 @@
...
+                .source("{\"field1\": 3, \"field2\": \"value-2-3\", \"field3\": \"value-3-3\", \"field4\": \"value-4-3\" }", XContentType.JSON)).actionGet();
+
+        tc.index(new IndexRequest("index1-4").id("0").setRefreshPolicy(RefreshPolicy.IMMEDIATE)
+                .source("{\"field1\": 4, \"field2\": \"value-2-4\", \"field3\": \"value-3-4\", \"field4\": \"value-4-4\" }", XContentType.JSON)).actionGet();
+
+    }
+
+    @Test
+    public void testDFMUnrestrictedUser() throws Exception {
+        // admin user sees all, no dfm restrictions apply
```
rursprung, line 59: why not move such descriptive comments up into a javadoc on the test method? then it's better visible that this is the description of the test method.
(same for the other tests below)

src/test/java/org/opensearch/security/dlic/dlsfls/DfmOverwritesAllTest.java
```diff
@@ -0,0 +1,227 @@
...
+        final Settings settings = Settings.builder().put(ConfigConstants.SECURITY_DFM_EMPTY_OVERRIDES_ALL, true).build();
```
rursprung, line 60: maybe move these few setup lines which seem to be the same for all tests into an `@BeforeEach`?

rursprung (commented):
as usual: this is not meant as a thorough review, for that i'm not familiar enough with this code base (and security things should never be reviewed by people who aren't familiar with them, i.e. me 😄)

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -85,23 +74,24 @@
...
 import org.opensearch.security.support.ConfigConstants;
 import org.opensearch.security.support.WildcardMatcher;
 import org.opensearch.security.user.User;
+import org.opensearch.tasks.Task;
+import org.opensearch.threadpool.ThreadPool;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import com.google.common.collect.Sets;
+import java.util.*;
```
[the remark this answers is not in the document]
peternied: Please revert the import changes.  We enumerate imports to help identified when new dependencies are added, with them being re-ordered like this its nigh impossible to tell without using other tooling.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
peternied: Nit; we should add formatting linter that can auto-fix this

src/test/java/org/opensearch/security/dlic/dlsfls/DfmOverwritesAllTest.java
```diff
@@ -0,0 +1,227 @@
...
+ * Tests that the dfm_empty_overwrites_all flag works correctly.
+ * Per default, if a user has a role that adds restrictions to an index
+ * regarding DLS, FLS or masked fields (DFM), and another role that has no restrictions
+ * on the same index, the restrictions from the first role still applies.
+ *
+ * If the dfm_empty_overwrites_all flag is set to true, the logic is reversed:
+ * If a user has a role that places no restrictions on an index, this trumps
+ * all other role that eventually do place restrictions on this index.
+ */
+public class DfmOverwritesAllTest extends AbstractDlsFlsTest {
```
peternied, line 39: Can we add unit tests in addition to these integration tests - I am unable to interpret if the assertions in these tests are correct behavior or not.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
rursprung: please note that this isn't resolved yet even though it has been marked as such.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
peternied: I've created [FEATURE] Add java checkstyle / formatter https://github.com/opensearch-project/security/issues/1752 to track this.

src/test/java/org/opensearch/security/dlic/dlsfls/DfmOverwritesAllTest.java
peternied: I do not see any comments or updates in this issue

peternied (changes requested):
Thank you for removing the import changes and comments around the tests, much more readable!

Very close now, a few callouts remain.

src/test/java/org/opensearch/security/dlic/dlsfls/DfmOverwritesAllTest.java
cliu123: There're unit tests. The comments of each test method explain the test cases.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
```diff
@@ -406,8 +418,26 @@ public EvaluatedDlsFlsConfig getDlsFls(User user, IndexNameExpressionResolver re
...
                         }
-                    } 
+                    } else if (dfmEmptyOverwritesAll) {
+                        noMaskedFieldConcreteIndices.addAll(concreteIndices);
+                    }
+                }
+            }
+
+            if (dfmEmptyOverwritesAll) {
+                if (log.isDebugEnabled()) {
```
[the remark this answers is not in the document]
cliu123: There're quite a few same if statements like this in the code base. It'll be good to clean them up.

src/test/java/org/opensearch/security/dlic/dlsfls/DfmOverwritesAllTest.java
cliu123: Done. Java doc is actually barely used in this repo, especially in the tests. But this is a great idea!

src/test/java/org/opensearch/security/dlic/dlsfls/DfmOverwritesAllTest.java
cliu123: There are a couple of reasons not to include this in `@BeforeEach`:
1. There might be some new tests in the future where the setting is set to other values, for example `false`.
2. The variable defined in `@BeforeEach` is not accessible from test methods.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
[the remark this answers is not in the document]
rursprung: these guards are usually added for performance reasons: even if logging is disabled, any other function calls needed to pass values into the debug call will still be executed. if these are expensive (e.g. calculating some debug string) then that's just wasted time if debug is disabled.

whether this was warranted in this case here i can't say as it just called `Map#keySet` a couple of times which is _probably_ relatively fast.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
```diff
@@ -366,19 +366,29 @@ public EvaluatedDlsFlsConfig getDlsFls(User user, IndexNameExpressionResolver re
...
 					String dls = ip.getDlsQuery(user);
 
 					if (dls != null && dls.length() > 0) {
 
 						for (String concreteIndex : concreteIndices) {
 							dlsQueriesByIndex.computeIfAbsent(concreteIndex, (key) -> new HashSet<String>()).add(dls);
 						}
-					}
+					} else if (dfmEmptyOverwritesAll) {
+                        noDlsConcreteIndices.addAll(concreteIndices);
```
rursprung, line 390: the formatting is still off here and should be fixed (even if no linter is in place for this yet)

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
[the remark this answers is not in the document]
cliu123: @rursprung That is a good point! This explains the value of `isDebugEnabled())` check. I'll add the check back in the PR. Going forward, we should have a consistent approach in all the places of `log.debug()`.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
cliu123: The indentation looks good in my IDE. This looks like space vs tab issue in Github. Similar issue like [this](https://github.com/github/VisualStudio/issues/1524#issuecomment-372632866).
Screen Shot 2022-04-09 at 10 02 42 PM

cliu123: The indentation looks good in my IDE. This looks like space vs tab issue in Github. Similar issue like https://github.com/github/VisualStudio/issues/1524#issuecomment-372632866.
Screen Shot 2022-04-09 at 10 55 46 PM

cliu123: Fixed.

src/test/java/org/opensearch/security/dlic/dlsfls/DfmOverwritesAllTest.java
peternied: These tests integration tests, because they test multiple systems working together - spinning up a OpenSearch cluster.  Unit tests would as example invoke the method `getDlsFls(...)` valid its outcome directly.  The value of unit tests is they are much faster to execute and can validate error cases easier.

I will not block this PR on a practice that is not widely adopted in this codebase - as I look around I mostly see integration tests.  Keep the validation strategy in mind as we continue adding improvements so we can ensure a maintainable codebase.

peternied (approved):

peternied:
@opensearch-project/security Could we get a second reviewer to take a look?

davidlago (approved):

Issue #1676, opened 2022-03-11
Add support for build version qualifiers.


dblock:
Coming from https://github.com/opensearch-project/opensearch-build/issues/1632, add support for `-Dbuild.version_qualifier=`. 

The distribution build process now supports passing additional optional qualifiers other than the 3 digit versions, e.g. `2.0.0-alpha1`. Add the ability to pass additional qualifier to build scripts, and generate artifacts with a version that includes the qualifier.

cliu123:
Looks like OpenSearch core doesn't accept plugin version with version_qualifier. Plugin installation failed: https://github.com/opensearch-project/security/runs/5797837160?check_suite_focus=true
```
2022-04-02T05:49:53.3391925Z -> Installing file:///docker-host/security-plugin.zip
2022-04-02T05:49:53.3426933Z ##[error]Exception in thread "main" java.lang.IllegalArgumentException: the version needs to contain major, minor, and revision, and optionally the build: 2.0.0.0-alpha1
2022-04-02T05:49:53.3428210Z 	at org.opensearch.Version.fromString(Version.java:186)
2022-04-02T05:49:53.3428625Z 	at org.opensearch.plugins.PluginInfo.readFromProperties(PluginInfo.java:222)
2022-04-02T05:49:53.3429203Z 	at org.opensearch.plugins.InstallPluginCommand.loadPluginInfo(InstallPluginCommand.java:811)
2022-04-02T05:49:53.3429760Z 	at org.opensearch.plugins.InstallPluginCommand.installPlugin(InstallPluginCommand.java:870)
2022-04-02T05:49:53.3430273Z 	at org.opensearch.plugins.InstallPluginCommand.execute(InstallPluginCommand.java:275)
2022-04-02T05:49:53.3430780Z 	at org.opensearch.plugins.InstallPluginCommand.execute(InstallPluginCommand.java:249)
2022-04-02T05:49:53.3431282Z 	at org.opensearch.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:100)
2022-04-02T05:49:53.3431744Z 	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138)
2022-04-02T05:49:53.3432163Z 	at org.opensearch.cli.MultiCommand.execute(MultiCommand.java:104)
2022-04-02T05:49:53.3432587Z 	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138)
[...]

PR #1742, opened 2022-04-05
Updates for build qualifiers

peternied:
### Description
Making sure to fully support build qualifiers as used in OpenSearch 2.0.0 builds

### Issues Resolved
* Resolves https://github.com/opensearch-project/security/issues/1676

### Check List


cliu123:
It'll be good to avoid hardcode versions. Please see the changes in [this branch](https://github.com/opensearch-project/security/compare/main...cliu123:upgrade_to_opensearch_2.0.0_alpha1?expand=1)(WIP).

build.gradle
```diff
@@ -326,22 +322,19 @@ task bundleSecurityAdminStandaloneTarGz(dependsOn: jar, type: Tar) {
 }
 
 task createPluginDescriptor() {
-    if (opensearch_version.contains("-SNAPSHOT")) {
-        opensearch_version=opensearch_version.substring(0, opensearch_version.length() - 9)
-    }
     List<String> descriptorProperties = [
         "description=Provide access control related features for OpenSearch",
-        "version=${version}",
+        "version=${version.tokenize('-')[0]}",
```
DarshitChanpura, line 327: what would this change achieve?

peternied: Previously our version string would only be the numeric version `2.0.0` now it can include `2.0.0-alpha` and this is not accepted in the descriptor file.  `String.tokenize(delim)[0]` is a quick way in groovy to substring before the given token.

cliu123:
> > It'll be good to avoid hardcode versions.
> 
> How are these properties hardcoded? They can be altered by passing in command line parameters the test matrix is from the OpenSearch 2.0.0 build logs, e.g. for K-NN the command is executed as `./gradlew assemble --no-daemon --refresh-dependencies -DskipTests=true -Dopensearch.version=2.0.0-alpha1-SNAPSHOT -Dbuild.snapshot=true -Dbuild.version_qualifier=alpha1`

I'm referring to versions being hardcoded in `gradle.properties` file.

peternied:
> I'm referring to versions being hardcoded in `gradle.properties` file.

@cliu123 These can be substituted with command line overrides, and there needs to be defaults when the build occurs without any parameters. If this value is stored in a gradle.build file or gradle.properties or some other convention - this sounds like a preference, how can we move forward with this PR?

cliu123:
@peternied Can we avoid mixing up changes? You don't have to merge the changes in the PR#1741 into this PR when this PR#1741 is in WIP.

peternied:
@cliu123 I needed to pull in the test fixes otherwise the build with the qualifier wouldn't function.  The integration tests of the distribution and the security plugin are blocked on these changes, so optimizing for speed is important.

cliu123:
> @cliu123 I needed to pull in the test fixes otherwise the build with the qualifier wouldn't function. The integration tests of the distribution and the security plugin are blocked on these changes, so optimizing for speed is important.

Speed is not the first priority. This PR is for version qualifier. There's no need to include other changes, especially the changes from another open PR. BTW, there's [a blocking issue](https://github.com/opensearch-project/OpenSearch/issues/2769) on OpenSearch 2.0.0-alpha1.

cliu123 (changes requested):
Speed is not the first priority. This PR is for version qualifier. There's no need to include other changes, especially the changes from another open PR. BTW, there's https://github.com/opensearch-project/OpenSearch/issues/2769 on OpenSearch 2.0.0-alpha1. Please hold off moving forward to alpha1 or alpha2 for now.

PR #1747, opened 2022-04-06
Add archive of reports for test/coverage

peternied:
### Description
This provides a more digestible version of the test results for
debugging/troubleshooting failures that occur in CI.

I moved the jacoco report to use the same location

See how this output looks from this GHA workflow summary https://github.com/opensearch-project/security/actions/runs/2204140083
image

### Check List

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,18 @@ jobs:
       uses: codecov/codecov-action@v1
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
-        files: ./build/jacoco/test/jacocoTestReport.xml
+        files: ./build/reports/jacoco/test/jacocoTestReport.xml
+
+    - uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: ${{ matrix.jdk }}-reports
+        path: |
+          ./build/reports/
+
+    - name: check archive for debugging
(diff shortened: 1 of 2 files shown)


cliu123:
Would you please elaborate what's the old and new behavior? 
Currently the CI raw log has a full list of failing tests. Why is this change needed?

DarshitChanpura (approved):

.github/workflows/ci.yml
```diff
@@ -39,13 +39,24 @@ jobs:
       run: ./gradlew clean build -Dbuild.snapshot=false -x test
 
     - name: Test
-      run: OPENDISTRO_SECURITY_TEST_OPENSSL_OPT=true ./gradlew test -i
```
cliu123, line 42: No need to diable `info` logs in CI logs.

peternied: All the output is still available in the logs if you download the bundle.  My browser chugs when I load the CI build output so this was also to make that part of the GHA workflow better to use.

cliu123: View raw logs will avoid downloading the bundle. And it can be rendered in brower without issues. Let's keep the info logs available to check.
Screen Shot 2022-04-21 at 4 04 39 PM

DarshitChanpura (approved):

Issue #1689, opened 2022-03-18
[BUG] Unable to configure debug/trace logging for security plugins


patcable:
**Describe the bug**
In the [Troubleshooting OpenID Connect](https://opensearch.org/docs/latest/troubleshoot/openid-connect) document it mentions that you can turn on trace logging for OpenID logging. I tried to use this with LDAP - replacing `com.amazon.dllc.auth.http.jwt` with `com.amazon.dlic.auth.ldap.backend`. Regardless of the value for `logger.plugins.security.name` when I try to start OS i get the following on stdout:
```
Mar 18 20:56:13 i-02720ef57ad2d3ad4 opensearch[93705]: Exception in thread "main" org.apache.logging.log4j.core.config.ConfigurationException: No name attribute provided for Logger plugins
Mar 18 20:56:13 i-02720ef57ad2d3ad4 opensearch[93705]:         at org.apache.logging.log4j.core.config.properties.PropertiesConfigurationBuilder.createLogger(PropertiesConfigurationBuilder.java:256)
Mar 18 20:56:13 i-02720ef57ad2d3ad4 opensearch[93705]:         at org.apache.logging.log4j.core.config.properties.PropertiesConfigurationBuilder.build(PropertiesConfigurationBuilder.java:178)
Mar 18 20:56:13 i-02720ef57ad2d3ad4 opensearch[93705]:         at org.opensearch.common.logging.LogConfigurator$2.getConfiguration(LogConfigurator.java:228)
Mar 18 20:56:13 i-02720ef57ad2d3ad4 opensearch[93705]:         at org.opensearch.common.logging.LogConfigurator$2.getConfiguration(LogConfigurator.java:197)
Mar 18 20:56:13 i-02720ef57ad2d3ad4 opensearch[93705]:         at org.apache.logging.log4j.core.config.ConfigurationFactory.getConfiguration(ConfigurationFactory.java:302)
Mar 18 20:56:13 i-02720ef57ad2d3ad4 opensearch[93705]:         at org.opensearch.common.logging.LogConfigurator$3.visitFile(LogConfigurator.java:236)
Mar 18 20:56:13 i-02720ef57ad2d3ad4 opensearch[93705]:         at org.opensearch.common.logging.LogConfigurator$3.visitFile(LogConfigurator.java:232)
Mar 18 20:56:13 i-02720ef57ad2d3ad4 opensearch[93705]:         at java.base/java.nio.file.Files.walkFileTree(Files.java:2725)
[...]

PR #1751, opened 2022-04-07
Switch to log4j logger

peternied:
### Description
In OpenSearch 1.3.0 we removed an explicit dependency on log4j
https://github.com/opensearch-project/security/pull/1563 this caused the
log4j-slf4j-impl-X.XX.X.jar file no longer to be included in the plugin.
When the plugin started up the default no-op logger was used instead.
This prevented the security plugin from logging anything, yikes.

When looking at the other opensearch plugins, none of them use slf4.
Rather than continue using a seperate logging process, moving to the
standard log4j Logger/LogManager.

Tested this change on 2.0.0-alpha distribution and logging works as
expected.

### Issues Resolved
* Resolves https://github.com/opensearch-project/security/issues/1689

### Testing
Downloaded the min versions of opensearch 2.0.0-alpha1, then build the plugin, verified that the log output included lines like `[2022-04-07T18:07:42,441][INFO ][o.o.s.OpenSearchSecurityPlugin] [dev-dsk-petern-1e-d7e6bfc0.us-east-1.amazon.com] 0 OpenSearch Security modules loaded so far: []` that were previously not present

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -92,7 +92,6 @@ dependencies {
     implementation 'org.greenrobot:eventbus:3.2.0'
     implementation 'commons-cli:commons-cli:1.3.1'
     implementation 'org.bouncycastle:bcprov-jdk15on:1.67'
-    implementation 'org.slf4j:slf4j-api:1.7.32'
     implementation 'org.ldaptive:ldaptive:1.2.3'
     implementation 'org.apache.httpcomponents:httpclient-cache:4.5.13'
     implementation 'io.jsonwebtoken:jjwt-api:0.10.8'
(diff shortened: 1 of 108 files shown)


peternied:
@cliu123 This seems like a bad bug to keep in the 1.3.X release line, what do you think about backporting this change?

davidlago:
#1563 removed the explicit dependency and lets core provide it instead. SLF4J provides that layer of abstraction, and the thrust behind that PR was to get us off of declaring it as an explicit dependency so that if new bugs/issues are found, only core needs to update the jar they provide and we should be all set.

I agree that this should be backported to the 1.x line. @peternied is this something we can fix without adding back the dependency? (i.e. figure out why the jar is not being picked up from core and address that instead?)

peternied:
> @peternied is this something we can fix without adding back the dependency? (i.e. figure out why the jar is not being picked up from core and address that instead?)

Yes the dependency can stay out as the library is available from OpenSearch's classpath.

davidlago (approved):

cliu123 (approved):

peternied:
Created https://github.com/opensearch-project/security/issues/1753 to backport this change

PR #1758, opened 2022-04-11
Add signal/wait model for TestAuditlogImpl

peternied:
### Description
I have been tracking test failures with testRestMethod very often
showing failures.  My theory is that the execution environment can
impact the order of operations sometimes causing the audit log not to
contain messages before it is checked.

Adding a new method `doThenWaitForMessages(...)` this ensures the
log queue is fresh, the triggering action completes, and the expected
number of messages were recieved.  There is a second long time window
that allows for the messages to be flushed, this is likely more than
enough - if the messages are recieved the count down latch immediately
continues execution so the tests will not wait if they are ready to
proceed.

While this new method is much more reliable not all tests were
encountering such issues, so I've keep the original convention.  This
can be migrated in one-offs or all at once if we see more troublesome
behavoir.  The previous methods/fields are depreciated to push future
tests to follow the new pattern.

Modifications to the rest helper not to throw exceptions were needed to
keep the Runnable declaration clean and small.

### Check List

### DIFF
--- a/src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
+++ b/src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
@@ -71,7 +71,8 @@ public class TestAuditlogImpl extends AuditLogSink {
         
         try {
             action.run();
-            final boolean foundAll = latch.await(1, TimeUnit.SECONDS);
+            final int maxSecondsToWaitForMessages = 1; 
+            final boolean foundAll = latch.await(maxSecondsToWaitForMessages, TimeUnit.SECONDS);
             if (!foundAll) {
                 throw new RuntimeException("Did not recieve all " + expectedCount +" audit messages after a short wait.");
             }
--- a/src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
+++ b/src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
@@ -41,7 +41,7 @@ public class TestAuditlogImpl extends AuditLogSink {
         super(name, settings, null, fallbackSink);
     }
 
-    public boolean doStore(AuditMessage msg) {
+    public synchronized boolean doStore(AuditMessage msg) {
         if (messagesRef.get() == null || countDownRef.get() == null) {
             throw new RuntimeException("No message latch is waiting");
         }
--- a/src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
+++ b/src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
@@ -73,7 +73,7 @@ public class TestAuditlogImpl extends AuditLogSink {
             action.run();
             final boolean foundAll = latch.await(1, TimeUnit.SECONDS);
             if (!foundAll) {
-                throw new RuntimeException("Did not recieve all audit messages after a short wait.");
+                throw new RuntimeException("Did not recieve all " + expectedCount +" audit messages after a short wait.");
             }
             if (messages.size() != expectedCount) {
                 throw new RuntimeException("Unexpected number of messages, was expecting " + expectedCount + ", recieved " + messages.size());
--- a/src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
+++ b/src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
@@ -17,6 +17,9 @@ package org.opensearch.security.auditlog.integration;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.opensearch.common.settings.Settings;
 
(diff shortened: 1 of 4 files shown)


src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
```diff
@@ -25,23 +28,60 @@
...
+        final List<AuditMessage> messages = new ArrayList<>();
+        countDownRef.set(latch);
+        messagesRef.set(messages);
+
+        TestAuditlogImpl.sb = new StringBuffer();
+        TestAuditlogImpl.messages = messages; 
+        
+        try {
+            action.run();
+            final boolean foundAll = latch.await(1, TimeUnit.SECONDS);
```
davidlago, line 74: nit: magic number, perhaps extract to constant. Not used anywhere else so not that big of a deal, up to you.

davidlago (approved):
Thank you for adding this! Let me know if this proves/disproves your hypothesis after the change merges and it runs for a few days :)

peternied:
> Thank you for adding this! Let me know if this proves/disproves your hypothesis after the change merges and it runs for a few days :)

I've run this in loops on my dev machine, and its been solid as a rock so far, it will be interesting to see how it looks inside of GitHub actions.

src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
```diff
@@ -25,23 +28,61 @@
...
+    private static final AtomicReference<CountDownLatch> countDownRef = new AtomicReference<>();
+    private static final AtomicReference<List<AuditMessage>> messagesRef = new AtomicReference<>();
 
     public TestAuditlogImpl(String name, Settings settings, String settingsPrefix, AuditLogSink fallbackSink) {
         super(name, settings, null, fallbackSink);
     }
 
-
-    public synchronized boolean  doStore(AuditMessage msg) {
+    public boolean doStore(AuditMessage msg) {
```
cliu123, line 44: Would this break thread safety?

src/test/java/org/opensearch/security/test/helper/rest/RestHelper.java
```diff
@@ -136,48 +137,48 @@ public HttpResponse[] executeMultipleAsyncPutRequest(final int numOfRequests, fi
...
+	public HttpResponse executeOptionsRequest(final String request) {
         return executeRequest(new HttpOptions(getHttpServerUri() + "/" + request));
     }
 
-	public HttpResponse executePutRequest(final String request, String body, Header... header) throws Exception {
+	public HttpResponse executePutRequest(final String request, String body, Header... header) {
 		HttpPut uriRequest = new HttpPut(getHttpServerUri() + "/" + request);
 		if (body != null && !body.isEmpty()) {
-			uriRequest.setEntity(new StringEntity(body));
+			uriRequest.setEntity(createStringEntity(body));
```
cliu123, line 155: Why does `new StringEntity(body)` need to be moved to a separate method and throw `RuntimeException` from there?

peternied: When creating runnable they cannot throw an exception, this allows calls to these methods to be more concise.

`() -> executePutRequest(...)` vs
`() -> { try { return executePutRequest(...); } catch (Exception e) { throw new RuntimeException(e);}`

While in a production workflow capturing these individual exceptions has utility, they are almost never expected to be handled in this test code so runtime exception is appropriate.

DarshitChanpura (approved):

src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
peternied: Removed this originally to test its utility, added it back.

davidlago (approved):

cliu123:
There're whole bunch of `Assert.assertTrue(TestAuditlogImpl.sb.toString().contains(...)` in the codebase. Now that a new method is being introduced, can we migrate all of them with the new method? Using the same method for exactly same purpose in all places would retain consistency in the codebase.

src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
```diff
@@ -25,23 +28,61 @@
...
+        }
         sb.append(msg.toPrettyString()+System.lineSeparator());
-        messages.add(msg);
+        messagesRef.get().add(msg);
+        countDownRef.get().countDown();
         return true;
     }
 
+    /** Unneeded after switching to `doThenWaitForMessages(...)` as data is automatically flushed */
+    @Deprecated
```
cliu123, line 55: Is there any reason for keeping 2 different methods for the exactly same purpose? If not, replacing the old method with the new method would retain consistency.

src/test/java/org/opensearch/security/test/helper/rest/RestHelper.java
```diff
@@ -197,13 +198,27 @@ public HttpResponse executeRequest(HttpUriRequest uriRequest, Header... header)
 			HttpResponse res = new HttpResponse(httpClient.execute(uriRequest));
 			log.debug(res.getBody());
 			return res;
+		} catch (final Exception e) {
+			throw new RuntimeException(e);
```
cliu123, line 202: Now that `Exception` is being thrown with the existing implementation, why does it need to be converted to `RuntimeException`?

src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java
peternied: The testRestMethod has been the source of trouble, lets incrementally invest in the space.

If none of these other tests need to be updated, it will save us time/trouble. Each assert that was updated required investigation to confirm the correct behavior.  With  the api deprecation, this will help ensure new tests use the more reliable technique.

src/test/java/org/opensearch/security/test/helper/rest/RestHelper.java
peternied: `httpClient.execute(...)` can still throw IOException

peternied:
> There're whole bunch of `Assert.assertTrue(TestAuditlogImpl.sb.toString().contains(...)` in the codebase. Now that a new method is being introduced, can we migrate all of them with the new method? Using the same method for exactly same purpose in all places would retain consistency in the codebase.

There are 2,375 instances of this pattern in the code base, I've created [Reduce the amount of assertTrue/False statements](https://github.com/opensearch-project/security/issues/1762) so we don't rewrite the test infrastructure with this change.

DarshitChanpura (approved):

peternied:
Note; we aren't likely to make a 1.4 so I'm not backporting this change to that branch at this time.

Issue #1797, opened 2022-04-27
[BUG] Allow reload of same certificate using hot reload


shikharj05:
**What is the bug?**
Today, hot reload API doesn't allow reloading same certificate as it [requires](https://github.com/opensearch-project/security/blob/22822b4697d85512a5fe5c28c15af2ca5bae4b22/src/main/java/org/opensearch/security/ssl/DefaultSecurityKeyStore.java#L563) expiry date to be of later timestamp than current certificate. 

**How can one reproduce the bug?**
1. Initialize a cluster with http/transport certs
2. Try reloading http/transport certs using reload API with same certificates

Currently, there is a [test](https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/ssl/SecuritySSLReloadCertsActionTests.java#L216-L224) as well which replicates the above behavior.

**What is the expected behavior?**
A same certificate (validated using certificate signature) should be allowed to be reloaded. 

**What is your host/environment?**
 - OS: N/A
 - Version - All

**Do you have any additional context?**
In certain domains, after certificate hot-reload, hot reloaded nodes are not able to connect successfully to nodes bootstrapped with new certificate; existing (hot reloaded) nodes are able to connect to new nodes only after bouncing OpenSearch process on the node (which would now bootstrap with new certificate).

PR #1798, opened 2022-04-27
Allow same certificate during hot reload

shikharj05:
### Description
Allows reload of same certificate as existing certificate used in TLS for transport/http connections.
* Category - Bug Fix
* Why these changes are required? allow reloading same certificate for TLS- https://github.com/opensearch-project/security/issues/1797
* What is the old behavior before changes and new behavior after changes? old behavior- reloading with same certificate generates the following exception-
 ```{"error":"OpenSearchSecurityException[Error while initializing transport SSL layer from PEM: java.lang.Exception: New certificates should not expire before the current ones.]; nested: Exception[New certificates should not expire before the current ones.];"}```
new behavior- 
If certificate signatures match, skip further validation and allow reload.

### Issues Resolved
https://github.com/opensearch-project/security/issues/1797

Needs to be backported to all versions

### Testing
Added test cases.

### Check List


peternied (changes requested):
This is a great change thank you for the contribution.  When looking at the tests there are some gaps around make sure the cluster is still operational that I think we need to address.

src/main/java/org/opensearch/security/ssl/DefaultSecurityKeyStore.java
```diff
@@ -624,6 +629,34 @@ private boolean hasValidExpiryDates(final X509Certificate[] currentX509Certs, fi
...
+     */
+
+    private boolean areSameCerts(final X509Certificate[] currentX509Certs, final X509Certificate[] newX509Certs) {
+
+        final Function<? super X509Certificate, String> certificateSignature = cert -> {
+            final byte[] signature = cert !=null && cert.getSignature() != null ? cert.getSignature() : null;
+            return new String(signature);
+        };
+
+        final List<String> currentCertSignatureList = Arrays.stream(currentX509Certs)
```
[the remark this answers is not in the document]
shikharj05: Thank you for looking at this request, I have updated the PR with suggested changes.

src/test/java/org/opensearch/security/ssl/SecuritySSLReloadCertsActionTests.java
```diff
@@ -63,98 +69,70 @@ public class SecuritySSLReloadCertsActionTests extends SingleClusterTest {
...
-        rh.enableHTTPClientSSL = true;
-        rh.trustHTTPServerCertificate = true;
-        rh.sendAdminCertificate = true;
-        rh.keystore = "ssl/reload/kirk-keystore.jks";
+        String clusterHealthResponse = rh.executeSimpleRequest("_cluster/health");
+        JSONParser parser = new JSONParser();
+        JSONObject clusterHealthResponseJson = (JSONObject) parser.parse(clusterHealthResponse);
+        Assert.assertEquals("green", clusterHealthResponseJson.get("status"));
+
+        String catNodesResponse = rh.executeSimpleRequest("_cat/nodes?format=json");
```
peternied, line 94: Nicely done!

src/test/java/org/opensearch/security/ssl/SecuritySSLReloadCertsActionTests.java
```diff
@@ -63,98 +69,70 @@ public class SecuritySSLReloadCertsActionTests extends SingleClusterTest {
...
+        JSONArray catNodesResponseJson = (JSONArray) parser.parse(catNodesResponse);
+        Assert.assertEquals(3, catNodesResponseJson.size());
```
peternied, line 96: Rather than hardcode this number, can we pull it from the SingleClusterTest's properties?

src/test/java/org/opensearch/security/ssl/SecuritySSLReloadCertsActionTests.java
```diff
@@ -251,6 +189,117 @@ public void testSSLReloadFail_NoReloadSet() throws Exception {
...
+
+    /**
+     *
+     * @param rh RestHelper to perform rest actions on the cluster
+     * @param updateChannel certType/channel being updated, either http or transport
+     * @param expectedCertResponse expected Certs after reload
+     * @return True if all assertions pass
+     * @throws Exception if rest api failed
+     */
+    private Boolean reloadAndVerifyResponse(RestHelper rh, String updateChannel, JSONObject expectedCertResponse) throws Exception {
```
peternied, line 233: What do you think about renaming this method `assertThatReloadCertificates(...)`, then you don't have to pass a bool out, just make it void

peternied (changes requested):
Much cleaner, thank you.  A couple of small callouts in the test code.

src/test/java/org/opensearch/security/ssl/SecuritySSLReloadCertsActionTests.java
shikharj05: Sure, updated to assertReloadCertificateSuccess(...)

src/test/java/org/opensearch/security/ssl/SecuritySSLReloadCertsActionTests.java
shikharj05: Thanks, updated.

peternied (approved):

DarshitChanpura (approved):

DarshitChanpura (approved):

PR #819, opened 2020-11-01
Tenant Permissions : added the possibility to specify tenants via par…

chrousto:
Please provide as much details as possible to get feedback/acceptance on your PR quickly

**Category: (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation)**
    Enhancement / New Feature : specify tenants in roles via parameter substitution (like for indexes and DLS)

    Github Issue # or road-map entry, if available: 
    #817 

**Description of changes:**
The idea would be that this user :
```yaml
new-user:
  hash: "*************"
  reserved: false
  hidden: false
  opendistro_security_roles:
  - "role-tenant1"
  attributes:
    attribute1: "tenant1"
  static: false
```
Would only have access to the tenant tenant1 if the role role-tenant1 was defined like this :

```yaml
role-tenant1:
 reserved: false
 hidden: false
 cluster_permissions:
 - "read"
 - "cluster:monitor/nodes/stats"
 - "cluster:monitor/task/get"
 tenant_permissions:
 - tenant_patterns:
   - ${attr.internal.attribute1}
   allowed_actions:
   - "kibana_all_write"
 static: false
_meta:
 type: "roles"
 config_version: 2
```

**Why these changes are required?**
    Enhancement of the way tenants are specified in roles definitions. Allows for more generic roles to be created with information being received from the authorization backend for instance.

**What is the old behavior before changes and new behavior after changes? (Please add any example/logs/screen-shot if available)**
   Before the change, tenants could only be specified statically or via patterns in role definitions

**Testing done: (Please provide details of testing done: Unit testing, integration testing and manual testing)**
     Testing done manually

**TO-DOs, if any: (Please describe pending items and provide Github issues# for each of them)**
    Provide proper tests units.

**Is it backport from master branch? (If yes, please add backport PR # and commits #)**
    No

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #817, opened 2020-10-30
Feature Request : parameter substitution in tenant patterns.


chrousto:
Hello all,

I have a question regarding parameter substitution and tenants patterns in roles that I think would constitute an interesting enhancement.
(I have posted this in the forum, [here](https://discuss.opendistrocommunity.dev/t/tenants-and-parameter-substitution/4038) but it did not get much attention so I am taking the initiative to repost this here as a Feature Request / enhancement, sorry for the cross-post)

The Feature is this :
I would like to have parameter substitution for tenant patterns in roles (for the moment, my understanding is that it is only available for index permissions as well as dls : https://opendistro.github.io/for-elasticsearch-docs/docs/security/access-control/users-roles/#create-users).

The idea would be that this user :
```yaml
new-user:
  hash: "*************"
  reserved: false
  hidden: false
  opendistro_security_roles:
  - "role-tenant1"
  attributes:
    attribute1: "tenant1"
  static: false
```
Would only have access to the tenant tenant1 if the role role-tenant1 was defined like this :

```yaml
role-tenant1:
 reserved: false
 hidden: false
 cluster_permissions:
 - "read"
 - "cluster:monitor/nodes/stats"
 - "cluster:monitor/task/get"
 tenant_permissions:
 - tenant_patterns:
   - ${attr.internal.attribute1}
   allowed_actions:
   - "kibana_all_write"
 static: false
_meta:
 type: "roles"
 config_version: 2
```
I did not find anything on the roadmap yet or any feature request regarding this, do you think this would be an interesting feature to add ?

Thanks in advance.
Christophe.

hardik-k-shah:
Hey @chrousto , this sounds interesting feature to me. This will help admin to configure tenant permission dynamically through user attributes instead just defining in role.  In future, this can be also extended for 3rd party IdP. So, that tenant can be configured based on attributes of 3rd party users too.   Thanks @chrousto for bringing your creative thoughts. We are very happy to review if you can create PR for your changes.  

@jimishs thoughts?

francisco-hoo:
Will be useful have this feature but not only for tenant-patterns but for tenant-permissions too. Imagine with only one role, attach dynamically permissions to users based on their attributes...

For example:
`attr.internal|saml.role` with multiple values such `Admin`  or `Reader`
`attr.internal|saml.accessControl` with multiple values such `kibana_all_read`  or `kibana_all_write`

With this we can attach different permissions, without creating multiple resources (Roles)

[...]

PR #1813, opened 2022-04-29
Tenant Permissions : added the possibility to specify tenants via parameter

peternied:
### Description
This is a continuation of #819 with many thanks to @chrousto

The idea would be that this user :
```yaml
new-user:
  hash: "*************"
  reserved: false
  hidden: false
  opendistro_security_roles:
  - "role-tenant1"
  attributes:
    attribute1: "tenant1"
  static: false
```
Would only have access to the tenant tenant1 if the role role-tenant1 was defined like this :

```yaml
role-tenant1:
 reserved: false
 hidden: false
 cluster_permissions:
 - "read"
 - "cluster:monitor/nodes/stats"
 - "cluster:monitor/task/get"
 tenant_permissions:
 - tenant_patterns:
   - ${attr.internal.attribute1}
   allowed_actions:
   - "kibana_all_write"
 static: false
_meta:
 type: "roles"
 config_version: 2
```

### Issues Resolved
* #817 

### Testing
Integration tests have been created for this feature

### Check List


DarshitChanpura (approved):

src/test/java/org/opensearch/security/multitenancy/test/MultitenancyTests.java
```diff
@@ -387,4 +391,54 @@ public void testDashboardsAlias65() throws Exception {
...
+
+    @Test
+    public void testTenantParametersSubstitution() throws Exception {
+        final Settings settings = Settings.builder()
+                .build();
+        setup(settings);
+        final RestHelper rh = nonSslRestHelper();
+
+        HttpResponse res;
+        final String url = ".kibana/_doc/5.6.0?pretty";
```
cliu123, line 403: Why `5.6.0`?

peternied: This was from how the test were originally authored, would you like a different value in its place?

peternied:
@cliu123 Did you have any other comments around this pull request that need to be addressed?

cliu123 (approved):

src/test/java/org/opensearch/security/test/helper/rest/RestHelper.java
```diff
@@ -357,6 +362,38 @@ public String toString() {
...
+			try {
+				currentNode = DefaultObjectMapper.readTree(body);
+			} catch (final Exception e) {
+				throw new RuntimeException(e);
+			}
+
+			// Break the path into parts, and scan into the json object
+			try (final Scanner jsonPathScanner = new Scanner(jsonDotPath).useDelimiter("\\.")) {
+				do {
+					final String pathEntry = jsonPathScanner.next();
```
cliu123, line 383: This should be use while loop instead of `do...whilte`. Otherwise, this line could throw exception is `jsonPathScanner` doesn't have next.

cliu123 (changes requested):
Most of the changes look great to me! I missed the `do...while` in last review though. This can be merged when it gets fixed.

src/test/java/org/opensearch/security/test/helper/rest/RestHelper.java
peternied: The only case this would happen is if the string was an empty string, then per the javadoc it would throw `@throws NoSuchElementException if no more tokens are available`.  This seems an appropriate failure that would only happen during test authorship, not in practice after a test has been merged.

cliu123: Great! It's nice to identify the use cases where the exception could happen! Let's say the empty string case, if `NoSuchElementException` gets thrown, the method `findValueInJson` doesn't catch or throw it. And we are not supposed to let this exception happen here at all. It makes more sense to avoid the exception here by validating the condition before the operations unless it has to run the logic in `do` block before evaluating the condition.

peternied: I think this is a matter of preference, I think the exception is good enough, if this becomes a problem over time we can enhance it.  Is this issue worth blocking the pull request?

src/test/java/org/opensearch/security/multitenancy/test/MultitenancyTests.java
cliu123: I think it looks fine here.

src/test/java/org/opensearch/security/test/helper/rest/RestHelper.java
cliu123: There's a valid case here as discussed above. When `jsonDotPath` is an empty string, `final String pathEntry = jsonPathScanner.next();` throws `NoSuchElementException` that is not even caught or thrown. I actually don't see a value of using `do...while` here. In fact, it's error-prone. 
This contribution is valuable, it's only a minor fix away from being merged.

cliu123: Did this get marked as resolved in accident? unresolving this conversation

peternied: This was resolved, see the new lines added above to cover this scenario

cliu123 (approved):

peternied:
@DarshitChanpura @cliu123 could I get another review?

cliu123 (approved):

DarshitChanpura (approved):

PR #1828, opened 2022-05-06
Add suppression for all removal warnings

peternied:
### Description

Use of the SecurityManager and AccessController have been deprecated and
will be removed in java versions after 17.  While this is an issue its
also one that will take a concerted effort to resolve.  These warning
messages making discovering build errors and other warnings more
difficult; hence adding this supression logic.

For tracking the effort to replace these components look into https://github.com/opensearch-project/OpenSearch/issues/1687

### Issues Resolved
* Related to https://github.com/opensearch-project/OpenSearch/issues/1687

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -379,6 +379,12 @@ tasks.withType(Checkstyle) {
     }
 }
 
+tasks.withType(JavaCompile) {
+    configure(options) {
+        options.compilerArgs << '-Xlint:removal' << '-Werror'
+    }
+}
+
 spotless {
   java {
     // note: you can use an empty string for all the imports you didn't specify explicitly, and '\\#` prefix for static imports
--- a/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
+++ b/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
@@ -78,6 +78,7 @@ public abstract class AbstractHTTPJwtAuthenticator implements HTTPAuthenticator 
     }
 
     @Override
+    @SuppressWarnings("removal")
     public AuthCredentials extractCredentials(RestRequest request, ThreadContext context)
             throws OpenSearchSecurityException {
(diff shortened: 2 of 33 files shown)


DarshitChanpura (approved):

cliu123 (changes requested):
A couple of questions here:
1. Why removing warning message instead of disabling Java Security Manager?
2. Quoted from the issue:
> We could keep it as long as we can, but once removed from the JDK, it will be a problem.

If disabling JSM, what the problem will be?

DarshitChanpura (approved):

cliu123:
> Deprecation messages are useful if we were adding a component that we did not realize was deprecated and wanted to gate that dependency.

Exactly, JSM being deprecated should be warned in this case, shouldn't it? OpenSearch core doesn't suppress these warns either.

peternied:
@cliu123 This pull request doesn't eliminate deprecation messages, it suppresses them on components that we have no plans to fix.  It is confusing for contributors or even maintainers such as myself to see 'warnings' in our build output that we should actively ignore, which is why I am suppressing warnings on these components.

If a new warning was to be added in a pull request we would have no mechanism to detect it.  By suppressing these messages and failing builds on any new messages this puts the codebase in a better posture to prevent new dependencies on deprecated/removed features.

Is there a reason to keep the existing warnings as they are today?

cliu123:
> This pull request doesn't eliminate deprecation messages, it suppresses them on components that we have no plans to fix.

I have a couple of questions here:
1. Will the deprecation messages still show? Could you elaborate more in the PR description: 
What exactly deprecatio messages are these changes suppressing? Is it the following?
```
WARNING: System::setSecurityManager will be removed in a future release
```
2. Given that JSM is deprecated and will be removed, why shouldn't security plugin warn users this? 
@reta @nknize @dblock might have more context on this. How are these deprecation warnings handled in OpenSearch core where JSM is also being used?

cliu123:
I see. Thanks for the clarification! @peternied @reta

cliu123 (approved):

PR #1829, opened 2022-05-06
Updates license headers

### LINKED ISSUE #1744: [BUG] Incorrect license headers
Coming from https://github.com/opensearch-project/opensearch-plugins/issues/138, some obsolete ALv2 headers are still present, e.g. [here](https://github.com/opensearch-project/security/blob/2972bd15a98dbaa8c571b9b302268de076d20ee0/src/test/java/org/opensearch/security/TracingTests.java#L17). 

1. Add automated license header Gradle checks to CI similar to [this](https://github.com/opensearch-project/asynchronous-search/commit/bc483fe7a253d20216ce5fc4d5427e42819b7aaf).
2. Update the default header for all new files created by IDEs similar to https://github.com/opensearch-project/OpenSearch/pull/510.
3. Fix any incorrect headers.

Licenses must match https://github.com/opensearch-project/opensearch-plugins/blob/main/HEADERS.md

DarshitChanpura:
### Description
Updates license headers
* Category : Bug fix

### Issues Resolved
* Resolves #1744

### Check List

### DIFF
--- a/config/checkstyle/sun_checks.xml
+++ b/config/checkstyle/sun_checks.xml
@@ -204,4 +204,12 @@
     <property name="severity" value="warning"/>
   </module>
 
+  <!-- Checks for Headers to not contain old license text -->
+  <module name="RegexpSingleline">
+    <property name="format" value="\*( Portions)? Copyright OpenSearch Contributors"/>
+    <property name="ignoreCase" value="true"/>
+    <property name="message" value="Incorrect license header" />
+    <property name="severity" value="error"/>
+  </module>
+
 </module>
--- a/plugin-security.policy
+++ b/plugin-security.policy
@@ -16,18 +16,14 @@
  */
  
  /*
- * Portions Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
  *
- * Licensed under the Apache License, Version 2.0 (the "License").
- * You may not use this file except in compliance with the License.
- * A copy of the License is located at
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
  *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * or in the "license" file accompanying this file. This file is distributed
- * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
- * express or implied. See the License for the specific language governing
- * permissions and limitations under the License.
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
  */
  
  
--- a/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
+++ b/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
@@ -1,16 +1,12 @@
 /*
- * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
  *
- *  Licensed under the Apache License, Version 2.0 (the "License").
- *  You may not use this file except in compliance with the License.
- *  A copy of the License is located at
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
  *
- *  http://www.apache.org/licenses/LICENSE-2.0
- *
- *  or in the "license" file accompanying this file. This file is distributed
- *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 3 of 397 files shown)


DarshitChanpura:
@opensearch-project/security team, let's get this reviewed and merged and then we can open a follow-up issue to generate new license headers on new file creation.

peternied (changes requested):
Taking this change without being sure that we correctly updated all of the files makes it hard to know if this is the correct pull request to merge.  I would rather we solved the problem of detecting incorrect license headers first and then rolled out the fix for the improperly documented files.

I believe that OpenSearch already has a [process](https://github.com/opensearch-project/OpenSearch/blob/main/buildSrc/src/main/groovy/org/opensearch/gradle/precommit/PrecommitTasks.groovy#L46) for doing this, can we see if we can adopt it or a similar process?

DarshitChanpura:
> Taking this change without being sure that we correctly updated all of the files makes it hard to know if this is the correct pull request to merge. I would rather we solved the problem of detecting incorrect license headers first and then rolled out the fix for the improperly documented files.
> 
> I believe that OpenSearch already has a [process](https://github.com/opensearch-project/OpenSearch/blob/main/buildSrc/src/main/groovy/org/opensearch/gradle/precommit/PrecommitTasks.groovy#L46) for doing this, can we see if we can adopt it or a similar process?

The license headers are updated according to the step mentioned in the bug attached here. I have replaced all the headers similar to [this](https://github.com/opensearch-project/security/blob/2972bd15a98dbaa8c571b9b302268de076d20ee0/src/test/java/org/opensearch/security/TracingTests.java#L17) in all the files to match the header mentioned [here](https://github.com/opensearch-project/opensearch-plugins/blob/main/HEADERS.md). 

The follow-up to this would to add automated checks i.e. one which you mentioned in the comment, using a groovy task or an alternative in our project which would need more time. 

I did try by adding a task using spotless, but that would replace all the license headers with the new one which is not correct, so ended up removing it and would need an extra set of hands in getting this fix through.

peternied:
> The follow-up to this would to add automated checks i.e. one which you mentioned in the comment, using a groovy task or an alternative in our project which would need more time.

Could we add a check for the bad license header to make sure we don't add any more violations like these back in?  I don't think we need a 'license header corrector' tool

config/checkstyle/sun_checks.xml
```diff
@@ -204,4 +204,12 @@
     <property name="severity" value="warning"/>
   </module>
 
+  <!-- Checks for Headers to not contain old license text -->
+  <module name="RegexpSingleline">
+    <property name="format" value="Portions Copyright OpenSearch Contributors"/>
```
DarshitChanpura, line 209: For now I'm checking the starting line of the old license header to see if any files contain it, if so checkstyle will throw violation.

DarshitChanpura:
> > The follow-up to this would to add automated checks i.e. one which you mentioned in the comment, using a groovy task or an alternative in our project which would need more time.
> 
> Could we add a check for the bad license header to make sure we don't add any more violations like these back in? I don't think we need a 'license header corrector' tool

Done

config/checkstyle/sun_checks.xml
peternied, line 209: Looks like that regex needs a bit of improvment, here 
```suggestion
    <property name="format" value="\*( Portions)? Copyright OpenSearch Contributors"/>
```

With this change I found 5 errors in the following files:
```
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/ssl/transport/DualModeSSLHandler.java:2: Incorrect license header [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/ssl/transport/SSLConfig.java:2: Incorrect license header [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/ssl/util/SSLConnectionTestUtil.java:2: Incorrect license header [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/ssl/util/TLSUtil.java:2: Incorrect license header [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/tools/SecurityAdmin.java:17: Incorrect license header [RegexpSingleline]
```

peternied (changes requested):

config/checkstyle/sun_checks.xml
DarshitChanpura: good catch... thank you!

peternied:
> should we update NOTICE.txt?

I don't think we should modify the NOTICE.txt at this time.

peternied (approved):

cliu123 (approved):

Issue #1787, opened 2022-04-25
Integrate auto-password generation into security plugin bootstrap logic


setiah:
The proposal is to integrate the PasswordSetup logic that generates strong secure passwords into the security plugin bootstrap logic, so that the secure passwords are auto-generated when the OpenSearch process w/ security plugin bootstraps. This feature to provide secure out of the box password should be behind a setting so that it can be disabled/enabled at will. Initially in 2.x versions, we might want to default it to false so that we don't break users currently using admin:admin but going forward, we may want to switch defaults to auto-generate strong passwords during first bootstrap. This will enable deprecation of admin:admin default password.

setiah:
The plan here is to add the bootstrap time auto-password generation logic behind a feature flag, with default disabled. In 2.x line, users can explicitly enable the flag to get out of the box strong passwords for in-built security users ("admin", "kibanaserver", "kibanaro", etc.). With the next major version, the flag can be changed to be enabled by default, which will bootstrap the cluster with default strong auto-generated passwords.

setiah:
@ryanbogan I tested the current [implementation](https://github.com/ryanbogan/security/commit/114ba36716cbd6fd7f3802295ef2de60b7849910) for auto-generating passwords on bootstrap. There are a couple of issues that need to be fixed - 

1. The passwords get auto-generated each time the node bootstraps even when the earlier bootstrap generated the password. Thus the previous password ends up getting overwritten. This needs to be fixed. The password generation should happen only once during bootstrapping.
2. The passwords are not shared in a user-friendly way in the logs. They seem to be scattered all over the place in the logs. The bootstrapping process on completion should display passwords in a legible way to the users.
3. The password generation and sharing logic should work for docker and rpm distribution as well. Users setting up OpenSearch with docker and rpm should be able to read auto-generated passwords in a user-friendly way.
4. Add mechanism for this to work with docker-compose, as in case of docker-compose there is no way to share auto-generated passwords during bootstrapping.

derek-ho:
Closing this due to implementation of OPENSEARCH_INITIAL_ADMIN_PASSWORD in 2.12

PR #1836, opened 2022-05-11
Password setup tool for built-in users

### LINKED ISSUE #1576: Replace `admin:admin` default credentials
**Is your feature request related to a problem? Please describe.**
The default `admin:admin` credential is too weak and a sure-fire way to get hacked. Though the documentation recommends users to change the default passwords, it can be easily missed. Such weak defaults can be very easily exploited and lead to massive data leaks (remember [Equifax attack](https://www.forbes.com/sites/kateoflahertyuk/2019/10/20/equifax-lawsuit-reveals-terrible-security-practices-at-time-of-2017-breach/?sh=446983d93d38)?). 

The second gap is, in the existing distribution, there's no default password policy to ensure stronger passwords. It needs to be configured explicitly by users. That is why such weak credentials are able to slip through the cracks.

**Describe the solution you'd like**
Remove the default passwords provided in `internal_users.yml`. Instead, have the `admin` user setup passwords for internal users once the OpenSearch process is up. These passwords can be set up post-installation using a REST API and `admin` certificates (for authentication). Setting password for `admin` user would look something like -
```
PATCH _opendistro/_security/api/internalusers/admin --key admin-key.pem --cert admin.pem --cacert root-ca.pem -d'
{
    "password": "quCn52_8\*N8"
}
'
```
To make this further easy, OpenSearch distribution could provide a tool such as an OpenSearch shell client to make it easy for admin user to configure all users. This shell client could also help with autogenerating strong passwords and can be further extended to support for more administrative functions in future. 
 
In addition to this, we should also add a default password policy to only allow passwords that match a certain strength criteria. Obv, these policies are overridable depending on the user/enterprise needs. Having a strong default policy will protect against insecure password configurations.

**Describe alternatives you've considered**

**Additional context**
This is related to https://github.com/opensearch-project/OpenSearch/issues/1618

**Task breakdown**
- [x] Remove default passwords for inbuilt users in internal_users.yml, deprecate now and remove password field from yml in 2.0
- [x] Configure strong default password policy and also make it overridable via yml.
- [x] Write tool for password setup (to be used by admin).
- [x] Add auto-password-generation logic to the tool to ease user experience during setup.
- [ ] #1787

setiah:
### Description
This PR primarily provides new password setup tool for setting up passwords for inbuilt security users (admin, kibanaserver, etc.) and also provides a new unified interface for extending tooling in opensearch-security. 

> Details of all the changes
1. Provides a new unified interface to simplify and extend tooling in opensearch-security.
2. Adds setup-password functionality to allow users to setup secure passwords for built-in users, instead of relying on weak defaults such as admin:admin. The functionality is provided in both manual and auto mode for auto-password generation.
3. Adds default secure password policy to opensearch.yml

> What is the old behavior before changes and new behavior after changes?

In the old behavior, users had to update the internal_users.yml file on all nodes with the hash of the password to change the passwords from admin/admin. While that still works, with the new tool, they can simply run the tool to setup secure passwords after installation. The tool runs as `opensearch-admin setup-passwords [--auto]` or to setup policy compliant passwords for inbuilt users (both manually and automatically). It can be extended to support other functions such as `setup-certificates`, `reload-configs`, through its simplified extensible interface. 

### Issues Resolved
#1576 

No
### Testing
Unit testing, Manual testing

### Check List


setiah:
@peternied can you help with reviewing this PR? 
cc: @ryanbogan

setiah:
@peternied can you help with reviewing this PR? 
cc: @ryanbogan

CEHENKLE:
@opensearch-project/security Hey folks -- can we get some code review luv over here?  

Thanks :) 

/C

src/test/java/org/opensearch/security/SlowIntegrationTests.java
```diff
@@ -30,6 +30,7 @@
 
 package org.opensearch.security;
 
+
```
cliu123, line 33: Please remove this empty line if not necessary.

peternied (changes requested):
Thanks for the pull request.  Making sure that new clusters have strong security posture is an important space for us.

This pull request changes the responsibility of the security plugin by adding a new OpenSearch-Admin script.  This is a public interface in the security repository that over time will likely need to be deprecated based on de-emphasizing security as a plugin.  We should avoid this, I would recommend baking these changes into a cluster initialization process.

Accepting custom passwords creates the opportunity for bad password practices.  OpenSearch would have a better posture if we had a standard password generation of say 20+ characters of entropy.  This could be done with modifications to the existing create user API or as part of the initialization process to standard out.

Please reduce the scope of this PR to an automatic (strong) password generation during the initialization process, and extract the admin tool/script creation into a separate draft PR so we can have that larger discussion over there, and weigh the pros and cons of adding a tool that will need to be deprecated vs postponing that effort until we have a broader plan for admin tools moving forward.

setiah:
@peternied 

There is a separate [issue](https://github.com/opensearch-project/security/issues/1787) tracking auto-password setup during cluster bootstrap, which leverages the PasswordSetup logic added by this tool to autogenerate passwords. But passwords are not just setup during the first time setup. They also need to be rotated many times during the lifecycle of a cluster based on security needs of an organization. This tool provides an easy way to for cluster administrator to set the passwords first time (until we have [this](https://github.com/opensearch-project/security/issues/1787)) and rotate/update them later. Hence it would be helpful.

>This pull request changes the responsibility of the security plugin by adding a new OpenSearch-Admin script. This is a public interface in the security repository that over time will likely need to be deprecated based on de-emphasizing security as a plugin. 

The public interface is for Security plugin tooling and is not needed in other plugins. All tools in Security plugin (hasher, securityadmin, install_demo_configurations, etc.) can extend this interface to inherit some common functions, and just build their business logic. This will enable java-fying the security plugin tools so they are unit testable like code, unlike some shell tools in security plugin repo today that need to be validated manually. Happy to consider if you have any other suggestions for the tool name.

>Accepting custom passwords creates the opportunity for bad password practices. This could be done with modifications to the existing create user API or as part of the initialization process to standard out.

I think you missed this change mentioned in the PR description - "Adds default secure password policy to opensearch.yml". The custom passwords are validated against this default secure password policy that is being adding with this change. If the password is not policy compliant, it gets rejected. 
Lack of a default secure password policy is a gap today which allows users to create weak passwords 

I'd request you to take a look again.

peternied:
> This tool provides an easy way to for cluster administrator to set the passwords first time (until we have https://github.com/opensearch-project/security/issues/1787) and rotate/update them later. Hence it would be helpful.

The security plugin already has an API, there is no extra burden for documentation, maintenance or release management.  A new endpoint such as  `PUT user/{user}/rotateCredentials`, or improving existing endpoints for this scenario seems very reasonable.  Exposing this functionality with an API expands the utility so admins could make API calls directly, use the Dashboard UX, or use another plugin to go through the same workflow.  If this is built as a tool its utility is limited, lets build this functionality so it can be reused.  Is there a reason to limit the functionality?

> I think you missed this change mentioned in the PR description - "Adds default secure password policy to opensearch.yml". 

This change was done in the `tools/install_demo_configuration.sh` file, outside of the recommended configuration.  There would be much more benefit to make this change as part of the security default configuration.  What do you think about separating password policy into another pull request?

setiah:
> The security plugin already has an API, there is no extra burden for documentation, maintenance or release management. A new endpoint such as `PUT user/{user}/rotateCredentials`, or improving existing endpoints for this scenario seems very reasonable. Exposing this functionality with an API expands the utility so admins could make API calls directly, use the Dashboard UX, or use another plugin to go through the same workflow. If this is built as a tool its utility is limited, lets build this functionality so it can be reused. Is there a reason to limit the functionality?

The functionality already exists today with a PATCH API. The tool just builds on the capability of existing PATCH API to simplify password setup process for all internal users. I do think there is value in having a tool based mechanism over and above REST API as it makes life easier for users. 


> This change was done in the `tools/install_demo_configuration.sh` file, outside of the recommended configuration. There would be much more benefit to make this change as part of the security default configuration. What do you think about separating password policy into another pull request?

I'm okay with this suggestion.

Issue #1839, opened 2022-05-13
[Bug] Switch to standard OpenSearch gradle build process 


peternied:
All plugins use a standard set of gradle tasks provided by OpenSearch, this allows for consistent build/test and other validation workflows.

Follow processes from;
* [Example build.gradle](https://github.com/opensearch-project/opensearch-plugin-template-java/blob/main/build.gradle)
* [Detailed building processes](https://github.com/opensearch-project/opensearch-plugins/blob/main/BUILDING.md#building-plugins-with-opensearch)

### Related Issues:
* https://github.com/opensearch-project/security/issues/1816
* https://github.com/opensearch-project/security/issues/1801
* https://github.com/opensearch-project/OpenSearch/issues/3455

Potential starting point branch https://github.com/peternied/security/tree/use-plugin-gradle

PR #1855, opened 2022-05-18
Bump version to 2.1.0.0

cliu123:

### Description
* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation) Maintenance
* Bump security plugin version to 2.1.0.0 and OpenSearch core version to 2.1.0.
* Upgrading Gradle to [the version that OpenSearch core uses.](https://github.com/opensearch-project/OpenSearch/blob/main/gradle/wrapper/gradle-wrapper.properties#L14)
```
* What went wrong:
A problem occurred evaluating root project 'bwc-test'.
> Failed to apply plugin class 'org.opensearch.gradle.info.GlobalBuildInfoPlugin'.
   > Gradle 7.4.1+ is required
```

### Testing
UTs

### Check List

### DIFF
--- a/bwc-test/build.gradle
+++ b/bwc-test/build.gradle
@@ -47,7 +47,7 @@ ext {
 
 buildscript {
     ext {
-        opensearch_version = System.getProperty("opensearch.version", "2.0.0-SNAPSHOT")
+        opensearch_version = System.getProperty("opensearch.version", "2.1.0-SNAPSHOT")
         opensearch_group = "org.opensearch"
     }
     repositories {
@@ -76,13 +76,13 @@ dependencies {
 String bwcVersion = "1.3.0.0";
 String baseName = "securityBwcCluster"
 String bwcFilePath = "src/test/resources/"
-String projectVersion = "2.0.0.0"
+String projectVersion = "2.1.0.0"
 
 2.times {i ->
     testClusters {
         "${baseName}$i" {
             testDistribution = "ARCHIVE"
-        versions = ["1.3.0","2.0.0"]
+        versions = ["1.3.0","2.1.0"]
             numberOfNodes = 3
             plugin(provider(new Callable<RegularFile>() {
                 @Override
--- a/bwc-test/build.gradle
+++ b/bwc-test/build.gradle
@@ -73,7 +73,7 @@ dependencies {
     testImplementation "org.opensearch.test:framework:${opensearch_version}"
 }
 
-String bwcVersion = "1.3.0.0";
+String bwcVersion = "2.0.0.0";
 String baseName = "securityBwcCluster"
 String bwcFilePath = "src/test/resources/"
 String projectVersion = "2.1.0.0"
@@ -82,7 +82,7 @@ String projectVersion = "2.1.0.0"
     testClusters {
         "${baseName}$i" {
             testDistribution = "ARCHIVE"
-        versions = ["1.3.0","2.1.0"]
+        versions = ["2.0.0","2.1.0"]
             numberOfNodes = 3
             plugin(provider(new Callable<RegularFile>() {
                 @Override
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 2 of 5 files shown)


bwc-test/build.gradle
```diff
@@ -76,13 +76,13 @@ dependencies {
...
 String baseName = "securityBwcCluster"
 String bwcFilePath = "src/test/resources/"
-String projectVersion = "2.0.0.0"
+String projectVersion = "2.1.0.0"
 
 2.times {i ->
     testClusters {
         "${baseName}$i" {
             testDistribution = "ARCHIVE"
-        versions = ["1.3.0","2.0.0"]
```
peternied, line 85: Should these versions be additive instead of replacing the previous version?

bwc-test/build.gradle
```diff
@@ -76,13 +76,13 @@ dependencies {
 String bwcVersion = "1.3.0.0";
 String baseName = "securityBwcCluster"
 String bwcFilePath = "src/test/resources/"
-String projectVersion = "2.0.0.0"
```
peternied, line 79: Can we pull this from the other gradle configuration? so we don't have to keep updating both files?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
```
peternied, line 3: Why the gradle upgrade, is it required? (I don't see details in the description about this)

cliu123: added the details to the PR description

bwc-test/build.gradle
cliu123: Upgrading to `2.0.0` is tested on `2.0` branch.

bwc-test/build.gradle
cliu123: This is the only occurrence. We don't have to pull it to other places.

bwc-test/build.gradle
peternied: My concern is if we add an incompatibility in version in 2.1 that doesn't work with 2.0.  Can you help me find where we test this scenario?

bwc-test/build.gradle
peternied: > This is the only occurrence.

How is this the case? This pull request has effectively the same change at build.gradle@60, bwc-test/build.gradle@50 and bwc-test/build.gradle@79

cliu123: I see what you're referring to now. I thought that you were referring to this line#79 specifically. This is security plugin version. The others are OpenSearch version.
The 2 files are in 2 project folders. `bwc-test` is a sub-project of security plugin project. I don't see how the versions in 2 different project can be extracted to a central place though.

bwc-test/build.gradle
cliu123: That's a great point! The test should validate the upgrade from 2.0.0 to 2.1.0.

peternied: According to our support matrix, we support the last major and the current previous major, so we should have tests for `1.3 -> 2.1` and `2.0 -> 2.1`, can we make sure we test both?

stated differently, we should have tests that cover: 
* last supported major version (1) and its most recent minor version (3) to current version
* current major version (2) and its most recent minor version (0) to current version 

if we really wanted to be thorough I think we would add checks for every minor version on the current major version (e.g. if we were on 2.4, we would have BWC tests for 1.3, 2.0, 2.1, 2.2, and 2.3) (e.g.2 if we were on 3.1 we would have BWC tests for 2.3 and 3.0)

I think we should codify your preferred plan for this is in a comment side by side with these values to make future changes clear.

bwc-test/build.gradle
peternied: Let's table this for after we've correctly rewritten the build.gradle https://github.com/opensearch-project/security/issues/1839

bwc-test/build.gradle
cliu123: >the current previous major

Do you mean current previous minor?

> According to our support matrix, we support the last major and the current previous major, so we should have tests for 1.3 -> 2.1 and 2.0 -> 2.1

Could you post the link to the support matrix? I don't see any plugins testing upgrade from multiple previous versions to the current version though.

cliu123: the current previous major

Do you mean current previous minor?

According to our support matrix, we support the last major and the current previous major, so we should have tests for 1.3 -> 2.1 and 2.0 -> 2.1

Could you post the link to the support matrix? We should follow if there is a consistent requirements, but I don't see any plugins testing upgrade from multiple previous versions to the current version though.

peternied: @cliu123 I've created https://github.com/opensearch-project/OpenSearch/issues/3440 to follow up for broad guidance.  Lets move ahead with the work you have done here.

DarshitChanpura (approved):

PR #1865, opened 2022-05-26
Bump version to 2.1.0.0

cliu123:
Signed-off-by: cliu123 <lc12251109@gmail.com>

### Description
* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation) Maintenance
* Bump security plugin version to 2.1.0.0 and OpenSearch core version to 2.1.0.
* Upgrading Gradle to [the version that OpenSearch core uses.](https://github.com/opensearch-project/OpenSearch/blob/main/gradle/wrapper/gradle-wrapper.properties#L14)
```
* What went wrong:
A problem occurred evaluating root project 'bwc-test'.
> Failed to apply plugin class 'org.opensearch.gradle.info.GlobalBuildInfoPlugin'.
   > Gradle 7.4.1+ is required
```
* Changes in `DlsFlsFilterLeafReader.java` is required for new Lucene version.

### Testing
UTs

### Check List
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1882, opened 2022-06-09
Revert "Bump version to 2.1.0.0"

cliu123:
Reverts opensearch-project/security#1865


peternied:
@cliu123 Please create a bug in our repo with the failure justifying this revert, then we should create another bug in OpenSearch to fix whatever is busted.

PR #1883, opened 2022-06-10
Bump version to 2.1.0.0

cliu123:
### Description
* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation) Test fix

### Testing

### Check List

### DIFF
--- a/src/test/java/org/opensearch/security/dlic/rest/api/AbstractRestApiUnitTest.java
+++ b/src/test/java/org/opensearch/security/dlic/rest/api/AbstractRestApiUnitTest.java
@@ -195,11 +195,11 @@ public abstract class AbstractRestApiUnitTest extends SingleClusterTest {
 		rh.sendAdminCertificate = sendAdminCertificate;
 	}
 
-	protected String checkReadAccess(int status, String username, String password, String indexName, String type,
+	protected String checkReadAccess(int status, String username, String password, String indexName, String actionType,
 			int id) throws Exception {
 		boolean sendAdminCertificate = rh.sendAdminCertificate;
 		rh.sendAdminCertificate = false;
-		String action = indexName + "/" + type + "/" + id;
+		String action = indexName + "/" + actionType + "/" + id;
 		HttpResponse response = rh.executeGetRequest(action,
 				encodeBasicHeader(username, password));
 		int returnedStatus = response.getStatusCode();
@@ -209,12 +209,12 @@ public abstract class AbstractRestApiUnitTest extends SingleClusterTest {
 
 	}
 
-	protected String checkWriteAccess(int status, String username, String password, String indexName, String type,
+	protected String checkWriteAccess(int status, String username, String password, String indexName, String actionType,
 			int id) throws Exception {
 
 		boolean sendAdminCertificate = rh.sendAdminCertificate;
 		rh.sendAdminCertificate = false;
-		String action = indexName + "/" + type + "/" + id;
+		String action = indexName + "/" + actionType + "/" + id;
 		String payload = "{\"value\" : \"true\"}";
 		HttpResponse response = rh.executePutRequest(action, payload,
 				encodeBasicHeader(username, password));
--- a/src/test/java/org/opensearch/security/dlic/rest/api/RolesMappingApiTest.java
+++ b/src/test/java/org/opensearch/security/dlic/rest/api/RolesMappingApiTest.java
@@ -337,8 +337,6 @@ public class RolesMappingApiTest extends AbstractRestApiUnitTest {
 		rh.sendAdminCertificate = false;
 		checkReadAccess(HttpStatus.SC_OK, "picard", "picard", "sf", "_doc", 1);
 		checkWriteAccess(HttpStatus.SC_OK, "picard", "picard", "sf", "_doc", 1);
-		// ES7 only supports one doc type, so trying to create a second one leads to 400  BAD REQUEST
-		checkWriteAccess(HttpStatus.SC_BAD_REQUEST, "picard", "picard", "sf", "public", 1);
 	}
 
 	private void checkAllSfForbidden() throws Exception {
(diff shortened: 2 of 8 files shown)


src/test/java/org/opensearch/security/dlic/rest/api/AbstractRestApiUnitTest.java
```diff
@@ -195,11 +195,11 @@ protected void checkGeneralAccess(int status, String username, String password)
 		rh.sendAdminCertificate = sendAdminCertificate;
 	}
 
-	protected String checkReadAccess(int status, String username, String password, String indexName, String type,
+	protected String checkReadAccess(int status, String username, String password, String indexName, String actionType,
```
cliu123, line 198: Fixing the naming here as OpenSearch 2.0+ supports actionType in the API path and dropped the support for type.

src/test/java/org/opensearch/security/dlic/rest/api/RolesMappingApiTest.java
```diff
@@ -337,8 +337,6 @@ private void checkAllSfAllowed() throws Exception {
 		rh.sendAdminCertificate = false;
 		checkReadAccess(HttpStatus.SC_OK, "picard", "picard", "sf", "_doc", 1);
 		checkWriteAccess(HttpStatus.SC_OK, "picard", "picard", "sf", "_doc", 1);
-		// ES7 only supports one doc type, so trying to create a second one leads to 400  BAD REQUEST
-		checkWriteAccess(HttpStatus.SC_BAD_REQUEST, "picard", "picard", "sf", "public", 1);
```
cliu123, line 341: Removing this assertion to align with the behavior change in OpenSearch 2.0.1.

peternied (changes requested):
This change drops BWC support for 1.3.0, we should instead add support for 2.0.0 - following the guidance from the [bwc version test matrix](https://github.com/opensearch-project/opensearch-plugins/blob/main/TESTING.md#bwc-version-test-matrix).

cliu123:
> This change drops BWC support for 1.3.0, we should instead add support for 2.0.0 - following the guidance from the [bwc version test matrix](https://github.com/opensearch-project/opensearch-plugins/blob/main/TESTING.md#bwc-version-test-matrix).

@peternied I see that you made the changes(`BWC Version test matrix`) on [the requirements for BWC tests](https://github.com/opensearch-project/opensearch-plugins/blob/main/TESTING.md#bwc-version-test-matrix):
```
At the time of this writing there are OpenSearch releases for 1.0, 1.1, 1.2, 1.3, and 2.0. To ensure bwc tests have appropriate coverage when building components for 2.1, the bwc test matrix should include 1.3, and 2.0.
```
Was there any discussion on this topic? I don't know any plugins plan to do this though.
Cc: @CEHENKLE @bbarani @VachaShah @saratvemulapalli @ohltyler for inputs.

saratvemulapalli:
@cliu123 we do not mandate this yet but that is our recommendation to test these versions.
There was a discussion to clarify this: https://github.com/opensearch-project/OpenSearch/issues/3440

cliu123:
@saratvemulapalli Thanks for fill me in the context!
@peternied Looks like we don't have to block the version bump on this as this is not mandated yet. We can re-visit this in a separate PR/issue if this will turn to a mandated global campaign across the entire OpenSearch org.

peternied:
@cliu123 Could you attempt to support BWC tests for 1.3.0 and 2.0.0 and if there are substantive issues we can create an issue to follow up in a separate PR?

cliu123:
@saratvemulapalli @peternied With the design of the BWC test framework , supporting multiple upgrade paths would require significant changes. There is not an available way as easy as to support build with multiple JDKs using matrix. I don't think mixing up these changes with the version bump is what we want to do. 
If necessary, let's track this with a separate issue. We can get inputs from @VachaShah and other people who have more context on the BWC test framework and see how this should be done more generically.

peternied:
> With the design of the BWC test framework , supporting multiple upgrade paths would require significant changes.

Could you quantify this statement with concrete issues? If we put off doing this work because it is hard, we should articulate this to the OpenSearch team so it can be solved for everyone.

cliu123:
@peternied This PR is supposed to include only the version changes. Other changes in BWC tests if required should be in a separate PR. I am not talking about how hard this change would be, but I just don't see why a version bump should be blocked by BWC test changes that are supposed to be looked into separately. IMO, things that haven't been well-defined shouldn't block another PR unless the PR being blocked has hard dependencies on that. 
@VachaShah Do you happen to have answer for @peternied 's question above as you have the most context on the BWC test framwork? I don't see a way to support multiple upgrade paths without creating more test clusters. If there's a generic solution for this that can be applied to all OpenSearch components, would you please provide? Thanks!

cliu123:
@peternied This PR is supposed to include only the version changes. Other changes in BWC tests if required should be in a separate PR. No matter how hard this change would be, I just don't see why a version bump should be blocked by BWC test changes that are supposed to be looked into separately. IMO, things that haven't been well-defined shouldn't block another PR unless the PR being blocked has hard dependencies on that. 
@VachaShah Do you happen to have answer for @peternied 's question above as you have the most context on the BWC test framwork? I don't see a way to support multiple upgrade paths without creating more test clusters. If there's a generic solution for this that can be applied to all OpenSearch components, would you please provide? Thanks!

cliu123:
@peternied This PR is supposed to include only the version changes. Other changes in BWC tests if required should be in a separate PR. We can explore a solution to support multiple upgrade paths. It is not worthwhile to block a version bump for BWC test changes that are supposed to be looked into separately. IMO, things that haven't been well-defined shouldn't block another PR unless the PR being blocked has hard dependencies on that.
@VachaShah Do you happen to have answer for @peternied 's question above as you have the most context on the BWC test framwork? I don't see any way to support multiple upgrade paths without creating more test clusters, and I believe it would require more changes than that. If there's a generic solution for this that can be applied to all OpenSearch components, would you please provide? Thanks!

peternied (approved):
I've looked into what it will take to update the BWC tests and created an issue https://github.com/opensearch-project/OpenSearch/issues/3607 to get the proper support built in at a framework level.  Lets move forward with the change you have and until the framework supports this scenario we will need to operate outside the guidance.

DarshitChanpura (approved):

cliu123:
> I've looked into what it will take to update the BWC tests and created an issue [opensearch-project/OpenSearch#3607](https://github.com/opensearch-project/OpenSearch/issues/3607) to get the proper support built in at a framework level. Lets move forward with the change you have and until the framework supports this scenario we will need to operate outside the guidance.

Created a separate issue to track the pending item: https://github.com/opensearch-project/security/issues/1889

Issue #1917, opened 2022-06-27
Test reliability is impacting CI workflows


peternied:
There has been an increased number of test failures that have started to prevent continuous integration for pull requests.  It looks like this has been accelerated by the recent cut over to the Randomizer test runner.  While there has been progress with pull requests like https://github.com/opensearch-project/security/pull/1914, we should continue to invest in this space to keep our build and test runs passing consistently.

List of tests that have seen issues blocking a pull request:
* [x] TenantInfoActionTest.testTenantInfoAPIAccess
* [x] ComplianceAuditlogTest.testComplianceEnable
* [x] ComplianceAuditlogTest. testSourceFilterMsearch

List of tests that have seen issues failures, but passed on retry:
* [ ] TransportUserInjectorIntegTest.testSecurityUserInjection
* [ ] AllowlistApiTest.testAllowlistAuditComplianceLogging
* [ ] LegacyTenantInfoActionTests.testTenantInfoAPIUpdate

As we see more tests with issues or resolve test issue I'll update this list

DarshitChanpura:
[Triage] Thank you for filing this. @peternied Please update this issue with when you will be done with it.

peternied:
Another instance of `ComplianceAuditlogTest.testComplianceEnable` impacting a pull request https://github.com/opensearch-project/security/pull/1923, failures in https://github.com/opensearch-project/security/actions/runs/2579777726/attempts/1

peternied:
Another instance in https://github.com/opensearch-project/security/pull/1924, `TenantInfoActionTest.testTenantInfoAPIAccess` was the blocking test

peternied:
[Triage] @cwperks after you've fix the blocking test lets close out this issue.

PR #1920, opened 2022-06-28
testComplianceEnable supports variable number of audit messages

peternied:
### Description
We were seeing test failures where on higher end computers there would
be duplicate audit messages for the index mapping creation.  Then when
run inside GitHub Actions there would only be 2 messages.  This doesn't
look like an overt product issue, overlogging of requests, but the test
case was not handling it well. Also improved the failure message response
for faster future debugging.

### Issues Resolved
* Related #1917

### Check List


DarshitChanpura (approved):

DarshitChanpura:
@peternied Should this be backported to 2.x?

cliu123 (approved):

PR #1937, opened 2022-07-13
Create a manually started workflow for bulk run of integration tests

peternied:
### Description
We've been looking to increase test reliability and that has lead to many runs and re-runs of the integration test in GitHub actions.  Creating a manually started workflow that runs the integration tests on both jdk11 and jdk17 10x times.

### Issues Resolved
* Related #1917

### Testing
See an example of this run [here](https://github.com/peternied/security/actions/runs/2665976803).

### Check List


DEVELOPER_GUIDE.md
```diff
@@ -135,6 +137,15 @@ curl -XGET https://localhost:9200/_plugins/_security/authinfo -u 'admin:admin' -
 
 Launch IntelliJ IDEA, choose **Project from Existing Sources**, and select directory with Gradle build script (`build.gradle`).
 
+## Running integration tests
```
cwperks, line 140: Thanks for adding a section on how to find the test output after CI has run!

cwperks:
The changes look good to me. Looks like there's a minor lint issue with the `integration-tests.yml` file. Clever use of `matrix`!

DarshitChanpura (approved):
This is a great addition. Thank you @peternied !

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,40 @@
+name: Bulk Integration Test
+
+on: [
+  workflow_dispatch,
+  # Uncomment below to automatically run on pushes to branches
+  # push,
```
cliu123, line 6: Can this comment be removed? I don't think we will want to run tests for 10 time automatically on each push.

peternied: Added this as a hint in cases where folks want to quickly iterate automatically rather than using the GitHub UX if they are troubleshooting an issue. I would like to leave the comment in, what do you think?

peternied:
> ### Testing
> See an example of this run [here](https://github.com/peternied/security/actions/runs/2665976803).

@cliu123 Here is an execution of this PR on my branch, 20 out of 20 passed, but you can drill in to see failures that were not blocking like our normal workflow.

.github/workflows/integration-tests.yml
cliu123: I think a quick click on the button to trigger the workflow will be easy enough. One can easily trigger the workflow AS NEEDED by clicking the button. 
If one has to run this workflow on every push, that would mean that the test reliability is too low to continue development. We wouldn't want to see that happens.

DarshitChanpura (approved):

cliu123:
Where is the button triggering this workflow? I mean I don't see the button in the [Checks page](https://github.com/opensearch-project/security/pull/1937/checks). How was the example fun triggered?

PR #2046, opened 2022-08-23
Add `allowlist.yml` to where it is not being handled in securityadmin tool.

### LINKED ISSUE #2047: [BUG] Security admin does not handle allowlist.yml in all the places where it handles whitelist.yml
**What is the bug?**
Security admin tool should handle `allowlist.yml` in all the places where it handles `whitelist.yml`.

cliu123:

### Description
* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation) Bug fix

### Issues Resolved
#2047

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/tools/SecurityAdmin.java
+++ b/src/main/java/org/opensearch/security/tools/SecurityAdmin.java
@@ -736,6 +736,7 @@ public class SecurityAdmin {
                 final boolean populateFileIfEmpty = true;
                 success = retrieveFile(restHighLevelClient, cd+"nodes_dn_"+date+".yml", index, "nodesdn", legacy, populateFileIfEmpty) && success;
                 success = retrieveFile(restHighLevelClient, cd+"whitelist_"+date+".yml", index, "whitelist", legacy, populateFileIfEmpty) && success;
+                success = retrieveFile(restHighLevelClient, cd+"allowlist_"+date+".yml", index, "allowlist", legacy, populateFileIfEmpty) && success;
                 return (success?0:-1);
             }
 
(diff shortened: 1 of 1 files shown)


DarshitChanpura (approved):

src/main/java/org/opensearch/security/tools/SecurityAdmin.java
```diff
@@ -1218,6 +1220,9 @@ private static int upload(RestHighLevelClient tc, String index, String cd, boole
         if (new File(cd+"audit.yml").exists()) {
             success = uploadFile(tc, cd + "audit.yml", index, "audit", legacy, resolveEnvVars) && success;
         }
+        if (new File(cd+"allowlist.yml").exists()) {
```
cwperks, line 1223: Does this mean that `allowlist.yml` is optional, but `whitelist.yml` is required?

cliu123: Good question! `allowlist.yml` is an alternative to and will be a replacement of `whitelist.yml`, so it does not necessarily exist in a cluster.

cliu123: @cwperks Good question! `allowlist.yml` is an alternative to and will be a replacement of `whitelist.yml`, so it does not necessarily exist in a cluster.

cwperks: There's an existence check around `allowlist.yml`, but no check around `whitelist.yml` a few lines above.

```
success = uploadFile(tc, cd+"whitelist.yml", index, "whitelist", legacy, resolveEnvVars) && success;
 if (new File(cd+"allowlist.yml").exists()) {
    success = uploadFile(tc, cd + "allowlist.yml", index, "allowlist", legacy, resolveEnvVars) && success;
}
```

When I run securityadmin after removing `whitelist.yml`, but have a file called `allowlist.yml` I get the following error: 

```ERR: Seems ../../../config/opensearch-security/whitelist.yml is not in legacy format: java.io.FileNotFoundException: ../../../config/opensearch-security/whitelist.yml (No such file or directory)```

I was wondering if it made sense to check for the existence of either and fail if neither exists.

cliu123: `whitelist.yml` is required, so there's no need to check the existence.

cwperks (approved):

PR #2033, opened 2022-08-17
Add security changes for point in time API

bharath-techie:
Signed-off-by: Bharathwaj G <bharath78910@gmail.com>

### Description
The existing model requires indices read access to 'pit delete' and 'list all pits' to all users. 
So changing the action names to 'cluster:admin/<action_name>".

Now , we can't add these cluster permissions to 'cluster_composite_ops' or 'ops_ro' since user has access to 'my_index' role which has access to 'cluster_composite_ops' which will make all users to access list and delete pit.

So changing the approach to new one below.
Add point in time permissions to 'manage_point_in_time' which is part of default static action groups.

Point in time apis - and associated action names :

1. Create PIT -     - "indices:data/read/point_in_time/create" ( indices:data/read is chosen because this will make sure users have read permission to the passed indices )
2. Delete PIT - "cluster:admin/point_in_time/delete" ( cluster:admin/*  is used because the apis are not dependent on indices like create pit and should just work based on this standalone permission )
3. List all PITs - "cluster:admin/point_in_time/read*"
4. PIT segments API - "indices:monitor/point_in_time/segments"

All the above actions are added to new action group 'manage_point_in_time'

* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation)
 New feature - Point in time feature
* Why these changes are required?
 These changes are required for access of various point in time APIs
* What is the old behavior before changes and new behavior after changes?
This is a new feature

Design document:
https://github.com/opensearch-project/OpenSearch/issues/3960

Api changes:
https://github.com/opensearch-project/OpenSearch/pull/4064 - create pit and delete pit api
https://github.com/opensearch-project/OpenSearch/pull/4016 - list all

### Issues Resolved
https://github.com/opensearch-project/OpenSearch/issues/3959

Is this a backport? If so, please add backport PR # and/or commits #

### Testing
Tested locally by running opensearch server alongside security plugin.
Only when index permissions and 'manage_point_in_time' action group permission is present , create api succeeds
For rest of the APIs, 'manage_point_in_time' action group permission controls whether the api can be accessible or not, which has been tested as well.

### Check List
- [ ] New functionality includes testing
- [ ] New functionality has been documented
- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
[...]

PR #2053, opened 2022-08-28
Changes to make PIT security model granular

bharath-techie:
Signed-off-by: Bharathwaj G <bharath78910@gmail.com>

### Description
As part of point in time opensearch review, we got comments which will help PIT security model more granular - where each PIT can be permitted or rejected based on user's indices permissions.
https://github.com/opensearch-project/OpenSearch/pull/4064#discussion_r952746806

This PR contains the changes for same.
As part of this change ,

- All PIT APIs which are user facing will be using 'indices:data/read' action that is inline with search semantics
- For 'List all PITs' API, we'll return only the PITs for which the user has access to.
   - If user has access to **all indices part of the PIT**, then the PIT will be permitted
   - If user has no access to _any index_  of PIT, then we'll not give access to that PIT
   - If user has no access to _any PITs_, then we'll throw permission error for the API request
- For 'Delete list of PITs' API
   - User needs to have access for all indices of all PITs, only then we'll allow deletion otherwise we'll fail the operation. ( This is inline with other APIs where explicit values are given instead of 'all' or '*' )
 - For 'Delete all PITs' API
    -  If user has access to **all indices part of the PIT**, then the PIT will be permitted and deleted
    - If user has no access to any index of PIT, then we'll not delete the PIT
    - If user has no access to _any PITs_, then we'll throw permission error for the API request

New feature - Modification to  point in time security model - It is a new feature and not released.
We're making these changes to enhance the security model of PIT as per review comments from opensearch community
https://github.com/opensearch-project/security/pull/2033 -- this is the old behavior and the above description gives an idea of new behavior. Basically old model gives user access to all PIT contexts or none of PIT contexts and the new model allows permission for PITs based on the underlying indices.

### Issues Resolved
https://github.com/opensearch-project/OpenSearch/issues/3960

2.3 / 2.x backport

### Testing
Tested all the changes with opensearch + security plugin locally. All changes are working as expected.

### Check List


bharath-techie:
@reta @Bukhtawar please review

bharath-techie:
Build is failing because all the dependencies are opensearch main branch ( 3.0.0 )

bharath-techie:
@peternied @cliu123 please review this PR.

src/main/resources/static_config/static_action_groups.yml
```diff
@@ -233,8 +233,10 @@ manage_point_in_time:
   static: true
   allowed_actions:
     - "indices:data/read/point_in_time/create"
-    - "cluster:admin/point_in_time/delete"
-    - "cluster:admin/point_in_time/read*"
+    - "indices:data/read/point_in_time/delete"
+    - "indices:data/read/point_in_time/readall"
+    - "indices:data/read/search"
+    - "cluster:admin/point_in_time/read_from_nodes"
```
Bukhtawar, line 239: Sorry, for my understanding what is `read_from_nodes`?

bharath-techie: This is the parent action that basically reads all active Point in time searches from nodes. 

This the action for NodesGetAllPitAction.

We can rename it as well, if the name is confusing.

src/main/java/org/opensearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -370,11 +370,11 @@ public final static class Resolved {
...
-        private Resolved(final ImmutableSet<String> aliases,
-                         final ImmutableSet<String> allIndices,
-                         final ImmutableSet<String> originalRequested,
-                         final ImmutableSet<String> remoteIndices,
-                         IndicesOptions indicesOptions) {
+        public Resolved(final ImmutableSet<String> aliases,
+                        final ImmutableSet<String> allIndices,
+                        final ImmutableSet<String> originalRequested,
+                        final ImmutableSet<String> remoteIndices,
+                        IndicesOptions indicesOptions) {
```
cliu123, line 377: Please avoid the format changes if not necessary.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
```diff
@@ -0,0 +1,344 @@
...
+import org.opensearch.common.unit.TimeValue;
+import org.opensearch.security.OpenSearchSecurityPlugin;
+import org.opensearch.security.resolver.IndexResolverReplacer;
+import org.opensearch.security.securityconf.SecurityRoles;
+import org.opensearch.security.user.User;
+
+/**
+ * This class evaluates privileges for point in time (Delete and List all) operations
+ */
+public class PitAccessEvaluator {
```
cliu123, line 64: nit: rename to `PitPrivilegeEvaluator`?

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
```diff
@@ -0,0 +1,344 @@
...
+
+    public PrivilegesEvaluatorResponse evaluate(final ActionRequest request, final ClusterService clusterService,
+                                                final User user, final SecurityRoles securityRoles, final String action,
+                                                final IndexNameExpressionResolver resolver,
+                                                final PrivilegesEvaluatorResponse presponse) {
+
+        // Skip pit evaluation for "NodesGetAllPITs" action, since it fetches all PITs across the cluster
+        // for privilege evaluation
+        if(action.startsWith("cluster:admin")) {
+            return presponse;
```
cliu123, line 77: How is the permission for `NodesGetAllPITs` evaluated if requests bypass privilege evaluator? Or is there no need to have access control on `NodesGetAllPITs`?

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
```diff
@@ -0,0 +1,344 @@
...
+                return handleGetAllPitsAccess(request, clusterService, user, securityRoles,
+                        action, resolver, presponse);
+            } else if (request instanceof DeletePitRequest) {
+                DeletePitRequest deletePitRequest = (DeletePitRequest) request;
+                List<String> pitIds = deletePitRequest.getPitIds();
+                if (pitIds.size() == 1 && "_all".equals(pitIds.get(0))) {
+                    return handleDeleteAllPitAccess(deletePitRequest, clusterService, user, securityRoles,
+                            action, resolver, presponse);
+                } else {
+                    return handleExplicitPitsAccess(deletePitRequest.getPitIds(), clusterService, user, securityRoles,
```
cliu123, line 93: Can `deletePitRequest.getPitIds()` be replaced with `pitIds` here to avoid unnecessary performance overhead?

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
```diff
@@ -0,0 +1,344 @@
...
+                            action, resolver, presponse);
+                }
+            } else if (request instanceof PitSegmentsRequest) {
+                PitSegmentsRequest pitSegmentsRequest = (PitSegmentsRequest) request;
+                List<String> pitIds = pitSegmentsRequest.getPitIds();
+                if (pitIds.size() == 1 && "_all".equals(pitIds.get(0))) {
+                    return handleGetAllPitSegmentsAccess(pitSegmentsRequest, clusterService, user, securityRoles,
+                            action, resolver, presponse);
+                } else {
+                    return handleExplicitPitsAccess(pitSegmentsRequest.getPitIds(), clusterService, user, securityRoles,
```
cliu123, line 103: Can `deletePitRequest.getPitIds()` be replaced with `pitIds` here to avoid unnecessary performance overhead?

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
bharath-techie: It'll still get evaluated, we are just not evaluating it as part of pit access evaluator.

We're not marking presponse as complete, so it'll continue down the privilegeEvaluator code flow.

In short, User needs the action permission to execute nodes get all pits.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
```diff
@@ -0,0 +1,344 @@
...
+                            action, resolver, presponse);
+                }
+            }
+        } catch(InterruptedException e) {
```
cliu123, line 107: Could there be any other types of exceptions other than `InterruptedException` here?

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
```diff
@@ -0,0 +1,344 @@
...
+
+    /**
+     * Handle access for Get All PITs access
+     */
+    private PrivilegesEvaluatorResponse handleGetAllPitsAccess(final ActionRequest request, final ClusterService clusterService,
+                                        final User user, SecurityRoles securityRoles, final String action,
+                                        IndexNameExpressionResolver resolver,
+                                        PrivilegesEvaluatorResponse presponse) throws InterruptedException {
+        List<ListPitInfo> pitInfos = getAllPitInfos((GetAllPitNodesRequest) request);
+        // if cluster has no PITs, then allow the operation to pass with empty response
```
cliu123, line 122: The authorization on a request should not be decided by whether there are PITs in a cluster. It should always be evaluated with permissions instead. If the user does not have right permissions, the request should be denied no matter whether there are PITs in a cluster or how many PITs there are.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
```diff
@@ -0,0 +1,344 @@
...
+        if(pitInfos.isEmpty()) {
+            presponse.allowed = true;
```
cliu123, line 124: The authorization on a request should not be decided by whether there are PITs in a cluster. It should always be evaluated with permissions instead. If the user does not have right permissions, the request should be denied no matter whether there are PITs in a cluster or how many PITs there are.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
```diff
@@ -0,0 +1,344 @@
...
+                log.debug("Evaluating PIT ID : " + pitId );
+            }
+            if (allPermittedIndices.size() == INDICES_SET.size()) {
+                if(isDebugEnabled) {
+                    log.debug(" Permitting PIT ID : " + pitId);
+                }
+                permittedPits.add(pitToPitInfoMap.get(pitId));
+            }
+        }
+        if (permittedPits.size() > 0) {
```
cliu123, line 157: Is there a case where `permittedPits.size()` equals to `0` as there is no PITs in the cluster, but the request should still be authorized?

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
```diff
@@ -0,0 +1,344 @@
...
+    private PrivilegesEvaluatorResponse handleDeleteAllPitAccess(DeletePitRequest deletePitRequest, ClusterService clusterService,
+                                          User user, SecurityRoles securityRoles, final String action,
+                                          IndexNameExpressionResolver resolver,
+                                          PrivilegesEvaluatorResponse presponse) throws InterruptedException {
+        List<String> permittedPits = new ArrayList<>();
+        List<String> pitIds = getAllPitIds();
+        // allow delete pit operation if there are no pits in the cluster ( response should be empty )
+        if(pitIds.isEmpty()) {
+            deletePitRequest.clearAndSetPitIds(pitIds);
+            presponse.allowed = true;
```
cliu123, line 178: The authorization on a request should not be decided by whether there are PITs in a cluster. It should always be evaluated with permissions instead. If the user does not have right permissions, the request should be denied no matter whether there are PITs in a cluster or how many PITs there are.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
bharath-techie: If the user does not have permissions, we do throw exception. I'll add some ITs which can assert the same.

Now this logic makes sure that we don't throw permission exception when there are simply no PITs created in the cluster.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
```diff
@@ -0,0 +1,344 @@
...
+            }
+            if (allPermittedIndices.size() == indicesSet.size()) {
+                if(isDebugEnabled) {
+                    log.debug(" Permitting PIT : " + pitId);
+                }
+                permittedPits.add(pitId);
+            }
+        }
+        // If there are any PITs for which the user has access to, then allow operation otherwise fail.
+        if(permittedPits.size() > 0) {
```
cliu123, line 199: When a user has right permissions, even if permittedPits.size() equals to 0, the request also should be authorized and get empty in the response.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
```diff
@@ -0,0 +1,344 @@
...
+    private PrivilegesEvaluatorResponse handleGetAllPitSegmentsAccess(PitSegmentsRequest pitSegmentsRequest, ClusterService clusterService,
+                                                                 User user, SecurityRoles securityRoles, final String action,
+                                                                 IndexNameExpressionResolver resolver,
+                                                                 PrivilegesEvaluatorResponse presponse) throws InterruptedException {
+        List<String> permittedPits = new ArrayList<>();
+        List<String> pitIds = getAllPitIds();
+        // allow pit segments operation if there are no pits in the cluster ( response should be empty )
+        if(pitIds.isEmpty()) {
+            pitSegmentsRequest.clearAndSetPitIds(pitIds);
+            presponse.allowed = true;
```
cliu123, line 219: The authorization on a request should not be decided by whether there are PITs in a cluster. It should always be evaluated with permissions instead. If the user does not have right permissions, the request should be denied no matter whether there are PITs in a cluster or how many PITs there are.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
```diff
@@ -0,0 +1,344 @@
...
+            }
+            if (allPermittedIndices.size() == indicesSet.size()) {
+                if(isDebugEnabled) {
+                    log.debug(" Permitting PIT : " + pitId);
+                }
+                permittedPits.add(pitId);
+            }
+        }
+        // If there are any PITs for which the user has access to, then allow operation otherwise fail.
+        if(permittedPits.size() > 0) {
```
cliu123, line 240: When a user has right permissions, even if permittedPits.size() equals to 0, the request also should be authorized and get empty in the response. There could be 0 PITs in the cluster.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
bharath-techie: That is the case we've handled in the line 124, we authorize the user. Because the whole custom logic just ensures if PIT IDs are present, the user should have associated indices permissions.

If not, at this point the user permissions are already evaluated and user does have action permissions, so we mark operation as success.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
bharath-techie: Same comment as above, we do evaluation of permissions, if user does not have permission, we'll not reach to this custom code.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
cliu123: >Now this logic makes sure that we don't throw permission exception when there are simply no PITs created in the cluster.

If a user does not have PIT search permissions, even if there are not PITs in the cluster, the request still should be denied. As I said, authorizing a request or not should only be decided by the permissions the user has(privilege evaluator).

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
bharath-techie: This is for the case where there are no PITs in cluster. We do evaluate the permissions.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
bharath-techie: Line 217 handles it.

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
bharath-techie: Other exceptions we handle it in onfailure

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
bharath-techie: @cliu123 i tried making this work, but looks like for index actions, there is no way to authorize the user when there are no indices to validate against. 

dnfOfEnabled flag is used for indices actions where for empty we permit , otherwise we throw forbidden error.

can you please suggest if there are any alternate approaches ?

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
bharath-techie: Ignore this comment, can we use the flag dnfofenabled config property similar to other permissions.

src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
```diff
@@ -0,0 +1,356 @@
...
+            @Override
+            public void onFailure(Exception e) {
+                throw new OpenSearchException("List all PITs failed", e);
+            }
+        };
+        LatchedActionListener latchedActionListener = new LatchedActionListener<>(listener, latch);
+
+        OpenSearchSecurityPlugin.GuiceHolder.getPitService().getAllPits(latchedActionListener);
+
+        if(!latch.await(15, TimeUnit.SECONDS)) {
```
Bukhtawar, line 321: @reta as per this [discussion](https://github.com/opensearch-project/OpenSearch/pull/4064#discussion_r963063762) we need to get rid of this latch since this has the potential to block http request handler threads. 

Do you have thoughts on how we could have a reference to the response listener or perform non-blocking waits,

reta: Sorry @Bukhtawar , I came a bit late, seems like this code is already gone :partying_face:

Bukhtawar: @bharath-techie would you be able to shed some light if you still have that commit handy

@reta apologies let's see if we that commit

src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
```diff
@@ -0,0 +1,193 @@
...
+        for(ListPitInfo pitInfo : pitInfos) {
+            pitToPitInfoMap.put(pitInfo.getPitId(), pitInfo);
+        }
+        List<ListPitInfo> permittedPits = new ArrayList<>();
+        for (String pitId : pitIds) {
+            String[] indices = pitToIndicesMap.get(pitId);
+            HashSet<String> indicesSet = new HashSet<>(Arrays.asList(indices));
+
+            final ImmutableSet<String> INDICES_SET = ImmutableSet.copyOf(indicesSet);
+            final IndexResolverReplacer.Resolved pitResolved =
```
peternied, line 130: Can you reuse the private function for this?

bharath-techie:
@elfisher @cliu123 please prioritize reviewing this PR as this is a 2.3 release blocker.

src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
```diff
@@ -0,0 +1,193 @@
...
+    private boolean isDebugEnabled = log.isDebugEnabled();
+
+    public PrivilegesEvaluatorResponse evaluate(final ActionRequest request, final ClusterService clusterService,
+                                                final User user, final SecurityRoles securityRoles, final String action,
+                                                final IndexNameExpressionResolver resolver,
+                                                boolean dnfOfEmptyResultsEnabled, final PrivilegesEvaluatorResponse presponse) {
+
+        // Skip pit evaluation for "NodesGetAllPITs" action, since it fetches all PITs across the cluster
+        // for privilege evaluation
+        if(action.startsWith("cluster:admin")) {
```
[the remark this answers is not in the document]
bharath-techie: no there are two actions with getAllPitNodesRequest, we skip it for the parent action which basically gets active PITs cluster wide.

src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
```diff
@@ -0,0 +1,193 @@
...
+                log.debug("Evaluating PIT ID : " + pitId );
+            }
+            if (allPermittedIndices.size() == INDICES_SET.size()) {
+                if(isDebugEnabled) {
+                    log.debug(" Permitting PIT ID : " + pitId);
+                }
+                permittedPits.add(pitToPitInfoMap.get(pitId));
+            }
+        }
+        if (permittedPits.size() > 0) {
```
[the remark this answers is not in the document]
bharath-techie: We are making this whole change because we want the pit permission to be granular - only the allowed PITs are shown to the user ( by setting the permitted pits in request )

src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
```diff
@@ -0,0 +1,193 @@
...
+                                                final IndexNameExpressionResolver resolver,
+                                                boolean dnfOfEmptyResultsEnabled, final PrivilegesEvaluatorResponse presponse) {
+
+        // Skip pit evaluation for "NodesGetAllPITs" action, since it fetches all PITs across the cluster
+        // for privilege evaluation
+        if(action.startsWith("cluster:admin")) {
+            return presponse;
+        }
+        try {
+            if (request instanceof GetAllPitNodesRequest) {
```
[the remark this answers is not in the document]
bharath-techie: It does work for some requests like delete and pit segments , but for get all pits, we want the behavior to be different.

In the sense , for 'all' PITs, we want to choose and permit the PITs for which the user has access to. ( out of all PITs in cluster )

For 'explicit' PITs, we want user to have access to all PITs or we deny the operation.

src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
```diff
@@ -0,0 +1,193 @@
...
+        List<String> pitIds = new ArrayList<>();
+        pitIds.addAll(pitInfos.stream().map(ListPitInfo::getPitId).collect(Collectors.toList()));
+        Map<String, String[]> pitToIndicesMap = OpenSearchSecurityPlugin.GuiceHolder.getPitService().getIndicesForPits(pitIds);
+        Map<String, ListPitInfo> pitToPitInfoMap =  new HashMap<>();
+
+        for(ListPitInfo pitInfo : pitInfos) {
+            pitToPitInfoMap.put(pitInfo.getPitId(), pitInfo);
+        }
+        List<ListPitInfo> permittedPits = new ArrayList<>();
+        for (String pitId : pitIds) {
```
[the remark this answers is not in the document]
bharath-techie: done.

src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
```diff
@@ -0,0 +1,193 @@
...
+        return presponse;
+    }
+
+    /**
+     * Handle access for Get All PITs access
+     */
+    private PrivilegesEvaluatorResponse handleGetAllPitsAccess(final ActionRequest request, final ClusterService clusterService,
+                                                               final User user, SecurityRoles securityRoles, final String action,
+                                                               IndexNameExpressionResolver resolver,
+                                                               boolean dnfOfEmptyResultsEnabled, PrivilegesEvaluatorResponse presponse) throws InterruptedException {
```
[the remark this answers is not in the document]
bharath-techie: this is outdated, no such exceptions are thrown in current code ( earlier we were using latch and we removed the entire logic because of such exceptions + thread blockage )

src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
```diff
@@ -0,0 +1,193 @@
...
+
+    protected final Logger log = LogManager.getLogger(this.getClass());
+    private boolean isDebugEnabled = log.isDebugEnabled();
+
+    public PrivilegesEvaluatorResponse evaluate(final ActionRequest request, final ClusterService clusterService,
+                                                final User user, final SecurityRoles securityRoles, final String action,
+                                                final IndexNameExpressionResolver resolver,
+                                                boolean dnfOfEmptyResultsEnabled, final PrivilegesEvaluatorResponse presponse) {
+
+        // Skip pit evaluation for "NodesGetAllPITs" action, since it fetches all PITs across the cluster
```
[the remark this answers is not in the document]
bharath-techie: immediately below

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
cliu123: @bharath-techie looks like the permission for `NodesGetAllPITs` is evaluated in [PitPrivilegesEvaluator.java](https://github.com/opensearch-project/security/pull/2053/files#diff-eb0f476c4343f7d278cdcc617b1de506134909cdccb93baf4effd7c914d0b043)? Would you please point me to where it gets evaluated and explain a bit more about how it gets evaluated?

cliu123: @bharath-techie looks like the permission for `NodesGetAllPITs` is designed to be evaluated in [PitPrivilegesEvaluator.java](https://github.com/opensearch-project/security/pull/2053/files#diff-eb0f476c4343f7d278cdcc617b1de506134909cdccb93baf4effd7c914d0b043)? Would you please point me to where it gets evaluated and explain a bit more about how it gets evaluated?

bharath-techie: As long as we don't mark the presponse as complete, it'll continue down the flow in `PrivilegesEvaluator`

```
 // check access for point in time requests
        if(pitAccessEvaluator.evaluate(request, clusterService, user, securityRoles,
                action0, resolver, dcm.isDnfofForEmptyResultsEnabled(), presponse).isComplete()) {
            return presponse;
        }
```

It'll go inside this if block for evaluation after we return without marking it complete https://github.com/opensearch-project/security/pull/2053/files#diff-e746c76b234a31b6f95d428b7c40dabf805b609b110c560902c967db8f83fa7aR303

elfisher:
Hi I want to confirm this will work with relevant [predefined roles](https://opensearch.org/docs/latest/security-plugin/access-control/users-roles/#predefined-roles). This should at minimum work with `all_access`, `read_all`, and `read_all_and_monitor`. Thanks!

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
bharath-techie: Addressed the comment, we're denying if there are no pits so resolving this and similar ones

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
peternied: Here is a potential alternative approach:
1) Collect authorization information for the user
2) Does the user have access to any PITs? (If no- permission is denied)
3) Does the user have access to the PITs in the request? (If no- permission is denied)

src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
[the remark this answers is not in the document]
peternied: Resolving this thread, the behavior of 'all' is still questionable but that is captured in other comments

peternied (changes requested):
Thanks for the updates, there are still many unresolved threads please take a look.

One way we could see about reviewing/merging this change faster, I think there is less discussion around the workflows that intersect with `handleExplicitPitsAccess`.  Consider splitting this PR into multiple parts to so we can make progress independently

src/test/resources/roles.yml
```diff
@@ -1128,7 +1128,35 @@ data_stream_3:
         - "*"
       allowed_actions:
         - "DATASTREAM_ALL"
+point_in_time_1:
+  reserved: true
+  hidden: false
+  description: "Migrated from v6 (all types mapped)"
+  cluster_permissions:
+    - "cluster:admin/point_in_time/read_from_nodes"
```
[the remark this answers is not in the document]
bharath-techie: Without cluster level permission we'd not be able to get all PITs from the cluster, as the request doesn't have indices as part of it. Any suggestions ?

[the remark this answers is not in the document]
peternied: Is there a place where all of the different Rest APIs are documented in the form below?  I would like to be able to cross-reference the actions and associated permissions that are needed to perform them, I think that might help our understanding.

#### Create point in time
```
POST "/<target indices>/point_in_time?*keep_alive*=1h&routing=&expand_wildcards=&preference=" 
```

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
bharath-techie: so as long as the user does not have PITs , we throw 403. The behavior is consistent and the above does not happen.

User might not have PITs because :
- PITs are present in the cluster but the user does not have access
- PITs are not present in the cluster

src/main/java/org/opensearch/security/privileges/PitAccessEvaluator.java
bharath-techie: That is what we are doing currently when the config is false ( which is default ).
This config is used in search and other index actions as well, so can we please take a look at this config? it is designed not to throw forbidden in case of empty results.

src/test/resources/roles.yml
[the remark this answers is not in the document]
bharath-techie: https://github.com/opensearch-project/OpenSearch/issues/3960 -- is the design doc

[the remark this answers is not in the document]
peternied: The doc and the comments on this PR have evolved over time and it is not clear to me what the permission setup, user action, and result should be for the many scenarios embodied - please make this as clear as possible.

Not only would this help us understand the specific scope and bounds of your request, this would be of great value for writing the end-user facing permissions documentation for this feature

Issue #2050, opened 2022-08-24
Remaining areas to replace legacy references of opendistro and move to inclusive language


cwperks:
There have been a few recent updates to the Documentation Website to update documentation in areas that were missed from the transition away from legacy `opendistro` references or to the use of inclusive language. 

Here are three examples:

- [ ] [Fix setting name for LDAP custom attribute filter](https://github.com/opensearch-project/documentation-website/pull/912)
- [ ] [Update documentation from opensearch_security_roles to opendistro_security_roles](https://github.com/opensearch-project/documentation-website/pull/892)
- [ ] [Change /_plugins/_security/.. routes to /_opendistro/_security/... for SAML until plugins route is supported](https://github.com/opensearch-project/documentation-website/pull/877)

I'm opening an issue to keep track of any instances that come up to keep track of the remaining areas that need to be updated.

peternied:
@cwperks Thanks for filing this, could we separate out the individual actions that we need to take, maybe create issues for the legacy API paths, update legacy settings, or change documentation around the roles.  This way we can close them as they are completed.

cwperks:
@peternied 3 new issues have been created to track each issue independently.

- https://github.com/opensearch-project/security/issues/2059
- https://github.com/opensearch-project/security/issues/2058
- https://github.com/opensearch-project/security/issues/2060

peternied:
@cwperks Thanks for filing these issues, we will close them as we work these issues.

PR #2056, opened 2022-08-29
Fix the inconsistent variable for DLS

RyanL1997:
Signed-off-by: Ryan Liang <jiallian@amazon.com>

### Description
Fixed the inconsistent variable with name "`documentWhitelist`" and changed it into "`documentAllowlist`" with all its references in `DlsFilterLevelActionHandler.java`.

### Issues Resolved
Resolve https://github.com/opensearch-project/security/issues/1665

### Check List
- [ ] ~New functionality includes testing~
- [ ] ~New functionality has been documented~
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2062, opened 2022-08-30
Switch to ClusterManager terminology

peternied:
### Description

Enforce usage of ClusterManager terminology in the codebase
Include mechanism to disable checkstyle rule

Signed-off-by: Peter Nied <petern@amazon.com>

### Issues Resolved
* Related to https://github.com/opensearch-project/security/issues/2050

### Testing
Added Checkstyle mechanism to prevent usage of non-inclusive terminology

### Check List

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -157,7 +157,7 @@ Checkstyle enforced several rules within this codebase.  Sometimes exceptions wi
 
 *Example violation*
 ```
-[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java:178: Usage should be switched to cluster manager [RegexpSingleline]
+[ant:checkstyle] [ERROR] /local/home/security/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java:178: Usage should be switched to cluster manager [RegexpSingleline]
 ```
 
 *Single Rule Suppression*
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -8,6 +8,7 @@ So you want to contribute code to this project? Excellent! We're glad you're her
   - [Using IntelliJ IDEA](#using-intellij-idea)
   - [Running integration tests](#running-integration-tests)
     - [Bulk test runs](#bulk-test-runs)
+    - [Checkstyle Violations](#checkstyle-violations)
   - [Submitting Changes](#submitting-changes)
   - [Backports](#backports)
 
@@ -145,6 +146,35 @@ Integration tests are automatically run on all pull requests for all supported v
 
 ### Bulk test runs
 To collect reliability data on test runs there is a manual GitHub action workflow called `Bulk Integration Test`.  The workflow is started for a branch on this project or in a fork by going to [GitHub action workflows](https://github.com/opensearch-project/security/actions/workflows/integration-tests.yml) and selecting `Run Workflow`.
+
+### Checkstyle Violations
+Checkstyle enforced several rules within this codebase.  Sometimes exceptions will be necessary for components that are set for deprecation but the new version is unavailable.  There are two formats of suppression that can be used when dealing with violations of this nature, one for disabling a single rule, or another for disabling all rules - its best to be as specific as possible.
+
+*Execute Checkstyle*
+```
+./gradlew checkstyleMain checkstyleTest 
+```
+
+*Example violation*
+```
+[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java:178: Usage should be switched to cluster manager [RegexpSingleline]
+```
+
+*Single Rule Suppression*
+```
+    // CS-SUPPRESS-SINGLE: RegexpSingleline See http://github/issues/1234
+    ...
+    Code that violates the rule
+    ...
+    // CS-ENFORCE-SINGLE
+```
+
+*Suppression All Checkstyle Rules*
+```
+  // CS-SUPRESS-ALL: Legacy code to be deleted in Z.Y.X see http://github/issues/1234
+  ...
+  // CS-ENFORCE-ALL
+```
 
 ## Submitting Changes
 
--- a/checkstyle/sun_checks.xml
+++ b/checkstyle/sun_checks.xml
@@ -201,7 +201,18 @@
     <property name="format" value="master"/>
     <property name="ignoreCase" value="true"/>
     <property name="message" value="Usage should be switched to cluster manager" />
-    <property name="severity" value="ignore"/>
+    <property name="severity" value="error"/>
+  </module>
+
+  <module name="SuppressWithPlainTextCommentFilter">
+    <property name="offCommentFormat" value="CS-SUPPRESS-ALL: .+"/> <!-- Require an explaination after surpressing -->
+    <property name="onCommentFormat" value="CS-ENFORCE-ALL"/>
+  </module>
+
+  <module name="SuppressWithPlainTextCommentFilter">
+    <property name="offCommentFormat" value="CS-SUPPRESS-SINGLE\: ([\w\|]+) .+"/> <!-- Require an explaination after surpressing -->
+    <property name="onCommentFormat" value="CS-ENFORCE-SINGLE"/>
+    <property name="checkFormat" value="$1"/>
   </module>
 
 </module>
--- a/src/test/java/org/opensearch/security/ccstest/CrossClusterSearchTests.java
+++ b/src/test/java/org/opensearch/security/ccstest/CrossClusterSearchTests.java
@@ -84,7 +84,9 @@ public class CrossClusterSearchTests extends AbstractSecurityUnitTest {
         }
 
         public ClusterTransportClientSettings(Settings clusterSettings, Settings transportSettings) {
-            super(clusterSettings, transportSettings);
+            super(Settings.builder()
+                .put(clusterSettings)
+                .putList("node.roles", "remote_cluster_client").build(), transportSettings);
         }
 
         public Settings clusterSettings() {
@@ -1001,7 +1003,8 @@ public class CrossClusterSearchTests extends AbstractSecurityUnitTest {
                     .source("{\"cluster\": \""+cl2Info.clustername+"\"}", XContentType.JSON)).actionGet();
         }
 
-        final Settings tcSettings = AbstractSecurityUnitTest.nodeRolesSettings(Settings.builder(), false, false) 
+        final Settings.Builder clusterClientSettings = Settings.builder().putList("node.roles", "remote_cluster_client");
+        final Settings tcSettings = AbstractSecurityUnitTest.nodeRolesSettings(clusterClientSettings, false, false) 
                 .put(minimumSecuritySettings(Settings.EMPTY).get(0))
                 .put("cluster.name", cl1Info.clustername)
                 .put("path.data", "./target/data/" + cl1Info.clustername + "/cert/data")
--- a/src/test/java/org/opensearch/security/ccstest/CrossClusterSearchTests.java
+++ b/src/test/java/org/opensearch/security/ccstest/CrossClusterSearchTests.java
@@ -1001,12 +1001,9 @@ public class CrossClusterSearchTests extends AbstractSecurityUnitTest {
                     .source("{\"cluster\": \""+cl2Info.clustername+"\"}", XContentType.JSON)).actionGet();
         }
 
-        final Settings tcSettings = Settings.builder()
+        final Settings tcSettings = AbstractSecurityUnitTest.nodeRolesSettings(Settings.builder(), false, false) 
                 .put(minimumSecuritySettings(Settings.EMPTY).get(0))
                 .put("cluster.name", cl1Info.clustername)
-                .put("node.data", false)
-                .put("node.master", false)
-                .put("node.ingest", false)
                 .put("path.data", "./target/data/" + cl1Info.clustername + "/cert/data")
                 .put("path.logs", "./target/data/" + cl1Info.clustername + "/cert/logs")
                 .put("path.home", "./target")
--- a/src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
+++ b/src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
@@ -43,6 +43,7 @@ import com.carrotsearch.randomizedtesting.RandomizedTest;
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope.Scope;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import io.netty.handler.ssl.OpenSsl;
 import org.apache.http.Header;
 import org.apache.http.HttpHost;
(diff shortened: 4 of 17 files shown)


peternied:
@RyanL1997 This change is similar to a recent PR https://github.com/opensearch-project/security/pull/2056 that you made, mind taking a look?

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
```diff
@@ -194,6 +195,17 @@ public void waitForInit(Client client) {
         }
     }
 
+    public static Settings.Builder nodeRolesSettings(final Settings.Builder settingsBuilder, final boolean isClusterManager, final boolean isDataNode) {
+        final ImmutableList.Builder<String> nodeRolesBuilder = ImmutableList.<String>builder();
+        if (isDataNode) {
+            nodeRolesBuilder.add("data");
+        }
+        if (isClusterManager) {
+            nodeRolesBuilder.add("cluster_manager");
```
cwperks, line 204: Use `DiscoveryNodeRole.CLUSTER_MANAGER_ROLE.roleName()`?

DEVELOPER_GUIDE.md
```diff
@@ -146,6 +147,35 @@ Integration tests are automatically run on all pull requests for all supported v
 ### Bulk test runs
 To collect reliability data on test runs there is a manual GitHub action workflow called `Bulk Integration Test`.  The workflow is started for a branch on this project or in a fork by going to [GitHub action workflows](https://github.com/opensearch-project/security/actions/workflows/integration-tests.yml) and selecting `Run Workflow`.
 
+### Checkstyle Violations
```
cwperks, line 150: Thanks for adding this section to the README showing how to run the gradle command.

checkstyle/sun_checks.xml
```diff
@@ -201,7 +201,18 @@
     <property name="format" value="master"/>
     <property name="ignoreCase" value="true"/>
     <property name="message" value="Usage should be switched to cluster manager" />
-    <property name="severity" value="ignore"/>
+    <property name="severity" value="error"/>
+  </module>
+
+  <module name="SuppressWithPlainTextCommentFilter">
+    <property name="offCommentFormat" value="CS-SUPPRESS-ALL: .+"/> <!-- Require an explaination after surpressing -->
```
cwperks, line 208: Why do this regex and the one below not match?

Should these both be `([\w\|]+) .+`?

cwperks (approved):
Changes look good to me 🚀

DarshitChanpura (dismissed):

peternied:
Jumping on here, it looks like `node.roles` has slightly different behavoir when cluster startup is happening, there is a bug that I need to fix.

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
peternied: There is no constant inside of OpenSearch for the config key which is the most obscure of these values, `node.roles`.  So I'm throwing my hands in the air?  IMO this is a good topic to revisit if we clean up the test construction workflow

checkstyle/sun_checks.xml
peternied: `CS-SUPPRESS-SINGLE` - suppresses a single named checkstyle rule, then afterward a reason should still be included.   This is useful for disabling a single unavoidable violation, such as where we have upstream libraries with non-inclusive terms in a class name, see the usage in MigrateApiAction.java.  This keeps rules enforced on all other checkstyle checks, such as AvoidStarImport.

`CS-SUPPRESS-ALL` - suppresses _all_ checkstyle rules.  This would be more for a scenario where we need to import a file from another codebase that has entirely different rules, and applying our own rules would be onerous or make merging future changes a nightmare.

#### Example usage where `RegexpSingleline` is suppressed and other checks are still enforced.
https://github.com/opensearch-project/security/blob/f1e804489c82c60442b84dee34e8fa603a4cb636/src/main/java/org/opensearch/security/dlic/rest/api/MigrateApiAction.java#L14-L73

checkstyle/sun_checks.xml
```diff
@@ -201,7 +201,18 @@
...
+    <property name="onCommentFormat" value="CS-ENFORCE-ALL"/>
+  </module>
+
+  <module name="SuppressWithPlainTextCommentFilter">
+    <property name="offCommentFormat" value="CS-SUPPRESS-SINGLE\: ([\w\|]+) .+"/> <!-- Require an explaination after surpressing -->
```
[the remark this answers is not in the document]
peternied: On line 215 `<property name="checkFormat" value="$1"/>` is taking the result of the regex expression group 1 and using it to determine which checks are enabled or not.

#### See how the regex is interpreted, the parenthesis creates the first match group that is accessible via `$1`
image
https://regexr.com/ <- great tool for writing/interpreting regular expressions

See also, SuppressWithPlainTextCommentFilter [[Docs](https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/filters/SuppressWithPlainTextCommentFilter.html)], where examples cover these usages

src/test/java/org/opensearch/security/ccstest/CrossClusterSearchTests.java
```diff
@@ -84,7 +84,9 @@ public ClusterTransportClientSettings() {
         }
 
         public ClusterTransportClientSettings(Settings clusterSettings, Settings transportSettings) {
-            super(clusterSettings, transportSettings);
+            super(Settings.builder()
+                .put(clusterSettings)
+                .putList("node.roles", "remote_cluster_client").build(), transportSettings);
```
peternied, line 89: This is required for cross-cluster to be enabled https://opensearch.org/docs/latest/replication-plugin/get-started/#prerequisites

Note; I did not investigate why this wasn't required before.

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
```diff
@@ -194,6 +197,28 @@ public void waitForInit(Client client) {
...
+        }
+
+        final Settings nodeRoleSettings = Settings.builder().putList(NODE_ROLE_KEY, nodeRolesBuilder.build()).build();
+        return mergeNodeRolesAndSettings(settingsBuilder, nodeRoleSettings);
+    }
+
+    public static Settings.Builder mergeNodeRolesAndSettings(final Settings.Builder settingsBuilder, final Settings otherSettings) {
```
peternied, line 213: Why add this method?

Settings.Builder.put(Settings) [[code](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/settings/Settings.java#L1032)] takes the map representation of the incoming settings and stomps over top _all_ intersecting keys.

Depending when `.put(settings)` was used this would stomp the `{"node.roles": ["cluster_manager"]}` -> `{"node.roles":["remote_cluster_client"]}` - this would stop the cluster from starting because it couldn't find any cluster_manager eligable nodes OR the inverse would happen and `remote_cluster_client` role would not be available on the cluster.

^ I started that sentence with _depending_, and yes that is because sometimes the test configuration is providing the default values before the cluster helper does its thing, and sometimes it's provided through a `NodeSettingsSupplier` which is applied afterward, but before the cluster starts up.

This method is available to handle the bi-directional merges of these settings and ensure the `node.roles` are merged instead of replaced.

### That was difficult and complex, why go through that trouble for only `node.roles`?
Why not just support something like `.merge(Settings)`? - IMO this is a better solution except of course if our cluster configuration is taking advantage of this functionality - its a feature not a bug!

I avoided tackling this problem because the data flow around these settings is non-trivial and debugging and troubleshooting hundreds of integration unit tests compared to fixing the smaller subset of cross cluster tests that are broken seemed like the better short-term path.

As we build on the new test framework, being rigorous around the data flow can help avoid this class of problems

src/test/java/org/opensearch/security/test/helper/cluster/ClusterHelper.java
```diff
@@ -171,9 +174,15 @@ public final synchronized ClusterInfo startCluster(final NodeSettingsSupplier no
         for (int i = 0; i < internalClusterManagerNodeSettings.size(); i++) {
             NodeSettings setting = internalClusterManagerNodeSettings.get(i);
             int nodeNum = nodeNumCounter--;
-            PluginAwareNode node = new PluginAwareNode(setting.clusterManagerNode,
-                    getMinimumNonSecurityNodeSettingsBuilder(nodeNum, setting.clusterManagerNode, setting.dataNode, internalNodeSettings.size(), tcpClusterManagerPortsOnly, tcpPortsAllIt.next(), httpPortsIt.next())
-                            .put(nodeSettingsSupplier == null ? Settings.Builder.EMPTY_SETTINGS : nodeSettingsSupplier.get(nodeNum)).build(), setting.getPlugins());
+            final Settings.Builder nodeSettingsBuilder = getMinimumNonSecurityNodeSettingsBuilder(nodeNum, setting.clusterManagerNode, setting.dataNode, internalNodeSettings.size(), tcpClusterManagerPortsOnly, tcpPortsAllIt.next(), httpPortsIt.next());
```
peternied, line 177: This is kind of painful - see AbstractSecurityUnitTest.mergeNodeRolesAndSettings where more background is provided.

Also, why not make this a function?  This code uses a lot of locals to pass information around and the method signature would be untenable.  Rather than refactor ClusterHelper for this case, I'm continuing the rich tradition of copy-pasting the update.

If we have strong feelings this could be an issue, but I doubt it would survive triage with the new test framework incoming

peternied (commented):
@cwperks @DarshitChanpura @RyanL1997 Mind taking another look?

RyanL1997 (dismissed):

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
cwperks: I believe `NodeRoleSettings.NODE_ROLES_SETTING.getKey()` resolves to `node.roles`.

peternied: My last attempt to push back - I do not think it's worthwhile.  If you feel strongly, we could create an issue to clean this up or ask again and I'll just do it.

cwperks: I think it's fine as is. For my own understanding I searched through core to find where a list of the potential values was coming from. If someone comes to modify this function in the future to add additional node types they will need to be aware of what the list of potential node types is.

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
cwperks: Thanks for the explanation for this method. 

Would making this more generic and having the method take in a key be useful?

i.e. `public static Settings.Builder mergeListSettingByKey(final Settings.Builder settingsBuilder, final Settings otherSettings, String settingKey)`

cwperks (approved):
Nice work on discovering the issue with `put(Settings)` when switching from using `node.master, node.data and node.ingest` to using `node.roles` with a list of roles!

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
peternied: ^ Next time we need to do this I think it would be worthwhile.

DEVELOPER_GUIDE.md
```diff
@@ -146,6 +147,35 @@ Integration tests are automatically run on all pull requests for all supported v
...
+Checkstyle enforced several rules within this codebase.  Sometimes exceptions will be necessary for components that are set for deprecation but the new version is unavailable.  There are two formats of suppression that can be used when dealing with violations of this nature, one for disabling a single rule, or another for disabling all rules - its best to be as specific as possible.
+
+*Execute Checkstyle*
+```
+./gradlew checkstyleMain checkstyleTest 
+```
+
+*Example violation*
+```
+[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java:178: Usage should be switched to cluster manager [RegexpSingleline]
```
cliu123, line 160: @peternied Could you avoid real name/alias(`petern`) in the code? If the full path is not needed, please remove it. Otherwise, please use a placeholder to instead of a real name.

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
cliu123: >If someone comes to modify this function in the future to add additional node types they will need to be aware of what the list of potential node types is.

@peternied Are there any other occurrences of this? If so, we can resolve this in a separate issue for clean-up. Otherwise, could we avoid unnecessary follow-up issues only for this minor change? IMHO, @cwperks is making a legitimate request for this.

DEVELOPER_GUIDE.md
peternied: Done!

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
cliu123: @peternied Thanks for the comments making review much easier!

>Would making this more generic and having the method take in a key be useful?

@cwperks very interesting idea! I might miss something. Are there any other use cases where `settingKey` can be something else other than `NODE_ROLE_KEY`? If so, making the method more generic would be necessary to serve those use cases.

DarshitChanpura: Interesting..thank you for the explanation. I like the idea of making it generic if there are other values as `List` objects that would "stomp" each other

src/test/java/org/opensearch/security/test/helper/cluster/ClusterHelper.java
DarshitChanpura: If this a one-place-only kind of usage. I'm not against this approach.

DarshitChanpura (approved):
thank you @peternied !!

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
peternied: We do not have any other use cases at the moment.  If we found ourselves using a function like this often I think it would be more worthwhile to make a pull request to update Settings.Builder with this functionality, but that seems out of scope for this PR.  What do you think @cwperks @cliu123 ?

cliu123: If there are not other use cases, then I am good with moving forward as-is. @cwperks What do you think?

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
cliu123: > If someone comes to modify this function in the future to add additional node types they will need to be aware of what the list of potential node types is.

This is true. Using `DiscoveryNodeRole.CLUSTER_MANAGER_ROLE.roleName()` instead would make people easier to find out the list of potential node types when this function is getting changed in the future.

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
cwperks: @cliu123 I believe anywhere where you see `putList` in the code would be eligible: https://github.com/opensearch-project/security/search?q=putList

src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java
cwperks: @peternied Something to consider about using the existing constant here: `CLUSTER_MANAGER_ROLE` is a good example since it was recently renamed. If core modifies or removes a role then we will get a compilation error when using the constant. When hardcoding a string the code will still compile.

peternied: It's been updated.

cwperks: Thank you @peternied!

cwperks (approved):

Issue #2090, opened 2022-09-15
[FEATURE] Support LDAP pool pruning


Martin-Kemp:
**Is your feature request related to a problem?**

LDAP pool pruning is a common requirement for connecting to productive ldap servers. Currently the pruning period and idle time is hard coded to 5 and 10 minutes respectively. It would be nice to be able to configure this through `config.yml`.

**What solution would you like?**

Configure ldap connection pool pruning period and idle time through `config.yml`. For example, currently you can enable pooling and set the min and max size. ideally `pool.idle_time` and `pool.pruning_period` should also be configurable. Like this:
```
pool.enabled: true
pool.min_size: 0
pool.max_size: 5
pool.idle_time: 2
pool.pruning_period: 1
```

**What alternatives have you considered?**

I'm currently running a custom image with above change.

**Do you have any additional context?**

Pruning period and idle time is currently set [here](https://github.com/opensearch-project/security/blob/main/src/main/java/com/amazon/dlic/auth/ldap2/LDAPConnectionFactoryFactory.java#L130)

Maybe I'm not understanding the code correctly and this config can actually be set somewhere because there is a setting referenced:

```
        result.setPruneStrategy(new IdlePruneStrategy(Duration.ofMinutes(this.settings.getAsLong("pruning.period", 5l)),
                Duration.ofMinutes(this.settings.getAsLong("pruning.idleTime", 10l))));
```
But I have no idea where to set `pruning.period` and `pruning.idleTime`.

peternied:
The fix for this issue has been merged, and will be released in the next 2.X distribution

PR #2091, opened 2022-09-15
Make ldap pool period and idle time configurable

Martin-Kemp:
### Description
Configure ldap connection pool pruning period and idle time through config.yml. For example, currently you can enable pooling and set the min and max size. ideally pool.idle_time and pool.pruning_period should also be configurable. Like this:
```
pool.enabled: true
pool.min_size: 0
pool.max_size: 5
pool.idle_time: 2
pool.pruning_period: 1
```
* Category (Enhancement)
To comply with strict ldap providers.
Defaults are the same, this just makes things configurable.

### Issues Resolved
https://github.com/opensearch-project/security/issues/2090

No

### Testing
Manual testing locally.

### Check List

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/ldap/util/ConfigConstants.java
+++ b/src/main/java/com/amazon/dlic/auth/ldap/util/ConfigConstants.java
@@ -82,6 +82,9 @@ public final class ConfigConstants {
 
     public static final String LDAP_POOL_TYPE = "pool.type";
 
+    public static final String LDAP_POOL_PRUNING_PERIOD = "pool.pruning_period";
+    public static final String LDAP_POOL_IDLE_TIME = "pool.idle_time";
+
     private ConfigConstants() {
 
     }
--- a/src/main/java/com/amazon/dlic/auth/ldap2/LDAPConnectionFactoryFactory.java
+++ b/src/main/java/com/amazon/dlic/auth/ldap2/LDAPConnectionFactoryFactory.java
@@ -127,8 +127,11 @@ public class LDAPConnectionFactoryFactory {
         }
 
         result.setValidator(getConnectionValidator());
-        result.setPruneStrategy(new IdlePruneStrategy(Duration.ofMinutes(this.settings.getAsLong(ConfigConstants.LDAP_POOL_PRUNING_PERIOD, 5l)),
-                Duration.ofMinutes(this.settings.getAsLong(ConfigConstants.LDAP_POOL_IDLE_TIME, 10l))));
+          
+        result.setPruneStrategy(new IdlePruneStrategy(
+            Duration.ofMinutes(this.settings.getAsLong(ConfigConstants.LDAP_POOL_PRUNING_PERIOD, this.settings.getAsLong("pruning.period", 5l))),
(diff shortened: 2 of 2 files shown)


src/main/java/com/amazon/dlic/auth/ldap/util/ConfigConstants.java
```diff
@@ -82,6 +82,9 @@ public final class ConfigConstants {
 
     public static final String LDAP_POOL_TYPE = "pool.type";
 
+    public static final String LDAP_POOL_PRUNING_PERIOD = "pool.pruning_period";
```
cwperks, line 85: Thank you for creating a ConfigConstant out of these values. It looks like there's a lack in documentation for these settings: https://github.com/opensearch-project/documentation-website/search?q=pruning.period and https://github.com/opensearch-project/documentation-website/search?q=pruning.idleTime.

I will create an issue on the documentation website to review the LDAP documentation and add or update where needed. 

While `lower_snake_case` appears to be the convention for settings keys, this may have the potential to break existing configs referencing these settings. 

What do you think about supporting the existing settings keys as well?

```
this.settings.getAsLong(ConfigConstants.LDAP_POOL_PRUNING_PERIOD, this.settings.getAsLong("pruning.period", 5l))
this.settings.getAsLong(ConfigConstants.LDAP_POOL_IDLE_TIME, this.settings.getAsLong("pruning.idleTime", 10l))
```

cwperks: @cwillum - Looks like the documentation site for LDAP Configuration needs a review. https://opensearch.org/docs/latest/security-plugin/configuration/ldap/

Martin-Kemp: > What do you think about supporting the existing settings keys as well?

Yes, good idea. I'll update it on Monday.

Martin-Kemp: > @cwillum - Looks like the documentation site for LDAP Configuration needs a review. https://opensearch.org/docs/latest/security-plugin/configuration/ldap/

The whole ldap2 backend type isn't documented, I'm not even sure if it's actually supposed to be used.

cwillum: I added an issue for review and update of LDAP documentation: [#1248](https://github.com/opensearch-project/documentation-website/issues/1248).

peternied (changes requested):
We need to support backward capability, registering the previous setting as deprecated `Setting.Property.Deprecated` on the current settings values, and then registering the new ones you've created. This would allow the proper behavior and allow for fallback for anyone that used the code as documentation.

I know this was mentioned on the thread with @cwperks, just making sure to document how to move forward with this PR.

Martin-Kemp:
> registering the previous setting as deprecated Setting.Property.Deprecated on the current settings values, and then registering the new ones you've created

@peternied, I'm not sure how to do this. Could you point me in the right direction? How do I set the property to Deprecated?

I've pushed another commit to support the existing keys as suggested by @cwperks

peternied (approved):
@Martin-Kemp No worries, I'll make a follow up PR to add the deprecation notice for the legacy settings, what you've done here is great!

cwperks (approved):

Issue #2087, opened 2022-09-12
Point in time security model


bharath-techie:
#### What is point in time ?

User can create a Point In Time for a set of indices. A Point In Time is a set of segments which are frozen in time taken at the moment of creation. We lock the segments of those indices’ shards and create contexts to access and query only those shards.

Such point in time contexts can be reused across multiple search queries and the data that is searched upon will be on the locked segments. This solution is used with search after for pagination with data consistency.

This can be related to ‘scroll’ , whereas ‘scroll’ context gets created for each query ( cannot be reused for different search queries ) and ‘scroll’ can only move forward.
Design doc : https://github.com/opensearch-project/OpenSearch/issues/3960
Meta issue: https://github.com/opensearch-project/OpenSearch/issues/3959

#### Point in time APIs

* Create point in time
    * The request implements ‘IndicesRequest.Replaceable“
    * action - “indices:data/read/point_in_time/create”
    * User can create point in time on *any index* for which the user has access to ( similar to search )
* Delete point in time
    * User can pass *list of PITs* to delete (or) user can choose to *delete all PITs*
    * action - “indices:data/read/point_in_time/delete”
* List all PITs 
    * This API lists all PITs that the user has access to
    * Rest user facing action to list PITs- “indices:data/read/point_in_time/readall”
    * Action to retreive all cluster PITs - “cluster:admin/point_in_time/readall”
* Point in time segments API
    * This shows the resource utilization of segments held because of a point in time context.
    * User can pass *list of PITs* to show segments info (or) user can choose to show *all PIT segments info*
    * Action - "indices:monitor/segments/point_in_time"

#### Permission model 

Action group:

```
manage_point_in_time:
  reserved: true
  hidden: false
  static: true
  allowed_actions:
    - "indices:data/read/point_in_time/create"
    - "indices:data/read/point_in_time/delete"
    - "indices:data/read/point_in_time/readall"
    - "cluster:admin/point_in_time/readall"
    - "indices:monitor/point_in_time/segments"
  type: "index"
  description: "Manage point in time actions"
```

Role which has access to all PIT actions :
```
point_in_time_1:
  reserved: true
  hidden: false
  description: "Migrated from v6 (all types mapped)"
  cluster_permissions:
    - "cluster:admin/point_in_time/read_from_nodes"
  index_permissions:
    - index_patterns:
        - "pit_1"
      dls: null
      fls: null
      masked_fields: null
      allowed_actions:
        - "manage_point_in_time"
```

[...]

PR #2094, opened 2022-09-19
Point in time security changes

bharath-techie:
Signed-off-by: Bharathwaj G <bharath78910@gmail.com>

### Description

 - - For 'Delete PIT' and 'PIT segments' API, when PIT IDs are passed as part of request, this custom evaluator decode the PITs to indices and resolve the indices with user permissions.

- - If user has permission to all indices of PIT , then PIT is permitted to the user
 
- - Only when the user has permissions for all PITs in the request, then we allow the operation.

- - For requests which operates on 'all' PITs, we skip the custom evaluator and evaluate via standard code

Alias and data stream behavior :

- PIT IDs always contain the resolved indices ( underlying indices ) when saved.
- Based on this, 
  - For alias, user must have either 'index' or 'alias' permission for any PIT operation.
  - For data stream, user must have both 'data stream' AND 'backing indices of data stream' permission ( eg : data-stream-11 + .ds-my-data-stream11-000001 ) for any PIT operation. 
  - With just data stream permission, user will be able to create pit but will not be able to use the PIT ID for other operations such as search without backing indices permission.

Refer to : 
https://github.com/opensearch-project/security/issues/2087#issuecomment-1247972936

   New feature - Modification to point in time security model
  We're making these changes to enhance the security model of PIT as per review comments from opensearch community

### Issues Resolved
https://github.com/opensearch-project/security/issues/2087

https://github.com/opensearch-project/OpenSearch/issues/3959

### Testing

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
@@ -78,7 +78,7 @@ public class PitPrivilegesEvaluator {
         Set<String> allPermittedIndices = getPermittedIndices(pitIndices, clusterService, user,
                 securityRoles, action, resolver);
         // Only if user has access to all PIT's indices, allow operation, otherwise continue evaluation in PrivilegesEvaluator.
-        if(pitIndices.size() == allPermittedIndices.size()) {
+        if(allPermittedIndices.containsAll(pitIndices)) {
             presponse.allowed = true;
             presponse.markComplete();
         }
--- a/src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
@@ -16,18 +16,15 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-
-import com.google.common.collect.ImmutableSet;
+import java.util.concurrent.TimeUnit;
 
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.admin.indices.segments.PitSegmentsRequest;
+import org.opensearch.action.search.CreatePitRequest;
 import org.opensearch.action.search.DeletePitRequest;
-import org.opensearch.action.search.GetAllPitNodesRequest;
-import org.opensearch.action.search.GetAllPitNodesResponse;
-import org.opensearch.action.search.ListPitInfo;
-import org.opensearch.action.search.SearchRequest;
 import org.opensearch.cluster.metadata.IndexNameExpressionResolver;
 import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.security.OpenSearchSecurityPlugin;
 import org.opensearch.security.resolver.IndexResolverReplacer;
 import org.opensearch.security.securityconf.SecurityRoles;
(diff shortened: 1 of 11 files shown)


src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
```diff
@@ -0,0 +1,101 @@
...
+                GuiceHolder.getPitService().getIndicesForPits(pitIds);
+        Set<String> pitIndices = new HashSet<>();
+        // add indices across all PITs to a set and evaluate if user has access to all indices
+        for(String[] indices: pitToIndicesMap.values()) {
+            pitIndices.addAll(Arrays.asList(indices));
+        }
+        Set<String> allPermittedIndices = getPermittedIndices(pitIndices, clusterService, user,
+                securityRoles, action, resolver);
+        // Only if user has access to all PIT's indices, allow operation, otherwise continue evaluation in PrivilegesEvaluator.
+        if(pitIndices.size() == allPermittedIndices.size()) {
```
peternied, line 81: Since we have the full list of expected indices, what do you think about confirming that the listed PIT indices are all contained in the list of the permitted indices?

src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * This method returns list of permitted indices for the PIT indices passed
+     */
+    private Set<String> getPermittedIndices(Set<String> pitIndices, ClusterService clusterService,
+                                            User user, SecurityRoles securityRoles, final String action,
+                                            IndexNameExpressionResolver resolver) {
+        final ImmutableSet<String> pitImmutableIndices = ImmutableSet.copyOf(pitIndices);
+        final IndexResolverReplacer.Resolved pitResolved =
```
peternied, line 95: How does this resolve with aliases or data streams?

peternied (changes requested):
Much simplified version of these checks, great work!  Other than the few comments, please add integration tests.

bharath-techie:
> Much simplified version of these checks, great work! Other than the few comments, please add integration tests.

One problem with integration tests is that current framework doesn't let me pass body to 'delete
 or 'get' REST requests. Any suggestions ?

I can validate only '_all' requests.

peternied:
For these scenarios that aren't included in the test framework, could you add overloads that cover these scenarios?

src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
bharath-techie: done

src/main/java/org/opensearch/security/privileges/PitPrivilegesEvaluator.java
bharath-techie: Good question - don't think this logic handles other than local indices.
```
 final ResolvedIndicesProvider resolvedIndicesProvider = new ResolvedIndicesProvider(request);

        getOrReplaceAllIndices(request, resolvedIndicesProvider, false);

        return resolvedIndicesProvider.resolved(indicesOptionsFrom(request))
```

I think i have to use this logic instead. Let me try debugging and update this.

src/test/java/org/opensearch/security/PitIntegrationTests.java
```diff
@@ -0,0 +1,185 @@
+/*
+bharath-techie marked this conversation as resolved.
```
peternied, line 2: I don't think you meant to include this line in the copyright header

peternied (changes requested):
Thanks for the update for the tests, There is still an open question about data streams/aliases.  Could you add tests for these scenarios - this will confirm the behavior is as expected.

bharath-techie:
> Thanks for the update for the tests, There is still an open question about data streams/aliases. Could you add tests for these scenarios - this will confirm the behavior is as expected.

I've addressed the question about data stream / aliases. And also have added the tests validating the same.

peternied:
@bharath-techie It looks like there are build issues impacting the CI.  I'm happy to approve once we see those tests are passing

bharath-techie:
@peternied 
Thanks for the review, builds will pass post the 2.4 backport of opensearch PRs. I'll update this PR post that.

bharath-techie:
Hi @peternied,
Build is green finally, please review the changes.

peternied (approved):
@bharath-techie Excellent work! LGTM

PR #2051, opened 2022-08-24
Support for HTTP/2 (server-side)

reta:
Signed-off-by: Andriy Redko <andriy.redko@aiven.io>

### Description
Implement SSL/TLS support for HTTP/2, see [1]

- HTTP/2 (default)
```
$ curl https://localhost:9200 -kv -u admin:admin

*   Trying 127.0.0.1:9200...
* Connected to localhost (127.0.0.1) port 9200 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* TLSv1.0 (OUT), TLS header, Certificate Status (22):
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS header, Certificate Status (22):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS header, Finished (20):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Request CERT (13):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.2 (OUT), TLS header, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* TLSv1.3 (OUT), TLS handshake, Certificate (11):
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256
* ALPN, server accepted to use h2
* Server certificate:
*  subject: DC=de; L=test; O=node; OU=node; CN=node-0.example.com
*  start date: Apr 22 03:43:47 2018 GMT
*  expire date: Apr 19 03:43:47 2028 GMT
*  issuer: DC=com; DC=example; O=Example Com Inc.; OU=Example Com Inc. Root CA; CN=Example Com Inc. Root CA
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
* Using HTTP2, server supports multiplexing
* Connection state changed (HTTP/2 confirmed)
[...]

PR #2106, opened 2022-09-23
Keep version upgrade test working past v2

peternied:
### Description
This change locks the version update to a known good 2.1.0-snapshot so no matter the future version numbers being used, this test scenario will still function.

When the version update is applied, it expects the build system for the future version to exist, otherwise `gradle properties` will fail to load.  This was causing issues when v3 updates were expecting v4 versions of opensearch build tools.

### Issues Resolved
- Related https://github.com/opensearch-project/security/pull/2051

### Testing
Tested locally, and CI will verify

### Check List


.github/workflows/ci.yml
```diff
@@ -126,12 +126,12 @@ jobs:
...
 
     - run:  |
-        ## EXISTING_OS_VERSION outputs the major version, example as 2
-        EXISTING_OS_VERSION=$(./gradlew properties | grep opensearch.version | cut -d':' -f2- | awk '{$1=$1};1' | cut -d '-' -f1 | cut -d '.' -f1)
-        ## INCREMENT_OS_VERSION in an increment of 1, example if EXISTING_OS_VERSION is 2, INCREMENT_OS_VERSION is 3
-        INCREMENT_OS_VERSION=$((++EXISTING_OS_VERSION)) 
-        ./gradlew clean updateVersion -DnewVersion=$INCREMENT_OS_VERSION.0.0-SNAPSHOT
-        test `./gradlew properties | grep opensearch.version | cut -d':' -f2- | awk '{$1=$1};1'` = $INCREMENT_OS_VERSION.0.0-SNAPSHOT
+        ## Make sure the current doesn't match the test version
+        test "$(./gradlew properties | grep opensearch.version)" != "opensearch_version: 2.1.0-SNAPSHOT"
```
reta, line 130: @peternied I am wondering what is special about `2.1.0-SNAPSHOT` version? shouldn't it be aligned with `opensearch.version`?

cwperks: Is this to test the version increment automation at build time? If so, should we create a meaningless tag like `VERIFY-VERSION-INCREMENT`?

DarshitChanpura: Will this change be reverted to original at some point?

peternied: This value should never be changed

.github/workflows/ci.yml
```diff
@@ -125,13 +125,15 @@ jobs:
...
     - run: ./gradlew clean assemble -Dbuild.version_qualifier=${{ env.TEST_QUALIFIER }} && test -s ./build/distributions/opensearch-security-${{ env.SECURITY_PLUGIN_VERSION_ONLY_NUMBER }}-${{ env.TEST_QUALIFIER }}-SNAPSHOT.zip
 
-    - run:  |
-        ## EXISTING_OS_VERSION outputs the major version, example as 2
-        EXISTING_OS_VERSION=$(./gradlew properties | grep opensearch.version | cut -d':' -f2- | awk '{$1=$1};1' | cut -d '-' -f1 | cut -d '.' -f1)
-        ## INCREMENT_OS_VERSION in an increment of 1, example if EXISTING_OS_VERSION is 2, INCREMENT_OS_VERSION is 3
-        INCREMENT_OS_VERSION=$((++EXISTING_OS_VERSION)) 
-        ./gradlew clean updateVersion -DnewVersion=$INCREMENT_OS_VERSION.0.0-SNAPSHOT
-        test `./gradlew properties | grep opensearch.version | cut -d':' -f2- | awk '{$1=$1};1'` = $INCREMENT_OS_VERSION.0.0-SNAPSHOT
+    - env:
```
cwperks, line 128: @peternied Can we also add a `name:` here and describe the test. I think adding something like `name: Verify version increment automation` or something referenceing that this is a test of the gradle task `updateVersion`.

I've tested this locally and will approve once a descriptive name is added.

cwperks (approved):
Thanks!

.github/workflows/ci.yml
peternied: I've moved this to a variable to make this more apparent in the workflow description.

PR #2116, opened 2022-09-28
Adding tenancy access to user info in security thread context

vamsi-amazon:
Signed-off-by: vamsi-amazon <reddyvam@amazon.com>

### Description
* Adding tenancy access info into thread context.
* Downstream plugins could utilize this information to base their behavior.

### Testing
Manual Testing.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
@@ -188,10 +188,28 @@ public class PrivilegesEvaluator {
             joiner.add(String.join(",", user.getRoles()));
             joiner.add(String.join(",", Sets.union(user.getSecurityRoles(), mappedRoles)));
             String requestedTenant = user.getRequestedTenant();
-            if (!Strings.isNullOrEmpty(requestedTenant)) {
-                joiner.add(requestedTenant);
-            }
+            joiner.add(requestedTenant);
+            String tenantAccessToCheck = getTenancyAccess(requestedTenant, mapTenants(user, mappedRoles));
+            joiner.add(tenantAccessToCheck);
+            log.info(joiner);
             threadContext.putTransient(OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT, joiner.toString());
+        }
+    }
+
+    private String getTenancyAccess(String requestedTenant, Map<String, Boolean> tenancyAccessMap) {
+        if(Strings.isNullOrEmpty(requestedTenant)) {
+            requestedTenant = "global_tenant";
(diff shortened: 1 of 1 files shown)


peternied:
Thanks for creating @vamsi-amazon - I noticed that this change is in draft, I'm going to hold off on a more detailed review unless there is something you'd like feedback on.  Feel free to make a comment on any lines of code you would like to solicit feedback on.

Just a note before flipping to 'ready for review', please add tests that can make sure the change operates correctly and isn't regressed by any future updates to this codebase.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -188,13 +188,31 @@ private void setUserInfoInThreadContext(User user, Set<String> mappedRoles) {
             joiner.add(String.join(",", user.getRoles()));
             joiner.add(String.join(",", Sets.union(user.getSecurityRoles(), mappedRoles)));
             String requestedTenant = user.getRequestedTenant();
-            if (!Strings.isNullOrEmpty(requestedTenant)) {
```
DarshitChanpura, line 191: why was this null check removed?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -188,13 +188,31 @@ private void setUserInfoInThreadContext(User user, Set<String> mappedRoles) {
...
-                joiner.add(requestedTenant);
-            }
+            joiner.add(requestedTenant);
+            String tenantAccessToCheck = getTenancyAccess(requestedTenant, mapTenants(user, mappedRoles));
+            joiner.add(tenantAccessToCheck);
+            log.info(joiner);
```
DarshitChanpura, line 194: do we need this log print?

vamsi-amazon: My bad, wrong patch applied. I changed it to debug.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
vamsi-amazon: The parsing of this string in the downstream plugins happens in a weird way. 
In case of empty requested Tenant(Global Tenant) 

if we don't add anything for requestedTenant.
        The string generated would look like this `myuser|Roles|securityRoles|No`
If we add irrespective
       The string generated would look like this `myuser|Roles|securityRoles||No`

The extra pipe would help downstream plugins not to mistake tenancyAccessRole with tenantName.

PR #2135, opened 2022-09-30
Allows for configuration of LDAP referral following

### LINKED ISSUE #195: RFE: Be able to disable following ldap referrals
Like the elasticsearch xpack plugin (https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html) it would be good to be able to disable following ldap referrals.  It can generate a lot of failed connections when connecting to AD LDAP servers.

scrawfor99:
These changes allow for the configuration of the LDAP referral response behavior. Behavior can be set by modifying the `FOLLOW_REFERRALS` field of the ConfigConstants.java file. By default, referrals are followed in exactly the same way as previous releases. It is also possible to manually modify individual search instances by passing `true|false` to the search calls processed by the `LdapHelper` class.

Turning off referral following will result in a search result with LDAP dn = "cn=couldNotFollowReferral" this behavior can be modified in `LdapHelper.java`.

- Resolves #195 

### Testing
Added tests to the `test/java/com/amazon/dlic/auth/ldap` tests cases for LDAP. For each test, there is a check for the functioning of the referrals when `shouldFollowReferrals` is `true` as well as a test for when it is `false`. This methodology is in-line with the previous tests for following referrals. 

### Check List


src/main/java/com/amazon/dlic/auth/ldap/backend/LDAPAuthenticationBackend.java
```diff
@@ -54,6 +54,7 @@ public class LDAPAuthenticationBackend implements AuthenticationBackend {
 
     private final Settings settings;
     private final Path configPath;
+    private final boolean followReferrals;
```
peternied, line 57: I like to name my booleans as questions where true/false has a clear answer, consider `followReferrals` -> `shouldFollowReferrals`

src/main/java/com/amazon/dlic/auth/ldap/backend/LDAPAuthorizationBackend.java
```diff
@@ -99,7 +99,7 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
     private final Settings settings;
     private final WildcardMatcher skipUsersMatcher;
     private final WildcardMatcher nestedRoleMatcher;
-
+    //private final boolean followReferrals;
```
peternied, line 102: Delete commented out code

peternied: In a couple of places

src/main/java/com/amazon/dlic/auth/ldap/util/LdapHelper.java
```diff
@@ -34,37 +34,46 @@
...
+import org.opensearch.common.settings.Settings;
 
 public class LdapHelper {
 
     private static SearchFilter ALL = new SearchFilter("(objectClass=*)");
+
     @SuppressWarnings("removal")
     public static List<LdapEntry> search(final Connection conn, final String unescapedDn, SearchFilter filter,
-            final SearchScope searchScope, final String[] returnAttributes) throws LdapException {
+            final SearchScope searchScope, final String[] returnAttributes, Settings settings) throws LdapException {
```
peternied, line 45: searching an Ldap connection doesn't seem like it should require an OpenSearch.Settings object, instead of passing the settings all the way down, we should pass a parameter to follow referrals or not

src/main/java/com/amazon/dlic/auth/ldap/util/LdapHelper.java
```diff
@@ -34,37 +34,46 @@
...
+                        request.setReferralHandler(new SearchReferralHandler());
+                        final Response<SearchResult> r = search.execute(request);
+                        // referrals will be followed to build the response
+                        final org.ldaptive.SearchResult result = r.getResult();
+                        entries.addAll(result.getEntries());
+                        return entries;
+                    } 
+                    else{
+
+                        return entries;
```
peternied, line 75: This doesn't look correct if `r.getResults()` wasn't called

src/main/java/com/amazon/dlic/auth/ldap/util/LdapHelper.java
```diff
@@ -34,37 +34,46 @@
...
             sm.checkPermission(new SpecialPermission());
         }
-
         try {
             final String baseDn = escapeDn(unescapedDn);
             return AccessController.doPrivileged(new PrivilegedExceptionAction<List<LdapEntry>>() {
                 @Override
                 public List<LdapEntry> run() throws Exception {
                     final List<LdapEntry> entries = new ArrayList<>();
                     final SearchRequest request = new SearchRequest(baseDn, filter);
```
peternied, line 58: One of the other requests in the source ticket was if there was any failures when attempting to resolve the referrals a partial response could still be returned [see comment](https://github.com/opensearch-project/security/issues/195#issuecomment-717058416)  I'm not sure I agree with this behavior, I think its easier not to swallow failures - did you consider that scenario, what do you think?

config/opensearch.yml.example
```diff
@@ -38,6 +38,9 @@ plugins.security.authcz.admin_dn:
 # BOTH - backend roles are mapped to Security roles mapped directly and via roles_mapping.yml in addition
 plugins.security.roles_mapping_resolution: MAPPING_ONLY
 
+# Define behavior for following referalls from LDAP reponses (default: true)
+plugins.security.FOLLOW_REFERRALS: true
```
peternied, line 42: I don't think this should be such a high level setting, and it isn't following the standard casing convensions, I think this should be specific for an LDAP configuration, please move the setting into the LDAP specific section of the configuration

peternied (changes requested):
A number of comments in the code, please follow up on those individually.

I am concerned that no tests broke as I am uncertain SearchRequest are still working as expected.  I think we might need to add some tests that confirm that search is still doing what's expected.  Please add a new LdapHelperTests.java, take a look at https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/securityconf/impl/v7/IndexPatternTests.java for how you could write mocks / unit tests for the `search` method

src/main/java/com/amazon/dlic/auth/ldap/util/LdapHelper.java
scrawfor99: This functions to return an empty response when `shouldFollowReferrals`  is `false` and the search would require referral following. I am not sure what the best behavior would be as the use cases leading to this request are so diverse. 

If you had suggestions on what you think should be returned in this case they would be appreciated.

src/main/java/com/amazon/dlic/auth/ldap/util/LdapHelper.java
scrawfor99: I do not think partial responses are the solution since it seems like it would cause further issues for any code expecting full responses. This would likely just move whatever error is resolved by adding the `shouldFollowReferrals` field further down the line. 

I could be wrong however.

config/opensearch.yml.example
scrawfor99: Oops, you are correct. This was not where this should go.

src/main/java/com/amazon/dlic/auth/ldap/backend/LDAPAuthorizationBackend.java
scrawfor99: Sorry, I am not sure why we are removing the shouldFollowReferrals attribute here. Since search is called from the AuthorizationBackend file, I was under the impression it was necessary to be able to specify the referral following behavior on these searches as well.

Perhaps I am mistaken though so just let me know.

src/main/java/com/amazon/dlic/auth/ldap/util/LdapHelper.java
scrawfor99: Behavior has been updated to elicit an Ldap response of `dn="cn=couldNotFollowReferrals"` when `shouldFollowReferrals` is `false` and following is attempted.

src/main/java/com/amazon/dlic/auth/ldap/util/LdapHelper.java
scrawfor99: Modified so that it now returns a singleton list with an Ldap entry that is `dn="cn=couldNotFollowReferral"` this can be modified by users as they see fit but allows for testing and provides a concise message about the search query.

peternied: I think the code should look like this with the referral handler unset unless `shouldFollowReferrals` is enabled.  What do you think of this?
```java
                        if (shouldFollowReferrals) {
                           request.setReferralHandler(new SearchReferralHandler());  
                        }

                        final Response<SearchResult> r = search.execute(request);
                        // referrals will be followed to build the response
                        final org.ldaptive.SearchResult result = r.getResult();
                        entries.addAll(result.getEntries());
                        return entries;
```

config/opensearch.yml.example
peternied: Maybe this wasn't updated?

src/main/java/com/amazon/dlic/auth/ldap/backend/LDAPAuthenticationBackend.java
```diff
@@ -228,7 +230,7 @@ private static LdapEntry existsSearchingUntilFirstHit(final String user, Connect
                     baseSettings.get(ConfigConstants.LDAP_AUTHCZ_BASE, DEFAULT_USERBASE),
                     f,
                     SearchScope.SUBTREE,
-                    returnAttributes);
+                    returnAttributes, true);
```
peternied, line 233: I think we should follow the setting consistantly in all cases, so this should still read from the setting

src/main/java/com/amazon/dlic/auth/ldap/backend/LDAPAuthenticationBackend.java
```diff
@@ -54,6 +54,7 @@ public class LDAPAuthenticationBackend implements AuthenticationBackend {
 
     private final Settings settings;
     private final Path configPath;
+    private final boolean shouldFollowReferrals;
```
peternied, line 57: Note; in this file there are many static methods that cannot use the field, maybe better to remove the field.

src/main/java/com/amazon/dlic/auth/ldap/util/LdapHelper.java
```diff
@@ -35,36 +35,45 @@
...
+                    if (shouldFollowReferrals) {
+                        request.setReferralHandler(new SearchReferralHandler());
+                        final Response<SearchResult> r = search.execute(request);
+                        // referrals will be followed to build the response
+                        final org.ldaptive.SearchResult result = r.getResult();
+                        entries.addAll(result.getEntries());
+                        return entries;
+                    } 
+                    else{
+                        entries.add(new LdapEntry("cn=couldNotFollowReferral"));
```
peternied, line 74: I don't think adding entries like this will be helpful to callers of the API.  I think the test can be updated to check for the lack of a referral retrieved value, what do you think?

src/test/java/com/amazon/dlic/auth/ldap/LdapBackendTestNewStyleConfig.java
```diff
@@ -341,14 +341,34 @@ public void testLdapAuthenticationReferral() throws Exception {
...
+                .putList(ConfigConstants.LDAP_HOSTS, "localhost:" + ldapPort)
+                .put(ConfigConstants.LDAP_AUTHC_USERSEARCH, "(uid={0})")
+                .put(ConfigConstants.FOLLOW_REFERRALS, false).build();
+
+
+        final Connection con = LDAPAuthorizationBackend.getConnection(settings, null);
+        try {
+            //If following is off then should fail to return the result provided by following
+            final LdapEntry ref1 = LdapHelper.lookup(con, "cn=Ref1,ou=people,o=TEST", ReturnAttributes.ALL.value(), settings.getAsBoolean(ConfigConstants.FOLLOW_REFERRALS, true));
+            Assert.assertEquals("cn=couldNotFollowReferral", ref1.getDn());
```
peternied, line 366: It looks like this test is a mirror of the test above it, lets just check that the expected user isn't found because they are referred.

src/test/java/com/amazon/dlic/auth/ldap2/LdapBackendTestNewStyleConfig2.java
```diff
@@ -402,14 +402,34 @@ public void testLdapAuthenticationReferral() throws Exception {
...
+        final Settings settings = Settings.builder()
+                .putList(ConfigConstants.LDAP_HOSTS, "localhost:" + ldapPort)
+                .put(ConfigConstants.LDAP_AUTHC_USERSEARCH, "(uid={0})")
+                .put(ConfigConstants.FOLLOW_REFERRALS, false).build();
+
+
+        final Connection con = LDAPAuthorizationBackend.getConnection(settings, null);
+        try {
+            //If following is off then should fail to return the result provided by following
+            final LdapEntry ref1 = LdapHelper.lookup(con, "cn=Ref1,ou=people,o=TEST", ReturnAttributes.ALL.value(), settings.getAsBoolean(ConfigConstants.FOLLOW_REFERRALS, true));
```
peternied, line 426:
```suggestion
            final LdapEntry ref1 = LdapHelper.lookup(con, "cn=Ref1,ou=people,o=TEST", ReturnAttributes.ALL.value(), /* follow referrals */ false);
```

This should not be loaded from a setting, instead hardcode the API call

peternied (changes requested):
Looks like a giant binary file `OPENSEARCH_LDAP_HOME` was included, please remove that file

scrawfor99:
The automated tests seem to be frozen on Github, so I am going to wait and check back in on this in a little. (I tried pushing again to no avail)

src/main/java/com/amazon/dlic/auth/ldap/util/ConfigConstants.java
```diff
@@ -35,6 +35,9 @@ public final class ConfigConstants {
     public static final String LDAP_AUTHZ_MAX_NESTED_DEPTH = "max_nested_depth";
     public static final int LDAP_AUTHZ_MAX_NESTED_DEPTH_DEFAULT = 30;
 
+    public static final String FOLLOW_REFERRALS = "follow_referrals";
+    public static final boolean FOLLOW_REFERRALS_DEFAULT = true;
```
DarshitChanpura, line 39: Should we default it to `true`?

src/main/java/com/amazon/dlic/auth/ldap/util/LdapHelper.java
```diff
@@ -40,30 +40,40 @@ public class LdapHelper {
...
+                         // referrals will be followed to build the response
+                        request.setReferralHandler(new SearchReferralHandler());
+                    }
+
                     final Response<SearchResult> r = search.execute(request);
                     final org.ldaptive.SearchResult result = r.getResult();
                     entries.addAll(result.getEntries());
+
+                    if (entries == null) {
+                        //Throw an exception if no entries can be found without referrals and referrals are off
```
DarshitChanpura, line 73: 👍 for documentation

src/main/java/com/amazon/dlic/auth/ldap2/LDAPAuthenticationBackend2.java
```diff
@@ -78,6 +79,7 @@ public LDAPAuthenticationBackend2(final Settings settings, final Path configPath
 
         this.userSearcher = new LDAPUserSearcher(settings);
         this.returnAttributes = settings.getAsList(ConfigConstants.LDAP_RETURN_ATTRIBUTES, Arrays.asList(ReturnAttributes.ALL.value())).toArray(new String[0]);
+        this.shouldFollowReferrals = settings.getAsBoolean(ConfigConstants.FOLLOW_REFERRALS, true);
```
DarshitChanpura, line 82: instead of `true` here  as default we should use `ConfigConstants.FOLLOW_REFERRALS_DEFAULT`

src/main/java/com/amazon/dlic/auth/ldap2/LDAPAuthorizationBackend2.java
```diff
@@ -91,6 +92,7 @@ public LDAPAuthorizationBackend2(final Settings settings, final Path configPath)
         this.connectionFactory = ldapConnectionFactoryFactory.createConnectionFactory(this.connectionPool);
         this.userSearcher = new LDAPUserSearcher(settings);
         this.returnAttributes = settings.getAsList(ConfigConstants.LDAP_RETURN_ATTRIBUTES, Arrays.asList(ReturnAttributes.ALL.value())).toArray(new String[0]);
+        this.shouldFollowReferrals = settings.getAsBoolean(ConfigConstants.FOLLOW_REFERRALS, true);
```
DarshitChanpura, line 95: instead of `true` here as default we should use `ConfigConstants.FOLLOW_REFERRALS_DEFAULT`

src/main/java/com/amazon/dlic/auth/ldap2/LDAPAuthenticationBackend2.java
scrawfor99: fixed everywhere

DarshitChanpura (approved):

src/main/java/com/amazon/dlic/auth/ldap/util/LdapHelper.java
```diff
@@ -40,30 +40,40 @@ public class LdapHelper {
...
+                    if (shouldFollowReferrals) {
+                         // referrals will be followed to build the response
+                        request.setReferralHandler(new SearchReferralHandler());
+                    }
+
                     final Response<SearchResult> r = search.execute(request);
                     final org.ldaptive.SearchResult result = r.getResult();
                     entries.addAll(result.getEntries());
+
+                    if (entries == null) {
```
peternied, line 72: This cannot be null, its value is set at 55, this check/exception should be removed

peternied (changes requested):

src/main/java/com/amazon/dlic/auth/ldap/util/LdapHelper.java
scrawfor99: Fixed!

peternied (approved):

DarshitChanpura (approved):

scrawfor99:
@peternied  and @DarshitChanpura  would you be able to look at this again? I think that this should be good to go when the tests are fixed.

DarshitChanpura (approved):

peternied:
@scrawfor99 Please rebase against origin/main so we can see a clean CI run, thanks!

scrawfor99:
Can this be merged?

PR #2136, opened 2022-10-03
Release notes 1.3.6

vinayak15:
Signed-off-by: Vinayak Kothari <vkothar@amazon.com>

### Description
Release Notes 1.3.6
Documentation
*

### Check List

### DIFF
--- a/release-notes/opensearch-security.release-notes-1.3.6.0.md
+++ b/release-notes/opensearch-security.release-notes-1.3.6.0.md
@@ -1,0 +1,11 @@
+## 2022-10-06 Version 1.3.6.0
+
+### Bug fixes
+
+* Scope updateVersion to only build.gradle ([#2121](https://github.com/opensearch-project/security/pull/2121))
+
+### Maintenance
+
+* Updates jackson version to 2.13.4 ([#2128](https://github.com/opensearch-project/security/pull/2128))
+* Update Kafka Client to 3.0.2 ([#2129](https://github.com/opensearch-project/security/pull/2129))
+* Staging for version increment automation  ([#2029](https://github.com/opensearch-project/security/pull/2029))


release-notes/opensearch-security.release-notes-1.3.6.0.md
```diff
@@ -0,0 +1,10 @@
+## 2022-10-03 Version 1.3.6.0
+
+### Bug fixes
+* Update Kafka Client to 3.0.2 ([#2129](https://github.com/opensearch-project/security/pull/2129))
```
cwperks, line 4: Nit: This should be in Maintenance

release-notes/opensearch-security.release-notes-1.3.6.0.md
```diff
@@ -0,0 +1,10 @@
...
+* Scope updateVersion to only build.gradle ([#2121](https://github.com/opensearch-project/security/pull/2121))
+* Staging for version increment automation ([#2029](https://github.com/opensearch-project/security/pull/2029))
```
cwperks, line 6: This could probably go in Maintenance too

release-notes/opensearch-security.release-notes-1.3.6.0.md
```diff
@@ -0,0 +1,10 @@
+## 2022-10-03 Version 1.3.6.0
```
cwperks, line 1: 2022-10-06 - The day of the release

release-notes/opensearch-security.release-notes-1.3.6.0.md
```diff
@@ -0,0 +1,10 @@
...
+
+### Bug fixes
+* Update Kafka Client to 3.0.2 ([#2129](https://github.com/opensearch-project/security/pull/2129))
+* Scope updateVersion to only build.gradle ([#2121](https://github.com/opensearch-project/security/pull/2121))
+* Staging for version increment automation ([#2029](https://github.com/opensearch-project/security/pull/2029))
+
+### Maintenance
+
+* Updates jackson version to be in sync with core ([#2128](https://github.com/opensearch-project/security/pull/2128))
```
cwperks, line 10: Let's change this to `Updates jackson version to 2.13.4

cwperks (changes requested):

release-notes/opensearch-security.release-notes-1.3.6.0.md
DarshitChanpura: I think we should leave it as is. If someone is interested in the version number they can check core.

cwperks: Ideally, we should include the version of the dependency that we update to in a release note on a dependency update. In this case, all impacted plug-ins are going to have something similar to this in their release notes so either should be ok.

release-notes/opensearch-security.release-notes-1.3.6.0.md
vinayak15: Okay I will change this

release-notes/opensearch-security.release-notes-1.3.6.0.md
cwperks: Hey @vinayak15 we have some discretion when it comes to categorizing the issues, but generally speaking security patches will be categorized into Maintenance since that is ongoing maintenance we perform on the plugin and is not a bug in the security plugin itself.

The second one is a bugfix for version increment automation where it wasn't scoped correctly.

The third one could be argued as maintenance or enhancement IMO. I would moreso argue for maintenance because its not an enhancement to the security plugin, rather it helps us automate the version increment process between releases. 

You can read more about plugin release notes here: https://github.com/opensearch-project/opensearch-plugins/blob/main/RELEASE_NOTES.md

cwperks (approved):
Thank you @vinayak15!

cliu123 (approved):

Issue #2159, opened 2022-10-12
[BUG] hasValidDNs is not handling the rdn order change of subject and issuerDN.


shubhmkr-amazon:
**What is the bug?**
After changing the order of subjectDN field of the new certificate giving a cert validation error response(500).  As it is comparing the whole DN field as a whole String.
Current cert Subject: C=DE, L=Test, O=Test, OU=SSL, CN=node-1.example.com
New cert  Subject: CN=node-1.example.com, OU=SSL, C=DE, L=Test, O=Test

**How can one reproduce the bug?**
Steps to reproduce the behaviour:
1. Run unit testcase 'testSSLReloadCerts_Pass_NonSSLOnly' with attached certificates.
2. Check the result it is giving error response(500) instead of 200.

**What is the expected behaviour?**
From my point of view new certificate should be successfully validated. (Response code 200).

According to rfc document:-  https://www.rfc-editor.org/rfc/rfc5280#section-7.1

```
Two relative distinguished names RDN1 and RDN2 match if
they have the same number of naming attributes
and for each naming attribute in RDN1 there is a matching naming
attribute in RDN2. Two distinguished names DN1 and DN2 match if they
have the same number of RDNs, for each RDN in DN1 there is a matching
RDN in DN2, and the matching RDNs appear in the same order in both DNs.

```
This section of text from rfc document is confusing and asked on multiple platform.  From going through them I found two interpretation:-

1. Order of RDN in the DN matters As the RDNSequence(DN)  is sequence of RDN and I think sequence itself mean ordered and for RDN its mentioned as set which means order does not matter.  ([rfc section:- 4.1.2.4](https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.4)).

2. Two DNs have to be considered equal even if they differ in the order of naming attributes in some matching relative distinguished name.

Some of the discussion related to this topic:-
https://stackoverflow.com/questions/33080136/ordering-of-rdn-attributes-in-a-cas-certification-sign-process
https://security.stackexchange.com/questions/93985/openssl-x509-is-there-an-order-in-subjectname-fields
https://web.archive.org/web/20180328032619/http://openssl.6102.n7.nabble.com/Is-ordering-of-distinguished-names-for-subject-and-issuer-in-OpenSSl-0-9-8-certificates-important-td43602.html
https://frasertweedale.github.io/blog-redhat/posts/2019-05-28-a-dn-is-not-a-string.html

Let me explain my understanding with example
Example 1.
[...]

PR #2170, opened 2022-10-17
Modifying Validation logic to compare DN value irrespective of RDNs order.

shubhmkr-amazon:
### Description
While validating DN field during hot reload compare RDN value irrespective of their order.

* Category - Bug Fix
Current logic for DN validation is giving error when the order of RDN are not same.

* What is the old behavior before changes and new behavior after changes? Old behavior- Comparing DN field by converting it to String which was giving error if the RDNs are in different order.
{"error":"OpenSearchSecurityException[Error while initializing transport SSL layer from PEM: java.lang.Exception: New Certs do not have valid Issue    r DN, Subject DN or SAN.]; nested: Exception[New Certs do not have valid Issuer DN, Subject DN or SAN..];"}

New behavior- Comparing DN field irrespective of the RDNs order.

### Related Issue
hasValidDNs is not handling the rdn order change of subject and issuerDN [#2159](https://github.com/opensearch-project/security/issues/2159)

### Testing
Tested on current test case. And also tested on local by generating self signed certificates.

### Check List

Signed-off-by: Shubham Kumar <shubhmkr@amazon.com>

### DIFF
--- a/src/main/java/org/opensearch/security/ssl/DefaultSecurityKeyStore.java
+++ b/src/main/java/org/opensearch/security/ssl/DefaultSecurityKeyStore.java
@@ -73,6 +73,7 @@ import org.bouncycastle.asn1.ASN1Sequence;
 import org.bouncycastle.asn1.ASN1String;
 import org.bouncycastle.asn1.ASN1TaggedObject;
 
+import org.bouncycastle.asn1.x500.X500Name;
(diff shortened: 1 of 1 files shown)


peternied (changes requested):
Thanks for the contribution, could you please add some tests that would verify the behavior works as expected?

PR #2175, opened 2022-10-18
[DRAFT] 2.4 Windows Github Action

scrawfor99:
Basic setup for creating a GitHub Action to run test on push, merge, etc. for Windows.

This is a draft PR and there are several things about it that require further work. Namely: 
- It is not clear how the installation files are fetched when Docker is not used (without Docker being used, cannot call Docker hub)
- The plugin installation process is not successful and the current version makes use of the broken Windows core version where security does not function. 
- The final integration tests are built with Docker in mind so this requires migration. 

### Testing

So far the following functionality has been manually checked on a Windows instance:
1. Enable longer filenames
2. Gradle assemble security 
3. Hardcoded 2.3.0 Snapshot pull (command prompt)
4. Zip extraction and old zip deletion --> The extracted file organization is that of a snapshot so is built for Docker
5. Starting .\bin\opensearch.bat (powershell) --> This currently fails due to bad security plugin install 
6. Curl get https://localhost:9200 -u 'admin:admin' -k -v (command prompt) 
The intermediate steps have not been confirmed to work yet.
### Check List


peternied:
@scrawfor99 It looks like the DCO wasn't signed on one of those commits, could you amend the commit and force push to your branch?

Note; it might be a good time to rebase while you are at it.

cwperks:
> 5. Starting .\bin\opensearch.bat (powershell) --> This currently fails due to bad security plugin install

What error message are you receiving? The available windows distribution does not currently contain the .bat file we're seeking to add with this PR and associated PR that adds the bat file. For development of this action you will need to add the file manually in `plugins\opensearch-security\tools` so its available to be executed.

scrawfor99:
> @scrawfor99 It looks like the DCO wasn't signed on one of those commits, could you amend the commit and force push to your branch?
> 
> Note; it might be a good time to rebase while you are at it.
Took care of this. I must think I accidentally mixed something up before.

scrawfor99:
> > 5. Starting .\bin\opensearch.bat (powershell) --> This currently fails due to bad security plugin install
> 
> What error message are you receiving? The available windows distribution does not currently contain the .bat file we're seeking to add with this PR and associated PR that adds the bat file. For development of this action you will need to add the file manually in `plugins\opensearch-security\tools` so its available to be executed.



> > 5. Starting .\bin\opensearch.bat (powershell) --> This currently fails due to bad security plugin install
> 
> What error message are you receiving? The available windows distribution does not currently contain the .bat file we're seeking to add with this PR and associated PR that adds the bat file. For development of this action you will need to add the file manually in `plugins\opensearch-security\tools` so its available to be executed.

So the original action involves building with gradle but the snapshot that is downloaded from the Windows-team link comes prebuilt so I think the solution is going to be building the security plugin and then move it into the core build instead of trying to build both. I am not sure if this is what you meant but it seems like the best option.

PR #2178, opened 2022-10-19
Cross cluster search integration tests

lukasz-soszynski-eliatra:
### Description
* Category (Integration tests)
* What is the old behaviour before changes and new behaviour after changes?

Integration tests related to the cross-cluster search feature created

### Issues Resolved

### Testing

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/CrossClusterSearchTests.java
+++ b/src/integrationTest/java/org/opensearch/security/CrossClusterSearchTests.java
@@ -1,0 +1,435 @@
+/*
+* Copyright OpenSearch Contributors
+* SPDX-License-Identifier: Apache-2.0
+*
+* The OpenSearch Contributors require contributions made to
+* this file be licensed under the Apache-2.0 license or a
+* compatible open source license.
+*
+*/
+package org.opensearch.security;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+import com.carrotsearch.randomizedtesting.annotations.ParametersFactory;
+import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
+import org.apache.commons.lang3.tuple.Pair;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.action.search.SearchResponse;
+import org.opensearch.client.Client;
+import org.opensearch.client.RestHighLevelClient;
+import org.opensearch.test.framework.TestSecurityConfig.Role;
+import org.opensearch.test.framework.TestSecurityConfig.User;
+import org.opensearch.test.framework.certificate.TestCertificates;
+import org.opensearch.test.framework.cluster.ClusterManager;
+import org.opensearch.test.framework.cluster.LocalCluster;
+import org.opensearch.test.framework.cluster.SearchRequestFactory;
+import org.opensearch.test.framework.cluster.TestRestClient;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.opensearch.action.support.WriteRequest.RefreshPolicy.IMMEDIATE;
+import static org.opensearch.client.RequestOptions.DEFAULT;
+import static org.opensearch.rest.RestStatus.FORBIDDEN;
+import static org.opensearch.security.Song.ARTIST_FIRST;
+import static org.opensearch.security.Song.FIELD_ARTIST;
+import static org.opensearch.security.Song.FIELD_GENRE;
+import static org.opensearch.security.Song.FIELD_LYRICS;
+import static org.opensearch.security.Song.FIELD_STARS;
+import static org.opensearch.security.Song.FIELD_TITLE;
+import static org.opensearch.security.Song.GENRE_JAZZ;
+import static org.opensearch.security.Song.GENRE_ROCK;
+import static org.opensearch.security.Song.QUERY_TITLE_MAGNUM_OPUS;
+import static org.opensearch.security.Song.SONGS;
+import static org.opensearch.security.Song.TITLE_MAGNUM_OPUS;
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+import static org.opensearch.test.framework.TestSecurityConfig.Role.ALL_ACCESS;
+import static org.opensearch.test.framework.cluster.SearchRequestFactory.queryStringQueryRequest;
+import static org.opensearch.test.framework.matcher.ExceptionMatcherAssert.assertThatThrownBy;
+import static org.opensearch.test.framework.matcher.OpenSearchExceptionMatchers.statusException;
+import static org.opensearch.test.framework.matcher.SearchResponseMatchers.isSuccessfulSearchResponse;
+import static org.opensearch.test.framework.matcher.SearchResponseMatchers.numberOfTotalHitsIsEqualTo;
+import static org.opensearch.test.framework.matcher.SearchResponseMatchers.searchHitContainsFieldWithValue;
+import static org.opensearch.test.framework.matcher.SearchResponseMatchers.searchHitDoesNotContainField;
+import static org.opensearch.test.framework.matcher.SearchResponseMatchers.searchHitsContainDocumentWithId;
+import static org.opensearch.test.framework.matcher.SearchResponseMatchers.searchHitsContainDocumentsInAnyOrder;
+
+/**
+* This is a parameterized test so that one test class is used to test security plugin behaviour when <code>ccsMinimizeRoundtrips</code>
+* option is enabled or disabled. Method {@link #parameters()} is a source of parameters values.
+*/
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class CrossClusterSearchTests {
+
+	private static final String SONG_INDEX_NAME = "song_lyrics";
+
+	private static final String PROHIBITED_SONG_INDEX_NAME = "prohibited_song_lyrics";
+
+	public static final String REMOTE_CLUSTER_NAME = "ccsRemote";
+	public static final String REMOTE_SONG_INDEX = REMOTE_CLUSTER_NAME + ":" + SONG_INDEX_NAME;
+
+	public static final String SONG_ID_1R = "remote-00001";
+	public static final String SONG_ID_2L = "local-00002";
+	public static final String SONG_ID_3R = "remote-00003";
+	public static final String SONG_ID_4L = "local-00004";
+	public static final String SONG_ID_5R = "remote-00005";
+	public static final String SONG_ID_6R = "remote-00006";
+
+	private static final Role LIMITED_ROLE = new Role("limited_role")
+		.indexPermissions("indices:data/read/search", "indices:admin/shards/search_shards")
+		.on(SONG_INDEX_NAME, "user-${user.name}-${attr.internal.type}");
+
+	private static final Role DLS_ROLE_ROCK = new Role("dls_role_rock")
+		.indexPermissions("indices:data/read/search", "indices:data/read/get", "indices:admin/shards/search_shards")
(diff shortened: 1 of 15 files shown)


src/integrationTest/java/org/opensearch/security/Song.java
```diff
@@ -26,19 +30,34 @@ class Song {
...
+
+	static final String GENRE_ROCK = "rock";
+	static final String GENRE_JAZZ = "jazz";
+	static final String GENRE_BLUES = "blues";
 
 	static final String QUERY_TITLE_NEXT_SONG = FIELD_TITLE + ":" + "\"" + TITLE_NEXT_SONG + "\"";
 	static final String QUERY_TITLE_POISON = FIELD_TITLE + ":" + TITLE_POISON;
 	static final String QUERY_TITLE_MAGNUM_OPUS = FIELD_TITLE + ":" + TITLE_MAGNUM_OPUS;
 
 	static final Object[][] SONGS = {
```
peternied, line 55: This is complex and a small error could blow up many tests for an unknown reason, what do you think about introducing a type representing a song scheme to ensure consistent construction?

src/integrationTest/java/org/opensearch/test/framework/cluster/ClusterManager.java
```diff
@@ -52,15 +52,17 @@
 public enum ClusterManager {
 
 	//3 nodes (1m, 2d)
-	DEFAULT(new NodeSettings(true, false), new NodeSettings(false, true), new NodeSettings(false, true)),
+	DEFAULT(new NodeSettings(true, false, false), new NodeSettings(false, true, false), new NodeSettings(false, true, false)),
 
 	//1 node (1md)
-	SINGLENODE(new NodeSettings(true, true)),
+	SINGLENODE(new NodeSettings(true, true, false)),
```
peternied, line 58: Interpreting these configurations is hard, what do you think of having an enum for NodeRoles with values like, ClusterManager, Data, RemoteCluster, and then just accepting a list for the node settings.  

E.g.
```java
SINGLENODE(List.of(CLUSTER_MANAGER, REMOTE_CLUSTER)),
```

src/integrationTest/java/org/opensearch/test/framework/cluster/SearchRequestFactory.java
```diff
@@ -26,14 +26,21 @@ public final class SearchRequestFactory {
 	private SearchRequestFactory() {
 
 	}
-	public static SearchRequest queryStringQueryRequest(String indexName, String queryString) {
+	public static SearchRequest queryStringQueryRequest(String indexName, String queryString, Boolean ccsMinimizeRoundtrips) {
```
peternied, line 29: The SearchRequest is mutable, these additional overloads seem like they are going to be more maintenance work as additional scenarios are added over the long run.  What would you think about creating a 'builder' function like `minimizeRoundTrips(...)` that can be used inline to add these parameter?

```java
public static SearchRequest minimizeRoundTrips(SearchRequest sr) {
   sr.setCcsMinimizeRoundtrips(true);
   return sr;
}
```

src/integrationTest/java/org/opensearch/security/CrossClusterSearchTests.java
```diff
@@ -0,0 +1,428 @@
...
+		.nodeSettings(Map.of(PLUGINS_SECURITY_RESTAPI_ROLES_ENABLED, List.of("user_" + ADMIN_USER.getName()  +"__" + ALL_ACCESS.getName())))
+		.remote(REMOTE_CLUSTER_NAME, remoteCluster)
+		.roles(LIMITED_ROLE, DLS_ROLE_ROCK, DLS_ROLE_JAZZ, FLS_EXCLUDE_LYRICS_ROLE, FLS_INCLUDE_TITLE_ROLE).authc(AUTHC_HTTPBASIC_INTERNAL)
+		.users(ADMIN_USER, LIMITED_USER, DLS_USER_ROCK, DLS_USER_JAZZ, FLS_INCLUDE_TITLE_USER, FLS_EXCLUDE_LYRICS_USER)
+		.build();
+
+	@ParametersFactory(shuffle = false)
+	public static Iterable<Object[]> parameters() {
+		return List.of(new Object[] { true }, new Object[] { false });
+	}
```
peternied, line 149: Do these create any complexity in debugging that you ran into while authoring?  Just doublechecking if we should update any documentation or consider another design

lukasz-soszynski-eliatra: During debugging, the test method is executed twice for each parameter value. If this is inconvenient then it is easy to temporarily remove one parameter value.

I also considered creating an abstract class which contains all test methods and a constructor which accepts the parameter `ccsMinimizeRoundtrips` value. Then two subclasses are used to just invoke the superclass constructor with the proper value of the parameter `ccsMinimizeRoundtrips`. But I think that debugging is easier when the approach with parameterized tests is used.

src/integrationTest/java/org/opensearch/security/CrossClusterSearchTests.java
```diff
@@ -0,0 +1,428 @@
...
+			client.assignRoleToUser(DLS_USER_ROCK.getName(), DLS_ROLE_ROCK.getName()).assertStatusCode(200);
+			client.assignRoleToUser(DLS_USER_JAZZ.getName(), DLS_ROLE_JAZZ.getName()).assertStatusCode(200);
+			client.assignRoleToUser(FLS_INCLUDE_TITLE_USER.getName(), FLS_INCLUDE_TITLE_ROLE.getName()).assertStatusCode(200);
+			client.assignRoleToUser(FLS_EXCLUDE_LYRICS_USER.getName(), FLS_EXCLUDE_LYRICS_ROLE.getName()).assertStatusCode(200);
+		}
+	}
+
+	@Test
+	public void shouldFindDocumentOnRemoteCluster_positive() throws IOException {
+		try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(LIMITED_USER)) {
```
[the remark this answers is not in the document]
lukasz-soszynski-eliatra: I am not sure if I correctly understand your comment. The local and remote cluster contains distinct document sets. The ids of documents on the local cluster contain the word "local" whereas on the remote cluster word "remote" is used in document ids. Therefore when the document's id value is checked during the test then the test verifies if the document was loaded from the local or remote cluster.

src/integrationTest/java/org/opensearch/security/Song.java
lukasz-soszynski-eliatra: You are right. This data structure is getting more and more complex. I will create a DTO class which contains fields like artist, title, etc. and will serialize the class to JSON with a Jackson library.

src/integrationTest/java/org/opensearch/security/CrossClusterSearchTests.java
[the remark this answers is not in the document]
peternied: You have already taken this into account, great idea forcing the IDs to be deterministic.  Works for me.

src/integrationTest/java/org/opensearch/security/Song.java
lukasz-soszynski-eliatra: Finally I manage to do this without Jackson.

src/integrationTest/java/org/opensearch/test/framework/cluster/SearchRequestFactory.java
lukasz-soszynski-eliatra: done

src/integrationTest/java/org/opensearch/test/framework/cluster/ClusterManager.java
lukasz-soszynski-eliatra: done

peternied (approved):

src/integrationTest/java/org/opensearch/test/framework/matcher/SearchHitContainsFieldWithValueMatcher.java
```diff
@@ -49,7 +49,7 @@
 			mismatchDescription.appendText("Source document is null, is fetch source option set to true?");
 			return false;
 		}
-		if(!source.containsKey(fieldName)) {
+		if(source.containsKey(fieldName) == false) {
```
DarshitChanpura, line 52: nit: This change should not be needed

src/integrationTest/java/org/opensearch/test/framework/cluster/TestRestClient.java
```diff
@@ -167,6 +168,13 @@ public HttpResponse patch(String path, String body) {
 		return executeRequest(uriRequest, CONTENT_TYPE_JSON);
 	}
 
+	public HttpResponse assignRoleToUser(String username, String roleName) {
+		Objects.requireNonNull(roleName, "Role name is required");
+		Objects.requireNonNull(username, "User name is required");
+		String body = String.format("[{\"op\":\"add\",\"path\":\"/opendistro_security_roles\",\"value\":[\"%s\"]}]", roleName);
```
DarshitChanpura, line 174: Shouldn't this contain `opensearch_security_roles`?

DarshitChanpura (changes requested):
Great work @lukasz-soszynski-eliatra ! I have request a small change.

src/integrationTest/java/org/opensearch/test/framework/matcher/SearchHitContainsFieldWithValueMatcher.java
lukasz-soszynski-eliatra: Please correct me if I am wrong, but I thought that some documents define the convention to use expressions like `source.containsKey(fieldName) == false` instead of `!source.containsKey(fieldName)`

src/integrationTest/java/org/opensearch/test/framework/cluster/TestRestClient.java
lukasz-soszynski-eliatra: Sorry, I did not understand your comment :( Could you elaborate, please?

src/integrationTest/java/org/opensearch/test/framework/matcher/SearchHitContainsFieldWithValueMatcher.java
DarshitChanpura: interesting...I'm not aware about such documents.. If you have one handy could you please link it here

src/integrationTest/java/org/opensearch/test/framework/cluster/TestRestClient.java
DarshitChanpura: My bad, I meant that since we are no longer using `opendistro` convention should this initialization not contain `opendistro` and use `opensearch` instead. Let me know if I need to elaborate more

src/integrationTest/java/org/opensearch/test/framework/matcher/SearchHitContainsFieldWithValueMatcher.java
lukasz-soszynski-eliatra: > There is an implicit rule that negative boolean expressions should use the form foo == false instead of !foo for better readability of the code. While this isn't strictly enforced, if might get called out in PR reviews as something to change.

Please see https://github.com/opensearch-project/OpenSearch/blob/main/DEVELOPER_GUIDE.md . 

But this file is placed in the main OpenSearch repository and not in the security plugin repository. Therefore I am not sure if the rule is applicable.

src/integrationTest/java/org/opensearch/test/framework/cluster/TestRestClient.java
lukasz-soszynski-eliatra: Actually, the older name is still used in API, probably due to backwards compatibility, please see https://opensearch.org/docs/1.2/security-plugin/access-control/api#patch-user

DarshitChanpura: You are correct. It is still the old name: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java#L109

DarshitChanpura (approved):

peternied (approved):

DarshitChanpura:
Bwc tests are broken on main. Tracking issue here: https://github.com/opensearch-project/security/issues/2221

PR #2198, opened 2022-10-26
Introduce null safe toString

mrk-andreev:
Some fields may be nullable. In this case toString() with fields contraction using + will throw npe. Use String.format instead.

### Description
* **Category**: Refactoring
* **Why these changes are required?** Fix possible NPE
* **What is the old behavior before changes and new behavior after changes?** Fix possible NPE


peternied:
I'm going to close this out, thanks for the contribution, but without the DCO on the commit we cannot merge this change.  Please push an update to your branch and comment on this pull request and we can see about merging.  Thanks!

Issue #2194, opened 2022-10-25
[Windows] Configuration values are improperly decoded


peternied:
**What is the bug?**
When users.yml is proccessed by DynamicSecurityConfig through the scenario tests the user strings are improperly treated as ANSI values.

Expected username `§ÄÖÜäöüß`, malformed username `Â§Ã„Ã–ÃœÃ¤Ã¶Ã¼ÃŸ`

**How can one reproduce the bug?**
On a windows machine run the following command
`./gradlew.bat test --tests org.opensearch.security.IntegrationTests.testSpecialUsernames` 

**Additional Context**
Its not yet clear if this is a scenario test issue or if this a problem with any configuration that is loaded normally.

Should attempt a manual repro on windows with an UTF-8 name like above in the configuration settings.

scrawfor99:
So I have some news on this issue: 

I took several steps to try to determine the most likely cause and believe I have narrowed it down. 

Right now, the integration test for the special usernames functions perfectly fine on-non special usernames for any internal users in the internal_users yaml file. This suggests that the issue is not a universal encoding error but instead an issue arising from the conversions of Windows-1252 into UTF-8 and vice versa. 

What complicates things is the location of this translation. When running Opensearch 2.4 with the security plugin, you can access the node information ONLY when you log in as one of the original internal user roles: admin, readall, kibanauser, etc. So while the security plugin successfully tests on users such as rexclude the Opensearch node responds 401 from a manual attempt -- likewise for any of the special names. 

Going deeper, it appears somewhere OpenSearch is populating from only the internal_users2.yml file instead of both. I am not sure where yet but I do know that this is likely related to at least one side of the problem. The next step will be to check whether moving a role from internal_users.yml to 2.yml makes it identifiable by OpenSearch on login and then further if this somehow helps with the special character issue.

scrawfor99:
I also do not know why we have two internal_users ymls and would reccomend we remove one.

peternied:
@scrawfor99  Setting aside the integrations tests cases, can you manually reproduce the issue with a `internal_users.yml` bootstrapped cluster?

Separately did you see that in base window OpenSearch (no security plugin), if you modify the config's `cluster.name` value to `§ÄÖÜäöüß`, when you call the information endpoint at `/` it comes back without being malformed, like below?  [1]

```
GET http://localhost:9200/
{
  "name" : "opensearch-node1-2-2-0",
  "cluster_name" : "§ÄÖÜäöüß",
[...]

PR #2206, opened 2022-10-29
Fix windows encoding issues

peternied:
### Description
Fix windows encoding issues

Adds spotbugs [1] to detect internationalization before they are added to the codebase, also fixed several encoding bugs that impact windows users.

[1] https://spotbugs.readthedocs.io/en/stable/index.html

### Issues Resolved
- Resolves https://github.com/opensearch-project/security/issues/2194

### Check List


.github/workflows/ci.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           -x spotlessCheck
           -x checkstyleMain
           -x checkstyleTest
+          -x spotbugsMain
```
scrawfor99, line 37: It looks like this workflow may be causing the plugin to not install properly? Is this just a lint error or something with this specific workflow?

spotbugs-include.xml
```diff
@@ -0,0 +1,5 @@
+<FindBugsFilter>
+    <Match>
+        <Bug category="I18N" />
+    </Match>
+</FindBugsFilter>
```
scrawfor99, line 5: Do you think this should be added as a separate PR or is fine to include it in this one even though it seems not directly related to the parsing issue--maybe it is and I am wrong?

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -419,7 +419,7 @@ public void logDocumentWritten(ShardId shardId, GetResult originalResult, Index
                     try (XContentParser parser = XContentHelper.createParser(NamedXContentRegistry.EMPTY, THROW_UNSUPPORTED_OPERATION, originalResult.internalSourceRef(), XContentType.JSON)) {
                         Object base64 = parser.map().values().iterator().next();
                         if(base64 instanceof String) {
-                            originalSource = (new String(BaseEncoding.base64().decode((String) base64)));
+                            originalSource = (new String(BaseEncoding.base64().decode((String) base64), StandardCharsets.UTF_8));
```
scrawfor99, line 422: Since we use this in numerous places and I believe there used to be a content type setting in a separate file, do we want to make a setting for the encoding method or does hard coding make more sense because we are setting it to the most broad option?

spotbugs-include.xml
cwperks: This is helpful to find instances where the code is relying on the default encoding. See https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html?highlight=i18n#dm-reliance-on-default-encoding-dm-default-encoding

cwperks (approved):

spotbugs-include.xml
peternied: I did not add more rules because we are in violation of _many_ different rules, but I didn't see any types that would create platform issues.

If you'd like to see the full list, delete line 3 and run the gradle task, you'll see hundreds of things!

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
peternied: When we were on unix environments the setting value was always UTF-8 both inside of security and OpenSearch so everything was happy.  Since OpenSearch widely forces the use of UTF-8 and security does in some spots this was the source of our issues.

.github/workflows/ci.yml
peternied: The broken workflow was because of the linelinter, fix now!

peternied:
@cwperks @DarshitChanpura Ready for review, thanks!

cwperks (approved):

peternied:
@opensearch-project/security Can I get another review on this please?

peternied:
Manual backports created:
- https://github.com/opensearch-project/security/pull/2217
- https://github.com/opensearch-project/security/pull/2218

Issue #2207, opened 2022-10-31
[FEATURE] Migrate Plugin install into an GitHub action


peternied:
**Is your feature request related to a problem?**
The plugin-install workflow user to separate paths to configure and run OpenSearch from the min build.  This would be really useful if it was reusable for this codebase as well as others.  Being that we are thinking more multiplatform this is also an opportunity where we should could invest in combining these workflows.

**What solution would you like?**
In the `plugin-install` workflow the there all the downloading, configuring, starting of OpenSearch would be replaced with a step like the following:

```yml
   - id: start-opensearch
     uses: uses: ./.github/actions/start-opensearch
     with:
         platform: ${{ matrix.platform }}
         opensearch-version: 3.0.0
         plugin-zip-path: ./security-dashboards-plugin.zip
         plugin-start-script: plugins/opensearch-security/tools/install_demo_configuration.sh
```

This would the be reusable for workflows within this Security codebase, but also for https://github.com/opensearch-project/security-dashboards-plugin/issues/1171 where this kind of workflow would simplify the extra work needed

**Do you have any additional context?**
Started a prototype with https://github.com/peternied/security/commit/1b880f4797c60ee4ba60598e548cbdd28c16426d

peternied:
FYI @scrawfor99 we talked about this offline to support the dashboard integration test workflows on linux/windows/mac

peternied:
Note; I think this action could be made more generic so it was usable by all plugin teams, but it might be good to start building it here before sharing it out to all the other plugins via opensearch-build or another repo

scrawfor99:
To start things off I have been working on creating a more streamlined workflow for the plugin install. My plan is to get it working on both JDK versions for all three OS' and then I will start splitting out steps in actions and files for drop-in functionality along the lines of the prototype you linked. For a preview checkout https://github.com/scrawfor99/security/tree/CombinedInstallWorkflows

scrawfor99:
Update 2: Still working on streamlining the workflow and adding macOS install. Because the GHA macOS runner does not come with docker (Linux does) we cannot simply reuse the Linux commands so I am having to write a new workflow similar to the Windows install.

peternied:
Maybe for now we just thrown an error if the OS is mac?  OpenSearch doesn't yet support the mac platform so there are no guarentees it will work or we could get support to get it running.

scrawfor99:
[...]

PR #2233, opened 2022-11-03
Add backwards compatibility tests for Windows and JDK11/17 

peternied:
### Description
Update the backward compatibility tests to run on windows in JDK11/17

### Issues Resolved
- Resolves https://github.com/opensearch-project/security/issues/2232
- Resolves https://github.com/opensearch-project/security/issues/1474

### Check List
- [ ] ~New functionality includes testing~
- [ ] ~New functionality has been documented~
- [X] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2239, opened 2022-11-04
GitHub Action for plugin install 

scrawfor99:
### Description
Converts the security plugin install into a git hub action action.yml process instead of a workflow. 

Partially resolves issue #2207 

### Testing
Completed check on 2.4 branch for windows and linux os with jdk 11 and 17. 

### Check List


.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,199 @@
...
+    required: true
+
+  plugin-name:
+    description: 'The the name of the plugin to use, such as security-dashboards-plugin'
+    required: true
+
+  plugin-start-script:
+    description: 'The file name for the configuration script for the plugin such as install_demo_configurations'
+    required: true
+  # plugins/opensearch-security/tools/install_demo_configuration.sh
```
peternied, line 20: Remove comment

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,199 @@
...
+
+  java-version:
```
peternied, line 22: Doesn't look used, remove?

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,199 @@
...
+    description: 'The java version to be used, must be either 11 or 17'
+    required: true
+
+  plugin-zip:
```
peternied, line 26: What do you think of making `plugin-zip` is the same as `plugin-name` so this parameter can be removed?

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,199 @@
...
+      shell: bash
+
+    - name: Download OpenSearch Core for Windows
+      if: ${{ inputs.platform == 'windows-latest' }}
+      run: |
+        Invoke-WebRequest https://artifacts.opensearch.org/snapshots/core/opensearch/${{ inputs.opensearch-version }}-SNAPSHOT/opensearch-min-${{ inputs.opensearch-version }}-SNAPSHOT-windows-x64-latest.zip -Outfile opensearch-${{ inputs.opensearch-version }}.zip
+        tar -xzf opensearch-${{ inputs.opensearch-version }}.zip
+        del opensearch-${{ inputs.opensearch-version }}.zip
+        Rename-Item opensearch-* opensearch
+      shell: pwsh
```
peternied, line 56: This can be combined into a single statement with `peternied/download-file` [1] rather than treating the platforms differently.  See also how this was done for the BWC tests [2] PR that is out for review

[1] https://github.com/peternied/download-file
[2] https://github.com/opensearch-project/security/pull/2233/files

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,175 @@
...
+
+    - name: Download OpenSearch Core for MacOS
+      if: ${{ inputs.platform == 'macos-latest' }}
```
peternied, line 59: Looks like this includes steps for mac, but we don't have mac artifacts?  Does it work or is this something that we should roll into another pull request?

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,175 @@
...
+        curl https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/${{ inputs.opensearch-version }}/latest/linux/x64/tar/builds/opensearch/dist/opensearch-min-${{ inputs.opensearch-version }}-linux-x64.tar.gz
+        tar -xzf opensearch-*.tar.gz
+        rm -f opensearch-*.tar.gz
+        mv opensearch-* opensearch
+      shell: bash
+
+    # Assemble the specified plugin
+    - name: Assemble the specified plugin for MacOS and Linux
+      if: ${{ inputs.platform != 'windows-latest' }}
+      run: ./gradlew clean assemble
```
peternied, line 70: I still recommend having the build step happen outside this action, but speaking of gradle, if you use `gradle/gradle-build-action@v2` its already multiplatform safe - did this in the CI workflow 

https://github.com/scrawfor99/security/blob/adc668f1a4e403c9ca4f96672166d6c498404c75/.github/workflows/ci.yml#L75-L79

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,175 @@
...
+    # Move and rename the plugin for installation
+    - name: Move and rename the plugin for installation on MacOS and Linux
+      if: ${{ inputs.platform != 'windows-latest' }}
+      run: mv build/distributions/${{ inputs.plugin-name }}-*.zip ${{ inputs.plugin-name }}.zip
+      shell: bash
+
+    - name: Move and rename the plugin for installation on Windows
+      if: ${{ inputs.platform == 'windows-latest' }}
+      run: move build\distributions\${{ inputs.plugin-name }}-*.zip .\${{ inputs.plugin-name }}.zip
+      shell: pwsh
```
peternied, line 89: I think we can reuse the same action for both platforms, we don't need to run on `pwsh` if `shell` works.

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,175 @@
...
+    - name: Sleep while OpenSearch finishes starting up
+      if: ${{ inputs.platform != 'windows-latest'}}
+      uses: whatnick/wait-action@v0.1.2
+      with:
+        time: '30s'
+
+    - name: Sleep while OpenSearch starts on Windows
+      if: ${{ inputs.platform == 'windows-latest'}}
+      run: Start-Sleep -s 30
+      shell: pwsh
```
peternied, line 159: Doesn't `whatnick/wait-action@v0.1.2` work on all platforms so we can condense this into one step?

peternied (changes requested):
Great work, lets try to condense some of these steps to have fewer platform specific code paths and fewer inputs, looks really close!

.github/actions/start-opensearch/action.yml
scrawfor99: We do not have mac support yet but since there are no calls for running the runner on Mac I figured it would be easier for whoever adds Mac support if the framework is already there.

.github/actions/start-opensearch/action.yml
scrawfor99: It does not, it only works on Linux.

.github/actions/start-opensearch/action.yml
scrawfor99: Oops. good catch

.github/actions/start-opensearch/action.yml
scrawfor99: That sounds good but right now the files on docker-host have different naming conventions so that would have to be changed first.

.github/actions/start-opensearch/action.yml
peternied: Please do not include code that is unused.  I've added some lines where folks have the following thoughts on commented-out code, and in this case, I think this code is basically following the same pattern even if it 'could be' active.  When its time to support the mac platform that will be a good time to revisit and update this code with testing.

- [1] https://kentcdodds.com/blog/please-dont-commit-commented-out-code
- [2 ]https://stackoverflow.com/questions/758279/checking-in-of-commented-out-code

.github/actions/start-opensearch/action.yml
scrawfor99: I am a bit hesitant to rely on actions hosted on our personal github accounts. Perhaps if it were published to the marketplace page https://github.com/marketplace?type=actions or if we were to add a package to opensearch which included a variety of different actions we wanted to reuse across the organization? I just do not know that it is best idea to set precedent for relying on actions that will break should the account no longer exist or which would allow for someone to effectively modify code without code review. I think this type of action reference would let someone (not saying they would) swap their own action code directly and then effectively change the repo's test code without any need for a PR. Since they are only changing code in their private repo I do not think a PR would be required and I think we would have to rely on people to properly enumerate their action versions in order to make the action changes require PR on the repo's side before taking affect. This is just my thoughts though and I could very well be wrong. 

 In the the meantime, I will swap the platform logic to simply install wget with chocolatey since that seems to be what you are suggesting with your own example.

peternied: > I am a bit hesitant to rely on actions hosted on our personal github accounts.

We absolutely should be suspicious of ANY GitHub Action, even those that look like official sources.  They are effectively running code on our behalf which could change at any time.

Good news! By using a version tag, it ensures that unless the repo is deleted that action will always be referenceable.

> In the the meantime, I will swap the platform logic to simply install wget with chocolatey since that seems to be what you are suggesting with your own example.

This is fine - but it sounds like very shortly we will be in the business of maintaining GitHub actions to do many different things, I'll start a discussion around this.  The Unix model of small special purpose [1] tools speaks to me, and I think we should optimize workflows like this one around the 'core' purpose and for everything else borrow.   


[1] https://www.brandur.org/small-sharp-tools

.github/actions/start-opensearch/action.yml
scrawfor99: Do you mind explaining why you do not want the build in the action? I am not sure I saw your initial thoughts on it sorry.

peternied: This simplifies the inputs for this workflow to roughly, "given a plugin zip file, setup and deploy an instance of OpenSearch with that plugin installed for testing".

Since we use different JDKs versions this is extra complexity we can have the caller of the workflow deal with, rather than integrating here.

.github/actions/start-opensearch/action.yml
scrawfor99: Alright, I will remove it.

.github/actions/start-opensearch/action.yml
scrawfor99: I definitely agree that using other pre-built actions is great when possible. It is easier for everyone and saves time and confusion. I just meant that maybe we would want to make sure that whatever actions we use are on the marketplace or merged into some to-be-made 'action' repo so that we did not have to worry about the possibility that the action code was changed without the versioning number being incremented and then the code we thought did one thing now did something else. Maybe that is not a real concern though. 

If you prefer I use the action you linked I can do that though--just let me know.

.github/actions/start-opensearch/action.yml
scrawfor99: Sounds good.

.github/actions/start-opensearch/action.yml
peternied: Seems like the docker start-up script is where this is happening, is there more to it?

.github/actions/start-opensearch/action.yml
peternied: Ask me if I'm biased without asking me if I'm biased :D - I do have a preference towards the action I made

Created an issue to talk about the logistic of shared actions.
- https://github.com/opensearch-project/.github/issues/105

.github/actions/start-opensearch/action.yml
peternied: Nit; If you are feeling adventurous https://github.com/peternied/action-sleep

.github/actions/start-opensearch/action.yml
scrawfor99: I am not sure what the docker startup script is--the first mention of anything docker related is at that line in the file.

.github/actions/start-opensearch/action.yml
scrawfor99: I am trying to get your action to work but haven been unable to figure out any logic other than simply calling it with an If condition for the different OSs. Perhaps you have a thought about this? But the snapshot files are named significantly differently between OSs and I have been trying for a time to get them to work any way.

.github/actions/start-opensearch/action.yml
scrawfor99: This seems like the code that was already in the file but I will put the action you just made in the spot if you think that is best.

.github/actions/start-opensearch/action.yml
scrawfor99: Hi @peternied , I tried for a while longer to make the code compatible with your download action testing changes on a private repo to avoid notifications (https://github.com/scrawfor99/security/tree/LoweredPermissionsInstall). That being said, I am still unable to get the changes to work without having at least as much OS-based logic as there already was. In fact, I only have Linux working now whereas both worked previously. What are your thoughts on either waiting to rebase the downloading logic until after this is merged or lending your knowledge on a good work around. I intended to build dashboard install platform for that team to use and then to supplement the security-dash-repo CI/CD process but would like to take care of this issue first.

What do you suggest and what would you like me to do?

scrawfor99:
@peternied this should be all set now.

.github/workflows/plugin_install.yml
```diff
@@ -3,120 +3,30 @@ name: Plugin Install
...
-          $Headers = @{ Authorization = $baseCredentials }
-          Invoke-WebRequest -SkipCertificateCheck -Uri 'https://localhost:9200' -Headers $Headers
+      - name: Run Opensearch with A Single Plugin
+        uses: ./.github/actions/start-opensearch
+        with:
+          platform: ${{ matrix.os }}
+          opensearch-version: 2.4.0
+          plugin-name: opensearch-security
+          plugin-start-script: install_demo_configuration
+          java-version: ${{ matrix.jdk }}
```
peternied, line 31: Remove unused parameter

.github/workflows/plugin_install.yml
```diff
@@ -3,120 +3,30 @@ name: Plugin Install
...
+          docker-file-name: security-plugin.zip
```
peternied, line 32: Remove unused parameter

.github/workflows/plugin_install.yml
```diff
@@ -3,120 +3,30 @@ name: Plugin Install
...
-
-      - name: Check OpenSearch Running
-        run: curl -XGET https://localhost:9200 -u 'admin:admin' -k -v
-
-      - name: Get Docker Logs
-        if: always()
-        run: docker logs ops
-
-      - name: Run sanity tests
-        run: ./gradlew integTestRemote -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="opensearch" -Dhttps=true -Duser=admin -Dpassword=admin
```
peternied, line 67: Can you add this step back at the end of this workflow?

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,128 @@
...
+    - name: Enable Longpaths if on Windows
+      if: ${{ inputs.platform == 'windows-latest' }}
+      run: git config --system core.longpaths true
+      shell: pwsh
+
+    # Download OpenSearch
+    - uses: peternied/download-file@v1
+      if: ${{ inputs.platform == 'windows-latest' }}
+      with:
+        url: https://artifacts.opensearch.org/snapshots/core/opensearch/${{ inputs.opensearch-version }}-SNAPSHOT/opensearch-min-${{ inputs.opensearch-version }}-SNAPSHOT-windows-x64-latest.zip
```
DarshitChanpura, line 39: This url is to be changed once windows artifacts are available via CI, correct?

to something like: 
```
https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/${{ inputs.opensearch-version }}/latest/windows/x64/zip/builds/opensearch/dist/opensearch-min-${{ inputs.opensearch-version }}-windows-x64-latest.zip
```

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,128 @@
...
+
+  plugin-name:
+    description: 'The the name of the plugin to use, such as security-dashboards-plugin'
+    required: true
+
+  plugin-start-script:
+    description: 'The file name for the configuration script for the plugin such as install_demo_configurations -- may not be needed for every plugin'
+    required: false
+
+  plugin-zip:
```
peternied, line 21: I think you need to add the `plugin-zip` parameter on the workflow

#### Doesn't look like the plugin actually installed
>  yes | opensearch-plugin install file:///**docker-host/.zip**
  chmod +x plugins/opensearch-security/tools/install_demo_configuration.sh
  yes | plugins/opensearch-security/tools/install_demo_configuration.sh
  chown 1001:1001 -R /opensearch
  su -c "/opensearch/bin/opensearch" -s /bin/bash opensearch
  EOF
  docker build -t opensearch-test:latest -f- . <<EOF
  FROM ubuntu:latest
  COPY --chown=1001:1001 os-ep.sh /docker-host/
  COPY --chown=1001:1001 opensearch-security.zip **/docker-host/.zip**
  COPY --chown=1001:1001 opensearch* /opensearch/
  RUN chmod +x /docker-host/os-ep.sh

_From https://github.com/opensearch-project/security/actions/runs/3429405272/jobs/5714978189#step:5:4480_

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,128 @@
...
+      shell: bash
+
+    - name: Check OpenSearch Running on Windows
+      if: ${{ inputs.platform == 'windows-latest'}}
+      run: |
+        $credentialBytes = [Text.Encoding]::ASCII.GetBytes("admin:admin")
+        $encodedCredentials = [Convert]::ToBase64String($credentialBytes)
+        $baseCredentials = "Basic $encodedCredentials"
+        $Headers = @{ Authorization = $baseCredentials }
+        Invoke-WebRequest -SkipCertificateCheck -Uri 'https://localhost:9200' -Headers $Headers;
```
peternied, line 127: Can we change both url to ~`https://admin:admin@localhost:9200/_cat/plugins`


#### `_cat/plugins` output

This is much better at confirming that plugin installed correctly

```
>curl https://admin:admin@localhost:9200/_cat/plugins -k
opensearch-node1-2-2-0 opensearch-alerting                  2.2.0.0
opensearch-node1-2-2-0 opensearch-anomaly-detection         2.2.0.0
opensearch-node1-2-2-0 opensearch-asynchronous-search       2.2.0.0
opensearch-node1-2-2-0 opensearch-cross-cluster-replication 2.2.0.0
opensearch-node1-2-2-0 opensearch-geospatial                2.2.0.0
opensearch-node1-2-2-0 opensearch-index-management          2.2.0.0
opensearch-node1-2-2-0 opensearch-job-scheduler             2.2.0.0
opensearch-node1-2-2-0 opensearch-knn                       2.2.0.0
opensearch-node1-2-2-0 opensearch-ml                        2.2.0.0
opensearch-node1-2-2-0 opensearch-notifications             2.2.0.0
opensearch-node1-2-2-0 opensearch-notifications-core        2.2.0.0
opensearch-node1-2-2-0 opensearch-observability             2.2.0.0
opensearch-node1-2-2-0 opensearch-performance-analyzer      2.2.0.0
opensearch-node1-2-2-0 opensearch-reports-scheduler         2.2.0.0
opensearch-node1-2-2-0 opensearch-security                  2.2.0.0
opensearch-node1-2-2-0 opensearch-sql                       2.2.0.0
```

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,128 @@
+name: 'Start OpenSearch'
+description: 'Downloads latest build of OpenSearch, installs a plugin, executes and script and then starts OpenSearch on localhost:9200'
+
+inputs:
+  platform:
```
peternied, line 5: Can we drop this parameter in favor of checks from the environment variables like `${{ runner.os != 'Windows'}}`?

> The operating system of the runner executing the job. Possible values are `Linux`, `Windows`, or `macOS`.

_From https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables_

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,128 @@
...
+    # Give the OpenSearch process some time to boot up before sending any requires, might need to increase the default time!
+    - name: Sleep while OpenSearch starts
+      uses: peternied/action-sleep@v1
+      with:
+        seconds: 30
+
+    # Verify that the server is operational
+    - name: Check OpenSearch Running on Linux and MacOS
+      if: ${{ inputs.platform != 'windows-latest'}}
+      run: curl https://localhost:9200 -u 'admin:admin' -k -v
```
peternied, line 117: Before the curl command in this block please add `docker logs ops` so we get debug information about the cluster starting up.

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,128 @@
...
+      shell: bash
+
+    - name: Check OpenSearch Running on Windows
+      if: ${{ inputs.platform == 'windows-latest'}}
+      run: |
```
peternied, line 122: Before the web request please cat the opensearch.log (or whatever its called) file for troubleshooting

peternied (changes requested):
Please add those tests back in, and I think there might be some issues with the workflow setting up the linux instance correctly.

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,128 @@
...
+        tar -xzf opensearch-*.tar.gz
+        rm -f opensearch-*.tar.gz
+      shell: bash
+
+    - name: Extract downloaded zip for Windows
+      if: ${{ inputs.platform == 'windows-latest' }}
+      run: |
+        tar -xzf opensearch-min-${{ inputs.opensearch-version }}-SNAPSHOT-windows-x64-latest.zip
+        del opensearch-min-${{ inputs.opensearch-version }}-SNAPSHOT-windows-x64-latest.zip
```
DarshitChanpura, line 58: What do you think of using `*` pattern as above, if you haven't tried already ? (i know PowerShell is a little painful)

.github/actions/start-opensearch/action.yml
scrawfor99: You mean to print the whole log file or to simply pipe the output of running opensearch into the file? I.e. do you mean cat the log or cat opensearch into the log?

.github/workflows/plugin_install.yml
DarshitChanpura: And `java-version` is not being used in the start-opensearch script so it is probly safe to remove it from here

.github/workflows/plugin_install.yml
```diff
@@ -3,120 +3,33 @@ name: Plugin Install
...
-          COPY --chown=1001:1001 opensearch* /opensearch/
-          RUN chmod +x /docker-host/os-ep.sh
-          RUN useradd -u 1001 -s /sbin/nologin opensearch
-          ENV PATH="/opensearch/bin:${PATH}"
-          WORKDIR /opensearch/
-          ENTRYPOINT /docker-host/os-ep.sh
-          EOF
-          docker run --name ops -d -p 9200:9200 -p 9600:9600 -i opensearch-test:latest
+      - name: Assemble target plugin
+        run: ./gradlew assemble
```
DarshitChanpura, line 22: this should have `clean` task too, to invoke a completely fresh build

.github/actions/start-opensearch/action.yml
scrawfor99: Yeah I tried, without specifying the whole path I keep getting errors about multiple inputs :/ . I can try again.

.github/workflows/plugin_install.yml
scrawfor99: That is what I thought too but I guess that the runners always start from clean states with no persistent data.

.github/actions/start-opensearch/action.yml
peternied: Print the whatever is in that log file so it can be read in the workflow log

.github/actions/start-opensearch/action.yml
scrawfor99: I think so. Right now this is the only path we have though. I got it from a github discussion a while back.

.github/workflows/plugin_install.yml
DarshitChanpura: oh okay.

peternied: Why do we need to invoke the clean task, when we get the github runners they are already a fresh environment?  When we use `clean` it deletes everything out of caches that are safe to reuse and slows down the job

.github/actions/start-opensearch/action.yml
DarshitChanpura: okay..we need to keep track of replace it once it becomes available

.github/actions/start-opensearch/action.yml
scrawfor99: Screen Shot 2022-11-09 at 11 59 04 AM
Foiled again :/

.github/actions/start-opensearch/action.yml
scrawfor99: Seems like the log file is empty? Id this expected? I don't think it is.

.github/actions/start-opensearch/action.yml
scrawfor99: Yeah I imagine once 2.4 is released we can swap.

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,127 @@
+name: 'Start OpenSearch'
```
cwperks, line 1: Thoughts on naming this: `Start OpenSearch with security`?

.github/actions/start-opensearch/action.yml
```diff
@@ -0,0 +1,127 @@
...
+inputs:
+  opensearch-version:
+    description: 'The version of OpenSearch that should be used, e.g "3.0.0"'
+    required: true
+
+  plugin-name:
+    description: 'The the name of the plugin to use, such as security-dashboards-plugin'
+    required: true
+
+  plugin-start-script:
```
cwperks, line 13: I know this is currently being used to run the `install_demo_configuration.sh` script, but I wonder if we can make it possible to specify settings to add to `opensearch.yml` before starting up the node. It would be useful to have a generic way of modifying the configuration before starting up the node to test different scenarios. @RyanL1997 is facing a similar problem while supplying custom configuration to `OpenSearch-Dashboards` before starting up which is needed for the aggregation view functional tests since the default configuration has that feature disabled.

.github/actions/start-opensearch/action.yml
scrawfor99: The current idea is for this to really be usable with __any__ plugin. That being said it is currently only tested on security so that should be fine.

.github/actions/start-opensearch/action.yml
scrawfor99: So you want to be able to specify a file name in opensearch.yml instead of specifying it in the workflow? That should be doable though I am not sure that the setting will really be used for anything but here.

.github/actions/start-opensearch/action.yml
peternied: I don't think `logs\gc.log.00` is correct, that is a garbage collection log, I think it should be named opensearch.log in that same folder

scrawfor99: Yeah that is what I thought too but the only file with any bytes in the directory is the gc file. When I try to cat opensearch.log I get an error because it is empty. What would you like me to do?

scrawfor99: Made an issue because the log file is empty: https://github.com/opensearch-project/security/issues/2251

.github/actions/start-opensearch/action.yml
scrawfor99: I created an issue so that we can consider adding this functionality after the fact: https://github.com/opensearch-project/security/issues/2252

scrawfor99:
@cwperks @peternied @DarshitChanpura This should be all set.

.github/workflows/plugin_install.yml
```diff
@@ -3,120 +3,32 @@ name: Plugin Install
...
-      - name: Check OpenSearch Running
-        run: curl -XGET https://localhost:9200 -u 'admin:admin' -k -v
-
-      - name: Get Docker Logs
-        if: always()
-        run: docker logs ops
-
-      - name: Run sanity tests
+      - name: Run sanity tests on Linux
+        if: ${{ matrix.os == 'ubuntu-latest' }}
```
peternied, line 33: Why not run these on Windows?

peternied: If its a matter of the command, you can use 

```yml
      uses: gradle/gradle-build-action@v2
      with:
        arguments: integTestRemote ...
```

peternied:
This pull request is pointed to the 2.4 branch, but we are past feature freeze.  Seems like we should point it to main, what do you think?

.github/workflows/plugin_install.yml
scrawfor99: To be honest, there was no reason... But all taken care of haha.

scrawfor99:
> This pull request is pointed to the 2.4 branch, but we are past feature freeze. Seems like we should point it to main, what do you think?

Main does not have a stable windows build is the reason this is on 2.4.

peternied:
If we can't get it working on main, then we should go into 2.X - do you have a link to the issue about windows being unavailable?

scrawfor99:
https://github.com/opensearch-project/opensearch-build/issues/2754 At the bottom of the page, it is mentioned that there was not a successful build for 3.0. This may have changed but I have been working on this for a while so I think when I started it was still not there. I can move this up to being based on 3 and see what happens but I am not sure how to confirm one-way or the other whether the main is not a viable base.

peternied:
I think you need to rebase your changes with something like `git rebase origin/main` and then force push to your branch.  Checkout the following guide if this is your first time https://www.digitalocean.com/community/tutorials/how-to-rebase-and-update-a-pull-request

scrawfor99:
I did and resolved all the conflicts. Not sure why it is not running the checks again but you can see it is up-to-date with main. I will try again.

peternied:
I don't think the rebase / force push worked.  If you look at the pull request there is 125 commits in this change.  Maybe squash and force push

peternied:
image

I don't see any changes?  Just in case I have found this useful too https://ohshitgit.com/

scrawfor99:
I cannot get it to work and now it closed itself.

cwperks:
If you push a change that removes all changes introduced in a PR it automatically closes the PR.

peternied:
Looks like your changes made it back, congrats on surviving a rebase related nightmare, looks like the plugin-install failed... I'll check back in when there are updates and its passing, so close!

scrawfor99:
image

peternied (approved):
Well done!

.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -0,0 +1,126 @@
+name: 'Start OpenSearch with One Plugin'
+description: 'Downloads latest build of OpenSearch, installs a plugin, executes and script and then starts OpenSearch on localhost:9200'
```
cwperks, line 2: nit: executes a script

.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -0,0 +1,126 @@
...
+
+inputs:
+  opensearch-version:
+    description: 'The version of OpenSearch that should be used, e.g "3.0.0"'
+    required: true
+
+  plugin-name:
+    description: 'The the name of the plugin to use, such as security-dashboards-plugin'
```
cwperks, line 10: nit: The appears twice in the beginning here

`security-dashboards-plugin` is a plugin for OpenSearch Dashboards, but this action looks like it starts an `opensearch` node. Should this reference an `opensearch` plugin?

.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -0,0 +1,126 @@
...
+
+    - name: Extract downloaded zip for Windows
+      if: ${{ runner.os == 'Windows' }}
+      run: |
+        tar -xzf opensearch-min-${{ inputs.opensearch-version }}-SNAPSHOT-windows-x64-latest.zip
+        del opensearch-min-${{ inputs.opensearch-version }}-SNAPSHOT-windows-x64-latest.zip
+      shell: pwsh
+
+    # Move and rename the plugin for installation
+    - name: Move and rename the plugin for installation
```
cwperks, line 58: I found this a bit confusing to follow. This action is intended to run after the checked out plugin code has been built? I'm wondering if we can specify a build command as a parameter and have this action also do the build. wdyt?

.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -0,0 +1,126 @@
...
+      if: ${{ runner.os == 'Linux'}}
+      run: |
+        cat > os-ep.sh <<EOF
+        yes | opensearch-plugin install file:///docker-host/${{ inputs.plugin-zip }}.zip
+        chmod +x plugins/${{ inputs.plugin-name }}/tools/${{ inputs.plugin-start-script }}.sh
+        yes | plugins/${{ inputs.plugin-name }}/tools/${{ inputs.plugin-start-script }}.sh
+        chown 1001:1001 -R /opensearch
+        su -c "/opensearch/bin/opensearch" -s /bin/bash opensearch
+        EOF
+        docker build -t opensearch-test:latest -f- . <<EOF
```
cwperks, line 73: Curious, is tagging this image with `opensearch-test:latest` needed?

.github/workflows/plugin_install.yml
```diff
@@ -3,120 +3,33 @@ name: Plugin Install
...
-        run: Start-Sleep -s 30
-
-      - name: Check OpenSearch Running
-        run: |
-          $credentialBytes = [Text.Encoding]::ASCII.GetBytes("admin:admin")
-          $encodedCredentials = [Convert]::ToBase64String($credentialBytes)
-          $baseCredentials = "Basic $encodedCredentials"
-          $Headers = @{ Authorization = $baseCredentials }
-          Invoke-WebRequest -SkipCertificateCheck -Uri 'https://localhost:9200' -Headers $Headers
+          arguements: integTestRemote -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="opensearch" -Dhttps=true -Duser=admin -Dpassword=admin
```
cwperks, line 35: Not sure this is working as intended:

```
Warning: Unexpected input(s) 'arguements', valid inputs are ['gradle-version', 'cache-disabled', 'cache-read-only', 'cache-write-only', 'gradle-home-cache-includes', 'gradle-home-cache-excludes', 'arguments', 'build-root-directory', 'gradle-executable', 'generate-job-summary', 'gradle-home-cache-strict-match', 'workflow-job-context', 'gradle-home-cache-cleanup']
```

.github/workflows/plugin_install.yml
```diff
@@ -3,120 +3,33 @@ name: Plugin Install
...
 
-      - name: Sleep while OpenSearch finishes starting up
-        uses: whatnick/wait-action@v0.1.2
+      - name: Run Opensearch with A Single Plugin
+        uses: ./.github/actions/start-opensearch-with-one-plugin
         with:
-          time: '30s'
+          opensearch-version: 3.0.0
+          plugin-name: opensearch-security
+          plugin-start-script: install_demo_configuration
```
cwperks, line 29: Can these parameters be multi-line? I see an opportunity for user's of this action to run commands before node startup to alter the configuration as necessary. Perhaps we can have all of:

```
chmod +x plugins/${{ inputs.plugin-name }}/tools/install_demo_configuration.sh
yes | plugins/${{ inputs.plugin-name }}/tools/install_demo_configuration.sh
```

as this parameter and decouple this even more from the security plugin?

.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -0,0 +1,126 @@
...
+  plugin-name:
+    description: 'The the name of the plugin to use, such as security-dashboards-plugin'
+    required: true
+
+  plugin-start-script:
+    description: 'The file name for the configuration script for the plugin such as install_demo_configurations -- may not be needed for every plugin'
+    required: false
+
+  plugin-zip:
+    description: 'The name of the zip file for the plugin hosted on docker-host i.e. security-plugin.zip '
```
cwperks, line 18: This may be confusing for someone who is unaware that this actions uses docker to start the node under the hood. What do you think of `A zip of the build of the plugin` or something similar?

cwperks (changes requested):
Great work on the abstraction of the github action to promote re-usability! Left a few suggestions, and one area that needs to be addressed. The section on integTests is not working, can you look into it?

.github/actions/start-opensearch-with-one-plugin/action.yml
scrawfor99: The plugin builds in the workflow file before this action is called. It was suggested to make as few operations happen in the action itself as possible so I moved the build outside of the action.

.github/actions/start-opensearch-with-one-plugin/action.yml
scrawfor99: The issue with swapping to any other naming convention is that it becomes further from the actual case. Saying a zip of the build of the plugin seems to suggest that it is just a zip but this is strictly for the file hosted on docker-host. So I think until we can rename the docker host file it is going to have to stay as it is.

.github/workflows/plugin_install.yml
scrawfor99: I am not sure what you mean by be multi-line. Swapping those two lines into the workflow instead of the action is possible but I do not know that we want that. If we did this and ran it outside the current step in the action (splitting it in two) it would require another plugin to know the exact spot in the script where they wanted to inject their own version of this script call as well as know the specific path. If instead we just swap it into an input parameter, in the exact same spot it does not really save any overhead and makes the parameters messier. Likewise, you could extract this code snippet and make it a separate action but I do not know that doing so would save any other plugin any more work. They would still have take the existing action and modify it to match whatever paths and file names they use.

.github/actions/start-opensearch-with-one-plugin/action.yml
scrawfor99: No, if you do not specify the a version tag for the image it will pull the latest code. That being said, I do think for readability it is nice to have the image tag there since people are most likely to be used to seeing docker images tagged like this. 

I can remove it if you prefer however.

cwperks: This command is building the container with an inline build script. Tagging the image is useful when publishing to docker hub, but doesn't look like its being used here. I know it wasn't introduced with this PR, but think its something worth pointing out.

.github/workflows/plugin_install.yml
cwperks: I think there needs to be a way to append to the config file before starting up a node. The concept of the setup script is nice and general, but I think it's too tightly coupled with the security plugin which has a tools folder with scripts. I'm not sure if any other plugins have tools folders with scripts. 

What do you think about having a step before the action in the workflow to create a setup script? The example below creates a file called `setup.sh` which installs the demo configuration and also appends to `opensearch.yml` to allow rest API security config modification.

```
- name: Create Setup Script
run: |
        cat > setup.sh <<EOF
        chmod +x plugins/${{ inputs.plugin-name }}/tools/install_demo_configuration.sh
        yes | plugins/${{ inputs.plugin-name }}/tools/install_demo_configuration.sh
        echo plugins.security.unsupported.restapi.allow_securityconfig_modification: true >> config/opensearch.yml
        EOF
```

Then we can set `plugin-setup-script` as the new `setup.sh` file and if there needs to be actions with different configs in the future then all it would take is defining your own creation of the setup script.

scrawfor99: Thank you for your feedback and example. Sorry if I am being daft, but I am not sure I understand what you are requesting. I think you want to create a separate step that creates a new file `setup.sh` which holds the two lines for the linux install command and then another line that adds the security REST api to the opensearch.yml. If this is the case, I caution against this as we would have to have two parameters to action.yml because Windows cannot use the same commands and now we have removed the file name parameter and replaced it with a linux-specific script. 

If we do want to add functionality for modifying the opensearch.yml before running the action I would probably leave that to another issue as I am not sure what that would look like or how we would use it. I could be misunderstanding your intent however.

scrawfor99:
I am not sure why this stopped working all of a sudden.

scrawfor99:
Seems there is something wrong with gradle and not the code so I will reflog my changes and then see where that gets us once it is fixed.

.github/workflows/plugin_install.yml
cwperks: Yes, some way of modifying the configuration before startup is needed. 

To try to clarify on why I think this is coupled with the security plugin is because `plugin-start-script` is used to run a script called `plugins/${{ inputs.plugin-name }}/tools/${{ inputs.plugin-start-script }}.sh` or `.bat` depending on the platform.

The security plugin has a `tools` directory with tools like the installation of the demo configuration, security admin and hash.

I just went through 3 other plugin repos:

- https://github.com/opensearch-project/observability
- https://github.com/opensearch-project/notifications
- https://github.com/opensearch-project/sql

and none have a `tools` directory in the same way.

scrawfor99: I separated the logic for handling a demo script path to only run if the there was demo path provided. Otherwise it will not run. That should mean that any other plugins can simply remove the script name from the action when they are setting things up.

scrawfor99: Sorry @cwperks but I still do not understand what you are hoping to see here. I know you have concerns about the re-usability of the action but I think that there is always going to be some variance between the different plugins. I do not believe that adding the options directly into the opensearch.yml is an option at the moment as we would have to do it between the time that opensearch installs the security plugin and the when it tries to run. This would mean that we would either have to split the action into two parts with a new step to run the demo script in the middle or just add two if-statemented steps for each OS inside the action itself. This seems more complicated then just writing a note that tells other plugins who do not use scripts to remove the lines. 

There is always going to be a bit of different between each code base unfortunately, as things get named in different ways etc. I think that this still offers a very strong starting point for other plugins to build or adapt to their own needs while also working as intended for the security repo. In the meantime, I can swap the line  `plugins/${{ inputs.plugin-name }}/tools/install_demo_configuration.sh` to be an entire path handed into the file instead of just a relative path containing the `tools` subdirectory. How does that sound? I am motivated to get this merged so please let me know what you would like done.

.github/workflows/plugin_install.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     strategy:
       matrix:
```
peternied, line 8: @scrawfor99  This is a good reason to have the strategy to `fail-fast: false`

.github/workflows/plugin_install.yml
scrawfor99: I looked into this and do not think that it will be a trivial change to reformat things with the setup.sh file as you suggested. A number of problems arise because this workflow functions on both Linux using docker and Windows using a manual setup. The only straightforward solution I think may be possible is having two separate setup.sh files -- one for each OS. This complicates things however and may not be in-line with the spirit of the original workflow. 

I went ahead and looked at the security-dashboards installation action to see what you were thinking as I know you had previously referenced that as an example. For that, I do not think that the setup steps for used will be directly applicable to this workflow. While there are echos into opensearch.yml they look like they are because the dashboards repository already has a system in place for operating based off of the contents of opensearch.yml. It looks like the same script running used in this action is also used there. 

I am of the opinion that if we want to implement further changes to the way the action works we can try to come up with a uniform solution for the various actions as part of another issue. This code is still widely usable by other plugins with only minor modifications and for the time being, I believe it is easy enough for others to remove the  the logic for the the script running as they see fit.

.github/workflows/plugin_install.yml
scrawfor99: Will do. See above, but I am hoping that for now we can accept the changes as they stand. If you would be willing to re-review it would be appreciated. The only thing that should have changed is that now there is the fail-fast policy you mentioned.

tools/install_demo_configuration.sh
```diff
@@ -425,6 +425,7 @@ else
 	echo "### or run ./securityadmin_demo.sh"
 	echo "### To use the Security Plugin ConfigurationGUI"
 fi
+eifjcbdkithivclitcurdbgnrcuhjfhffjdrbghfjelu
```
cwperks, line 428: Looks like this inadvertently got in here.

scrawfor99: Oops haha. Thank you!

.github/workflows/plugin_install.yml
cwperks: It can only be used by the security plug-in in the current form because of the structure of having scripts in the `tools` folder.

What I am thinking is adding steps in the workflow file (`plugin_install.yml`) to create a setup step for each platform.

`plugin_install.yml`
```
- name: Create Setup Script for Linux
   if: ${{ runner.os == 'Linux' }}
   run: |
        cat > setup.sh <<EOF
        chmod +x plugins/${{ inputs.plugin-name }}/tools/install_demo_configuration.sh
        yes | plugins/${{ inputs.plugin-name }}/tools/install_demo_configuration.sh
        echo plugins.security.unsupported.restapi.allow_securityconfig_modification: true >> config/opensearch.yml
        EOF

- name: Create Setup Script for Windows
   if: ${{ runner.os == 'Windows' }}
   shell: pwsh
   run: |
        echo. >> setup.bat
        echo 'y', 'y', 'N' | .\opensearch-${{ inputs.opensearch-version }}-SNAPSHOT\plugins\${{ inputs.plugin-name }}\tools\install_demo_configuration.bat >> setup.bat

- name: Run Opensearch with A Single Plugin
        uses: ./.github/actions/start-opensearch-with-one-plugin
        with:
          opensearch-version: 3.0.0
          plugin-name: opensearch-security
          plugin-setup-script: setup
          docker-host-plugin-zip: security-plugin.zip
```

and in the action:

`action.yml`
```
- name: Install Plugin into OpenSearch
   ...

- name: Setup OpenSearch for Linux
   run: `${{ inputs.plugin-setup-script}}`.sh

- name: Setup OpenSearch for Windows
   run: `${{ inputs.plugin-setup-script}}`.bat

- name: Run OpenSearch with plugin
   ....
```

Thank you for filing an issue for follow-up: https://github.com/opensearch-project/security/issues/2252

DarshitChanpura: I agree with @cwperks here (ty for the sample code). Making this modular and modifiable will give provide an opportunity for this workflow to be re-usable by other plugin and makes this flow much cleaner.

peternied (approved):

cwperks (approved):
Great work, thank you @scrawfor99!

Issue #2250, opened 2022-11-09
[BUG] Backward compatibility tests do not execute test cases


peternied:
**What is the bug?**
If you run the bwc test workflow on your own computer following along from the ci.yml [1] on your local machine the clusters are setup, but the test files are never executed.

**What is the expected behavior?**
The SecurityBackwardsCompatibilityIT [2] tests should be executed against all the different cluster configurations and pass / fail as planned

**Do you have any additional context?**
W

[1] https://github.com/opensearch-project/security/blob/40e2e9c628b20b932140288af890b7f67cbabbe7/.github/workflows/ci.yml#L86
[2] https://github.com/opensearch-project/security/blob/40e2e9c628b20b932140288af890b7f67cbabbe7/bwc-test/src/test/java/SecurityBackwardsCompatibilityIT.java

Issue #2232, opened 2022-11-03
Update BWC tests to operate on Windows


peternied:
Backwards compatibility workflow [1] needs to be updated to run on windows and ubuntu.

[1] https://github.com/opensearch-project/security/blob/main/.github/workflows/ci.yml#L86

peternied:
Note - this is not required for the 2.4 release because there is not backward combability requirements for Windows since we never released that platform before

Issue #1474, opened 2021-10-27
Add Windows Support


dblock:
Coming from https://github.com/opensearch-project/opensearch-plugins/issues/95, add Windows support.

- [x] Passing CI on Windows
- [x] opensearch-project/documentation-website#1714
- [x] https://github.com/opensearch-project/security/issues/1476
- [x] #2232

peternied:
BWC tests are not running on windows, added a new task and reopenning

Issue #2228, opened 2022-11-03
[BUG] Errors/Broken operations during rolling upgrade of clusters from 1.3 to 2.0


saikaranam-amazon:
**What is the bug?**
Errors/Broken search results during rolling upgrade of clusters from 1.3 to 2.0

**How can one reproduce the bug?**
1. Create 1.3 cluster with atleast 2 nodes
2. Create index with 2 primaries to allocate atleast one primary per node.
3. Upgrade one of the node to 2.0 OS version.
4. Invoke search query to invoke search on all the shards from 1.3 node.
5. See that there are failures to execute the search on this request
```
"took" : 40,
  "timed_out" : false,
  "_shards" : {
    "total" : 2,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 1,
    "failures" : [
      {
        "shard" : 1,
        "index" : "test-index",
        "node" : "O7kxX-lMTAKvXBj91-LQ8Q",
        "reason" : {
          "type" : "exception",
          "reason" : "java.lang.ClassNotFoundException: com.amazon.opendistroforelasticsearch.security.user.User",
          "caused_by" : {
            "type" : "class_not_found_exception",
            "reason" : "class_not_found_exception: com.amazon.opendistroforelasticsearch.security.user.User"
          }
        }
      }
    ]
  },
  "hits" : {
    "total" : {
      "value" : 2,
      "relation" : "eq"
    },
    "max_score" : 1.0,
    "hits" : [
      {
        ".....
          }
        }
      }
    ]
```

Notice that the failed shard count is 1

**What is the expected behavior?**
Rolling upgrade of clusters should complete without any issues.

**What is your host/environment?**
 - OS: 1.3 to OS: 2.0 upgrade
 - Plugins: Security

saikaranam-amazon:
Upon investigation,
- The issue is first seen on https://github.com/opensearch-project/security/issues/1259 and the code was reverted on 2.0 due to this https://github.com/opensearch-project/security/issues/1653
- But during the rolling upgrade, requests initiated from the co-ordinator node with 1.3 OS version fails to fetch the docs on the shards hosted on 2.0 version nodes with the above failure.

krishna-ggk:
Looks like this was just discovered few days back - https://github.com/opensearch-project/security/issues/2168

scrawfor99:
Hi @saikaranam-amazon are you also running the tarball installation? Thank you.

saikaranam-amazon:
[...]

PR #2253, opened 2022-11-09
Generalize Backwards Compatibility tests so we can test from any version to any version

peternied:
### Description
With an issue reported indicating that there are serialization issue between 1.3 and 2.X, making sure that we can reproduce the errors.  This new workflow(s) will make sure that we aren't breaking BWC with changes we are adding to 2.X and will give us the flexibility to test certain migration workflows.

Fixing an issue where the BWC tests were not actually building or executing causing the clusters to spin up and then immediately spin down.  We will need to invest more energy into running multiple kinds of  security plugin specific scenarios through the test system.

### Issues Resolved
- Resolves https://github.com/opensearch-project/security/issues/2250
- Resolves https://github.com/opensearch-project/security/issues/2232
- Resolves https://github.com/opensearch-project/security/issues/1474
- Related https://github.com/opensearch-project/security/issues/2228

### Check List

### DIFF
--- a/.github/actions/create-bwc-build/action.yaml
+++ b/.github/actions/create-bwc-build/action.yaml
@@ -38,6 +38,6 @@ runs:
 
     - name: Copy current distro into the expected folder
       run: |
-        mkdir -p ./bwc-test/src/test/resources/${{ steps.get-opensearch-version.outputs.built-version }}
-        cp ${{ inputs.plugin-branch }}/build/distributions/opensearch-security-${{ steps.get-opensearch-version.outputs.built-version }}.zip ./bwc-test/src/test/resources/${{ steps.get-opensearch-version.outputs.built-version }}
+        mkdir -p ./bwc-test/src/test/resources/${{ steps.get-opensearch-version.outputs.version }}
+        cp ${{ inputs.plugin-branch }}/build/distributions/opensearch-security-${{ steps.get-opensearch-version.outputs.version }}.zip ./bwc-test/src/test/resources/${{ steps.get-opensearch-version.outputs.version }}
       shell: bash
--- a/.github/actions/create-bwc-build/action.yaml
+++ b/.github/actions/create-bwc-build/action.yaml
@@ -9,7 +9,7 @@ inputs:
 outputs:
   built-version:
     description: 'The version of OpenSearch that was associated with this branch'
-    value: ${{ steps.get-opensearch-version.outputs.built-version }}
+    value: ${{ steps.get-opensearch-version.outputs.version }}
 
 runs:
   using: "composite"
--- a/.github/actions/create-bwc-build/action.yaml
+++ b/.github/actions/create-bwc-build/action.yaml
@@ -1,0 +1,47 @@
+name: 'Create a backwards compatible ready build'
+description: 'Checkouts the official version of a the Security plugin and builds it so it can be used for BWC tests'
+
+inputs:
+  plugin-branch:
+    description: 'The branch of the plugin that should be built, e.g "2.2", "1.x"'
+    required: true
+  
+outputs:
+  built-version:
+    description: 'The version of OpenSearch that was associated with this branch'
+    value: ${{ steps.get-opensearch-version.outputs.built-version }}
+
+runs:
+  using: "composite"
+  steps:
+      # Configure longpath names if on Windows
+    - name: Enable Longpaths if on Windows
+      if: ${{ runner.os == 'Windows' }}
+      run: git config --system core.longpaths true
+      shell: pwsh
+
+    - uses: actions/checkout@v3
+      with:
+        repository: opensearch-project/security
+        ref: ${{ inputs.plugin-branch }}
+        path: ${{ inputs.plugin-branch }}
+
+    - name: Build
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: assemble -Dbuild.snapshot=false
+        build-root-directory: ${{ inputs.plugin-branch }}
+
+    # This might not work in windows
+    - id: get-opensearch-version
+      run: |
+        export extractedVersion=$(./gradlew properties -Dbuild.snapshot=false | grep ^version: | awk '{split($0, a, ": ");print a[2]}')
+        echo "::set-output name=built-version::$extractedVersion"
+      working-directory: ${{ inputs.plugin-branch }}
+      shell: bash
+
+    - name: Copy current distro into the expected folder
+      run: |
+        mkdir -p ./bwc-test/src/test/resources/${{ steps.get-opensearch-version.outputs.built-version }}
+        cp ${{ inputs.plugin-branch }}/build/distributions/opensearch-security-${{ steps.get-opensearch-version.outputs.built-version }}.zip ./bwc-test/src/test/resources/${{ steps.get-opensearch-version.outputs.built-version }}
+      shell: bash
--- a/.github/actions/create-bwc-build/action.yaml
+++ b/.github/actions/create-bwc-build/action.yaml
@@ -14,7 +14,6 @@ outputs:
 runs:
   using: "composite"
   steps:
-      # Configure longpath names if on Windows
     - name: Enable Longpaths if on Windows
       if: ${{ runner.os == 'Windows' }}
       run: git config --system core.longpaths true
@@ -32,13 +31,10 @@ runs:
         arguments: assemble -Dbuild.snapshot=false
         build-root-directory: ${{ inputs.plugin-branch }}
 
-    # This might not work in windows
     - id: get-opensearch-version
-      run: |
-        export extractedVersion=$(./gradlew properties -Dbuild.snapshot=false | grep ^version: | awk '{split($0, a, ": ");print a[2]}')
-        echo "::set-output name=built-version::$extractedVersion"
-      working-directory: ${{ inputs.plugin-branch }}
-      shell: bash
+      uses: peternied/get-opensearch-version@v1
+      with:
+        working-directory: ${{ inputs.plugin-branch }}
 
     - name: Copy current distro into the expected folder
       run: |
--- a/.github/actions/run-bwc-suite/action.yaml
+++ b/.github/actions/run-bwc-suite/action.yaml
@@ -1,5 +1,5 @@
-name: 'Create a backwards compatible ready build'
-description: 'Checkouts the official version of a the Security plugin and builds it so it can be used for BWC tests'
+name: 'Runs the backward bompatiblity test suite'
+description: 'Tests backwards compability between a previous and next version of this plugin'
 
 inputs:
   plugin-previous-branch:
--- a/.github/actions/run-bwc-suite/action.yaml
+++ b/.github/actions/run-bwc-suite/action.yaml
@@ -1,0 +1,46 @@
+name: 'Create a backwards compatible ready build'
+description: 'Checkouts the official version of a the Security plugin and builds it so it can be used for BWC tests'
+
+inputs:
+  plugin-previous-branch:
+    description: 'The branch of the plugin that should be built for the previous version, e.g "2.2", "1.x"'
+    required: true
+
+  plugin-next-branch:
+    description: 'The branch of the plugin that should be built for the next version, e.g "2.3", "main"'
+    required: true
+
+  report-artifact-name:
+    description: 'The name of the artifacts for this run, e.g. "BWC-2.1-to-2.4-results"'
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+
+    - id: build-previous
+      uses: ./.github/actions/create-bwc-build
+      with:
+        plugin-branch: ${{ inputs.plugin-previous-branch }}
+
+    - id: build-next
+      uses: ./.github/actions/create-bwc-build
+      with:
+        plugin-branch: ${{ inputs.plugin-next-branch }}
+
+    - name: Run BWC tests
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: |
+          bwcTestSuite
+          -Dtests.security.manager=false
+          -Dbwc.version.previous=${{ steps.build-previous.outputs.built-version }}
+          -Dbwc.version.next=${{ steps.build-next.outputs.built-version }} -i
+        build-root-directory: bwc-test
+
+    - uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: ${{ inputs.report-artifact-name }}
+        path: |
+          ./bwc-test/build/reports/
--- a/.github/workflows/bwc-tests.yml
+++ b/.github/workflows/bwc-tests.yml
@@ -1,0 +1,42 @@
+name: Backwards Compability Suite
+
+on: [workflow_dispatch]
+
+jobs:
+  last-supported-major:
+    name: Make sure that the last supported major version can mov
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 11
+
+    - name: Checkout Security Repo
+      uses: actions/checkout@v2
+
+    - id: build-previous
+      uses: ./.github/actions/run-bwc-suite
+      with:
+        plugin-previous-branch: "1.3"
+        plugin-next-branch: "2.x"
+        report-artifact-name: BWC-Last-Supported-Major
+
+  next-major:
+    name: Make sure that the current version is compatible with
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 11
+
+    - name: Checkout Security Repo
+      uses: actions/checkout@v2
+
+    - id: build-previous
+      uses: ./.github/actions/run-bwc-suite
+      with:
+        plugin-previous-branch: "2.x"
+        plugin-next-branch: "main"
+        report-artifact-name: BWC-Next-Major
--- a/.github/workflows/bwc-tests.yml
+++ b/.github/workflows/bwc-tests.yml
@@ -3,8 +3,8 @@ name: Backwards Compability Suite
 on: [workflow_dispatch]
 
 jobs:
-  last-supported-major:
-    name: Make sure that the last supported major version can mov
+  last-supported-major-to-current:
+    name: Make sure that the last supported major version can move to the most current version
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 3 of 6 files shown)


peternied:
The build is totally broken because a JDK18 features is being used - waiting on the following issue to be merged to resolve this
https://github.com/opensearch-project/OpenSearch/pull/5205

.github/actions/create-bwc-build/action.yaml
```diff
@@ -0,0 +1,47 @@
...
+      run: |
+        export extractedVersion=$(./gradlew properties -Dbuild.snapshot=false | grep ^version: | awk '{split($0, a, ": ");print a[2]}')
+        echo "::set-output name=built-version::$extractedVersion"
+      working-directory: ${{ inputs.plugin-branch }}
+      shell: bash
+
+    - name: Copy current distro into the expected folder
+      run: |
+        mkdir -p ./bwc-test/src/test/resources/${{ steps.get-opensearch-version.outputs.built-version }}
+        cp ${{ inputs.plugin-branch }}/build/distributions/opensearch-security-${{ steps.get-opensearch-version.outputs.built-version }}.zip ./bwc-test/src/test/resources/${{ steps.get-opensearch-version.outputs.built-version }}
```
scrawfor99, line 46: Do we want to swap the `opensearch-security` part for a generic variable which is parsed from the repo/branch? I do not know if other plugins use these same tests but if so I am sure they would love this.

.github/actions/run-bwc-suite/action.yaml
```diff
@@ -0,0 +1,46 @@
+name: 'Create a backwards compatible ready build'
+description: 'Checkouts the official version of a the Security plugin and builds it so it can be used for BWC tests'
+
+inputs:
+  plugin-previous-branch:
+    description: 'The branch of the plugin that should be built for the previous version, e.g "2.2", "1.x"'
+    required: true
+
+  plugin-next-branch:
+    description: 'The branch of the plugin that should be built for the next version, e.g "2.3", "main"'
```
scrawfor99, line 10: Can we just increment automatically by saying 2.n --> 2.n+1 and  1.x --> 2.0, 2.x --> 3.0? Or is this not quite consistent with our naming conventions?

.github/workflows/bwc-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    - id: build-previous
+      uses: ./.github/actions/run-bwc-suite
+      with:
+        plugin-previous-branch: "1.3"
+        plugin-next-branch: "2.x"
+        report-artifact-name: BWC-Last-Supported-Major
+
+  next-major:
+    name: Make sure that the current version is compatible with
```
scrawfor99, line 26: Seems like some text got cut off at the end of this name

bwc-test/build.gradle
```diff
@@ -148,24 +163,7 @@ List<Provider<RegularFile>> plugins = [
 // Creates a test cluster with 3 nodes of the old version.
 2.times {i ->
     task "${baseName}#oldVersionClusterTask$i"(type: StandaloneRestIntegTestTask) {
-        exclude '**/*Test*'
-        exclude '**/*Sanity*'
         useCluster testClusters."${baseName}$i"
-        if (System.getProperty("mixedCluster") != null) {
```
scrawfor99, line 154: We do not need any of these tests any more? I know you added coverage in the action but seems like we may still be mixing some of these tests after the changes?

bwc-test/src/test/java/SecurityBackwardsCompatibilityIT.java
```diff
@@ -95,19 +102,16 @@ private void assertPluginUpgrade(String uri) throws Exception {
...
-                    Assert.assertTrue(pluginNames.contains("opensearch-security"));
-                    break;
-                case UPGRADED:
-                    Assert.assertTrue(pluginNames.contains("opendistro_security"));
-                    break;
+            Set<String> pluginNames = plugins.stream().map(map -> (String) map.get("name")).collect(Collectors.toSet());
+
+            final Version minNodeVersion = this.minimumNodeVersion();
+
+            if (minNodeVersion.major <= 1) {
```
scrawfor99, line 109: This is a clever way of doing this.

peternied:
Note; CI is blocked until the following issue(s) are resolved:
- https://github.com/opensearch-project/opensearch-build/issues/2886

.github/actions/create-bwc-build/action.yaml
peternied: I figured for this initial version I would just hardcode in these values.  If anyone else wants to pick it up we can add it later, seem fair?

.github/actions/run-bwc-suite/action.yaml
peternied: This parameter is meant to allow for different kinds of testing e.g. between to small changes by SHAs, between 1.2 -> 2.2 or between 1.2 -> main.  While these scenarios at not necessary supported, this allows testing them in the event we wanted to test them.

We could describe the version flow pattern and pick up the N+1 version, but I figure that makes more sense out of band of this change.

.github/actions/create-bwc-build/action.yaml
```diff
@@ -0,0 +1,47 @@
+name: 'Create a backwards compatible ready build'
+description: 'Checkouts the official version of a the Security plugin and builds it so it can be used for BWC tests'
+
+inputs:
+  plugin-branch:
+    description: 'The branch of the plugin that should be built, e.g "2.2", "1.x"'
+    required: true
```
scrawfor99, line 7: I have not needed to specify the security version in the past since we care on a given version branch. Is this meant to be the backwards version we are testing? Maybe rename if so.

.github/actions/create-bwc-build/action.yaml
scrawfor99: This sounds reasonable to me!

.github/actions/create-bwc-build/action.yaml
```diff
@@ -0,0 +1,47 @@
...
+      # Configure longpath names if on Windows
+    - name: Enable Longpaths if on Windows
+      if: ${{ runner.os == 'Windows' }}
+      run: git config --system core.longpaths true
+      shell: pwsh
+
+    - uses: actions/checkout@v3
+      with:
+        repository: opensearch-project/security
+        ref: ${{ inputs.plugin-branch }}
```
scrawfor99, line 26: I have had success with just `uses: actions/checkout@v{n}` so you may not need the lower lines.

.github/actions/create-bwc-build/action.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - name: Build
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: assemble -Dbuild.snapshot=false
+        build-root-directory: ${{ inputs.plugin-branch }}
+
+    # This might not work in windows
+    - id: get-opensearch-version
+      run: |
+        export extractedVersion=$(./gradlew properties -Dbuild.snapshot=false | grep ^version: | awk '{split($0, a, ": ");print a[2]}')
```
scrawfor99, line 38: While I love this, I have a suspicion Windows will not handle the `awk` properly. Could be wrong though.

.github/actions/run-bwc-suite/action.yaml
scrawfor99: This makes sense then. I just wanted to make sure I understood your plan.

.github/actions/run-bwc-suite/action.yaml
```diff
@@ -0,0 +1,46 @@
...
+      with:
+        arguments: |
+          bwcTestSuite
+          -Dtests.security.manager=false
+          -Dbwc.version.previous=${{ steps.build-previous.outputs.built-version }}
+          -Dbwc.version.next=${{ steps.build-next.outputs.built-version }} -i
+        build-root-directory: bwc-test
+
+    - uses: actions/upload-artifact@v3
+      if: always()
```
scrawfor99, line 42: Why have the `if` when it is set to `always`?

.github/workflows/bwc-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  last-supported-major:
+    name: Make sure that the last supported major version can mov
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 11
```
scrawfor99, line 13: As you know, you may run into issues with using jdk 11. Do we want/need to account for other jdk versions? What happens when we move from a version that supports jdk 17/18 to one that does not?

.github/actions/create-bwc-build/action.yaml
peternied: This allows the tests to be run against arbitrary builds, rather than only versions that are build on build.ci.opensearch.org.  We could run BWC test on windows if 2.3 builds on windows even if we don't have an official build.

.github/actions/create-bwc-build/action.yaml
peternied: I need the ref/path because these files are placed in another directory pointing to the other reference of the security build.

.github/actions/create-bwc-build/action.yaml
peternied: Bash on windows is amazing! See the git hub action passes at this part, command executes [1] then see the value is used [2] to create the directory variable.

I also tested on my windows desktop in the Git Bash
```
petern@SEA-1800605833 MINGW64 ~/Documents/GitHub/security (main)
$ export extractedVersion=$(./gradlew properties -Dbuild.snapshot=false | grep ^version: | awk '{split($0, a, ": ");print a[2]}')
Picked up JAVA_TOOL_OPTIONS: -Dlog4j2.formatMsgNoLookups=true

petern@SEA-1800605833 MINGW64 ~/Documents/GitHub/security (main)
$ echo $extractedVersion
3.0.0.0
```

Unfortunately the tests are still broken because we don't have 2.5.0 version :/


[1]  https://github.com/opensearch-project/security/actions/runs/3463162053/jobs/5784787988#step:4:590
[2] https://github.com/opensearch-project/security/actions/runs/3463162053/jobs/5784787988#step:4:603

bwc-test/build.gradle
```diff
@@ -70,19 +67,37 @@ repositories {
...
+String previousVersion = System.getProperty("bwc.version.previous", "2.4.0.0")
+String nextVersion = System.getProperty("bwc.version.next", "3.0.0.0")
+
+String bwcVersion = previousVersion
 String baseName = "securityBwcCluster"
 String bwcFilePath = "src/test/resources/"
-String projectVersion = "3.0.0.0"
+String projectVersion = nextVersion
+
+def previousVersionMatcher =  previousVersion =~ /(.+?)(\.\d+)$/
```
cwperks, line 87: Looks like an opportunity for a re-usable function here

.github/actions/run-bwc-suite/action.yaml
peternied: This forces the reports to be captured/uploaded even if there is a failure in the previous step, this way we always get logs.

bwc-test/build.gradle
peternied: This filter logic is really complex and I'm not sure it accomplishes much.   System properties are set indicating what kind of test is running that can be read by the test, e.g. `systemProperty 'tests.rest.bwcsuite', 'old_cluster'`  When looking at our previous BWC tests they didn't do much different between the different kinds of testing.


Created the following issue to figure out what tests we should add:
- https://github.com/opensearch-project/security/issues/2259

.github/workflows/bwc-tests.yml
```diff
@@ -0,0 +1,42 @@
+name: Backwards Compability Suite
+
+on: [workflow_dispatch]
+
+jobs:
+  last-supported-major:
```
cwperks, line 6: Thoughts on making this even more verbose? `last-supported-major-to-current`

.github/workflows/bwc-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+      uses: actions/checkout@v2
+
+    - id: build-previous
+      uses: ./.github/actions/run-bwc-suite
+      with:
+        plugin-previous-branch: "1.3"
+        plugin-next-branch: "2.x"
+        report-artifact-name: BWC-Last-Supported-Major
+
+  next-major:
```
cwperks, line 25: and for here `current-to-next-unreleased-major`?

cwperks (commented):
Thank you @peternied! I took a first pass and really like the flexibility this gives us in BWC testing. Left a few thoughts. There's a few descriptions that look incomplete which should be addressed, but otherwise this looks good to me.

.github/workflows/bwc-tests.yml
peternied: I wrote these tests as specific one off- I imagine this file might become a much bigger matrix of all testing we want done, but wanted to start smaller, I'll create a follow up issue.  

- https://github.com/opensearch-project/security/issues/2260

bwc-test/src/test/java/SecurityBackwardsCompatibilityIT.java
```diff
@@ -95,19 +102,16 @@ private void assertPluginUpgrade(String uri) throws Exception {
...
+                assertThat(pluginNames, hasItem("opensearch_security"));
```
DarshitChanpura, line 110: did you mean `opendistro_security` here?

bwc-test/src/test/java/SecurityBackwardsCompatibilityIT.java
```diff
@@ -52,16 +66,9 @@ protected final Settings restClientSettings() {
                 .build();
     }
 
-    public void testPluginUpgradeInAMixedCluster() throws Exception {
-        assertPluginUpgrade("_nodes/" + CLUSTER_NAME + "-0/plugins");
-    }
-
-    public void testPluginUpgradeInAnUpgradedCluster() throws Exception {
```
DarshitChanpura, line 59: why remove these tests? If I understand correctly, they are testing backward compatibility during upgrade

DarshitChanpura (changes requested):
Much cleaner and compartmentalized!!

bwc-test/build.gradle
```diff
@@ -70,19 +67,37 @@ repositories {
...
+    testImplementation 'org.apache.logging.log4j:log4j-core:2.17.1'
 }
 
-String bwcVersion = "2.4.0.0";
+loggerUsageCheck.enabled = false
+testingConventions.enabled = false
+validateNebulaPom.enabled = false
+
+String previousVersion = System.getProperty("bwc.version.previous", "2.4.0.0")
+String nextVersion = System.getProperty("bwc.version.next", "3.0.0.0")
```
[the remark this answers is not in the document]
scrawfor99: Gotcha. That all sounds good then. Thank you for clarifying, I just know I have seen you do some fancy parsing stuff in the past and was not sure if this would be a good opportunity to do likewise.

bwc-test/build.gradle
scrawfor99: Perfect! I know you still have some tests, I just wanted to get a read on what you thought about testing moving forward.

.github/actions/create-bwc-build/action.yaml
scrawfor99: Oh okay. This makes sense then. Thank you for explaining.

.github/actions/create-bwc-build/action.yaml
scrawfor99: Understood, I must have missed then when reading the file paths. Makes sense then!

.github/actions/create-bwc-build/action.yaml
scrawfor99: Well I stand corrected then! I think my use of Powershell must have messed with a similar implementation I had made a while back then. Regardless this seems great then. Maybe another opportunity for you to make one of your little sub actions if you wanted even. I bet a lot of people/teams would be interested in it :)

.github/actions/run-bwc-suite/action.yaml
scrawfor99: Oh I see. Looks good then.

.github/workflows/bwc-tests.yml
scrawfor99: Makes perfect sense to me!

bwc-test/src/test/java/SecurityBackwardsCompatibilityIT.java
peternied: More details here
https://github.com/opensearch-project/security/pull/2253/files/aca655b9ffef0411e414a64611d1cf28cf31d03b#r1021991312

These tests are not testing functionality of the security plugin and to only run certain tests during certain times a bunch of logic was in the gradle script.  Until we have a richer set of tests it seems easier to start with a test and then build out how it should run based on the different circumstances

bwc-test/src/test/java/SecurityBackwardsCompatibilityIT.java
peternied: These tests don't work because of the build breaks, when they do I'll make sure it checks the correct value.  Leaving this open until then

.github/actions/create-bwc-build/action.yaml
peternied: Darn your reasonable request https://github.com/peternied/get-opensearch-version!

scrawfor99: Lovely

scrawfor99 (dismissed):

peternied:
@cwperks @DarshitChanpura @scrawfor99 CI has finally been sorted out since we've got a 2.5.0 build, please take another look.

DarshitChanpura (approved):
Ty @peternied for resolving this pain!

cwperks:
@peternied Quick q: Is it safe to ignore these Windows errors? ```Failed to create working dir using hard links. Falling back to copy
org.opensearch.gradle.testclusters.OpenSearchNode$LinkCreationException: Failed to create hard link D:\a\security\security\bwc-test\build\testclusters\securityBwcCluster0-2\distro\3.0.0-ARCHIVE\bin\opensearch pointing to C:\Users\runneradmin\.gradle\caches\transforms-3\c38e0a71e65414d6057b5385bc6304be\transformed\opensearch-3.0.0-windows-x64.zip\opensearch-3.0.0\bin\opensearch```

peternied:
@cwperks Interesting, since they are falling back seems safe to me.  Took a look at the related code, seems like maybe its slowing the process down a little.  I think we could dig in more in the future if we were looking for a speed boost

https://github.com/opensearch-project/OpenSearch/blob/a0d30739402ab361d2a4c0fc6970ffbd2fe07039/buildSrc/src/main/java/org/opensearch/gradle/testclusters/OpenSearchNode.java#L1048

cwperks (approved):

scrawfor99:
This looks really good to me!

PR #2266, opened 2022-11-17
Command-line only plugin install action 

scrawfor99:
### Description
This PR builds on the recent PR for adding a plugin install action. This version drops the docker use from the Ubuntu runners and relies entirely on a standardized process for both operating systems. 

### Testing
Passes CI and successfully completes install action on Windows and Linux with JDK 11 and JDK 17. 

### Check List

### DIFF
--- a/.github/actions/start-opensearch-with-one-plugin/action.yml
+++ b/.github/actions/start-opensearch-with-one-plugin/action.yml
@@ -63,7 +63,7 @@ runs:
 
     # Move and rename the plugin for installation
     - name: Move and rename the plugin for installation
-      run: mv ./build/distributions/${{ inputs.plugin-name }}-*.zip ${{ inputs.plugin-name }}.zip
+      run: mv ./build/distributions/${{ inputs.plugin-name }}-*.zip $pwd/${{ inputs.plugin-name }}.zip
       shell: bash
 
     # Install the plugin, runs its start-script, and start the OpenSearch server
--- a/.github/actions/start-opensearch-with-one-plugin/action.yml
+++ b/.github/actions/start-opensearch-with-one-plugin/action.yml
@@ -88,7 +88,7 @@ runs:
       run: |
         chmod +x ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/bin/opensearch-plugin
         echo "Installing plugin" 
-        echo "yes" | ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/bin/opensearch-plugin install file:./${{ inputs.plugin-name }}.zip
+        echo "yes" | ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/bin/opensearch-plugin install file:$pwd/${{ inputs.plugin-name }}.zip
         echo "Installed plugin"
         echo "yes yes no" | ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/plugins/${{ inputs.plugin-name }}/tools/${{ inputs.plugin-start-script }}
         echo "Ran setup script"
--- a/.github/actions/start-opensearch-with-one-plugin/action.yml
+++ b/.github/actions/start-opensearch-with-one-plugin/action.yml
@@ -74,6 +74,9 @@ runs:
         echo in distributions 
         cd distributions 
         ls
+        echo $pwd
+        cd .. 
+        cd ..
         mv ./build/distributions/${{ inputs.plugin-name }}-${{ inputs.opensearch-version }}.0-SNAPSHOT.zip ./${{ inputs.plugin-name }}.zip
       shell: bash
 
--- a/.github/actions/start-opensearch-with-one-plugin/action.yml
+++ b/.github/actions/start-opensearch-with-one-plugin/action.yml
@@ -88,9 +88,9 @@ runs:
       run: |
         chmod +x ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/bin/opensearch-plugin
         echo "Installing plugin" 
-        yes > ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/bin/opensearch-plugin install file:$(pwd)/${{ inputs.plugin-name }}.zip
+        echo "yes" | ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/bin/opensearch-plugin install file:$(pwd)/${{ inputs.plugin-name }}.zip
         echo "Installed plugin"
-        echo "yes yes no" > ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/plugins/${{ inputs.plugin-name }}/tools/${{ inputs.plugin-start-script }}.sh
+        echo "yes yes no" | ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/plugins/${{ inputs.plugin-name }}/tools/${{ inputs.plugin-start-script }}.sh
         echo "Ran setup script"
       shell: bash
 
--- a/.github/actions/start-opensearch-with-one-plugin/action.yml
+++ b/.github/actions/start-opensearch-with-one-plugin/action.yml
@@ -73,8 +73,7 @@ runs:
         docker build -t opensearch-test -f- . <<EOF
         FROM ubuntu:latest
         COPY --chown=1001:1001 os-ep.sh /docker-host/
-        COPY --chown=1001:1001 ${{ inputs.plugin-name }}.zip /docker-host/${{ inputs.docker-host-
-        plugin-zip }}.zip
+        COPY --chown=1001:1001 ${{ inputs.plugin-name }}.zip /docker-host/${{ inputs.docker-host-plugin-zip }}.zip
         COPY --chown=1001:1001 opensearch* /opensearch/
         RUN chmod +x /docker-host/os-ep.sh
         RUN useradd -u 1001 -s /sbin/nologin opensearch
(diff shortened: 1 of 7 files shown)


.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -63,37 +61,24 @@ runs:
...
-        RUN useradd -u 1001 -s /sbin/nologin opensearch
-        ENV PATH="/opensearch/bin:${PATH}"
-        WORKDIR /opensearch/
-        ENTRYPOINT /docker-host/os-ep.sh
-        EOF
+
+        chmod +x ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/bin/opensearch-plugin
+        /bin/bash -c "yes | ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/bin/opensearch-plugin install file:$(pwd)/${{ inputs.plugin-name }}.zip"
+        chmod +x ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/plugins/${{ inputs.plugin-name }}/tools/${{ inputs.plugin-start-script }}.sh
+        /bin/bash -c "yes | ./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/plugins/${{ inputs.plugin-name }}/tools/${{ inputs.plugin-start-script }}.sh"
```
DarshitChanpura, line 68: nice!

.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -34,19 +30,21 @@ runs:
...
+    # Download OpenSearch
     - uses: peternied/download-file@v1
       if: ${{ runner.os == 'Linux' }}
       with:
-        url: https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/${{ inputs.opensearch-version }}/latest/linux/x64/tar/builds/opensearch/dist/opensearch-min-${{ inputs.opensearch-version }}-linux-x64.tar.gz
+        url: https://artifacts.opensearch.org/snapshots/core/opensearch/${{ inputs.opensearch-version }}-SNAPSHOT/opensearch-min-${{ inputs.opensearch-version }}-SNAPSHOT-linux-x64-latest.tar.gz
 
     # Extract downloaded zip
-    - name: Extract downloaded zip for Linux
+    - name: Extract downloaded zip
```
DarshitChanpura, line 40: this step looks like it is for Linux only

DarshitChanpura (commented):
Looks good @scrawfor99 !

A general comment: See if we can have a variable that is initialized with `-SNAPSHOT` or `` based on the parameter passed to the script and replace current use of `-SNAPSHOT` with that variable. Thoughts?

In `action.yml`:
```
..
..
runs:
  using: "composite"
  env:
     useSnapshot: ${{ fromJSON('{"false": "danger", "true": "good"}')[inputs.use-snapshot == 'true'] }}
     # Replace all -SNAPSHOT with this variable
..
..
```

In `plugin-install.yml`:
```
..
..
plugin-start-script: install_demo_configuration
use-snapshot: true
..
..

```
Maybe this works, maybe it doesn't. But if its takes long to make this change then its not worth it.

.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -34,19 +30,21 @@ runs:
       with:
         url: https://artifacts.opensearch.org/snapshots/core/opensearch/${{ inputs.opensearch-version }}-SNAPSHOT/opensearch-min-${{ inputs.opensearch-version }}-SNAPSHOT-windows-x64-latest.zip
 
+    # Download OpenSearch
     - uses: peternied/download-file@v1
       if: ${{ runner.os == 'Linux' }}
       with:
-        url: https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/${{ inputs.opensearch-version }}/latest/linux/x64/tar/builds/opensearch/dist/opensearch-min-${{ inputs.opensearch-version }}-linux-x64.tar.gz
+        url: https://artifacts.opensearch.org/snapshots/core/opensearch/${{ inputs.opensearch-version }}-SNAPSHOT/opensearch-min-${{ inputs.opensearch-version }}-SNAPSHOT-linux-x64-latest.tar.gz
```
peternied, line 37: Why change the source?  `https://ci.opensearch.org/...` [1] is the freshest place to get them from according to the OpenSearch-Build team [2]

[1] https://github.com/opensearch-project/opensearch-build/blob/main/README.md#latest-distribution-url
[2] https://github.com/opensearch-project/opensearch-build/issues/2754

scrawfor99:
Update: I spent today trying to incorporate the ability to define a setup.sh/.bat script in the workflow into this command-line only install action. I got it working on Linux with no docker use. The issue I am running into prevent Windows from working is described in detail here https://superuser.com/questions/1753564/how-to-pass-multiple-automatic-reponses-to-user-prompts-between-batch-files-usin. 

If anyone would like to look at Linux version which does work it can be viewed on https://github.com/scrawfor99/security/tree/SetupForPluginInstall and then by swapping the os in the workflow matrix to 'ubuntu-latest' instead of windows. I also went ahead and created two files that can be used to replicate the scenario on a local windows machine or a remote desktop. The contents of the two are pasted below (they are really short) 


test.bat: 

```
@echo off
echo                                     
echo.
echo Yes or No? (y/n)
set /p Input=Enter Yes or No:
If /I "%Input%"=="y" goto yes1
goto no1
:yes1
echo You said yes 
goto yorn1
:no1
echo You said no
goto yorn1
:yorn1
echo Yes or No? (y/n)
set /p Input=Enter Yes or No:
If /I "%Input%"=="y" goto yes2
goto no2
:yes2
echo You said yes 
goto yorn2
:no2
echo You said no
goto yorn2
:yorn2
echo Yes or No? (y/n)
set /p Input=Enter Yes or No:
If /I "%Input%"=="y" goto yes3
goto no3
:yes3
echo You said yes 
goto exitprocess
:no3
echo You said no 
goto exitprocess
:exitprocess
```


run_test.bat:
```
(echo y&&echo y&&echo n) | .\test.bat
```
It should work if you can call `.\run_test.bat` in the same directory as the test.bat file and see responses that say you said 'yes' twice and 'no' once.

.github/actions/start-opensearch-with-one-plugin/action.yml
scrawfor99: Yeah I briefly flew too close to the Sun and thought I could combine the two.

.github/actions/start-opensearch-with-one-plugin/action.yml
scrawfor99: So I swapped the source for two reasons: 1) We do not have the same type of catalogue for Windows as we do Linux so what was available to Linux was not for Windows which made the processes for the two more divergent then I liked 2) Related to that, the most recent distributions seemed to have security already installed which would require that I make the script uninstall and reinstall the plugin to actually test what they are meant to test. Likewise, they probably would have had other plugins installed and I am not sure how much I could account for the differences in what those install and removal processes looked like. With the minimums snapshots, there is basically a barebones framework of OpenSearch to work with which I think makes it easier to install into but also gives you a better idea of whether your particular plugin is working right. 

Good question though--I actually thought a bit about this.

scrawfor99:
> 

For the SNAPSHOT, I think that it is always going to be there so long as we use the distributions I selected. If you check out the response I sent to Peter, there were a couple reasons I chose to make both Linux and Windows runners use the min distributions. I am not sure what you guys will think but for now I think they should always have SNAPSHOT as a suffix. I can swap to your suggestion if that changes though!

scrawfor99:
> Looks good @scrawfor99 !
> 
> A general comment: See if we can have a variable that is initialized with `-SNAPSHOT` or `<empty-string>` based on the parameter passed to the script and replace current use of `-SNAPSHOT` with that variable. Thoughts?
> 
> In `action.yml`:
> 
> ```
> ..
> ..
> runs:
>   using: "composite"
>   env:
>      useSnapshot: ${{ fromJSON('{"false": "", "true": "-SNAPSHOT"}')[inputs.use-snapshot == 'true'] }}
>      # Replace all -SNAPSHOT with this variable
> ..
> ..
> ```
> 
> In `plugin-install.yml`:
> 
> ```
> ..
> ..
> plugin-start-script: install_demo_configuration
> use-snapshot: true
> ..
> ..
> ```
> 
> Maybe this works, maybe it doesn't. But if its takes long to make this change then its not worth it.

For the SNAPSHOT, I think that it is always going to be there so long as we use the distributions I selected. If you check out the response I sent to Peter, there were a couple reasons I chose to make both Linux and Windows runners use the min distributions. I am not sure what you guys will think but for now I think they should always have SNAPSHOT as a suffix. I can swap to your suggestion if that changes though!

.github/actions/start-opensearch-with-one-plugin/action.yml
peternied: > the most recent distributions seemed to have security already installed

That doesn't sound right, the `opensearch-min-*` is suppose to be only the core version of OpenSearch with no plugins, could you see about making a bug on OpenSearch-Build if that is the case.

DarshitChanpura:
> Thank you for your review @DarshitChanpura!
> 
> For the SNAPSHOT, I think that it is always going to be there so long as we use the distributions I selected. If you check out the response I sent to Peter, there were a couple reasons I chose to make both Linux and Windows runners use the min distributions. I am not sure what you guys will think but for now I think they should always have SNAPSHOT as a suffix. I can swap to your suggestion if that changes though!

yes having SNAPSHOT is absolutely ok. My question is more in terms of defining a boolean variable that can be passed dynamically and based on that the script decides whether to append `-SNAPSHOT` after the version

scrawfor99:
Closing in favor of #2271

PR #2197, opened 2022-10-26
 [FasterXML/Woodstox] Woodstox Version Bump to 6.4.0

scrawfor99:
Updated FasterXML/Woodstox version to 6.4.0 

Backport to 2.x., 1.x, 1.3

### Check List
- [ ] ~New functionality includes testing~
- [ ] ~New functionality has been documented~
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2269, opened 2022-11-21
Force woodstox to 6.4.0

scrawfor99:
Signed-off-by: Stephen Crawford <steecraw@amazon.com>

### Description
Fixes a remaining dependency to Woodstox 6.2.6 library as a follow-up to https://github.com/opensearch-project/security/pull/2197. I incorrectly looked at only the direct dependencies previously not accounting for the transient dependencies introduced by other libraries. This force should make the Woodstox version match the desired 6.4.0. 

### Issues Resolved
Further resolves the Woodstox dependency problem. 

### Testing 
After the change, runnning `./gradlew dependencies` shows all references to Woodstox-core being bumped to 6.4.0.

### Check List
- [] ~New functionality has been documented~

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,6 +70,7 @@ configurations.all {
         force "org.apache.commons:commons-lang3:3.4"
         force "org.springframework:spring-core:5.3.20"
         force "com.google.guava:guava:30.0-jre"
+        force "com.fasterxml.woodstox:woodstox-core:6.4.0"
     }
 }
 


cwperks:
@scrawfor99 Can you list the dependencies and versions being upgraded in the PR title?

build.gradle
```diff
@@ -70,6 +70,14 @@ configurations.all {
         force "org.apache.commons:commons-lang3:3.4"
         force "org.springframework:spring-core:5.3.20"
         force "com.google.guava:guava:30.0-jre"
+        force "com.fasterxml.woodstox:woodstox-core:6.4.0"
+        force "org.scala-lang:scala-library:2.13.9"
+        force "org.yaml:snakeyaml:1.32"
+        force 'io.netty:netty-buffer:4.1.78.Final'
```
cwperks, line 76: Do we need to force the resolution of the netty version here?

DarshitChanpura (approved):

build.gradle
scrawfor99: There was a netty cve in the previous push without the netty change. I know that in later versions we have a netty.version variable but this does not seem to exist for 1.x.

scrawfor99: Yeah, I think I may have seen the CVE for 1.3 on Netty and then because I just copy+pasted the same changes it may have been included. I can remove it if you think I should but I imagine keeping it is not a bad idea either.

scrawfor99: Alright I can remove snakeyaml and then swap netty to be 79

cwperks:
@scrawfor99 The `snakeyaml` flag can be ignored, the SNAPSHOT build is outdated. Netty can be added back into the resolutionStrategy section, but we will need a way to be notified of upstream changes from core to try to match the version. In the `main` branch we are able to reference versions from `versions.properties` using `${version.netty}`, but its not possible to do that in the 1.3 branch.

cwperks:
You can monitor builds here: https://build.ci.opensearch.org/job/distribution-build-opensearch/. Looks like there are some recent attempts to build 1.3.7, but none have succeeded.

scrawfor99:
@cwperks @DarshitChanpura what do you two think about the current state of things? Are we keeping netty and snakeyaml for now since 1.3 does not have the version properties or did you prefer I remove them? I would like to get this merged ASAP so just following up.

build.gradle
```diff
@@ -70,6 +70,13 @@ configurations.all {
         force "org.apache.commons:commons-lang3:3.4"
         force "org.springframework:spring-core:5.3.20"
         force "com.google.guava:guava:30.0-jre"
+        force "com.fasterxml.woodstox:woodstox-core:6.4.0"
+        force "org.scala-lang:scala-library:2.13.9"
+        force "org.yaml:snakeyaml:1.32"
```
cwperks, line 75: Can you remove snakeyaml and netty from this section?

scrawfor99: Done

scrawfor99:
Removing it causes WhiteSource Security to fail like before. I am not sure why we wanted to remove it if it causes the check to report a vulnerability.

DarshitChanpura:
> Removing it causes WhiteSource Security to fail like before. I am not sure why we wanted to remove it if it causes the check to report a vulnerability.

For the sake of this PR let's add netty and snakeyaml back, and change the PR description. We can cleanup later. @cwperks Let's fix the whitesource errors so that CI is unblocked.

scrawfor99:
Haha whatever is fine; I just want to do whatever is best.

cwperks:
I think we should add a step in the CI to checkout the corresponding branch of core, assemble the jar and publish to maven local so that the build is not dependent on updated artifacts from infra team.

The additional steps at the beginning of CI would be:

1. Checkout corresponding branch of core
2. Run `./gradlew assemble` or the corresponding command needed to generate a jar to publish to maven
3. Run `./gradlew publishToMavenLocal`

cwperks:
It may not be necessary to run `./gradlew assemble`. `./gradlew publishToMavenLocal` does the compilation, jarring and publishing to the local maven m2 repository.

cwperks (approved):

DarshitChanpura:
Merging this as the whitesource failures are due to stale artifacts and should resolve itself once latest artifacts become avaialble.

Issue #2023, opened 2022-08-12
[Infrastructure] Intermittent 1.3 branch CI failures


cwperks:
The recent [update](https://github.com/opensearch-project/security/pull/2004) of dependencies in 1.3 took 6 times of the CI running to pass CI. 

This was a minor fix to increment dependencies flagged by whitesource as having vulnerabilities and `jackson-databind` was upgraded to align with 1.x and with core's `version.properties`.

Some of the tests that failed intermittently include:

```
-org.opensearch.security.HttpIntegrationTests > testTenantInfo FAILED
    java.lang.AssertionError at HttpIntegrationTests.java:685

- org.opensearch.security.auditlog.integration.BasicAuditlogTest > testRestMethod FAILED
    java.lang.AssertionError at BasicAuditlogTest.java:854

- org.opensearch.security.auditlog.AuditLogTestSuite > org.opensearch.security.auditlog.sink.WebhookAuditLogTest.postGetHttpTest FAILED
    java.net.BindException: Address already in use (Bind failed)

- org.opensearch.security.ssl.SSLTest > testCRL FAILED
    java.lang.RuntimeException at SSLTest.java:798
        Caused by: org.opensearch.http.BindHttpException at AbstractHttpServerTransport.java:203
            Caused by: java.net.BindException at Net.java:-2

- org.opensearch.security.ssl.SSLTest > testTLSv12 FAILED
    java.lang.RuntimeException at SSLTest.java:874
        Caused by: org.opensearch.http.BindHttpException at AbstractHttpServerTransport.java:203
            Caused by: java.net.BindException at Net.java:-2
```

scrawfor99:
For reference here is a link to a specific instance of the types of improvements that may improve stability look at issue #2278

scrawfor99:
I have been working on this and I think that the issue is caused because the webhookAuditLogTest tests are called multiple times directly and also from AuditLogTestSuite. The tests actually pass 3 times (all called from the suite) before failing the first time they are directly called by webhook. Will look into this further on Monday.

peternied:
I think I've tackled many of these failures in https://github.com/opensearch-project/security/pull/2291 - take a look

scrawfor99:
I will definitely take a look at it and see where we are at with your fixes!

scrawfor99:
Closed since changes fixed test failures.

PR #2270, opened 2022-11-21
Update Dependency Versions:  Woodstox 6.4.0, Scala-lang 2.13.9, Jackson-Databind 2.14.0

scrawfor99:
Signed-off-by: Stephen Crawford <steecraw@amazon.com>

### Description
Fixes a remaining dependency to Woodstox 6.2.6 library as a follow-up to https://github.com/opensearch-project/security/pull/2197. I incorrectly looked at only the direct dependencies previously not accounting for the transient dependencies introduced by other libraries. This force should make the Woodstox version match the desired 6.4.0. 

I also corrected the versions for jackson_databind, kafka, and snakeyaml. 

### Issues Resolved
Further resolves the Woodstox dependency problem. 

### Testing 
After the change, runnning `./gradlew dependencies` shows all references to Woodstox-core being bumped to 6.4.0 and all the other WhiteSource issues being fixed. 

### Check List
- [] ~New functionality has been documented~


scrawfor99:
@cwperks @DarshitChanpura this build failure does not seem related to the changes made in this commit. What do you guys think?

DarshitChanpura:
I was just going through the test failures and I agree

DarshitChanpura:
I've triggered a re-run to see if it was one-off failure, cause most of the tests are failing due to `Address already in use` error

cwperks:
@DarshitChanpura 1.3 build is known to be flaky - https://github.com/opensearch-project/security/issues/2023

It may take a couple of retries.

DarshitChanpura:
okay.. do we want to proceed with merge @cwperks ?

cwperks:
@DarshitChanpura Let's make sure this PR and the one against 1.x match. I don't think we need `snakeyaml` listed in here and we may not need netty in the section either, but I'd like to wait for an updated SNAPSHOT from core and then re-run the CI to see what version of netty it would resolve to without forcing the resolution strategy. If we need to force the resolution strategy than that is fine, but it would be nice to have a mechanism to know when there is a version mismatch with the version in core's version.properties.

cwperks:
Actually this can be done locally as well.

If you checkout version 1.3 of core locally and run the following:

```
./gradlew assemble
./gradlew publishToMavenLocal
```

once those are complete switch to the patch for 1.3 of security and run

```
./gradlew clean assemble
```

That should produce a .zip in `build/distributions`

From there you check what jars are included for snakeyaml and netty.

DarshitChanpura:
@scrawfor99 Were you able to test the behavior with the steps @cwperks mentioned?

scrawfor99:
@DarshitChanpura, no because I cannot use my Mac (1.3 does not work on M1) to test it and I am getting errors trying to create a new remote desktop currently. I tried two different accounts and various settings with no avail so am going to table it for now. It should not be too big a change either way once it is testable.

scrawfor99:
Update: I am away for the remainder of the day for travel but I will be confirming this works tomorrow and updating accordingly. I did set up a linux remote desktop to test this.

CEHENKLE (dismissed):

cwperks (approved):

cwperks:
Confirmed via creating a local snapshot of the latest from 1.3 of core and publishing to maven local. It is not necessary to force the netty resolution as it will resolve to the version of netty from core. This will make maintenance easier because it will not require an update every time core updates the netty dependency.

cwperks:
@scrawfor99 Thank you for looking into CI stability on the 1.3 branch. Can you do the manual backport and test to see if it improves CI stability for 1.3?

cwperks (approved):

scrawfor99:
@cwperks thank you for rerunning the failures as needed until things passed. I looked into the issue and the backport is complicated. I suspect that is why it did not happen in the first place. I am going to create an issue for the changes to be backported. After resolving the basic issues with the backport I encountered a problem where the new logic expects to be able to find various files in `auditlog/` however I do not think this folder exists in the earlier version.

PR #2271, opened 2022-11-21
Universal Command-line only Plugin Installation Action & Workflow

### LINKED ISSUE #2207: [FEATURE] Migrate Plugin install into an GitHub action

### LINKED ISSUE #2252: [FEATURE] During OpenSearch service starting for GHA workflow be able to supply script/config modification
It was brought to my attention by @cwperks that it would be nice if the settings contained in our startup scripts i.e. `install_demo_configuration`  could be added directly to the `opensearch.yml`  file. This would then allow for the various workflows and GitHub Actions to not require that users specify a path to the demo configurations for their plugins (https://github.com/opensearch-project/security/pull/2239). 

Be able to specify in some way the configurations needed for a plugin or a redirect to a file containing these configurations within the opensearch.yml file. 

Currently the same functionality is implemented by passing a path to a script containing all the configuration steps.

scrawfor99:
### Description
This pull request introduces a refactored design of the plugin-installation workflow and action. As the final step in introducing the plugin-install action, the new action & workflow allow a user to specify any setup steps they want for their plugin and completely removes the need for passing any sort of configuration file path to the action itself. 

- Closes #2252
- Closes #2207 

### Testing
The new design works on Linux and Windows runners running JDK 11 and JDK 17 and have been verified to successfully install the opensearch-security plugin. 

### Check List


.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -59,41 +58,38 @@ runs:
...
-        COPY --chown=1001:1001 ${{ inputs.plugin-name }}.zip /docker-host/${{ inputs.docker-host-plugin-zip }}.zip
-        COPY --chown=1001:1001 opensearch* /opensearch/
-        RUN chmod +x /docker-host/os-ep.sh
-        RUN useradd -u 1001 -s /sbin/nologin opensearch
-        ENV PATH="/opensearch/bin:${PATH}"
-        WORKDIR /opensearch/
-        ENTRYPOINT /docker-host/os-ep.sh
-        EOF
+        chmod +x ./opensearch-3.0.0-SNAPSHOT/bin/opensearch-plugin
+        /bin/bash -c "yes | ./opensearch-3.0.0-SNAPSHOT/bin/opensearch-plugin install file:$(pwd)/opensearch-security.zip"
```
DarshitChanpura, line 66: `opensearch-3.0.0` should be replaced with `opensearch-${{ inputs.opensearch-version }}`

.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -59,41 +58,38 @@ runs:
...
+    - name: Run Setup Script for Windows
+      if: ${{ runner.os == 'Windows' && inputs.setup-script != '' }}
+      run: .\setup.bat
       shell: pwsh
 
     # Run OpenSearch
     - name: Run OpenSearch with plugin on Linux
       if: ${{ runner.os == 'Linux'}}
-      run: docker run --name ops -d -p 9200:9200 -p 9600:9600 -i opensearch-test:latest
+      run: /bin/bash -c "./opensearch-${{ inputs.opensearch-version }}-SNAPSHOT/bin/opensearch &"
```
DarshitChanpura, line 92: This command is set to run in background here with `&`. Any specific reason why background and not normally?

.github/workflows/plugin_install.yml
```diff
@@ -1,36 +1,120 @@
...
-
-      - name: Run Opensearch with A Single Plugin
-        uses: ./.github/actions/start-opensearch-with-one-plugin
-        with:
-          opensearch-version: 3.0.0
-          plugin-name: opensearch-security
-          plugin-start-script: install_demo_configuration
-          docker-host-plugin-zip: security-plugin.zip
-
-      - name: Run sanity tests
```
DarshitChanpura, line 33: Why did we remove the sanity test step?

DarshitChanpura (commented):
Generic questions:
1. How will `.github/actions/start-opensearch-with-one-plugin/action.yml` be used?
2. I see a bunch of similarities between `.github/actions/start-opensearch-with-one-plugin/action.yml` and `.github/workflows/plugin_install.yml`. What is the difference between these two and why are both required?

.github/actions/start-opensearch-with-one-plugin/action.yml
scrawfor99: Ooops, I missed swapping that. 


As for the questions: 

1) You can use it to install a generic plugin into a version of OpenSearch. This is useful for testing but could also be adapted into a start up command if you wanted. 

2) The workflow uses the action. A workflow is run on GitHub Actions for testing and automation. An action is basically an external script you can call from a workflow. So you can download a file for example or in this case install a plugin into opensearch.

.github/actions/start-opensearch-with-one-plugin/action.yml
scrawfor99: Because you want it to be a background so that it is able run detached outside of docker.

.github/workflows/plugin_install.yml
```diff
@@ -1,36 +1,120 @@
...
+    - name: Move and rename the plugin for installation
+      run: mv ./build/distributions/${{ inputs.plugin-name }}-*.zip ${{ inputs.plugin-name }}.zip
+      shell: bash
+
+    # Install the plugin
+    - name: Install Plugin into OpenSearch for Linux
+      if: ${{ runner.os == 'Linux'}}
+      run: |
+        chmod +x ./opensearch-3.0.0-SNAPSHOT/bin/opensearch-plugin
+        /bin/bash -c "yes | ./opensearch-3.0.0-SNAPSHOT/bin/opensearch-plugin install file:$(pwd)/opensearch-security.zip"
```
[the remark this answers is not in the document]
scrawfor99: This you cannot swap because the workflow does not get inputs in the same way as an action. So this you have to specify yourself.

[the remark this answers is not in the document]
DarshitChanpura: okay.. we can use env variable here. Thoughts?
https://docs.github.com/en/actions/learn-github-actions/environment-variables

.github/actions/start-opensearch-with-one-plugin/action.yml
DarshitChanpura: okie...in my understanding... everything runs outside docker now..so I'm still not sure how this would help (not saying it shouldn't be as it is)

.github/workflows/plugin_install.yml
```diff
@@ -1,36 +1,120 @@
...
-          opensearch-version: 3.0.0
-          plugin-name: opensearch-security
-          plugin-start-script: install_demo_configuration
-          docker-host-plugin-zip: security-plugin.zip
-
-      - name: Run sanity tests
-        uses: gradle/gradle-build-action@v2
-        with:
-          arguments: integTestRemote -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="opensearch" -Dhttps=true -Duser=admin -Dpassword=admin
+name: 'Start OpenSearch with One Plugin'
```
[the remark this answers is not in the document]
DarshitChanpura: `name: 'Install a Plugin in OpenSearch'`
Thoughts?

(Also, I meant to add the comment to line 1 but somehow ended up in line 2)

.github/workflows/plugin_install.yml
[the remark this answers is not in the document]
DarshitChanpura: Note. This is not urgent and can be done as cleanup (across both repos). But this will definitely help in version bump scenarios.

.github/actions/start-opensearch-with-one-plugin/action.yml
scrawfor99: Basically, the process fails if you do not detach the call. So by making it a background process everything works because then you can have it operate in the background and then go make a curl to it. Otherwise you cannot/would need to open a new terminal during the execution to curl to it. This is my understanding at least--it fails without the & so I added it at Craig's suggestion.

.github/workflows/plugin_install.yml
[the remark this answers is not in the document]
scrawfor99: That sounds good I will look at it

.github/workflows/plugin_install.yml
scrawfor99: So I realized that during the conflict resolution it messed up some of the files so I think that is why things looked so weird to you before. I readded this and fixed the copying over of the action to the workflow.

.github/workflows/plugin_install.yml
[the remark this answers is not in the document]
scrawfor99: Swapped to environment variables like you suggested.

.github/actions/start-opensearch-with-one-plugin/action.yml
cwperks: @DarshitChanpura This call is blocking if it is not detached and the rest of the action does not execute.

.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -11,7 +11,7 @@ inputs:
     required: true
 
   setup-script:
-    description: 'The file name for the configuration script for the plugin such as install_demo_configurations -- may not be needed for every plugin'
+    description: 'An input indicating whether a setup script should be run for the plugin. Leave empty to indicate one should not be run.'
     required: true
```
DarshitChanpura, line 15: based on above description should this parameter be false?

.github/actions/start-opensearch-with-one-plugin/action.yml
RyanL1997: @scrawfor99 So is there any difference between running it in this way and running it with `docker run`?

.github/actions/start-opensearch-with-one-plugin/action.yml
scrawfor99: Good catch!

scrawfor99:
@cwperks @DarshitChanpura hoping this can get merged. Thanks!

.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -10,14 +10,10 @@ inputs:
     description: 'The name of the plugin to use, such as opensearch-security'
     required: true
 
-  plugin-start-script:
-    description: 'The file name for the configuration script for the plugin such as install_demo_configurations -- may not be needed for every plugin'
+  setup-script:
```
peternied, line 13: Please change this parameters value to be the script name such as 'setup.bat' or 'setup.sh', hardcoding where the script is and how its named its difficult to follow

scrawfor99:
@peternied this should be all set. Note that the Windows & JDK 11 install seems to be flaky for whatever reason with the change to the file name version. I am not sure why this is.

.github/workflows/plugin_install.yml
```diff
@@ -20,15 +25,31 @@ jobs:
...
       - name: Run Opensearch with A Single Plugin
         uses: ./.github/actions/start-opensearch-with-one-plugin
         with:
-          opensearch-version: 3.0.0
-          plugin-name: opensearch-security
-          plugin-start-script: install_demo_configuration
-          docker-host-plugin-zip: security-plugin.zip
+          opensearch-version: ${{ env.OPENSEARCH_VERSION }}
+          plugin-name: ${{ env.PLUGIN_NAME }}
+          setup-script: setup
```
DarshitChanpura, line 52: this should be renamed to `setup-script-name` to conform with the change in the action.yml above

.github/actions/start-opensearch-with-one-plugin/action.yml
```diff
@@ -59,41 +58,38 @@ runs:
...
+        'y' | .\opensearch-${{ inputs.opensearch-version }}-SNAPSHOT\bin\opensearch-plugin.bat install file:$(pwd)\${{ inputs.plugin-name }}.zip
+      shell: pwsh
+
+    # Run any configuration scripts
+    - name: Run Setup Script for Linux
+      if: ${{ runner.os == 'Linux' && inputs.setup-script != '' }}
+      run: | 
+        echo "running linux setup"
+        chmod +x ./${{ inputs.setup-script }}.sh
+        ./${{ inputs.setup-script }}.sh
```
DarshitChanpura, line 81: this should be changed to setup-script-name to conform with above

DarshitChanpura (changes requested):
two quick changes requested. LGTM otherwise!

.github/workflows/plugin_install.yml
scrawfor99: Fixed!

.github/actions/start-opensearch-with-one-plugin/action.yml
scrawfor99: Fixed!

DarshitChanpura (approved):

peternied (approved):

PR #2277, opened 2022-11-25
Username validation for special characters

rutuja-amazon:
Add username validation:

- [x] Username should not include special characters

### DIFF
--- a/src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
+++ b/src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
@@ -14,6 +14,7 @@ package org.opensearch.security.dlic.rest.api;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
@@ -90,6 +91,12 @@ public class InternalUsersApiAction extends PatchableResourceApiAction {
 
         if (username == null || username.length() == 0) {
             badRequestResponse(channel, "No " + getResourceName() + " specified.");
+            return;
+        }
+
+        Pattern usernamePattern = Pattern.compile("[$&+,:;=\\\\?#|/'<>^*()%!]");
+        if (usernamePattern.matcher(username).find()) {
+            badRequestResponse(channel, "Username has special characters, not permitted.");
             return;
         }
 
--- a/src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
+++ b/src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
@@ -14,7 +14,7 @@ package org.opensearch.security.dlic.rest.api;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.List;
-import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
@@ -50,6 +50,10 @@ import static org.opensearch.security.dlic.rest.support.Utils.addRoutesPrefix;
 import static org.opensearch.security.dlic.rest.support.Utils.hash;
 
 public class InternalUsersApiAction extends PatchableResourceApiAction {
+    static final List<String> RESTRICTED_FROM_USERNAME = ImmutableList.of(
+        ":" // Not allowed in basic auth, see https://stackoverflow.com/a/33391003/533057
+    );
+
     private static final List<Route> routes = addRoutesPrefix(ImmutableList.of(
             new Route(Method.GET, "/user/{name}"),
             new Route(Method.GET, "/user/"),
@@ -94,9 +98,10 @@ public class InternalUsersApiAction extends PatchableResourceApiAction {
             return;
         }
 
-        Pattern usernamePattern = Pattern.compile("[$&+,:;=\\\\?#|/'<>^*()%!]");
-        if (usernamePattern.matcher(username).find()) {
-            badRequestResponse(channel, "Username has special characters, not permitted.");
+        final List<String> foundRestrictedContents = RESTRICTED_FROM_USERNAME.stream().filter(username::contains).collect(Collectors.toList());
+        if (!foundRestrictedContents.isEmpty()) {
+            final String restrictedContents = foundRestrictedContents.stream().map(s -> "'" + s + "'").collect(Collectors.joining(","));
+            badRequestResponse(channel, "Username has restricted characters " + restrictedContents + " that are not permitted.");
             return;
         }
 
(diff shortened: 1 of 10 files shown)


shikharj05:
Thank you for the change. Can you add tests as well?

rutuja-amazon:
Added unit tests.

cwperks:
@rutuja-amazon Should I add any backport labels to this PR?

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -133,6 +133,10 @@ public void testUserApi() throws Exception {
         response = rh.executePutRequest(ENDPOINT + "/internalusers/", "{\"hash\": \"123\"}", new Header[0]);
         Assert.assertEquals(HttpStatus.SC_METHOD_NOT_ALLOWED, response.getStatusCode());
 
+        // username has special characters
+        response = rh.executePutRequest(ENDPOINT + "/internalusers/n@ag:ilum", "{\"hash\": \"123\"}", new Header[0]);
+        Assert.assertEquals(HttpStatus.SC_METHOD_NOT_ALLOWED, response.getStatusCode());
```
cliu123, line 138: Is this test case specifically related to this fix? If yes, why isn't it also `HttpStatus.SC_BAD_REQUEST`.

rutuja-amazon: Yes, we need to updated all test cases that use special characters in `username` to now fail at creation

cliu123: Agreed. Then shouldn't this `PUT` request with special characters in `username` get `HttpStatus.SC_BAD_REQUEST`, which would be consistent with the fix?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -93,6 +94,12 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
             return;
         }
 
+        Pattern usernamePattern = Pattern.compile("[$&+,:;=\\\\?@#|/'<>.^*()%!-]");
```
shikharj05, line 97: I think ``.`` should be allowed

rutuja-amazon: Permitted `.`. Do we also allow `@` in username?

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
rutuja-amazon: Sounds good, updated to `HttpStatus.SC_BAD_REQUEST`.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
DarshitChanpura: `@` shouldn't be allowed and only one `.` should be allowed. Thoughts? @cwperks @peternied @davidlago

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -93,6 +94,12 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
             return;
         }
 
+        Pattern usernamePattern = Pattern.compile("[$&+,:;=\\\\?@#|/'<>^*()%!-]");
```
peternied, line 97: > Username should not include special characters

Why not, are there bugs associated with these characters?

peternied (changes requested):
This change impacts forward combability as username might have been viable in older versions of OpenSearch that will no longer work.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
rutuja-amazon: Do we also allow `-`, `_` and `@` along with `.`? Please suggest.
In the current fix I have permitted the four special characters above.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
peternied: Thanks for looking out for issues of this nature, could you be more specific as to the risks associated with individual characters?  I am reticent to limit these usernames without cause.

If there is an issue please follow our reporting process [1] also quoted below.

> ## Reporting a Vulnerability
>
>If you discover a potential security issue in this project we ask that you notify AWS/Amazon Security via our [vulnerability reporting page](http://aws.amazon.com/security/vulnerability-reporting/) or directly via email to aws-security@amazon.com. Please do **not** create a public GitHub issue.

[1] https://github.com/opensearch-project/security/security/policy

rutuja-amazon:
Requesting addition of backport labels @cwperks

DarshitChanpura:
@rutuja-amazon which branches would you like this change to be backported to?

peternied:
@rutuja-amazon Since this is change is based on a LOW security finding we should discuss everything in the public on this pull request.  Pulling in the context from our internal threads:

> ## Bug
> ### Repro 
> Create two users through the dashboards create internal user:
>   - username: hello password: world:a_great_day
>   - username: hello:world password: a_great_day
> 
> Attempt to logging to login via basic auth into both accounts
> curl -k -s https://hello:world:a_great_day@localhost:9200/_plugins/_security/api/account | jq
> curl -k -s https://hello:world:a_great_day@localhost:9200/_plugins/_security/api/account | jq
>
> You cannot login as the user `hello:world`.
>
> ### Additional details
> Note; this is expected as the username cannot contain a colon [1].
>
> As per CVSS rating (3.1), it's been categorized as LOW risk.
>
> [1] https://stackoverflow.com/questions/11612854/http-https-basic-authentication-colon-in-username


For the moment, lets block **only** the colon character.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
peternied: Added more details on the main thread https://github.com/opensearch-project/security/pull/2277#issuecomment-1333993589

peternied:
@rutuja-amazon I've created https://github.com/rutuja-amazon/security/pull/1, if you merge that pull request into this change we can get this issue fixed, what do you think?

rutuja-amazon:
> @rutuja-amazon I've created [rutuja-amazon#1](https://github.com/rutuja-amazon/security/pull/1), if you merge that pull request into this change we can get this issue fixed, what do you think?

Issue is with other special characters as well, as discussed offline

peternied:
I've created an issue to track the set of special characters that should be included or not, and at this time the only exclusion that seems justified is colon `:`.  Please feel free to discuss on the issue, but as this pull request stands its too restrictive without justification for me approve and merge.

- https://github.com/opensearch-project/security/issues/2309

DarshitChanpura (approved):

rutuja-amazon:
Updated PR to only allow `:` for this use-case. Requesting merge of this PR with backported labels.

cwperks (commented):
The changes look good to me, once the code hygiene check is fixed I will approve this PR.

peternied:
> The changes look good to me, once the code hygiene check is fixed I will approve this PR.

I'll take care of this to help get this merged more quickly...

cwperks (approved):

peternied:
Note; `CI / build-artifact-names (pull_request)` will fail because this hasn't merged the latest from main, and I'm am attempting to avoid rebasing these changes.

peternied (approved):

peternied:
Failing CI check is known fixed in main, merging!

Issue #2305, opened 2022-12-06
Windows start up issues?


peternied:
Plugin install workflow seems to be seeing issues with Windows JDK 11/17.  After several rerun attempts did not see any altered behavior and additional logging seems to indicate the security index isn't being created in a timely fashion.  

#### Failure
From https://github.com/opensearch-project/security/actions/runs/3625326485/jobs/6113250804
```
Invoke-WebRequest: D:\a\_temp\2d6b1604-a1a3-410d-a119-3204283a8b44.ps1:6
Line |
   6 |  Invoke-WebRequest -SkipCertificateCheck -Uri 'https://localhost:9200/ …
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | OpenSearch Security not initialized.
```

#### Logs after node start
```
[2022-12-06T00:43:22,300][INFO ][o.o.s.OpenSearchSecurityPlugin] [fv-az206-877] Node started
[2022-12-06T00:43:22,304][INFO ][o.o.s.c.ConfigurationRepository] [fv-az206-877] Will not attempt to create index .opendistro_security and default configs if they are absent. Use securityadmin to initialize cluster
[2022-12-06T00:43:22,304][INFO ][o.o.s.OpenSearchSecurityPlugin] [fv-az206-877] 0 OpenSearch Security modules loaded so far: []
[2022-12-06T00:43:22,304][INFO ][o.o.s.c.ConfigurationRepository] [fv-az206-877] Background init thread started. Install default config?: false
[2022-12-06T00:43:22,304][INFO ][o.o.s.c.ConfigurationRepository] [fv-az206-877] Try to load config ...
[2022-12-06T00:43:22,304][INFO ][o.o.s.c.ConfigurationRepository] [fv-az206-877] security index not exists (yet)
Error: 2-06T00:43:22,304][ERROR][o.o.s.c.ConfigurationLoaderSecurity7] [fv-az206-877] Failure no such index [.opendistro_security] retrieving configuration for [INTERNALUSERS, ACTIONGROUPS, CONFIG, ROLES, ROLESMAPPING, TENANTS, NODESDN, WHITELIST, ALLOWLIST, AUDIT] (index=.opendistro_security)
[...]

PR #2308, opened 2022-12-07
Windows Install Action Fix

scrawfor99:
### Description
This fix resolves the issue with the Windows install. The -y flag in the demo scripts was not actually setting the config settings correctly. 

### Issues Resolved
Resolves issue #2305 

### Testing
I ran the full workflow suite and everything worked. 

### Check List

### DIFF
--- a/tools/install_demo_configuration.bat
+++ b/tools/install_demo_configuration.bat
@@ -62,6 +62,11 @@ if %cluster_mode% == 0 (
     )
 )
 
+if %assumeyes% == 1 (
+    set "initsecurity=1"
+    set "cluster_mode=1"
+)
+
 set BASE_DIR=%SCRIPT_DIR%\..\..\..\
 if not exist %BASE_DIR% (
     echo "basedir does not exist"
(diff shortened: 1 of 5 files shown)


tools/install_demo_configuration.bat
```diff
@@ -62,6 +62,11 @@ if %cluster_mode% == 0 (
     )
 )
 
+if %assumeyes% == 1 (
```
DarshitChanpura, line 65: for my knowledge, what would this do?

cwperks: It sets all of the individual flags to true, `assumeyes` mean yes to all of `initsecurity`, `clustermode` and `skip_updated` (which is already defaulted to true).

Good catch @scrawfor99 !

cwperks (approved):

peternied (approved):
Nice catch!

tools/install_demo_configuration.bat
cwperks: No is the default, if `-y` is not specified then you get prompted for input to choose. 

`-y` advertises that it accepts all prompts, but its not working as advertised without this fix.

scrawfor99:
For some reason JDK 11 Linux build fails and says there is an issue with the sinkProvider. Was this an issue we are aware of?

cwperks:
I think its flaky. I see `TimeoutException[Timeout after 10SECONDS while retrieving configuration for [ROLESMAPPING](index=.opendistro_security)];` in the logs.

tools/install_demo_configuration.bat
DarshitChanpura: ah..gotcha.. thank you @cwperks!

PR #2318, opened 2022-12-09
Fix windows action backport to be callable from other repos

scrawfor99:
### Description
Addresses changes that standardize the plugin install workflow across versions and makes older versions compatible with being called from other repos. 

### Check List


peternied (approved):

peternied (commented):
Oops, why do other repos need to call the 2.4 version instead of the main version?  IMO this action should be viable no matter the version of the plugin you are installing

scrawfor99:
Right now there is not a building 3.0 distribution so in order to call the workflow properly we need to backport the changes to a version that is building correctly. I see your point however and I think that this can be closed if the 2.4 version is merged.

PR #2334, opened 2022-12-13
Changing logging type to give warning for basic auth with no creds

abhivka7:
### Description
* Changing logging type to give warning for basic auth with no creds


src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -267,7 +267,7 @@ public boolean authenticate(final RestRequest request, final RestChannel channel
 
                 if(authDomain.isChallenge() && httpAuthenticator.reRequestAuthentication(channel, null)) {
                     auditLog.logFailedLogin("<NONE>", false, null, request);
-                    log.trace("No 'Authorization' header, send 401 and 'WWW-Authenticate Basic'");
+                    log.warn("No 'Authorization' header, send 401 and 'WWW-Authenticate Basic'");
```
peternied, line 270: This is a potential denial of surface vector since unauthenticated callers will be able to force this message into the log - leading to full disks.  I would much rather this was configurable for OpenSearch administrators, via the log4j configuration file or a configuration setting.  What do you think?

peternied (changes requested):
Could you provide more background on this change, I do not see a linked issue and the pull request description template wasn't filled out.  

Also - please make sure to include the DCO verification on your commits.

abhivka7:
Opened new pull request: https://github.com/opensearch-project/security/pull/2347

src/main/java/org/opensearch/security/auth/BackendRegistry.java
abhivka7: Ideally logs should reveal if unauthenticated callers  are trying to login. Currently there is zero visibility in it.
Regarding disks being full, log rotation should take care of it.

Issue #2346, opened 2022-12-14
[BUG] Basic auth with no creds does not give a warning.


abhivka7:
**What is the bug?**
Using basic auth with no creds should give a warning in logs. But right now the logging level is trace and not warn.

**How can one reproduce the bug?**
1. Create an opensearch domain with basic auth, and try to do a get request on it without passing credentials. It will just send the message: "Unauthorized" and only log if trace logging is enabled.

**What is the expected behavior?**
As an admin, I don't have any visibility into why my users are not able to login without enabling [trace logs](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L270) 

**What is your host/environment?**
 - Security plugin enabled with OpenSearch 1.3 domain

DarshitChanpura:
[Triage] @abhivka7 Thank you for filing this bug. Could you please elaborate more on the use case that you have in mind for this scenario?

cwperks:
Closing this issue.

This has been merged into main and 2.x:

- https://github.com/opensearch-project/security/pull/2347
- https://github.com/opensearch-project/security/pull/2364

PR #2347, opened 2022-12-14
Changing logging type to give warning for basic auth with no creds

abhivka7:
Signed-off-by: Abhi Kalra <abhivka@amazon.com>

### Description
* Category 
Bug Fix

Changing logging type to give warning for basic auth with no creds.

When using basic auth without credentials, users just get a message telling "Unauthorized" which is not very descriptive. As an admin, I don't have any visibility into why my users are not able to login without enabling this log which is currently at trace level. Hence, changing log level to warn.

### Issues Resolved
https://github.com/opensearch-project/security/issues/2346

We want to backport it to latest OS versions.

 Signed-off-by: Abhi Kalra <abhivka@amazon.com>

### DIFF
--- a/src/main/java/org/opensearch/security/auth/BackendRegistry.java
+++ b/src/main/java/org/opensearch/security/auth/BackendRegistry.java
@@ -267,7 +267,7 @@ public class BackendRegistry {
 
                 if(authDomain.isChallenge() && httpAuthenticator.reRequestAuthentication(channel, null)) {
                     auditLog.logFailedLogin("<NONE>", false, null, request);
-                    log.trace("No 'Authorization' header, send 401 and 'WWW-Authenticate Basic'");
+                    log.warn("No 'Authorization' header, send 401 and 'WWW-Authenticate Basic'");
                     return false;
                 } else {
                     //no reRequest possible


scrawfor99 (approved):
This seems reasonable.

scrawfor99 (changes requested):
In light of the previous PR's coversation (https://github.com/opensearch-project/security/pull/2334), I think it may be necessary to first implement a precaution against DoS attacks or justification for why this change cannot wait until such a precaution is in place.

abhivka7:
> plement a precaution against DoS attacks

Security plugin does not cover any implementation of precaution against DoS attack. That has to be done from infrastructure change. if needed we can create a backlog to see what can be done from our end. 
Anyways, having a log to tell if there is an attack is still needed and therefore the PR should be merged.

davidlago:
I'm struggling to see how this is any different than logging any other messages for unauthenticated requests. If you get millions of them, yes, your logs are filling up but also you have other resources under strain as well that will probably be an issue and/or be detected and acted upon (and you might find those logs useful after all?)

Based on @peternied 's comment [here](https://github.com/opensearch-project/security/pull/2334#discussion_r1047333931), perhaps the compromise is to let this type of error log level be configurable, default it to current level and let people override. Right now it seems like OP _is_ already resorting to setting it to trace to be able to see these, which is an even worse scenario if we're thinking of DoS, so adding this as a configurable level will consider all users' needs while preserving our current default behavior.

wdyt?

varun-lodaya:
I think we should decouple the problem of logs filling up disk from adding more logs, since there are tons of places where we log lack of FGAC permissions at request level or SAML JWT failures for every request. Eg - 

- https://github.com/opensearch-project/security/blob/b836040c577894a08db9377d46a3d4eaca426fb8/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java#L173
- https://github.com/opensearch-project/security/blob/f431ec2201e1466b7c12528347a1f54cf64387c9/src/main/java/org/opensearch/security/configuration/PrivilegesInterceptorImpl.java#L73

Moreover, as @davidlago  mentioned, if a user is sending millions of request, they are probably already stressing other system resources.

cwperks (approved):

scrawfor99 (approved):

RyanL1997 (approved):

PR #2350, opened 2022-12-14
Upgrade CXF to 3.5.5 to address CVE-2022-46363

scrawfor99:
### Description
Upgrade CXF to 3.5.5 to address CVE-2022-46363

### Check List


TRIAGING.md
```diff
@@ -20,7 +20,7 @@ If you have an issue you'd like to bring forth please consider getting a link to
 
 ### Is there an agenda for each week?
 
-Meetings are lightly structured as follows: 
+Meetings are structured as follows:
```
DarshitChanpura, line 23: This is good catch. But the general idea of the PR is to keep the changes in the PR to the expected scope. This could be a good separate PR along-with any other changes that you find in the document.

In this particular case we should remove this change from here as 2.x line doesn't contain TRIAGING.md and so the auto-cut backport might fail.

build.gradle
```diff
@@ -332,7 +332,7 @@ dependencies {
 
     runtimeOnly 'net.minidev:accessors-smart:2.4.7'
 
-    runtimeOnly 'org.apache.cxf:cxf-core:3.4.5'
+    runtimeOnly 'org.apache.cxf:cxf-core:3.5.5'
```
DarshitChanpura, line 335: we should update all cxf related dependencies here

DarshitChanpura (changes requested):

TRIAGING.md
scrawfor99: Yeah, it was just a change I needed to push because the workflow did not run for some reason the first push.

build.gradle
scrawfor99: Good catch--fixed.

TRIAGING.md
scrawfor99: Removed.

DarshitChanpura (approved):

cwperks (approved):

ulir:
Hi,
quick question: this ticket has the `backport 2.4` label - does this mean there's going to be a 2.4.1 release containing the fix?
Couldn't sort this out reading https://opensearch.org/releases.html - from what I read there, there's not going to be any 2.4.z releases as soon as 2.5.0 is out?
Thanks for a quick clarification!
-Uli

scrawfor99:
Hi @ulir,

You are correct in your understanding of the backporting structure. In short, we backport fixes to the latest version of all supported majors versions. So since `main` tracks for the next major (3.0.0), we also backport to `2.4` and `1.3` for any potential patch releases (which would be semantically versioned as `2.4.x` or `1.3.x`). Then because we also want to include any changes in an upcoming version of the supported majors which are still under development, we want to backport to `2.x` as well for a `2.5` version. Because we are never going to have a `1.4`, backporting to `1.x` is not strictly necessary but you may see instances where changes are also backported to this branch as well.

PR #2375, opened 2023-01-04
When excluding fields also exclude the term + `.keyword`

peternied:
### Description
When excluding fields also exclude the term + `.keyword`

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/configuration/DlsFlsFilterLeafReader.java
+++ b/src/main/java/org/opensearch/security/configuration/DlsFlsFilterLeafReader.java
@@ -142,6 +142,7 @@ class DlsFlsFilterLeafReader extends SequentialStoredFieldsLeafReader  {
 
                 if (firstChar == '!' || firstChar == '~') {
                     excludesSet.add(incExc.substring(1));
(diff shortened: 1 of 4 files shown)


cwperks (approved):

src/main/java/org/opensearch/security/configuration/DlsFlsFilterLeafReader.java
```diff
@@ -142,6 +142,7 @@ class DlsFlsFilterLeafReader extends SequentialStoredFieldsLeafReader  {
 
                 if (firstChar == '!' || firstChar == '~') {
                     excludesSet.add(incExc.substring(1));
+                    excludesSet.add(incExc.substring(1) + KEYWORD);
```
DarshitChanpura, line 145: For my knowledge: What would this change accomplish?

peternied: When documents are parsed fields are mapped to types, for string mappings a keyword field is automatically created.  By additional excluding keyword fields if you wanted to block a field `actors`, we would also block `actors.keyword` to ensure those field values are not included in queries.

PR #2384, opened 2023-01-06
Add a test in order to catch incorrect handling of index parsing during Snapshot Restoration

### LINKED ISSUE #1652: [BUG] Restoring snapshot: indices exclusion triggers security_exception (creating OK, listing OK)
**Describe the bug**
I'm using the S3 repository plugin to store snapshots. I've have tested this previously with the same scripting, but now I can't restore snapshots anymore. Don't know the exact cause, two things have changed: I have moved to 1.2.4 and I have moved from SAML to openid. The calls for the snapshot create/restore/list/etc are still being done through basic auth. 

The strange thing is, I can list all the snapshots, I can create snapshots, I just can't restore them. I'm not including global state or the security index:

Error:

```
{
    "error": {
        "root_cause": [
            {
                "type": "security_exception",
                "reason": "no permissions for [] and User [name=admin_XXXX, backend_roles=[admin], requestedTenant=null]"
            }
        ],
        "type": "security_exception",
        "reason": "no permissions for [] and User [name=admin_XXXX, backend_roles=[admin], requestedTenant=null]"
    },
    "status": 403
}
```

Listing works:

image

Creating snapshot works:

image

Restoring fails:

image

all_access is mapped to the admin backend role:

image

**Expected behavior**
Snapshot is restored

**Plugins**
Default docker 1.2.4 plus s3 repo plugin
Also tried default docker 1.1 plus s3 repo plugin

**Screenshots**

**Host/Environment (please complete the following information):**
Docker 1.2.4 image on kubernetes

**Additional context**

scrawfor99:
### Description
Further addresses issue #1652 by adding a Security-side test which checks that the behavior of SnapshotRestore is consistent no matter the order of positive or negative wildcards. 

This test mirrors the parsing test added to the core codebase but is able to confirm that the issue is resolved and will likely catch other issues which could happen between index resolution and the SnapshotRestore operations. 

### Issues Resolved
#1652 

### Check List


peternied (commented):
Looks good - before I approve, could you tell me that the test fails when run against an older version of OpenSearch?

scrawfor99:
@peternied, I just checked it on 1.3 and 2.1 and it passed so I inverted the test case because I realized that checking for something was not going to work when the issue was treating things like wildcards when they shouldn't be. I swapped it over to now assert that the result does not contain "foo" or "bar" but does contain "baz" and that seems to have fixed the issue. I would have liked to one line it but this way it is easier to read. 

Proof 2.1 fails: 
Screen Shot 2023-01-06 at 4 53 42 PM

src/test/java/org/opensearch/security/SnapshotRestoreTests.java
```diff
@@ -231,106 +235,145 @@ public void testSnapshotCheckWritePrivileges() throws Exception {
...
+        String wild_first_body = rh.executePostRequest("_snapshot/all/all_1/_restore?wait_for_completion=true","{\"indices\": \"b*,-bar\", \"rename_pattern\": \"(.+)\", \"rename_replacement\": \"wild_first_restored_index_$1\"}", encodeBasicHeader("nagilum", "nagilum")).getBody();
+        assert(!wild_first_body.contains("wild_first_restored_index_foo"));
+        assert(!wild_first_body.contains("wild_first_restored_index_bar"));
+        assert(wild_first_body.contains("wild_first_restored_index_baz"));
+        String neg_first_body = rh.executePostRequest("_snapshot/all/all_1/_restore?wait_for_completion=true","{\"indices\": \"-bar,b*\", \"rename_pattern\": \"(.+)\", \"rename_replacement\": \"negate_first_restored_index_$1\"}", encodeBasicHeader("nagilum", "nagilum")).getBody();
+        assert(!neg_first_body.contains("negate_first_restored_index_foo"));
+        assert(!neg_first_body.contains("negate_first_restored_index_bar"));
+        assert(neg_first_body.contains("negate_first_restored_index_baz"));
```
peternied, line 324: Nit: `assertThat` provides much cleaner output, and IMO is also much easier to read

```java
assertThat(neg_first_body, not(containsString("negate_first_restored_index_foo")));
assertThat(neg_first_body, containsString("negate_first_restored_index_baz"));
```

More reasons I like these matchers https://stackoverflow.com/a/1701143/533057

peternied (changes requested):
Since you are writing more interesting validation maybe we can clean that up to easier to read

src/test/java/org/opensearch/security/SnapshotRestoreTests.java
```diff
@@ -231,106 +240,145 @@ public void testSnapshotCheckWritePrivileges() throws Exception {
...
+            tc.admin().cluster().createSnapshot(new CreateSnapshotRequest("all", "all_1").indices(listOfIndexesToTest).includeGlobalState(false).waitForCompletion(true)).actionGet();
+       }
+
+        RestHelper rh = nonSslRestHelper();
+
+        Assert.assertEquals(HttpStatus.SC_OK, rh.executePostRequest("_snapshot/all/all_1/_restore?wait_for_completion=true","{\"indices\": \"b*,-bar\", \"rename_pattern\": \"(.+)\", \"rename_replacement\": \"wild_first_restored_index_$1\"}", encodeBasicHeader("nagilum", "nagilum")).getStatusCode());
+        Assert.assertEquals(HttpStatus.SC_OK, rh.executePostRequest("_snapshot/all/all_1/_restore?wait_for_completion=true","{\"indices\": \"-bar,b*\", \"rename_pattern\": \"(.+)\", \"rename_replacement\": \"neg_first_restored_index_$1\"}", encodeBasicHeader("nagilum", "nagilum")).getStatusCode());
+        String wild_first_body = rh.executePostRequest("_snapshot/all/all_1/_restore?wait_for_completion=true","{\"indices\": \"b*,-bar\", \"rename_pattern\": \"(.+)\", \"rename_replacement\": \"wild_first_restored_index_$1\"}", encodeBasicHeader("nagilum", "nagilum")).getBody();
```
cwperks, line 322: nit: please use lowerCamelCase for variable names

src/test/java/org/opensearch/security/SnapshotRestoreTests.java
```diff
@@ -231,106 +240,145 @@ public void testSnapshotCheckWritePrivileges() throws Exception {
...
+    public void testSnapshotRestoreSpecialIndicesPatterns() throws Exception {
+        // Run with ./gradlew test --tests org.opensearch.security.SnapshotRestoreTests.testSnapshotRestoreSpecialIndicesPatterns
+
+        final List<String> listOfIndexesToTest = Arrays.asList("foo", "bar", "baz");
+
+        final Settings settings = Settings.builder()
+                .putList("path.repo", repositoryPath.getRoot().getAbsolutePath())
+                .build();
+
+        setup(Settings.EMPTY, new DynamicSecurityConfig().setSecurityActionGroups("action_groups_packaged.yml"), settings, true, currentClusterConfig);
```
cwperks, line 304: is `.setSecurityActionGroups("action_groups_packaged.yml")` required here?

src/test/java/org/opensearch/security/SnapshotRestoreTests.java
```diff
@@ -231,106 +240,145 @@ public void testSnapshotCheckWritePrivileges() throws Exception {
...
+                        "\"indices\": \"testsnap2\","+
+                        "\"ignore_unavailable\": false,"+
+                        "\"include_global_state\": true"+
+                        "}";
+
+        Assert.assertEquals(HttpStatus.SC_OK, rh.executePutRequest("_snapshot/bckrepo/"+putSnapshot.hashCode()+"?wait_for_completion=true&pretty", putSnapshot, encodeBasicHeader("snapresuser", "nagilum")).getStatusCode());
         Assert.assertEquals(HttpStatus.SC_FORBIDDEN, rh.executePostRequest("_snapshot/bckrepo/"+putSnapshot.hashCode()+"/_restore?wait_for_completion=true&pretty","{ \"include_global_state\": true, \"rename_pattern\": \"(.+)\", \"rename_replacement\": \"restored_index_$1\" }", encodeBasicHeader("snapresuser", "nagilum")).getStatusCode());
     }
 
+    @Test
```
cwperks, line 294: Nice job writing a test for this in the security repo! This looks good to me. If we are to add more scenarios for testing than I would suggest breaking these down into individual tests.

cwperks (approved):
Looks good to me. Thank you @scrawfor99!

src/test/java/org/opensearch/security/SnapshotRestoreTests.java
scrawfor99: That makes sense. I will break it out if we add more. The current test framework is a little weird with three distinct formats depending on the era it was written.

scrawfor99:
@cwperks would you be able to merge this for me--I still cannot. Thank you!

cwperks:
Merged!

scrawfor99:
Thank you!

cwperks:
@andrross Can this be backported to 2.x?

PR #2409, opened 2023-01-17
Add action indices:admin/index_template/* to cluster_manage_index_tem…

### LINKED ISSUE #2407: [BUG] Default action group cluster_manage_index_templates should include permissions for composable index templates
**What is the bug?**
The default action group `cluster_manage_index_templates` currently only contains the permission `indices:admin/template/*` for legacy templates. It should also include `indices:admin/index_template/*` to support composable templates.

1. Create a role with `cluster_permissions` set to `cluster_manage_index_templates`
2. Create a user with that role
3. Send a PutIndexTemplateRequest to OpenSearch
4. See error `[security_exception] no permissions for [indices:admin/index_template/put]`

Composable templates can be managed by users with `cluster_manage_index_templates` permission.

 - 2.4.1

srt:
…plates (#2407)

Signed-off-by: Stefan Reuter <stefan.reuter@reucon.com>

### Description
Adds permissions to manage composable templates to the default action group  `cluster_manage_index_templates`.

### Issues Resolved
#2407 

### Check List

### DIFF
--- a/src/main/resources/static_config/static_action_groups.yml
+++ b/src/main/resources/static_config/static_action_groups.yml
@@ -209,6 +209,7 @@ cluster_manage_index_templates:
   allowed_actions:
   - "indices:admin/template/*"
   - "indices:admin/index_template/*"
+  - "cluster:admin/component_template/*"
   type: "cluster"
   description: "Manage index templates"
 cluster_manage_pipelines:
--- a/src/main/resources/static_config/static_action_groups.yml
+++ b/src/main/resources/static_config/static_action_groups.yml
@@ -208,6 +208,7 @@ cluster_manage_index_templates:
   static: true
   allowed_actions:
   - "indices:admin/template/*"
+  - "indices:admin/index_template/*"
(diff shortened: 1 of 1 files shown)


peternied (commented):
@srt Thanks for the contribution, I just started the CI workflow and myself and other maintainers will review your change in detail.

cwperks (approved):

peternied (changes requested):
Do you know how `cluster:admin/component_template` fits into this space, it seems like it could be related [1], what do you think about including it as well?

- [1] https://opensearch.org/docs/latest/opensearch/index-templates/#composable-index-templates

DarshitChanpura:
I agree with @peternied 's comments above. As `_component_template`s can be used to compose an index template, it might be a good step to include that action here.

srt:
Good point, I have added cluster:admin/component_template/*

peternied (approved):
Thanks!

RyanL1997 (approved):

PR #2428, opened 2023-02-03
Fix testCheckForDeprecatedSettingFoundLegacy was broken with changes in calling patterns

peternied:
### Description
We've seen failures with this mock test cases seeing different parameters.  This failure was completely blocking https://github.com/opensearch-project/security/pull/2427

I don't see an obvious change that could cause this, it could be changes in compiler settings of how the 'params' argument was treated differently, or a shift in Mockito version from core.

### Testing
Reproduced locally, and then with this fix it has been resolved

### Check List

### DIFF
--- a/src/test/java/org/opensearch/security/setting/DeprecatedSettingsTest.java
+++ b/src/test/java/org/opensearch/security/setting/DeprecatedSettingsTest.java
@@ -58,6 +58,6 @@ public class DeprecatedSettingsTest {
 
         checkForDeprecatedSetting(settings, "legacyKey", "properKey");
 
-        verify(logger).deprecate(eq("legacyKey"), anyString(), any());
+        verify(logger).deprecate(eq("legacyKey"), anyString(), any(), any());
     }
 }


cwperks (approved):

cwperks:
The bwc failures are related to the build issues on the 2.x line. Those cannot be resolved until this change is backported to 2.x.

DarshitChanpura:
BWC tests will be fixed once an artifact is available with #2427 merged. Is that right?

DarshitChanpura (approved):

RyanL1997 (approved):
Should we also backport this fix?

cwperks:
@DarshitChanpura Not with new artifacts, the bwcTests directly call on `./gradlew assemble` to assemble the specific version being tested. See: https://github.com/opensearch-project/security/blob/main/.github/actions/create-bwc-build/action.yaml

If the other PR merges first and these tests are re-ran then I expect them to pass.

scrawfor99 (approved):

peternied:
Thanks for jumping on this team, as soon as the standard CI clears I'll override and merge

PR #2444, opened 2023-02-17
Dynamically Enable/Disable multi-tenancy, private tenant and set Defa…

abhivka7:
…ult Tenant

### Description
* Category : Enhancement
These changes are required to dynamically enable/disable multi-tenancy, private tenant and set a default tenant.
Earlier users had to change YAMl files in each node and restart node to enable/disable multi-tenancy and private tenant. With our changes, users will be able to change these settings dynamically. We will also give admins the option to set a default tenant for all users for first time log in.
More details can be found here:
https://github.com/opensearch-project/OpenSearch/issues/5853
https://github.com/opensearch-project/security-dashboards-plugin/issues/1302

### Issues Resolved
https://github.com/opensearch-project/security-dashboards-plugin/issues/1302

Nope

### Testing
We have done manual testing for all changes on our local server. We have also written some unit tests for our api changes. We will write more unit tests and integration tests based on reviewer feedbacks soon.

### Check List
- [ ✅ ] New functionality includes testing
- [] New functionality has been documented. : Will document in further commits.
- [ ✅ ] Commits are signed per the DCO using --signoff


src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -136,10 +136,11 @@ public void run() {
                                     ConfigHelper.uploadFile(client, cd+"roles_mapping.yml", securityIndex, CType.ROLESMAPPING, DEFAULT_CONFIG_VERSION);
                                     ConfigHelper.uploadFile(client, cd+"internal_users.yml", securityIndex, CType.INTERNALUSERS, DEFAULT_CONFIG_VERSION);
                                     ConfigHelper.uploadFile(client, cd+"action_groups.yml", securityIndex, CType.ACTIONGROUPS, DEFAULT_CONFIG_VERSION);
+                                    final boolean populateEmptyIfFileMissing = true;
```
scrawfor99, line 139: Should this be moved to a config constants file?

src/main/java/org/opensearch/security/dlic/rest/api/TenancyConfigAction.java
```diff
@@ -0,0 +1,104 @@
...
+    public List<Route> routes() {
+        return allRoutes;
+    }
+
+    @Override
+    protected void handleGet(RestChannel channel, RestRequest request, Client client, final JsonNode content) throws IOException{
+
+        final SecurityDynamicConfiguration<?> configuration = load(getConfigName(), true);
+        filter(configuration);
+        successResponse(channel,configuration);
```
scrawfor99, line 65: If you want to do anything special for responses you will need to add `onResponse` logic here.

scrawfor99 (commented):
Looks good, just a couple comments and general questions. I left a note for cleaning the style stuff. Thank you for you contribution.

src/main/java/org/opensearch/security/configuration/PrivilegesInterceptorImpl.java
```diff
@@ -95,9 +96,10 @@ private boolean isTenantAllowed(final ActionRequest request, final String action
      */
     @Override
     public ReplaceResult replaceDashboardsIndex(final ActionRequest request, final String action, final User user, final DynamicConfigModel config,
+                                                final TenancyConfigModel tenancyconfig,
                                                 final Resolved requestedResolved, final Map<String, Boolean> tenants) {
 
-        final boolean enabled = config.isDashboardsMultitenancyEnabled();//config.dynamic.kibana.multitenancy_enabled;
+        final boolean enabled = tenancyconfig.isDashboardsMultitenancyEnabled();
```
anijain-Amazon, line 102: nit: Camel Case

src/main/java/org/opensearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -481,6 +482,7 @@ protected void successResponse(RestChannel channel) {
 		try {
 			final XContentBuilder builder = channel.newBuilder();
 			builder.startObject();
+			builder.endObject();
```
anijain-Amazon, line 485: Were we not calling `builder.startObject();` before ?

src/main/java/org/opensearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -109,20 +109,21 @@ protected void handleApiRequest(final RestChannel channel, final RestRequest req
 		try {
 			// validate additional settings, if any
 			AbstractConfigurationValidator validator = getValidator(request, request.content());
-			if (!validator.validate()) {
+			if (validator!= null && !validator.validate()) {
```
anijain-Amazon, line 112: If we change the condition to `validator == null || !validator.validate()` then the use of checks in the lines below is not required.

src/main/java/org/opensearch/security/dlic/rest/api/TenancyConfigAction.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    protected void handleDelete(RestChannel channel, final RestRequest request, final Client client, final JsonNode content) throws IOException{
+        notImplemented(channel, Method.DELETE);
+    }
+
+    @Override
+    protected void handlePut(RestChannel channel, final RestRequest request, final Client client, final JsonNode content) throws IOException{
+        successResponse(channel);
```
anijain-Amazon, line 81: If this is not doing anything, shouldn't it be `notImplemented` as well ?

src/main/java/org/opensearch/security/tools/SecurityAdmin.java
```diff
@@ -1382,7 +1387,7 @@ private static boolean validateConfigFile(String file, CType cType, int version)
 
 	private static String[] getTypes(boolean legacy) {
 		if (legacy) {
-			return new String[]{"config", "roles", "rolesmapping", "internalusers", "actiongroups", "nodesdn", "audit"};
+			return new String[]{"config", "roles", "rolesmapping", "internalusers", "actiongroups", "nodesdn", "audit","tenancyconfig"};
```
cwperks, line 1390: This legacy block is only for CType for ES6. This can be removed from this block.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -136,10 +136,11 @@ public void run() {
...
                                     if(DEFAULT_CONFIG_VERSION == 2) {
                                         ConfigHelper.uploadFile(client, cd+"tenants.yml", securityIndex, CType.TENANTS, DEFAULT_CONFIG_VERSION);
+                                        ConfigHelper.uploadFile(client, cd+"tenancy_config.yml", securityIndex, CType.TENANCYCONFIG, DEFAULT_CONFIG_VERSION, populateEmptyIfFileMissing);
```
cwperks, line 142: Should this only be populated if the file exists? I see `tenants.yml` does not create this config entry if no `tenants.yml` file is supplied.

src/main/java/org/opensearch/security/dlic/rest/api/AbstractApiAction.java
cwperks, line 112: I see that this PR introduces a validator. When will the validator ever be null? Are the changes in this file needed?

cwperks (commented):
@abhivka7 Thank you for the PR! Overall the PR looks good, but I wanted to run an idea by you to ensure that introducing a new `tenantconfig.yml` file is the right way to go.

I see that `config.yml` already provides:

```
@Override
public boolean isDashboardsMultitenancyEnabled() {
    return config.dynamic.kibana.multitenancy_enabled;
}
```

was any consideration given to also adding `private_tenant_enabled` and `default_tenant` to this section of `config.yml`? I think that could potentially simplify this PR. The beginning of `config.yml` could look like:

```
config:
  dynamic:
    # Set filtered_alias_mode to 'disallow' to forbid more than 2 filtered aliases per index
    # Set filtered_alias_mode to 'warn' to allow more than 2 filtered aliases per index but warns about it (default)
    # Set filtered_alias_mode to 'nowarn' to allow more than 2 filtered aliases per index silently
    #filtered_alias_mode: warn
    #do_not_fail_on_forbidden: false
    multitenancy:
      enabled: true
      private_tenant_enabled: true
      default_tenant: ''
```

then to get these into `SecurityInfoAction` you could follow a pattern similar to https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L152-L169 to subscribe to changes in this config.

src/main/java/org/opensearch/security/dlic/rest/api/TenancyConfigAction.java
```diff
@@ -0,0 +1,104 @@
...
+
+public class TenancyConfigAction extends PatchableResourceApiAction{
+
+    private static final List<Route> allRoutes = new ImmutableList.Builder<Route>()
+            .addAll(addRoutesPrefix(
+                    ImmutableList.of(
+                            new Route(Method.GET, "/tenancyconfig/"),
+                            new Route(Method.PUT, "/tenancyconfig/"),
+                            new Route(Method.PATCH, "/tenancyconfig/")
+                    )
```
shikharj05, line 43: Is DELETE not required?

src/main/java/org/opensearch/security/securityconf/TenancyConfigModel.java
```diff
@@ -0,0 +1,26 @@
...
+public class TenancyConfigModel {
+    private final TenancyConfigV7 tenancyConfig;
+
+    public TenancyConfigModel(TenancyConfigV7 tenancyConfig) {
+        this.tenancyConfig = tenancyConfig;
+    }
+
+    public boolean isDashboardsMultitenancyEnabled() { return this.tenancyConfig.multitenancy_enabled; };
+    public boolean isDashboardsPrivateTenantEnabled() { return this.tenancyConfig.private_tenant_enabled; };
+    public String dashboardsDefaultTenant() { return this.tenancyConfig.default_tenant; };
```
shikharj05, line 24: nit: rename to getDashboardsDefaultTenant

src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java
```diff
@@ -184,6 +189,7 @@ public void onChange(Map<CType, SecurityDynamicConfiguration<?>> typeToConfig) {
         final WhitelistingSettings whitelist = (WhitelistingSettings) cr.getConfiguration(CType.WHITELIST).getCEntry("config");
         final AllowlistingSettings allowlist = (AllowlistingSettings) cr.getConfiguration(CType.ALLOWLIST).getCEntry("config");
         final AuditConfig audit = (AuditConfig)cr.getConfiguration(CType.AUDIT).getCEntry("config");
+        final TenancyConfigModel tcm ;
```
shikharj05, line 192: nit: extra space

src/main/java/org/opensearch/security/securityconf/impl/v7/TenancyConfigV7.java
```diff
@@ -0,0 +1,16 @@
+package org.opensearch.security.securityconf.impl.v7;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+public class TenancyConfigV7 {
```
shikharj05, line 5: implement a toString() method?

src/test/java/org/opensearch/security/dlic/rest/api/TenancyConfigApiTest.java
```diff
@@ -0,0 +1,93 @@
...
+        response = rh.executeGetRequest(ENDPOINT);
+        Assert.assertEquals(HttpStatus.SC_UNAUTHORIZED, response.getStatusCode());
+
+        rh.sendHTTPClientCredentials = true;
+        response = rh.executeGetRequest(ENDPOINT);
+        Assert.assertEquals(HttpStatus.SC_FORBIDDEN, response.getStatusCode());
+    }
+
+    @Test
+    public void testTenancyConfigAPIUpdate() throws Exception {
```
shikharj05, line 47: these can be split into different tests for each property

shikharj05 (commented):
Thank you for the work, added few comments.

config/tenancy_config.yml
```diff
@@ -0,0 +1,8 @@
+_meta:
+  type: "tenancyconfig"
+  config_version: 2
+
+tenancy_config:
+  multitenancy_enabled: true
+  private_tenant_enabled: true
+  default_tenant: ""
```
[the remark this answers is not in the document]
abhivka7: Thanks for the sugestion @scrawfor99 . I used `./gradlew spotlessApply` as per your suggestion. It corrected style for some files but not tenancy_config.yml. Do you need any specific change in this file that you want me to do manually?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
abhivka7: Sure will do in next commit. Thanks for the suggestion.

src/main/java/org/opensearch/security/configuration/PrivilegesInterceptorImpl.java
abhivka7: Thanks @anijain-Amazon for finding this miss. Will update in next commit.

src/main/java/org/opensearch/security/dlic/rest/api/AbstractApiAction.java
abhivka7: This was added while testing and I forgot to remove it. Thanks @anijain-Amazon  for pointing this out. Will remove this check condition in next commit.

abhivka7: This was added while testing and I forgot to remove it. Thanks @cwperks  for pointing this out. Will remove this check condition in next commit.

src/main/java/org/opensearch/security/dlic/rest/api/AbstractApiAction.java
abhivka7: No we were not doing it. That was a miss I guess.

src/main/java/org/opensearch/security/dlic/rest/api/TenancyConfigAction.java
abhivka7: I don't think there will be a use case where we require DELETE. We should not delete tenancy_config and we can replace existing config using PATCH. 
Correct me if I am wrong @devardee

src/main/java/org/opensearch/security/dlic/rest/api/TenancyConfigAction.java
abhivka7: @scrawfor99 We do not want to do anything special with our responses. I don't think we need that logic here.

src/main/java/org/opensearch/security/dlic/rest/api/TenancyConfigAction.java
abhivka7: We want to enable PUT request for tenancyConfig. That's why we are not using `notImplemented ` .

src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java
abhivka7: Thanks @shikharj05 for pointing this out. Will fix in next commit.

src/main/java/org/opensearch/security/securityconf/TenancyConfigModel.java
abhivka7: Ok. Will change in next commit.

src/main/java/org/opensearch/security/securityconf/impl/v7/TenancyConfigV7.java
abhivka7: Sure. Will do in next commit.

src/main/java/org/opensearch/security/tools/SecurityAdmin.java
abhivka7: Thanks for pointing this out @cwperks. Will remove this in next commit.

src/test/java/org/opensearch/security/dlic/rest/api/TenancyConfigApiTest.java
abhivka7: Thanks for the suggestion @shikharj05 . I will split these and update in next commit.

cwperks:
@abhivka7 @shikharj05 See comment above. Have you considered doing this inside of `config.yml` under `config.dynamic` instead of introducing a new config file? 

i.e.

```
config:
  dynamic:
    multitenancy:
      enabled: true
      private_tenant_enabled: true
      default_tenant: ''
```

abhivka7:
> @abhivka7 Thank you for the PR! Overall the PR looks good, but I wanted to run an idea by you to ensure that introducing a new `tenantconfig.yml` file is the right way to go.
> 
> I see that `config.yml` already provides:
> 
> ```
> @Override
> public boolean isDashboardsMultitenancyEnabled() {
>     return config.dynamic.kibana.multitenancy_enabled;
> }
> ```
> 
> was any consideration given to also adding `private_tenant_enabled` and `default_tenant` to this section of `config.yml`? I think that could potentially simplify this PR. The beginning of `config.yml` could look like:
> 
> ```
> config:
>   dynamic:
>     # Set filtered_alias_mode to 'disallow' to forbid more than 2 filtered aliases per index
>     # Set filtered_alias_mode to 'warn' to allow more than 2 filtered aliases per index but warns about it (default)
>     # Set filtered_alias_mode to 'nowarn' to allow more than 2 filtered aliases per index silently
>     #filtered_alias_mode: warn
>     #do_not_fail_on_forbidden: false
>     multitenancy:
>       enabled: true
>       private_tenant_enabled: true
>       default_tenant: ''
> ```
> 
> then to get these into `SecurityInfoAction` you could follow a pattern similar to https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L152-L169 to subscribe to changes in this config.

@cwperks We were initially working on the same approach that you suggested. But going forward, we plan on introducing some other fields and features for multi-tenancy. Therefore we plan on separating this file and introducing a new one.

src/main/java/org/opensearch/security/configuration/PrivilegesInterceptorImpl.java
cliu123: With this change, would `config.dynamic.kibana.multitenancy_enabled` be ignored? Would this work in mixed clusters? Based on my understanding, security index is node-scoped. If there're 2 settings(the old and the new), could there be 2 different values(`false` and `true`) of multitenancy_enabled setting in a mixed cluster?

src/main/java/org/opensearch/security/rest/TenantInfoAction.java
```diff
@@ -106,7 +106,7 @@ public void accept(RestChannel channel) throws Exception {
                 try {
 
                     final User user = (User)threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
-                    
+
```
cliu123, line 109: Please remove format-only changes.

src/main/java/org/opensearch/security/securityconf/impl/SecurityDynamicConfiguration.java
```diff
@@ -52,7 +52,7 @@
 import org.opensearch.security.securityconf.StaticDefinable;
 
 public class SecurityDynamicConfiguration<T> implements ToXContent {
-    
+
```
cliu123, line 55: Please remove format-only changes.

src/main/java/org/opensearch/security/securityconf/impl/SecurityDynamicConfiguration.java
```diff
@@ -87,7 +87,7 @@ public static <T> SecurityDynamicConfiguration<T> fromJson(String json, CType ct
         } else {
             sdc = new SecurityDynamicConfiguration<T>();
         }
-        
+
```
cliu123, line 90: Please remove format-only changes.

src/main/java/org/opensearch/security/securityconf/impl/v7/ConfigV7.java
```diff
@@ -138,6 +138,7 @@ public static class Kibana {
         public String server_username = "kibanaserver";
         public String opendistro_role = null;
         public String index = ".kibana";
+
```
cliu123, line 141: Please remove format-only changes.

src/test/java/org/opensearch/security/test/DynamicSecurityConfig.java
```diff
@@ -150,8 +151,12 @@ public List<IndexRequest> getDynamicConfig(String folder) {
                     .id(CType.TENANTS.toLCString())
                     .setRefreshPolicy(RefreshPolicy.IMMEDIATE)
                     .source(CType.TENANTS.toLCString(), FileHelper.readYamlContent(prefix+securityTenants)));
-        }
```
cliu123, line 153: Please remove format-only changes.

peternied (changes requested):
Consider reusing the existing configuration and configuration API for tenancy instead of introducing a new one. This would simplify the design and avoid adding complexity. 

Could you explain why this new configuration approach was chosen?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
abhivka7: It was already always true. We just moved the line of declaring this variable. From next commit onwards, we can change this from config constants file.

src/main/java/org/opensearch/security/configuration/PrivilegesInterceptorImpl.java
abhivka7: @cliu123 going forward we will only use `config.dynamic.kibana.multitenancy_enabled` and this will get priority over `config.dynamic.kibana.multitenancy_enabled`.  

> If there're 2 settings(the old and the new), could there be 2 different values(false and true) of multitenancy_enabled setting in a mixed cluster?

In this case we will give priority to new settings.

RyanL1997, line 102: Hi @abhivka7, thanks for this contribution. I see the point that you are trying to extend the original multi-tenancy feature with this change. So do we need a feature flag for this dynamically enable/disable multi-tenancy feature? Or it will be applied directly by enable multi tenancy?

devardee:
> Consider reusing the existing configuration and configuration API for tenancy instead of introducing a new one. This would simplify the design.
> 
> Could you explain why this new configuration approach was chosen?

We want to decouple tenancy related configuration from securityconfig for the following reasons:

1. Securityconfig currently contains information about the authc and authz aspect of the Security Plugin, We believe that Tenancy as such is not related to plugins security model (since, it is used to enhance OSD user experience) hence we are planning on moving tenancy configuration to a separate document inside security-index.

2. If we include the new Tenancy fields in Securityconfig we need to add the changes in both ConfigModelV7 and ConfigModelV6 and also think about the migrate api, having a new SecurityDynamicConfiguration Model for Tenancy configuration on the other hand will not need us to worry about the migrate api, duplicating code and possibility on introducing breaking changes.

3. All future enhancements to Tenancy will go into the Tenancy configuration document instead of securityconfig doc

cwperks:
@devardee, there is already a multi-tenancy key (`config.dynamic.kibana.multitenancy_enabled`) in `config.yml` so the coupling is already there. 

I'm not sure I follow the concerns on the migrate API.

----

For this section I am assuming the new section in `config.yml` could be under `osd` instead of `kibana`:

```
config:
  dynamic:
    osd:
      multitenancy_enabled: true
      private_tenant_enabled: true
      default_tenant: ''
```

What I am thinking is changing the `DynamicConfigModel` to add 2 additional abstract methods:

```
public abstract boolean isPrivateTenantEnabled();
public abstract String getDefaultTenant();
```

then inside `DynamicConfigModelV7` the implementation for `public abstract boolean isDashboardsMultitenancyEnabled();` can be updated and implementations for the new methods could be implemented as follows:

```
@Override
public boolean isDashboardsMultitenancyEnabled() {
    return config.dynamic.osd.multitenancy_enabled || config.dynamic.kibana.multitenancy_enabled;
}

@Override
public boolean isPrivateTenantEnabled() {
    return config.dynamic.osd.private_tenant_enabled;
}

@Override
public boolean getDefaultTenant() {
    return config.dynamic.osd.default_tenant;
}
```

This new section could be created in `ConfigModelV7` and omitted from `ConfigModelV6`. The corresponding implementations in DynamicConfigModelV6 could return the default values. 

Would this approach be feasible? I think what's presented in this PR will work, but I also like the idea of less configuration files.

peternied:
Thank you for the reply @abhivka7 

> Securityconfig currently contains information about the authc and authz aspect of the Security Plugin, We believe that Tenancy as such is not related to plugins security model (since, it is used to enhance OSD user experience) hence we are planning on moving tenancy configuration to a separate document inside security-index.

There is an existing setting already in the [security configuration](https://github.com/opensearch-project/security/blob/main/config/config.yml#L70), removing/invalidating this setting is a breaking change - that is a con with this proposal.

> If we include the new Tenancy fields in Securityconfig we need to add the changes in both ConfigModelV7 and ConfigModelV6 and also think about the migrate api, having a new SecurityDynamicConfiguration Model for Tenancy configuration on the other hand will not need us to worry about the migrate api, duplicating code and possibility on introducing breaking changes.

I'm not sure how this is a problem.  Could you provide more background on what the migrate api is and how it is impacted with this change vs reusing the existing config?

> All future enhancements to Tenancy will go into the Tenancy configuration document instead of securityconfig doc

Could you provide reference to what future enhancements are?  Without more context this argument does not inform the current choice.

I am in aligned with Craig's [proposal](https://github.com/opensearch-project/security/pull/2444#issuecomment-1440606625) that reusing the existing systems.

src/main/java/org/opensearch/security/dlic/rest/api/TenancyConfigAction.java
DarshitChanpura: For my understanding, why is it returning successResponse directly?

src/main/java/org/opensearch/security/rest/SecurityInfoAction.java
```diff
@@ -112,6 +112,9 @@ public void accept(RestChannel channel) throws Exception {
                     builder.field("principal", (String)threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_SSL_PRINCIPAL));
                     builder.field("peer_certificates", certs != null && certs.length > 0 ? certs.length + "" : "0");
                     builder.field("sso_logout_url", (String)threadContext.getTransient(ConfigConstants.SSO_LOGOUT_URL));
+                    builder.field("tenancy_enabled", evaluator.multitenancyEnabled());
```
DarshitChanpura, line 115: the name `tenancy_enabled` is slightly misleading. Can we have `multi_tenancy_enabled` instead?

src/main/java/org/opensearch/security/securityconf/impl/v7/TenancyConfigV7.java
DarshitChanpura, line 5: Why the name V7 here? There is only one version of this class. We can move this class out of v7 folder and rename it to `TenancyConfig`

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -522,9 +529,13 @@ public Set<String> getAllConfiguredTenantNames() {
 
     public boolean multitenancyEnabled() {
         return privilegesInterceptor.getClass() != PrivilegesInterceptor.class
-                && dcm.isDashboardsMultitenancyEnabled();
+                && tcm.isDashboardsMultitenancyEnabled();
```
DarshitChanpura, line 532: Can we remove this method from DynamicConfigModel and its inheritor DynamicConfigModelV7 since it is not used anymore?

src/test/java/org/opensearch/security/dlic/rest/api/TenancyConfigApiTest.java
DarshitChanpura: I would suggest to keep it in one test, as the test sets up a cluster before executing the test and tears down after. This will impact on the performance of the tests.

DarshitChanpura (commented):
Thank you @abhivka7 for adding this PR. One generic comment is to add a brief javadoc for the new class added and check/add a new-line at the end of all modified files.

peternied:
Checking back in with this pull request.  There are still a number of outstanding comments.  I have written up a proposal that I think might simplify the issues around proxying config data from place to place.  I'll see about getting a proof of concept up and running as I think it would be a far better model isolating intention from the storage medium of the configuration value.

Love to get feedback on this - thanks!

- https://github.com/opensearch-project/security/issues/2559

src/main/java/org/opensearch/security/configuration/PrivilegesInterceptorImpl.java
abhivka7: We don't meed a flag for these changes. We will keep multi-tenancy enabled for new domains and changes can be applied directly.

src/main/java/org/opensearch/security/rest/SecurityInfoAction.java
abhivka7: Ok will make this change.

peternied:
I have created a draft pull request featuring a version of this dynamic configuration that would support this scenario and short up the limitations that have been called in the comments.

- https://github.com/opensearch-project/security/pull/2568

abhivka7:
Created another PR for these changes: https://github.com/opensearch-project/security/pull/2607. Closing this one.

PR #2459, opened 2023-02-22
Support multitenancy for the anonymous user

sduenas:
### Description

* Category Bug fix

Multitenancy for the anonymous user is not supported. It's implied in Opensearch-Dashboards that this isn't the expected behaviour. From the UI, anonymous users can switch between tenants but in the end all the operations are run on the
Global tenant (even when the UI says the user is in a different tenant).

The current version of the plugin supports multitenancy for all the users but not for the anonymous user.
The header 'securitytenant' is ignored when this user is authorized. Therefore, all the operations run
with it, use the default tenant - normally the 'Global' tenant -.
This patch fixes this issue assigning the selected tenant - defined by the 'securitytenant' header -
to the anonymous user.

### Issues Resolved

https://github.com/opensearch-project/security-dashboards-plugin/issues/1336

### Testing
- unit testing
- manual testing

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/auth/BackendRegistry.java
+++ b/src/main/java/org/opensearch/security/auth/BackendRegistry.java
@@ -338,8 +338,12 @@ public class BackendRegistry {
             }
 
             if(authCredenetials == null && anonymousAuthEnabled) {
-            	threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_USER, User.ANONYMOUS);
-            	auditLog.logSucceededLogin(User.ANONYMOUS.getName(), false, null, request);
+                final String tenant = Utils.coalesce(request.header("securitytenant"), request.header("security_tenant"));
+                User anonymousUser = new User(User.ANONYMOUS.getName(), new HashSet<String>(User.ANONYMOUS.getRoles()), null);
+                anonymousUser.setRequestedTenant(tenant);
+
+                threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_USER, anonymousUser);
+                auditLog.logSucceededLogin(anonymousUser.getName(), false, null, request);
                 if (isDebugEnabled) {
                     log.debug("Anonymous User is authenticated");
                 }
--- a/src/test/java/org/opensearch/security/multitenancy/test/MultitenancyTests.java
+++ b/src/test/java/org/opensearch/security/multitenancy/test/MultitenancyTests.java
@@ -437,4 +437,35 @@ public class MultitenancyTests extends SingleClusterTest {
         assertThat(res.getStatusCode(), equalTo(HttpStatus.SC_OK));
         assertThat(res.findValueInJson("_source.tenant"), equalTo(tenantNameAppended));
     }
+
+    @Test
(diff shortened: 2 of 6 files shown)


scrawfor99 (approved):
Thank you for this PR. Looks good to me.

cwperks (commented):
@sduenas Thank you for the PR. Could you add a test for this fix?

sduenas:
> @sduenas Thank you for the PR. Could you add a test for this fix?

Sure, I just was wondering what the best place for the test would be. I wasn't sure what would be the best place for that. I see there are tests about multi-tenancy but there aren't unit test for `BackendRegistry` which is the class that I modified. If someone of the maintainers or core developers can give me some advice, I wouldn't mind to add them.

Btw, tests are failing due to the bug reported in #2462. I mainly tested this code with 2.x but I thought it was worth to integrated it with `main` and later be backported to 2.x.

cwperks:
@sduenas Possibly in [AnonymousAuthenticationTest](https://github.com/opensearch-project/security/blob/main/src/integrationTest/java/org/opensearch/security/http/AnonymousAuthenticationTest.java)?

RyanL1997 (commented):
Hi @sduenas, thanks for the fix. As @cwperks mentioned above, a test case will make this pr ready to merge. In generally, the fix looks good to me!

sduenas:
@cwperks @RyanL1997 I've updated the PR with some basic tests. I finally added the tests to `MultitenancyTests.java` because I wasn't sure how to configure a local cluster with the anonymous access active in the integration tests. I hope that's ok.
I also rebase the PR with the latest commits on `main` branch.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -338,8 +338,12 @@ public boolean authenticate(final RestRequest request, final RestChannel channel
             }
 
             if(authCredenetials == null && anonymousAuthEnabled) {
-            	threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_USER, User.ANONYMOUS);
-            	auditLog.logSucceededLogin(User.ANONYMOUS.getName(), false, null, request);
+                final String tenant = Utils.coalesce(request.header("securitytenant"), request.header("security_tenant"));
```
cwperks, line 341: Thank you for adding a test! The test looks good to me. One last question, what happens if the request does not contain the securitytenant header? Should it skip setting the requested tenant or is it ok that the value is set with empty/null?

sduenas: The behavior is the same of any other user. If you check the code for authenticated user, this header will be null if it's not set for that user. I don't mind to change it, I was just trying to be consistent with the same pattern used for the authenticated user.

cwperks (approved):

DarshitChanpura (approved):
ty for this contribution @sduenas !

DarshitChanpura:
CI failures would be fixed via #2451. We can retry failed tasks once an artifact with that PR becomes available.

RyanL1997 (approved):

scrawfor99 (approved):

PR #2482, opened 2023-02-24
Clock skew tolerance for oidc token validation

### LINKED ISSUE #2439: BadCredentialsException: The token cannot be accepted yet
Description:
The current JWT validation process does not account for a time window for the "not before" (nbf) claim. This can cause authentication issues when there is a slight time difference between token generation and validation systems (few seconds).

Solution:
Add a configurable time window (in seconds) for nbf claim validation to allow for slight time differences. This will ensure that the user is authenticated even if the time difference is within the specified window.

actually the  clockOffset used is zero 
JwtUtils.validateJwtNotBefore(claims, 0, false); 

https://github.com/opensearch-project/security/blob/07cf4bb5166733f89cc8454d244fc2d4ffc81a4e/src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/JwtVerifier.java#L112

AndreaPQ:
### Description
Added new oidc authenticator configuration parameter "clockSkewToleranceSeconds"  for both
jwt nbf and exp validation 

* Category : Enhancement
This will ensure that the user is authenticated even if there is a slight time difference between token generation and validation systems.
old behavior: If the token generation and validation systems have a system clock misalignment, even by a few seconds, the user will not be authenticated.
new behavior: If the systems have a system clock misaligned within the tolerance parameter, the user is correctly authenticated

### Issues Resolved
#2439 

NO

### Testing
Unit test added

### Check List
- [X] New functionality has been documented (added new parameter in reference configuration)


config/config.yml
```diff
@@ -131,6 +131,7 @@ config:
             signing_key: "base64 encoded HMAC key or public RSA/ECDSA pem key"
             jwt_header: "Authorization"
             jwt_url_parameter: null
+            clockSkewToleranceSeconds: 300
```
scrawfor99, line 134: Nit: What do you think about using underscores to match the rest of the parameters in this section?

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java
```diff
@@ -108,20 +111,94 @@ public void testRoles() throws Exception {
...
 	}
 
+
 	@Test
 	public void testExp() throws Exception {
 		Settings settings = Settings.builder().put("openid_connect_url", mockIdpServer.getDiscoverUri()).build();
 
 		HTTPJwtKeyByOpenIdConnectAuthenticator jwtAuth = new HTTPJwtKeyByOpenIdConnectAuthenticator(settings, null);
-
+
```
scrawfor99, line 120: Nit: Could you remove these empty spaces in the line break?

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java
```diff
@@ -108,20 +111,94 @@ public void testRoles() throws Exception {
...
+		AuthCredentials creds = jwtAuth.extractCredentials(
+				new FakeRestRequest(ImmutableMap.of("Authorization", "bearer "+token),
+						new HashMap<String, String>()),
+				null);
+
+		Assert.assertNotNull(creds);
 	}
 
+	@Test
+	public void testNbf() throws Exception {
```
scrawfor99, line 154: Nit: Consider renaming this test to something like "testTooEarlyWithoutSkew". Just something that gives you an idea what it does from the name alone because in the logs, it will only show the test class and name.

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java
```diff
@@ -108,20 +111,94 @@ public void testRoles() throws Exception {
...
+		String token=TestJwts.createSigned(jwt_token, TestJwk.OCT_1);
+		AuthCredentials creds = jwtAuth.extractCredentials(
+				new FakeRestRequest(ImmutableMap.of("Authorization", "bearer "+token),
+						new HashMap<String, String>()),
+				null);
+
+		Assert.assertNull(creds);
+	}
+
+	@Test
```
scrawfor99, line 177: See above. I don't know what the best name for the tests would be but consider something that gives you an idea what they do.

scrawfor99 (commented):
Thank you for this PR. Overall, I think it is very strong. I left a few nit-picky comments which you are free to take or leave. I love that you put tests in and hope you will continue to contribute on more PRs later.

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java
AndreaPQ: the name of this test matches the already existing test "testExp"

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java
AndreaPQ: see above

config/config.yml
AndreaPQ: ok

DarshitChanpura: Why 300s here? Shouldn't a lower number, maybe 60s, be sufficient?

@cwperks @peternied Thoughts on this?

DarshitChanpura (commented):
This is a great contribution @AndreaPQ! Thank you for taking the initiative to fix the issue.
Thank you for adding the tests too. I only have one comment.

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
```diff
@@ -56,16 +56,20 @@ public abstract class AbstractHTTPJwtAuthenticator implements HTTPAuthenticator
     private final String subjectKey;
     private final String rolesKey;
 
+    public static final int DEFAULT_CLOCKSKEWTOLERANCESECONDS = 300;
```
peternied, line 59: Please upper snake case
```suggestion
    public static final int DEFAULT_CLOCK_SKEW_TOLERANCE_SECONDS = 300;
```

config/config.yml
AndreaPQ: I searched online for which default values are commonly used , e.g.
https://learn.microsoft.com/en-us/dotnet/api/microsoft.identitymodel.tokens.tokenvalidationparameters.defaultclockskew?view=msal-web-dotnet-latest

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java
```diff
@@ -108,20 +111,94 @@ public void testRoles() throws Exception {
...
 		AuthCredentials creds = jwtAuth.extractCredentials(
 				new FakeRestRequest(ImmutableMap.of("Authorization", TestJwts.MC_COY_EXPIRED_SIGNED_OCT_1),
 						new HashMap<String, String>()),
 				null);
 
 		Assert.assertNull(creds);
+	}	
+
+	@Test
+	public void testExpInSkew() throws Exception {
```
peternied, line 129: Let's clean these tests up by refactoring the comment code.  I'd recommend creating a private function that accepts an expiration time and settings object as parameters with a return of AuthCredentials.

This should alter the test methods to be the setup for the settings, the action to generate the credentials object, and the verification messages.

config/config.yml
```diff
@@ -131,6 +131,7 @@ config:
             signing_key: "base64 encoded HMAC key or public RSA/ECDSA pem key"
             jwt_header: "Authorization"
             jwt_url_parameter: null
+            jwt_clock_skew_tolerance_seconds: 300
```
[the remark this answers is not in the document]
AndreaPQ: I searched online for which default values are commonly used , e.g.
https://learn.microsoft.com/en-us/dotnet/api/microsoft.identitymodel.tokens.tokenvalidationparameters.defaultclockskew?view=msal-web-dotnet-latest

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java
peternied: The name doesn't make sense even if you know the acronym, lets update all the names of the tests to describe what is being tested.

config/config.yml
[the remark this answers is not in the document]
DarshitChanpura: I see the recommended value as 5s with the range being 0-60s.
https://techdocs.akamai.com/api-definitions/docs/json-web-token-jwt-val#:~:text=Clock%20skew%20amount%20specifies%20the,recommended%20default%20value%20is%205.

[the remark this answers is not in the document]
peternied: This change seems in line with the RFC guidance [1] and as @AndreaPQ has pointed out Microsoft is using this value.  

I was unable to find counter-examples that indicated smaller time thresholds like 5 seconds, which is originally what came to my mind when considering the clock skew.  In absence of more information, I'd say let's keep it as implemented.

[1]
>  Implementers MAY provide for some small leeway, usually no more than a few minutes,
From https://www.rfc-editor.org/rfc/rfc7519#section-7.2

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/JwtVerifier.java
```diff
@@ -32,9 +32,11 @@ public class JwtVerifier {
 	private final static Logger log = LogManager.getLogger(JwtVerifier.class);
 
 	private final KeyProvider keyProvider;
+	private final int clockSkewToleranceSeconds;
 
-	public JwtVerifier(KeyProvider keyProvider) {
+	public JwtVerifier(KeyProvider keyProvider, int clockSkewToleranceSeconds ) {
```
cwperks, line 37: wdyt about passing this setting in directly into the call here: https://github.com/opensearch-project/security/blob/main/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java#L107 instead of constructing a verifier with this setting. The setting could then be passed into `validateClaims` where it is needed.

```
jwt = jwtVerifier.getVerifiedJwtToken(jwtString, clockSkewToleranceSeconds);
```

peternied (changes requested):
Thanks for the contribution!  See comments inline, lets get those tests cleaned up - should be pretty trivial.

config/config.yml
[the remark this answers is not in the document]
cwperks: If the current default is 0 then it should remain the same default after the change to avoid changing existing cluster configs to users unknowingly. This change will also need to be accompanied with documentation for the version its released in.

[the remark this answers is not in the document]
peternied: Nice find @DarshitChanpura.  @AndreaPQ as you reported this issue, do you know by how much the tokens were expired, would default to 5 seconds mitigated the problem? 

I think 5 seconds is a good default as it is fairly close to the original behavoir.  Since this value was added to the configuration it can be tweaked by cluster operators to find the 'right spot' for their operations

cwperks (commented):
Thank you for the contribution @AndreaPQ. This change looks near ready to me. A few minor comments.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/JwtVerifier.java
AndreaPQ: The key provider is also used to validate the token and is currently passed in the constructor

config/config.yml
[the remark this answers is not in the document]
AndreaPQ: I think we should consider the general case and take advantage of  evaluations made by others .
I propose 300 seconds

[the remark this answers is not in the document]
cwperks: 5 minutes is too large IMO. Having a configurable setting should solve the need for when the auth server and the OpenSearch server have clocks out of sync. 

To make sure I understand exactly what the skew will mean.

| OpenSearch Node | AuthServer |
| ------------------ | ----------- |
| 02-24-2023 14:00:00 (1677265200 epoch) | 02-24-2023 14:00:04 (1677265200 epoch - ahead of OpenSearch Node by 4 s) |                   

User logs in successfully to auth server and server gives back a JWT with the following claims:

```
{
  ...
  "nbf": 1677265204,
  "exp": 1677268804 // 1 hr later
}
```

The OpenSearch node receives it immediately, but since the `nbf` from the auth server token is in the future (from the OpenSearch node's perspective) it is not valid yet. (Current time in epoch on the OpenSearch node `1677265200`)

With the skew setting it will offset from the OpenSearch node's perspective and add to the current timestamp (`Instant.now() + <skew>`) when determining its clock's current time in epoch. With a 5 second offset the OpenSearch node would calculate the current time as `1677265205` and the token would be considered valid. Likewise, the skew is used when validating the expiry so it will shorten the length of the session.

Again, I think 5 minutes is too long for a default as a lot of clusters would effectively shorten sessions by 5 minutes if the clocks of the auth server and OpenSearch node are in sync. I believe the default should remain 0 or closer to 0 for a setting like skew.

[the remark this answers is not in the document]
AndreaPQ: If the Two system are in synch the session Will be longer by 5 minutes, the tollerance works different for nbf and exp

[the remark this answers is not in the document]
peternied: @cwperks Could you describe your desired outcome of this pull request?

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java
AndreaPQ: Sorry I don't understand your request, could you explain it ? thanks
Is it OK if I create this function in class TestJwts ?

```
	static String createMcCoySignedOct1(int nbf, int exp)
	{
		JwtToken jwt_token = create(
			MCCOY_SUBJECT, TEST_AUDIENCE, 
			ROLES_CLAIM, TEST_ROLES_STRING, 
			JwtConstants.CLAIM_NOT_BEFORE, nbf, 
			JwtConstants.CLAIM_EXPIRY, exp);

		return createSigned(jwt_token, TestJwk.OCT_1);
	}
```

AndreaPQ:
By the way I noticed that there is an error in the HTTPJwtKeyByOpenIdConnectAuthenticatorTest.testExp test

https://github.com/opensearch-project/security/blob/7547d035ff11a00257f6c7e8316499e2138790e8/src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java#L118

The Authorization header is lacking the "Bearer" prefix , so the test will always fail regardless of the expiry date of the token.

For reference, see here
https://github.com/opensearch-project/security/blob/7547d035ff11a00257f6c7e8316499e2138790e8/src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java#L69

Should I fix it in this PR or should I open a separate issue?

config/config.yml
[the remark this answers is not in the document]
AndreaPQ: The skew tolerance should be taken into account  when evaluating the expiration date of the token too, see 
https://www.rfc-editor.org/rfc/rfc7523#section-3  point 4

> 4. The JWT MUST contain an "exp" (expiration time) claim that
> limits the time window during which the JWT can be used.  The
> authorization server MUST reject any JWT with an expiration time
> that has passed, subject to allowable clock skew between
> systems.

at Curity they reccomend 30s
https://curity.io/resources/learn/jwt-best-practices/#9-dealing-with-time-based-claims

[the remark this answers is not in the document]
cwperks: I see, looking inside `JwtUtils` of jose (https://github.com/apache/cxf/blob/main/rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwt/JwtUtils.java#L44-L81) the skew is added to the OpenSearch server's current clock time when determining if the `nbf` claim is valid, but for expiry the skew is added to the `exp` claim from the token when comparing to the OpenSearch server current clock. 

From the curity link:

```
This should be values of a few seconds, and we don't recommend using more than 30 seconds for this purpose, as this would rather indicate problems with the server, not a common clock skew.
```

They mention a maximum of 30s and that greater clock skew is indicative of a bigger problem. wdyt about setting the default value for this setting to a value <30s?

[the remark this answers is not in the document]
peternied: Thanks, what do you think @AndreaPQ to setting the default value to 30 seconds?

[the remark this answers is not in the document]
AndreaPQ: both 300s and 30s are good alternatives; your  team should review what others are doing and make a choice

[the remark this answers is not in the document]
peternied: I've updated this PR to use 30s as the default.

peternied:
> Should I fix it in this PR or should I open a separate issue? @peternied ?

@AndreaPQ could you make a follow up PR - don't need an issue if you would rather just include the details in the description

peternied (approved):

Issue #2483, opened 2023-02-24
[BUG] When attempting to delete a system index as a non-super admin user, the response misleads and always contains the .opendistro_security index


cwperks:
**What is the bug?**

When trying to delete a system index like `.opendistro_knn_models` (defined [here](https://github.com/opensearch-project/security/blob/main/tools/install_demo_configuration.sh#L380)) the response always contains `.opendistro_security` even though that is not the index that is attempting to be deleted. The response is:

```
[ERROR][o.o.k.t.TrainingJobRunner] no permissions for [] and User [name=admin, backend_roles=[admin], requestedTenant=null]
[WARN ][o.o.s.p.SecurityIndexAccessEvaluator] indices:admin/delete for '.opendistro_security' index is not allowed for a regular user
```

The log statement in the code in [SecurityIndexAccessEvaluator](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java#L124) always logs the `securityIndex` instead of the index from the request.

PR #2525, opened 2023-03-08
Fix error message when system index is blocked

peternied:
### Description
There was a bug in the error message when interacting with a system index causing it to always return the security index instead of the index(es) that were blocked.

Also includes unit test cases to verify the SecurityIndexAccessEvaluator

### Issues Resolved
- Resolves https://github.com/opensearch-project/security/issues/2483

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
@@ -30,6 +30,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -47,7 +48,7 @@ import org.opensearch.tasks.Task;
 
 public class SecurityIndexAccessEvaluator {
     
-    protected final Logger log = LogManager.getLogger(this.getClass());
+    Logger log = LogManager.getLogger(this.getClass());
     
     private final String securityIndex;
     private final AuditLog auditLog;
@@ -103,7 +104,7 @@ public class SecurityIndexAccessEvaluator {
                     presponse.allowed = false;
                     return presponse.markComplete();
                 }
-            } else if (requestedResolved.getAllIndices().contains(securityIndex) || matchAnySystemIndices(requestedResolved)) {
+            } else if (matchAnySystemIndices(requestedResolved)) {
                 if(filterSecurityIndex) {
                     Set<String> allWithoutSecurity = new HashSet<>(requestedResolved.getAllIndices());
                     allWithoutSecurity.remove(securityIndex);
@@ -121,7 +122,8 @@ public class SecurityIndexAccessEvaluator {
                     return presponse;
                 } else {
                     auditLog.logSecurityIndexAttempt(request, action, task);
-                    log.warn("{} for '{}' index is not allowed for a regular user", action, securityIndex);
+                    final String foundSystemIndexes = getProtectedIndexes(requestedResolved).stream().collect(Collectors.joining(", "));
+                    log.warn("{} for '{}' index is not allowed for a regular user", action, foundSystemIndexes);
                     presponse.allowed = false;
                     return presponse.markComplete();
                 }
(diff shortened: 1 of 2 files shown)


src/test/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluatorTest.java
```diff
@@ -0,0 +1,152 @@
...
+        final PrivilegesEvaluatorResponse response = evaluator.evaluate(request, null, UNPROTECTED_ACTION, resolved, presponse);
+
+        verifyNoInteractions(presponse);
+        assertThat(response, is(presponse));
+
+        verify(log).isDebugEnabled();
+    }
+
+    @Test
+    public void disableCacheOrRealtimeOnSystemIndex() {
```
cwperks, line 97: Nice!

src/test/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluatorTest.java
```diff
@@ -0,0 +1,152 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class SecurityIndexAccessEvaluatorTest {
```
cwperks, line 43: This is a good addition. Nice suite of tests in this file.

cwperks (approved):

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -121,7 +122,8 @@ public PrivilegesEvaluatorResponse evaluate(final ActionRequest request, final T
                     return presponse;
                 } else {
                     auditLog.logSecurityIndexAttempt(request, action, task);
-                    log.warn("{} for '{}' index is not allowed for a regular user", action, securityIndex);
+                    final String foundSystemIndexes = getProtectedIndexes(requestedResolved).stream().collect(Collectors.joining(", "));
+                    log.warn("{} for '{}' index is not allowed for a regular user", action, foundSystemIndexes);
```
DarshitChanpura, line 126: `... '{}' indexes are not allowed...` Thoughts?

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -121,7 +122,8 @@ public PrivilegesEvaluatorResponse evaluate(final ActionRequest request, final T
                     return presponse;
                 } else {
                     auditLog.logSecurityIndexAttempt(request, action, task);
-                    log.warn("{} for '{}' index is not allowed for a regular user", action, securityIndex);
+                    final String foundSystemIndexes = getProtectedIndexes(requestedResolved).stream().collect(Collectors.joining(", "));
```
DarshitChanpura, line 125: 👌🏽

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -149,6 +151,14 @@ public PrivilegesEvaluatorResponse evaluate(final ActionRequest request, final T
     }
 
     private boolean matchAnySystemIndices(final Resolved requestedResolved){
-        return systemIndexEnabled && systemIndexMatcher.matchAny(requestedResolved.getAllIndices());
+        return !getProtectedIndexes(requestedResolved).isEmpty();
+    }
+
+    private List<String> getProtectedIndexes(final Resolved requestedResolved) {
+        final List<String> protectedIndexes = requestedResolved.getAllIndices().stream().filter(securityIndex::equals).collect(Collectors.toList());
```
DarshitChanpura, line 158: I'm slightly confused. Won't this filter only match `securityIndex` and collect them. I'm not quite sure how is it matching against all protected indices? Or is the goal to match against securityIndex only

peternied: The security index should always be checked, this statement will create a collection with 0 items, or 1 item being the security index, if it was found.  

In the following line, if there are system indexes (not the security index), those will be added to the results in the branch statement, so long as the configuration switch `systemIndexEnabled` is true.

DarshitChanpura (approved):

Issue #2545, opened 2023-03-10
[Question] What capabilities should a token that is passed to an extension have? Should it be a JWT?


peternied:
This issue should be used for detailing the options for the tokens that will be passed to extensions. This issue should have a comparison of the options and justifications for a decision.  

1. Restate problem
2. List capabilities required. 
3. Make a T-chart with pros and cons of options and check off capabilities.
4. Say best one.

RyanL1997:
### Token Based Authentication for Extensions ###

### Problem

The plugins can interact opensearch as the user. However, extensions do not have this capability. Extensions need a way to perform actions as the users. For achieving this, the extensions require token to make a request against opensearch cluster as the user.

### List capabilities/required - How is that token used?

1. Generated by security plugin
2. Included pass by Extension Rest Handler to extensions
3. Once it got passed, the extensions will store/used the token
    1. The extension should not be able to modify the token
    2. The extension should/should not(?) be able to read the content of the token
    3. For one requests, the token can be reused multiple times as the user
    4. The token will be expired once the client request is done

![Extensions for security](https://user-images.githubusercontent.com/109499885/225071058-492bff0e-93fa-4ff6-8229-e85ab6845c95.jpg)

4. When OS gets called by the token, it will act like a user, and the token identifies where that request comes from

### Pro/Cons

**Option 1: JWT**

One of the most common solutions used for representing claims that are to be transferred between two parties. JWT consists of three parts separated by dots: a header, a payload, and a signature. Payload contains the actual claims being transferred. The signature is used to verify the integrity of the token.

Pros: 

* JWT is already implemented in Security Plugin/Feature Identity
* JWT supports tamper protection/expiration
* JWT typically has smaller size, since it doesn’t include encryption

Cons: 

* Even JWT is using digital signature to verify its authenticity, its payload can be viewed by extension, we may need extra encryption to secure sensitive user information before we cross the trust boundary. 
* ~Payload size is limited~

**Option 2: JWE**

JWE is commonly used for encrypting a payload so that it can be securely transmitted between two parties. JWE Consists of five parts separated by dots: a header, a payload, a ciphertext, an initialization vector (IV), authentication tag, and optionally an encrypted key. The header contains information about the encrypted data. The IV is used to randomize the encryption and the authentication tag is used to verify the integrity of the data.

Pros:

* Encrypts the payload so that it can be transmitted securely between two parties.
* Supports encryption and decryption of data
* Supports multiple encryption algorithms

Cons: 

[...]

Issue #2536, opened 2023-03-09
[Question] How to support extensions parity for user information


peternied:
# How to support extensions parity for user information

This issue is meant to decide how information user should be provided to extensions based on exiting behavior for plugins

# Background
Many existing plugins reference common-utils [User](https://github.com/opensearch-project/common-utils/blob/main/src/main/java/org/opensearch/commons/authuser/User.java) object.  For plugins that implement security checks on resources these are the limits of the existing security information that is exposed.

```mermaid
sequenceDiagram
    title How Plugin access to User Context
    actor User
    box white OpenSearch Process
       participant RestController
       participant SecurityPlugin
       participant Plugin as Plugin
    end

    User->>+RestController: Send search request
    RestController->>+SecurityPlugin: Authenticate user
    SecurityPlugin-->>-RestController: Store user object in thread context
    RestController->>+Plugin: Invoke plugin method
    Plugin->>+RestController: Read user from thread context
    RestController-->>-Plugin: Return user object
    Plugin-->>User: Return search results
```

#### Example user object represented as JSON
```json
{
    "name": "UserNameOrUserEmailAddress"
    "backend_roles": ["BackendRoleFromLDAP", "BackendRoleFromSAML"]
    "roles": ["security_rest_api_access", "alerting_full_access"]
    "custom_attribute_names": ["attr1", "attr2"]
    "user_requested_tenant": "private"
}
```

# Problem
User context information is stored within the thread context.  Extensions should never have access to the thread context and there must be a way to reason decisions about user with existing data.

> NOTE: Exposing backend_roles / roles directly to an extension should not be done to prevent tight coupling and information disclosure, instead extensions should use make request to a checkPermissions API with an action name & resources.  This path is being built for parity to let plugins have a clear migration path to extensions.

# Option 0: Don't provide user information

In this option extensions do not have support for existing scenarios where reasoning is done about the user identity or resource checks.

For each plugin -> extension migration the plugin would need to find any dependencies on User data and remove them.  This might require modification or deprecation of features

### Pros:
 - Prevents exposing user information to extensions

### Cons
 - Feature deprecation or modification might be substantial effort
 - Regression in functionality for plugin users

# Option 1: Include user information on all requests to extensions

[...]

Issue #254, opened 2020-02-28
Refactor DefaultOpenDistroKeyStore initSSLConfig method


debjanibnrj:
DefaultOpenDistroKeyStore initializes the SSL Config and seems to have a lot of duplicate code for initializing the Keystore and Pemstore for the http and transport certificates. 

For more context see: https://github.com/opendistro-for-elasticsearch/security/pull/238#discussion_r384198532

PR #2567, opened 2023-03-17
[Security/Extension] JWT Vendor for extensions

RyanL1997:
### Description
Initial commit of JWT token generator for extensions
* Category New feature

### Issues Resolved
* Relate: https://github.com/opensearch-project/security/issues/2545
* Relate: Generate an auth token for an Extension Request (WIP)

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
+++ b/src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
@@ -1,0 +1,134 @@
+package org.opensearch.security.authtoken.jwt;
+
+import com.google.common.base.Strings;
+import org.apache.cxf.jaxrs.json.basic.JsonMapObjectReaderWriter;
+import org.apache.cxf.rs.security.jose.jwk.JsonWebKey;
+import org.apache.cxf.rs.security.jose.jwk.KeyType;
+import org.apache.cxf.rs.security.jose.jwk.PublicKeyUse;
+import org.apache.cxf.rs.security.jose.jws.JwsUtils;
+import org.apache.cxf.rs.security.jose.jwt.JoseJwtProducer;
+import org.apache.cxf.rs.security.jose.jwt.JwtClaims;
+import org.apache.cxf.rs.security.jose.jwt.JwtToken;
+
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.cxf.rs.security.jose.jwt.JwtUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.OpenSearchSecurityException;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.common.transport.TransportAddress;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.security.user.User;
+import org.opensearch.security.securityconf.ConfigModel;
+import org.opensearch.threadpool.ThreadPool;
+
+
+public class JwtVendor {
+    private static final Logger logger = LogManager.getLogger(JwtVendor.class);
+
+    private static JsonMapObjectReaderWriter jsonMapReaderWriter = new JsonMapObjectReaderWriter();
+
+    private JsonWebKey signingKey;
+    private ConfigModel configModel;
+    private ThreadContext threadContext;
+
+    static JsonWebKey createJwkFromSettings(Settings settings) throws Exception {
+        String exchangeKey = settings.get("exchange_key");
+
+        if (!Strings.isNullOrEmpty(exchangeKey)) {
+
+            JsonWebKey jwk = new JsonWebKey();
+
+            jwk.setKeyType(KeyType.OCTET);
+            jwk.setAlgorithm("HS512");
+            jwk.setPublicKeyUse(PublicKeyUse.SIGN);
+            jwk.setProperty("k", exchangeKey);
+
+            return jwk;
+        } else {
+            Settings jwkSettings = settings.getAsSettings("jwt").getAsSettings("key");
+
+            if (jwkSettings.isEmpty()) {
+                throw new Exception(
+                        "Settings for key exchange missing. Please specify at least the option exchange_key with a shared secret.");
+            }
+
+            JsonWebKey jwk = new JsonWebKey();
+
+            for (String key : jwkSettings.keySet()) {
+                jwk.setProperty(key, jwkSettings.get(key));
+            }
+
+            return jwk;
+        }
+    }
+
+    //Getting roles from User
+    public Map<String, String> prepareClaimsForUser(User user, ThreadPool threadPool) {
+        Map<String, String> claims = new HashMap<>();
+        this.threadContext = threadPool.getThreadContext();
+        final TransportAddress caller = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
+        Set<String> mappedRoles = mapRoles(user, caller);
+        claims.put("sub", user.getName());
+        claims.put("roles", String.join(",", mappedRoles));
+        return claims;
+    }
+    public Set<String> mapRoles(final User user, final TransportAddress caller) {
+        return this.configModel.mapSecurityRoles(user, caller);
+    }
+
+    private String createJwt(Map<String, String> claims, Settings settings) {
+        JoseJwtProducer jwtProducer = new JoseJwtProducer();
+        try {
+            signingKey = createJwkFromSettings(settings);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+
+        jwtProducer.setSignatureProvider(JwsUtils.getSignatureProvider(signingKey));
+        JwtClaims jwtClaims = new JwtClaims();
+        JwtToken jwt = new JwtToken(jwtClaims);
+
+        jwtClaims.setNotBefore(System.currentTimeMillis() / 1000);
+        long expiryTime = System.currentTimeMillis() / 1000 + (60 * 5);
+
+        if (claims.containsKey("sub")) {
+            jwtClaims.setSubject(claims.get("sub"));
+        } else {
+            throw new OpenSearchSecurityException("Cannot create jwt, 'sub' claim is required");
+        }
+
+        jwtClaims.setIssuedAt(Instant.now().toEpochMilli());
+
+        // TODO: Should call preparelaims();
+        if (claims.containsKey("roles")) {
+            jwtClaims.setProperty("roles", claims.get("roles"));
+        }
+
+        if (claims.containsKey("exp")) {
+            int customTime = Integer.parseInt(claims.get("exp"));
+            jwtClaims.setExpiryTime(System.currentTimeMillis() / 1000 + (60 * customTime));
+        } else {
+            jwtClaims.setExpiryTime(expiryTime);
+        }
+
+        String encodedJwt = jwtProducer.processJwt(jwt);
+
+        if (logger.isDebugEnabled()) {
+            logger.debug(
+                    "Created JWT: "
+                            + encodedJwt
+                            + "\n"
+                            + jsonMapReaderWriter.toJson(jwt.getJwsHeaders())
+                            + "\n"
+                            + JwtUtils.claimsToJson(jwt.getClaims())
+            );
+        }
+        return encodedJwt;
+    }
+}
--- a/src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
+++ b/src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
@@ -124,6 +124,7 @@ public class SecurityRestFilter {
                 if (!checkAndAuthenticateRequest(request, channel, client)) {
                     User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
                     if (userIsSuperAdmin(user, adminDNs) || (whitelistingSettings.checkRequestIsAllowed(request, channel, client) && allowlistingSettings.checkRequestIsAllowed(request, channel, client))) {
+                        //TODO: If the request is going to the extension, issue a JWT for authenticated user.
                         original.handleRequest(request, channel, client);
                     }
                 }


RyanL1997:
While we are having the discussion in this PR, I will create a META issue for **Generate an auth token for an Extension Request** and add a unit test.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -0,0 +1,151 @@
...
+        return this.configModel.mapSecurityRoles(user, caller);
+    }
+
+    public String createJwt(Map<String, String> claims) {
+
+        jwtProducer.setSignatureProvider(JwsUtils.getSignatureProvider(signingKey));
+        JwtClaims jwtClaims = new JwtClaims();
+        JwtToken jwt = new JwtToken(jwtClaims);
+
+        jwtClaims.setNotBefore(System.currentTimeMillis() / 1000);
```
peternied, line 113: Use dependency injection to replace `System.currentTimeMillis()` and `Instant.now()` references, so the expiry/nbe fields can all be unit tested

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -0,0 +1,151 @@
...
+    static JsonWebKey createJwkFromSettings(Settings settings) throws Exception {
+        String signingKey = settings.get("signing_key");
+
+        if (!Strings.isNullOrEmpty(signingKey)) {
+
+            JsonWebKey jwk = new JsonWebKey();
+
+            jwk.setKeyType(KeyType.OCTET);
+            jwk.setAlgorithm("HS512");
+            jwk.setPublicKeyUse(PublicKeyUse.SIGN);
```
peternied, line 71: Looks like the other option for this enum is ENCRYPT, by switching this enum are we making JWEs?

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -0,0 +1,151 @@
...
+
+public class JwtVendor {
+    private static final Logger logger = LogManager.getLogger(JwtVendor.class);
+
+    private static JsonMapObjectReaderWriter jsonMapReaderWriter = new JsonMapObjectReaderWriter();
+
+    private JsonWebKey signingKey;
+    private JoseJwtProducer jwtProducer;
+    private ConfigModel configModel;
+    private ThreadContext threadContext;
```
peternied, line 50: Let's remove these references, this Vendor should have as few implementation details about the security plugin as possible.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -0,0 +1,151 @@
...
+        Set<String> mappedRoles = mapRoles(user, caller);
+        claims.put("sub", user.getName());
+        claims.put("roles", String.join(",", mappedRoles));
+        return claims;
+    }
+    public Set<String> mapRoles(final User user, final TransportAddress caller) {
+        return this.configModel.mapSecurityRoles(user, caller);
+    }
+
+    public String createJwt(Map<String, String> claims) {
```
peternied, line 107: No property bags 👎 

What if this function signature was `String createToken(String issuer, String subject, String audience, int timeToExpireSeconds) {...}` I think that is all it needs everything else is defaults, what do you think?

peternied (commented):
Thanks for sending this out @RyanL1997

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -0,0 +1,151 @@
...
+            for (String key : jwkSettings.keySet()) {
+                jwk.setProperty(key, jwkSettings.get(key));
+            }
+
+            return jwk;
+        }
+    }
+
+    //TODO:Getting roles from User
+    public Map<String, String> prepareClaimsForUser(User user, ThreadPool threadPool) {
```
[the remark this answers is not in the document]
RyanL1997: Yes, we can continue our discussion over here: https://github.com/opensearch-project/security/issues/254. For this one, I was just write up the function we may need for implementing it. 100% we can change/remove it anytime. And also, this function should be locate in a separate class if we choose to go down this path.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -0,0 +1,151 @@
...
+
+            JsonWebKey jwk = new JsonWebKey();
+
+            jwk.setKeyType(KeyType.OCTET);
+            jwk.setAlgorithm("HS512");
+            jwk.setPublicKeyUse(PublicKeyUse.SIGN);
+            jwk.setProperty("k", signingKey);
+
+            return jwk;
+        } else {
```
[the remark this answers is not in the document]
cwperks: @peternied This `else` clause is what makes this flexible to support creating JWKs with any properties that are not the default HMAC SHA512 symmetric signing.

This would allow user's to configure asymmetric encryption for the JWT signing if desired.

i.e.

```
config:
    signing_key: base64.encode(<secret here>)
```

vs

```
config:
    key:
        kid: Gps4Ea8bRzBNXMrzE8ciJZKNrlTKPP2MPEBPDSUXPpo
        kty: RSA
        alg: RS256
        use: sig
        n: pGGGyC01Krfq4kR6ebiFm8L3OLdAIL7KCA4gw9iVCdo-12aAftxwTIfv59bhlktOlOhsTQ883wDn4XnquMUBW5DffZUXyf81wLP6aWR-iySANF7_bEnu-HFyl40X8QmpJImXADHjDL3D4C5ckhRqUnIqET3eQ6TWcWGnoEG6wpmE5UlZinB7koAFcLnucPcHBvLLvpMDKxN6GW6jjwn5PKQqfim5TF_xQCXlACfe-dd5x2ZVSzKmErfim-ZhLDr4D83kKSJjch7iROhs7sbh6bj_6OvIeiTDUHDN7dMZZJr-LCXyvRpJZVEZXXRlxgj9WV6UEq7UbKwmkc5653RBRw
        e: AQAB
        x5c:
            [
                "MIICmzCCAYMCBgGHCWjKXDANBgkqhkiG9w0BAQsFADARMQ8wDQYDVQQDDAZtYXN0ZXIwHhcNMjMwMzIyMTI1OTM1WhcNMzMwMzIyMTMwMTE1WjARMQ8wDQYDVQQDDAZtYXN0ZXIwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQCkYYbILTUqt+riRHp5uIWbwvc4t0AgvsoIDiDD2JUJ2j7XZoB+3HBMh+/n1uGWS06U6GxNDzzfAOfheeq4xQFbkN99lRfJ/zXAs/ppZH6LJIA0Xv9sSe74cXKXjRfxCakkiZcAMeMMvcPgLlySFGpScioRPd5DpNZxYaegQbrCmYTlSVmKcHuSgAVwue5w9wcG8su+kwMrE3oZbqOPCfk8pCp+KblMX/FAJeUAJ97513nHZlVLMqYSt+Kb5mEsOvgPzeQpImNyHuJE6GzuxuHpuP/o68h6JMNQcM3t0xlkmv4sJfK9GkllURlddGXGCP1ZXpQSrtRsrCaRznrndEFHAgMBAAEwDQYJKoZIhvcNAQELBQADggEBAFuycpje7FnIGBwurU/RwylGO1yx5NX/7LORv1q1fzaQoz4ti3BZTSwTM/K2NsWv3xJAYNm5sqL5CwcJ9PlSOVWRpUt0ce/zBQmAylYUWnfyym7p+JXV9317eT3BeKV04LfGTvVPSmFQRigOuyrihOQ7AQg8zFRJUWvfGFrt3Jl8XRQ0qZAFLCoi9177onEVtdCXSiyIdjIEFE8GTyeRyqm0ed7l8HyLRWIjXCud17qGxZyaL1VqiCHfJGgJBES2LqCau8vKqnN8sLO+gnj/jE8QsJQ6mF+kWKK1/JMUwWscrnsB+rxktttzqo/WfKQiB1CjmoGkYIBlMljLpdaUdQo="
            ]
        x5t: 3nw2MTcS2gwFcGdnlGaB0RPOYG8
        x5t#S256: 2VUpjQgUQW1TyPGP6PSt0wLUDkTINRqmJxLBr2F-Ps0
```

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
scrawfor99, line 107: I would probably use an explicit declaration of the arguments if possible.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -0,0 +1,151 @@
...
+            return jwk;
+        }
+    }
+
+    //TODO:Getting roles from User
+    public Map<String, String> prepareClaimsForUser(User user, ThreadPool threadPool) {
+        Map<String, String> claims = new HashMap<>();
+        this.threadContext = threadPool.getThreadContext();
+        final TransportAddress caller = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
+        Set<String> mappedRoles = mapRoles(user, caller);
```
scrawfor99, line 98: What is the caller being used for? Seems like we can extract the roles just from the user no?

scrawfor99 (commented):
Hi Ryan, thank you for putting this together. Overall, I don't think there was much that seemed like it needed re-visiting. The obvious questions will be around the claims we include in the tokens but I did not leave comments for those. Great job.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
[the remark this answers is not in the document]
peternied: I recommend we start exposing as little surface area as possible around these new features. 

As this is part of flow where generating the sensitive tokens e.g. on-behalf-of user tokens for extensions.  I would advocate that we minimize the risk of misconfiguration by locking down these values in code.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -0,0 +1,151 @@
...
+        return claims;
+    }
+    public Set<String> mapRoles(final User user, final TransportAddress caller) {
+        return this.configModel.mapSecurityRoles(user, caller);
+    }
+
+    public String createJwt(Map<String, String> claims) {
+
+        jwtProducer.setSignatureProvider(JwsUtils.getSignatureProvider(signingKey));
+        JwtClaims jwtClaims = new JwtClaims();
```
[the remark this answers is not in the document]
RyanL1997: Sounds good, will the `issuer` always be OpenSearch Core?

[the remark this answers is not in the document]
peternied: It should identify the cluster, while true for our deployments this won't change.

However, this is cheap/easy flexibility for enabling multi-tenant clusters down the road.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
RyanL1997: Just wanna have a quick confirm, should the `String audience` be the identity of extension?

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
RyanL1997: > Yes, we can continue our discussion over here: https://github.com/opensearch-project/security/issues/254. For this one, I was just write up the function we may need for implementing it. 100% we can change/remove it anytime. And also, this function should be locate in a separate class if we choose to go down this path.

This is one is also associated with roles we discussed above. And we can remove it anytime.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
RyanL1997: After check with @cwperks, I think the caller is for one of ways to map roles to host. It contains the hostname and ip address. But still, this entire role method should not be locate in this JwtVendor class in the future. Since we haven't map having a consensus on the solution of payload, I will just leave it here for now.

cwperks: @scrawfor99 caller is used when calculating the roles to use for privilege evaluation. See details on role resolution here: https://github.com/opensearch-project/security/issues/2573#issuecomment-1478172731

When serializing the User for common-utils it performs the roles resolution so that plugins can later on do roles injection to evaluate a user's permissions asynchronously.

When thinking of backwards compatibility of replacing a plugin as an extension, in order for it to be a drop in replacement we would need to perform the roles resolution before the request is sent to the extension and pass those roles somehow (could be through claims in the JWT as presented here)

This function serializes the Security plugin's User object onto a special header in the ThreadContext that is readable by common-utils: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L198-L210

The `mappedRoles` param would be the same as Ryan has here.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
peternied: Lets encrypt!

cwperks: FYI this field is used for good intentions and does not provide any functionality. See the RFC here: https://www.rfc-editor.org/rfc/rfc7517#section-4.2

i.e. this field indicates to parties that have the JWK what it's intended usage is.

peternied: Thanks for the detail @cwperks definitely not what I was hoping for.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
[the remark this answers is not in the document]
peternied: We've got another issue about user information parity, that might cover it, but I don't think it clearly calls out that we don't include this information inside the authentication token.  Or is there another issue where we have this called out, if not want to make one so we can track?
- https://github.com/opensearch-project/security/issues/2536

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
RyanL1997: Done.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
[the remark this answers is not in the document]
RyanL1997: Sure, thanks for the information! I will create separate issue attach to this for the tracking.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -0,0 +1,168 @@
...
+        try {
+            this.signingKey = createJwkFromSettings(settings);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+        this.jwtProducer = jwtProducer;
+        this.timeProvider = timeProvider;
+    }
+
+    static JsonWebKey createJwkFromSettings(Settings settings) throws Exception {
```
scrawfor99, line 77: Could you add the comments about the defaults you chose for this? It looks like you were onboard below, but I do not see the changes. Thank you :)

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
scrawfor99: This should be swapped over I believe.

src/test/java/org/opensearch/security/authtoken/jwt/JwtVendorTest.java
```diff
@@ -0,0 +1,80 @@
...
+        String subject = "admin";
+        String audience = "extension_0";
+        Integer expirySeconds = -300;
+
+        Settings settings =  Settings.builder().put("signing_key", "abc123").build();
+        JwtVendor jwtVendor = new JwtVendor(settings);
+
+        jwtVendor.createJwt(issuer, subject, audience, expirySeconds);
+    }
+}
```
scrawfor99, line 80: Would do you think about adding a test for a bad not before time? That is a common part of JWTs as I understand so we may consider having a test for the scenario. That being said it does not look like you ever set a not before time on creation so perhaps that is why you did not test it.

Do you think we should add a not before time or no? It is basically a marker for when the token becomes live. 

Overall great work on this! I think it is in a good spot and if you could add the comment about the default values and address Peter's other comment your thumbs-up'd I would be ready to approve.

Great work Ryan!

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
peternied: https://github.com/opensearch-project/security/issues/2545 has details about how the JWE would be implemented that have more nuanced even if the flag was all thats required.
 
I'm going to resolve this coversation it looks like my initial question `by switching this enum are we making JWEs?` is nope - more work is required

peternied:
Hey, @RyanL1997, I noticed that this code is generating a JWT, but it doesn't seem to be connected with any user scenarios. Can you provide some insight into what you had in mind for the evolution of this pull request so it would be merged?

To enhance the functionality of this pull request, I recommend that we add the ability to generate a validated token for an authenticated user and accept the token and validate it.

Let me know your thoughts on this.

src/test/java/org/opensearch/security/authtoken/jwt/JwtVendorTest.java
RyanL1997: That's actually a good one. Because for the current design `iat` and `nbf` should be the same. I'm pinging @cwperks and @peternied here to ask that, is there any extended usage of the claim `nbf`?

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
RyanL1997: Good catch! The line you are highlighting here is the one we use for the functionality. The other constructor with an additional parameter `final LongSupplier timeProvider` is for the test cases, such as the one I did for expected expiry, In the future, once we implement the verification of this token, we can also use this testing-only constructor to do some other test, such as the validation of the expiry. But yeah, I will leave a comment.

RyanL1997:
Hi, @peternied. I had a conversation with @cwperks today about the authentication backend. First for answering your questions:

> Can you provide some insight into what you had in mind for the evolution of this pull request so it would be merged

I think the purpose of this PR is to setup a basic structure of this JIT token, such as, some of the components we discussed above for token claims: subject, audience ..etc. I guess the only tricky part for now is about the how are we gonna deal with the roles, encrypted/not-encrypted/include/not-included. For these parts, we can discuss it in another review session with a documentations and modify them later, but I think we can use this basic structure for future development. For now, I just started to write a draft PR for authentication backend, and I do need some of the basic concepts from this PR for token verification.

src/test/java/org/opensearch/security/authtoken/jwt/JwtVendorTest.java
RyanL1997: @scrawfor99 After talking with @cwperks, I think the original purpose of this `nbf` claim is for issue a token for future usage, which is not our case, at least for now. But, I think it is ok to have it here, so that we can also have some test for the authentication backend to check. Still, we can always go back and change it later.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -0,0 +1,151 @@
...
+        String signingKey = settings.get("signing_key");
+
+        if (!Strings.isNullOrEmpty(signingKey)) {
+
+            JsonWebKey jwk = new JsonWebKey();
+
+            jwk.setKeyType(KeyType.OCTET);
+            jwk.setAlgorithm("HS512");
```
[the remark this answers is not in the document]
RyanL1997: Here is a simple T-chart of HMAC vs RSA based on the research I did:
| Name/Features |        Pros            |         Cons         |
| ----------------- | ----------------- | -----------------|
| HMAC (Hash-based Message Authentication Code) | - Performance: HMAC is generally faster than RSA, as it uses symmetric key cryptography, which is computationally less expensive. <br /> <br /> - Flexibility: HMAC can work with various hash functions, such as SHA-256, SHA-384, or SHA-512, depending on the desired security level and performance. | - Shared secret key: HMAC relies on a shared secret key between the sender and receiver. <br /> <br /> - Not suitable for non-repudiation, as anyone with the secret key can create a valid signature. |
|RSA (Rivest-Shamir-Adleman)| - Widely adopted: RSA is a widely-used and well-established public key cryptography algorithm, with extensive support in various libraries and systems.<br /> <br /> - Public key cryptography: RSA uses a public and private key pair, allowing for easier key management and distribution. The public key can be shared openly, while the private key remains secret. | - Performance: RSA is generally slower than HMAC, as it involves more complex mathematical operations and requires larger key sizes to achieve the same level of security. <br /> <br /> - Key size: RSA requires larger keys for equivalent security levels compared to symmetric algorithms like HMAC. For example, a 3072-bit RSA key provides roughly the same level of security as a 256-bit symmetric key.| 

Another algorithm, ECDSA (Elliptic Curve Digital Signature Algorithm), is pretty similar to RSA, but a little bit faster (the performance ranking from most efficiency to less efficiency should be: HMAC > ECDSA > RSA). However, back to our case, I agree that HMAC is good for our scenario, due to its performance and simplicity. More importantly, it meets our requirement of supporting both data integrity and authenticity, so that the payload of our token cannot be tampered.

cwperks:
HI @peternied , since this a component of a larger feature (Security for Extensions), should we merge this into a feature branch until the feature is ready to merge to main?

peternied:
Let's merge changes into the main branch when they benefit security plugin customers. I recommend moving this pull request to a `feature/jwt` branch, and once it's functional enough for customer use and end-to-end validation, merge it into the main branch.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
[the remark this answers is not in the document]
cwperks: @peternied @RyanL1997 I'm curious to get your thoughts on an idea of encrypting the value of the claim itself for sensitive claims. For cases with an external IdP, I think we will need to store the roles in a claim of the token because they cannot be looked up when the security plugin receives the token as part of a request. Since they cannot be looked up, they will need to be part of the claims of the token. The current JWT backend (and OICD and SAML since they also use JWTs) already assumes that roles are included as a claim of the token. 

Encryption of the sensitive claim can be done using a utility like (Reference SO post):

```
import java.util.Arrays;
import java.util.Base64;

import javax.crypto.Cipher;
import javax.crypto.SecretKey;
import javax.crypto.spec.SecretKeySpec;

public class EncryptionDecryptionUtil {

    public static String encrypt(final String secret, final String data) {


        byte[] decodedKey = Base64.getDecoder().decode(secret);

        try {
            Cipher cipher = Cipher.getInstance("AES");
            // rebuild key using SecretKeySpec
            SecretKey originalKey = new SecretKeySpec(Arrays.copyOf(decodedKey, 16), "AES");
            cipher.init(Cipher.ENCRYPT_MODE, originalKey);
            byte[] cipherText = cipher.doFinal(data.getBytes("UTF-8"));
            return Base64.getEncoder().encodeToString(cipherText);
        } catch (Exception e) {
            throw new RuntimeException(
                    "Error occured while encrypting data", e);
        }

    }

    public static String decrypt(final String secret,
            final String encryptedString) {


        byte[] decodedKey = Base64.getDecoder().decode(secret);

        try {
            Cipher cipher = Cipher.getInstance("AES");
            // rebuild key using SecretKeySpec
            SecretKey originalKey = new SecretKeySpec(Arrays.copyOf(decodedKey, 16), "AES");
            cipher.init(Cipher.DECRYPT_MODE, originalKey);
            byte[] cipherText = cipher.doFinal(Base64.getDecoder().decode(encryptedString));
            return new String(cipherText);
        } catch (Exception e) {
            throw new RuntimeException(
                    "Error occured while decrypting data", e);
        }
    }
}
```

The `secret` that this function takes is different than the `signing_key` that this PR introduces and would be another setting on the same level. I wrote a test with this behavior here: https://github.com/RyanL1997/security/compare/jwt-generator-for-extensions...cwperks:security:jwt-generator-for-extensions

Here's the output of the test. The first line is the encrypted claim that the extension would see and below is decrypted that the security plugin would be able to decrypt from the claim inside the token. 

```
org.opensearch.security.authtoken.jwt.JwtVendorTest > testCreateJwtWithRoles STANDARD_OUT
    rolesClaim: U5CjroB/LS95E5nrKl+WMw==
    roles: IT,HR
```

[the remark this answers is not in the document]
RyanL1997: Hi @cwperks, thanks for putting this together. This is making sense to me.

scrawfor99 (approved):
Thanks for putting this together Ryan. It looks like all my previous comments were addressed. I am happy to merge this PR and then follow up as needed for velocity.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -0,0 +1,151 @@
...
+
+            return jwk;
+        }
+    }
+
+    //TODO:Getting roles from User
+    public Map<String, String> prepareClaimsForUser(User user, ThreadPool threadPool) {
+        Map<String, String> claims = new HashMap<>();
+        this.threadContext = threadPool.getThreadContext();
+        final TransportAddress caller = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
```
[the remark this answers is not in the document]
scrawfor99: @RyanL1997, did this get split into the separate issue? It is the last area where I have a concern around merging. As long as there is an issue, I am fine with following-up in this case but could you share the link please? Thank you.

src/test/java/org/opensearch/security/authtoken/jwt/JwtVendorTest.java
scrawfor99, line 80: We may want to add further negative test cases down the road. Since you are only looking at the vending this seems fine for now, but once we have verification as well, we will want to test expiration compliance and signature mismatches.

scrawfor99 (approved):

src/test/java/org/opensearch/security/authtoken/jwt/JwtVendorTest.java
RyanL1997: 100% We need an authc backend to do the verification, so that we can test them. I''m working on that.

RyanL1997:
Hi @scrawfor99, here is a link for place holder. Me and Craig are working on a documentation, and I will transfer that into this issue.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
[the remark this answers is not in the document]
cwperks: For the experimental release we are aiming for sensible defaults. In this case HMAC SHA512 is chosen as the sensible defaults. Customizability will be added, but less of a priority then making sure all key functionality works first.

Issue #2581, opened 2023-03-22
Broken link in the DEVELOPER_GUIDE.md


nurSaadat:
**What is the bug?**
Broken link on line 49 which leads to a 404 page.

**How can one reproduce the bug?**
1. Go to [DEVELOPER_GUIDE.md](https://github.com/opensearch-project/security/blob/main/DEVELOPER_GUIDE.md)
2. In the last paragraph of "Native platforms" section, click on the "here" link.
3. See a page with 404 error and a message `The 'opensearch-project/security' repository doesn't contain the 'plugin-descriptor.properties' path in 'main'.`

**What is the expected behavior?**
The link redirects to a valid page.

**Do you have any screenshots?**
![image](https://user-images.githubusercontent.com/39532643/227051072-5496c2ad-378b-4c94-a6c4-301973976a30.png)
![image](https://user-images.githubusercontent.com/39532643/227051471-a63b790b-1391-45dc-8f01-db5263b643d5.png)

peternied:
@nurSaadat Thanks for filing, I'd be happy to review a PR that resolve this

PR #2582, opened 2023-03-23
Remove broken link from DEVELOPER_GUIDE.md

nurSaadat:
### Description
* Category
Documentation
These changes keep the documentation up-to-date according to the changes which were introduced in earlier contributions. 
Before the changes, when clicking on the "here" link in the "Native platforms" section of [DEVELOPER_GUIDE.md](https://github.com/opensearch-project/security/blob/main/DEVELOPER_GUIDE.md) the user was redirected to a page with 404 error and a message `The 'opensearch-project/security' repository doesn't contain the 'plugin-descriptor.properties' path in 'main'.`

After the changes, the "Worth noting" paragraph was put after building code snippet, so that the user will be able to access the `plugin-descriptor.properties` file, which is created programmatically after [#1623](https://github.com/opensearch-project/security/pull/1623)

### Issues Resolved
*  Resolve: #2581

### Testing
Tested manually by reviewing the edited .md document in the forked repo.
image

### Check List

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -45,9 +45,6 @@ cd $OPENSEARCH_HOME
 
 The `curl localhost:9200` call should succeed again. Kill the server with `Ctrl+c`. We are now ready to install the security plugin.
 
->Worth noting:\
-> The version of OpenSearch and the security plugin must match as there is an explicit version check at startup. This can be a bit confusing as, for example, at the time of writing this guide, the `main` branch of this security plugin builds version `1.3.0.0-SNAPSHOT` compatible with OpenSearch `1.3.0-SNAPSHOT` that gets built from branch `1.x`. Check the expected compatible version [here](https://github.com/opensearch-project/security/blob/main/plugin-descriptor.properties#L27) and make sure you get the correct branch from OpenSearch when building that project.
-
 ## Building
 
 First create a fork of this repo and clone it locally. You should then change to the directory containing the clone and run this to build the project:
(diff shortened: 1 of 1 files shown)


cwperks (dismissed):

scrawfor99 (dismissed):
Great find! Thank you for your contribution :)

DEVELOPER_GUIDE.md
```diff
@@ -56,6 +53,9 @@ First create a fork of this repo and clone it locally. You should then change to
 ./gradlew clean assemble
 ```
 
+>Worth noting:\
```
peternied, line 56: Please update this section to provide more detail on how to find this file as @cwperks mentioned in his https://github.com/opensearch-project/security/pull/2582#issuecomment-1481144513

peternied: Note, I would prefer an option that was faster than a full build, such as referencing a part of the build.gradle or using a fast command line option

https://github.com/opensearch-project/security/blob/575c2bc1e7283398ef448c404f480bd7b1eddef5/build.gradle#L18

https://github.com/opensearch-project/security/blob/575c2bc1e7283398ef448c404f480bd7b1eddef5/.github/workflows/ci.yml#L133

peternied (changes requested):
Thanks for fixing the broken link and updating the version numbers, but I think this section needs more detail on how to determine the compatible version of OpenSearch.

DEVELOPER_GUIDE.md
nurSaadat: Thank you @peternied for outlining the easier and faster way to find compatible OpenSearch version!

Probably then it is better to remove the method which includes looking for `plugin-descriptor.properties` file in a build. It is harder and takes more time, as well as it is quite lengthy to describe.

peternied (approved):
Thanks, much better!

scrawfor99 (approved):
Thank you for all your help with this!

Issue #2584, opened 2023-03-24
Add Extension Service Accounts


scrawfor99:
## Problem Statement

Based on the discussions on decision document review and comment discussions, I am moving forward with implementing service accounts for Extensions. 

As part of implementing service accounts, there are several individual items that will need to be addressed. 

Specifically the Security Plugin: 

- Needs to know what extensions exist
- Needs to know how to make a service account
- Needs to be able to give the service account to an extension  

After addressing these items, the a mock up of a service account can be provided. The mock up will include details on how the Security Plugin can authenticate a service account and how a service account is treated differently than a user account. 

$\textcolor{cyan}{\textsf{Needs to know what extensions exist}}$

For the Security Plugin to know what extensions exist, we can just implement a new API similar to the existing `_cat/plugins` API. this will allow the Security Plugin to know which extensions are installed. 

The other alternative would be to parse the service accounts to find the extensions associated with each but this would be a costly process taking O(n) to resolve a single extension. 

You could alternatively inject the Extensions Manager into the Guice Holder for the Security Plugin. 

Until proven otherwise, it appears to make much more sense to choose the first option.

$\textcolor{cyan}{\textsf{Needs to know how to make a service account}}$

When an extension is installed, we will want core to share information about the extension with the Security Plugin. The Security Plugin can then use that information to create a service account for the extension along the lines of the comments on this [issue](https://github.com/opensearch-project/security/issues/2572) and this [write up](https://github.com/opensearch-project/security/issues/2566). 

$\textcolor{cyan}{\textsf{Needs to be able to give the service account to the extension}}$

After a service account is created for an extension, the Security Plugin will need to give that account to the extension. The most straightforward way of doing this is by sending the service account credentials back to core and then having core provide those details to the extension as part of the installation process. There is not a clear alternative to this solution since the extension will not actually be speaking directly with the Security Plugin at any point of installation.

cwperks:
@scrawfor99 It should be possible to inject the `ExtensionsManager` to get the list of installed extensions which updates as new extensions are installed as well. Anything binded in `Node.java` is injectable: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/node/Node.java#L1020

You can inject it into the `GuiceHolder` here: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L1172-L1179

-----

[...]

PR #2594, opened 2023-03-28
Security User Refactor

scrawfor99:
### Description
This PR is meant to refactor the concept of a user from the Security Plugin's point of view. This will allow us to introduce the basics of service accounts on the Security plugin side. 

### Issues Resolved
This PR will contribute to resolving #2584. 

### Testing
The existing InternalUsersTests are being used to verify the previous functionality is maintained. After merging this change, specific Service Account operations can be performed and tests added to exercise service account logic. 

### Check List


cwperks:
Hi @scrawfor99 , I think you are on the right track with the extensions registration process and creation of the service account at that time for extensions that require one, but I think we should start by introducing the notion of a service account.

I see in this PR that `service: true` is added in the attribute map of the internal user and I think this should be a field of a user in the style of `reserved: true` if the service accounts will be stored in the internal user list. 

For introduction of the concept of a service account, I would expect:

1. That service accounts are designated by a special property `service` on a user. If this property exists and is set to `true` than it is a service account.
2. Service accounts can only be mapped to at most one role - that role will dictate how a service account token can interact with the cluster. This can be used to implement system indices - i.e. service account tokens scoped to a system index pattern can create, delete and modify the index.
3. Service accounts cannot be mapped to backend roles
4. Service accounts are passwordless and only become useful once a token is generated for the account
5. internaluser APIs should account for service accounts
    - List internalusers should not include service accounts
    - Create internaluser prohibits creation of account
    - Delete internaluser prohibits deletion of account - once they are created programmatically on registration of extension they will exist and be useless until a token is created
    - Patch user/users prohibits patching the account
6. There should be a convention in naming the role that can be associated with a service account. If by convention we have a service account named `<extension/extensionId>` then maybe a role can exist with the same name for quick retrieval?

Once the concept of the service account is introduced, an API to list service accounts should be created and we can also start linking it to the extensions use-case like you are doing here to create one on extension registration.

scrawfor99:
Hi @cwperks, thank you for your thoughtful comments. For specifications about defining service accounts, I would suggest we use the associated decision doc/question #2597. I think that some of the expectations you listed for service accounts sound like great ideas, but I also think we may need to come to a better understanding about some others. For instance, having an API to list service accounts sounds great. However, I am not sure why we need to restrict the number of roles a service account can have. 

How about we head over to the decision doc to hash things out and then we can come back here as I implement based on our decision?

cwperks:
@scrawfor99 I will move the comments over to the decision doc. I was thinking that service account tokens would be narrowly used in order to allow extensions to write to reserved indices for those extensions that have special protections - similar to plugins with a system index, but without the need for stashing the thread context.

It would be possible to increase the scope of the service account token for use-cases where an extension runs as a daemon if its been granted by the cluster admin. Single roles makes it a one stop place to look up the permissions. User's may have multiple roles, but for a service account limiting it to a single role would make it explicitly clear what it can do. I have seen other systems that implement service accounts hardcode the roles definition in code and the permissions are not modifiable at all.

scrawfor99:
This will resolve #2609 since we can now make use of the internalUsersApiAction to access the Service Accounts. There may be slight modifications required in a follow-up but the general functionality is there.

scrawfor99:
I created an issue for the failing test #2648. It passes locally and on all other configurations. I have tried re-running things but it did not fix the issue. I hope this can still not be blocked.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -479,7 +482,7 @@ public List<RestHandler> getRestHandlers(Settings settings, RestController restC
                                 cr, cs, principalExtractor,
                                 evaluator,
                                 threadPool,
-                                Objects.requireNonNull(auditLog), sks,
+                                Objects.requireNonNull(auditLog), sks, userService,
```
peternied, line 485:
```suggestion
                                Objects.requireNonNull(auditLog), sks,
                                Objects.requireNonNull(userService),
```

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1166,16 +1172,17 @@ public static class GuiceHolder implements LifecycleComponent {
         private static RepositoriesService repositoriesService;
         private static RemoteClusterService remoteClusterService;
         private static IndicesService indicesService;
-
         private static PitService pitService;
+        private static ExtensionsManager extensionsManager;
```
peternied, line 1176: I don't think this is used, lets remove

src/main/java/org/opensearch/security/user/AccountCreateOrUpdateException.java
```diff
@@ -0,0 +1,51 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.user;
+
+public class AccountCreateOrUpdateException extends Exception {
```
peternied, line 29: Nit: extend RuntimeException so callers don't have to handle this, unless you have thoughts on this? https://stackoverflow.com/a/14012233/533057

src/main/java/org/opensearch/security/user/ServiceAccountRegistrationException.java
```diff
@@ -0,0 +1,51 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.user;
+
+public class ServiceAccountRegistrationException extends Exception {
```
peternied, line 29: Nit: Could we get away with a single exception type?

src/main/java/org/opensearch/security/user/ServiceAccountRegistrationResponse.java
```diff
@@ -0,0 +1,75 @@
...
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.user;
+
+
+public class ServiceAccountRegistrationResponse {
```
peternied, line 30: This class doesn't seem to be used?

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,311 @@
...
+ */
+public class UserService {
+
+    protected final Logger log = LogManager.getLogger(this.getClass());
+    ClusterService clusterService;
+    static ConfigurationRepository configurationRepository;
+    String securityIndex;
+    Client client;
+
+    final String NO_PASSWORD_OR_HASH_MESSAGE = "Please specify either 'hash' or 'password' when creating a new internal user.";
```
peternied, line 77: Make these static

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,311 @@
...
+    // This should be for all internal users instead of just service accounts
+
+
+    // Make a SecurityPluginUser object
+    // Work backwards from remaking this as a User Service -- can mostly reuse from current state
+    // Want to check that when code is created or updated, you check if it is a service account and turn on properties
+    // For example: Creator/owner of service account (extension or user)
+    // All end-to-end tests can be written without extensions and we can test everything
+    // Should not need to test create, update, list, get as unit tests but will want to test service account creation and deletion
+    // I.e. service account is invalid because no owner -- can leave tests till later.
```
peternied, line 106: Not sure these comments are relevant, lets clean these up or promote to javadocs

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,311 @@
...
+    protected static final SecurityDynamicConfiguration<?> load(final CType config, boolean logComplianceEvent) {
+        SecurityDynamicConfiguration<?> loaded = configurationRepository.getConfigurationsFromIndex(Collections.singleton(config), logComplianceEvent).get(config).deepClone();
+        return DynamicConfigFactory.addStatics(loaded);
+    }
+
+    protected void saveAndUpdateConfiguration(final Client client, final CType cType,
+                                              final SecurityDynamicConfiguration<?> configuration) {
+        final IndexRequest ir = new IndexRequest(this.securityIndex);
+
+        // final String type = "_doc";
```
peternied, line 137: Remove commented code

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,311 @@
...
+
+        String accountName = securityJsonNode.get("name").asString();
+
+        if (accountName == null || accountName.length() == 0) { // Fail if field is present but empty
+            throw new AccountCreateOrUpdateException(NO_ACCOUNT_NAME_MESSAGE);
+        }
+
+        final Map<String, String> accountAttributes = new HashMap<>();
+
+        if (!securityJsonNode.get("service").isNull() && securityJsonNode.get("service").asString() == "true") { // If this is a service account
```
peternied, line 179: Its difficult to see the schema changes that are being added for service accounts as is.  It might be cleaner to separate the service account updates into follow up PR - what do you think?

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,311 @@
...
+            // final DiscoveryExtensionNode extensionInformation = OpenSearchSecurityPlugin.GuiceHolder.getExtensionsManager().getExtensionIdMap().get(extensionUniqueId);
+            // extensionInformation.getSecurityConfiguration(); TODO: Need to make it so that we can get the extension configuration information
+            // extensionInformation.parseToJson();
+            // Add default role option for extensions which do not specify their own role
+            accountAttributes.put("service", "true"); // This attribute signifies that the account is a service account
```
peternied, line 184: This is schema so we need to be careful about these values.  I don't think 'service account' is very meaningful as a field on its own

I see the defining characteristic of a service account is an account with a relationship to an another system, e.g. this account is for a plugin or an extension.  So maybe that relationship is `"usedBy": "..."`.  

[If we defer the service account parts we can migrate this discussion to that PR or a new issue]

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,311 @@
...
+        }
+
+        internalUsersConfiguration.remove(accountName);
+        contentAsNode.remove("name");
+
+        internalUsersConfiguration.putCObject(accountName, DefaultObjectMapper.readTree(contentAsNode,  internalUsersConfiguration.getImplementingClass()));
+
+        if (!securityJsonNode.get("service").isNull() && securityJsonNode.get("service").asString() == "true") { // Internal users update the config as part
+            saveAndUpdateConfiguration(client, CType.INTERNALUSERS, internalUsersConfiguration);
+        }
```
peternied, line 243: The config always needs to be updated or the change won't replicate to the other nodes

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,311 @@
...
+        return internalUsersConfiguration;
+    }
+
+    public static List<String> listServiceAccounts() {
```
peternied, line 247: Unused method

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,311 @@
...
+        return internalUserAccounts;
+    }
+
+    public static Set<String> listUserAccounts() {
+
+        final SecurityDynamicConfiguration<?> internalUsersConfiguration = load(getConfigName(), false);
+        return internalUsersConfiguration.getCEntries().keySet();
+    }
+
+    public static boolean accountExists(String accountName) {
```
peternied, line 287: this should not be static

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,311 @@
...
+
+    /**
+     * This function will handle the creation or update of a user account.
+     *
+     * @param accountDetailsAsString A string JSON of different account configurations.
+     * @return
+     * @throws ServiceAccountRegistrationException
+     * @throws IOException
+     */
+    public SecurityDynamicConfiguration<?> createOrUpdateAccount(String accountDetailsAsString) throws IOException, AccountCreateOrUpdateException, ServiceAccountRegistrationException {
```
peternied, line 161: I think this needs a refactor to be more maintainable, different validations can be moved into functions to make this more straight forward.

Afterward - please add unit test to verify this functionality

peternied (changes requested):
Can we minimize the amount of 'service account' differences that are needed?  We can add more difference later if we need specific functionality.  For v1 I think the only think we need is a way to know that an account is associated with an extension - how does this sound to you?

src/main/java/org/opensearch/security/user/ServiceAccountRegistrationResponse.java
scrawfor99: Yeah, this is for once the core side is hooked up. It will be what is sent back to core. I can remove it for now if you want? What do you think?

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: This needs to be set like this for use with the internalUserApiAction.

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: The API for them is not written yet. They were requested as part of the Service Account specifications.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -132,46 +120,23 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
...
-            // sanity check, this should usually not happen
-            final String hash = ((Hashed) internalUsersConfiguration.getCEntry(username)).getHash();
-            if (hash == null || hash.length() == 0) {
-                internalErrorResponse(channel,
-                    "Existing user " + username + " has no password, and no new password or hash was specified.");
+        try {
+            ((ObjectNode) content).put("name", username);
+            internalUsersConfiguration = userService.createOrUpdateAccount(content.toString());
+        } catch (Exception ex) {
+            if (ex.toString().contains(userService.getSERVICE_ACCOUNT_PASSWORD_MESSAGE()) || ex.toString().contains(userService.getNO_PASSWORD_OR_HASH_MESSAGE()) || ex.toString().contains(userService.getSERVICE_ACCOUNT_PASSWORD_MESSAGE()) || ex.toString().contains(userService.getNO_ACCOUNT_NAME_MESSAGE()) || ex.toString().contains(userService.getSERVICE_ACCOUNT_HASH_MESSAGE()) || ex.toString().contains(userService.getRESTRICTED_CHARACTER_USE_MESSAGE())) {
```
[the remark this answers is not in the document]
scrawfor99: Will do.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
scrawfor99: Will do

src/main/java/org/opensearch/security/user/ServiceAccountRegistrationResponse.java
peternied: I would prefer we removed all unused code / comments.  Is there a meta issue with this step called out that we could reference in future PRs?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
scrawfor99: I removed this for now, it will be needed for grabbing the extension information later though.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -405,4 +406,31 @@ private static String formatDate(long date) {
     public static int getDefaultConfigVersion() {
         return ConfigurationRepository.DEFAULT_CONFIG_VERSION;
     }
+
+    public Map<CType, SecurityDynamicConfiguration<?>> getConfigurationsFromIndex(Collection<CType> configTypes) {
```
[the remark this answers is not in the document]
scrawfor99: Will do

src/main/java/org/opensearch/security/user/ServiceAccountRegistrationResponse.java
scrawfor99: Alright I will remove it; it is documented as part of the api handling so should be all set

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: Done

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,311 @@
...
+
+    final String NO_PASSWORD_OR_HASH_MESSAGE = "Please specify either 'hash' or 'password' when creating a new internal user.";
+    final String RESTRICTED_CHARACTER_USE_MESSAGE = "A restricted character(s) was detected in the account name. Please remove: ";
+
+    final String SERVICE_ACCOUNT_PASSWORD_MESSAGE = "A password cannot be provided for a service account. Failed to register service account: ";
+
+    final String SERVICE_ACCOUNT_HASH_MESSAGE = "A password hash cannot be provided for service account. Failed to register service account: ";
+
+    final String NO_ACCOUNT_NAME_MESSAGE = "No account name was specified in the request.";
+    protected static CType getConfigName() {
```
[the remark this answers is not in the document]
scrawfor99: Will do

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: Oops

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: Oops my bad

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: I would prefer to leave this here since I am splitting it into methods. The more PRs the harder it is to keep velocity up. I will refactor to make it clearer.

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: This was discussed over in https://github.com/opensearch-project/security/issues/2597#issuecomment-1490826252. So for now, I would prefer to stick with the `service` term. We can definitely swap around the specifics at a later point. I am more than open to suggestions for naming things. I think that since the code is isolated it would only require a minimal change if we decided to use an alternative naming convention. Does leaving this for now sound alright with you?

scrawfor99:
List internal users & service accounts code

```
public static List<String> listServiceAccounts() {

        final SecurityDynamicConfiguration<?> internalUsersConfiguration = load(getConfigName(), false);

        List<String> serviceAccounts = new ArrayList<>();
        for (Map.Entry<String, ?> entry : internalUsersConfiguration.getCEntries().entrySet()) {

            final InternalUserV7 internalUserEntry = (InternalUserV7) entry.getValue();
            final Map accountAttributes = internalUserEntry.getAttributes();
            final String accountName = entry.getKey();
            if (accountAttributes.containsKey("service") && accountAttributes.get("service") == "true") {
                serviceAccounts.add(accountName);
            }
        }
        return serviceAccounts;
    }

    public static List<String> listInternalUsers() {

        final SecurityDynamicConfiguration<?> internalUsersConfiguration = load(getConfigName(), false);

        List<String> internalUserAccounts = new ArrayList<>();
        for (Map.Entry<String, ?> entry : internalUsersConfiguration.getCEntries().entrySet()) {

            final InternalUserV7 internalUserEntry = (InternalUserV7) entry.getValue();
            final Map accountAttributes = internalUserEntry.getAttributes();
            final String accountName = entry.getKey();
            if (!accountAttributes.containsKey("service") || accountAttributes.get("service") == "false") {
                internalUserAccounts.add(accountName);
            }
        }
        return internalUserAccounts;
    }

    public static Set<String> listUserAccounts() {

        final SecurityDynamicConfiguration<?> internalUsersConfiguration = load(getConfigName(), false);
        return internalUsersConfiguration.getCEntries().keySet();
    }

```

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,228 @@
...
+
+        String accountName = securityJsonNode.get("name").asString();
+
+        if (accountName == null || accountName.length() == 0) { // Fail if field is present but empty
+            throw new UserServiceException(NO_ACCOUNT_NAME_MESSAGE);
+        }
+
+        final Map<String, String> accountAttributes = new HashMap<>();
+
+        if (!securityJsonNode.get("service").isNull() && securityJsonNode.get("service").asString() == "true") { // If this is a service account
```
RyanL1997, line 157: This may be a dumb question: will the service account's name also have naming restriction check?

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,228 @@
...
+
+            accountAttributes.put("service", "true"); // This attribute signifies that the account is a service account
```
RyanL1997, line 159: I remember we had a conversation about the activation/deactivation about service account before. I was wondering can this flag be changed dynamically for those activation and deactivation?

RyanL1997 (commented):
Hi @scrawfor99, thanks for putting this together!

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,228 @@
+/*
+ * Copyright 2015-2018 _floragunn_ GmbH
```
peternied, line 2: This doesn't seem correct

scrawfor99: It certainly is not haha. Copied from the wrong file it appears. Good catch.

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: It will be able to be changed with an API once we implement that. Right now, it will just be part of the configuration file.

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: No dumb questions! I don't think so. The service account name is just going to be the extension unique ID for now, so I don't plan to have any restrictions. The reason we have the ":" restriction on the normal usernames is for basic auth but service accounts will use Bearer so I don't think we will need to worry.

scrawfor99:
Plugin install failing as mentioned here: #2657

scrawfor99:
@opensearch-project/security if we could get reviews on this as soon as possible that would be great so that the changes can be merged and we can continue to move forward. Thank you.

cwperks (commented):
@scrawfor99 Took another pass and left some comments around the design of the new fields in the attribute map of a user. One comment I'd like to see addressed about stringifying the JsonNode only to read it back in.

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,186 @@
...
+        String accountName = securityJsonNode.get("name").asString();
+
+        if (accountName == null || accountName.length() == 0) { // Fail if field is present but empty
+            throw new UserServiceException(NO_ACCOUNT_NAME_MESSAGE);
+        }
+
+        if (!securityJsonNode.get("attributes").get("owner").isNull() && !securityJsonNode.get("attributes").get("owner").equals(accountName)) { // If this is a service account
+            verifyServiceAccount(securityJsonNode, accountName);
+            String password = generatePassword();
+            contentAsNode.put("password", password);
```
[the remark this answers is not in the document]
scrawfor99: It is not stored, this is pattern for the internalUsers. See below in the code, it is then removed from the config. I don't strictly know that it needs to be added but when it is added we can reuse the existing internal user code below this directly since it expects a password to be provided.

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,186 @@
...
+        SecurityJsonNode securityJsonNode = new SecurityJsonNode(contentAsNode);
+
+        final SecurityDynamicConfiguration<?> internalUsersConfiguration = load(getConfigName(), false);
+        String accountName = securityJsonNode.get("name").asString();
+
+        if (accountName == null || accountName.length() == 0) { // Fail if field is present but empty
+            throw new UserServiceException(NO_ACCOUNT_NAME_MESSAGE);
+        }
+
+        if (!securityJsonNode.get("attributes").get("owner").isNull() && !securityJsonNode.get("attributes").get("owner").equals(accountName)) { // If this is a service account
```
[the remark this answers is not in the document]
scrawfor99: Hi Craig, thank you for your good questions. Basically, the accountName is not an attribute. The accountName is the key to the user in the configuration. The owner is whoever 'owns' the account. For example, if you installed the HellowWolrld extension, it would make a service account for HelloWorld where you would be the owner even though the account name is HelloWorld. 

To your second point, the internalUsers are not the same as the User objects. The User objects that Ryan built the token generator around have a specific set of requirements for them that are more restrictive then the internalUsers. InternalUsers are mostly a data structure in practice and implementation. 

As a result of this, there are some differences with how the existing APIs were constructed. For instance, you can only specify four fields in the requests involving the internalUsers API one of which is the `attributes` field. If you were to try to specify individual fields the existing validator would need to be changed to accept them. The current implementation does allow an administrator to convert an account from a service account to normal account or vice versa. I don't know that this is necessarily a problem however.

src/test/java/org/opensearch/security/TransportUserInjectorIntegTest.java
```diff
@@ -106,11 +106,12 @@ public void testSecurityUserInjection() throws Exception {
             Assert.fail("Expecting exception");
         } catch (OpenSearchSecurityException ex) {
             exception = ex;
-            log.warn(ex.toString());
+            log.debug(ex.toString());
```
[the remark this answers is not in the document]
scrawfor99: Warn will always print out. If the test is expected to fail, I was not sure why we would print out the exception message. I swapped to debug since that would be used when we were trying to diagnose unexpected behavior.

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * This function will handle the creation or update of a user account.
+     *
+     * @param accountDetailsAsString A string JSON of different account configurations.
+     * @return InternalUserConfiguration with the new/updated user
+     * @throws UserServiceException
+     * @throws IOException
+     */
+    public SecurityDynamicConfiguration<?> createOrUpdateAccount(String accountDetailsAsString) throws IOException {
```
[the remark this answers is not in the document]
scrawfor99: Yeah good point. I will fix this.

src/main/java/org/opensearch/security/user/UserService.java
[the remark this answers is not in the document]
scrawfor99: Looking this over, I removed it since it was not needed.

scrawfor99:
@RyanL1997, @DarshitChanpura, @reta can I get a review here so that we can potentially merge this? I addressed Craig's comments but he is not around this week.

reta:
> @RyanL1997, @DarshitChanpura, @reta can I get a review here so that we can potentially merge this? I addressed Craig's comments but he is not around this week.

@scrawfor99 my apologies, I do not (yet) know this part of the plugin well enough, looking nonetheless but would let others to approve with confidence

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -478,7 +549,6 @@ public void testPasswordRules() throws Exception {
         HttpResponse response = rh
                 .executeGetRequest("_plugins/_security/api/" + CType.INTERNALUSERS.toLCString());
         Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
-        System.out.println(response.getBody());
```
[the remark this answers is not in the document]
scrawfor99: Hi Ryan, I was removing that since I believe we generally avoid System.out.println when possible. I don't know if that line snuck through on a previous PR but I don't think it is a pattern we endorse so figured I would remove it.

RyanL1997 (dismissed):
Nice work! Thanks @scrawfor99

RyanL1997:
Btw, I saw Peter was requesting some changes, but for some reason I couldn't see that thru the log. Just wanna double check with you to make sure all the comments have been resolved.

scrawfor99:
Hi @RyanL1997, thank you. Yeah, Peter's changes were addressed a while ago, I will dismiss if we get two approvals.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -128,50 +115,35 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
...
         // changes
 
-        // sanity checks, hash is mandatory for newly created users
-        if (!userExisted && securityJsonNode.get("hash").asString() == null) {
-            badRequestResponse(channel, "Please specify either 'hash' or 'password' when creating a new internal user.");
+        try {
+            if (request.hasParam("owner")) {
+                ((ObjectNode) content).put("owner", request.param("owner"));
+            }
+            if (request.hasParam("isEnabled")) {
```
DarshitChanpura, line 133: What do you think about changing this flag to `isServiceAccount` to allow for checks whether this account is a service account or not? And, admins can toggle this feature to switch a normal account to service-account and vice-versa

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,181 @@
...
+        final SecurityDynamicConfiguration<?> internalUsersConfiguration = load(getConfigName(), false);
+        String accountName = securityJsonNode.get("name").asString();
+
+        if (accountName == null || accountName.length() == 0) { // Fail if field is present but empty
+            throw new UserServiceException(NO_ACCOUNT_NAME_MESSAGE);
+        }
+
+        if (!securityJsonNode.get("attributes").get("owner").isNull() && !securityJsonNode.get("attributes").get("owner").equals(accountName)) { // If this is a service account
+            verifyServiceAccount(securityJsonNode, accountName);
+            String password = generatePassword();
```
DarshitChanpura, line 111: why not use `securityJsonNode.put('name')...` here to avoid duplicate declaration on line 115?
Having said that, I know that `put()` doesn't exist in SecurityJsonNode for some unknown reason (probably to make it a read-only node). Do you think creating it might be worthwhile?

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * This will be swapped in for a real solution once one is decided on.
+     *
+     * @return A password for a service account.
+     */
+    private String generatePassword() {
+        String generatedPassword = "superSecurePassword";
+        return generatedPassword;
+    }
+}
```
DarshitChanpura, line 181: 👏🏼 Great job moving in all validations and user related service methods into this new class. Better for readability!

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -292,6 +321,49 @@ private void verifyPatch(final boolean sendAdminCert, Header... restAdminHeader)
...
                 HttpStatus.SC_CREATED);
 
+        // Add enabled service account then get it
+
+
+        rh.sendAdminCertificate = sendAdminCert;
+        response = rh.executePutRequest(ENDPOINT + "/internalusers/happyServiceLive",
+                ENABLED_SERVICE_ACCOUNT_BODY, restAdminHeader);
+        Assert.assertEquals(response.getBody(), HttpStatus.SC_CREATED, response.getStatusCode());
+        rh.sendAdminCertificate = sendAdminCert;
```
DarshitChanpura, line 331: Probly can remove this re-assignment and the ones below as they are redundant because we are not changing reference for `rh`

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
scrawfor99: The idea is to abstract as much as possible away from being specifically about "Service Accounts". By using the `owner` attribute, we are able to treat service accounts identical to any other account and avoid locking us into a specific convention. In theory, an administrator could create a service account with them being the owner in this scenario and that would be unrelated to extensions.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -128,50 +115,35 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
...
+                ((ObjectNode) content).put("isEnabled", request.param("isEnabled"));
```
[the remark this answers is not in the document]
scrawfor99: Unfortunately, the SecurityJsonNode does not have any way to put or remove fields from it. The ObjectNode allows this to be done directly. I can make the swap but it will require modifying the behavior of the SecuirtyJsonNode to include `put` and `remove`. 

Which would you prefer?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -128,50 +115,35 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
...
         final boolean userExisted = internalUsersConfiguration.exists(username);
 
         // when updating an existing user password hash can be blank, which means no
         // changes
 
-        // sanity checks, hash is mandatory for newly created users
-        if (!userExisted && securityJsonNode.get("hash").asString() == null) {
-            badRequestResponse(channel, "Please specify either 'hash' or 'password' when creating a new internal user.");
+        try {
+            if (request.hasParam("owner")) {
```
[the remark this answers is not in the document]
scrawfor99: The `owner` field is used for determining whether an account is a service account (an account owned by an entity than the account's subject). This can be documented once we add more details around service accounts to the codebase but does not impact any existing functionality.

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: See above. I can remove use of the contentAsNode pattern by adding methods to the SecurityJsonNode object. I was not sure which implementation would be preferred. 

I am interpreting this as you would prefer that change so I will go that route.

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
scrawfor99: Will do

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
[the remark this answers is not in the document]
scrawfor99: Swapping per below

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: EDIT: I looked into this, looks like the existing SecurityJsonNode is based off of Jackson Databind's JSONNode class. That class is not constructed for removal or addition of elements. I am going to stick with this implementation for the time being unless that is a blocker to this PR.

scrawfor99: EDIT: Hi @DarshitChanpura Unless it is going to block this PR, I am going to suggest we leave it as is. I know it is not the most convenient solution but the securityJsonNode class has 40+ usages and was not intended to be directly modifiable. I have been working on modifying it but it requires changing the class from creating final instances of itself to creating mutable versions. I suspect a part of the reason the class was created how it is, is so that there would be an immutable version of a JSON node for implementations. Let me know if this will be a blocker and I can look more into it but I am going to suggest we leave this as is otherwise. 

P.S. I did make the other changes you suggested.

DarshitChanpura: Alright, we can leave it as is. Thank you for taking time and going through SecurityJsonNode 😃. We can cover this in a separate PR in future

DarshitChanpura (commented):
Thank you for addressing all the comments. I only have one last comment. The changes look good to me otherwise!

scrawfor99:
@DarshitChanpura, exactly, it is so that an account can be turned on/off and auth tokens will not be created for any disabled accounts.

DarshitChanpura (approved):
Thank you for addressing all my comments @scrawfor99 ! I'm happy to merge this.

src/main/java/org/opensearch/security/user/UserService.java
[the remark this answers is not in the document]
cwperks: I reviewed the APIs and UI to ensure there is no avenue for a normal user to modify their own attributes. In general, it would be great to define a schema for what a service account entails vs. relying on keys to be present in a generic map of attributes.

A few of the concerns I have with this design are:

- An admin can at any time delete one of these `owner` or `isEnabled` keys and what would that affect be?
- If the user updates the value of the `accountName` to something that doesn't exist then there is no referential integrity
- I'm still not quite sure what `owner` means, you mentioned its a placeholder to provide the name of the person of initially installed an extension? 
- Attributes are read in from external IdPs or provided as a claim in a jwt, would this cause any problems?

Similar to what @RyanL1997 has done with support for extension REST handlers, would it make sense to merge a change like this into a feature branch while more of the feature is being worked on?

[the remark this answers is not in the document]
scrawfor99: Hi @cwperks,

Thank you for taking the time to look into this PR. I know you mentioned having some concerns about the implementation. Reviewing your concerns it seems they relate to the immutability of the service accounts. In general, it does not seem like there needs to be immutability of service accounts for several reasons. Fields such as `isEnabled` should be configurable by nature of what they represent. Another point is discussed in this comment [above](https://github.com/opensearch-project/security/pull/2594#discussion_r1160030423). Basically, it is not entirely clear the hard distinction between a normal internal user account and a service account. They are intended to be interchangeable for all intents and purposes and swapping between the two should not necessarily be forbidden. The owner denotes the controller of the account. If it is a user, it is their account. If instead, it is a service account, the service cannot "own" an account as they are just a program. The owner is then the person who installed that program. I don't see any reason why reading from a JWT would be an issue. 

To your point about merging against the feature/extensions branch, I am against this practice in general. I think that all changes which can be safely merged into the main branch should be. By merging into the feature branch, we add debt we will later have to address and make it more and more complicated to get an accurate state of the code base. 

If you would like to open up a PR swapping things to how you prefer, I would be happy to review the changes and make any switches that everyone prefers.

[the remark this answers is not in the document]
cwperks: @scrawfor99 If there's code that relies on attribute keys being present in this map then there needs to be special protections in place to ensure that the keys cannot just be deleted from the map as that would put the user/account into a corrupt state. I see that @peternied also had the same concerns: https://github.com/opensearch-project/security/pull/2594#discussion_r1160030423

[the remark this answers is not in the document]
scrawfor99: @cwperks, Peter's concern appears to have been about the meaning-fulllness of the terms not the inclusion of the terms themselves. I disagree about needing explicit safeguards for either `isEnabled` or `ownner`. If you look at the map now, there are numerous values that could be deleted or changed and cause a "corrupted state." I see no reason to treat these attributes any differently then something like `roles` which could be deleted to the result of having no access.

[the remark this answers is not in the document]
cwperks: @scrawfor99 What is preventing this scenario from happening/is this a concern with the design? From the design put forward for allowing extensions to write to reserved indices for the extension the extension would utilize a token associated with the (service) account designated to the extension. These indices come with special protections that even forbid an admin from deleting the index and rendering the extension into a corrupt state. However, with this design what is preventing this:

1. Admin errantly updates the attributes of the account associated with the extension and deletes the `owner` key from the attribute map
2. Since the `owner` key is no longer present the account associated with the extension is no longer one of these special accounts so its now forbidden from writing to the indices with system index protection.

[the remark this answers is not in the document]
scrawfor99: Hi @cwperks, there is nothing preventing an admin from deleting the `owner` or turning off `isEnabled`. This would make the service account no longer vend its associated token and it would no longer have access to the associated index. This is by design. I view this as an acceptable trade off for having service accounts be functionally identical to other internal user accounts. An admin can swap between the two, enable or disable both without hard deletion, and use the existing internaluser apis to access service accounts. For a P0 implementation these are the exact types of trade offs I am in favor of. We can always switch things down the road but these code changes introduce no immediate risks--service accounts are not used for anything currently and the api changes are not documented on the website--and are important for the extensions work.

Issue #2559, opened 2023-03-16
[FEATURE] Create a way to add action oriented APIs to update configuration values


peternied:
**Is your feature request related to a problem?**
The security config of OpenSearch covers a lot of surface area.  Some of these settings are very sensitive, others operators might want to allow for more control over.  There is no clean way to provide granular access to these settings.

Lets take into consideration a recent configuration setting that was added `jwt_clock_skew_tolerance_seconds`.  The legwork required to expose GET / POST / DELETE (restore to default value) is large and difficult.

**What solution would you like?**
There should be a way to declare that there should be an configuration based action handler to support exposing this setting such as the following:
```java
new ConfigActionDescription().Builder()
   .valueType(ConfigActionType.Integer)
   .route("_plugins/_security/config/jwt/{auth_domain}/skew_tolerance_seconds")
   .permission("securityconfig:admin/config/jwt/skew_tolerance_seconds")
   .configPath("config.dynamic.authc.{auth_domain}.http_authenticator.config.jwt_clock_skew_tolerance_seconds")
   .onDeleteDefaultValue(DEFAULT_CLOCK_SKEW_TOLERANCE_SECONDS)
   .build();
```
* Note; type safety constraints might make this more slightly more complex

After registering the ConfigActionDescription object a handler would be built up that would support integer values following a straight forward convention of retrieve, update and set to default.  When attempting to access the endpoint, the permission would would be required to be prefixed with 'securityconfig:' and read from the users roles.

<details>

<summary>Example REST APIs</summary>

#### Retrieve
**Request**
```
GET "_plugins/_security/config/jwt/jwt_auth_domain/skew_tolerance_seconds"
```
**Response**
```
200 OK 
{
   "value": 30
}
```
#### Update
**Request**
```
PUT "_plugins/_security/config/jwt/jwt_auth_domain/skew_tolerance_seconds"
{
   "value":77
}
```
**Response**
```
200 OK
{
   "value": 77
}
```
#### Set to default
**Request**
```
DELETE "_plugins/_security/config/jwt/jwt_auth_domain/skew_tolerance_seconds"
```
**Response**
```
200 OK
{
   "value": 77
}
```
</details>

**Do you have any additional context?**
I think this would be a good way to work through problems around duplicate data and filtered access for dynamically modifying tenant config.  I think using a 'action' oriented framework would help create abstraction layers from being exposed to downstream users, e.g. `GET _plugins/_security/config/tenancy/multitenancy_enabled` 
[...]

PR #2607, opened 2023-03-30
Dynamic tenancy configurations

abhivka7:
### Description
* Category : Enhancement
These changes are required to dynamically enable/disable multi-tenancy, private tenant and set a default tenant.
Earlier users had to change YAML files in each node and restart node to enable/disable multi-tenancy and private tenant. With our changes, users will be able to change these settings dynamically. We will also give admins the option to set a default tenant for all users for first time log in.
More details can be found here:
https://github.com/opensearch-project/OpenSearch/issues/5853
https://github.com/opensearch-project/security-dashboards-plugin/issues/1302

Previous PR for the same issue: https://github.com/opensearch-project/security/pull/2444

### Issues Resolved
https://github.com/opensearch-project/security-dashboards-plugin/issues/1302

Nope

### Testing
We have done manual testing for all changes on our local server. We have also unit tests for our api changes. 

### Check List
- [ ✅ ] New functionality includes testing
- [] New functionality has been documented. : Will document in further commits.
- [ ✅ ] Commits are signed per the DCO using --signoff

### DIFF
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -115,6 +115,11 @@ import org.opensearch.search.internal.SearchContext;
 import org.opensearch.search.query.QuerySearchResult;
 import org.opensearch.security.action.configupdate.ConfigUpdateAction;
 import org.opensearch.security.action.configupdate.TransportConfigUpdateAction;
+import org.opensearch.security.action.tenancy.TenancyConfigRestHandler;
+import org.opensearch.security.action.tenancy.TenancyConfigRetrieveActions;
+import org.opensearch.security.action.tenancy.TenancyConfigRetrieveTransportAction;
+import org.opensearch.security.action.tenancy.TenancyConfigUpdateAction;
+import org.opensearch.security.action.tenancy.TenancyConfigUpdateTransportAction;
 import org.opensearch.security.action.whoami.TransportWhoAmIAction;
 import org.opensearch.security.action.whoami.WhoAmIAction;
 import org.opensearch.security.auditlog.AuditLog;
@@ -469,6 +474,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin 
                         Objects.requireNonNull(cs), Objects.requireNonNull(adminDns), Objects.requireNonNull(cr)));
                 handlers.add(new SecurityConfigUpdateAction(settings, restController, Objects.requireNonNull(threadPool), adminDns, configPath, principalExtractor));
                 handlers.add(new SecurityWhoAmIAction(settings, restController, Objects.requireNonNull(threadPool), adminDns, configPath, principalExtractor));
+                handlers.add(new TenancyConfigRestHandler());
                 handlers.addAll(
                         SecurityRestApiActions.getHandler(
                                 settings,
@@ -505,6 +511,10 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin 
         if(!disabled && !SSLConfig.isSslOnlyMode()) {
             actions.add(new ActionHandler<>(ConfigUpdateAction.INSTANCE, TransportConfigUpdateAction.class));
             actions.add(new ActionHandler<>(WhoAmIAction.INSTANCE, TransportWhoAmIAction.class));
+
+            actions.add(new ActionHandler<>(TenancyConfigRetrieveActions.INSTANCE, TenancyConfigRetrieveTransportAction.class));
+            actions.add(new ActionHandler<>(TenancyConfigUpdateAction.INSTANCE, TenancyConfigUpdateTransportAction.class));
+
         }
         return actions;
     }
--- a/src/main/java/org/opensearch/security/action/tenancy/EmptyRequest.java
+++ b/src/main/java/org/opensearch/security/action/tenancy/EmptyRequest.java
@@ -1,0 +1,35 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.action.tenancy;
+
+import java.io.IOException;
+
+import org.opensearch.action.ActionRequest;
+import org.opensearch.action.ActionRequestValidationException;
+import org.opensearch.common.io.stream.StreamInput;
+
+public class EmptyRequest extends ActionRequest {
+
+    public EmptyRequest(final StreamInput in) throws IOException {
+        super(in);
+    }
+
+    public EmptyRequest() throws IOException {
+        super();
+    }
+
+    @Override
+    public ActionRequestValidationException validate()
+    {
+        return null;
+    }
+}
--- a/src/main/java/org/opensearch/security/action/tenancy/TenancyConfigRestHandler.java
+++ b/src/main/java/org/opensearch/security/action/tenancy/TenancyConfigRestHandler.java
@@ -1,0 +1,65 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.action.tenancy;
+
+import java.io.IOException;
+import java.util.List;
+
+import com.google.common.collect.ImmutableList;
+
+import org.opensearch.client.node.NodeClient;
+import org.opensearch.rest.BaseRestHandler;
+import org.opensearch.rest.RestRequest;
+import org.opensearch.rest.action.RestToXContentListener;
+
+import static org.opensearch.rest.RestRequest.Method.GET;
+import static org.opensearch.rest.RestRequest.Method.PUT;
+
+public class TenancyConfigRestHandler extends BaseRestHandler {
+
+    public TenancyConfigRestHandler() {
+        super();
+    }
+
+    @Override
+    public String getName() {
+        return "Multi Tenancy actions to Retrieve / Update configs.";
+    }
+
+    @Override
+    public List<Route> routes() {
+        return ImmutableList.of(
+                new Route(GET, "/_plugins/_security/api/tenancy/config"),
+                new Route(PUT, "/_plugins/_security/api/tenancy/config")
+        );
+    }
+
+    @Override
+    protected RestChannelConsumer prepareRequest(final RestRequest request, final NodeClient nodeClient) throws IOException {
+
+        switch (request.method()) {
+            case GET:
+                return channel -> nodeClient.execute(
+                        TenancyConfigRetrieveActions.INSTANCE,
+                        new EmptyRequest(),
+                        new RestToXContentListener<>(channel));
+            case PUT:
+                return channel -> nodeClient.execute(
+                        TenancyConfigUpdateAction.INSTANCE,
+                        TenancyConfigUpdateRequest.fromXContent(request.contentParser()),
+                        new RestToXContentListener<>(channel));
+            default:
+                throw new RuntimeException("Not implemented");
+        }
+    }
+
+}
--- a/src/main/java/org/opensearch/security/action/tenancy/TenancyConfigUpdateTransportAction.java
+++ b/src/main/java/org/opensearch/security/action/tenancy/TenancyConfigUpdateTransportAction.java
@@ -1,0 +1,155 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.action.tenancy;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import org.opensearch.action.ActionListener;
+import org.opensearch.action.index.IndexResponse;
+import org.opensearch.action.support.ActionFilters;
+import org.opensearch.action.support.HandledTransportAction;
+import org.opensearch.client.Client;
+import org.opensearch.common.inject.Inject;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.security.configuration.ConfigurationRepository;
+import org.opensearch.security.dlic.rest.api.AbstractApiAction;
+import org.opensearch.security.securityconf.impl.CType;
+import org.opensearch.security.securityconf.impl.SecurityDynamicConfiguration;
+import org.opensearch.security.securityconf.impl.v7.ConfigV7;
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.tasks.Task;
+import org.opensearch.threadpool.ThreadPool;
+import org.opensearch.transport.TransportService;
+
+public class TenancyConfigUpdateTransportAction extends HandledTransportAction<TenancyConfigUpdateRequest, TenancyConfigRetrieveResponse> {
+
+    private static final Logger log = LogManager.getLogger(TenancyConfigUpdateTransportAction.class);
+
+    private final String securityIndex;
+    private final ConfigurationRepository config;
+    private final Client client;
+    private final ThreadPool pool;
+
+    @Inject
+    public TenancyConfigUpdateTransportAction(final Settings settings,
+                                              final TransportService transportService,
+                                              final ActionFilters actionFilters,
+                                              final ConfigurationRepository config,
+                                              final ThreadPool pool,
+                                              final Client client) {
+        super(TenancyConfigUpdateAction.NAME, transportService, actionFilters, TenancyConfigUpdateRequest::new);
+
+        this.securityIndex = settings.get(ConfigConstants.SECURITY_CONFIG_INDEX_NAME, ConfigConstants.OPENDISTRO_SECURITY_DEFAULT_CONFIG_INDEX);
+
+        this.config = config;
+        this.client = client;
+        this.pool = pool;
+    }
+
+    /** Load the configuration from the security index and return a copy */
+    protected final SecurityDynamicConfiguration<?> load() {
+        return config.getConfigurationsFromIndex(Collections.singleton(CType.CONFIG), false).get(CType.CONFIG).deepClone();
+    }
+
+    private Set<String> getAcceptableDefaultTenants() {
+        Set<String> acceptableDefaultTenants = new HashSet<String>();
+        acceptableDefaultTenants.add(ConfigConstants.TENANCY_GLOBAL_TENANT_DEFAULT_NAME);
+        acceptableDefaultTenants.add(ConfigConstants.TENANCY_GLOBAL_TENANT_NAME);
+        acceptableDefaultTenants.add(ConfigConstants.TENANCY_PRIVATE_TENANT_NAME);
+        return acceptableDefaultTenants;
+    }
+
+    private Set<String> getAllConfiguredTenantNames() {
+
+        return this.config.getConfiguration(CType.TENANTS).getCEntries().keySet();
+    }
+
+    protected void validate(ConfigV7 updatedConfig) {
+        if(!updatedConfig.dynamic.kibana.private_tenant_enabled && (updatedConfig.dynamic.kibana.default_tenant).equals(ConfigConstants.TENANCY_PRIVATE_TENANT_NAME)) {
+            throw new IllegalArgumentException("Private tenant can not be disabled if it is the default tenant.");
+        }
+
+        Set<String> acceptableDefaultTenants = getAcceptableDefaultTenants();
+
+        if(acceptableDefaultTenants.contains(updatedConfig.dynamic.kibana.default_tenant)) {
+            return;
+        }
+
+        Set<String> availableTenants = getAllConfiguredTenantNames();
+
+        if(!availableTenants.contains(updatedConfig.dynamic.kibana.default_tenant)){
+            throw new IllegalArgumentException(updatedConfig.dynamic.kibana.default_tenant + " can not be set to default tenant. Default tenant should be selected from one of the available tenants.");
+        }
+
+    }
+
+    @Override
+    protected void doExecute(final Task task, final TenancyConfigUpdateRequest request, final ActionListener<TenancyConfigRetrieveResponse> listener) {
+
+        // Get the current security config and prepare the config with the updated value
+        final SecurityDynamicConfiguration dynamicConfig = load();
+        final ConfigV7 config = (ConfigV7)dynamicConfig.getCEntry("config");
+
+        final TenancyConfigs tenancyConfigs = request.getTenancyConfigs();
+        if(tenancyConfigs.multitenancy_enabled != null)
+        {
+            config.dynamic.kibana.multitenancy_enabled = tenancyConfigs.multitenancy_enabled;
+        }
+
+        if(tenancyConfigs.private_tenant_enabled != null)
+        {
+            config.dynamic.kibana.private_tenant_enabled = tenancyConfigs.private_tenant_enabled;
+        }
+
+        if(tenancyConfigs.default_tenant != null)
+        {
+            config.dynamic.kibana.default_tenant = tenancyConfigs.default_tenant;
+        }
+
+        validate(config);
+
+        dynamicConfig.putCEntry("config", config);
+
+        // When performing an update to the configuration run as admin
+        try (final ThreadContext.StoredContext stashedContext = pool.getThreadContext().stashContext()) {
+            // Update the security configuration and make sure the cluster has fully refreshed
+            AbstractApiAction.saveAndUpdateConfigs(this.securityIndex, this.client, CType.CONFIG, dynamicConfig, new ActionListener<IndexResponse>(){
+
+                @Override
+                public void onResponse(final IndexResponse response) {
+                    // After processing the request, restore the user context
+                    stashedContext.close();
+                    try {
+                        // Lookup the current value and notify the listener
+                        client.execute(TenancyConfigRetrieveActions.INSTANCE, new EmptyRequest(), listener);
+                    } catch (IOException ioe) {
+                        log.error(ioe);
+                        listener.onFailure(ioe);
+                    }
+                }
+
+                @Override
+                public void onFailure(Exception e) {
+                    log.error(e);
+                    listener.onFailure(e);
+                }
+            });
+        }
+    }
+}
--- a/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
@@ -525,6 +525,15 @@ public class PrivilegesEvaluator {
                 && dcm.isDashboardsMultitenancyEnabled();
     }
 
+    public boolean privateTenantEnabled() {
+        return privilegesInterceptor.getClass() != PrivilegesInterceptor.class
+                && dcm.isDashboardsPrivateTenantEnabled();
+    }
+
+    public String dashboardsDefaultTenant() {
+        return dcm.getDashboardsDefaultTenant();
+    }
+
     public boolean notFailOnForbiddenEnabled() {
         return privilegesInterceptor.getClass() != PrivilegesInterceptor.class
                 && dcm.isDnfofEnabled();
--- a/src/test/java/org/opensearch/security/multitenancy/test/TenancyDefaultTenantTests.java
+++ b/src/test/java/org/opensearch/security/multitenancy/test/TenancyDefaultTenantTests.java
@@ -1,0 +1,96 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.multitenancy.test;
+
+import org.apache.hc.core5.http.Header;
+import org.apache.hc.core5.http.HttpStatus;
+import org.junit.Test;
+
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.security.test.SingleClusterTest;
+import org.opensearch.security.test.helper.rest.RestHelper.HttpResponse;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.IsEqual.equalTo;
+import static org.hamcrest.core.StringContains.containsString;
+
+public class TenancyDefaultTenantTests extends SingleClusterTest {
+    private final Header asAdminUser = encodeBasicHeader("admin", "admin");
+    private final Header asUser = encodeBasicHeader("kirk", "kirk");
+
+    @Override
+    protected String getResourceFolder() {
+        return "multitenancy";
+    }
+
+    @Test
+    public void testDefaultTenantUpdate() throws Exception {
+        setup();
+
+        final HttpResponse getSettingResponse = nonSslRestHelper().executeGetRequest("/_plugins/_security/api/tenancy/config", asAdminUser);
+        assertThat(getSettingResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertThat(getSettingResponse.findValueInJson("default_tenant"), equalTo(ConfigConstants.TENANCY_GLOBAL_TENANT_DEFAULT_NAME));
+
+        HttpResponse getDashboardsinfoResponse = nonSslRestHelper().executeGetRequest("/_plugins/_security/dashboardsinfo", asAdminUser);
+        assertThat(getDashboardsinfoResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertThat(getDashboardsinfoResponse.findValueInJson("default_tenant"), equalTo(ConfigConstants.TENANCY_GLOBAL_TENANT_DEFAULT_NAME));
+
+        final HttpResponse setPrivateTenantAsDefaultResponse = nonSslRestHelper().executePutRequest("/_plugins/_security/api/tenancy/config", "{\"default_tenant\": \"Private\"}", asAdminUser);
+        assertThat(setPrivateTenantAsDefaultResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        getDashboardsinfoResponse = nonSslRestHelper().executeGetRequest("/_plugins/_security/dashboardsinfo", asAdminUser);
+        assertThat(getDashboardsinfoResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertThat(getDashboardsinfoResponse.findValueInJson("default_tenant"), equalTo(ConfigConstants.TENANCY_PRIVATE_TENANT_NAME));
+    }
+
+    @Test
+    public void testDefaultTenant_UpdateFailed() throws Exception {
+        setup();
+
+        final HttpResponse disablePrivateTenantResponse = nonSslRestHelper().executePutRequest("/_plugins/_security/api/tenancy/config", "{\"private_tenant_enabled\":false}", asAdminUser);
+        assertThat(disablePrivateTenantResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+
+
+        final HttpResponse setPrivateTenantAsDefaultFailResponse = nonSslRestHelper().executePutRequest("/_plugins/_security/api/tenancy/config", "{\"default_tenant\": \"Private\"}", asAdminUser);
+        assertThat(setPrivateTenantAsDefaultFailResponse.getStatusCode(), equalTo(HttpStatus.SC_BAD_REQUEST));
+        assertThat(setPrivateTenantAsDefaultFailResponse.findValueInJson("error.reason"), containsString("Private tenant can not be disabled if it is the default tenant."));
+
+        final HttpResponse enablePrivateTenantResponse = nonSslRestHelper().executePutRequest("/_plugins/_security/api/tenancy/config", "{\"private_tenant_enabled\":true}", asAdminUser);
+        assertThat(enablePrivateTenantResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+
+        final HttpResponse setPrivateTenantAsDefaultResponse = nonSslRestHelper().executePutRequest("/_plugins/_security/api/tenancy/config", "{\"default_tenant\": \"Private\"}", asAdminUser);
+        assertThat(setPrivateTenantAsDefaultResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+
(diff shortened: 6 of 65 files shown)


DEVELOPER_GUIDE.md
```diff
@@ -45,16 +45,8 @@ cd $OPENSEARCH_HOME
...
-> 
-> The line to look for: `opensearch_version = System.getProperty("opensearch.version", "x")`
-> 
-> Alternatively, you can find the compatible version of OpenSearch by running in project root folder
-> ```
-> ./gradlew properties -q | grep -E '^version:' | awk '{print $2}'
-> ```
+> The version of OpenSearch and the security plugin must match as there is an explicit version check at startup. This can be a bit confusing as, for example, at the time of writing this guide, the `main` branch of this security plugin builds version `1.3.0.0-SNAPSHOT` compatible with OpenSearch `1.3.0-SNAPSHOT` that gets built from branch `1.x`. Check the expected compatible version [here](https://github.com/opensearch-project/security/blob/main/plugin-descriptor.properties#L27) and make sure you get the correct branch from OpenSearch when building that project.
```
anijain-Amazon, line 49: This change seems incorrect for main branch

src/test/java/org/opensearch/security/multitenancy/test/TenancyDefaultTenantTests.java
```diff
@@ -0,0 +1,88 @@
...
+        return "multitenancy";
+    }
+
+    @Test
+    public void testDefaultTenant_endToEndTest() throws Exception {
+        setup();
+
+        final HttpResponse getSettingResponse = nonSslRestHelper().executeGetRequest("/_plugins/_security/api/tenancy/default_tenant", asAdminUser);
+        assertThat(getSettingResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertThat(getSettingResponse.findValueInJson("value"), equalTo(""));
```
anijain-Amazon, line 41: nit: Can we read `""` from a constant for Global tenant instead.

src/main/java/org/opensearch/security/action/tenancy/DefaultTenantRestHandler.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    public List<Route> routes() {
+        return ImmutableList.of(
+                new Route(GET, "/_plugins/_security/api/tenancy/default_tenant"),
+                new Route(PUT, "/_plugins/_security/api/tenancy/default_tenant")
+        );
+    }
+
+    public void validate(final RestRequest request) throws Exception {
```
peternied, line 53: AuthZ checks are not enforced at the REST Layer, so these errors can disclose the security configuration details such as the list of tenants through error message flows.  This needs to be fixed

Validation should be moved into the transport layer actions.

src/main/java/org/opensearch/security/action/tenancy/DefaultTenantRestHandler.java
```diff
@@ -0,0 +1,116 @@
...
+
+        boolean isPrivateTenantEnabled = this.evaluator.privateTenantEnabled();
+        boolean isMultitenancyEnabled = this.evaluator.multitenancyEnabled();
+
+        if( !isMultitenancyEnabled ) {
+            throw new IllegalArgumentException("Default tenant can not be set if multi-tenancy is disabled.");
+        }
+
+        JSONParser parser = new JSONParser();
```
peternied, line 62: Remove the de/serialization code - Transport Actions should already have this handled.

src/main/java/org/opensearch/security/action/tenancy/DefaultTenantRestHandler.java
```diff
@@ -0,0 +1,116 @@
...
+        JSONObject request_content = (JSONObject) parser.parse(request.content().utf8ToString());
+
+        String default_tenant = (String) request_content.get("value");
+
+        if(default_tenant.equals(""))
+        {
+            return;
+        }
+
+        if(default_tenant.equals("Private")) {
```
peternied, line 72: Should not use inline literals for these values that would be consider breaking changes if they were modified - lets make this a static constant

src/main/java/org/opensearch/security/action/tenancy/DefaultTenantRestHandler.java
```diff
@@ -0,0 +1,116 @@
...
+            if(!isPrivateTenantEnabled)
+            {
+                throw new IllegalArgumentException("Default tenant can not be set to Private if Private tenant is disabled.");
```
peternied, line 75: Isn't the private tenant denoted as `__private__`?

src/main/java/org/opensearch/security/action/tenancy/EmptyRequest.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public EmptyRequest(final StreamInput in) throws IOException {
+        super(in);
+    }
+
+    public EmptyRequest() throws IOException {
+        super();
+    }
+
+	@Override
```
peternied, line 30: Nit: whitespace issue

src/main/java/org/opensearch/security/action/tenancy/DefaultTenantUpdateTransportAction.java
```diff
@@ -0,0 +1,106 @@
...
+    /** Load the configuration from the security index and return a copy */
+    protected final SecurityDynamicConfiguration<?> load() {
+        return config.getConfigurationsFromIndex(Collections.singleton(CType.CONFIG), false).get(CType.CONFIG).deepClone();
+    }
+
+
+
+    @Override
+    protected void doExecute(final Task task, final StringSettingUpdateRequest request, final ActionListener<StringSettingRetrieveResponse> listener) {
+
```
peternied, line 73: All of the transport action update methods have nearly the same code, lets refactor so its only written once.  I think you could try something like the following.

```java
public ??? doUpdate(Consumer<ConfigV7> updateAction, ...)`
```

Then for this method it could look something like...

```java
@Override
protected void doExecute(final Task task, final StringSettingUpdateRequest request, final ActionListener<StringSettingRetrieveResponse> listener) {
    doUpdate(
        (config) -> { config.dynamic.kibana.default_tenant = request.getValue(); },
        DefaultTenantRetrieveAction.INSTANCE,
        listener
    );
}
```

src/main/java/org/opensearch/security/action/tenancy/PrivateTenantEnabledRestHandler.java
```diff
@@ -0,0 +1,97 @@
...
+                        PrivateTenantEnabledRetrieveAction.INSTANCE,
+                        new EmptyRequest(),
+                        new RestToXContentListener<>(channel));
+            case PUT:
+                try {
+                    validate(request);
+                } catch (Exception e)
+                {
+                    throw new IllegalArgumentException(e.getMessage());
+                }
```
peternied, line 88: If this function throws exceptions, it shouldn't be wrapped.  Including this comment for when the validation logic is moved into the transport layer.

src/main/java/org/opensearch/security/action/tenancy/StringSettingUpdateRequest.java
```diff
@@ -0,0 +1,64 @@
...
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    @Override
+    public ActionRequestValidationException validate() {
+        if (getValue() == null || Objects.equals(getValue(), "Abhi1")) {
```
peternied, line 44: This does not look correct

src/test/java/org/opensearch/security/multitenancy/test/TenancyDefaultTenantTests.java
```diff
@@ -0,0 +1,88 @@
...
+        assertThat(getAuthinfoResponse.findValueInJson("default_tenant"),equalTo(""));
+
+        final HttpResponse disablePrivateTenantResponse = nonSslRestHelper().executePutRequest("/_plugins/_security/api/tenancy/private_tenant_enabled", "{\"value\":false}", asAdminUser);
+        assertThat(disablePrivateTenantResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+
+
+
+        final HttpResponse setPrivateTenantAsDefaultFailResponse = nonSslRestHelper().executePutRequest("/_plugins/_security/api/tenancy/default_tenant", "{\"value\": \"Private\"}", asAdminUser);
+        assertThat(setPrivateTenantAsDefaultFailResponse.getStatusCode(), equalTo(HttpStatus.SC_BAD_REQUEST));
+        assertThat(setPrivateTenantAsDefaultFailResponse.findValueInJson("error.reason"), containsString("Default tenant can not be set to Private if Private tenant is disabled."));
```
peternied, line 53: Separate out error test cases into their own test case to make it easier to read the flow of the end to end test case

src/test/java/org/opensearch/security/multitenancy/test/TenancyDefaultTenantTests.java
```diff
@@ -0,0 +1,88 @@
...
+        assertThat(enablePrivateTenantResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+
+        final HttpResponse setPrivateTenantAsDefaultResponse = nonSslRestHelper().executePutRequest("/_plugins/_security/api/tenancy/default_tenant", "{\"value\": \"Private\"}", asAdminUser);
+        assertThat(setPrivateTenantAsDefaultResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+
+        final HttpResponse getSettingResponseAfterUpdate = nonSslRestHelper().executeGetRequest("/_plugins/_security/api/tenancy/default_tenant", asAdminUser);
+        assertThat(getSettingResponseAfterUpdate.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertThat(getSettingResponseAfterUpdate.findValueInJson("value"), equalTo("Private"));
+
+        getAuthinfoResponse = nonSslRestHelper().executeGetRequest("/_plugins/_security/authinfo", asAdminUser);
```
peternied, line 65: End to end test should replicate the workflow that is used by dashboards.  If we just call the config APIs / confirm we are reading out the same value we put in it only indicates that the config was updated not that the users are seeing the correct behavior.

Same issue for the TenancyPrivateTenantEnabledTests

src/test/java/org/opensearch/security/multitenancy/test/TenancyPrivateTenantEnabledTests.java
```diff
@@ -0,0 +1,97 @@
...
+        assertThat(getSettingResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertThat(getSettingResponse.findValueInJson("value"), equalTo("true"));
+
+        HttpResponse getAuthinfoResponse = nonSslRestHelper().executeGetRequest("/_plugins/_security/authinfo", asAdminUser);
+        assertThat(getAuthinfoResponse.findValueInJson("private_tenant_enabled"),equalTo("true"));
+
+        final HttpResponse createDocInGlobalTenantResponse = nonSslRestHelper().executePostRequest(".kibana/_doc?refresh=true", createIndexPatternDoc("globalIndex"), asAdminUser);
+        assertThat(createDocInGlobalTenantResponse.getStatusCode(), equalTo(HttpStatus.SC_CREATED));
+        final HttpResponse createDocInUserTenantResponse = nonSslRestHelper().executePostRequest(".kibana/_doc?refresh=true", createIndexPatternDoc("userIndex"), onUserTenant, asAdminUser);
+        assertThat(createDocInUserTenantResponse.getStatusCode(), equalTo(HttpStatus.SC_CREATED));
```
peternied, line 59: This documents aren't used to verify the end to end scenario, they should be used or removed along with the `createIndexPatternDoc()` method

peternied (changes requested):
Thanks for iterating on this @abhivka7 huge improvement.  Many comments on here, but most of these should be fairly straight forward.  Let me know if you need help.

Another area I'd like us to shine some visibility on is the permission names and the REST API paths.  Can you enumerate those in a comment and walk through them?

From the draft pull request there was a lot of back and forth and those permissions / paths never really gelled, lets make sure we nail these down since we will have to live with them for a long time.

DEVELOPER_GUIDE.md
abhivka7: My bad.

anijain-Amazon:
@abhivka7 We are using separate Rest Actions for every configuration being added as part of this PR like `default_tenant` and `private_tenant` 

Since these are individual properties, the approach will require adding multiple classes for every property added in the future as well. This will in-turn end up increasing the metaspace of OS JVM 

Can we try an approach where separate endpoints/routes are not required for every property ?

peternied:
@anijain-Amazon Thanks for the comments, data consistency was a considerable trade-offs with the previous iteration that used a single API, see the discussion in https://github.com/opensearch-project/security/pull/2444.

We've got an issue centered around granular access of security config, we would welcome your feedback in this space.

- https://github.com/opensearch-project/security/issues/2559


> Since these are individual properties, the approach will require adding multiple classes for every property added in the future as well. This will in-turn end up increasing the metaspace of OS JVM.

This is a good consideration.  Is this an immediate issue, can you quantify the severity?

I expect all OpenSearch components + plugins + extensions to be impacting the memory usage of OpenSearch as we incorporate more features.  I don't see an issue in the OpenSearch repo, could you create one?

anijain-Amazon:
> data consistency was a considerable trade-offs with the previous iteration that used a single API, see the discussion in https://github.com/opensearch-project/security/pull/2444.

I could see the comments in previous iteration are w.r.t. using a new document vs reusing the existing security config document. Didn't find anything mentioning the flaws of using a single API. 

Having separate APIs for every property requires the users/clients to take the overhead of ensuring transaction while updating multiple properties OR consider the drawbacks of having partial failures. For instance, if a user wants to update `propertyA` and `propertyB`, then either they have to do it individually OR do it in a transaction to ensure consistency.

> We've got an issue centered around granular access of security config, we would welcome your feedback in this space.

Ack will take a look into it. This should help bring more flexibility for OpenSearch users to control admin privileges. 

> This is a good consideration. Is this an immediate issue, can you quantify the severity?

My concern is how will this approach scale with more properties in future. From a quick glance at this PR, it looks like we are adding at least 3 classes per property which can be simplified. Although, this will immediately not affect OpenSearch but will sure eat up the JVM Metaspace in future with say like 30-50 properties requiring 60-150 classes for updating their settings.

anijain-Amazon:
> I expect all OpenSearch components + plugins + extensions to be impacting the memory usage of OpenSearch as we incorporate more features. I don't see an issue in the OpenSearch repo, could you create one?

Since the classes are statically referenced, I expect OpenSearch will detect any increase in its JVM Metaspace while running performance tests. IMHO as the plugin contributors we should try to restrict the number of classes until absolutely mandated by the underlying design pattern.

prabhat-chaturvedi:
This PR should have v2.7.0 label on this PR? @peternied can you help please?

src/main/java/org/opensearch/security/rest/SecurityInfoAction.java
```diff
@@ -112,6 +112,9 @@ public void accept(RestChannel channel) throws Exception {
                     builder.field("principal", (String)threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_SSL_PRINCIPAL));
                     builder.field("peer_certificates", certs != null && certs.length > 0 ? certs.length + "" : "0");
                     builder.field("sso_logout_url", (String)threadContext.getTransient(ConfigConstants.SSO_LOGOUT_URL));
+                    builder.field("multitenancy_enabled", evaluator.multitenancyEnabled());
```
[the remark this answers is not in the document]
abhivka7: Understood. Implementing in new revision.

src/main/java/org/opensearch/security/action/tenancy/DefaultTenantRestHandler.java
abhivka7: Thanks @peternied for bringing this up. Will move validation to transport layer.

src/main/java/org/opensearch/security/action/tenancy/DefaultTenantRestHandler.java
abhivka7: There is some ambiguity around that. For eg: [Here](https://github.com/opensearch-project/security-dashboards-plugin/blob/b3b3120144e50d229bb09aec50991a383df6dafb/public/apps/configuration/utils/tenant-utils.tsx#L116)

When (tenant === '__user__') we resolve tenant name to 'Private'. I will be using the same in frontend changes.

src/main/java/org/opensearch/security/action/tenancy/DefaultTenantUpdateTransportAction.java
abhivka7: @peternied As per suggestions from @anijain-Amazon, I am implementing one single API and therefore single class for updating all properties.

src/main/java/org/opensearch/security/action/tenancy/PrivateTenantEnabledRestHandler.java
abhivka7: OK. Thanks @peternied.

abhivka7:
Thanks a lot @anijain-Amazon for bringing this up. After a lot of considerations and POC, I have decided to move to a single API model for cleaner user experience and better metaspace utilisation. I will raise a new revision with that approach. Please take a look and give your feedbacks. 
cc: @prabhat-chaturvedi @shikharj05 @devardee @peternied

src/test/java/org/opensearch/security/multitenancy/test/TenancyDefaultTenantTests.java
abhivka7: Those tests will be covered in Security Dashboards plugin PR. Anyways the Dashboards also will be calling this API (DashboardsInfo) as replicated in test to get final config so that should be fine.

src/main/java/org/opensearch/security/action/tenancy/DefaultTenantRestHandler.java
abhivka7: Sure. Implemented in new commit.

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsRestHandler.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public String getName() {
+        return "Multi Tenancy actions for Retrieve / Update configurations.";
+    }
+
+    @Override
+    public List<Route> routes() {
+        return ImmutableList.of(
+                new Route(GET, "/_plugins/_security/api/tenancy/multitenancy_configurations"),
```
peternied, line 41: Seems like this is a good opportunity to decouple this API from the security plugin if we want to separate tenancy features in the feature, what do you think of `/_tenancy/config`?

If that is too drastic, what about `/_plugins/_security/api/tenancy/config`?

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsRetrieveActions.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package org.opensearch.security.action.tenancy;
+
+import org.opensearch.action.ActionType;
+
+public class MultiTenancyConfigurationsRetrieveActions extends ActionType<MultiTenancyConfigurationsRetrieveResponse> {
+
+    public static final org.opensearch.security.action.tenancy.MultiTenancyConfigurationsRetrieveActions INSTANCE = new org.opensearch.security.action.tenancy.MultiTenancyConfigurationsRetrieveActions();
+    public static final String NAME = "cluster:feature/tenancy/multitenancy_configurations/read";
```
peternied, line 19:
```suggestion
    public static final String NAME = "cluster:feature/tenancy/config/read";
```

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsRestHandler.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.opensearch.client.node.NodeClient;
+import org.opensearch.rest.BaseRestHandler;
+import org.opensearch.rest.RestRequest;
+import org.opensearch.rest.action.RestToXContentListener;
+
+import static org.opensearch.rest.RestRequest.Method.GET;
+import static org.opensearch.rest.RestRequest.Method.PUT;
+
+public class MultiTenancyConfigurationsRestHandler extends BaseRestHandler {
```
peternied, line 27: The name doesn't seem right, rename all these classes to `TenancyConfig` in place of `MultiTenancyConfigurations`.  Its a single configuration and mutli-tenancy is overly verbose.  We shouldn't change the config value for backwards compatibility, but the new classes don't need to follow suite.

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsUpdateAction.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package org.opensearch.security.action.tenancy;
+
+import org.opensearch.action.ActionType;
+
+public class MultiTenancyConfigurationsUpdateAction extends ActionType<MultiTenancyConfigurationsRetrieveResponse> {
+
+    public static final org.opensearch.security.action.tenancy.MultiTenancyConfigurationsUpdateAction INSTANCE = new org.opensearch.security.action.tenancy.MultiTenancyConfigurationsUpdateAction();
+    public static final String NAME = "cluster:feature/tenancy/multitenancy_configurations/update";
```
peternied, line 19:
```suggestion
    public static final String NAME = "cluster:feature/tenancy/config/update";
```

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsUpdateTransportAction.java
```diff
@@ -0,0 +1,146 @@
...
+
+    /** Load the configuration from the security index and return a copy */
+    protected final SecurityDynamicConfiguration<?> load() {
+        return config.getConfigurationsFromIndex(Collections.singleton(CType.CONFIG), false).get(CType.CONFIG).deepClone();
+    }
+
+    public Set<String> getAllConfiguredTenantNames() {
+
+        return this.config.getConfiguration(CType.TENANTS).getCEntries().keySet();
+    }
```
peternied, line 72: Move this to be inline, or make private.

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsUpdateTransportAction.java
```diff
@@ -0,0 +1,146 @@
...
+
+    protected void validate(ConfigV7 updatedConfig) throws RuntimeException {
+        if(!updatedConfig.dynamic.kibana.private_tenant_enabled && (updatedConfig.dynamic.kibana.default_tenant).equals(ConfigConstants.MULTITENANCY_PRIVATE_TENANT_NAME) ) {
```
peternied, line 75: Please clean up the spacing using a formatter.

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsUpdateTransportAction.java
```diff
@@ -0,0 +1,146 @@
...
+    protected final SecurityDynamicConfiguration<?> load() {
+        return config.getConfigurationsFromIndex(Collections.singleton(CType.CONFIG), false).get(CType.CONFIG).deepClone();
+    }
+
+    public Set<String> getAllConfiguredTenantNames() {
+
+        return this.config.getConfiguration(CType.TENANTS).getCEntries().keySet();
+    }
+
+    protected void validate(ConfigV7 updatedConfig) throws RuntimeException {
```
peternied, line 74: RuntimeException by definition isn't a checked exception, please remove
```suggestion
    protected void validate(ConfigV7 updatedConfig) {
```

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsUpdateTransportAction.java
```diff
@@ -0,0 +1,146 @@
...
+        if(!updatedConfig.dynamic.kibana.private_tenant_enabled && (updatedConfig.dynamic.kibana.default_tenant).equals(ConfigConstants.MULTITENANCY_PRIVATE_TENANT_NAME) ) {
+            throw new IllegalArgumentException("Private tenant can not be disabled if it is the default tenant.");
+        }
+
+        if( (updatedConfig.dynamic.kibana.default_tenant).equals(ConfigConstants.MULTITENANCY_GLOBAL_TENANT_DEFAULT_NAME) ||
+                (updatedConfig.dynamic.kibana.default_tenant).equals(ConfigConstants.MULTITENANCY_GLOBAL_TENANT_NAME) ||
+                (updatedConfig.dynamic.kibana.default_tenant).equals(ConfigConstants.MULTITENANCY_PRIVATE_TENANT_NAME) ) {
```
peternied, line 81: This is hard to parse and isn't clear at a glance.  Instead create a collection called something like acceptableDefaultTenants that contains those 3 names from the constants, then check if `updatedConfig.dynamic.kibana.default_tenant` is in that collection and return

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsUpdateTransportAction.java
```diff
@@ -0,0 +1,146 @@
...
+            return;
+        }
+
+        Set<String> availableTenants = getAllConfiguredTenantNames();
+
+        if(!availableTenants.contains(updatedConfig.dynamic.kibana.default_tenant)){
+            throw new IllegalArgumentException("Default tenant should be selected from one of the available tenants.");
```
peternied, line 88: The error message should include the name of the tenant that was not found for troubleshooting

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -525,6 +525,15 @@ public boolean multitenancyEnabled() {
                 && dcm.isDashboardsMultitenancyEnabled();
     }
 
+    public boolean privateTenantEnabled() {
+        return privilegesInterceptor.getClass() != PrivilegesInterceptor.class
```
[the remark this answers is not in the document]
peternied: As long as this is exercised in end to end tests, ok.

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsRestHandler.java
peternied: Note; there are many places where the term 'configurations' is used, please do a pass over the whole PR to clean these up.

src/main/java/org/opensearch/security/action/tenancy/EmptyRequest.java
peternied: Still looks like the spacing is off in this file.

src/test/java/org/opensearch/security/multitenancy/test/TenancyDefaultTenantTests.java
peternied: We need the end to end coverage of this scenarios in the security plugin.  Otherwise breaking changes could be make in this repo.

The example from TenancyMultitenancyEnabledTests.java shows how when the feature is activated behavior changes, please use this as a starting point.  Let me know if you need help to test these scenarios.

src/test/java/org/opensearch/security/multitenancy/test/TenancyDefaultTenantTests.java
```diff
@@ -0,0 +1,96 @@
...
+
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.security.test.SingleClusterTest;
+import org.opensearch.security.test.helper.rest.RestHelper.HttpResponse;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.IsEqual.equalTo;
+import static org.hamcrest.core.StringContains.containsString;
+
+public class TenancyDefaultTenantTests extends SingleClusterTest {
```
peternied, line 26: Note, I know the refactor changes from 3 APIs to 1 and I think the tests being separated like this is fine 👍

peternied (changes requested):
Refactoring looks good, notable callouts:
- API / permissions naming needs iteration, I've made suggestions, but product team might need to be engaged.
- Need end to end verification in the test cases.  Cannot depend on test cases in other repositories.
- Looks like many test cases are broken. Check the CI results.

src/main/java/org/opensearch/security/action/tenancy/EmptyRequest.java
abhivka7: Github was showing a tab differently than my IDE. Fixing.

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurations.java
```diff
@@ -0,0 +1,18 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.action.tenancy;
+
+public class MultiTenancyConfigurations {
```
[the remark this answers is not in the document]
abhivka7: Going forward we want to contribute more to tenancy in OpenSearch and therefore add new properties. It makes more sense to have an object for same for better accessibility and keeping everything together in a structured form.

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsRestHandler.java
abhivka7: Cool.

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsRestHandler.java
abhivka7: Cool. Going with ```/_plugins/_security/api/tenancy/config```

src/main/java/org/opensearch/security/action/tenancy/MultiTenancyConfigurationsUpdateTransportAction.java
abhivka7: Cool.

src/test/java/org/opensearch/security/multitenancy/test/TenancyDefaultTenantTests.java
abhivka7: I have added that for TenancyPrivateTenantEnabledTests. Not sure what else to add for Default tenant tests. What test scenario are you thinking @peternied ?

src/main/java/org/opensearch/security/action/tenancy/TenancyConfigUpdateTransportAction.java
```diff
@@ -0,0 +1,155 @@
...
+        this.client = client;
+        this.pool = pool;
+    }
+
+    /** Load the configuration from the security index and return a copy */
+    protected final SecurityDynamicConfiguration<?> load() {
+        return config.getConfigurationsFromIndex(Collections.singleton(CType.CONFIG), false).get(CType.CONFIG).deepClone();
+    }
+
+    private Set<String> getAcceptableDefaultTenants() {
```
peternied, line 70: Nit: this is a good candidate for a static immutable collection

src/test/java/org/opensearch/security/multitenancy/test/TenancyDefaultTenantTests.java
```diff
@@ -0,0 +1,96 @@
...
+        HttpResponse getDashboardsinfoResponse = nonSslRestHelper().executeGetRequest("/_plugins/_security/dashboardsinfo", asAdminUser);
+        assertThat(getDashboardsinfoResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertThat(getDashboardsinfoResponse.findValueInJson("default_tenant"), equalTo(ConfigConstants.TENANCY_GLOBAL_TENANT_DEFAULT_NAME));
+
+        final HttpResponse setPrivateTenantAsDefaultResponse = nonSslRestHelper().executePutRequest("/_plugins/_security/api/tenancy/config", "{\"default_tenant\": \"Private\"}", asAdminUser);
+        assertThat(setPrivateTenantAsDefaultResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        getDashboardsinfoResponse = nonSslRestHelper().executeGetRequest("/_plugins/_security/dashboardsinfo", asAdminUser);
+        assertThat(getDashboardsinfoResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertThat(getDashboardsinfoResponse.findValueInJson("default_tenant"), equalTo(ConfigConstants.TENANCY_PRIVATE_TENANT_NAME));
+    }
```
peternied, line 52: Lets tweak this test case so that a document is made in the private tenant index and the global.  Before switching the value, check that the expected document is found when no tenant is provided, when specifying the tenant the expected document is found.  Then the same after the value is updated.

peternied (commented):
Thanks for the update, one last change to TenancyDefaultTenantTests and then it looks good to me

peternied:
@scrawfor99 Could you take a look at this PR?

scrawfor99:
> @scrawfor99 Could you take a look at this PR?

I reviewed the frontend changes yesterday, sorry for the confusion. I will review these now.

src/test/java/org/opensearch/security/multitenancy/test/TenancyMultitenancyEnabledTests.java
```diff
@@ -0,0 +1,89 @@
...
+        assertThat(getDashboardsinfoResponse.findValueInJson("multitenancy_enabled"),equalTo("true"));
+
+        final HttpResponse createDocInGlobalTenantResponse = nonSslRestHelper().executePostRequest(".kibana/_doc?refresh=true", createIndexPatternDoc("globalIndex"), asAdminUser);
+        assertThat(createDocInGlobalTenantResponse.getStatusCode(), equalTo(HttpStatus.SC_CREATED));
+        final HttpResponse createDocInUserTenantResponse = nonSslRestHelper().executePostRequest(".kibana/_doc?refresh=true", createIndexPatternDoc("userIndex"), onUserTenant, asAdminUser);
+        assertThat(createDocInUserTenantResponse.getStatusCode(), equalTo(HttpStatus.SC_CREATED));
+
+        final HttpResponse searchInUserTenantWithMutlitenancyEnabled = nonSslRestHelper().executeGetRequest(".kibana/_search", onUserTenant, asAdminUser);
+        assertThat(searchInUserTenantWithMutlitenancyEnabled.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertThat(searchInUserTenantWithMutlitenancyEnabled.findValueInJson("hits.hits[0]._source.index-pattern.title"), equalTo("userIndex"));
```
scrawfor99, line 63: Perhaps you can add another test with the unhappy case for this scenario? I.e. trying to find the value of `userIndex` when MultiTenancy is disabled?

scrawfor99 (commented):
Great job with this! I left just a few comments here and there and a couple questions. Overall, I think this is in great shape.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -505,6 +511,10 @@ public UnaryOperator<RestHandler> getRestHandlerWrapper(final ThreadContext thre
         if(!disabled && !SSLConfig.isSslOnlyMode()) {
             actions.add(new ActionHandler<>(ConfigUpdateAction.INSTANCE, TransportConfigUpdateAction.class));
             actions.add(new ActionHandler<>(WhoAmIAction.INSTANCE, TransportWhoAmIAction.class));
+
+            actions.add(new ActionHandler<>(TenancyConfigRetrieveActions.INSTANCE, TenancyConfigRetrieveTransportAction.class));
```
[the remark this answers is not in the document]
anijain-Amazon: Just a guess..I think when we show the current configuration to the user on Dashboards this API is also used to fetch the applicable tenants when their is none saved in the local storage. 

Personally, I would find it super useful while debugging issues related to incorrect tenancy access.

src/main/java/org/opensearch/security/action/tenancy/TenancyConfigRestHandler.java
```diff
@@ -0,0 +1,65 @@
...
+                        TenancyConfigRetrieveActions.INSTANCE,
+                        new EmptyRequest(),
+                        new RestToXContentListener<>(channel));
+            case PUT:
+                return channel -> nodeClient.execute(
+                        TenancyConfigUpdateAction.INSTANCE,
+                        TenancyConfigUpdateRequest.fromXContent(request.contentParser()),
+                        new RestToXContentListener<>(channel));
+            default:
+                throw new RuntimeException("Not implemented");
```
anijain-Amazon, line 61: We can also add the value of `request.method()` if it's not implemented for better context when it's logged.

src/main/java/org/opensearch/security/action/tenancy/EmptyRequest.java
```diff
@@ -0,0 +1,35 @@
...
+
+package org.opensearch.security.action.tenancy;
+
+import java.io.IOException;
+
+import org.opensearch.action.ActionRequest;
+import org.opensearch.action.ActionRequestValidationException;
+import org.opensearch.common.io.stream.StreamInput;
+
+public class EmptyRequest extends ActionRequest {
```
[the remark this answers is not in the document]
anijain-Amazon: This is used as the generic type for [HandledTransportAction](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/HandledTransportAction.java#L48) as it expects an instance of [ActionRequest](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/ActionRequest.java)

src/main/java/org/opensearch/security/action/tenancy/TenancyConfigUpdateTransportAction.java
anijain-Amazon: +1 And, we can use more readable initialisers like `Set.of()`

src/main/java/org/opensearch/security/action/tenancy/EmptyRequest.java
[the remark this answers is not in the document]
anijain-Amazon: It is being used for `TenancyConfigRetrieveTransportAction`

src/main/java/org/opensearch/security/action/tenancy/TenancyConfigUpdateTransportAction.java
```diff
@@ -0,0 +1,155 @@
...
+        Set<String> acceptableDefaultTenants = getAcceptableDefaultTenants();
+
+        if(acceptableDefaultTenants.contains(updatedConfig.dynamic.kibana.default_tenant)) {
+            return;
+        }
+
+        Set<String> availableTenants = getAllConfiguredTenantNames();
+
+        if(!availableTenants.contains(updatedConfig.dynamic.kibana.default_tenant)){
+            throw new IllegalArgumentException(updatedConfig.dynamic.kibana.default_tenant + " can not be set to default tenant. Default tenant should be selected from one of the available tenants.");
```
anijain-Amazon, line 97: Intentionally not adding the `availableTenants` in the logs.

src/test/java/org/opensearch/security/multitenancy/test/TenancyDefaultTenantTests.java
peternied: Oops - the `default_tenant` doesn't change behavior of the backend system, the value is passed via dashboardsinfo to Dashboards where it is used.  No changes are needed - thanks for calling this out.

peternied (approved):

scrawfor99 (approved):
Hi @abhivka7, I see that a few of my questions got answered. Reviewing the remaining ones and the state of the PR, I am fine to merge this without any further conversation. Thank you for the contribution and being so quick to respond throughout. Great job!

abhivka7:
Thanks @peternied and @scrawfor99 for reviewing and merging this PR. Also thanks @prabhat-chaturvedi @anijain-Amazon @devardee @shikharj05 @cliu123 for helping with this feature.

Issue #2615, opened 2023-03-31
[Extensions] Create dynamic configuration section in config.yml for extensions


cwperks:
For the initial experimental release of extensions, the security plugin needs to support a couple dynamic configuration options to support the issuance and verification of auth tokens. All configuration options should fall under a new key in the `config.dynamic` portion of the security plugin `config.yml` file. Below is an example of the dynamic configuration options.

```
config:
  dynamic:
    extensions:
      signing_key: <base64_encoded_signing_key>
      encryption_key: <base64_encoded_encryption_key>
```

The 2 configuration options above represent 2 initial options needed to support the auth token workflows, but this `extensions:` area of the config.yml file may be extended in the future with more dynamic configuration options.

The 2 settings above represent:

1. `signing_key` - This is a base64 encoded secret that will be used to create a JWK to sign the JWTs that are issued - by default the signing is done using symmetric encryption (HMAC SHA512). More configurable keys will be supported in the future, but the default is chosen for the experimental release.
2. `encryption_key` - This is another base64 encoded secret, but this will be utilized to encrypt sensitive information in the payload of the JWT. 

`signing_key` and `encryption_key` are also utilized for an authentication backend to support these tokens issued by the security plugin and used to verify the tokens and decrypt the encrypted sections of the payload.

There are 2 java files around config in this codebase: ConfigV6 and ConfigV7. This change will only support OS >=2 so changes can be only applied into ConfigV7.

peternied:
Why does an Admin need to be able to provide there own key?  What do you think about the security plugin manage the generation on its own, and supply an API to rotate the keys?

"Extensions needs this feature" doesn't translate to the feature name needs to include the work extensions.  To me this is the DelegatingAuthenticationBackend - but I'm happy to consider other names.

scrawfor99:
[Triage] This is part of the Extensions project.

MaciejMierzwa:
Hi,
Are there any more decisions pending for this task? If not my understanding, for now, is that we want to add configuration from the example above and make sure it's loading correctly into [ConfigV7](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/impl/v7/ConfigV7.java) file. Is that correct?

cwperks:
[...]

PR #2671, opened 2023-04-11
Extensions config for JWT signing/encryption key

MaciejMierzwa:
### Description
This PR adds a section for extension configuration with signing/encryption keys for JWT generation.

### Issues Resolved
[Create dynamic configuration section in config.yml for extensions](https://github.com/opensearch-project/security/issues/2615)

### Testing
Manual tests were performed by checking if configurations were deserialized into POJO correctly. Existing tests in: [ConfigTests](https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/ConfigTests.java) check if the config file is deserialized correctly

### Check List


cwperks:
Hi @MaciejMierzwa , could you please sign the commits with `git commit --signoff`? You can see more information about DCO here: https://github.com/opensearch-project/.github/blob/main/CONTRIBUTING.md#developer-certificate-of-origin

src/main/java/org/opensearch/security/securityconf/impl/v7/ConfigV7.java
```diff
@@ -33,10 +33,7 @@
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
-import com.fasterxml.jackson.annotation.JsonAnyGetter;
-import com.fasterxml.jackson.annotation.JsonAnySetter;
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.*;
```
cwperks, line 36: If you are using Intellij, can you set it to no wildcard imports? This post from SO should help with setup for no wildcard imports: https://stackoverflow.com/a/3348855

src/main/java/org/opensearch/security/securityconf/impl/v7/ConfigV7.java
```diff
@@ -123,6 +120,7 @@ public static class Dynamic {
         public String hosts_resolver_mode = "ip-only";
         public String transport_userrname_attribute;
         public boolean do_not_fail_on_forbidden_empty;
+        public Extensions extensions = new Extensions();
     
         @Override
         public String toString() {
```
cwperks, line 126: Can you update this `toString` and also add `toString()` to the `Extensions` class similar to how `Kibana` is done?

cwperks (commented):
Thank you for opening a PR @MaciejMierzwa!

Can we also add 2 helped methods on [DynamicConfigFactory](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java) to get these values from the dynamic configuration?

In this abstract class there can be 2 new abstract methods:

```
public abstract String getExtensionsSigningKey();
public abstract String getExtensionsEncryptionKey();
```

There are 2 subclasses of this class: `DynamicConfigModelV6` and `DynamicConfigModelV7`. 

For V6 you can return these values as null since these are not pertinent for V6, for V7 they can be obtained by `config.dynamic.extensions.signing_key` and `config.dynamic.extensions.encryption_key`.

src/main/java/org/opensearch/security/securityconf/impl/v7/ConfigV7.java
MaciejMierzwa: Is there a use case for allowing this part of the configuration to be printed? Since it stores encryption/signing keys I think it would be a good idea to hide it as much as possible.

MaciejMierzwa:
Regarding [comment](https://github.com/opensearch-project/security/pull/2671#pullrequestreview-1379654910) about adding abstract methods. What do you think about just adding concrete methods into DynamicConfigModelV7 (without adding those to the abstract class)? I could also separate DynamicConfigModel abstract class into 2 interfaces, or maybe add changes to V6 class so config would be also available there.
Returning null looks like a programming smell that will bite somebody sometime in the future.

cwperks:
@MaciejMierzwa Any V6 classes you see are for the old security configuration that supported ES6 and before. From ES7 and all versions of OpenSearch, V7 is the configuration that is utilized. For extensions related work, it will only be supported in OpenSearch so its only applicable to V7. If you added the same Extension section to `ConfigV6` and referenced those in the method implementations in `DynamicConfigModelV6` than that would also be ok, but it would never be used.

There are some methods that may return null if no value has been set in the `config`. See `Kibana.opendistro_role` which is set to `null` by default:

- https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/impl/v6/ConfigV6.java#L91
- https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV6.java#L131-L134

src/main/java/org/opensearch/security/securityconf/impl/v7/ConfigV7.java
MaciejMierzwa: So it's ok to include this secret strings in toString() methods. Pushed my changes

cwperks:
This PR connects with a Draft PR that @RyanL1997 opened yesterday which consumes these settings in an Authenticator for the tokens that will be issued for extensions. 

See https://github.com/opensearch-project/security/pull/2672/files#diff-efd217975f23af4bb84808e2771bfc85211f12b59bb619da430a3363902207ceR264-R272 for one of the locations these settings will be consumed. The other location will they will be consumed is the [JWTVendor](https://github.com/opensearch-project/security/blob/feature/extensions/src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java) where they are used to sign the token that is issued and encrypt sensitive claims like the roles and the backend roles.

cwperks:
@MaciejMierzwa On second thought, I wonder if it would be possible to have a single Abstract method. Something like:

```
public abstract Settings getDynamicExtensionsSettings();
```

The `signing_key` is used for the default symmetric HS512 signing for the JWT issued for extensions, but there is an open issue for JWK configurability: https://github.com/opensearch-project/security/issues/2680

The `http_authenticator` config is returned as a settings Object for an `http_authenticator`. See below for an example JWT backend config.

```
jwt_auth_domain:
  enabled: true
  order: 0
  http_authenticator:
    type: jwt
    challenge: false
    config:
      signing_key: "base64 encoded key"
      jwt_header: "Authorization"
      jwt_url_parameter: null
      subject_key: null
      roles_key: null
  authentication_backend:
     type: noop
```

Everything in the `config:` area of this backend would be returned as a `Settings` Object.

See relevant areas of the code:

1. https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java#L279

```
Settings.builder().loadFromSource(ad.getValue().http_authenticator.configAsJson(), XContentType.JSON).build()
```

2. https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/impl/v7/ConfigV7.java#L324-L331

Would it be possible to make a single abstract method to return all dynamic settings for extensions?

scrawfor99 (commented):
Hi Maciej, thank you for this. Overall your changes look great. If you could add a couple tests to verify the expected behavior of your changes, I think that would be all that needs to be done for this to be merged. For instance, you could confirm that the config changes do in fact take place in a couple short tests.

MaciejMierzwa:
Hi, @cwperks let me know if the changes that I made in the latest commit are more or less what you meant. I replaced method returning Settings object instead of two following methods:
`    public abstract String getExtensionsSigningKey();
    public abstract String getExtensionsEncryptionKey();`

cwperks:
Hi @MaciejMierzwa, the new changes look great! Just a couple of things remaining and then this looks good to go:

1. Can you merge the latest changes from main into this branch? There was a build fixing [PR](https://github.com/opensearch-project/security/pull/2679) that renewed some expired certs used in tests
2. As @scrawfor99 mentioned, we should add a test for this change and verify the dynamic nature of the configuration. It would also be useful to test any validation logic if the settings need to be validated. There's an example of security configuration integ tests that @lukasz-soszynski-eliatra has contributed to this repo. See https://github.com/opensearch-project/security/blob/main/src/integrationTest/java/org/opensearch/security/SecurityConfigurationTests.java#L201-L214 for an example test that updates configuration dynamically. To run integ tests locally you can use `./gradlew integrationTest --tests SecurityConfigurationTests.shouldUseSecurityAdminTool -i`

MaciejMierzwa:
Hi @cwperks, @scrawfor99, pushed in integration test that realoads cluster extensions configuration. After recent merge from main branch random commits are not recognized as signed-off. I'll probably have to nuke this PR and rebase everything.

scrawfor99:
Hi @MaciejMierzwa, thank you for adding tests. You should be able to rebase following the DCO output steps. This will let you overwrite the DCO issue. I went ahead and manually set this version as signed for you as well.

RyanL1997:
Hi @MaciejMierzwa, thank you for putting these together and adding the test! In generally this looks good to me. I just initiated the rerun of the rebase. In addition, as @cwperks mentioned above, my work of authentication backend will take in these configurations.

scrawfor99:
@MaciejMierzwa  The spotless issue is a result of using the default encoding in the JWTVendor. It can be fixed by swapping to a non-default encoding--specifying UTF-8 etc.

MaciejMierzwa:
Yea, I think this part of the code got to my branch during a merge. Added charset UTF-8

cwperks:
@MaciejMierzwa @scrawfor99 Has a PR open in this repo to fix the issue https://github.com/opensearch-project/security/pull/2715. We need to wait for the PR to be merged and then you can merge the latest changes from main into this branch to re-run the CI or one of the maintainers can also synchronize the branches assuming there are no merge conflicts.

src/integrationTest/java/org/opensearch/security/SecurityConfigurationTests.java
```diff
@@ -209,7 +210,28 @@ public void shouldUseSecurityAdminTool() throws Exception {
...
+					.until(() -> client.get("_plugins/_security/api/rolesmapping/readall").getStatusCode(), equalTo(200));
+		}
+	}
+
+	@Test
+	public void shouldReloadExtensionsConfigurationFromFile() throws Exception {
+		SecurityAdminLauncher securityAdminLauncher = new SecurityAdminLauncher(cluster.getHttpPort(), cluster.getTestCertificates());
+		File config = configurationDirectory.newFile("config.yml");
+		ConfigurationFiles.createConfigFile(config);
+		int exitCode = securityAdminLauncher.updateConfig(config);
```
cwperks, line 222: Nice! This is great that you were able to show the dynamic config settings being update. 🚀

cwperks (approved):

cwperks:
Merging this into the feature branch for support for extension REST handlers.

Issue #2619, opened 2023-03-31
[Extension] Authentication Backend for JIT Token Validation


RyanL1997:
### Problem:
To accomplish the objective of enabling extensions to act on behalf of a user, we have devised a just-in-time (JIT) Token to manage incoming request authentication. This JIT Token will be transmitted to the extension and subsequently returned to the security plugin for privilege validation concerning both the user and the extension. Consequently, it is necessary to develop an authentication backend to efficiently handle these validation processes.

### Design
Here is a graph of how user requests processed through the security-extension feature _[graph1]_:
![Extensions for security](https://user-images.githubusercontent.com/109499885/225071058-492bff0e-93fa-4ff6-8229-e85ab6845c95.jpg)
If we zoom up to step 4. & 5 in _[graph1]_ then we get _[graph2]_ as following:
![OpenSearch_JIT_Extension drawio](https://user-images.githubusercontent.com/109499885/229668480-cc551278-6e17-4dbd-b937-8ef6f09f472d.png)
This graph shows that during the creation of the JIT Token, the user's roles will be encrypted (see: #2620 ) and combined with other claims. Subsequently, this JIT Token will be sent to the extension for further processing. 

Once the extension sends back this token to security plugin, which is step 6 in _[graph1]_ then we get _[graph3]_ as following:
![OpenSearch_JIT_Extension drawio (1)](https://user-images.githubusercontent.com/109499885/229669028-a30d5d1c-8c4b-4da4-bd58-3d6edb2ec9fd.png)
Once the security plugin receives the JIT Token back, it will be forwarded into the authentication backend. The following tasks will be initiated: 
* Decrypt roles / backend_roles fields inside of the JIT Token
* Two validations will happen: 
    * Validation of extensions privilege - Service Gate (1st gate and we process to the next gate after the validation)
        * Similar to user roles this field is called governing policies, and the major difference is that you cannot map those policies to users. We need to do is to compare these policies with the specific actions from the request.
        * If the Service Gate validation fails, it will directly abort the target action, and it will not process to the User Gate.
    * Validation of user privilege - User Gate (2nd gate)
        * Decrypt `roles` / `backend_role`
        * Validate the privileges by using the roles concepts (like current role concept)

### Appendix
[WIP] I will put link this section to some of the issues with more concepts of `roles.yml`.

RyanL1997:
Hi @cwperks, I have transferred the doc into this issue. Feel free to leave some thoughts.

[...]

PR #2672, opened 2023-04-11
[Security/Extension] Extension Authentication Backend

RyanL1997:
### Description
Extension Authentication Backend

Some additional modification in JWTVendor:
- Add the support of BWCMode
~~- _BWCMode On_: plain text roles will be passed into claim. It also accepts nonString configuration, but it will be converted into `string` as the payload. The roles information will be stored in the payload field `dec_r` (standing for decrypted roles).~~
~~- _BMCMode Off_: encrypted roles will be passed into claim, and it will always be a encrypted `string`. The roles information will be stored in the payload field `enc_r` (standing for encrypted roles).~~

### Category
New feature

### Issues Resolved
* Resolve: #2619

### Check List


RyanL1997:
Testing is WIP

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,274 @@
...
+        this.signingKey = signingKey;
+        this.encryptionKey = encryptionKey;
+        init();
+    }
+
+    private void init() {
+        JwtParser _jwtParser = null;
+
+        try {
+            if(signingKey == null || signingKey.length() == 0) {
```
scrawfor99, line 84: Should this throw an error instead of just logging?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -826,6 +827,8 @@ public Collection<Object> createComponents(Client localClient, ClusterService cl
 
         securityRestHandler = new SecurityRestFilter(backendRegistry, auditLog, threadPool,
                 principalExtractor, settings, configPath, compatConfig);
+        //TODO: CREATE A INSTANCE OF HTTPExtensionAuthenticationBackend
+        HTTPExtensionJwtAuthenticator acInstance = new HTTPExtensionJwtAuthenticator();
```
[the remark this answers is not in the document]
RyanL1997: AC is standing for AuthCredentials. Nice to point it out. I will also including that in the terminology list I'm creating.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,274 @@
...
+            // try to get roles from claims, first as Object to avoid having to catch the ExpectedTypeException
+            Object subjectObject = claims.get(subjectKey, Object.class);
+            if(subjectObject == null) {
+                log.warn("Failed to get subject from JWT claims, check if subject_key '{}' is correct.", subjectKey);
+                return null;
+            }
+            // We expect a String. If we find something else, convert to String but issue a warning
+            if(!(subjectObject instanceof String)) {
+                log.warn("Expected type String for roles in the JWT for subject_key {}, but value was '{}' ({}). Will convert this value to String.", subjectKey, subjectObject, subjectObject.getClass());
+            }
```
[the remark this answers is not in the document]
RyanL1997: In the original `HTTPJwtAuthenticator.java` it does the same thing. I believe the reason to do this is that we also need to handle the case for some nonString claims. In the test case, we also have a test case called `testNonStringClaim`. I think this scenario only happens when the user enable the BWCMode, which means instead of passing encoded roles, plain text roles will be passed into the claim. For the disabling of BMCMode, the role will always be encrypted, so that it will always be an encrypted `string`.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
RyanL1997: Yes this will directly skip the `else` and throw another error log L110 with an runtime exception at [L112](https://github.com/opensearch-project/security/pull/2672/files#diff-efd217975f23af4bb84808e2771bfc85211f12b59bb619da430a3363902207ceR112). The way we want to do it in this way is because we would like to log some specific information for user to know the reason of failure creation of this token. There are also some `try`/`catch` statement in the `else` for checking encryption algorithm, and each of them will work in the same pattern. It will tell the user why they failed first by these error log and give them the runtime exception error at L112.

RyanL1997:
I'm working on the fix of just 1 test case: `testRolesArray` and some code hygiene issue. Other than that this PR is ready for review.

RyanL1997:
The CI will be fixed after the merge of https://github.com/opensearch-project/security/pull/2715

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,282 @@
...
+        return creds;
+    }
+
+    private AuthCredentials extractCredentials0(final RestRequest request) {
+        if (jwtParser == null) {
+            log.error("Missing Signing Key. JWT authentication will not work");
+            return null;
+        }
+
+        String jwtToken = request.header(HttpHeaders.AUTHORIZATION);
```
cwperks, line 151: Does this mean that this takes anything passed in the `Authorization` header as long as it doesn't contain the word `Basic` anywhere? In the next block if it contains the word `bearer` it will strip that out and trim the rest?

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,282 @@
...
+            }
+
+            if (rolesObject == null) {
+                log.warn(
+                        "Failed to get roles from JWT claims. Check if this key is correct and available in the JWT payload.");
+                roles = new String[0];
+            } else {
+                final String rolesClaim = rolesObject.toString();
+                //IF THE USER IS IN BWC MODE WE ARE EXPECTING UNENCRYPTED ROLES. IF THE BWC MODE IS OFF WE ARE EXPECTING ENCRYPTED ROLES
+                if (rolesClaim != null && !BWCMode) {
```
cwperks, line 196: The null check isn't needed here or in the else if, this will never be null.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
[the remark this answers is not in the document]
cwperks: The message here shouldn't contain the word `roles`. Can you update the message to `Expected type String for subject...`

This should be updated in the `HTTPJwtAuthenticator` as well. There is overlap between here and the `HTTPJwtAuthenticator` class. Maybe we can create a utility class with static methods or maybe add these to an abstract class that both inherit from?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
cwperks, line 831: Should this be behind the extensions feature flag?

cwperks (commented):
Hey @RyanL1997 , I took a first pass at this and like the progress and tests that have been added here. Do you have any thoughts about configuration of this backend? Is this a special backend that if extensions is enabled then is evaluated before the list of auth domains in the `authc` section of the security plugin `config.yml` file? Have you been able to install the security plugin with this change into an opensearch node and show the change in use?

scrawfor99 (commented):
Hi @RyanL1997, things look very good here. I would be happy to approve if you are able to address Craig's comments. I think he covered everything.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,282 @@
...
+            final String subject = extractSubject(claims, request);
+
+            final String audience = claims.getAudience();
+
+            //TODO: GET ROLESCLAIM DEPENDING ON THE STATUS OF BWC MODE. ON: enc_r / OFF: dec_r
+            Object rolesObject = null;
+            String[] roles = null;
+
+            if (BWCMode) {
+                rolesObject = claims.get("dec_r");
```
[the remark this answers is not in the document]
RyanL1997: Yes, I think this is an good idea. Updating it now.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,282 @@
...
+    private static final String BEARER = "bearer ";
+
+    //TODO: TO SEE IF WE NEED THE FINAL FOR FOLLOWING
+    private JwtParser jwtParser;
+    private String subjectKey;
+
+    private String signingKey;
+    private String encryptionKey;
+
+    private Boolean BWCMode;
```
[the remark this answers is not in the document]
RyanL1997: Yes, I will update this~

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,282 @@
...
+                }
+
+            }
+        } catch (Throwable e) {
+            log.error("Error creating JWT authenticator. JWT authentication will not work", e);
+            throw new RuntimeException(e);
+        }
+
+        subjectKey = "sub";
+        //TODO: requireIssuer should be the cluster name. There is one authentication backend for all extensions, so there will be multiple audiences.
```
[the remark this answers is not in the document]
RyanL1997: Done~

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,274 @@
...
+
+    // FOR TESTING
+    public HTTPExtensionJwtAuthenticator(String signingKey, String encryptionKey){
+        this.signingKey = signingKey;
+        this.encryptionKey = encryptionKey;
+        init();
+    }
+
+    private void init() {
+        JwtParser _jwtParser = null;
```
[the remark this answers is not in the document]
RyanL1997: Good catch. That was a redundancy. I have removed this.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,274 @@
...
+
+        if (jwtToken == null || jwtToken.length() == 0) {
+            if(log.isDebugEnabled()) {
+                log.debug("No JWT token found in '{}' header", HttpHeaders.AUTHORIZATION);
+            }
+            return null;
+        }
+
+        final int index;
+        if((index = jwtToken.toLowerCase().indexOf(BEARER)) > -1) { //detect Bearer
```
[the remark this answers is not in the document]
RyanL1997: Yes, I think it is also work with the `.match()`. I was following the same expression in `HTTPJwtAuthenticator.java` ([L170](https://github.com/opensearch-project/security/blob/6997f97f950683c1125ac5b0bb797c40fe8877e8/src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java#L170)). So, maybe we should just leave like this.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package com.amazon.dlic.auth.http.jwt;
```
peternied, line 12: What are your thoughts on this package, com.amazon doesn't make much sense to me.  I'd suggest a move this under a `org.opensearch.security.auth` package or child package.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+import org.opensearch.security.auth.HTTPAuthenticator;
+import org.opensearch.security.authtoken.jwt.EncryptionDecryptionUtil;
+import org.opensearch.security.securityconf.DynamicConfigModel;
+import org.opensearch.security.user.AuthCredentials;
+
+public class HTTPExtensionJwtAuthenticator implements HTTPAuthenticator {
+
+    protected final Logger log = LogManager.getLogger(this.getClass());
+
+    private static final Pattern BASIC = Pattern.compile("^\\s*Basic\\s.*", Pattern.CASE_INSENSITIVE);
```
peternied, line 53: Why does this authenticating backend check for BASIC auth headers, shouldn't it only support bearer auth?

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+    public HTTPExtensionJwtAuthenticator() {
+        super();
+        init();
+    }
+
+    // FOR TESTING
+    public HTTPExtensionJwtAuthenticator(String signingKey, String encryptionKey,Boolean bwcPluginCompatibilityMode){
+        this.signingKey = signingKey;
+        this.encryptionKey = encryptionKey;
+        this.bwcPluginCompatibilityMode = bwcPluginCompatibilityMode;
```
peternied, line 73: If any extensions needs the BWC mode all extensions will get roles information? What do you think about making this determination when the token is generated instead?

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+        final SecurityManager sm = System.getSecurityManager();
+
+        if (sm != null) {
+            sm.checkPermission(new SpecialPermission());
+        }
+
+        AuthCredentials creds = AccessController.doPrivileged(new PrivilegedAction<AuthCredentials>() {
+            @Override
+            public AuthCredentials run() {
+                return extractCredentials0(request);
```
peternied, line 129: Is their a specific action that needs the permission?  The InternalAuthenticationBackend doesn't seem to need this PrivilegedAction, what is this class doing?

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+            }
+        });
+
+        return creds;
+    }
+
+    private AuthCredentials extractCredentials0(final RestRequest request) {
```
peternied, line 136: Take a look at these classes from @scrawfor99 's PR, they do a good job isolating the specific functionality from the overall context.  While you might not need extra classes here, I think the parsing extracting logic would be cleaner

- [RestTokenExtractor](https://github.com/opensearch-project/OpenSearch/blob/e2269cbd40eeb44a9062991090387ca543639fef/server/src/main/java/org/opensearch/identity/tokens/RestTokenExtractor.java)
- [BasicAuthToken](https://github.com/opensearch-project/OpenSearch/blob/e2269cbd40eeb44a9062991090387ca543639fef/server/src/main/java/org/opensearch/identity/tokens/BasicAuthToken.java)

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+            return null;
+        }
+    }
+
+    @Override
+    public boolean reRequestAuthentication(final RestChannel channel, AuthCredentials creds) {
+        final BytesRestResponse wwwAuthenticateResponse = new BytesRestResponse(RestStatus.UNAUTHORIZED,"");
+        wwwAuthenticateResponse.addHeader("WWW-Authenticate", "Bearer realm=\"OpenSearch Security\"");
+        channel.sendResponse(wwwAuthenticateResponse);
+        return true;
```
peternied, line 229: Instead lets return false like in the different proxy authenticators

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+import org.opensearch.rest.BytesRestResponse;
+import org.opensearch.rest.RestChannel;
+import org.opensearch.rest.RestRequest;
+import org.opensearch.rest.RestStatus;
+import org.opensearch.security.auth.HTTPAuthenticator;
+import org.opensearch.security.authtoken.jwt.EncryptionDecryptionUtil;
+import org.opensearch.security.securityconf.DynamicConfigModel;
+import org.opensearch.security.user.AuthCredentials;
+
+public class HTTPExtensionJwtAuthenticator implements HTTPAuthenticator {
```
peternied, line 49: While extensions will use this functionality.  This class is a delegate authenticating backend, lets remove the name extension and reframe to its purpose, what do you think of this?

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+                }
+            }
+
+            if (subject == null) {
+                log.error("No subject found in JWT token");
+                return null;
+            }
+
+            if (audience == null) {
+                log.error("No audience found in JWT token");
```
DarshitChanpura, line 202: is `audience` not mandatory here? If not, why?

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+    public boolean reRequestAuthentication(final RestChannel channel, AuthCredentials creds) {
+        final BytesRestResponse wwwAuthenticateResponse = new BytesRestResponse(RestStatus.UNAUTHORIZED,"");
+        wwwAuthenticateResponse.addHeader("WWW-Authenticate", "Bearer realm=\"OpenSearch Security\"");
+        channel.sendResponse(wwwAuthenticateResponse);
+        return true;
+    }
+
+    @Override
+    public String getType() {
+        return "extension_jwt";
```
DarshitChanpura, line 234: should this hard-coded value be part of Constants file?

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
RyanL1997: Fixed. That makes sense to me too. I was doing this because I was following the class of `HTTPJwtAuthenticator`.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
RyanL1997: That was the question I raised up during one of our stand-up meetings. I assume that the `HTTPJwtAuthenticator` were doing this could be due to leniency in handling to support non-standard header formats that do not include the “Bearer” keyword explicitly. However, for our case, we are require the keyword "Bearer" for our auth header. I will fix this in the next commit.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+                try {
+                    key = getPublicKey(decoded, "EC");
+                } catch (Exception e) {
+                    log.debug("No public ECDSA key, try other algos ({})", e.toString());
+                }
+
+                if(key != null) {
+                    jwtParser = Jwts.parser().setSigningKey(key);
+                } else {
+                    jwtParser = Jwts.parser().setSigningKey(decoded);
```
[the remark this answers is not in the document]
RyanL1997: Correct, that `getPublicKey()` cannot handle the case for Base64 encoding.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -166,9 +166,10 @@ public String createJwt(String issuer, String subject, String audience, Integer
             throw new Exception("The expiration time should be a positive integer");
         }
 
+        //TODO: IF USER ENABLES THE BWC MODE, WE ARE EXPECTING TO SET PLAIN TEXT ROLE AS `dec_r`
         if (roles != null) {
             String listOfRoles = String.join(",", roles);
-            jwtClaims.setProperty("roles", EncryptionDecryptionUtil.encrypt(claimsEncryptionKey, listOfRoles));
+            jwtClaims.setProperty("er", EncryptionDecryptionUtil.encrypt(claimsEncryptionKey, listOfRoles));
```
scrawfor99, line 172: I am not sure what "er" means. I know there are some conflicting opinions, but I am of the mind that we lose meaning at a certain point of abbreviating things. I am guessing this stands for "____ roles"? Perhaps we can make this a more meaningful key?

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+        if (jwtToken != null && BASIC.matcher(jwtToken).matches()) {
+            jwtToken = null;
+        }
+
+        if (jwtToken == null || jwtToken.length() == 0) {
+            if(log.isDebugEnabled()) {
+                log.debug("No JWT token found in '{}' header", HttpHeaders.AUTHORIZATION);
+            }
+            return null;
+        }
```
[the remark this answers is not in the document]
RyanL1997: Thanks for the suggestion! Already fixed

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+                final String rolesClaim = rolesObject.toString();
+                //IF THE USER IS IN BWC MODE WE ARE EXPECTING UNENCRYPTED ROLES. IF THE BWC MODE IS OFF WE ARE EXPECTING ENCRYPTED ROLES
+                if (!bwcPluginCompatibilityMode) {
+                    //TODO: WHERE TO GET THE ENCRYTION KEY
+                    final String decryptedRoles = EncryptionDecryptionUtil.decrypt(encryptionKey, rolesClaim);
+                    roles = Arrays.stream(decryptedRoles.split(",")).map(String::trim).toArray(String[]::new);
+                } else if (bwcPluginCompatibilityMode){
+                    roles = Arrays.stream(rolesClaim.split(",")).map(String::trim).toArray(String[]::new);
+                }
+            }
```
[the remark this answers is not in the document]
RyanL1997: Thanks for the suggestions, fixed~

scrawfor99 (approved):
Looks good Ryan! Great job :)

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
DarshitChanpura: i think `er` stands for `encrypted roles`.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
RyanL1997: I think we can leave like this. I was following the pattern of [HTTPJwtAuthenticator](https://github.com/opensearch-project/security/blob/main/src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java#L219) and [HTTPBasicAuthenticator](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/http/HTTPBasicAuthenticator.java#L77).

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
RyanL1997: Fixed!

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
RyanL1997: @peternied and @DarshitChanpura , Thanks for the information and suggestions. Yes I think this is a good idea. We can do that as the follow-up PR for linking this backend with the configuration setup.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+
+                if(key != null) {
+                    jwtParser = Jwts.parser().setSigningKey(key);
+                } else {
+                    jwtParser = Jwts.parser().setSigningKey(decoded);
+                }
+
+            }
+        } catch (Throwable e) {
+            log.error("Error creating JWT authenticator. JWT authentication will not work", e);
```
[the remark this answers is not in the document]
RyanL1997: Good suggestion! Fixed.

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPExtensionJwtAuthenticator.java
```diff
@@ -0,0 +1,272 @@
...
+
+            final String audience = claims.getAudience();
+
+            //TODO: GET ROLESCLAIM DEPENDING ON THE STATUS OF BWC MODE. ON: er / OFF: dr
+            Object rolesObject = null;
+            String[] roles = null;
+
+            if (bwcPluginCompatibilityMode) {
+                rolesObject = claims.get("dr");
+            } else if (!bwcPluginCompatibilityMode) {
```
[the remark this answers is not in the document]
RyanL1997: Good catch! I switched that to a `else`.

src/main/java/org/opensearch/security/http/HTTPOnBehalfOfJwtAuthenticator.java
```diff
@@ -0,0 +1,277 @@
...
+
+            if (rolesObject == null) {
+                log.warn(
+                        "Failed to get roles from JWT claims. Check if this key is correct and available in the JWT payload.");
+                roles = new String[0];
+            } else {
+                final String rolesClaim = rolesObject.toString();
+
+                // Extracting roles based on the compatbility mode
+                String rolesPostEncryptionDecryption = rolesClaim;
```
cwperks, line 193: nit: Can we change this variable name to `decryptedRoles`?

cwperks (commented):
Thank you @RyanL1997! I left 2 comments that are pretty minor. Once addressed I will approve this PR and we can merge it into the feature branch.

src/main/java/org/opensearch/security/http/HTTPOnBehalfOfJwtAuthenticator.java
RyanL1997: Sure, I have fixed.

src/main/java/org/opensearch/security/http/HTTPOnBehalfOfJwtAuthenticator.java
```diff
@@ -0,0 +1,277 @@
...
+            if(log.isDebugEnabled()) {
+                log.debug("Invalid or expired JWT token.", e);
+            }
+            return null;
+        }
+    }
+
+    @Override
+    public boolean reRequestAuthentication(final RestChannel channel, AuthCredentials creds) {
+        final BytesRestResponse wwwAuthenticateResponse = new BytesRestResponse(RestStatus.UNAUTHORIZED,"");
```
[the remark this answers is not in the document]
RyanL1997: Fixed

cwperks (approved):

DarshitChanpura (approved):
Nicely done!

RyanL1997:
Merging, and the CI is failing due to the new introduced term check for 'extension' at previous PR: Extensions config for JWT signing/encryption key (#2671). I have created a follow-up issue at: https://github.com/opensearch-project/security/issues/2760

Issue #1858, opened 2022-05-20
[FEATURE] usage of JWKS with JWT (w/o OpenID connect)


rursprung:
**Is your feature request related to a problem?**
when using JWT it is cumbersome to manage the public keys / certificates directly in the config file (or as environment variables), making adding/removing issuers and general key rotation a pain.

**What solution would you like?**
it should be possible to define a JWKS endpoint for JWT-based authentication.

note: it _might_ be that this already works (because JWT is used within OpenID connect), but it's definitely undocumented and so i don't know how we could set this up. so at the very least this would require a documentation update, worst-case it might need some enhancements.

**What alternatives have you considered?**
see above:
* store the keys in the config file
* store the keys in environment variables which are referenced in config files

both of these also require one auth domain per issuer, which is suboptimal.

**Do you have any additional context?**
* JWK & JWKS specification: [RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517)
* this is already fully supported when using OpenID connect: [OpenID Connect OpenSearch Security Documentation](https://opensearch.org/docs/latest/security-plugin/configuration/openid-connect/)

DarshitChanpura:
[Triage] Thank you for filing this feature request.

cwperks:
@rursprung I was able to get a cluster running locally with keycloak with the config found here: https://github.com/cwperks/osd-oidc-example/tree/master

When running through this example I see that the `openid_connect_url` contains a link to the `jwks_uri` in the endpoint's response. See example security config below:

```
authc:
      basic_internal_auth_domain:
        description: "Authenticate via HTTP Basic against internal users database"
        http_enabled: true
        transport_enabled: true
        order: 1
        http_authenticator:
          type: basic
          challenge: false
        authentication_backend:
          type: intern
      openid_auth_domain:
        http_enabled: true
        transport_enabled: true
        order: 1
        http_authenticator:
          type: openid
          challenge: false
          config:
            subject_key: preferred_username
            roles_key: roles
            openid_connect_url: http://localhost:8080/auth/realms/master/.well-known/openid-configuration
        authentication_backend:
          type: noop
```

The `http://localhost:8080/auth/realms/master/.well-known/openid-configuration` returns

```
{
   "issuer":"http://localhost:8080/auth/realms/master",
   "authorization_endpoint":"http://localhost:8080/auth/realms/master/protocol/openid-connect/auth",
[...]

PR #2708, opened 2023-04-19
[FEATURE] usage of JWKS with JWT (w/o OpenID connect)

sebastianmichalski:
### Description
Extending auth with jwks_uri parameter to omit passing openid_connect_url with openid-configuration
* Category: feature
* Why these changes are required?  To add possibility to define a JWKS endpoint for JWT-based authentication
* What is the old behavior before changes and new behavior after changes? Old behavior remains the same, new one omits the part with calling openid-configuration endpoint of IdP

### Issues Resolved
https://github.com/opensearch-project/security/issues/1858

### Check List


src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/KeySetRetriever.java
```diff
@@ -65,6 +66,16 @@ public class KeySetRetriever implements KeySetProvider {
 		}
 	}
 
+    KeySetRetriever(SSLConfig sslConfig, boolean useCacheForOidConnectEndpoint, String jwksUri) {
```
scrawfor99, line 69: Nit: Can you keep the SSLConfig and cache args. in the same spot of the constructor by placing your `jwksUri` field first?

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/KeySetRetriever.java
```diff
@@ -65,6 +66,16 @@ public class KeySetRetriever implements KeySetProvider {
...
+        this.sslConfig = sslConfig;
+        this.jwksUri = jwksUri;
+
+        if (useCacheForOidConnectEndpoint) {
```
scrawfor99, line 73: Is it possible to separate this out here and above to avoid duplicating code?

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java
```diff
@@ -58,6 +58,23 @@ public void basicTest() {
 		Assert.assertEquals(3, creds.getAttributes().size());
 	}
 
+
+    @Test
+    public void jwksUriTest() {
```
scrawfor99, line 63: Can you add a negative/unhappy test scenario as well? Demonstrate the behavior is as expected with an invalid configuration? Thank you.

scrawfor99 (commented):
Hi @sebastianmichalski, thank you for taking the time to put together this PR. I left a couple nit-picky comments and then also would appreciate it if you could expand your test to include an 'unhappy' scenario. Thank you for all your work! Generally things look good :)

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/KeySetRetriever.java
sebastianmichalski: Unfortunately, no. Two constructors/methods with the same name cannot have same parameter types order 😉

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/KeySetRetriever.java
sebastianmichalski: Sure, I will fix this

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java
sebastianmichalski: No problem, I'm on it 👍

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/KeySetRetriever.java
scrawfor99: Doh, haha oops.

nibix:
@cwperks You are mentioning a good point.

Effectively, OpenSearch has two JWT authenticator implementations:

- `jwt` in `HTTPJwtAuthenticator`
- `openid` in `HTTPJwtKeyByOpenIdConnectAuthenticator`

Even though one of these is called `openid`, it is not a full OpenId implementation - it only implements fetching of JWKS keys from the "well known" OIDC config endpoint. Otherwise, it is just a JWT authenticator based on a JWKS file linked in the well known OIDC endpoint.

Thus, the requirement from #1858 can be very easily implemented in `HTTPJwtKeyByOpenIdConnectAuthenticator`, as it already provides 95% of the infrastructure.

The other JWT authenticator,  `HTTPJwtAuthenticator`, however does only support a single key and does not know anything about JWKS. Adding support for JWKS to that class would be actually a complete rewrite of that class.

Thus, for kicking off the conversation, we have implemented the new feature in  `HTTPJwtKeyByOpenIdConnectAuthenticator`. It would now totally make sense to consolidate the two JWK implementations into one. 

However, this needs some product mgmt decisions:

- How should this look like
- To which extent is backwards compat required
- How shall backwards compat be achieved

nibix:
@cwperks Thank you for the list! 

However, I think that the following attributes are already supported by `HTTPJwtKeyByOpenIdConnectAuthenticator` (via `AbstractHTTPJwtAuthenticator`):

- `jwt_url_parameter`
- `jwt_header`
- `roles_key`
- `subject_key`
- `jwt_clock_skew_tolerance_seconds`

The attribute `signing_key` is indeed not supported, but I am not sure whether it would make sense to directly support this, as we get the keys in the case of  `HTTPJwtKeyByOpenIdConnectAuthenticator` from the JWKS file.

This leaves for following attributes which we can add

- `required_audience`
- `required_issuer`

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticator.java
```diff
@@ -32,9 +32,15 @@ protected KeyProvider initKeyProvider(Settings settings, Path configPath) throws
 
 		int refreshRateLimitTimeWindowMs = settings.getAsInt("refresh_rate_limit_time_window_ms", 10000);
 		int refreshRateLimitCount = settings.getAsInt("refresh_rate_limit_count", 10);
-
-		KeySetRetriever keySetRetriever = new KeySetRetriever(settings.get("openid_connect_url"),
-				getSSLConfig(settings, configPath), settings.getAsBoolean("cache_jwks_endpoint", false));
+        var jwksUri = settings.get("jwks_uri");
```
peternied, line 35: Quick drive by comment, looks like many of these files have misaligned whitespace, please update these.

sebastianmichalski: Sure, on it

DarshitChanpura:
> The attribute signing_key is indeed not supported, but I am not sure whether it would make sense to directly support this, as we get the keys in the case of HTTPJwtKeyByOpenIdConnectAuthenticator from the JWKS file.

Hey @nibix, Can you elaborate a little on why would the param `signing_key` not be required?

nibix:
> Hey @nibix, Can you elaborate a little on why would the param `signing_key` not be required?

The signing keys now come from the referenced JWKS.

cwperks (approved):
Thank you @sebastianmichalski . This will need to be accompanied with documentation to instruct users to use OIDC auth backend for simple JWT authentication with a JWKS config instead of using the JWT backend. It would be nice if this could be extended to the JWT backend as the issue describes, but this solution does solve the request from the issue filed.

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
```diff
@@ -55,6 +55,8 @@ public abstract class AbstractHTTPJwtAuthenticator implements HTTPAuthenticator
     private final String jwtUrlParameter;
     private final String subjectKey;
     private final String rolesKey;
+	private final String requiredAudience;
```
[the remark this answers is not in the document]
DarshitChanpura: interesting. If you are using Intelli J you can check your tab and indent settings to see if they match. They both should be 4. https://www.jetbrains.com/help/idea/indentation.html#6e8d6896 ( for vscode: https://stackoverflow.com/questions/34174207/how-to-change-indentation-in-visual-studio-code)

Once you have these settings in sync, it should probably resolve indentation issue.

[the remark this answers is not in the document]
sebastianmichalski: Yes, I know the IDE setting, so we want to force tab indentation everywhere even if spotless does not do the job 👍 I will reformat this file and push it in a minute then.

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
```diff
@@ -43,194 +43,203 @@
 import org.opensearch.security.user.AuthCredentials;
 
 public abstract class AbstractHTTPJwtAuthenticator implements HTTPAuthenticator {
-    private final static Logger log = LogManager.getLogger(AbstractHTTPJwtAuthenticator.class);
+	private final static Logger log = LogManager.getLogger(AbstractHTTPJwtAuthenticator.class);
```
DarshitChanpura, line 46: nit: seems like the indentation issue is still present.

DarshitChanpura (commented):
LGTM overall. Left some nit comments.

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
```diff
@@ -43,194 +43,203 @@
...
+			this.keyProvider = this.initKeyProvider(settings, configPath);
+			jwtVerifier = new JwtVerifier(keyProvider, clockSkewToleranceSeconds);
+
+		} catch (Exception e) {
+			log.error("Error creating JWT authenticator. JWT authentication will not work", e);
+			throw new RuntimeException(e);
+		}
+	}
+
+	@Override @SuppressWarnings("removal") public AuthCredentials extractCredentials(RestRequest request, ThreadContext context)
```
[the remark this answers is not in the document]
sebastianmichalski: So if formatting should not be changed how should I treat the class then? :) You said I should set Tab in IntelliJ. 

My question is then still concern with spotless, which added tabs for new lines while not touching old code at all, which leads to indentation problems: https://github.com/opensearch-project/security/pull/2708#discussion_r1182357726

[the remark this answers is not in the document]
sebastianmichalski: In other words, which way should we go?

1. keeping indentations as it was in file and new added lines should have tabs according to spotless (but those lines will be misaligned with old-formatted lines)
2. forcing tab indentation in files we are changing according to spotless, even when file was formatted with 4 spaces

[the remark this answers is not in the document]
sebastianmichalski: Not necessary, I just reverted one commit and kept indentation in this one file as You wished in Your PR, but keep in mind for the future, that some files have four spaces some others tabs as indents, and it was before I started working on this issue 😉

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/KeySetRetriever.java
```diff
@@ -204,6 +213,13 @@ private CloseableHttpClient createHttpClient(HttpCacheStorage httpCacheStorage)
 		return builder.build();
 	}
 
+	private void configureCache(boolean useCacheForOidConnectEndpoint) {
+		if (useCacheForOidConnectEndpoint) {
+			cacheConfig = CacheConfig.custom().setMaxCacheEntries(10).setMaxObjectSize(1024L * 1024L).build();
```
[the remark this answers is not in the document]
sebastianmichalski: Good question. I just extracted method to not repeat myself in both constructors.

DarshitChanpura (dismissed):
LGTM. The comment about extracting the hard-coded values is not a blocker from merging.

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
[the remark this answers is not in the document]
DarshitChanpura: hmm.. noted. Thank you!

cwperks:
@sebastianmichalski Quick question: Is it possible to add an integration test to show required audience and required issuer working with OIDC backend?

sebastianmichalski:
> @sebastianmichalski Quick question: Is it possible to add an integration test to show required audience and required issuer working with OIDC backend?

Is this test not sufficient? https://github.com/opensearch-project/security/pull/2708/files#diff-7495fb84f47dfc4449f13b5a9eb8cc3ea42e5b5a58529ef133e373947918bb2fR64

cwperks:
@sebastianmichalski How about a negative test? The backend is configured with required issuer, but the token does not contain the claim? Same for required audience? Let me know what you think.

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticatorTest.java
```diff
@@ -58,6 +59,86 @@ public void basicTest() {
...
+
+		Assert.assertNotNull(creds);
+		Assert.assertEquals(TestJwts.MCCOY_SUBJECT, creds.getUsername());
+		Assert.assertEquals(TestJwts.TEST_AUDIENCE, creds.getAttributes().get("attr.jwt.aud"));
+		Assert.assertEquals(0, creds.getBackendRoles().size());
+		Assert.assertEquals(4, creds.getAttributes().size());
+		Assert.assertFalse(creds.getAttributes().get("attr.jwt.iss").contains(jwtAuth.getRequiredIssuer()));
+	}
+
+	@Test
```
cwperks, line 108: I'm trying to understand, based on the equivalent test in [HTTPJwtAuthenticatorTest](https://github.com/opensearch-project/security/blob/main/src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java#L356-L364) it looks like this should be returning `null` creds because if the audience claim in the token is different from the configured `requiredAudience` then it would not be considered a valid token. Since the creds are not null here, does that mean it would successfully authenticate?

sebastianmichalski: The reason that the one You found is providing null instead of `AuthCredentials` is that we are using different `extractCredentials0() `method (from `HTTPJwtAuthenticator` instead of `AbstractHTTPJwtAuthenticator`). The one more verbose is trying to parse claims and checks ie. audience, while the other one in the same time just takes claims from token without any checks there (or at least the checks agains audience/issuer):

`jwtParser.parseClaimsJws(jwtToken)` throws Exception which leads to `return null;` in line 205:
image

in the meantime `jwt.getClaims();` is just returning the claims which are just validated only against signature, expiry and nbf.

sebastianmichalski: The topic of this issue was to just provide jwks_uri to use it instead of changing signing key manually I guess, not to change current plugin behavior. Should we change this behavior within this issue?

nibix: @cwperks I have to apologize for the irritation here. Of course, having `required_audience` and `required_issuer` as config attributes needs the accompanying logic. We will add this logic ASAP.

sebastianmichalski: Sure, now it's doable when I know what to do with those params 😉

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/JwtVerifier.java
```diff
@@ -112,6 +116,20 @@ private void validateClaims(JwtToken jwt) throws BadCredentialsException, JwtExc
...
+			validateRequiredAudienceAndIssuer(claims);
+		}
+	}
+
+	private void validateRequiredAudienceAndIssuer(JwtClaims claims) {
+		String audience = claims.getAudience();
+		String issuer = claims.getIssuer();
+
+		if (!audience.equals(requiredAudience)) {
+			throw new JwtException("Invalid issuer");
```
cwperks, line 128: Can this exception text be updated to read `Invalid audience`?

Issue #2667, opened 2023-04-10
[Extensions] Parse configuration file of extensions during installation, pass this information to the Security Plugin


scrawfor99:
After merging #2594, we will need to have a way for core to parse the configuration files of extensions during installation and then pass that information to the Security plugin. Once in the Security plugin, a service account can be created based on the information provided in the extension's configuration files. 

Completion of the issue will look like a PR which adds functionality for parsing a configuration file in core during bootstrap and then passing that information from core into the Security plugin. The base PR should include tests which demonstrate the successful parsing of a configuration file and creation of a payload which can be sent to the Security plugin. 

A follow-up PR will need to be created after this issue is completed which adds the automatic handling of the created payload and the subsequent creation of the service account for the extension.

scrawfor99:
Hi @cwperks, I think that between that and the work that was already done in the ExtensionsManager class. This issue should be all set. This code gets most of the info we would realistically need:

```
 private ExtensionsSettings readFromExtensionsYml(Path filePath) throws IOException {
        Yaml yaml = new Yaml();
        try (InputStream inputStream = Files.newInputStream(filePath)) {
            Map<String, Object> obj = yaml.load(inputStream);
            if (obj == null) {
                inputStream.close();
                throw new IOException("extensions.yml is empty");
            }
            List<HashMap<String, ?>> unreadExtensions = new ArrayList<>((Collection<HashMap<String, ?>>) obj.get("extensions"));
            List<Extension> readExtensions = new ArrayList<Extension>();
            for (HashMap<String, ?> extensionMap : unreadExtensions) {
                try {
                    // checking to see whether any required fields are missing from extension.yml file or not
                    String[] requiredFields = {
                        "name",
                        "uniqueId",
                        "hostAddress",
                        "port",
                        "version",
                        "opensearchVersion",
                        "minimumCompatibleVersion" };
                    List<String> missingFields = Arrays.stream(requiredFields)
                        .filter(field -> !extensionMap.containsKey(field))
                        .collect(Collectors.toList());
                    if (!missingFields.isEmpty()) {
                        throw new IOException("Extension is missing these required fields : " + missingFields);
                    }
```

PR #2713, opened 2023-04-21
[Extensions] Add a internal user provider interface

scrawfor99:
### Description
This PR introduces an internal user provider interface and a security internal user provider which implements this interface. The purpose of this interface is to let us build the plugin-side handler for the extension installation listener in core. Once this PR is merged, we can make changes in core to add the actual interface there that the provider will implement. That will then let us listen to core for extension registration events. 

### Issues Resolved
This PR is 1/2 of the solution to #2667. 

### Testing
Adds isolated tests for the provider functionality. 

### Check List


src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -186,6 +161,55 @@ public void onResponse(IndexResponse response) {
...
+        final boolean userExisted = internalUsersConfiguration.exists(username);
+
+        if (!userExisted) {
+            notFound(channel, "Resource '" + username + "' not found.");
+            return;
+        }
+
+        String authToken = "";
+        try {
+            if (request.uri().contains("/internalusers/" + username + "/authtoken")) {
```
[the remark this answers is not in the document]
scrawfor99: The `InternalUsersApiAction` already includes several endpoints so I thought it was appropriate to add this here. It is just a special instance of the `GET` functionality.

I will add some extra documentation.

src/main/java/org/opensearch/security/user/InternalUserProvider.java
```diff
@@ -0,0 +1,20 @@
...
+import java.io.IOException;
+import org.opensearch.security.securityconf.impl.SecurityDynamicConfiguration;
+
+/**
+ * This interface will be defined in core following these changes. For now this is used for testing in the Security Plugin.
+ *
+ */
+public interface InternalUserProvider {
+
+    public void putInternalUser(String userInfo) throws java.io.IOException;
```
peternied, line 13: I'd recommend against throwing IOException, can we try /catch where this is coming from or if you think its needed throw specific exceptions?

src/main/java/org/opensearch/security/user/InternalUserProvider.java
```diff
@@ -0,0 +1,20 @@
...
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import java.io.IOException;
+import org.opensearch.security.securityconf.impl.SecurityDynamicConfiguration;
+
+/**
+ * This interface will be defined in core following these changes. For now this is used for testing in the Security Plugin.
+ *
+ */
+public interface InternalUserProvider {
```
peternied, line 11: What do you think about making this a 'ServiceAccountManager'?

Then it would have only the use cases needed in OpenSearch core, during extensions registration for find or create with enable / disable?

src/main/java/org/opensearch/security/user/InternalUserProvider.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public void putInternalUser(String userInfo) throws java.io.IOException;
+
+    public SecurityDynamicConfiguration<?> getInternalUser(String userInfo);
```
peternied, line 15: `SecurityDynamicConfiguration` shouldn't be part of the interface, we should return a public interface/class instead

src/main/java/org/opensearch/security/user/InternalUserProvider.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public void removeInternalUser(String userInfo);
+
+    public String getInternalUserAuthToken(String userInfo) throws IOException;
```
peternied, line 19: I don't think this return type should be a string in OpenSearch, it should be some kind of interface around the auth token.  @cwperks Did you have ideas about how we would pass refresh token in the java ecosystem?

peternied (changes requested):
The interface is vague, can we tighten in the use case for extensions with service accounts?

scrawfor99:
Hi @peternied, sorry for the late reply. This is more of a testing area than anything else at the moment. I appreciate you leaving early feedback though. I will certainly address your concerns once I start narrowing down on this work.

Issue #2714, opened 2023-04-21
[Bug] Core changes to ImmutableOpenMap causing various failures 


scrawfor99:
Builds against main (and potentially other versions) are encountering issues related to the `valuesIt()` method used by the `clusterService` when it tries to look at the list of indices. 

```bash
/home/runner/work/security/security/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java:652: error: cannot find symbol
                   return clusterService.state().getMetadata().getIndices().valuesIt();
```
These changes are the result of the Core refactor to no longer use `ImmutableOpenMap`. 

_Originally posted by @willyborankin in https://github.com/opensearch-project/security/issues/2557#issuecomment-1517988828_

scrawfor99:
I expect this change caused the issue: https://github.com/opensearch-project/OpenSearch/commit/1ab22e1cb3d52101b72741447f7d68333b43cc0a

Working on swapping over now.

scrawfor99:
[Triage] @scrawfor99 is working on a fix for this issue at #2715.

DarshitChanpura:
Closing this as related PRs have been merged.

PR #2715, opened 2023-04-21
Resolve ImmutableOpenMap issue from core refactor

scrawfor99:
### Description
Removes use of the ImmutableOpenMap from Security code to resolve issue #2714. 

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/test/framework/matcher/AliasExistsMatcher.java
+++ b/src/integrationTest/java/org/opensearch/test/framework/matcher/AliasExistsMatcher.java
@@ -60,7 +60,7 @@ class AliasExistsMatcher extends TypeSafeDiagnosingMatcher<Client> {
 			}
 			return true;
 		} catch (InterruptedException | ExecutionException e) {
-			mismatchDescription.appendText("Error occured during checking if cluster contains alias ")
+			mismatchDescription.appendText("Error occurred during checking if cluster contains alias ")
 				.appendValue(e);
 			return false;
 		}
--- a/src/integrationTest/java/org/opensearch/test/framework/matcher/AliasExistsMatcher.java
+++ b/src/integrationTest/java/org/opensearch/test/framework/matcher/AliasExistsMatcher.java
@@ -11,6 +11,7 @@ package org.opensearch.test.framework.matcher;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
@@ -23,7 +24,6 @@ import org.opensearch.action.admin.indices.alias.get.GetAliasesRequest;
 import org.opensearch.action.admin.indices.alias.get.GetAliasesResponse;
 import org.opensearch.client.Client;
 import org.opensearch.cluster.metadata.AliasMetadata;
-import org.opensearch.common.collect.ImmutableOpenMap;
 
 import static java.util.Objects.requireNonNull;
 import static java.util.Spliterator.IMMUTABLE;
@@ -41,8 +41,8 @@ class AliasExistsMatcher extends TypeSafeDiagnosingMatcher<Client> {
 	protected boolean matchesSafely(Client client, Description mismatchDescription) {
 		try {
 			GetAliasesResponse response = client.admin().indices().getAliases(new GetAliasesRequest(aliasName)).get();
-			ImmutableOpenMap<String, List<AliasMetadata>> aliases = response.getAliases();
-			Set<String> actualAliasNames = StreamSupport.stream(spliteratorUnknownSize(aliases.valuesIt(), IMMUTABLE), false)
+			Map<String, List<AliasMetadata>> aliases = response.getAliases();
+			Set<String> actualAliasNames = StreamSupport.stream(spliteratorUnknownSize(aliases.values().iterator(), IMMUTABLE), false)
 					.flatMap(Collection::stream)
 					.map(AliasMetadata::getAlias)
 					.collect(Collectors.toSet());
--- a/src/integrationTest/java/org/opensearch/test/framework/matcher/AliasExistsMatcher.java
+++ b/src/integrationTest/java/org/opensearch/test/framework/matcher/AliasExistsMatcher.java
@@ -10,6 +10,7 @@
 package org.opensearch.test.framework.matcher;
 
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -17,6 +18,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
+import com.carrotsearch.hppc.cursors.ObjectObjectCursor;
 import org.hamcrest.Description;
 import org.hamcrest.TypeSafeDiagnosingMatcher;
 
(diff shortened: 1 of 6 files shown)


reta (dismissed):

scrawfor99:
Will have to wait until the next build: https://build.ci.opensearch.org/job/distribution-build-opensearch/7748/ before we can actually confirm this is fixed. Most recent build gets pulled as part of build but that does not include the Core changes.

RyanL1997 (dismissed):

src/integrationTest/java/org/opensearch/test/framework/matcher/AliasExistsMatcher.java
```diff
@@ -41,8 +41,8 @@ public AliasExistsMatcher(String aliasName) {
 	protected boolean matchesSafely(Client client, Description mismatchDescription) {
 		try {
 			GetAliasesResponse response = client.admin().indices().getAliases(new GetAliasesRequest(aliasName)).get();
-			ImmutableOpenMap<String, List<AliasMetadata>> aliases = response.getAliases();
-			Set<String> actualAliasNames = StreamSupport.stream(spliteratorUnknownSize(aliases.valuesIt(), IMMUTABLE), false)
+			Map<String, List<AliasMetadata>> aliases = response.getAliases();
```
RyanL1997, line 44: I think for this `getAliases()`, we also need to fix it, because it is also depending on the stale library of `ImmutableOpenMap`

RyanL1997 (commented):
Hi @scrawfor99 , thanks for this fix. Sorry I had to dismissed my previous approve, because I was trying this fix on my local machine, and I think we need to fix some of the functions. I left a comment down below.

src/integrationTest/java/org/opensearch/test/framework/matcher/AliasExistsMatcher.java
scrawfor99: So I took a look into this and I think it will actually be resolved once the new build is done. For whatever reason, the build with the core change is still not complete: https://build.ci.opensearch.org/job/distribution-build-opensearch/7750/

peternied (approved):

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -649,7 +649,7 @@ private boolean checkFilteredAliases(Resolved requestedResolved, String action,
             indexMetaDataCollection = new Iterable<IndexMetadata>() {
                 @Override
                 public Iterator<IndexMetadata> iterator() {
-                    return clusterService.state().getMetadata().getIndices().valuesIt();
+                    return (Iterator<IndexMetadata>) clusterService.state().getMetadata().getIndices().values();
```
peternied, line 652:
```
2023-04-24T19:23:06.3310228Z   1> java.lang.ClassCastException: class java.util.Collections$UnmodifiableCollection cannot be cast to class java.util.Iterator (java.util.Collections$UnmodifiableCollection and java.util.Iterator are in module java.base of loader 'bootstrap')
2023-04-24T19:23:06.3310965Z   1> 	at org.opensearch.security.privileges.PrivilegesEvaluator$1.iterator(PrivilegesEvaluator.java:652) ~[main/:?]
```

Think you might be missing a `.iterator()` call before attempting to return, can you do this without the unchecked suppression?

You can repro the test failure with `org.opensearch.security.IndexIntegrationTests.testIndexResolveMinus` for a quick turnaround

RyanL1997 (approved):

src/integrationTest/java/org/opensearch/test/framework/matcher/GetSettingsResponseContainsIndicesMatcher.java
```diff
@@ -31,12 +34,16 @@ class GetSettingsResponseContainsIndicesMatcher extends TypeSafeDiagnosingMatche
 
 	@Override
 	protected boolean matchesSafely(GetSettingsResponse response, Description mismatchDescription) {
-		ImmutableOpenMap<String, Settings> indexToSettings = response.getIndexToSettings();
+
+		final Map<String, Settings> indexToSettings = new HashMap<>();
+		for (ObjectObjectCursor<String, Settings> cursor : response.getIndexToSettings()) {
+			indexToSettings.put(cursor.key, cursor.value);
+		}
```
cwperks, line 41: Can this be simplified to:

```suggestion
		final Set<String> indexKeys = new HashSet<>();
		for (ObjectObjectCursor<String, Settings> cursor : response.getIndexToSettings()) {
			indexKeys.add(cursor.key);
		}
```

cwperks: Looks like either would work, but since we are not using the values in the map and only getting a keySet it would take less memory.

cwperks (commented):
Thank you @scrawfor99, the changes look good to me. I left one suggestion so let me know what you think, it wouldn't block this PR from being approved and merged. If the suggestion doesn't work then let me know and I will approve this PR.

peternied (approved):

src/integrationTest/java/org/opensearch/test/framework/matcher/GetSettingsResponseContainsIndicesMatcher.java
scrawfor99: Sorry Craig, I didn't see this before the PR got merged. I had left it as a map because that is what the original implementation was. I agree that a Set should work fine here as well. If you wanted to swap it you could open up a quick PR to change it in main.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -674,14 +674,17 @@ public Iterator<IndexMetadata> iterator() {
 
             final List<AliasMetadata> filteredAliases = new ArrayList<AliasMetadata>();
 
-            final ImmutableOpenMap<String, AliasMetadata> aliases = indexMetaData.getAliases();
+            final Map<String, AliasMetadata> aliases = new HashMap<>();
```
andrross, line 677: Why did you make this change? Doesn't `indexMetaData.getAliases()` still return an ImmutableOpenMap? Seems like you should avoid making a copy of the map in any case.

scrawfor99: Hi @andrross, I made this swap because we could not use `keysIt` anymore. We needed `aliases` to be a normal map to then grab and iterator over the keyset.

andrross: @nknize What's happening here? Pretty sure making a copy of the the map is the exact opposite of your intention with this change.

scrawfor99: This was the issue that arose when the change was made to core: https://github.com/opensearch-project/security/issues/2714. I just made changes to get things Green. Sorry if this was not the intended implementation...

andrross: Sorry @scrawfor99! Definitely not trying to blame you here! I don't understand why `keysIt` doesn't work anymore, and I don't think the intent with @nknize's change is to force downstream dependencies to copy data structures when doing a simple iteration.

nknize: No copies unless you need to make changes to the map (even then the core should return an `unmodifiableMap` so you would only be able to change a clone). 

If you just need an iterator over the keys then use `.keySet().iterator()`

Issue #2668, opened 2023-04-10
[Extensions] Generate a password for an extension's service account on installation


scrawfor99:
After it is possible to read an extension's configuration file and pass that information to the Security plugin, we will need to generate a password for the extension to then use for the creation of its auth header. The password will be used for basic authentication of the service account associated with the extension and should not be stored in the Security plugin. Instead, the password should be immediately hashed and the hash stored as part of the service account details. The password will then be sent back to core where it can be handed to the extension. 

Completion of this issue will look like a PR which generates a password for an extension's service account, hashes it, and then stores the hash and sends the password back to core.

scrawfor99:
[Update 4/11] I am working on this in tandem with the token generation issue [here](https://github.com/scrawfor99/security/tree/connectTokenGenerator). I believe I have found a bearer auth based solution which will avoid requiring passwords for the service accounts. That being said, I am running into some trouble storing the token information for revocation. We don't want to store it in the security index since it is not a configuration file based map. I have not found any other examples in the security plugin which don't use the CType pattern, so working on it.

Issue #2611, opened 2023-03-30
[Extensions] Connect auth token generator to service accounts 


scrawfor99:
**NOTE: #2567 must be merged before this is actionable.**

In order to implement Service Accounts (https://github.com/opensearch-project/security/issues/2597), we need the Security Plugin to be able to vend an authorization token back to core.

When core requests a Service Account for an extension (#2609), the response from the Security Plugin should include an authorization token associated with the Service Account. The authorization token can be made using the token generator introduced in #2567. This should provide the framework for creating a JWT that can be passed back to core and then later verified by the Security Plugin. 

The authorization token should correspond to the Service Account associated with the extensionId that core provides. This PR should be straightforward to implement since the framework for generating a JWT is already introduced. 

This issue will be complete when there is a PR that takes in an `extensionUniqueId`, fetches the corresponding Service Account from the Internal Users Storage, and finally creates a JWT based on this information. The final PR should contain tests that show that a JWT is created from an arbitrary string (representing the ID) and that this token can later be verified.

scrawfor99:
[Triage] This is part of the Extensions project.

scrawfor99:
I am currently working on this ahttps://github.com/scrawfor99/security/tree/connectTokenGenerator. I have everything but the actual storage of the tokens figured out.

PR #2716, opened 2023-04-21
[Extensions] Generate auth tokens for service accounts

scrawfor99:
### Description
This is a PR to introduce the functionality for generating an auth token for an extension. The auth token consists of a an encoded basic auth token created from a randomly generated password. The token is then passed back to the client so that changes can be made in core in order to later perform authentication and authorization using the token. 

### Issues Resolved
This issue addresses This PR covers issue https://github.com/opensearch-project/security/issues/2611 and issue https://github.com/opensearch-project/security/issues/2668.

### Testing
New tests are added to the UserApiTest file.

### Check List


src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -144,8 +148,27 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
...
+                        "Existing user " + username + " has no password, and no new password or hash was specified.");
+                return;
+            }
+            contentAsNode.put("hash", hash);
+        }
+
+        internalUsersConfiguration.remove(username);
+
+        // checks complete, create or update the user
+        internalUsersConfiguration.putCObject(username, DefaultObjectMapper.readTree(contentAsNode,  internalUsersConfiguration.getImplementingClass()));
```
DarshitChanpura, line 166: Do you think it might be worth while to extract `readTree` call to a variable and then use that here for better readability and debugging if needed?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -158,6 +181,55 @@ public void onResponse(IndexResponse response) {
...
+        final boolean userExisted = internalUsersConfiguration.exists(username);
+
+        if (!userExisted) {
+            notFound(channel, "Resource '" + username + "' not found.");
+            return;
+        }
+
+        String authToken = "";
+        try {
+            if (request.uri().contains("/internalusers/" + username + "/authtoken")) {
```
DarshitChanpura, line 208: What do you think about checking `.endsWith("/authtoken")` here?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -158,6 +181,55 @@ public void onResponse(IndexResponse response) {
...
+        }  catch (UserServiceException ex) {
+            badRequestResponse(channel, ex.getMessage());
+            return;
+        }
+        catch (IOException ex) {
+            throw new IOException(ex);
+        }
+
+        if (!authToken.isEmpty()) {
+            createdResponse(channel, "'" + username + "' authtoken updated generated "  + authToken);
```
DarshitChanpura, line 226: nit: can remove `updated` here

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -158,6 +181,55 @@ public void onResponse(IndexResponse response) {
...
+        }
+
+        String authToken = "";
+        try {
+            if (request.uri().contains("/internalusers/" + username + "/authtoken")) {
+
+                authToken = userService.generateAuthToken(username);
+            } else {
+
+                internalUsersConfiguration.removeOthers(username);
```
DarshitChanpura, line 213: Have you tested if there is any change in behaviour for other `GETs`.
E.g. `/internalusers/` API

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -322,9 +323,12 @@ private void verifyPatch(final boolean sendAdminCert, Header... restAdminHeader)
                 HttpStatus.SC_CREATED);
 
         // Add enabled service account then get it
+
+        rh.sendAdminCertificate = sendAdminCert;
```
DarshitChanpura, line 327: Don't think we need this, as we are not changing `sendAdminCert` value at any point below

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
scrawfor99: Yeah, the old tests still run exactly as expected.

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
scrawfor99: Oops nice catch.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
scrawfor99: Will do.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
scrawfor99: Yeah sounds good.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
scrawfor99: Sure, will do.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -144,8 +148,28 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
             throw new IOException(ex);
         }
 
+        // for existing users, hash is optional
+        if (userExisted && securityJsonNode.get("hash").asString() == null) {
```
RyanL1997, line 152: For my understanding, why don't we need hash for existing users?

RyanL1997 (commented):
Hi @scrawfor99 , nice work for the service account! In generally, I think this PR is in a good shape, and I just left some questions for my understanding.

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -409,6 +405,51 @@ private void verifyPatch(final boolean sendAdminCert, Header... restAdminHeader)
...
+        HttpResponse response = rh.executePutRequest(ENDPOINT + "/internalusers/happyServiceLive",
+                ENABLED_SERVICE_ACCOUNT_BODY, restAdminHeader);
+        Assert.assertEquals(response.getBody(), HttpStatus.SC_CREATED, response.getStatusCode());
+        rh.sendAdminCertificate = sendAdminCert;
+        response = rh.executeGetRequest(ENDPOINT + "/internalusers/happyServiceLive", restAdminHeader);
+        Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
+
+        response = rh.executeGetRequest(ENDPOINT + "/internalusers/happyServiceLive/authtoken",
+                ENABLED_SERVICE_ACCOUNT_BODY, restAdminHeader);
+        Assert.assertEquals(response.getBody(), HttpStatus.SC_CREATED, response.getStatusCode());
```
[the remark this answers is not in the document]
scrawfor99: So for asserts, you can provide 3 arguments in which case the first one is what will be output if the assertion fails. It would be the whole body though in this case.

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -409,6 +405,51 @@ private void verifyPatch(final boolean sendAdminCert, Header... restAdminHeader)
...
+        String tokenFromResponse = response.getBody();
+        byte[] decodedResponse = Base64.getUrlDecoder().decode(tokenFromResponse);
+        String[] decodedResponseString = new String(decodedResponse).split(":", 2);
+        String username = decodedResponseString[0];
+        String password = decodedResponseString[1];
+        Assert.assertEquals("Username is: " + username,username, "happyServiceLive");
+
+        // Add disabled service account then try to get its auth token
```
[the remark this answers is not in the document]
scrawfor99: In theory it is all exercising the same type of fake rest request logic. That being said there will be more complicated tests once we have the core code implemented as well. So the full integration tests will have to be its own file.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
scrawfor99: You don't need the hash for existing users since it is already stored in our records. So as long as we have it we are fine with them not providing it.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -50,17 +52,19 @@
...
 public class InternalUsersApiAction extends PatchableResourceApiAction {
     static final List<String> RESTRICTED_FROM_USERNAME = ImmutableList.of(
-        ":" // Not allowed in basic auth, see https://stackoverflow.com/a/33391003/533057
+            ":" // Not allowed in basic auth, see https://stackoverflow.com/a/33391003/533057
     );
 
     private static final List<Route> routes = addRoutesPrefix(ImmutableList.of(
             new Route(Method.GET, "/user/{name}"),
             new Route(Method.GET, "/user/"),
+            new Route(Method.GET, "/user/{name}/authtoken"),
```
peternied, line 61: Shouldn't be a `GET`, I think this should be a `POST` since its modifying the user and it isn't going to generate the same result if called multiple times.  Used https://www.baeldung.com/rest-http-put-vs-post to look into this, I'd be curious what you think.

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -178,4 +202,79 @@ private String generatePassword() {
...
+        final SecurityDynamicConfiguration<?> internalUsersConfiguration = load(getUserConfigName(), false);
+
+        if (!internalUsersConfiguration.exists(accountName)) {
+            throw new UserServiceException(FAILED_ACCOUNT_RETRIEVAL_MESSAGE);
+        }
+
+        String authToken = null;
+        try {
+            ObjectMapper mapper = new ObjectMapper();
+            JsonNode accountDetails = mapper.readTree(internalUsersConfiguration.getCEntry(accountName).toString());
```
peternied, line 224: Can you use DefaultObjectMapper.readTree?

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -178,4 +202,79 @@ private String generatePassword() {
...
+            final ObjectNode contentAsNode = (ObjectNode) accountDetails;
```
peternied, line 225: While it isn't required to an ObjectNode, JsonNode can use `withObject` to create a new object with the modifications [1]; however what you are doing here is fine.

- [1] https://fasterxml.github.io/jackson-databind/javadoc/2.14/com/fasterxml/jackson/databind/JsonNode.html#withObject-com.fasterxml.jackson.core.JsonPointer-

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -178,4 +202,79 @@ private String generatePassword() {
...
+            SecurityJsonNode securityJsonNode = new SecurityJsonNode(contentAsNode);
+
+            if (securityJsonNode.get("isService").isNull()) { // If this is not a service account
+                throw new UserServiceException(AUTH_TOKEN_GENERATION_MESSAGE);
+            }
+            if (Objects.requireNonNull(securityJsonNode.get("isService").asString()).equalsIgnoreCase("false")) { // If this is not a service account
```
peternied, line 231: Very Nitty; `requireNonNull` will throw an exception if the item is null, but the message might not be very useful without what is null if line number don't match up.

IMO I'd recommend using Optional with `.orElseThrow(...);` to prevent null pointers and let you clearly say what you wanted

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -86,7 +87,7 @@ public void testSecurityRoles() throws Exception {
                 .executeGetRequest(ENDPOINT + "/" + CType.INTERNALUSERS.toLCString());
         Assert.assertEquals(response.getBody(), HttpStatus.SC_OK, response.getStatusCode());
         Settings settings = Settings.builder().loadFromSource(response.getBody(), XContentType.JSON).build();
-        Assert.assertEquals(133, settings.size());
+        Assert.assertEquals(171, settings.size());
```
peternied, line 90: I really hate this, consider cleaning this up in a follow up PR.  As it stands these tests just config an arbitrary size instead of the contents.

If there is a math based source to this value make a constant and preform a calcuation that can be followed by a human such as `final int USER_SETTING_SIZE = 9 /* lines per entry*/ * 19 /* number of accounts*/;` so that if 19 -> 20 its should be obvious as part of the code changes why this was done

peternied (dismissed):
Minor comments, feel free to address in a future PR

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
scrawfor99: Yeah that is a good point. I did not think about the distinction because it feels like you are "getting" something. I see that it should be the same result each time though, and this may not be the case.

peternied: The result doesn't have to be the same each time, I think it makes sense for their to be different responses in this case. Typically 'PUT' calls do return a consistent result that is all

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: Yeah will do.

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: Oh good to know, I will take a look into whether it makes more sense to swap over or not. This was the previous pattern before the refactor but that does not mean it is the "correct" pattern.

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: Oh, I did not realize this. I will swap over.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
scrawfor99: Yeah, sorry, I should have been clearer. I meant that I see that `GET` should give the same response each time and since this won't I should use a different request type.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -158,6 +182,64 @@ public void onResponse(IndexResponse response) {
...
+        if (!userExisted) {
+            notFound(channel, "Resource '" + username + "' not found.");
+            return;
+        }
+
+        String authToken = "";
+        try {
+            if (request.uri().contains("/internalusers/" + username + "/authtoken") && request.uri().endsWith("/authtoken")) {  // Handle auth token fetching
+
+                authToken = userService.generateAuthToken(username);
```
DarshitChanpura, line 220: Should we have a transport action setup here to control who can/cannot execute this endpoint. Or should we leave it as is?
If not, we can leave it as is.

scrawfor99: I think for now, we want to leave this as is. It could be subject to change in the future though.

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -178,4 +202,79 @@ private String generatePassword() {
...
+                throw new UserServiceException(AUTH_TOKEN_GENERATION_MESSAGE);
+            }
+            if (securityJsonNode.get("isEnabled").isNull()) { // If there is no enabled flag
+                throw new UserServiceException(AUTH_TOKEN_GENERATION_MESSAGE);
+            }
+            if (Objects.requireNonNull(securityJsonNode.get("isEnabled").asString()).equalsIgnoreCase("false")) { // If the service account is not active
+                throw new UserServiceException(AUTH_TOKEN_GENERATION_MESSAGE);
```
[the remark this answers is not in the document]
scrawfor99: Previously I had gotten feedback that combined ORs were hard to read so I did them as separate cases here.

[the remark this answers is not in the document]
DarshitChanpura: hmm.. interesting choice. My understanding is if is fine upto 3 conditions but should be broken up when >3.
I'll leave it upto you.

build.gradle
```diff
@@ -75,16 +75,16 @@ licenseFile = rootProject.file('LICENSE.txt')
 noticeFile = rootProject.file('NOTICE.txt')
```
peternied, line 75: Please turn off your auto formatter, it makes it harder to review your changes.  If you'd like a format to be followed, lets see about enforcing it via a checkstyle rule for consistency in the codebase in a separate PR

peternied (approved):

build.gradle
scrawfor99: Oops, I did not even know it was on. Sorry.

src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
```diff
@@ -114,9 +128,17 @@ public void setAttributes(Map<String, String> attributes) {
...
+        }
+
+        public boolean getIsService() {
+            return this.isService;
+        }
+
         @Override
         public String toString() {
-            return "InternalUserV7 [hash=" + hash + ", reserved=" + reserved + ", hidden=" + hidden + ", _static=" + _static + ", backend_roles="
+            return "InternalUserV7 [hash=" + hash + ", isEnabled=" + isEnabled + ", isService=" + isService + ", reserved=" + reserved + ", hidden=" + hidden + ", _static=" + _static + ", backend_roles="
```
cwperks, line 141: This is awkward with the mix of casing. Do you think the work `is` is important in these variable names?

src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
```diff
@@ -114,9 +128,17 @@ public void setAttributes(Map<String, String> attributes) {
             this.attributes = attributes;
         }
 
+        public boolean getIsEnabled() {
+            return this.isEnabled;
+        }
+
+        public boolean getIsService() {
```
cwperks, line 135: Shorten this to `isService()` and `isEnabled()`

cwperks (commented):
Thank you @scrawfor99. I left a few comments that are cosmetic around variable naming. This is looking close to me.

src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
scrawfor99: I would probably keep them, since I prefer the boolean names to say the positive case in them. I think `enabled` would be fine but `service=true/false` sounds ambiguous to me.

DarshitChanpura (changes requested):
Request 1 change. Looks good to me otherwise.

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -178,4 +203,77 @@ private String generatePassword() {
...
+            Optional.ofNullable(securityJsonNode.get("isService"))
+                .map(SecurityJsonNode::asString)
+                .filter("true"::equalsIgnoreCase)
+                .orElseThrow(() -> new UserServiceException(AUTH_TOKEN_GENERATION_MESSAGE));
+
+
+            Optional.ofNullable(securityJsonNode.get("isEnabled"))
+                .map(SecurityJsonNode::asString)
+                .filter("true"::equalsIgnoreCase)
+                .orElseThrow(() -> new UserServiceException(AUTH_TOKEN_GENERATION_MESSAGE));
```
DarshitChanpura, line 238: nicely done 👏

DarshitChanpura (dismissed):
LGTM. Thanks @scrawfor99 !

DarshitChanpura (approved):

peternied (approved):

DarshitChanpura (approved):

peternied (approved):

Issue #2553, opened 2023-03-14
Update SecurityIndexAccessEvaluator to support new permission


peternied:

samuelcostae:
@peternied  @scrawfor99  Sorry if the answer is obvious, but I have a question, each Extension will have it's own index? or is it only one index called "system_index"  that will contain info for all extensions? And Piggybacking on that, each extension would have it is own User and Role?

peternied:
Additional background on system indexes configuration / security plugin interaction https://opensearch.org/docs/1.3/security/configuration/system-indexes/

Plugins implement the `SystemIndexPlugin` interface when they can describe N system indexes, [[link]](https://github.com/opensearch-project/OpenSearch/blob/c334bbde881c61e66b7b4768e2d43455f1c827d9/server/src/main/java/org/opensearch/plugins/SystemIndexPlugin.java#L41).  Extensions will do the same (but that is out of scope for this task).

Check out how different plugins use them by searching through the org [[search results]](https://github.com/search?q=org%3Aopensearch-project+SystemIndexPlugin&type=code).

samuelcostae:
Hi!

"The new permissions allows for read and write operations on the permitted index(es)"

So a user mapped to the following role:

```
myExtensionRole:
  description: "Extension Role"
  cluster_permissions:
    - '*'
  index_permissions:
    - index_patterns:
        - '.myExtensionIndex'
      allowed_actions:
        - "system:admin/system_index"
```

Should be able to read/write  without any of the existing  allowed actions ? (e.g.  "indices:admin/get" )

Cause i was thinking, couldn't just block the access if it is trying to access and extension index without the new permission and if it has the new permission, proceed with the remaining existing permission checks?

samuelcostae:
I created a draft PR: https://github.com/opensearch-project/security/pull/2681  

Here are some scenarios:

Given a index called 'system_index_b' configured as a 'system index'
The role below would be able to access system_index_b

```
extensions_role_b:
  description: "Limited ExtensionsRole"
  cluster_permissions:
    - '*'
  index_permissions:
    - index_patterns:
        - 'system_index_b'
      allowed_actions:
        - '*'
        - 'system:admin/system_index'
```

The role below would not be able to access system_index_b
```
extensions_role_c:
  description: "Limited ExtensionsRole"
  cluster_permissions:
    - '*'
  index_permissions:
    - index_patterns:
        - 'system_index_c'
      allowed_actions:
        - '*'
[...]

PR #2750, opened 2023-05-08
Extension registration

### LINKED ISSUE #2595: Allow extensions to dynamically register roles
Plugins can register roles by adding new roles to [roles.yml](https://github.com/opensearch-project/security/blob/main/config/roles.yml). The roles are added to the cluster on bootstrap of a fresh cluster installation, by running securityadmin.sh or by using the API to create the roles. 

Extensions need a way to register roles with a list of actions and have the roles sourced to the opensearch security index on extension installation.

samuelcostae:
### Description
Draft PR to get feedback and questions .
This changes aim to Provide an endpoint where Extension can register themselves in the cluster and list its protected indices that should not be accessed  by any user.

 Category: new feature
  

### Issues Resolved
#2595

### Testing
A unique unit  test added so far to assist devlopment 

### Check List


peternied (changes requested):
It looks like this is duplicating the list of extensions that are specified by the OpenSearch extension yaml file and then loaded into the extension manager.

What is you're thinking about reconcile having two sources of information about extensions?

samuelcostae:
@peternied Thanks a million for the offer, I eventually figured out the put  the request was arriving at the new class, and it was because i hadn't implemented a Validator (Added a dummy one in the meantime to always return true).

I have two questions here regarding the requests.  I'm not sure what is the process to  add a document to a indice in the OS  cluster. In the draft PR i have i used  
`        client.index(new IndexRequest("index_name").source(contentAsNode)); `

Is that the correct approach to add a document in a cluster?

I also think im missing something pretty obvious regarding why the request's body im sending here on ExtensionRegistrationApiActionTest.java 

`        RestHelper.HttpResponse responsea = keyStoreRestHelper.executePutRequest( ENDPOINT, correctExtensionRequest);`

AS the body / content is arriving empty, even though from other tests it seems pretty straight forward.

Thanks

samuelcostae:
> It looks like this is duplicating the list of extensions that are specified by the OpenSearch extension yaml file and then loaded into the extension manager.
> 
> What is you're thinking about reconcile having two sources of information about extensions?



From the discussion on #2553,  I had understood that  extensions would communicate with OS or Security via REST  requests, mainly the posts before/after  this @cwperks 's post:    https://github.com/opensearch-project/security/issues/2553#issuecomment-1526825194 . 

> What is you're thinking about reconcile having two sources of information about extensions?

Having two sources is most of times not good, on this PR I had no intention to add a new source of truth  for the extension data. My understanding is that wherever the extension would have its information stored (either the settings.yaml or whichever external souce to be defined/discussed in the future), this endpoint would  be touched with that info and the relevant info would be  saved to the index.

But I might be missing some context regarding how the best approach to saving info to internally an index.

src/main/java/org/opensearch/security/dlic/rest/api/ExtensionRegistrationApiAction.java
```diff
@@ -0,0 +1,164 @@
...
+            throws IOException{
+
+        createdResponse(channel, " updated");
+
+        ;
+    }
+
+    @Override
+    protected void handlePut(RestChannel channel, final RestRequest request, final Client client, final JsonNode content) throws IOException {
+
```
samuelcostae, line 104: The request body  arrives here empty .

An im missing something obvious?

Issue #2752, opened 2023-05-09
[Feature/Extensions] Supporting `cluster_permissions` nomenclature


DarshitChanpura:
Current security roles model uses `cluster_permissions` to evaluate Transport Layer actions. This is because the plugins run  in-process with OS cluster. 
With Extensions being out-of-process entities, they can request OS cluster only via REST APIs.
And so there is a need to authorize these requests at REST layer.

This issue covers authorizing REST request against the legacy and new `cluster_permissions` model.

E.g: For [hello-world sample extension](https://github.com/opensearch-project/opensearch-sdk-java/tree/main/src/main/java/org/opensearch/sdk/sample/helloworld), the endpoint `/hello` should succeed when either of the following roles are mapped to a user.

```yml
legacy_hw_greet:
  reserved: true
  cluster_permissions:
    - 'cluster:admin/opensearch/hw/greet'


extension_hw_greet:
  reserved: true
  cluster_permissions:
    - 'hw:greet'
```

This enables minimal effort for admin to enable use of extensions.

- [ ] Allow registering legacy transport actionName when registering a NamedRoute

DarshitChanpura:
Done via https://github.com/opensearch-project/security/pull/2753

PR #2753, opened 2023-05-09
Authorize rest requests

DarshitChanpura:
### Description
Continuation of #2601. Adds support for legacy naming conventions
* Category : Enhancement
* Why these changes are required? To support legacy action names to authorize REST requests

### Issues Resolved
https://github.com/opensearch-project/security/issues/2752

### Testing
Manual testing.

1. Setup opensearch-cluster and helloWorld extension following this [developer guide](https://github.com/opensearch-project/opensearch-sdk-java/blob/main/DEVELOPER_GUIDE.md).
2. Check-out this branch.
3. Install security in opensearch.
4. Ensure certificates are setup correctly in config folder and in `opensearch.yml`
5. Following should be the output of each curl command:
```console
➜  ~ curl -XGET https://new-user:admin@localhost:9200/_extensions/_hw/hello --insecure
Hello, World!%
➜  ~ curl -XPUT https://new-user:admin@localhost:9200/_extensions/_hw/hello/HW --insecure -H 'Content-Type: application/json' -d '{}'
Updated the world's name to HW%
➜  ~ curl -XGET https://new-user:admin@localhost:9200/_extensions/_hw/hello --insecure
Hello, HW!%
➜  ~ curl -XPOST https://new-user:admin@localhost:9200/_extensions/_hw/hello --insecure -H 'Content-Type: application/json' -d '{ }'
no permissions for [hw:greet_with_adjective] and User [name=new-user, backend_roles=[], requestedTenant=null]%
➜  ~ curl -XDELETE https://new-user:admin@localhost:9200/_extensions/_hw/goodbye --insecure
no permissions for [hw:goodbye] and User [name=new-user, backend_roles=[], requestedTenant=null]%
```

### Check List


src/main/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluator.java
```diff
@@ -0,0 +1,160 @@
...
+    private boolean impliesLegacyPermission(String action, SecurityRoles roles) {
+        boolean isAlreadyLegacy = action.startsWith("cluster:admin/open");
+        if (isAlreadyLegacy) {
+            return false; // this check was already made, so return false
+        }
+
+        log.info("Checking legacy permissions for {}", action);
+
+        action = action.replace(':', '/');
+        String opendistroAction = "cluster:admin/opendistro/" + action;
```
cwperks, line 155: I don't think we should assume that the name and legacy name of an action are overlapping. 

For AD, as an extension the transport action for get detector info is:

```
cluster:admin/opendistro/ad/detector/info
```

and will not be changed.

In this PR of the SDK (https://github.com/opensearch-project/opensearch-sdk-java/pulls) I introduced a concept of an `extensionShortName` (abbreviation) for an extension. For Anomaly Detection that could be `ad` like is done in this transport action name, but it could just as well be `anomaly` and then the REST API action name could be `anomaly:detector/info`. `detector/info` would be the `name` of the REST action corresponding to the detector info, but that name is entirely up to the extension developer and could change. Could we have a more flexible way of defining a legacy name?

scrawfor99 (approved):
Hi @DarshitChanpura, looks good outside two blank lines and Craig;s comment.

scrawfor99 (commented):
What are your plans for testing these changes? Are you just trying to let the existing tests cover the changes? Is there a way to add explicit checks for the changes you have introduced?

DarshitChanpura:
> What are your plans for testing these changes? Are you just trying to let the existing tests cover the changes? Is there a way to add explicit checks for the changes you have introduced?

This will covered once end to end testing is implemented.

src/main/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluator.java
cwperks: @DarshitChanpura FYI there are some transport action naming conventions in opensearch and you can find more details in this issue: https://github.com/opensearch-project/security/issues/2769#issuecomment-1544987612

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -138,19 +145,53 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
         return user != null && adminDNs.isAdmin(user);
     }
 
+    private boolean authorizeRequest(RestHandler original, RestRequest request, RestChannel channel, User user) throws Exception {
+        if (original instanceof RestSendToExtensionAction) {
+            List<RestHandler.Route> outOfProcessRoutes = original.routes();
```
cwperks, line 150: Why `outOfProcessRoutes`? The routes may be declared by an extension using the SDK, but the routes are present in OpenSearch and opensearch will forward the request to the extension that has registered the route.

DarshitChanpura: it failed code-ql due to extensions keyword here, so I changed the name, but I'll change it back now and suppress the check for this file.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -138,19 +145,53 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
...
+            Optional<RestHandler.Route> handler = outOfProcessRoutes.stream()
+                    .filter(rh -> rh.getMethod().equals(request.method()))
+                    .filter(rh -> restPathMatches(request.path(), rh.getPath()))
+                    .findFirst();
+            if (handler.isPresent() && handler.get() instanceof NamedRoute) {
+                String action = ((NamedRoute)handler.get()).name();
```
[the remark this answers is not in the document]
DarshitChanpura: IMO this should happen during extension initialization or plugin install whichever happens later.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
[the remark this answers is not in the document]
DarshitChanpura: aah nice..thanks I have a test case locally but ran into issues mocking threadpool among other things. I'll take a look and add the tests.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
cwperks: What about `routes`?

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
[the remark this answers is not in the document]
scrawfor99: I know from our conversation yesterday that the ThreadContext stuff has been causing some issues. One way around this is to mock the ThreadPool and then skip over the functionality that requires using the final class. 

Mockito is supposed to support final classes as part of mockito-inline and then later part of mockito 5+, but I was not able to get that running... Instead I doReturned'd the spy'd method which was going to require use of the ThreadContext and set the return value as the expected output. This is not perfect but lets you test the majority of the flow without being able to directly run all of the spy'd methods.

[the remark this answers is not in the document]
DarshitChanpura: I did something similar, where I ended up check the whether `evaluate()` returns things as expected.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
```diff
@@ -484,6 +485,69 @@ public boolean impliesClusterPermissionPermission(String action) {
...
+            }
+
+            log.info("Checking legacy permissions for {}", action);
+
+            action = action.split(":")[1]; // e.g. `hw:greet` would check for action `greet` for ext `hw`
+
+            /* Regex: `/(?:cluster:admin\/\b(open(distro|search))\b\/[a-zA-Z]+\/|\*)action\/?(?:\*|[\/a-zA-Z0-9]*)/gm`
+             *  matches:
+             *  *action*
+             *  cluster:admin/opensearch/abcd/action
```
cwperks, line 527: My main issue with this design is that it implies that even newly minted routes will have a legacy name associated with them. 

Is it possible to have the flexibility to let a plugin developer who is converting their plugin to an extension the power to define an old action name (the action name as it is in the plugin) and a newer name where they are not constrained to the `cluster:*` naming convention of a transport action?

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
DarshitChanpura: we need to suppress it for the file otherwise it fails for `RestSendToExtensionAction`

DarshitChanpura:
CI fails due to stale artifact issue. A recent push was mad to core to update `com.github.luben:zstd-jni` to 1.5.5-3 and it was added to version.properties. However, CI fails because it is unable to find `versions.zstd` which indicates that github cache might not have been updated.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
cwperks: Looking at AD, the current transport action name for get detector info is `cluster:admin/opendistro/ad/detector/info`.

The convention put forward in the SDK in this PR (https://github.com/opensearch-project/opensearch-sdk-java/pull/622 - its been sitting for a while) would use the `extensionShortName` to prefix the names of routes. 

For HelloWorld extension that would mean that if the short name of the extension is `hw` then all names of (REST) actions would be prefixed with `hw:`. Granting full access to the extension would be `hw:*` or granting individual actions with `hw:greet` for example. `hw` is defined in `extensions.yml` though and could be changed to `hello`.

In this design it looks like it would be required for AD as an extension to have a shortened name of `ad` and would not be able to be changed to something like `anomaly`. That may be ok, but would be great if it had the flexibility.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -26,7 +26,10 @@
 
 package org.opensearch.security.filter;
 
+// CS-SUPPRESS-SINGLE: RegexpSingleline Extension is used to refer to certificate extensions, keeping this rule disable for the whole file
```
peternied, line 29: Please activate/deactivate this suppression only in the places where you are making changes.  There is only one other cluster of references to extensions in this file.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -210,4 +253,30 @@ public void onWhitelistingSettingChanged(WhitelistingSettings whitelistingSettin
...
     }
+
+    /**
+     * Determines if the request's path is a match for the configured handler path.
+     *
+     * @param requestPath The path from the {@link NamedRoute}
+     * @param handlerPath The path from the {@link RestHandler.Route}
+     * @return true if the request path matches the route
+     */
+    private boolean restPathMatches(String requestPath, String handlerPath) {
```
peternied, line 264: Not sure what this function works on it looks like basic string comparision, needs unit tests which can clarify

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -138,6 +145,42 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
         return user != null && adminDNs.isAdmin(user);
     }
 
+    private boolean authorizeRequest(RestHandler original, RestRequest request, RestChannel channel, User user) throws Exception {
+        if (original instanceof RestSendToExtensionAction) {
```
peternied, line 149: Lets remove this check, we aren't using the `RestSendToExtensionAction` for anything, no reason any other named route couldn't be used if it was added by a plugin

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -138,6 +145,42 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
...
+                String action = ((NamedRoute)handler.get()).name();
+                PrivilegesEvaluatorResponse pres = evaluator.evaluate(user, action);
+                if (log.isDebugEnabled()) {
+                    log.debug(pres.toString());
+                }
+
+                if (pres.isAllowed()) {
+                    // TODO make sure this is audit logged
+                    log.debug("Request has been granted");
+                    // auditLog.logGrantedPrivileges(action, request, task);
```
peternied, line 165: In a follow up PR lets look into these todo items for auditing

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -138,6 +145,42 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
...
+                    String err;
+                    if(!pres.getMissingSecurityRoles().isEmpty()) {
+                        err = String.format("No mapping for %s on roles %s", user, pres.getMissingSecurityRoles());
+                    } else {
+                        err = String.format("no permissions for %s and %s", pres.getMissingPrivileges(), user);
+                    }
+                    log.debug(err);
+                    // TODO Figure out why ext hangs intermittently after single unauthorized request
+                    channel.sendResponse(new BytesRestResponse(RestStatus.UNAUTHORIZED, err));
+                    return true;
```
peternied, line 177: Shouldn't this return false, as in the request wasn't authorized?

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -138,6 +145,42 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
...
+            if (handler.isPresent() && handler.get() instanceof NamedRoute) {
+                String action = ((NamedRoute)handler.get()).name();
+                PrivilegesEvaluatorResponse pres = evaluator.evaluate(user, action);
+                if (log.isDebugEnabled()) {
+                    log.debug(pres.toString());
+                }
+
+                if (pres.isAllowed()) {
+                    // TODO make sure this is audit logged
+                    log.debug("Request has been granted");
```
peternied, line 164: Is there a performance impact of running these checks against rest + transport actions?  If unclear, how about creating a new issue on the meta.

peternied (dismissed):
Can we change an existing action in the security plugin to use NamedRoute so we can write an integration test that confirms AuthZ for rest works?

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
DarshitChanpura: @scrawfor99 Is working on this.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
DarshitChanpura: done

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
scrawfor99: Correction: Actually Sam and/or Maciej are doing this.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
```diff
@@ -484,6 +485,69 @@ public boolean impliesClusterPermissionPermission(String action) {
...
+         *   reserved: true
+         *   cluster_permissions:
+         *     - 'cluster:admin/opensearch/hw/greet'
+         *
+         *
+         * @param action - The action to be checked against its legacy version
+         * @return true, if a legacy version was found and validated, false otherwise
+         */
+        @Override
+        public boolean impliesLegacyPermission(String action) {
```
peternied, line 514: Why did we need to alter the config model?  Permissions themselves aren't any different, the only difference is how the RestPrivilegeEvaluator checks them, if any match from the NamedRoutes instead of checking a single action name

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -338,7 +338,10 @@ public PrivilegesEvaluatorResponse evaluate(
         );
 
         if (isClusterPerm(action0)) {
-            if (!securityRoles.impliesClusterPermissionPermission(action0)) {
+            // We add a check here for `impliesLegacyPermission` to ensure that no new action names miss evaluation here
+            // e.g. ad:get/detector/info will be evaluated as `cluster:admin/opensearch/ad/get/detector/info` against user's granted
+            // permissions
```
RyanL1997, line 343: Is that means in this case, the `get/detector/info` is the legacy permission? I was just trying to figure out where is this legacy permission coming from.

src/test/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluatorTest.java
```diff
@@ -0,0 +1,170 @@
...
+        Set<String> result = privilegesEvaluator.mapRoles(any(), any());
+
+        assertEquals(mappedRoles, result);
+        verify(configModel).mapSecurityRoles(any(), any());
+    }
+
+    @Test
+    public void testEvaluate_Successful_NewPermission() {
+        String action = "hw:greet";
+        SecurityRoles securityRoles = mock(SecurityRoles.class);
```
RyanL1997, line 126: Just for my knowledge, so the are the mapped used for the security privilege evaluation? Cuz I think you were using the securityRoles here.

src/main/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluator.java
```diff
@@ -0,0 +1,132 @@
...
+                log.info(
+                    "No permission match for {} [Action [{}]] [RolesChecked {}]. No permissions for {}",
+                    user,
+                    action,
+                    securityRoles.getRoleNames(),
+                    presponse.missingPrivileges
+                );
+            } else {
+                if (isDebugEnabled) {
+                    log.debug("Allowed because we have ext permissions for {}", action0);
```
RyanL1997, line 116: nit: should we change the wording "ext"/"extension" here?

RyanL1997 (commented):
Hi @DarshitChanpura, thanks for the great work! For this  I just left some comments/questions.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
DarshitChanpura: legacy is the current nomenclature of defining actions. i.e `cluster:admin/opensearch/...`, with the new permissioning scheme we may have it defined as `ad:detector/info` but the role might still say `cluster:admin/opensearch/ad/detector/info`, and to address this mismatch I added the extra check `impliesLegacyPermissions(action0)`

src/test/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluatorTest.java
DarshitChanpura: `mapRoles` internally maps roles from securityRoles via `this.configModel.mapSecurityRoles(user, caller);`
Check this out: https://github.com/opensearch-project/security/blob/8557fce67470dcb14bf9440530ea76284693a0bd/src/main/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluator.java#L128

src/main/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluator.java
DarshitChanpura: I was thinking about removing 'ext' completely. wdyt?

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
DarshitChanpura: This was added to ensure that the new naming convention still matches the old roles defined. i.e. `ad:detector/info` will match against `cluster:admin/opensearch/ad/detector/info`. Without this, the new permissions will not match against legacy ones.

build.gradle
```diff
@@ -289,7 +289,7 @@ configurations {
             force "io.netty:netty-transport:${versions.netty}"
             force "io.netty:netty-transport-native-unix-common:${versions.netty}"
             force "org.apache.bcel:bcel:6.6.0" // This line should be removed once Spotbugs is upgraded to 4.7.4
-            force "com.github.luben:zstd-jni:${versions.zstd}"
+            force "com.github.luben:zstd-jni:1.5.5-3"
```
peternied, line 292: Revert this change

build.gradle
```diff
@@ -412,7 +412,7 @@ dependencies {
     runtimeOnly 'com.fasterxml.woodstox:woodstox-core:6.4.0'
     runtimeOnly 'org.apache.ws.xmlschema:xmlschema-core:2.2.5'
     runtimeOnly 'org.apache.santuario:xmlsec:2.2.3'
-    runtimeOnly "com.github.luben:zstd-jni:${versions.zstd}"
+    runtimeOnly "com.github.luben:zstd-jni:1.5.5-3"
```
peternied, line 415: Revert this change

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -145,7 +151,56 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
...
 
-    private boolean checkAndAuthenticateRequest(RestRequest request, RestChannel channel, NodeClient client) throws Exception {
+    private boolean authorizeRequest(RestHandler original, RestRequest request, RestChannel channel, User user) {
+
+        List<RestHandler.Route> restRoutes = original.routes();
+        Optional<RestHandler.Route> handler = restRoutes.stream()
+            .filter(rh -> rh.getMethod().equals(request.method()))
+            .filter(rh -> restPathMatches(request.path(), rh.getPath()))
+            .findFirst();
+        if (handler.isPresent() && handler.get() instanceof NamedRoute) {
```
peternied, line 161: Invert the flow of control to lower the intending making it easier to read
```suggestion
        if (!handler.isPresent() || !(handler.get() instanceof NamedRoute)) {
           return false;
        }
```

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -145,7 +151,56 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
...
+                }
+                log.debug(err);
+                // TODO Figure out why ext hangs intermittently after single unauthorized request
+                channel.sendResponse(new BytesRestResponse(RestStatus.UNAUTHORIZED, err));
+                return false;
+            }
+        }
+
+        // if handler is not an instance of NamedRoute then we pass through to eval at Transport Layer.
+        // TODO Change this to false once all plugins have migrated to use NamedRoutes
```
peternied, line 199: We will migrate all plugins to use NamedRoutes?  IMO I think we should accept that we will have standard routes used by some plugins and core forever.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -145,7 +151,56 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
         return user != null && adminDNs.isAdmin(user);
     }
 
-    private boolean checkAndAuthenticateRequest(RestRequest request, RestChannel channel, NodeClient client) throws Exception {
+    private boolean authorizeRequest(RestHandler original, RestRequest request, RestChannel channel, User user) {
```
peternied, line 154: Shouldn't this return a `PrivilegesEvaluatorResponse` so it can be passed into the rest of the ecosystem, such as in the transport layer?

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -145,7 +151,56 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
...
+            if (!pres.isAllowed()) {
+                String action = route.name();
+                pres = evaluator.evaluate(user, Set.of(action));
+            }
+
+            if (log.isDebugEnabled()) {
+                log.debug(pres.toString());
+            }
+            if (pres.isAllowed()) {
+                // TODO make sure this is audit logged
```
peternied, line 180: Are these todo for this pull request or should there be issues that track these items?

peternied: (Many places)

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
peternied: This is a good integration test scenario to add

peternied (commented):
Taken another pass, looks like some comments might have been lost after the great refactoring due to the spotless changes.

config/internal_users.yml
```diff
@@ -61,3 +61,8 @@ snapshotrestore:
   backend_roles:
   - "snapshotrestore"
   description: "Demo snapshotrestore user, using external role mapping"
+
+new-user:
```
cwperks, line 65: I don't think we should add a demo user here. If this is used for an integration test, can this be moved to a test file?

config/roles.yml
```diff
@@ -355,3 +355,22 @@ security_analytics_ack_alerts:
   reserved: true
   cluster_permissions:
     - 'cluster:admin/opensearch/securityanalytics/alerts/*'
+
+# Roles for Hello World extension, roles that extensions would like to add should be ultimately be defined elsewhere
```
cwperks, line 359: Same for these. If there are specific demo configurations that we would like to ship by default would it make sense to have separate files in different folders under `config/` here? Maybe put these into `config/extension_demo/roles.yml`?

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -145,7 +151,56 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
...
+            } else {
+                // auditLog.logMissingPrivileges(action, request, task);
+                String err;
+                if (!pres.getMissingSecurityRoles().isEmpty()) {
+                    err = String.format("No mapping for %s on roles %s", user, pres.getMissingSecurityRoles());
+                } else {
+                    err = String.format("no permissions for %s and %s", pres.getMissingPrivileges(), user);
+                }
+                log.debug(err);
+                // TODO Figure out why ext hangs intermittently after single unauthorized request
```
cwperks, line 192: I added this TODO statement when encountering an issue with extensions. We should test this with plugins to ensure the issue is not present.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
cwperks: IMO we don't need this since `NamedRoutes` support registering legacy action names as well.

cwperks (commented):
Took another pass at this PR. It looks close to me, but IMO I don't think we need to do the regex matching and rely on NamedRoute registration.

src/main/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluator.java
RyanL1997: Ya I think that makes 100% sense.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
DarshitChanpura: I will address all TODOs

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
DarshitChanpura: this method evaluates new permissions against legacy permission. For example, a route `GET /x` is registered with a unique name `getX` and no legacy permission actions. We can't directly evaluate using `impliesClusterPermission`, instead we would need this method to translate `getX` to a legacy name, i.e.`cluster:admin/opensearch/plugin1/getX` and then evaluated.

DarshitChanpura: We won't need this at all if we are mandating actionNames to be associated with each PR. Then we will rely on those to evaluate against the user's permissions.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
peternied: Note; you can also use a `final boolean doesRouteSupportRestAuthorization = /* gunk */;` to protect yourself in this case

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
DarshitChanpura: For now, I have made the checks plugin name agnostic. As long as the unique action name matches, it will return true.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -109,28 +117,26 @@ public SecurityRestFilter(
...
-                    User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
-                    if (userIsSuperAdmin(user, adminDNs)
-                        || (whitelistingSettings.checkRequestIsAllowed(request, channel, client)
-                            && allowlistingSettings.checkRequestIsAllowed(request, channel, client))) {
+        return (request, channel, client) -> {
+            org.apache.logging.log4j.ThreadContext.clearAll();
+            if (!checkAndAuthenticateRequest(request, channel)) {
+                User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+                if (userIsSuperAdmin(user, adminDNs)
+                    || (whitelistingSettings.checkRequestIsAllowed(request, channel, client)
```
scrawfor99, line 137: Swap to allowlisting?

scrawfor99 (commented):
Hi @DarshitChanpura, looks close to me. Nothing major I noticed outside of Craig's most recent comments

build.gradle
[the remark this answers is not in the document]
DarshitChanpura: Seems like an old change you commented on. This has been fixed in the most recent push

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
DarshitChanpura: we are checking for both to be backwards compatible. Check out line below where I've added a check for allowlistSettings

build.gradle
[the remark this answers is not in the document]
scrawfor99: Oops!

scrawfor99:
Hi @DarshitChanpura, just following up here to see if there is anything you need help with to get this across the finish line. It looks really close, so let me know if there is anything you need and if not, if the CI gets passing we can provide final reviews.

DarshitChanpura:
@opensearch-project/security I have modified `GET /whoami` route to now be a NamedRoute and added integration tests  for the same. Could I get a re-review of this PR?

REST_AUTHZ_FOR_PLUGINS.md
```diff
@@ -0,0 +1,134 @@
+# Authorization at REST Layer for plugins
+
+This feature is introduced as an added layer of security on top of existing TransportLayer authorization framework. In order to leverage these feature some core changes need to be made at Route registration level. This document talks about you can achieve this.
```
scrawfor99, line 3:
```suggestion
This feature is introduced as an added layer of security on top of existing TransportLayer authorization framework. In order to leverage these feature some core changes need to be made at Route registration level. This document talks about how you can achieve this.
```

REST_AUTHZ_FOR_PLUGINS.md
```diff
@@ -0,0 +1,134 @@
...
+
+The security plugin must be installed and operational in OpenSearch cluster for this feature to work.
+
+### How does NamedRoute authorization work?
+
+Once the routes are defined as NamedRoute, they, along-with their handlers, will be registered the same way as Route. Once the request comes in `SecurityRestFilter.java` applies an authorization check which extracts information about the NamedRoute.
+Next we get the unique named associated with that route and evaluate those against existing `cluster_permissions` across all roles of the requesting user. If the authz check succeeds the request chain proceeds as normal. Else, a 403 response is returned to the user.
+
+NOTE:
+1. These action names defined in roles must exactly match the names of registered routes, or else, the request would be deemed unauthorized.
```
scrawfor99, line 15:
```suggestion
1. The action names defined in roles must exactly match the names of registered routes, or else, the request would be deemed unauthorized.
```

REST_AUTHZ_FOR_PLUGINS.md
```diff
@@ -0,0 +1,134 @@
...
+    return ImmutableList.of(
+            new NamedRoute.Builder().method(GET).path("/uri").uniqueName("plugin:uri").actionNames(Set.of("cluster:admin/opensearch/plugin/uri")).build()
+        );
+}
+```
+
+`actionNames()` are optional. They correspond to any current actions defined as permissions in roles.
+Ensure that these name-to-route mappings are easily accessible to the cluster admins to allow granting access to these APIs.
+
+### How does authorization at REST Layer work?
```
scrawfor99, line 43:
```suggestion
### How does authorization in the REST Layer work?
```

REST_AUTHZ_FOR_PLUGINS.md
```diff
@@ -0,0 +1,134 @@
+# Authorization at REST Layer for plugins
+
+This feature is introduced as an added layer of security on top of existing TransportLayer authorization framework. In order to leverage these feature some core changes need to be made at Route registration level. This document talks about you can achieve this.
+
+## Pre-requisites
+
+The security plugin must be installed and operational in OpenSearch cluster for this feature to work.
```
scrawfor99, line 7:
```suggestion
The security plugin must be installed and operational in your OpenSearch cluster for this feature to work.
```

REST_AUTHZ_FOR_PLUGINS.md
```diff
@@ -0,0 +1,134 @@
...
+
+### How does NamedRoute authorization work?
+
+Once the routes are defined as NamedRoute, they, along-with their handlers, will be registered the same way as Route. Once the request comes in `SecurityRestFilter.java` applies an authorization check which extracts information about the NamedRoute.
```
scrawfor99, line 11:
```suggestion
Once the routes are defined as NamedRoute, they, along-with their handlers, will be registered the same way as Route objects. When a request comes in, `SecurityRestFilter.java` applies an authorization check which extracts information about the NamedRoute.
```

REST_AUTHZ_FOR_PLUGINS.md
```diff
@@ -0,0 +1,134 @@
...
+Next we get the unique named associated with that route and evaluate those against existing `cluster_permissions` across all roles of the requesting user. If the authz check succeeds the request chain proceeds as normal. Else, a 403 response is returned to the user.
```
scrawfor99, line 12:
```suggestion
Next we get the unique named associated with that route and evaluate it against existing `cluster_permissions` across all roles of the requesting user. If the authorization check succeeds, the request chain proceeds as normal. If it fails, a 403 response is returned to the user.
```

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -145,7 +150,54 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
...
+                    err = String.format("no permissions for %s and %s", pres.getMissingPrivileges(), user);
+                }
+                log.debug(err);
+                channel.sendResponse(new BytesRestResponse(RestStatus.UNAUTHORIZED, err));
+                return false;
+            }
+        }
+
+        // if handler is not an instance of NamedRoute then we pass through to eval at Transport Layer.
+        return true;
```
scrawfor99, line 197: If I am reading this correctly, we are always going to be evaluating both on the REST and trasnport layer correct? The only case where things would happen only in the REST layer would be if the request were unauthorized?

src/main/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluator.java
```diff
@@ -0,0 +1,132 @@
...
+
+    public RestLayerPrivilegesEvaluator(
+        final ClusterService clusterService,
+        final ThreadPool threadPool,
+        AuditLog auditLog,
+        final ClusterInfoHolder clusterInfoHolder,
+        AtomicReference<NamedXContentRegistry> namedXContentRegistry
+    ) {
+
+        super();
```
scrawfor99, line 53: Is there a reason we need to super() Object here?

src/main/java/org/opensearch/security/tools/SecurityAdmin.java
```diff
@@ -577,39 +577,44 @@ public static int execute(final String[] args) throws Exception {
...
-            Response whoAmIRes = restHighLevelClient.getLowLevelClient().performRequest(new Request("GET", "/_plugins/_security/whoami"));
-            if (whoAmIRes.getStatusLine().getStatusCode() != 200) {
-                System.out.println("Unable to check whether cluster is sane because return code was " + whoAmIRes.getStatusLine());
-                return (-1);
-            }
+            if (whoami) {
+                Response whoAmIRes = restHighLevelClient.getLowLevelClient()
+                    .performRequest(new Request("GET", "/_plugins/_security/whoami"));
+                if (whoAmIRes.getStatusLine().getStatusCode() != 200) {
+                    System.out.println("Unable to check whether cluster is sane because return code was " + whoAmIRes.getStatusLine());
```
scrawfor99, line 584:
```suggestion
                    System.out.println("Unable to check whether cluster is successfully initialized because return code was " + whoAmIRes.getStatusLine());
```

Sane may have connotations so we may want to use another phrase.

scrawfor99 (commented):
Hi @DarshitChanpura, great work with this! Well done. I left a few wording suggestions and two comments but overall looks good to me. We will just want to address the current failures with the CI and I think the rest should be about ready code-wise.

src/main/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluator.java
DarshitChanpura: good catch!

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
DarshitChanpura: > If I am reading this correctly, we are always going to be evaluating both on the REST and trasnport layer correct?

Yes.

> The only case where things would happen only in the REST layer would be if the request were unauthorized?

I'm not quite sure I understand this. Can you elaborate a little?

scrawfor99: I meant to clarify that the only time a request would end in the REST layer would be if it were unauthorized by the REST layer auth checker. Otherwise, it will always hit the transport layer checks.

scrawfor99 (dismissed):

DEVELOPER_GUIDE.md
```diff
@@ -78,6 +79,51 @@ mv config/* $OPENSEARCH_HOME/config/opensearch-security/
 rm -rf config/
 ```
 
+### Installing demo extension users and roles
+
+If you are working with an extension and want to set up demo users for the Hello-World extension, append following items to files inside `$OPENSEARCH_HOME/config/opensearch-security/`:
+1. In **internal_users.yml**
+```yaml
+new-user:
```
cwperks, line 87: nit: should probably be a name other than new-user. Maybe hw-user?

REST_AUTHZ_FOR_PLUGINS.md
```diff
@@ -0,0 +1,134 @@
...
+who_am_i_role_no_perm:
+  reserved: true
+  cluster_permissions:
+    - 'some_invalid_perm'
+
+```
+
+#### internal_users.yml
+```yaml
+who_am_i-user:
```
cwperks, line 96: nit: mix of dashes and underscores

REST_AUTHZ_FOR_PLUGINS.md
```diff
@@ -0,0 +1,134 @@
...
+who_am_i_role_no_perm:
+  reserved: true
+  users:
+    - "who_am_i_no_perm-user"
+```
+
+Follow [DEVELOPER_GUIDE](DEVELOPER_GUIDE.md) to setup OpenSearch cluster and initialize security plugin. Once you have verified that security plugin is installed correctly and OpenSearch is running, execute following curl requests:
+1. `curl -XGET https://who_am_i-user:admin@localhost:9200/_plugins/_security/whoami --insecure` should succeed.
+2. `curl -XGET https://who_am_i_legacy-user:admin@localhost:9200/_plugins/_security/whoami --insecure` should succeed.
+3. `curl -XGET https://who_am_i_no-perm-user:admin@localhost:9200/_plugins/_security/whoami --insecure` should fail.
```
cwperks, line 133: Should this say whoamiprotected now?

REST_AUTHZ_FOR_PLUGINS.md
```diff
@@ -0,0 +1,134 @@
+# Authorization at REST Layer for plugins
```
cwperks, line 1: General feedback on this doc. I think it should convey to plugin developers that they can opt not to write a transport action for a route if its unnecessary. From what I understand, plugin developers will write a transport action to protect and endpoint and not because they need to write a transport action. This is fine, but I think the doc should touch on the subject.

src/integrationTest/java/org/opensearch/security/rest/WhoAmITests.java
```diff
@@ -0,0 +1,107 @@
...
+import org.opensearch.test.framework.cluster.LocalCluster;
+import org.opensearch.test.framework.cluster.TestRestClient;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class WhoAmITests {
```
cwperks, line 31: 🙌

src/main/java/org/opensearch/security/action/whoami/WhoAmIAction.java
```diff
@@ -31,7 +31,7 @@
 public class WhoAmIAction extends ActionType<WhoAmIResponse> {
 
     public static final WhoAmIAction INSTANCE = new WhoAmIAction();
-    public static final String NAME = "cluster:admin/opendistro_security/whoami";
+    public static final String NAME = "cluster:admin/opendistro_security/whoamiprotected";
```
cwperks, line 34: Is this change necessary?

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -145,7 +152,54 @@ private boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
...
+                String action = route.name();
+                pres = evaluator.evaluate(user, Set.of(action));
+            }
+
+            if (log.isDebugEnabled()) {
+                log.debug(pres.toString());
+            }
+            if (pres.isAllowed()) {
+                log.debug("Request has been granted");
+                auditLog.logGrantedPrivileges(user.getName(), request);
```
cwperks, line 183: Can we create a f/u item to write automated tests for the audit logging here?

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -217,4 +271,30 @@ public void onWhitelistingSettingChanged(WhitelistingSettings whitelistingSettin
...
     }
+
+    /**
+     * Determines if the request's path is a match for the configured handler path.
+     *
+     * @param requestPath The path from the {@link NamedRoute}
+     * @param handlerPath The path from the {@link RestHandler.Route}
+     * @return true if the request path matches the route
+     */
+    private boolean restPathMatches(String requestPath, String handlerPath) {
```
cwperks, line 282: Units like this should have tests with them. I've seen code similar to this in core, can this be extracted to a library?

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
```diff
@@ -496,6 +497,78 @@ public boolean impliesClusterPermissionPermission(String action) {
...
```
cwperks, line 527: Is this still needed now that action names can be associated directly in a NamedRoute? IMO I'd like to see this removed.

cwperks (commented):
Took another pass. @DarshitChanpura Is it possible to remove the `impliesLegacyPermission`? This PR looks close, but I'd like to get resolution on this first.

REST_AUTHZ_FOR_PLUGINS.md
DarshitChanpura: I kept this on purpose, to differentiate between a role and a user since they have same prefix.

src/main/java/org/opensearch/security/action/whoami/WhoAmIAction.java
DarshitChanpura: No.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
DarshitChanpura: I will remove this code, but that will mean that action names will have to be passed in order for legacy check to work.

cwperks: IMO it makes sense to put that burden on the extension/plugin dev and not try to formulate what the legacy transport action name is associated with what is being converted.

When Anomaly Detection plugin is being converted to an extension, the extension dev would have to convert each route handler to using a NamedRoute to afford it the same protections they have in place today with transport action names. If they want to keep their existing roles compatible with how they are defined as a plugin today then they would register the `actionNames` list of `NamedRoute` and the evaluator introduced in this PR would evaluate permissions against the `actionNames` registered with the route. There is no inferring the name of a transport action name.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
DarshitChanpura: Test file already exists. Check it out https://github.com/opensearch-project/security/blob/9b05d44a6ba15bca521e1ccbd6e33124facddcc0/src/test/java/org/opensearch/security/filter/RestPathMatchesTest.java

> can this be extracted to a library?

It's only used at 2 places and I don't think it will be used at other places in future. I think keeping it here is fine.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
DarshitChanpura: done. Added an item on meta.

src/test/java/org/opensearch/security/filter/RestPathMatchesTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Test
+    public void testPartialMatch() throws InvocationTargetException, IllegalAccessException {
+        String requestPath = "_plugins/security/api/x/y";
+        String handlerPath = "_plugins/security/api/x/z";
+        assertFalse((Boolean) restPathMatches.invoke(securityRestFilter, requestPath, handlerPath));
+    }
+
+    @Test
+    public void testNamedParamsMatch() throws InvocationTargetException, IllegalAccessException {
```
RyanL1997, line 47: I was a bit confused by this test case that where to get the actual value of `{id}`. Or is the method `restPathMatches` seems to understand these placeholders, and considers "123" in the requestPath to match "{id}" in the handlerPath?

REST_AUTHZ_FOR_PLUGINS.md
DarshitChanpura: will add a note here.

RyanL1997 (commented):
Hi @DarshitChanpura, in generally LGTM beyond some of the suggestions that Craig and Stephen left and thanks for all the work on this one!

src/main/java/org/opensearch/security/action/whoami/WhoAmIAction.java
DarshitChanpura: No. it's a protected version of `/whoami`

src/test/java/org/opensearch/security/filter/RestPathMatchesTest.java
DarshitChanpura: yes. Parameter {id} will resolve to 123 for this request path which is what we want. {id} acts as a placeholder in the handler for any actual parameters passed in the request.

RyanL1997 (dismissed):

DarshitChanpura:
> Can we change an existing action in the security plugin to use NamedRoute so we can write an integration test that confirms AuthZ for rest works?

@peternied Added a new endpoint `GET /whoamiprotected` and integrations tests for it `WhoAmITests.java`.

src/test/java/org/opensearch/security/SystemIntegratorsTests.java
```diff
@@ -179,7 +179,7 @@ public void testInjectedUser() throws Exception {
         Assert.assertTrue(resc.getBody().contains("\"remote_address\":\"8.8.8.8:8\""));
         Assert.assertTrue(resc.getBody().contains("\"backend_roles\":[\"role1\",\"role2\"]"));
         // mapped by username
-        Assert.assertTrue(resc.getBody().contains("\"roles\":[\"opendistro_security_all_access\""));
+        Assert.assertTrue(resc.getBody().contains("\"opendistro_security_all_access\""));
```
cwperks, line 182: Was this change needed?

scrawfor99 (dismissed):
Great job DC!

src/test/java/org/opensearch/security/SystemIntegratorsTests.java
DarshitChanpura: Yes. because nagilum is now associated with another role called `who_am_i`, the request body won't always start with `\"roles\":[`

src/main/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluator.java
```diff
@@ -0,0 +1,130 @@
...
+        final SecurityRoles securityRoles = getSecurityRoles(mappedRoles);
+
+        final boolean isDebugEnabled = log.isDebugEnabled();
+        if (isDebugEnabled) {
+            log.debug("Evaluate permissions for {} on {}", user, clusterService.localNode().getName());
+            log.debug("Action: {}", action0);
+            log.debug("Mapped roles: {}", mappedRoles.toString());
+        }
+
+        for (String action : action0) {
```
cwperks, line 101: nit: I had to reference up above that `action0` is a set. This looks like its looping through a singular action.

DarshitChanpura: It will be singular for `name`, but it will be in loop for `actionNames`

cwperks (dismissed):

src/test/java/org/opensearch/security/SystemIntegratorsTests.java
cwperks: Similar to the assertion above with backend roles, I think this should assert all expected roles

RyanL1997 (approved):

scrawfor99 (approved):

cwperks (approved):

PR #2765, opened 2023-05-11
Adds a check to skip serialization-deserialization if request is for same node

### LINKED ISSUE #2724: [Performance Bug] Repeated deserialization of headers for a direct channel request
**What is the issue?**
During a direct channel request i.e. a request which originated and is handled by the same node, the security plugin kicks in and deserializes `userHeader` and `originalRemoteAddress` objects again hence consuming more time to process the request.

A simple bulk request on a multi node cluster should reproduce the issue.

Do not deserialize the objects again for direct channel requests, and re-use the previously deserialized values.

 - OS v2.3
 - Security Plugin

`SecurityIntercepter.sendMessageDecorate` stashes the current thread context due to which the transient headers are cleared. `SecurityRequestHandler.messageReceivedDecorate` deserializes the userHeader and originalRemoteAddress objects just to put them in thread context's transient headers. 

For a direct channel request (channelType == "direct"), `SecurityIntercepter.sendMessageDecorate` may retain these values in transient headers and `SecurityRequestHandler.messageReceivedDecorate` will no longer need to deserialize these again.

DarshitChanpura:
### Description
* Category:
Bug Fix

To improve performance of large requests handled on same node

### Issues Resolved
- Resolves #2724 

### Testing
Current test suite should pass.

### Check List


src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
```diff
@@ -177,21 +170,21 @@ else if(!Strings.isNullOrEmpty(injectedUserHeader)) {
 
             boolean skipSecurityIfDualMode = getThreadContext().getTransient(ConfigConstants.SECURITY_SSL_DUAL_MODE_SKIP_SECURITY) == Boolean.TRUE;
 
-                if(skipSecurityIfDualMode) {
-                    if(getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS) == null) {
-                        getThreadContext().putTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS, request.remoteAddress());
-                    }
+            if(skipSecurityIfDualMode) {
```
DarshitChanpura, line 173: fixed indentation to promote better readability, as it was misaligned in the file

DarshitChanpura (commented):
Confirmed working behavior by running some test suites locally. Waiting for GH runner to run complete test suite.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -127,6 +128,8 @@ public <T extends TransportResponse> void sendRequestDecorate(AsyncSender sender
         final String origCCSTransientMf = getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_MASKED_FIELD_CCS);
 
         final boolean isDebugEnabled = log.isDebugEnabled();
+        final boolean isDirectRequest = HeaderHelper.isDirectRequest(getThreadContext());
+
         try (ThreadContext.StoredContext stashedContext = getThreadContext().stashContext()) {
```
DarshitChanpura, line 133: Could not yet find a solution to not stash the context for same node.

Tried with these changes:
```
try (ThreadContext.StoredContext stashedContext = (isDirectRequest ? null: getThreadContext().stashContext())) {
    final TransportResponseHandler<T> restoringHandler = isDirectRequest ? handler : new RestoringTransportResponseHandler<T>(handler, stashedContext);
```

But the test fails with ClusterManagernotDiscoveredException everytime on test start.



```

File: src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java [hunk 3]
```diff
@@ -137,11 +137,12 @@ protected void messageReceivedDecorate(final T request, final TransportRequestHa
...
             //bypass non-netty requests
             if(channelType.equals("direct")) {
-                final String userHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER);
+                // for direct channel requests we don't serialize the user object in sendRequestDecorate
+                final User user = getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
                 final String injectedRolesHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROLES_HEADER);
                 final String injectedUserHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER_HEADER);
 
-                if(Strings.isNullOrEmpty(userHeader)) {
+                if(user != null) {
```
cwperks, line 145: Why remove the empty check here?

DarshitChanpura: Because for direct requests we don't pass headers (which would then be serialized and put as a transient), rather we directly put the transient object, User, in the thread context via `sendMessageDecorate` in SecurityInterceptor and retrieve it as an object here. In doing so we skip serialization and deserialization of direct requests.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -224,7 +225,14 @@ private void ensureCorrectHeaders(final Object remoteAdr, final User origUser, f
...
             if(remoteAddressHeader == null) {
-                getThreadContext().putHeader(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS_HEADER, Base64Helper.serializeObject(((TransportAddress) remoteAdr).address()));
+                TransportAddress transportAddress = (TransportAddress) remoteAdr;
+
+                if(isDirectRequest) {
+                    // if request is going to be handled by same node, we directly put transient value.
+                    getThreadContext().putTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS, transportAddress);
+                } else {
+                    getThreadContext().putHeader(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS_HEADER, Base64Helper.serializeObject(transportAddress.address()));
+                }
```
krishna-ggk, line 235: Per entry if-else clauses can be prone to error as it needs matching checks on receiver end. Can we generalize to either use transients vs header for all config entries?

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
krishna-ggk: It is perhaps good idea to continue stashing from security point of view too. That way, any existing transients assumed to be cleared are not carry forwarded.

For example let us assume current thread context had a,b,c entries and we explicitly passed around only a & b. Now carry forwarding c may be a security concern.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -127,6 +128,8 @@ public <T extends TransportResponse> void sendRequestDecorate(AsyncSender sender
         final String origCCSTransientMf = getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_MASKED_FIELD_CCS);
 
         final boolean isDebugEnabled = log.isDebugEnabled();
+        final boolean isDirectRequest = HeaderHelper.isDirectRequest(getThreadContext());
```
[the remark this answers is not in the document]
nibix: Yep, that is the wrong end.

In `SecurityInterceptor` we need to check whether `connection.getNode()` is equal to the local node, similar to how `TransportService` is doing it here:

https://github.com/opensearch-project/OpenSearch/blob/1e82b894902a4e34ed2cf71e97511ace72d0205c/server/src/main/java/org/opensearch/transport/TransportService.java#L1620

The local node could be retrieved by `clusterService.localNode()`. I am not sure, though, whether this is a 100% correct approach, as `TransportService` as a slightly more involved way to get the local node. See:

https://github.com/opensearch-project/OpenSearch/blob/1e82b894902a4e34ed2cf71e97511ace72d0205c/server/src/main/java/org/opensearch/transport/TransportService.java#L306

Alternatively, one might try to get ahold of a reference to `TransportService`. However, this only works with hacks in a non-Guice context:

https://github.com/opensearch-project/security/blob/5c6342038e0a6b4afe577bcfcba30213bcb35776/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L1198

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
nibix: > But the test fails with ClusterManagernotDiscoveredException everytime on test start.

I would guess that this is related to the wrong check for `isDirectRequest`. But, yes, I would also recommend to stash anyway and just restore the transients which you want to keep. That would be the most conservative and most minimal change.

DarshitChanpura:
CI is currently failing with following to change introduced in core: https://github.com/opensearch-project/OpenSearch/pull/7465
```
 java.lang.NoClassDefFoundError: org/opensearch/core/common/io/stream/BaseStreamInput
```

DarshitChanpura:
> CI is currently failing with following to change introduced in core: [opensearch-project/OpenSearch#7465](https://github.com/opensearch-project/OpenSearch/pull/7465)
> 
> ```
>  java.lang.NoClassDefFoundError: org/opensearch/core/common/io/stream/BaseStreamInput
> ```

Seems like a stale artifact issue that will be fixed on its own.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -142,17 +144,15 @@ public <T extends TransportResponse> void sendRequestDecorate(AsyncSender sender
                             || k.equals(ConfigConstants.OPENDISTRO_SECURITY_DOC_ALLOWLIST_HEADER)
                             || k.equals(ConfigConstants.OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE)
                             || k.equals(ConfigConstants.OPENDISTRO_SECURITY_DLS_MODE_HEADER)
-                            || k.equals(ConfigConstants.OPENDISTRO_SECURITY_DLS_FILTER_LEVEL_QUERY_HEADER)                            
+                            || k.equals(ConfigConstants.OPENDISTRO_SECURITY_DLS_FILTER_LEVEL_QUERY_HEADER)
```
hardik-k-shah, line 147: possible to keep formatting changes separate from this PR?

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
hardik-k-shah: should we say `isLocalNodeRequest` to add more readability?  or are we categorizing local node request to  `direct` channel through out the code-base ?

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -224,25 +224,42 @@ private void ensureCorrectHeaders(final Object remoteAdr, final User origUser, f
...
+                }
             }
         }
 
-
+        User user = getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
         String userHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER);
 
-        if(userHeader == null) {
+        if(isDirectRequest) {
```
hardik-k-shah, line 241: can be merge with above if..else..

krishna-ggk (commented):
Can we add tests to ensure both direct and non-direct paths are explicitly validated?

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -224,25 +224,42 @@ private void ensureCorrectHeaders(final Object remoteAdr, final User origUser, f
...
-                getThreadContext().putHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER, Base64Helper.serializeObject(origUser));
+                // if request is going to be handled by same node, we directly put transient value as the thread context is not going to be stah.
+                getThreadContext().putTransient(ConfigConstants.OPENDISTRO_SECURITY_USER, origUser);
+            } else if(StringUtils.isNotEmpty(injectedRolesString)) {
+                getThreadContext().putTransient(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROLES, injectedRolesString);
+            } else if(StringUtils.isNotEmpty(injectedUserString)) {
+                getThreadContext().putTransient(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER, injectedUserString);
             }
-            else if(StringUtils.isNotEmpty(injectedRolesString)) {
+        } else {
```
parasjain1, line 251: Do we not intend to retain the existing `userHeader == null` check for this `else` code block, or are we assuming that this scenario used to arise only for direct requests?

parasjain1 (commented):
Left a comment for a query I have regarding the changes.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -127,6 +127,13 @@ public <T extends TransportResponse> void sendRequestDecorate(AsyncSender sender
         final String origCCSTransientMf = getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_MASKED_FIELD_CCS);
 
         final boolean isDebugEnabled = log.isDebugEnabled();
+        boolean isSameNodeRequest = false;
+        try {
+            isSameNodeRequest = cs.localNode().equals(connection.getNode()); // using DiscoveryNode equals comparison here
+        } catch (AssertionError e) {
+            // do nothing
+        }
```
krishna-ggk, line 135: This doesn't seem right - which assertion is failing?

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -219,30 +224,46 @@ private void ensureCorrectHeaders(final Object remoteAdr, final User origUser, f
...
-                getThreadContext().putHeader(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS_HEADER, Base64Helper.serializeObject(((TransportAddress) remoteAdr).address()));
+                transportAddress = (TransportAddress) remoteAdr;
             }
         }
 
+        // we put headers as transient for same node requests
+        if (isSameNodeRequest) {
 
-        String userHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER);
+            if(transportAddress != null) {
```
RyanL1997, line 238: For my knowledge, what's this `transportAddress`? Is this the same as `userHeader`?

DarshitChanpura: its the variable that holders the address of the remote cluster node (could be same cluster node too, but its store as remoteAddress in thread context)

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
nibix: In order to form the initial cluster state, it is already necessary to exchange transport messages which pass through here. TransportService uses a different approach to retrieve the reference to the local node. As this also needs to pass through the early transport messages, it is likely that here the local node information is already available:

https://github.com/opensearch-project/OpenSearch/blob/1e82b894902a4e34ed2cf71e97511ace72d0205c/server/src/main/java/org/opensearch/transport/TransportService.java#L306

DarshitChanpura: @nibix Apologies, but I'm not super clear on whether you want the implementation to be changed.

nibix: Well, I think catching an Error, especially an AssertionError is a code smell. It is problematic for two reasons:

- You could be also catching other AssertionErrors which are more serious and not meant to be ignored.
- Assertions are usually only enabled in dev environments. However, that does not mean that the error will just go away in a prod environment. It is possible that it is just deferred to a different code place which then bails with an NPE or similar thing. Thus, code that works in dev could break in prod.

Thus, I would recommend to use an approach which does not require catching AssertionErrors.

nibix: Hm, I guess that will be also the very early initialization phase when localNode in TransportService is null, right?

I would still believe that TransportService will know earlier about localNode than ClusterService. Also, if we use the localNode from TransportService, we could use a simple null check instead of catching the AssertionError.

DarshitChanpura: Will the null check not defeat the purpose? How will it be different from catching an AssertionError when cluster state is not accessible from this class?

DarshitChanpura: thank you for the explanation!. I'll update the condition here. If you have any resources handy that I can take a look for assertions and their usage in Prod vs staging, it would be helpful for me.

DarshitChanpura:
@nibix After updating GuiceHolder to contains an instance of TransportService and use that instance to fetch the localNode information, I'm getting different results:

```
cs.localNode() = {DiscoveryNode@12410} "{node_utest_n1_fnull_t353609252151333_num3}{cHyj4v__T_-MMs2s_____w}{_Rk3Y5VWTUeuzUrwCfr9Rg}{127.0.0.1}{127.0.0.1:6212}{m}{shard_indexing_pressure_enabled=true}"

connection.getNode() = {DiscoveryNode@12410} "{node_utest_n1_fnull_t353609252151333_num3}{cHyj4v__T_-MMs2s_____w}{_Rk3Y5VWTUeuzUrwCfr9Rg}{127.0.0.1}{127.0.0.1:6212}{m}{shard_indexing_pressure_enabled=true}"

localNode = {DiscoveryNode@12404} "{node_utest_n1_fnull_t353609252151333_num1}{1S3HEQAAQACnOySu_____w}{LUJGxVdsSoqtJ3kB3RZ4wg}{127.0.0.1}{127.0.0.1:7352}{d}{shard_indexing_pressure_enabled=true}"
```

the localNode value from ClusterState seems to be matching the value of the node the connection originated from (resulting in passing tests), but localNode value (retrieved from TransportService instance) is not matching connection.getNode() resulting in test failures.
Am i missing something here in terms of fetching localNode info from TransportService?

nibix:
> @nibix After updating GuiceHolder to contain an instance of TransportService and use that instance to fetch the localNode information, I'm getting different results

Oh, hm, interesting. I think I have an idea where this is coming from. Am I right in the assumption that you get that error when testing in a junit based test with a cluster embedded in the JVM? 

If yes: This indeed collides with the GuiceHolder. Unfortunately, the GuiceHolder is at closer consideration a quite dirty hack to get ahold of dependencies which area only available by deprndecy injection. It uses for this a static attribute in the SecurityPlugin class. While in an JVM embedded cluster the nodes and also the instances of the SecurityPlugin can coexist very well together, this static class is a collision. The different nodes write in an undetermined fashion references to their own TransportService into it.

That's why you see other results there, at least if my assumption is correct.

Of course the more important part of the consideration would be "how can this be fixed". To be honest I also don't know at the moment. I will have to look at the code and think about it for a while. 😔

DarshitChanpura:
> Oh, hm, interesting. I think I have an idea where this is coming from. Am I right in the assumption that you get that error when testing in a junit based test with a cluster embedded in the JVM?

Yes, I saw this while running one of the failing unit tests.

> If yes: This indeed collides with the GuiceHolder. Unfortunately, the GuiceHolder is at closer consideration a quite dirty hack to get ahold of dependencies which area only available by deprndecy injection. It uses for this a static attribute in the SecurityPlugin class. While in an JVM embedded cluster the nodes and also the instances of the SecurityPlugin can coexist very well together, this static class is a collision. The different nodes write in an undetermined fashion references to their own TransportService into it.

If that is indeed the case, then do you think it might be worthwhile to look into using cluster state() as an alternative for getting localNode reference, cause that seems to be available at almost all time (except during cluster bootstrap using it Zen Discover mechanism, although I'm not a 100% sure whether the request will flow through security with some cluster state info during bootstrap).

> Of course the more important part of the consideration would be "how can this be fixed". To be honest I also don't know at the moment. I will have to look at the code and think about it for a while. 😔

Thank you for helping with this.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
nibix: I have not found good resources on this so far, but I am still looking. I can hopefully update you later on this.

nibix:
> If that is indeed the case, then do you think it might be worthwhile to look into using cluster state() as an alternative for getting localNode reference, cause that seems to be available at almost all time (except during cluster bootstrap using it Zen Discover mechanism, although I'm not a 100% sure whether the request will flow through security with some cluster state info during bootstrap).

Yes, I think using cluster state will be the best solution in the current situation. 

Still, on a medium term, it might be recommendable to see whether it would be possible to get rid of the GuiceHolder hack and get access to the services provided by this on a better way. This probably needs a discussion with the core team.

DarshitChanpura:
@reta @dblock Thoughts? The problem at hand is to fetch localNode info for every transport request. There are two options:
1. using cluster state to retrieve localNode info. (seems to work all the time, with an AssertionError being thrown at bootstrap because it is expecting ClusterStateApplier as the calling class, but a catch block helps solve the problem)
3. using transport service via GuiceHolder. (doesn't work all the time)

For more details see the discussion above..

DarshitChanpura:
> @DarshitChanpura the local node is the one where the plugin is bootstrapped, right? if that's the case, I think the most straightforward way is for security plugin to implement `ClusterPlugin` as well, which has the hook `onNodeStarted()`, a well defined place where the local node could be queried for.

@reta SecurityPlugin already implements ClusterPlugin and so I tried to initialize a variable: `localNode = cs.localNode()` inside `onNodeStarted()` and then to access it in SecurityInterceptor. However node initialization fails with `Unable to start node` NPE due to ClusterState `cs` which suggests that we might have to go through a different route to get localNode info (may not be a straighforward implementation).

reta:
> However node initialization fails with `Unable to start node` NPE due to ClusterState `cs` which suggests that we might have to go through a different route to get localNode info (may not be a straighforward implementation).

This is super surprising,  could you access the local node using TransportService, not ClusterState ?

DarshitChanpura:
> > However node initialization fails with `Unable to start node` NPE due to ClusterState `cs` which suggests that we might have to go through a different route to get localNode info (may not be a straighforward implementation).
> 
> This is super surprising, could you access the local node using TransportService, not ClusterState ?

I couldn't. This is what I have currently implemented via GuiceHolder but it is not working. The only working solution I could get was via this:
```code
        try {
            isSameNodeRequest = cs.localNode().equals(connection.getNode()); // using DiscoveryNode equals comparison here
        } catch (AssertionError e) {
            // do nothing
            log.info(e);
        }
```

nibix:
> This is super surprising, could you access the local node using TransportService, not ClusterState ?

Unfortunately, plugins can only get ahold of TransportService via hacks which can cause other trouble. See https://github.com/opensearch-project/security/pull/2765#issuecomment-1561556458 for details.

It would be cool if some more straight-forward possibility could be created to get a reference to TransportService from plugin code.

DarshitChanpura:
@reta @nibix How should we proceed w.r.t this PR?

reta:
> @reta @nibix How should we proceed w.r.t this PR?

I most straightforward option I see is to consider enhancing `onNodeStarted` hook with `DiscoveryNode node`  argument - this (to me) looks like a missed contextual information for the hook from the beginning

DarshitChanpura:
> The most straightforward option I see is to consider enhancing `onNodeStarted` hook with `DiscoveryNode node` argument - this (to me) looks like a missed contextual information for the hook from the beginning

This will involve adding a new declaration in ClusterPlugin class in core and modifying `onNodeStarted` here to pass localNode info:
https://github.com/opensearch-project/OpenSearch/blob/277eb3d59fb9edcae8575577ec135032e3563f2e/server/src/main/java/org/opensearch/node/Node.java#L1345

This might be out of scope for this PR and might need a discussion in core repo. Thoughts @reta ?

DarshitChanpura:
@reta I have pushed the changes. Thank you for your valuable guidance!!

This PR's CI is now blocked until https://github.com/opensearch-project/OpenSearch/pull/7919 is merged and an artifact with this change becomes available.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -86,13 +87,13 @@ public class SecurityInterceptor {
...
 
     public SecurityInterceptor(final Settings settings,
-            final ThreadPool threadPool, final BackendRegistry backendRegistry,
-            final AuditLog auditLog, final PrincipalExtractor principalExtractor,
-            final InterClusterRequestEvaluator requestEvalProvider,
-            final ClusterService cs,
-            final SslExceptionHandler sslExceptionHandler,
-            final ClusterInfoHolder clusterInfoHolder,
-            final SSLConfig SSLConfig) {
+                               final ThreadPool threadPool, final BackendRegistry backendRegistry,
```
peternied, line 90: Could you revert the format changes, it makes it harder to see what is actually changed?

peternied (commented):
This is good, how are we measure the perf impact of this change?

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
DarshitChanpura: interesting, I didn;t make this change (I verified using git blame locally)

peternied:
> > This is good, how are we measure the perf impact of this change?
> 
> Great question. I was thinking about taking before and after running times of the test-suite. Any other suggestions are much appreciated.

Could you post those findings on this pull request so we know what to expect if this change is merged?

DarshitChanpura:
> @DarshitChanpura Can you resolve the conflicts on this PR? I see that the PR in core has been merged. After the conflicts are resolved is this PR good to finish review on?

Yes.

cwperks:
@DarshitChanpura Is the plugin install failure related to the change of the extension point in core?

> Task :compileJava
/home/runner/work/security/security/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java:1796: error: method does not override or implement a method from a supertype
    @Override

scrawfor99 (commented):
Looks good; can we add some tests to confirm handling of the states? You can mock the node behavior to flag it as being local and then confirm information from a fake context is stored transiently. 

It would be best to see that a request is handled the same regardless of whether it is resolved on the same or a different node with the exception of a performance boost on the local resolution. 

This goes to the comment above mentioning the loss of entries in the thread context.

cwperks:
@DarshitChanpura What is pending on this PR?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1876,6 +1879,14 @@ private static String handleKeyword(final String field) {
         return field;
     }
 
+    public static DiscoveryNode getLocalNode(){
+        return localNode;
+    }
+
+    public static void setLocalNode(DiscoveryNode node){
```
DarshitChanpura: need this for SecurityInterceptor tests.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1876,6 +1879,14 @@ private static String handleKeyword(final String field) {
         return field;
     }
 
+    public static DiscoveryNode getLocalNode() {
+        return localNode;
+    }
+
+    public static void setLocalNode(DiscoveryNode node) {
```
DarshitChanpura, line 1886: need this for security Interceptor tests

cwperks: Is it possible to add `localNode` to `SecurityInterceptor` constructor and remove this static method only needed for testing?

DarshitChanpura:
Looking into test-failures now.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
cwperks: FYI direct !== same node - @DarshitChanpura and I figured this out today after debugging the failing tests:

In OpenSearch, the channelType is a parameter that indicates the type of communication channel used for inter-node communication. There are two types of channels: direct and transport.

direct: This type of channel is used for communication between nodes within the same JVM (Java Virtual Machine). It's used for local node-to-node communication, which means it's used when the nodes are on the same machine or in the same OpenSearch process.

transport: This type of channel is for communication between nodes that are not in the same JVM. It's used for remote node-to-node communication, which means it's used when the nodes are on different machines or in different OpenSearch processes.

In summary, the direct channelType is used for local communication within the same JVM, while the transport channelType is used for remote communication between different JVMs.

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
```diff
@@ -142,7 +141,31 @@ protected void messageReceivedDecorate(
             }
 
             // bypass non-netty requests
-            if (channelType.equals("direct")) {
+            if (getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER) != null
```
cwperks, line 144: This logic in the receiver was simplified using @krishna-ggk 's suggestion here: https://github.com/opensearch-project/security/pull/2765#discussion_r1192128758

The receiver does not need any logic to determine where the request came from, if the transient headers are present then there is no need to deserialize. If the serialized (non-transient) headers are present then it will deserialize them.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -263,30 +267,49 @@ && getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_ORIGIN_HEADE
...
+        if (isSameNodeRequest) {
+            if (transportAddress != null) {
+                getThreadContext().putTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS, transportAddress);
+            }
 
-        if (userHeader == null) {
             if (origUser != null) {
-                getThreadContext().putHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER, Base64Helper.serializeObject(origUser));
+                // if request is going to be handled by same node, we directly put transient value as the thread context is not going to be
+                // stah.
```
scrawfor99, line 286:
```suggestion
                // stashed.
```

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
```diff
@@ -272,58 +284,10 @@ protected void messageReceivedDecorate(
 
                 // network intercluster request or cross search cluster request
                 // CS-SUPPRESS-SINGLE: RegexpSingleline Used to allow/disallow TLS connections to extensions
-                if (HeaderHelper.isInterClusterRequest(getThreadContext())
+                if (!(HeaderHelper.isInterClusterRequest(getThreadContext())
```
scrawfor99, line 287: We either need to update the comment above this or remove the exclamation point I think.

scrawfor99 (approved):
Looks good to me, one comment where I think we want to remove a comment from above some code you changed.

RyanL1997 (approved):

DarshitChanpura:
@hardik-k-shah @krishna-ggk @parasjain1 Could you please re-review this?

DarshitChanpura:
Merging now. Any comments can be addressed in follow-ups.

PR #2768, opened 2023-05-12
`deserializeSafeFromHeader` uses `context.getHeader(headerName)` instead of `context.getHeaders()`

### LINKED ISSUE #2757: [Performance Bug] `HeaderHelper.getSafeFromHeader` uses ThreadContext.getHeaders() causing major performance overhead
**What is the bug?**
Within `HeaderHelper.deserializeSafeFromHeader` we make a call to `HeaderHelper.getSafeFromHeader` which “safely” retrieves a header value from ThreadContext. For retrieving a header value for given header name `ThreadContext.getHeaders()` is being called which creates a new map for all existing headers and returns the same. This creation of new map every time `getSafeFromHeader` is called causes a major overhead, especially for search use cases.

Retrieve the header value from ThreadContext without creating a new map

**Additional Context & Proposed Solution**
Instead of using `ThreadContext.getHeaders()` we could use `ThreadContext.getHeader(key)`.

One caveat is that `ThreadContext.getHeader(key)` will return the current header value if present, else the default value. If default value is not present, it’ll return a `null` value. 

While in `HeaderHelper.getSafeFromHeader` we expect to retrieve the current header value OR `null` (not the default value).

I'm proposing the following changes to fix this -

Within OpenSearch -
* Add a method `ThreadContext.getHeaderOrDefault(key, defaultVal)` which will return the current header value if present, `defaultVal` otherwise.
* Optionally, can also add a ThreadContext.isHeaderPresent(key) method which returns true if header is present, false otherwise.

Within Security Plugin -
* Replace call to `ThreadContext.getHeaders()` with `ThreadContext.getHeaderOrDefault(headerName, null)` to avoid unnecessarily creating an additional map

 - OS1.3
 - Security Plugin

parasjain1:
**Description**
Category:  Performance Bug Fix

`HeaderHelper.deserializeSafeFromHeader` is invoked O(segment) times during every search request and introduced a huge overhead. This optimisation has resulted in good gains with respect to search throughput and latency in our experiments.

**Issues Resolved**
* Resolves #2757 

### Testing
* Current test suite should pass.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/support/HeaderHelper.java
+++ b/src/main/java/org/opensearch/security/support/HeaderHelper.java
@@ -57,15 +57,8 @@ public class HeaderHelper {
             return null;
         }
 
-        String headerValue = null;
-        	
-        Map<String, String> headers = context.getHeaders();
-        if (!headers.containsKey(headerName) || (headerValue = headers.get(headerName)) == null) {
-            return null;
-        }
-
         if (isInterClusterRequest(context) || isTrustedClusterRequest(context) || isDirectRequest(context)) {
-            return headerValue;
+            return context.getHeader(headerName);
         }
 
         return null;
--- a/src/main/java/org/opensearch/security/support/HeaderHelper.java
+++ b/src/main/java/org/opensearch/security/support/HeaderHelper.java
@@ -27,7 +27,6 @@
 package org.opensearch.security.support;
 
 import java.io.Serializable;
-import java.util.Map;
 
 import com.google.common.base.Strings;
 


src/main/java/org/opensearch/security/support/HeaderHelper.java
```diff
@@ -57,15 +57,8 @@ public static String getSafeFromHeader(final ThreadContext context, final String
...
-        	
-        Map<String, String> headers = context.getHeaders();
-        if (!headers.containsKey(headerName) || (headerValue = headers.get(headerName)) == null) {
-            return null;
-        }
-
         if (isInterClusterRequest(context) || isTrustedClusterRequest(context) || isDirectRequest(context)) {
-            return headerValue;
+            return context.getHeader(headerName);
         }
```
krishna-ggk, line 63: It is surprising that this small change makes a sizeable difference.

krishna-ggk (commented):
Changes are good - can you share the experiment results?

src/main/java/org/opensearch/security/support/HeaderHelper.java
parasjain1: > Changes are good - can you share the experiment results?

* HeaderHelper.deserializeSafeFromHeader ‘s overall CPU usage dropped to 0.52% from 5%-12.5% seen across multiple CPU profiles obtained for prior runs.
* Search Latency saw a drop of ~75%, Search Throughput increased by ~175%
* Indexing Latency dropped by  70%, throughput increased by more than ~50%.

cwperks:
@parasjain1 Can you sign the commits?

cwperks:
Looking into the issue with plugin install. I suspect it could be related to this PR: https://github.com/opensearch-project/OpenSearch/pull/7465

parasjain1:
> @parasjain1 Can you sign the commits?

Done.

DarshitChanpura (approved):
Great fix @parasjain1 !!

krishna-ggk (approved):

cwperks (approved):

peternied (approved):

PR #2786, opened 2023-05-23
Add backend filtering for legacy internal user and service accounts

### LINKED ISSUE #2704: [Extensions] Filter internal users for Service Accounts and add associated API
As a follow-up to #2594, a filtering option can be added to the existing internal users functionality. The filtering should allow a client to fetch just true internal user (not service) accounts or just service accounts. This filtering should be applied to the backend as well as the internal users list capabilities of the frontend code. An API route will need to be added to the `handleGet` method of the `internalUsersAPI` since there is not currently group filtering functionality. 

The backend logic can look something like this inside the `UserService`: 
```
public static List<String> listServiceAccounts() {

        final SecurityDynamicConfiguration<?> internalUsersConfiguration = load(getConfigName(), false);

        List<String> serviceAccounts = new ArrayList<>();
        for (Map.Entry<String, ?> entry : internalUsersConfiguration.getCEntries().entrySet()) {

            final InternalUserV7 internalUserEntry = (InternalUserV7) entry.getValue();
            final Map accountAttributes = internalUserEntry.getAttributes();
            final String accountName = entry.getKey();
            if (accountAttributes.containsKey("service") && accountAttributes.get("service") == "true") {
                serviceAccounts.add(accountName);
            }
        }
        return serviceAccounts;
    }

    public static List<String> listInternalUsers() {

        final SecurityDynamicConfiguration<?> internalUsersConfiguration = load(getConfigName(), false);

        List<String> internalUserAccounts = new ArrayList<>();
        for (Map.Entry<String, ?> entry : internalUsersConfiguration.getCEntries().entrySet()) {

            final InternalUserV7 internalUserEntry = (InternalUserV7) entry.getValue();
            final Map accountAttributes = internalUserEntry.getAttributes();
            final String accountName = entry.getKey();
            if (!accountAttributes.containsKey("service") || accountAttributes.get("service") == "false") {
                internalUserAccounts.add(accountName);
            }
        }
        return internalUserAccounts;
    }

    public static Set<String> listUserAccounts() {

        final SecurityDynamicConfiguration<?> internalUsersConfiguration = load(getConfigName(), false);
        return internalUsersConfiguration.getCEntries().keySet();
    }
```

For the frontend logic, the PR will need to extend the existing code that fetches the internal users and displays it as part of the Security tab. 

Completion of this issue should look like two PR:

One in the backend code base that introduces the filtering capabilities shown above as well as the associated API routes.
One in the frontend code base that updates the internal users tab to allow for filtering the shown users based on whether they are service accounts or not.

samuelcostae:
### Description
Update UserAPIAction to Filter internal accounts and Service accounts.  

Dashboards Plugin  changes via PR https://github.com/opensearch-project/security-dashboards-plugin/pull/1502

### Issues Resolved
- Resolves #2704 

### Testing
Unit Tests created 

### Check List


src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -55,21 +59,17 @@ public class InternalUsersApiAction extends PatchableResourceApiAction {
...
             new Route(Method.GET, "/user/{name}"),
             new Route(Method.GET, "/user/"),
+            new Route(Method.GET, "/user/serviceaccounts"),
+            new Route(Method.GET, "/user/internalaccounts"),
             new Route(Method.POST, "/user/{name}/authtoken"),
             new Route(Method.DELETE, "/user/{name}"),
-            new Route(Method.PUT, "/user/{name}"),
-
-            // corrected mapping, introduced in OpenSearch Security
-            new Route(Method.GET, "/internalusers/{name}"),
```
DarshitChanpura, line 66: why remove these new routes?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -100,6 +100,43 @@ public List<Route> routes() {
...
+            userService.listInternalUsers();
+            successResponse(channel, configuration);
+            return;
+
+            }
+        if (request.rawPath().equalsIgnoreCase("/_plugins/_security/api/user/serviceaccounts")){
+            userService.listServiceAccounts();
+            successResponse(channel, configuration);
+            return;
+         }
```
DarshitChanpura, line 120:
```suggestion
        String destination = request.rawPath().split("/")[5];

        if (destination.equalsIgnoreCase("internalaccounts")) {
            userService.listInternalUsers();
            successResponse(channel, configuration);
            return;
        } else if (destination.equalsIgnoreCase("serviceaccounts")) {
            userService.listServiceAccounts();
            successResponse(channel, configuration);
            return;
        } 
```

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -276,4 +274,51 @@ public static void saveAndUpdateConfigs(final String indexName, final Client cli
...
+
+        List<String> serviceAccounts = new ArrayList<>();
+        for (Map.Entry<String, ?> entry : internalUsersConfiguration.getCEntries().entrySet()) {
+
+            final InternalUserV7 internalUserEntry = (InternalUserV7) entry.getValue();
+            final Map accountAttributes = internalUserEntry.getAttributes();
+            final String accountName = entry.getKey();
+            if (accountAttributes.containsKey("service") && accountAttributes.get("service") == "true") {
+                serviceAccounts.add(accountName);
+            }
```
DarshitChanpura, line 290: You can rewrite this as:
```suggestion
        if (accountAttributes.getOrDefault("service", "false").equals("true")) {
                serviceAccounts.add(accountName);
            }
```
for better performance and readability.

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -276,4 +274,51 @@ public static void saveAndUpdateConfigs(final String indexName, final Client cli
...
+
+        List<String> internalUserAccounts = new ArrayList<>();
+        for (Map.Entry<String, ?> entry : internalUsersConfiguration.getCEntries().entrySet()) {
+
+            final InternalUserV7 internalUserEntry = (InternalUserV7) entry.getValue();
+            final Map accountAttributes = internalUserEntry.getAttributes();
+            final String accountName = entry.getKey();
+            if (!accountAttributes.containsKey("service") || accountAttributes.get("service") == "false") {
+                internalUserAccounts.add(accountName);
+            }
```
DarshitChanpura, line 307: similar to the comment above:
```suggestion
        if (accountAttributes.getOrDefault("service", "false").equals("false")) {
                serviceAccounts.add(accountName);
            }
```

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -276,4 +274,51 @@ public static void saveAndUpdateConfigs(final String indexName, final Client cli
...
+        }
+        return internalUserAccounts;
+    }
+
+    public static Set<String> listUserAccounts() {
```
DarshitChanpura, line 312: Is this method used? if not, will it be used in future?

nit: `listAllAccounts`

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
DarshitChanpura: okay. Since we are adding two new routes for `/users`, we should add them here for `/internalusers` too

samuelcostae: I corrected it. 

As it is a new endpoint, I left the ithe it onlyt only with the new mapping(/internalusers), as retrocompatibility wouldnt be an issue

samuelcostae:
@DarshitChanpura I had to a good bit of refactoring as I noticed that before, the beahviour in the PR was returning only the list of names of the accounts of the specified type, but the existing implementations on internalUsersAPI return the full account details.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -100,6 +107,39 @@ public List<Route> routes() {
...
         return Endpoint.INTERNALUSERS;
     }
+    @Override
+    protected void handleGet(final RestChannel channel, RestRequest request, Client client, final JsonNode content) throws IOException{
+        final String resourcename = request.param("name");
+
+        SecurityDynamicConfiguration<?> configuration = load(getConfigName(), true);
+        filter(configuration);
+
+        String requestDestination = request.rawPath().split("/api")[1];
```
samuelcostae, line 117: chnaged to split on /api instead of a spefic slash, as different requests could have different number slashes

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -276,4 +279,35 @@ public static void saveAndUpdateConfigs(final String indexName, final Client cli
             throw ExceptionsHelper.convertToOpenSearchException(e);
         }
     }
+
+    public void removeNonServiceAccounts(SecurityDynamicConfiguration<?> configuration) {
```
samuelcostae, line 283: Due to  the refactoring, I changed the method names to better describe their new behaviour

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -136,6 +138,38 @@ public void testParallelPutRequests() throws Exception {
         deleteUser("test1");
     }
 
+    @Test
```
samuelcostae, line 141: In general, a separate test case for each of the checks would be more appropriate, but as added them all in one test case as that is how the other tests on the class are setup and I'm not sure if there is some underlying reason for that.

scrawfor99: There is no particular reason that I am aware of. I believe an initial choice was made and then everyone has copied the pattern since. If you are able to make individual tests that would be excellent.

scrawfor99 (commented):
Overall looks good to me. I left a comment about tests, but I believe this should ready for final review.

scrawfor99:
@samuelcostae could you rebase against main and resolve any conflicts.

DarshitChanpura:
@samuelcostae Please let us know if you would require any help to move this along.

samuelcostae:
> @samuelcostae Please let us know if you would require any help to move this along.

I believe the only step missing now (not counting the front end) would be documentation related. Do we he have any resources regarding conventions, what to include/not include, etc?

peternied:
@scrawfor99 @DarshitChanpura mind taking another look at this change?

scrawfor99 (approved):
Looks good to me. Thanks @samuelcostae!

scrawfor99:
Before merging, we will need to get the CI passing. Looks like a rebase and resolving of a mapped reference will be required.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -66,6 +71,8 @@ public class InternalUsersApiAction extends PatchableResourceApiAction {
             // corrected mapping, introduced in OpenSearch Security
             new Route(Method.GET, "/internalusers/{name}"),
             new Route(Method.GET, "/internalusers/"),
+            new Route(Method.GET, SERVICE_ACCOUNTS_ENDPOINT),
+            new Route(Method.GET, INTERNAL_ACCOUNTS_ENDPOINT),
```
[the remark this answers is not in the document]
samuelcostae: I didn't want to remove/change any existing endpoints   to avoid backward compatibility issues and because the issue description mentioned "adding endpoints for those". 

_As a follow-up, we can a modify /internalusers handler to check for a query parameter called filter_by and it can have either of two values service-accounts & internal-accounts. This way we avoid creating new endpoints._


If editing the existing endpoint is ok, i think this makes sense

DarshitChanpura:
@samuelcostae Would you please address the merge conflicts?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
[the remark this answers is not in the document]
samuelcostae: updated to use the existing endpoint. Will update the Dashboard plugins accordingly

samuelcostae:
> @samuelcostae Would you please address the merge conflicts?

Done. It was just the order of imports to the fixes to match core changes that i also had done on this branch

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -115,6 +124,32 @@ protected Endpoint getEndpoint() {
...
+        filter(configuration);
+
+        String filterBy = request.param("filterBy", "all");
+
+        if (filterBy != "internal" && filterBy != "service") {
+            userService.filterAccountsByType(configuration, filterBy);
+        }
+
+        // no specific resource requested, return complete config
+        if (resourcename == null || resourcename.length() == 0) {
```
DarshitChanpura, line 141: can leverage `Strings.isNullOrEmpty(resourceName)` from core

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -314,4 +317,24 @@ public static void saveAndUpdateConfigs(
             throw ExceptionsHelper.convertToOpenSearchException(e);
         }
     }
+
+    public void filterAccountsByType(SecurityDynamicConfiguration<?> configuration, String requestedAccountType) {
```
DarshitChanpura, line 321: seems like this method removes:
- internal accounts with service flag as true
- service accounts with service flag as false

If that is the case, can we rename it as `removeInvalidAccounts`

Also can you add javadoc for this one.

DarshitChanpura (dismissed):
Looks good overall. Last couple of comments from my end.

scrawfor99 (approved):
Looks good to me. Thanks Sam.

DarshitChanpura (dismissed):
LGTM

samuelcostae:
@RyanL1997  or @peternied  Could you review this please?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -115,6 +125,32 @@ protected Endpoint getEndpoint() {
...
     }
 
+    @Override
+    protected void handleGet(final RestChannel channel, RestRequest request, Client client, final JsonNode content) throws IOException {
+        final String resourcename = request.param("name");
+
+        SecurityDynamicConfiguration<?> configuration = load(getConfigName(), true);
+        filter(configuration);
+
+        String filterBy = request.param("filterBy", "all");
```
peternied, line 135: Default value of all, e.g. no results?  IMO this should be nullable

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -115,6 +125,32 @@ protected Endpoint getEndpoint() {
...
+
+        if (filterBy.equalsIgnoreCase("internal") || filterBy.equalsIgnoreCase("service")) {
```
peternied, line 137: Move these types of filter into an enum, these are public API members after this code ships so we need to be sure they won't break.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -115,6 +125,32 @@ protected Endpoint getEndpoint() {
...
+            userService.filterAccountsByType(configuration, filterBy);
```
peternied, line 138: If an unsupported filter type is found, an error should be returned indicating the allowed values otherwise you won't know if you used the term `users` and it didn't return any results.

src/main/java/org/opensearch/security/securityconf/impl/SecurityDynamicConfiguration.java
```diff
@@ -285,7 +286,11 @@ public SecurityDynamicConfiguration<T> deepClone() {
     @JsonIgnore
     public void remove(String key) {
         centries.remove(key);
+    }
 
+    @JsonIgnore
+    public void remove(List<String> keySet) {
+        keySet.stream().forEach(this::remove);
```
peternied, line 293: Nit: Seems unneeded, its only used in 2 places right new to one another

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -314,4 +317,32 @@ public static void saveAndUpdateConfigs(
...
+     *
+     * @param configuration SecurityDynamicConfiguration object containing all accounts
+     * @param requestedAccountType The type of account to be kept. Should be "service" or "internal"
+     *
+     */
+    public void filterAccountsByType(SecurityDynamicConfiguration<?> configuration, String requestedAccountType) {
+        List<String> toBeRemoved = new ArrayList<>();
+
+        for (Map.Entry<String, ?> entry : configuration.getCEntries().entrySet()) {
+            final InternalUserV7 internalUserEntry = (InternalUserV7) entry.getValue();
```
peternied, line 335: Since you are casting to InternalUserV7, there is a property `boolean service()`, can we use that instead?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
samuelcostae: FilterBy "service" will return Service Accounts
FilterBy "Internal" will return Internal Accounts
The default "All" will return All Accounts

peternied:
@samuelcostae I've seen responses on some comments on but I didn't see any code pushed.

src/main/java/org/opensearch/security/user/UserService.java
scrawfor99: You can use the field. 

The `service` field is used when called with the internal users api.

For example:

```
    try {
            if (request.hasParam("service")) {
                ((ObjectNode) content).put("service", request.param("service"));
            }
            if (request.hasParam("enabled")) {
                ((ObjectNode) content).put("enabled", request.param("enabled"));
            }
            ((ObjectNode) content).put("name", username);
            internalUsersConfiguration = userService.createOrUpdateAccount((ObjectNode) content);
```

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
samuelcostae: I made the change, but wouldn't "Include" be read as "add" and cause confusion as well?

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -337,9 +342,9 @@ public void filterAccountsByType(SecurityDynamicConfiguration<?> configuration,
             final String accountName = entry.getKey();
             boolean isServiceAccount = Boolean.parseBoolean(accountAttributes.getOrDefault("service", "false").toString());
 
-            if (requestedAccountType.equalsIgnoreCase("internal") && isServiceAccount) {
+            if (requestedAccountType.equalsIgnoreCase(AccountType.INTERNAL.name()) && isServiceAccount) {
```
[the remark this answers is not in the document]
samuelcostae: Do you mean doing this instead? 
`        if (requestedAccountType == AccountType.SERVICE) {
            for (Map.Entry<String, ?> entry : configuration.getCEntries().entrySet()) {
                final InternalUserV7 internalUserEntry = (InternalUserV7) entry.getValue();
                final Map accountAttributes = internalUserEntry.getAttributes();
                final String accountName = entry.getKey();
                boolean isServiceAccount = Boolean.parseBoolean(accountAttributes.getOrDefault("service", "false").toString());

                if (isServiceAccount == false) {
                    toBeRemoved.add(accountName);
                }
            }
        }
        else if (requestedAccountType == AccountType.INTERNAL) {

            for (Map.Entry<String, ?> entry : configuration.getCEntries().entrySet()) {
                final InternalUserV7 internalUserEntry = (InternalUserV7) entry.getValue();
                final Map accountAttributes = internalUserEntry.getAttributes();
                final String accountName = entry.getKey();
                boolean isServiceAccount = Boolean.parseBoolean(accountAttributes.getOrDefault("service", "false").toString());

                if (isServiceAccount == true) {
                    toBeRemoved.add(accountName);
                }
            }
        }
        configuration.remove(toBeRemoved);
        }`

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -52,10 +54,19 @@
 import static org.opensearch.security.dlic.rest.support.Utils.hash;
 
 public class InternalUsersApiAction extends PatchableResourceApiAction {
```
peternied, line 56: An extensive refactor [1] to how classes are declared and default behaviors was merged recently, you'll need to resolve these conflicts before this can be reviewed.

- [1] https://github.com/opensearch-project/security/pull/3123

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -314,4 +317,49 @@ public static void saveAndUpdateConfigs(
...
+     *
+     * @param configuration SecurityDynamicConfiguration object containing all accounts
+     * @param requestedAccountType The type of account to be kept. Should be "service" or "internal"
+     *
+     */
+    public void includeAccountsIfType(SecurityDynamicConfiguration<?> configuration, AccountType requestedAccountType)
+        throws UserServiceException {
+        if (requestedAccountType != AccountType.INTERNAL && requestedAccountType != AccountType.SERVICE) {
+            throw new UserServiceException("Invalid user type {} " + requestedAccountType);
+        }
```
peternied, line 335: The enum type should be used to determine if the values are correct, instead of having the caller check if the enum usage is correct.

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -314,4 +317,49 @@ public static void saveAndUpdateConfigs(
...
+     *
+     * Accounts with the 'service' attribute set to true, are considered of type 'service'.
+     * Accounts with the 'service' attribute set to false or without the 'service' attribute, are considered of type 'internal'.
+     *
+     * @param configuration SecurityDynamicConfiguration object containing all accounts
+     * @param requestedAccountType The type of account to be kept. Should be "service" or "internal"
+     *
+     */
+    public void includeAccountsIfType(SecurityDynamicConfiguration<?> configuration, AccountType requestedAccountType)
+        throws UserServiceException {
```
peternied, line 332: I don't think this exception is required, exceptions as part of normal flow of control are frowned upon, lets removed https://stackoverflow.com/questions/729379/why-not-use-exceptions-as-regular-flow-of-control

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -314,4 +317,49 @@ public static void saveAndUpdateConfigs(
...
+                final Map accountAttributes = internalUserEntry.getAttributes();
+                final String accountName = entry.getKey();
+                final boolean isServiceAccount = Boolean.parseBoolean(accountAttributes.getOrDefault("service", "false").toString());
+
+                if (isServiceAccount == true) {
+                    toBeRemoved.add(accountName);
+                }
+            }
+        }
+        configuration.remove(toBeRemoved);
```
peternied, line 363: This method body has larges amounts of duplicate logic, lets see about improving it.  I had ChatGPT rewrite this code, but it can be condensed even more with logic abstracted which will make it easier to read and more obvious if there are bugs.

```
        final List<String> toBeRemoved = configuration.getCEntries().entrySet().stream()
            .filter(entry -> {
                InternalUserV7 internalUserEntry = (InternalUserV7) entry.getValue();
                Map<String, Object> accountAttributes = internalUserEntry.getAttributes();
                boolean isServiceAccount = Boolean.parseBoolean(accountAttributes.getOrDefault("service", "false").toString());
                return (requestedAccountType == AccountType.SERVICE && !isServiceAccount) ||
                       (requestedAccountType == AccountType.INTERNAL && isServiceAccount);
            })
            .map(Map.Entry::getKey)
            .collect(Collectors.toList());
        
        configuration.remove(toBeRemoved);
        ```

File: src/main/java/org/opensearch/security/user/UserService.java [hunk 22]
```diff
@@ -314,4 +317,49 @@ public static void saveAndUpdateConfigs(
...
+     * Removes accounts that are not of the requested type from the SecurityDynamicConfiguration object passed.
+     *
+     * Accounts with the 'service' attribute set to true, are considered of type 'service'.
+     * Accounts with the 'service' attribute set to false or without the 'service' attribute, are considered of type 'internal'.
+     *
+     * @param configuration SecurityDynamicConfiguration object containing all accounts
+     * @param requestedAccountType The type of account to be kept. Should be "service" or "internal"
+     *
+     */
+    public void includeAccountsIfType(SecurityDynamicConfiguration<?> configuration, AccountType requestedAccountType)
```
peternied, line 331: Please author unit tests for this file.

src/main/java/org/opensearch/security/user/AccountType.java
```diff
@@ -0,0 +1,6 @@
+package org.opensearch.security.user;
+
+public enum AccountType {
```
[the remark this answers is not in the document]
peternied: Please add javadocs

src/main/java/org/opensearch/security/user/AccountType.java
```diff
@@ -0,0 +1,26 @@
...
+        return this.name;
+    }
+    public static AccountType fromString(String name) {
+        for (AccountType b : AccountType.values()) {
+            if (b.name.equalsIgnoreCase(name)) {
+                return b;
+            }
+        }
+        return null;
+    }
```
peternied, line 24: Its strange to get a null when attempting to resolve an enum from a value, either throw an error, or make the return type Optional<AccountType> so callers can handle the case where the value wasn't parsed explicitly

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -115,12 +126,42 @@ protected Endpoint getEndpoint() {
         return Endpoint.INTERNALUSERS;
     }
 
+    @Override
+    protected void handleGet(final RestChannel channel, RestRequest request, Client client, final JsonNode content) throws IOException {
```
peternied, line 130: This method doesn't seem to have much of an order, please declare values closer to when they are used, collapse logic into a few layers as possible.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -115,12 +126,42 @@ protected Endpoint getEndpoint() {
...
+        filter(configuration);
+
+        String includeIfType = request.param("filterBy", "any");
+
+        if (includeIfType != "any") {
+            try {
+                userService.includeAccountsIfType(configuration, AccountType.valueOf(includeIfType.toUpperCase()));
+            } catch (UserServiceException ex) {
+                badRequestResponse(channel, ex.getMessage());
+            }
```
peternied, line 143: Lets avoid using exceptions as flow of control, the only invalid case is when the account type isn't known AFAIK, maybe we could change the method to accept `null` or `Optional.none()` and not have to special case.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -115,12 +126,42 @@ protected Endpoint getEndpoint() {
...
+        // no specific resource requested, return complete config
+        if (Strings.isNullOrEmpty(resourcename)) {
+            successResponse(channel, configuration);
+            return;
+        }
+        if (!configuration.exists(resourcename)) {
+            notFound(channel, "Resource '" + resourcename + "' not found.");
+            return;
+        }
+        configuration.removeOthers(resourcename);
```
peternied, line 155: Its strange that there are two different times when filtering is applied on this request's output, can this be better reconciled?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -52,10 +54,19 @@
...
+        request.param("name");
+        request.param("filterBy");
+    }
+
     static final List<String> RESTRICTED_FROM_USERNAME = ImmutableList.of(
         ":" // Not allowed in basic auth, see https://stackoverflow.com/a/33391003/533057
     );
 
+    public static final String LEGACY_OPENDISTRO_PREFIX = "_opendistro/_security";
+    public static final String PLUGINS_PREFIX = "_plugins/_security";
```
peternied, line 69: Unused?

src/test/java/org/opensearch/security/test/helper/rest/RestHelper.java
```diff
@@ -527,7 +527,7 @@ private void throwIfNotValueNode(final JsonNode node) {
             }
         }
 
-        private JsonNode findObjectInJson(final String jsonDotPath) {
+        public JsonNode findObjectInJson(final String jsonDotPath) {
```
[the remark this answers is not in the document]
peternied: Within test code they shouldn't be interacting with the format of the response directly, but instead verifying lists, strings, or numeric values are what is expected, by returning a non-primitive type to the tests it makes them more complex to author and read.

peternied:
@samuelcostae Thanks for the updates - sorry its taken a little bit to circle back around to this pull requests.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
samuelcostae: Unresolving it, taking longer than I expected. Having difficulties as tests are timing out for some reason

peternied: While you are working on the tests could you push your latest changes so they can be looked at while you troubleshoot?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
samuelcostae: Should it return everything or nothing if that is case?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
samuelcostae: This should become 'outdated' with the abstractApiAction changes from the refactor, but i believe it makes sense.  If someone Request JhonDoe's account (Which is an internal account) and ask to filter by service accounts,  it makes sense to me that it the account would not be displayed.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
scrawfor99: If nothing is specified, you can return all. If an invalid account type is provided just return nothing. I am not opposed to throwing the exception personally, but the benefit of applying the filter always is that it would be more future proof. 

I.e. you filter for "purple," you get nothing because we don't have purple accounts. But in the future we could and the call would already work. 

Empty filter = everything.

scrawfor99:
This is associated with #2704

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -121,6 +135,21 @@ private void internalUsersApiRequestHandlers(RequestHandler.RequestHandlersBuild
...
+        userService.includeAccountsIfType(users, AccountType.fromString(userType));
+        return ValidationResult.success(SecurityConfiguration.of(users.getCType().toString(), users));
+
+    }
+
+    protected final String filterParam(final RestRequest request) {
+        final String name = request.param("filterBy");
+        if (Strings.isNullOrEmpty(name)) {
+            return "any";
+        }
```
peternied, line 151: Why not using the enum?

src/main/java/org/opensearch/security/user/AccountType.java
```diff
@@ -0,0 +1,26 @@
+package org.opensearch.security.user;
+
+public enum AccountType {
+
+    INTERNAL("internal"),
+    SERVICE("service");
```
peternied, line 6: I think this is more useful if it represents the 'user filter types', so you would include a 'no filter' value which could be the default when using `fromString(..)`

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
samuelcostae: Yep, it goes with your other comment, the enum represents the account type. I will change the enum to be  "UserFilterType"

src/main/java/org/opensearch/security/user/AccountType.java
samuelcostae: As I changed the enum to represent the filter, i will return "ANY" instead of null

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -314,4 +317,47 @@ public static void saveAndUpdateConfigs(
...
+                final String accountName = entry.getKey();
+                final boolean isServiceAccount = Boolean.parseBoolean(accountAttributes.getOrDefault("service", "false").toString());
+
+                if (isServiceAccount == true) {
+                    toBeRemoved.add(accountName);
+                }
+            }
+        }
+        configuration.remove(toBeRemoved);
+    }
```
DarshitChanpura, line 362: Opportunity to clean up some re-usable code:

```suggestion
    public void includeAccountsIfType(SecurityDynamicConfiguration<?> configuration, UserFilterType requestedAccountType) {
        if (requestedAccountType != UserFilterType.INTERNAL && requestedAccountType != UserFilterType.SERVICE) {
            return;
        }
        List<String> toBeRemoved = new ArrayList<>();

        if (requestedAccountType == UserFilterType.SERVICE) {
            accountToRemoveFromConfig(configuration, toBeRemoved, false);
        } else if (requestedAccountType == UserFilterType.INTERNAL) {
            accountToRemoveFromConfig(configuration, toBeRemoved, true);
        }
        configuration.remove(toBeRemoved);
    }
    
    private void accountsToRemoveFromConfiguration(SecurityDynamicConfiguration<?> configuration, ArrayList<String> toBeRemoved, boolean isServiceAccountRequested) {
        for (Map.Entry<String, ?> entry : configuration.getCEntries().entrySet()) {
                final InternalUserV7 internalUserEntry = (InternalUserV7) entry.getValue();
                final Map accountAttributes = internalUserEntry.getAttributes();
                final String accountName = entry.getKey();
                final boolean isServiceAccount = Boolean.parseBoolean(accountAttributes.getOrDefault("service", "false").toString());

                if (isServiceAccount == isServiceAccountRequested) {
                    toBeRemoved.add(accountName);
                }
            }
    }
```

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -144,6 +148,55 @@ public void testParallelPutRequests() throws Exception {
...
+        response = rh.executeGetRequest(ENDPOINT + "/internalusers?filterBy=internal");
+
+        Assert.assertEquals(response.getBody(), HttpStatus.SC_OK, response.getStatusCode());
+        JsonNode list = DefaultObjectMapper.readTree(response.getBody());
+        Assert.assertEquals(INTERNAL_ACCOUNTS_IN_SETTINGS, list.size());
+
+        response = rh.executeGetRequest(ENDPOINT + "/internalusers?filterBy=service");
+        Assert.assertEquals(response.getBody(), HttpStatus.SC_OK, response.getStatusCode());
+        list = DefaultObjectMapper.readTree(response.getBody());
+        assertThat(list, Matchers.emptyIterable());
```
DarshitChanpura, line 170: Shouldn't this be 1 (SERVICE_ACCOUNTS_IN_SETTINGS)?

DarshitChanpura (commented):
Looks close. Final two comments.

src/main/java/org/opensearch/security/user/UserService.java
samuelcostae: I agree that code replication in general is not good, but the alternative in that case would be moving the 4 repeated lines into a new method, adding (imo) unnecessary complexity and a very 'meagre' method. 

The functional approach pasted above  has the same structure that I had done initially, with the RequestedAccountType being checked in every entry.

Can you confirm if we should move back to this initial approach?

Thanks

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
samuelcostae: That assert being made before the service account entry is created. 

It was suggested at some point to check the assertion  befo.re and after creating the entry . Line 185 has the `        Assert.assertEquals(SERVICE_ACCOUNTS_IN_SETTINGS, list.size());`

peternied (commented):
Last blocker in my mind is the structure of the UserService code, please iterate on @DarshitChanpura suggestion to slim down that code.

src/main/java/org/opensearch/security/user/UserService.java
samuelcostae: Im working on it, but what is the best approach to create SecurityDynamicConfiguration from the yml userfile?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -272,6 +301,7 @@ public Map<String, RequestContentValidator.DataType> allowedKeys() {
                             .put("opendistro_security_roles", DataType.ARRAY)
                             .put("hash", DataType.STRING)
                             .put("password", DataType.STRING)
+                            .put("service", DataType.STRING)
```
peternied, line 304: Maybe I missed this before, why is this needed? Service is part of the attributes field, not a top level field on the document

src/main/java/org/opensearch/security/user/UserFilterType.java
```diff
@@ -0,0 +1,39 @@
...
+    INTERNAL("internal"),
+    SERVICE("service");
+
+    private String name;
+
+    UserFilterType(String name) {
+        this.name = name;
+    }
+
+    public String getName() {
```
peternied, line 26: Nit; Looks like this getter is not use by any of the code, lets remove it to keep the codebase / change smaller

src/main/java/org/opensearch/security/user/UserService.java
peternied: A good reference for this is kind of test RestLayerPrivilegesEvaluatorTest.java, so there is no need to interact with any yml - instead there is only mocked version of the configuration that does exactly what is needed for the test case.

src/main/java/org/opensearch/security/user/UserService.java
peternied: With the most recent revision this is much more improved, good work! While I think it could still be even better - lets leave that to a future exercise if you are interested, but not block this review.  Progress not perfection!

src/main/java/org/opensearch/security/user/UserFilterType.java
peternied: Nit: optional if you'd like, feel free to close without resolving if you'd prefer not.

src/test/java/org/opensearch/security/UserServiceUnitTests.java
```diff
@@ -0,0 +1,78 @@
...
+        Assert.assertEquals(SERVICE_ACCOUNTS_IN_SETTINGS, config.getCEntries().size());
+
+        config = readConfigFromYml(usersYmlFile, CType.INTERNALUSERS);
+
+        userService.includeAccountsIfType(config, UserFilterType.INTERNAL);
+        Assert.assertEquals(INTERNAL_ACCOUNTS_IN_SETTINGS, config.getCEntries().size());
+
+    }
+
+    private SecurityDynamicConfiguration<?> readConfigFromYml(String file, CType cType) throws Exception {
```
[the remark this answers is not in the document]
peternied: I know I also posted a recommendation to use mocks.  If this works, it works :D

src/test/java/org/opensearch/security/UserServiceUnitTests.java
```diff
@@ -0,0 +1,78 @@
...
+        UserService userService = new UserService(clusterService, configurationRepository, builder.build(), client);
+        config = readConfigFromYml(usersYmlFile, CType.INTERNALUSERS);
+
+        final int SERVICE_ACCOUNTS_IN_SETTINGS = 0;
+        final int INTERNAL_ACCOUNTS_IN_SETTINGS = 67;
+
+        userService.includeAccountsIfType(config, UserFilterType.SERVICE);
+        Assert.assertEquals(SERVICE_ACCOUNTS_IN_SETTINGS, config.getCEntries().size());
+
+        config = readConfigFromYml(usersYmlFile, CType.INTERNALUSERS);
```
peternied, line 56: Please separate the tests into patterns like the following, in this case, there are multiple 'actions' in a single method
```java
// Setup
// Action
// Verification
```

src/test/java/org/opensearch/security/UserServiceUnitTests.java
```diff
@@ -0,0 +1,78 @@
...
+    @Mock
+    Client client;
+
+    @Test
+    public void testUserTypeFilter() throws Exception {
+
+        String usersYmlFile = "./internal_users.yml";
+        Settings.Builder builder = Settings.builder();
+        UserService userService = new UserService(clusterService, configurationRepository, builder.build(), client);
+        config = readConfigFromYml(usersYmlFile, CType.INTERNALUSERS);
```
peternied, line 48: Move this into a setup method so its done before each test method executes.

src/test/java/org/opensearch/security/UserServiceUnitTests.java
```diff
@@ -0,0 +1,78 @@
...
+        final int SERVICE_ACCOUNTS_IN_SETTINGS = 0;
+        final int INTERNAL_ACCOUNTS_IN_SETTINGS = 67;
+
+        userService.includeAccountsIfType(config, UserFilterType.SERVICE);
+        Assert.assertEquals(SERVICE_ACCOUNTS_IN_SETTINGS, config.getCEntries().size());
+
+        config = readConfigFromYml(usersYmlFile, CType.INTERNALUSERS);
+
+        userService.includeAccountsIfType(config, UserFilterType.INTERNAL);
+        Assert.assertEquals(INTERNAL_ACCOUNTS_IN_SETTINGS, config.getCEntries().size());
```
peternied, line 59: Since a consistent data set is being used, I'd recommend at least 3 assertThat(..) statements, the resulting output is an expected size (already done), and some known users are contained in the output, and some known users are not contained.

src/test/java/org/opensearch/security/UserServiceUnitTests.java
```diff
@@ -0,0 +1,78 @@
...
+    SecurityDynamicConfiguration<?> config;
+    @Mock
+    ClusterService clusterService;
+    @Mock
+    ConfigurationRepository configurationRepository;
+    @Mock
+    Client client;
+
+    @Test
+    public void testUserTypeFilter() throws Exception {
```
peternied, line 43: Please make sure there is test cases for for filter to type `any` as well

peternied (commented):
Adding some feedback about the new unit tests - really close.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
samuelcostae: I had to add because when doing the put request to create an account in the tests,  the application expected the 'service' entry to be on the root of the json,  and the 'service' entry inside 'attributes' wasn;t being used

image

src/test/java/org/opensearch/security/UserServiceUnitTests.java
samuelcostae: yep

src/test/java/org/opensearch/security/UserServiceUnitTests.java
samuelcostae: yep, it was just a draft commit upload to check that way to get the config was valid

src/test/java/org/opensearch/security/UserServiceUnitTests.java
samuelcostae: yep, it was just a draft commit upload to check that way to get the config was valid

src/test/java/org/opensearch/security/UserServiceUnitTests.java
samuelcostae: yep, it was just a draft commit upload to check that way to get the config was valid

src/main/java/org/opensearch/security/user/UserService.java
samuelcostae: Thanks, but i didn't do anything xD, just signed-off the commit suggestion xD

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
```diff
@@ -55,7 +58,8 @@ protected String getEndpointPrefix() {
 
     private static final String ENABLED_SERVICE_ACCOUNT_BODY = "{"
         + " \"attributes\": { \"service\": \"true\", "
-        + "\"enabled\": \"true\"}"
+        + " \"enabled   \": \"true\"},"
+        + " \"service\": \"true\" "
```
peternied, line 62: This should be reverted, I'd expect the attribute on line 60 to handle this case.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
peternied: I'd recommend making a unit test to sent the expected response (with service in the attributes) and use that to troubleshot what isn't working.  @scrawfor99 Do you have any additional thoughts?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -179,8 +208,8 @@ ValidationResult<SecurityConfiguration> createOrUpdateAccount(
         try {
             final var username = securityConfiguration.entityName();
             final var content = (ObjectNode) securityConfiguration.requestContent();
-            if (request.hasParam("service")) {
-                content.put("service", request.param("service"));
+            if (request.hasParam("attributes")) {
+                content.put("attributes", request.param("attributes"));
```
samuelcostae, line 212: @peternied  @scrawfor99  with this change to the put request I was able to not use the 'service' in the root of the object. Does that make sense?

src/test/java/org/opensearch/security/dlic/rest/api/UserApiTest.java
samuelcostae: https://github.com/opensearch-project/security/pull/2786/files#r1334459144

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
samuelcostae: https://github.com/opensearch-project/security/pull/2786/files#r1334459144

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
peternied: Looks like this was a bug in the original implemetnation, thanks for finding and fixing.

peternied (approved):

peternied:
@scrawfor99 Could you take another look?

scrawfor99 (approved):
Great job Sam :)

PR #2793, opened 2023-05-26
Introduce formatting guidelines and documentation; add a new editorconfig file for use with IDEs 

scrawfor99:
### Description
I often see comments on pull requests from people asking to not change the whitespace in files etc. Basically, there is no standardized practice contributors are expected to follow and so some files have space while others have tabs. 

This change introduces guidelines for styling and applies these guidelines to all files in the code base.  It also adds a security.editorconfig file which can be used with IntelliJ or VS Code to import the standardized settings to the IDE. 

The rules being added are:

```
## Common Practice

Please visit [security.editorconfig](./security.editorconfig) to download the complete formatting settings.
Use with IntelliJ is natively supported, while Visual Studio Code supports use through the EditorConfig Extension.

### Whitespace

* Indentations should use 1 tab (of size 4 spaces) in Java files
* Indentations should be 2 spaces in YAML files
* There should be a preceding space before all 'if', 'for', 'while', 'switch', 'try', 'catch', and 'synchronized'
  parentheses
* There should be a preceding space before all non-unary operators

### Imports

* Imports should use single class names (no wildcards)
* JavaDocs should use fully qualified class names
* Java.* imports should be placed at the top, followed by Javax.*, then all other imports, and finally static imports. 

```

If any other specifications should be met, please leave a comment and I will update the requirements. 

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -346,7 +346,6 @@ task integrationTest(type: Test) {
 check.dependsOn integrationTest
 
 dependencies {
-    implementation 'jakarta.annotation:jakarta.annotation-api:1.3.5'
     implementation "org.opensearch.plugin:transport-netty4-client:${opensearch_version}"
     implementation "org.opensearch.client:opensearch-rest-high-level-client:${opensearch_version}"
     implementation "org.apache.httpcomponents.client5:httpclient5-cache:${versions.httpclient5}"


cwperks (changes requested):
Thank you for the PR @scrawfor99! I have noticed that `spotlessApply` works differently on thi repo then on other repositories. I think we should opt to configure spotless in line with other repos and enforce that PRs run `spotlessApply` before getting checked in then to do this with an editor config.

See example here where I took the spotless config from job-scheduler and applied it to this repo: https://github.com/opensearch-project/security/compare/main...cwperks:configure-spotless?expand=1

For Job-Scheduler the gradle formatting config is primarily driven by this file: https://github.com/opensearch-project/job-scheduler/blob/main/gradle/formatting.gradle

scrawfor99:
> 

Hi @cwperks, yeah I am trying to see what it is doing now. I need to look at the code I think since it complains even when you enforce java imports first so not sure what the specific requirements are. I will make that part of this change as well.

cwperks:
@scrawfor99 I removed the output of running `spotlessApply` these changes show how to setup the `formatting.gradle`: https://github.com/opensearch-project/security/compare/main...cwperks:configure-spotless-plain?expand=1

reta:
@scrawfor99 certainly :+1: for Spotless (no formatting wars :D), but I think:
 - you should exclude build files (build.gradle)
 - you should exclude `.github` folder and *.MD (markdown) files

src/main/java/org/opensearch/security/auditlog/AuditLog.java
```diff
@@ -1,34 +1,15 @@
 /*
- * Copyright 2015-2018 _floragunn_ GmbH
```
peternied, line 2: Need to keep the original license headers on files like this

peternied (commented):
Thanks for taking the initiative on this @scrawfor99 This PR changes pretty much every file in the codebase making it hard to see what is being changed or if the style changes are easier / harder to read.  What do you think about making the rules updates more incremental?

Specific concerns:
- I don't like the markdown line breaks mid-sentence, that isn't how I write.
- What about onboarding yaml/xml/md/json/gradle/... one at a time, or at least isolated from the java style changes.
- Looks like you are changing the license header that looks different from the default in OpenSearch - is there an issue where this is discussed?
- Looks like legacy licenses are being overwritten

scrawfor99:
> 

Hi @peternied,

Thank you for following up. 

I will go ahead and make a more incremental rollout process like you suggested. 

I will also swap the line breaks in the markdowns. I will also followup on the comments you left regarding the license headers. Because of how the formatting works, it can be hard to tell what is being changes as you mentioned. I don't believe the changes from the other repos are meant to swap the headers so I will double check to make sure that everything is correct. I will make a tracking issue and link the related PRs so that it is easier to tell what is being changed. I will then refactor by package instead of having the IDE refactor everything at once.

Issue #2573, opened 2023-03-21
[META] On-Behalf-Of Authentication


RyanL1997:
### Description
When security is installed, extensions will need an auth token in order to interact with the OpenSearch cluster. This auth token will be in the form of a JWT. Extensions are a replacement for plugins, so any information from a user that plugins utilize today should be contained as a claim in the JWT sent to an extension. 

Example header + payload:

```
Header:
{"alg":"HS512"}

Payload:
{
  "iss": "<cluster_name>",
  "iat":1676908684,
  "exp":1676908744,
  "sub":"<principal_identifier_token>",
  "er":"<encrypted_mapped_roles>", # r for roles
  "br": "<encrypted_backend_roles>", # br for backend_roles
  "aud": "extension/{extensionUniqueId}"
}
```

Useful reference class to see how JWTs are generated within the security plugin on successful SAML authentication: https://github.com/opensearch-project/security/blob/main/src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java

For the initial extensions design, these tokens will allow the extension to interact with the OpenSearch cluster using the same privileges as the initiating user. 

For the initial implementation, the JWTs can be signed with an HMAC 512 hash by default. If any encryption is performed, then the extension will require a mechanism for decrypting the JWE to view the payload of the JWT. The signing key should be configured in the security configuration. Maybe in the `config.dynamic` portion of `config.yml`?

### Design
- [x] JWT vs JWE: https://github.com/opensearch-project/security/issues/2545
- [x] https://github.com/opensearch-project/security/pull/2567
- [x] #2620 
- [x] #2619 
    - [x] #2672 
- [x] https://github.com/opensearch-project/security/issues/2615
    - [x] https://github.com/opensearch-project/security/pull/2671
    - [x] https://github.com/opensearch-project/security/pull/2857
- [x] https://github.com/opensearch-project/security/issues/3000
- [x] #2865
- [x] #2764

### Implementation
- [x] https://github.com/opensearch-project/security/pull/3047
- [x] #2891
    - [x] https://github.com/opensearch-project/security/pull/3008
- [x] https://github.com/opensearch-project/security/issues/3102
    - [x] https://github.com/opensearch-project/security/pull/3117
- [x] https://github.com/opensearch-project/security/issues/2707

### Integration
- [x] https://github.com/opensearch-project/security/issues/2945
    - [x] https://github.com/opensearch-project/security/pull/3179
[...]

PR #2857, opened 2023-06-13
[Feature/Extension] Add oboauthcbackend registry and set up e2e endpoint testing flow

RyanL1997:
### Description
Add oboauthcbackend registry and set up e2e endpoint testing flow

New feature

### Issues Resolved
* Relate https://github.com/opensearch-project/security/issues/2573

### Testing

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
@@ -61,6 +61,7 @@ public class OnBehalfOfJwtAuthenticationTest {
 
 	@Test
 	public void shouldAuthenticateWithJwtToken_positive() {
+		// TODO: This integration test should use an endpoint to get an OnBehalfOf token, not generate it 
 		try(TestRestClient client = cluster.getRestClient(tokenFactory.generateValidToken())){
 
 			TestRestClient.HttpResponse response = client.getAuthInfo();
--- a/src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
@@ -1,6 +1,7 @@
 package org.opensearch.security.http;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.security.KeyPair;
 import java.util.Base64;
 import java.util.List;
@@ -22,6 +23,7 @@ import org.opensearch.action.search.SearchResponse;
 import org.opensearch.client.Client;
 import org.opensearch.client.RestHighLevelClient;
 import org.opensearch.test.framework.JwtConfigBuilder;
+import org.opensearch.test.framework.OnBehalfOfConfig;
 import org.opensearch.test.framework.TestSecurityConfig;
 import org.opensearch.test.framework.cluster.ClusterManager;
 import org.opensearch.test.framework.cluster.LocalCluster;
@@ -54,14 +56,7 @@ public class OnBehalfOfJwtAuthenticationTest {
 	public static final String CLAIM_ROLES = "backend-user-roles";
 
 	public static final String USER_SUPERHERO = "superhero";
-	public static final String USERNAME_ROOT = "root";
-	public static final String ROLE_ADMIN = "role_admin";
-	public static final String ROLE_DEVELOPER = "role_developer";
-	public static final String ROLE_QA = "role_qa";
-	public static final String ROLE_CTO = "role_cto";
-	public static final String ROLE_CEO = "role_ceo";
 	public static final String ROLE_VP = "role_vp";
-	public static final String POINTER_BACKEND_ROLES = "/backend_roles";
 	public static final String POINTER_USERNAME = "/user_name";
 
 	public static final String QA_DEPARTMENT = "qa-department";
@@ -73,17 +68,31 @@ public class OnBehalfOfJwtAuthenticationTest {
 	public static final String QA_SONG_INDEX_NAME = String.format("song_lyrics_%s", QA_DEPARTMENT);
 
 	private static final KeyPair KEY_PAIR = Keys.keyPairFor(SignatureAlgorithm.RS256);
-	private static final String PUBLIC_KEY = new String(Base64.getEncoder().encode(KEY_PAIR.getPublic().getEncoded()), US_ASCII);
 
 	static final TestSecurityConfig.User ADMIN_USER = new TestSecurityConfig.User("admin").roles(ALL_ACCESS);
 
 	private static final String JWT_AUTH_HEADER = "jwt-auth";
 
-	private static final JwtAuthorizationHeaderFactory tokenFactory = new JwtAuthorizationHeaderFactory(
-			KEY_PAIR.getPrivate(),
-			CLAIM_USERNAME,
-			CLAIM_ROLES,
-			JWT_AUTH_HEADER);
+	public static final String issuer = "cluster_0";
+	public static final String subject = "testUser";
+	public static final String audience = "audience_0";
+	public static final Integer expirySeconds = 100000;
+	public static final String headerName = "Bearer";
+	public static final List<String> roles = List.of("admin", "HR");
+
+	private static final String signingKey = Base64.getEncoder().encodeToString("jwt signing key for an on behalf of token authentication backend for testing of extensions".getBytes(StandardCharsets.UTF_8));
+	private static final String encryptionKey = Base64.getEncoder().encodeToString("encryptionKey".getBytes(StandardCharsets.UTF_8));
+
+	private static final OnBehalfOfJwtAuthorizationHeaderFactory tokenFactory = new OnBehalfOfJwtAuthorizationHeaderFactory(
+			signingKey,
+			issuer,
+			subject,
+			audience,
+			roles,
+			expirySeconds,
+			headerName,
+			encryptionKey
+	);
 
 	public static final String SONG_ID_1 = "song-id-01";
 
@@ -94,7 +103,7 @@ public class OnBehalfOfJwtAuthenticationTest {
 	public static final LocalCluster cluster = new LocalCluster.Builder()
 			.clusterManager(ClusterManager.SINGLENODE).anonymousAuth(false)
 			.nodeSettings(Map.of("plugins.security.restapi.roles_enabled", List.of("user_" + ADMIN_USER.getName()  +"__" + ALL_ACCESS.getName())))
-			.authc(AUTHC_HTTPBASIC_INTERNAL).users(ADMIN_USER).roles(DEPARTMENT_SONG_LISTENER_ROLE).config()
+			.authc(AUTHC_HTTPBASIC_INTERNAL).onBehalfOf(new OnBehalfOfConfig().signing_key(signingKey).encryption_key(encryptionKey))
 			.build();
 
 	@Rule
@@ -112,13 +121,15 @@ public class OnBehalfOfJwtAuthenticationTest {
 
 	@Test
 	public void shouldAuthenticateWithJwtToken_positive() {
-		try(TestRestClient client = cluster.getRestClient(tokenFactory.generateValidToken(USER_SUPERHERO))){
+		try(TestRestClient client = cluster.getRestClient(tokenFactory.generateValidToken())){
 
 			TestRestClient.HttpResponse response = client.getAuthInfo();
 
 			response.assertStatusCode(200);
 			String username = response.getTextFromJsonBody(POINTER_USERNAME);
-			assertThat(username, equalTo(username));
+			assertThat("testUser", equalTo(username));
+		} catch (Exception e) {
+			throw new RuntimeException(e);
 		}
 	}
 }
--- a/src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
@@ -1,5 +1,11 @@
 package org.opensearch.security.http;
 
+import java.io.IOException;
+import java.security.KeyPair;
+import java.util.Base64;
+import java.util.List;
+import java.util.Map;
+
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
 import io.jsonwebtoken.SignatureAlgorithm;
 import io.jsonwebtoken.security.Keys;
@@ -10,6 +16,7 @@ import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.client.Client;
@@ -20,12 +27,6 @@ import org.opensearch.test.framework.cluster.ClusterManager;
 import org.opensearch.test.framework.cluster.LocalCluster;
 import org.opensearch.test.framework.cluster.TestRestClient;
 import org.opensearch.test.framework.log.LogsRule;
-
-import java.io.IOException;
-import java.security.KeyPair;
-import java.util.Base64;
-import java.util.List;
-import java.util.Map;
 
 import static java.nio.charset.StandardCharsets.US_ASCII;
 import static org.apache.http.HttpHeaders.AUTHORIZATION;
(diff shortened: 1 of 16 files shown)


RyanL1997:
I will update the test in next commit

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfToken.java
```diff
@@ -0,0 +1,155 @@
...
+
+                    final String source = "self-issued";
+                    final User user = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+        
+                    builder.startObject();
+                    builder.field("user", user.getName());
+                    final String token = vendor.createJwt(/* TODO: Update the issuer to represent the cluster */"OpenSearch",
+                        user.getName(),
+                        source,
+                        tokenDuration,
```
RyanL1997, line 134: Hi @peternied, I see the `tokenDuration` is here. So this should be the representation of the expiry right?

peternied: Yes

RyanL1997: I'm changing this into seconds cuz the `createJwt()` is taking in expiry in seconds

RyanL1997 (commented):
Hi @peternied and @cwperks, I have tried to setup the testing with endpoint. The first thing I tried is the manual testing, and I'm updating the findings here:

- I'm able to get the obo token by using this command:
```
curl -XPOST https://localhost:9200/_plugins/_security/api/user/onbehalfof --insecure -u "admin:admin" --data '{"reason":"Test generation"}' -H "Content-Type: application/json"
```
The response is something like this:
```
{
  "user": "admin",
  "onBehalfOfToken": "eyJhbGciOiJIUzUxMiJ9.eyJpc3MiOiJPcGVuU2VhcmNoIiwiaWF0IjoxNjg2ODA4NzE1LCJzdWIiOiJhZG1pbiIsImF1ZCI6InNlbGYtaXNzdWVkIiwibmJmIjoxNjg2ODA4NzE1LCJleHAiOjE3NzMyMDg3MTUsImVyIjoiSzZJTFlQcUVuakZTcmprSXJzbnlMZz09In0.U_YzpWv4_iIR6_LEc5ty3EVa31MqiaCTUW-mH8DorRZZqaPUFFCgyYiIp4XVJmqqwMJUHfWQUHwxVGPB6imOvg",
  "duration": 24
}
```
- However, the token seems doesn't have the `roles` in its claim:
```curl -XGET https://localhost:9200/ --insecure -H "Authorization: Bearer $OBOToken"
{
  "error": {
    "root_cause": [
      {
        "type": "security_exception",
        "reason": "no permissions for [cluster:monitor/main] and User [name=admin, backend_roles=[], requestedTenant=null]"
      }
    ],
    "type": "security_exception",
    "reason": "no permissions for [cluster:monitor/main] and User [name=admin, backend_roles=[], requestedTenant=null]"
  },
  "status": 403
}
```

After some debugging, I left some comments on the code and I think [this comment](https://github.com/opensearch-project/security/pull/2857#discussion_r1230466169) is pointing to the root cause of the above 403 error.

RyanL1997:
Looking into this error in build CIs:
```
Error: 023-06-19T23:26:40,557][ERROR][org.opensearch.security.http.OnBehalfOfAuthenticator] Error while creating JWT authenticator
  1> io.jsonwebtoken.io.DecodingException: Illegal base64 
```

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthorizationHeaderFactory.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.apache.hc.core5.http.Header;
+import org.apache.hc.core5.http.message.BasicHeader;
+
+import org.opensearch.common.settings.Settings;
+import org.opensearch.security.authtoken.jwt.JwtVendor;
+
+import static java.util.Objects.requireNonNull;
+
+class OnBehalfOfJwtAuthorizationHeaderFactory {
```
peternied, line 24: Remove this class, this is acting like a mock for unit tests, any tests that depend on this class should be converted to unit tests where they are faster to run / debug

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
```diff
@@ -0,0 +1,136 @@
...
+			.nodeSettings(Map.of(
+					"plugins.security.allow_default_init_securityindex", true,
+					"plugins.security.restapi.roles_enabled", List.of("user_admin__all_access")
+			))
+			.authc(AUTHC_HTTPBASIC_INTERNAL)
+			.onBehalfOf(new OnBehalfOfConfig().signing_key(signingKey).encryption_key(encryptionKey))
+			.build();
+
+	@Test
+	public void shouldAuthenticateWithOBOToken() {
```
peternied, line 90: This should be a unit test

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+		try (TestRestClient client = cluster.getRestClient(ADMIN_USER_NAME, DEFAULT_PASSWORD)) {
+
+			client.assertCorrectCredentials(ADMIN_USER_NAME);
+
+			TestRestClient.HttpResponse response = client.postJson(OBO_ENDPOINT_PREFIX, OBO_TOKEN_REASON);
+			response.assertStatusCode(200);
+
+			Map<String, Object> oboEndPointResponse = response.getBodyAs(Map.class);
+			assertThat(oboEndPointResponse, allOf(
```
peternied, line 115: Nice 🔥

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
```diff
@@ -0,0 +1,136 @@
...
+			response.assertStatusCode(200);
+			String username = response.getTextFromJsonBody(POINTER_USERNAME);
+			assertThat("testUser", equalTo(username));
+		} catch (Exception e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	@Test
+	public void shouldAuthenticateWithOBOTokenEndPoint() {
```
peternied, line 104: What do you think about adding a test that retrieves the token and the tamper with it attempts to use it?

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+		try (TestRestClient client = cluster.getRestClient(adminOboAuthHeader)) {
+
+			TestRestClient.HttpResponse response = client.getAuthInfo();
+			response.assertStatusCode(200);
+
+			String username = response.getTextFromJsonBody(POINTER_USERNAME);
+			assertThat(username, equalTo(ADMIN_USER_NAME));
+		}
+	}
```
peternied, line 135: We should also have tests that check that OBO tokens cannot be used to 
- Get a new OBO token
- Reset the password (?)

Note; these could be follow up tasks

src/main/java/org/opensearch/security/auth/internal/NoOpAuthenticationBackend.java
```diff
@@ -46,7 +46,9 @@ public String getType() {
 
     @Override
     public User authenticate(final AuthCredentials credentials) {
-        return new User(credentials.getUsername(), credentials.getBackendRoles(), credentials);
+        User user = new User(credentials.getUsername(), credentials.getBackendRoles(), credentials);
+        user.addSecurityRoles(credentials.getSecurityRoles());
```
peternied, line 50: Why was this needed?

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
```diff
@@ -20,94 +20,130 @@
...
+            if (rolesObject == claims.get("er")) {
+                decryptedRoles = EncryptionDecryptionUtil.decrypt(encryptionKey, rolesClaim);
+            }
+            roles = Arrays.stream(decryptedRoles.split(",")).map(String::trim).collect(Collectors.toList());
+        }
+
+        return roles;
+    }
+
+    private String[] extractBackendRolesFromClaims(Claims claims) {
```
peternied, line 127: IMO We shouldn't add backend roles in this token - discussion is in https://github.com/opensearch-project/security/issues/2865

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
```diff
@@ -20,94 +20,130 @@
...
+    }
+
+    private List<String> extractSecurityRolesFromClaims(Claims claims) {
+        Object rolesObject = ObjectUtils.firstNonNull(claims.get("er"), claims.get("dr"));
+        List<String> roles;
+
+        if (rolesObject == null) {
+            log.warn(
+                    "Failed to get roles from JWT claims. Check if this key is correct and available in the JWT payload.");
+            roles = List.of();
```
peternied, line 112: If this message appeared in the logs there is no context to know which token was bad or how to troubleshoot it more.  I'd expect any token like this should be reported as malformed and no more message is needed

peternied (commented):
@RyanL1997 Up to you what you'd like to address right now.  I'll do a more thorough review/following when the feature branch is being merged to main.

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
RyanL1997: Ya this test case can be removed

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthorizationHeaderFactory.java
RyanL1997: Should we keep this for some of the unhappy cases?

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
RyanL1997: Thanks for the reminder! I have created an issue to track these edge cases (https://github.com/opensearch-project/security/issues/2891)

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthorizationHeaderFactory.java
scrawfor99: You can either mock just mock the unhappy cases you want. We don't actually care about the creation of the tokens for the unit tests, just the handling so you can skip that step and avoid the computation involved in using an actual token.

src/integrationTest/resources/config.yml
```diff
@@ -16,5 +16,5 @@ config:
           type: "internal"
           config: {}
     on_behalf_of:
-      signing_key: "signing key"
-      encryption_key: "encryption key"
+      signing_key: "VGhpcyBpcyB0aGUgand0IHNpZ25pbmcga2V5IGZvciBhbiBvbiBiZWhhbGYgb2YgdG9rZW4gYXV0aGVudGljYXRpb24gYmFja2VuZCBmb3IgdGVzdGluZyBvZiBleHRlbnNpb25z"
```
scrawfor99, line 19: NIT: Add a comment for what these are--how did you get them and how can we update them etc.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfToken.java
scrawfor99, line 1: Can we rename this `CreateOnBehalfofTokenAction`? This will make it consistent with the other actions.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfToken.java
```diff
@@ -0,0 +1,152 @@
...
+            this.vendor = null;
+        }
+    }
+
+    public CreateOnBehalfOfToken(final Settings settings, final ThreadPool threadPool) {
+        this.threadPool = threadPool;
+    }
+
+    @Override
+    public String getName() {
```
scrawfor99, line 72: Another place where I think have a different name may make this more obvious what is going on. If you were to call 

```
CreateOnBehalfOfToken myToken = new CreateOnBehalfOfToken(...) 
myToken.getName() 
```

I would not expect "CreateOnBehalfOfToken" -- There is nothing wrong with treating it like this but I think the naming could be clarified.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfToken.java
```diff
@@ -0,0 +1,152 @@
...
+
+    private RestChannelConsumer handlePost(RestRequest request, NodeClient client) throws IOException {
+        return new RestChannelConsumer() {
+            @Override
+            public void accept(RestChannel channel) throws Exception {
+                final XContentBuilder builder = channel.newBuilder();
+                BytesRestResponse response;
+                try {
+                    if (vendor == null) {
+                        channel.sendResponse(new BytesRestResponse(RestStatus.SERVICE_UNAVAILABLE, "on_behalf_of configuration is not being configured"));
```
scrawfor99, line 103: Nit: Is this message correct? I thought this was for calling the token creation not for configuring the Security settings of the token creator?

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfToken.java
```diff
@@ -0,0 +1,152 @@
...
+                    }
+
+                    final Map<String, Object> requestBody = request.contentOrSourceParamParser().map();
+                    final String reason = (String)requestBody.getOrDefault("reason", null);
+
+                    final Integer tokenDuration = Optional.ofNullable(requestBody.get("duration"))
+                            .map(value -> (String)value)
+                            .map(Integer::parseInt)
+                            .map(value -> Math.min(value, 72 * 3600)) // Max duration is 72 hours
+                            .orElse(24 * 3600); // Fallback to default;
```
scrawfor99, line 114: Seems quite long--I was under the impression we were thinking 5 minutes or so? Could this not lead to up to 3 days of an invalid role being active?

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfToken.java
```diff
@@ -0,0 +1,152 @@
...
+
+                    final String source = "self-issued";
+                    final User user = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+                    final TransportAddress caller = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
+                    Set<String> mappedRoles = mapRoles(user, caller);
+
+                    builder.startObject();
+                    builder.field("user", user.getName());
+                    final String token = vendor.createJwt(/* TODO: Update the issuer to represent the cluster */"OpenSearch",
```
scrawfor99, line 123: Nit: Move comment out of argument.

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
scrawfor99: Perhaps we can just combine the roles at this stage and then drop the backenedRoles from the later AuthCredentials usage.

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
```diff
@@ -149,63 +182,37 @@ private AuthCredentials extractCredentials0(final RestRequest request) {
         }
 
         final int index;
-        if((index = jwtToken.toLowerCase().indexOf(BEARER_PREFIX)) > -1) { //detect Bearer
+        if(jwtToken != null && (index = jwtToken.toLowerCase().indexOf(BEARER_PREFIX)) > -1) { //detect Bearer
```
scrawfor99, line 185: What is the "-1" doing? You should be able to just use a matcher no?

scrawfor99 (commented):
Hi Ryan, overall looks really close. Great job with everything. I left some comments in addition to Peter's. Nothing too major.

src/integrationTest/resources/config.yml
RyanL1997: Good idea I can do a comment above to show the plain text version of them.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfToken.java
RyanL1997: Yes, so basically this line is handling the case that when a user doesn't have the `on_behalf_of` section configured in the security's config and still trying to call the obo endpoint. According to this scenario, Craig also suggested that we should also provide some useful error msg to show user what is going on, and that's why I put something like "Obo setting is not there, so this service is not available" Otherwise, it will just go through the config file and give user an error like:
```
{
  "error": {
    "root_cause": [
      {
        "type": "json_generation_exception",
        "reason": "Can not start an object, expecting field name (context: Object)"
      }
    ],
    "type": "json_generation_exception",
    "reason": "Can not start an object, expecting field name (context: Object)"
  },
```
Which is also make sense, but it may confuse the user to understand what is going on.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfToken.java
RyanL1997: Ya this is the one I will modify it to something like 5 mins (300 seconds). I will do that in the next couple commits.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -533,7 +534,7 @@ private User impersonate(final RestRequest request, final User originalUser) thr
             //loop over all http/rest auth domains
             for (final AuthDomain authDomain: restAuthDomains) {
                 final AuthenticationBackend authenticationBackend = authDomain.getBackend();
-                if (authDomain.getBackend() instanceof NoOpAuthenticationBackend  ) {
+                if (authDomain.getHttpAuthenticator() instanceof OnBehalfOfAuthenticator) {
```
peternied, line 537: This doesn't seem right, the auth domain associated with the OnBehalfOfAuthenticator isn't configured correctly

src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java
```diff
@@ -312,6 +314,12 @@ private void buildAAA() {
             }
         }
 
+        Settings oboSettings = getDynamicOnBehalfOfSettings();
+        if (oboSettings.get("signing_key") != null && oboSettings.get("encryption_key") != null) {
+            final AuthDomain _ad = new AuthDomain(new NoOpAuthenticationBackend(Settings.EMPTY, null), new OnBehalfOfAuthenticator(getDynamicOnBehalfOfSettings()), false, -1);
```
peternied, line 319: ⚠️ NoOpAuthenticationBackend always returns a user object even if they don't exist - that's a problem

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
RyanL1997: I have added another test case~

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthorizationHeaderFactory.java
RyanL1997: Ya I have removed it since most of them were in the unit testing of this authc backend.

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
RyanL1997: I think the `indexOf` method returns the index of the first occurrence of the specified substring. If the substring is not found, it returns -1.  But yeah, I have changed it into the matcher.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfToken.java
RyanL1997: Done~

src/integrationTest/resources/config.yml
RyanL1997: Done

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfToken.java
RyanL1997: done~

src/main/java/org/opensearch/security/auth/BackendRegistry.java
RyanL1997: I think this is the solution I can think of for only skipping the OBO auth backend but not the other backends.(cc: @cwperks )

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
RyanL1997: Fixed~

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
RyanL1997: Ya I have changed it, align with our discussion on #2865 that only including backend roles when the bwc mode is on.

RyanL1997:
> Perhaps we can just combine the roles at this stage and then drop the backenedRoles from the later AuthCredentials usage.

@scrawfor99 , I have updated the code a little bit that the extraction of the backend roles will be called optionally based on the bwc mode, since we are not including backend roles in default.

src/main/java/org/opensearch/security/auth/internal/NoOpAuthenticationBackend.java
RyanL1997: AuthCredential class didn't have mechanism of passing the security roles, and it only had backend roles included before. So this is the way that the obo token can have security roles included in its payload.

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+	@Test
+	public void shouldNotAuthenticateWithATemperedOBOToken() {
```
cwperks, line 102: This is a good start, but we should aim to cover more scenarios in this suite. 

For this test why not try passing something known to be invalid like `invalid-token` instead of requesting a valid token from the endpoint and removing the last character?

src/integrationTest/resources/config.yml
```diff
@@ -16,5 +16,7 @@ config:
           type: "internal"
           config: {}
     on_behalf_of:
-      signing_key: "signing key"
-      encryption_key: "encryption key"
+      # The decrypted signing key is: This is the jwt signing key for an on behalf of token authentication backend for testing of extensions
```
cwperks, line 19: *decoded - these are base64 encoded strings

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -0,0 +1,155 @@
...
+                    final Map<String, Object> requestBody = request.contentOrSourceParamParser().map();
+                    final String reason = (String)requestBody.getOrDefault("reason", null);
+
+                    final Integer tokenDuration = Optional.ofNullable(requestBody.get("duration"))
+                            .map(value -> (String)value)
+                            .map(Integer::parseInt)
+                            .map(value -> Math.min(value, 10 * 60)) // Max duration is 10 minutes
+                            .orElse(5 * 60); // Fallback to default of 5 minutes;
+
+                    final String source = "self-issued";
```
cwperks, line 116: This may be fine merging into the `feature/extensions` branch, but would need to be addressed for a merge in `main`. Will this be changed into an API parameter?

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -0,0 +1,155 @@
...
+                    final User user = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+                    final TransportAddress caller = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
+                    Set<String> mappedRoles = mapRoles(user, caller);
+
+                    builder.startObject();
+                    builder.field("user", user.getName());
+
+                    /* TODO: Update the issuer to represent the cluster */
+                    final String token = vendor.createJwt(
+                            "OpenSearch",
```
cwperks, line 126: Is it possible to get the cluster name here?

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -0,0 +1,155 @@
...
+                            tokenDuration,
+                            mappedRoles.stream().collect(Collectors.toList()),
+                            user.getRoles().stream().collect(Collectors.toList()));
+                    builder.field("onBehalfOfToken", token);
+                    builder.field("duration", tokenDuration);
+                    builder.endObject();
+
+                    response = new BytesRestResponse(RestStatus.OK, builder);
+                } catch (final Exception exception) {
+                    System.out.println(exception.toString());
```
cwperks, line 138: Remove the sysout

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
cwperks: For obo tokens that are ferried to an extension when handling a REST request then backend roles should be included if the plugin being converted to an extension relies on them. For obo tokens generated during the call to the CreateOboToken endpoint I don't think they should be included either. 

Since some obo tokens may have them as a claim then this method should remain.

src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java
cwperks: On behalf of tokens are only supported for internal users?

src/test/resources/config.yml
```diff
@@ -97,5 +97,5 @@ config:
     hosts_resolver_mode: "ip-only"
     transport_userrname_attribute: null
     on_behalf_of:
-      signing_key: "signing key"
-      encryption_key: "encryption key"
+      signing_key: "VGhpcyBpcyB0aGUgand0IHNpZ25pbmcga2V5IGZvciBhbiBvbiBiZWhhbGYgb2YgdG9rZW4gYXV0aGVudGljYXRpb24gYmFja2VuZCBmb3IgdGVzdGluZyBvZiBleHRlbnNpb25z"
```
cwperks, line 100: For this and the files below, are these settings required in all of those test config files?

cwperks (commented):
Thank you @RyanL1997. I took another pass at this PR.

src/test/resources/config.yml
RyanL1997: Yes, I believe these files we will need them to run the tests. Remember those `signing key` caused some Base64 decoding issue when there was a space in between.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
RyanL1997: Done

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
RyanL1997: Done~

src/integrationTest/resources/config.yml
RyanL1997: Nice catch. Fixed

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
RyanL1997: Cuz I would like to test if this token can be tampered based on this [comment](https://github.com/opensearch-project/security/pull/2857#discussion_r1236028482). But yeah, it would perform exactly the same of `invalid-token`.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -0,0 +1,157 @@
...
+
+    @Subscribe
+    public void onConfigModelChanged(ConfigModel configModel) {
+        this.configModel = configModel;
+    }
+
+    @Subscribe
+    public void onDynamicConfigModelChanged(DynamicConfigModel dcm) {
+        this.dcm = dcm;
+        if (dcm.getDynamicOnBehalfOfSettings().get("signing_key") != null
```
cwperks, line 62: It may make sense to create a class called `OnBehalfOfSettings` and turn this into a method like `isValid()`. This is fine here, but there is another open PR around signing key configuration which would impact a check like this: https://github.com/opensearch-project/security/pull/2805

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -0,0 +1,157 @@
...
+
+                    final Integer tokenDuration = Optional.ofNullable(requestBody.get("duration"))
+                        .map(value -> (String) value)
+                        .map(Integer::parseInt)
+                        .map(value -> Math.min(value, 10 * 60)) // Max duration is 10 minutes
+                        .orElse(5 * 60); // Fallback to default of 5 minutes;
+
+                    final String service = (String) requestBody.getOrDefault("service", "self-issued");
+                    final User user = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+                    final TransportAddress caller = threadPool.getThreadContext()
```
cwperks, line 122: I'm wondering if host mapping makes sense here. If this API creates a token and gives it to a service. If the service makes a call from a different host/IP address would host mapping also come into the picture on that request?

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
```diff
@@ -0,0 +1,249 @@
...
+        jwtParser = initParser(settings.get("signing_key"));
+    }
+
+    private JwtParser initParser(final String signingKey) {
+        if (signingKey == null || signingKey.length() == 0) {
+            throw new RuntimeException("Unable to find on behalf of authenticator signing key");
+        }
+
+        try {
+            final String minimalKeyFormat = signingKey.replace("-----BEGIN PUBLIC KEY-----\n", "").replace("-----END PUBLIC KEY-----", "");
```
cwperks, line 71: Is there any duplicate code here with the JWTAuthenticator that can be abstracted?

src/test/resources/config.yml
cwperks: I'm confused why the change is necessary in 5 of the test config files. Is it possible to have a config file specifically for the tests added in this PR and configure the settings in that file?

cwperks (commented):
Took another pass and left a couple of comments. The comments could be addressed in a follow-up PR. Thank you @RyanL1997 !

scrawfor99 (approved):
Hi @RyanL1997, looks good to me beyond the comments Craig already left. Thank you for all your hard work with this.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
peternied: Good callout - @RyanL1997 could you create a follow up issue, we should figure out what the 'correct' way to handle this scenario - we should be intentional with these scenarios

peternied:
@RyanL1997 Could you fix the missing DCO from https://github.com/opensearch-project/security/pull/2857/commits/0f7b4931b330ba2affe79f8e065ab06c4be8e672 ?

RyanL1997:
> Could you fix the missing DCO from https://github.com/opensearch-project/security/commit/0f7b4931b330ba2affe79f8e065ab06c4be8e672 ?

Hi @peternied, may I ask that if there is a more efficient way of fixing that? I was trying to use the interactive rebase and trying to do the `git commit --amend --signoff`. However, https://github.com/opensearch-project/security/commit/0f7b4931b330ba2affe79f8e065ab06c4be8e672 is 101 commits away from the current `HEAD` so it is kinda hard for me to go through the entire commit list and re-resolve all the conflicts based on all of the previous commits (including the large scale formatting refactor).

Here is some of the commands I was using:
1. Trying to interactive rebase to https://github.com/opensearch-project/security/commit/0f7b4931b330ba2affe79f8e065ab06c4be8e672: 
```
git rebase --signoff -i HEAD~101
```
2. In the text editor that comes up, replace "pick" with "edit" for the target commit I want to sign off, then save and close the editor.
3. And then the countless conflicts just popped up.

RyanL1997:
> I'm confused why the change is necessary in 5 of the test config files. Is it possible to have a config file specifically for the tests added in this PR and configure the settings in that file?

Hi @cwperks, nice catch actually. I just realized that we wont need all of them since we have fixed the authbackend list. Right now we only need the one for the restapi testing.

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
RyanL1997: Done~

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
RyanL1997: Done

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
RyanL1997: Here we go, and lets move our conversation here: https://github.com/opensearch-project/security/issues/3000

PR #2301, opened 2022-12-06
OpenSSLTest is not using the OpenSSL Provider

reta:
Signed-off-by: Andriy Redko <andriy.redko@aiven.io>

### Description
OpenSSLTest should be using the OpenSSL Provider. There are few issues with `netty-tcnative` and OpenSSL:
 - it uses OpenSSL 1.0 and needs `compat-openssl10` 
 - https://github.com/netty/netty-tcnative/issues/746
 - https://github.com/netty/netty-tcnative/issues/742
 - https://github.com/netty/netty-tcnative/issues/703

### Issues Resolved
Closes https://github.com/opensearch-project/security/issues/2208

Is this a backport? If so, please add backport PR # and/or commits #

### Testing
[Please provide details of testing done: unit testing, integration testing and manual testing]

### Check List
- [ ] New functionality includes testing
- [ ] New functionality has been documented
- [X] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2861, opened 2023-06-14
[Enhancement] Parallel test jobs for CI

### LINKED ISSUE #2798: Speed up CI test workflows in GHA
The CI job `build` can take more than one hour. As low turnaround times are desirable for speedy development, this should be reduced.

https://github.com/opensearch-project/security/actions/runs/4983823949/jobs/8921355684

239952005-c5c32b38-e1f8-4096-bdbc-b48cab789f01

pawel-gudel-eliatra:
### Description
Currently "build" job in Github Actions runs for around 60-80 minutes on average. Long time for tests to be done can slow down development process. Easiest way to make this situation little better is to split tests into separate task and reconfigure Github Actions to run tests in parallel. This change allows to cut run of CI from 60-80 minutes to around 28-30 minutes.

We've determined which tests run are the longest and moved them into separate tasks (defined in `build.gradle` file). We do not want to break test on Developers side so running `gradle test` command would run full test suite (`test` task depends on new ones). This also allows new test to be added run automatically.

### Issues Resolved
#2798 

### Testing
There's no changes in test themselves.

### Check List

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,6 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       separateTestsNames: ${{ steps.set-matrix.outputs.separateTestsNames }}
-      testsNamesExclude: ${{ steps.set-matrix.outputs.testsNamesExclude }}
     steps:
     - name: Set up JDK for build and test
       uses: actions/setup-java@v2
@@ -26,58 +25,15 @@ jobs:
       run: |
         echo "separateTestsNames=$(./gradlew listTasksAsJSON -q --console=plain | tail -n 1)" >> $GITHUB_OUTPUT
 
-  tests:
-    name: tests
-    strategy:
-      fail-fast: false
-      matrix:
-        platform: ["windows-latest", "ubuntu-latest"]
-        jdk: [11, 17]
-    runs-on: ${{ matrix.platform }}
-
-    steps:
-    - name: Set up JDK for build and test
-      uses: actions/setup-java@v2
-      with:
-        distribution: temurin
-        java-version: ${{ matrix.jdk }}
-
-    - name: Checkout security
-      uses: actions/checkout@v2
-
-    - name: Build and Test
-      uses: gradle/gradle-build-action@v2
-      with:
-        arguments: |
-          citest -Dbuild.snapshot=false
-          -x test
-
-    - name: Coverage
-      uses: codecov/codecov-action@v1
-      with:
-        token: ${{ secrets.CODECOV_TOKEN }}
-        files: ./build/reports/jacoco/test/jacocoTestReport.xml
-
-    - uses: actions/upload-artifact@v3
-      if: always()
-      with:
-        name: ${{ matrix.platform }}-JDK${{ matrix.jdk }}-reports
-        path: |
-          ./build/reports/
-
-    - name: check archive for debugging
-      if: always()
-      run: echo "Check the artifact ${{ matrix.platform }}-JDK${{ matrix.jdk }}-reports for detailed test results"
-
   test:
     name: test
     needs: generate-test-list
     strategy:
       fail-fast: false
       matrix:
-        platform: ["windows-latest", "ubuntu-latest"]
-        jdk: [11, 17]
         gradle_task: ${{ fromJson(needs.generate-test-list.outputs.separateTestsNames) }}
+        platform: [windows-latest, ubuntu-latest]
+        jdk: [11, 17]
     runs-on: ${{ matrix.platform }}
 
     steps:
@@ -120,7 +76,7 @@ jobs:
       fail-fast: false
       matrix:
         jdk: [17]
-        platform: ["ubuntu-latest", "windows-latest"]
+        platform: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.platform }}
 
     steps:
@@ -149,7 +105,7 @@ jobs:
       fail-fast: false
       matrix:
         jdk: [11, 17]
-        platform: ["ubuntu-latest", "windows-latest"]
+        platform: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.platform }}
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,6 @@ jobs:
       id: set-matrix
       run: |
         echo "separateTestsNames=$(./gradlew listTasksAsJSON -q --console=plain | tail -n 1)" >> $GITHUB_OUTPUT
-        echo "testsNamesExclude=$(./gradlew listTasksAsParam -q --console=plain | tail -n 1)" >> $GITHUB_OUTPUT
 
   build:
     name: build
@@ -51,7 +50,7 @@ jobs:
       uses: gradle/gradle-build-action@v2
       with:
         arguments: |
-          build citest ${{ needs.generate-test-list.outputs.testsNamesExclude }} -Dbuild.snapshot=false
+          build citest -Dbuild.snapshot=false
           -x integrationTest
           -x spotlessCheck
           -x checkstyleMain
@@ -111,7 +110,6 @@ jobs:
           -x spotbugsMain
           -x spotbugsIntegrationTest
           -x checkstyleIntegrationTest
-          -x citest
           -x test
 
     - uses: actions/upload-artifact@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Generate list of tasks
       id: set-matrix
       run: |
-        echo "separateTestsNames=$(./gradlew listTasksAsJSON -q --console=plain | tail -n 1)" >> $GITHUB_OUTPUT
+        echo "separateTestsNames=$(./gradlew listTasksAsJSON -q --console=plain | head -n 1)" >> $GITHUB_OUTPUT
 
   build:
     name: build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,30 @@ env:
   GRADLE_OPTS: -Dhttp.keepAlive=false
 
 jobs:
+  generate-test-list:
+    runs-on: ubuntu-latest
+    outputs:
+      separateTestsNames: ${{ steps.set-matrix.outputs.separateTestsNames }}
+      testsNamesExclude: ${{ steps.set-matrix.outputs.testsNamesExclude }}
+    steps:
+    - name: Set up JDK for build and test
+      uses: actions/setup-java@v2
+      with:
+        distribution: temurin # Temurin is a distribution of adoptium
+        java-version: 17
+
+    - name: Checkout security
+      uses: actions/checkout@v2
+
+    - name: Generate list of tasks
+      id: set-matrix
+      run: |
+        echo "separateTestsNames=$(./gradlew listTasksAsJSON -q --console=plain | tail -n 1)" >> $GITHUB_OUTPUT
+        echo "testsNamesExclude=$(./gradlew listTasksAsParam -q --console=plain | tail -n 1)" >> $GITHUB_OUTPUT
+
   build:
     name: build
+    needs: generate-test-list
     strategy:
       fail-fast: false
       matrix:
@@ -29,18 +51,66 @@ jobs:
       uses: gradle/gradle-build-action@v2
       with:
         arguments: |
-          build test -Dbuild.snapshot=false
+          build test ${{ needs.generate-test-list.outputs.testsNamesExclude }} -Dbuild.snapshot=false
           -x integrationTest
           -x spotlessCheck
           -x checkstyleMain
           -x checkstyleTest
           -x spotbugsMain
+          -x spotbugsIntegrationTest
+          -x checkstyleIntegrationTest
 
     - name: Coverage
       uses: codecov/codecov-action@v1
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
         files: ./build/reports/jacoco/test/jacocoTestReport.xml
+
+    - uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: ${{ matrix.platform }}-JDK${{ matrix.jdk }}-reports
+        path: |
+          ./build/reports/
+
+    - name: check archive for debugging
+      if: always()
+      run: echo "Check the artifact ${{ matrix.platform }}-JDK${{ matrix.jdk }}-reports for detailed test results"
+
+  build-test:
+    name: build-test
+    needs: generate-test-list
+    strategy:
+      fail-fast: false
+      matrix:
+        jdk: [11, 17]
+        platform: ["ubuntu-latest", "windows-latest"]
+        gradle_task: ${{ fromJson(needs.generate-test-list.outputs.separateTestsNames) }}
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+    - name: Set up JDK for build and test
+      uses: actions/setup-java@v2
+      with:
+        distribution: temurin # Temurin is a distribution of adoptium
+        java-version: ${{ matrix.jdk }}
+
+    - name: Checkout security
+      uses: actions/checkout@v2
+
+    - name: Build and Test
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: |
+          build ${{ matrix.gradle_task }} -Dbuild.snapshot=false
+          -x integrationTest
+          -x spotlessCheck
+          -x checkstyleMain
+          -x checkstyleTest
+          -x spotbugsMain
+          -x spotbugsIntegrationTest
+          -x checkstyleIntegrationTest
+          -x test
 
     - uses: actions/upload-artifact@v3
       if: always()
@@ -84,7 +154,6 @@ jobs:
           -x spotbugsMain
 
   backward-compatibility:
-
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Generate list of tasks
       id: set-matrix
       run: |
-        echo "separateTestsNames=$(./gradlew listTasksAsJSON -q --console=plain | head -n 1)" >> $GITHUB_OUTPUT
+        echo "separateTestsNames=$(./gradlew listTasksAsJSON -q --console=plain | tail -n 1)" >> $GITHUB_OUTPUT
 
   build:
     name: build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,57 +34,6 @@ jobs:
       matrix:
         jdk: [11, 17]
         platform: ["ubuntu-latest", "windows-latest"]
-    runs-on: ${{ matrix.platform }}
-
-    steps:
-    - name: Set up JDK for build and test
-      uses: actions/setup-java@v2
-      with:
-        distribution: temurin # Temurin is a distribution of adoptium
-        java-version: ${{ matrix.jdk }}
-
-    - name: Checkout security
-      uses: actions/checkout@v2
-
-    - name: Build and Test
-      uses: gradle/gradle-build-action@v2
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -209,13 +277,50 @@ task opensslTest(type: Test) {
...
+}
+
 task copyExtraTestResources(dependsOn: testClasses) {
     copy {
         from 'src/test/resources'
         into 'build/testrun/test/src/test/resources'
     }
 }
-tasks.test.dependsOn(copyExtraTestResources, opensslTest)
+tasks.test.dependsOn(copyExtraTestResources, opensslTest, taskNames)
```
willyborankin, line 323: Could you please exclude  `opensslTest` here since you already defined the `sslTest`  task for it?

build.gradle
```diff
@@ -152,12 +153,79 @@ tasks.whenTaskAdded {task ->
...
+            "org.opensearch.security.ssl.OpenSSL*",
+        ]
+    ],
+    indicesTest: [
+        includeTestsMatching: [
+            "org.opensearch.security.*indices*"
+        ],
+        excludeTestsMatching: [
+            "org.opensearch.security.sanity.tests.*",
+            "org.opensearch.security.ssl.OpenSSL*",
```
willyborankin, line 192: I think they appeared here due to the `opensslTest` task

build.gradle
```diff
@@ -152,12 +153,79 @@ tasks.whenTaskAdded {task ->
...
+            "org.opensearch.security.ssl.*"
+        ]
+    ],
+    securityIntegrationTest: [
+        includeTestsMatching: [
+            "org.opensearch.security.*Integ*"
+        ],
+        excludeTestsMatching: [
+            "org.opensearch.security.sanity.tests.*",
+            "org.opensearch.security.ssl.OpenSSL*",
```
willyborankin, line 183: I think they appeared here due to the `opensslTest` task

willyborankin (changes requested):
Hi @pawel-gudel-eliatra, thank you for the contribution. Could you please remove `opensslTest` task since you already defined the new one `sslTest`. I'm wondering how will it work in IDE? 
As I understand IDE starts all list of tasks in case you select any test or I'm wrong?

build.gradle
peternied: Can we remove the other tasks likes `opensslTest` and consolidate in a single way to filter/execute them?

build.gradle
```diff
@@ -152,12 +153,79 @@ tasks.whenTaskAdded {task ->
     }
 }
 
+def splitTestConfig = [
```
peternied, line 156: This looks highly manual in a way we could miss tests.  How hard would it be to split on test class if a number of parallel runners was provided?

.github/workflows/ci.yml
```diff
@@ -6,8 +6,30 @@ env:
...
+        java-version: 17
+
+    - name: Checkout security
+      uses: actions/checkout@v2
+
+    - name: Generate list of tasks
+      id: set-matrix
+      run: |
+        echo "separateTestsNames=$(./gradlew listTasksAsJSON -q --console=plain | tail -n 1)" >> $GITHUB_OUTPUT
+        echo "testsNamesExclude=$(./gradlew listTasksAsParam -q --console=plain | tail -n 1)" >> $GITHUB_OUTPUT
```
peternied, line 28: I'd really prefer we didn't have these exclusions as it spreads the responsiblity of executing the tests based on a matrix of tests and several gradle task, can we see about removing them and grouping all the tests together into a single test task?

build.gradle
nibix: This approach will never miss tests, as tests which are not matched by these patterns will fall in to the bucket that is executed by the general `test` task.

> How hard would it be to split on test class if a number of parallel runners was provided?

Not sure if I understand. Can you elaborate?

Generally, it should be noted that this is a quick-win approach as asked for by @davidlago . As described in https://github.com/opensearch-project/security/issues/2798#issuecomment-1568539046 we also would like to restructure the tests themselves to consume fewer resources. However, that won't be something we can achieve within a couple of weeks.

.github/workflows/ci.yml
nibix: Apologies, but I don't really understand this. Can you elaborate more?

If we had all tests in a single task, then how shall we achieve the parallel execution in `workflows/ci.yml`?

build.gradle
pawel-gudel-eliatra: This is only manual for test which we decided to split. Any new test classess will be containted within default "test" job.

build.gradle
pawel-gudel-eliatra: Thanks for noticing that. I have to dig deeper in how `**/OpenSSL*.class` correlates to test names. Split on `org.opensearch.security.ssl.OpenSSL*` has moved some of the long running tests to `sslTest` job but not all of them.

build.gradle
pawel-gudel-eliatra: Yes, I've checked that and we only needs `.sanity.tests.*` here. It's been already fixed.

build.gradle
pawel-gudel-eliatra: @willyborankin Fixed.

@peternied I've removed `opensslTest` as all of its tests are included in `sslTest` task.

build.gradle
nibix: We did some profiling before and found that the execution time is not really equally distributed over the classes and packages. There are some very heavy classes and some more lightweight ones. The partitioning in this PR has been chosen to balance the execution time.

scrawfor99: In line with what Peter mentioned, is there a way we can perform this in a load balanced way? I know you mentioned identifying the tests manually, but I am unclear on what will happen if new tests are added. Will they just go into the default catch all? If we can find a way to balance the runners based on load we could optimize performance even when new tests are added. 

One possibility would be to track the execution time of the previous run and direct the splitting script to move tests from one runner to another based on the last execution. This would lead to improved performance for small changes where no significant tests were added to the repo since we could use the roughly optimized version of the test loads from the previous run.

.github/workflows/ci.yml
```diff
@@ -6,8 +6,30 @@ env:
...
+        distribution: temurin # Temurin is a distribution of adoptium
+        java-version: 17
+
+    - name: Checkout security
+      uses: actions/checkout@v2
+
+    - name: Generate list of tasks
+      id: set-matrix
+      run: |
+        echo "separateTestsNames=$(./gradlew listTasksAsJSON -q --console=plain | tail -n 1)" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
nibix: I totally agree that maintaining this on a long term will be infeasible. I believe that this needs to be considered as a short-term approach to immediately treat the pain of the long CI running times.

But we should aim to replace this within 6 to 12 months by a long-term solution.

build.gradle
peternied: > One possibility would be to track... [Good idea]

tracking / analyzing / acting seems like next step(s) that we might want to revisit if this is a problem

nibix: I just want to double-check: Is this now a discussion about next steps or are these requests for this PR? :-)

.github/workflows/ci.yml
pawel-gudel-eliatra: @peternied I'm not sure whether you understand what is the point of this PR or even what was the problem in the first place. Please let me know if you need more information.

> Trigger a single test. This is broken today because the openssl tasks run even with the filter condition. I understand if this is out of scope and I'll write up a separate issue

This is broken and will become even more broken after this PR. It looks like this is broken by design because (old) _opensslTest_ was dependency of "test" task so it was always started when "test" was run. This PR will make this issue BIGGER because all separated tests are dependencies.

Simplest way to fix is the same as within old one - just _-x_ tasks which you do not want to run. This PR gives you a simple method:

`./gradlew test --tests org.opensearch.security.my.test $(./gradlew listTasksAsParam -q --console=plain | tail -n 1)`

> Trigger a subset of the tests as if I was a github action runner: ./gradlew dlicDlsflsTest - Pretty sure this is what you are doing now, but when I look at the way the gradle task is constructed there is alot of -x'ed items

We must exclude tests which are started from "test" task because we do not want them to be run twice. The point of this PR is to shorten time for test run on Github Actions and that would be pointless if we:
1. Start test from `org.opensearch.security.dlic.*`  by running `gradlew build dlicDlsflsTest`
2. Run them AGAIN as a part of `gradlew build test` command.

build.gradle
pawel-gudel-eliatra: > In line with what Peter mentioned, is there a way we can perform this in a load balanced way?

No.

> Will they just go into the default catch all?

Yes.

> If we can find a way to balance the runners based on load we could optimize performance even when new tests are added.

You are right BUT Github Actions does not have a interface to do that. We would have to:
1. Keep old test runs somewhere besides Github.
2. Analyze them (by what? what metric should be used?)
3. Somehow inject the new code into the gradle? (how?)
4. Who will maintain this Rube-Goldberg machine of configs and parsers?
We have issues with long-running tasks right now, adding this complicated system will take weeks if not months. Creating manual configuration is faster and easier and solves problem by a little bit. Remember that this is not a final solution. This is only a patch for currently-somewhat-broken configuration. Core issue is within tests itself, not with how they are run.

.github/workflows/ci.yml
nibix: @peternied 

That sounds very reasonable to me, we are looking into this at the moment.

I just have one question: We are a bit confused about the openssl test - why was it split into a separate task? It seems to complicate the management of the tests a bit. Do you have any preferences on how to handle this test?

build.gradle
peternied: Nothing in this thread is a requirement for merging from me. Even if we have 'imbalanced' or 'may-require-maintenance-in-time' solution its still a great improvement to accelerate time to merge

build.gradle
pawel-gudel-eliatra: I've restored openSSL tests because we don't know whether issues from #2301 are still affecting tests and I don't want to break existing workflows.

.github/workflows/ci.yml
peternied: With 33f8318 adding the OpenSSL tests into their own execution I'm going to resolve this thread as there is nothing I see for follow up

build.gradle
peternied: @scrawfor99 What do you think about resolving this thread?  I don't think this thread has much actionable that we'd like to see from @pawel-gudel-eliatra

Would you want to open up an follow up issue on the test rebalancing / monitoring?

build.gradle
peternied: We should remove the either `opensslTest` or `opensslCiTest`.  I would recommend only keeping `opensslCiTest` around

.github/workflows/ci.yml
```diff
@@ -6,13 +6,86 @@ env:
...
+        arguments: |
+          build citest -Dbuild.snapshot=false
+          -x integrationTest
+          -x spotlessCheck
+          -x checkstyleMain
+          -x checkstyleTest
+          -x spotbugsMain
+          -x spotbugsIntegrationTest
+          -x checkstyleIntegrationTest
+          -x test
```
peternied, line 61: _Pulling this out from another comment thread_

This command doesn't look right, I'd expect to only pass what actions should be run, not a giant list of things that shouldn't be run, this is error prone IMO.  Lets work backwards from these use case if we can

- Trigger a subset of the tests as if I was a github action runner:
      `./gradlew dlicDlsflsTest` - Pretty sure this is what you are doing now, but when I look at the way the gradle task is constructed there is alot of -x'ed items

- Trigger all tests subsets in parallel (on local machine):
       `./gradlew test`
This would use the gradle / junit task parallelization to run

- **[Nice to Have]** Trigger single test:
     `./gradlew test --tests org.opensearch.security.my.test`, this would only trigger that test alone, no matter what other tasks are in play. This is broken today because the openssl tasks run even with the filter condition. I understand if this is out of scope and I'll write up a separate issue, but I'd also like to be sure this problem isn't made worse with this change,

build.gradle
```diff
@@ -105,6 +106,83 @@ tasks.whenTaskAdded {task ->
...
+    doLast {
+        System.out.println(new JsonBuilder(taskNames))
+    }
+}
+
+task listTasksAsParam {
+    doLast {
+        System.out.println('-x ' + (taskNames).join(' -x '))
+    }
+}
```
peternied, line 185: These seem like hacks, can they be removed?

build.gradle
```diff
@@ -163,18 +241,123 @@ task opensslTest(type: Test) {
...
+        }
+    }
+
+    copy {
+        from 'src/test/resources'
+        into 'build/testrun/citest/src/test/resources'
+    }
+}
+
+task citest(type: Test) {
```
peternied, line 263: This task seems like it is duplicated with the section in 312, can we trim this down by declaring and reusing the test configuration or something like that?

peternied (changes requested):
I still need to see resolution with the openssl tests and how the the tasks are used and cleanup of the command line parameters.  Is there anything you'd need help with to make forward progress?

nibix:
@peternied 

> I still need to see resolution with the openssl tests

As there was lots of back and forth on the openssl tests, let me just verify what the resolution of the openssl tests would be. Is it just removing openssltest in favour of opensslcitest?

.github/workflows/ci.yml
nibix: > Trigger a subset of the tests as if I was a github action runner:

This is possible.

> Trigger all tests subsets in parallel (on local machine):

This is also now possible. We have achieved this by splitting off the `citest` task to keep `test` for exclusive use on local machines.

>  Trigger single test:

This is also possible. Since the restoration of the openssltest the old issue applies again that this test is also executed in addition.

But coming to the `-x` exclusions in general: These are necessary because the CI task uses `gradle build test`. `build` pulls in a lot of additional task dependencies. However, I personally see spontaneously no reason why `build` is necessary, we are just using this because it was used this way before. But I think we should try to change

```
build citest -Dbuild.snapshot=false
          -x integrationTest
          -x spotlessCheck
          -x checkstyleMain
          -x checkstyleTest
          -x spotbugsMain
          -x spotbugsIntegrationTest
          -x checkstyleIntegrationTest
          -x test
```

into 

```
citest -Dbuild.snapshot=false
```

and see if this works.

build.gradle
pawel-gudel-eliatra: No, it is not. `citest` is "generic" one which covers everything not covered by specific tests (declared in `splitTestConfig` object). Loop on line 312 is generating tasks for `splitTestConfig`.

build.gradle
pawel-gudel-eliatra: They are  and no, unless someone has better idea how to dynamically generate list of tasks for Github Actions "matrix" strategy [(link to relevant part of ci.yml)](https://github.com/pawel-gudel-eliatra/security/blob/split-ci-tests/.github/workflows/ci.yml#L24)

Why I've decided to use gradle for that:
1. Gradle has knowledge about all tasks. `build.gradle` is source of truth.
2. Generating list of tasks does not need external utilities like: Bash, Python etc.

.github/workflows/ci.yml
pawel-gudel-eliatra: @nibix It will work:
```
root@feaf17923c58:/app# ./gradlew citest --dry-run
=======================================
OpenSearch Build Hamster says Hello!
  Gradle Version        : 8.1.1
  OS Info               : Linux 6.3.8-200.fc38.x86_64 (amd64)
  JDK Version           : 17 (Eclipse Temurin JDK)
  JAVA_HOME             : /root/.sdkman/candidates/java/17.0.7-tem
  Random Testing Seed   : 3B71B6CF19C67EF0
  In FIPS 140 mode      : false
=======================================
:compileJava SKIPPED
:processResources SKIPPED
:classes SKIPPED
:compileTestJava SKIPPED
:copyPluginPropertiesTemplate SKIPPED
:pluginProperties SKIPPED
:processTestResources SKIPPED
:testClasses SKIPPED
:copyExtraTestResources SKIPPED
:citest SKIPPED
:opensslTest SKIPPED
:test SKIPPED
:jacocoTestReport SKIPPED

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

See https://docs.gradle.org/8.1.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 786ms
```
Unfortunately we have dependency between `jacocoTestReport` and `test` and that's why we had to use `-x test` in the first place.

build.gradle
peternied: Digging in `listTasksAsParam` doesn't look like it is used, so we can remove?

Do we need the `doLast` block? can the printing on the json string be at the high level e.g. 

```
task listTasksAsJSON {
    System.out.println(new JsonBuilder(taskNames))
}
```

build.gradle
peternied: True, what about the parameters after `filter`?  seems like there is 20+ lines of code that are repeated in 2+ places that makes the layout of these look similar and changes should be consistent across

pawel-gudel-eliatra:
> > removing openssltest in favour of opensslcitest
> 
> +1 it looks like that block of tests should be run on its own that's all

Does the issue in #2301 is fixed? Can you please point us to commit which fixes it? If it's not then we should not change the current workflow.

build.gradle
pchmielnik: > True, what about the parameters after `filter`? seems like there is 20+ lines of code that are repeated in 2+ places that makes the layout of these look similar and changes should be consistent across


I have extracted the parts related to configuration into a separate function called `setCommonTestConfig` in order to avoid code duplication.

.github/workflows/ci.yml
peternied: Any progress on this thread?

nibix: @peternied The `jacocoTestReport` seems to force explicit dependencies on all test tasks. See the execution log above, the `citest` triggers `jacocoTestReport` which in turn triggers `test` and `opensslTest` as a dependency. That makes the exclusions necessary. So far, we did not find a reasonable way to avoid that.

build.gradle
pawel-gudel-eliatra: > Digging in listTasksAsParam doesn't look like it is used, so we can remove?
Removed.

> Do we need the doLast block?
Without `doLast` output was sometimes mixed with other tasks output. Removed per your request.

build.gradle
pawel-gudel-eliatra: @pchmielnik's changes has been applied on repository.

build.gradle
pawel-gudel-eliatra: Restored `doLast` because Github Task were not working properly.

.github/workflows/ci.yml
peternied: Could an issue be made to track this for circling back onto at a later time?

build.gradle
peternied: Could you add a comment?  IMO this a good case for comments to explain what problem this addresses since its non-obvious

build.gradle
scrawfor99: hi @peternied, definitely. We can follow up in a separate PR. I do not want to slow @pawel-gudel-eliatra's great work with this.

build.gradle
pawel-gudel-eliatra: @peternied Removed.

build.gradle
pawel-gudel-eliatra: Added.

.github/workflows/ci.yml
nibix: Done: https://github.com/opensearch-project/security/issues/2913

peternied (approved):
Total duration
:fire: [24m 54s](https://github.com/opensearch-project/security/actions/runs/5399528370/usage) :fire:

peternied:
@willyborankin anything else you are looking for out of this PR?

willyborankin (approved):
LGTM

peternied:
@pawel-gudel-eliatra Would you be open to create a backport PR to the 2.x branch of this repository?  The instructions for that process are in the post [[above](https://github.com/opensearch-project/security/pull/2861#issuecomment-1612078394)] this one.

PR #2866, opened 2023-06-15
Finishes reformatting files by formatting all *.java files outside of **/test && **/integrationTest

scrawfor99:
### Description
Follows up #2863

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
+++ b/src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
@@ -108,10 +108,7 @@ public class AuditConfig {
     }
 
     @VisibleForTesting
-    public AuditConfig(
-            final boolean auditLogEnabled,
-            final Filter filter,
-            final ComplianceConfig compliance) {
+    public AuditConfig(final boolean auditLogEnabled, final Filter filter, final ComplianceConfig compliance) {
         this.auditLogEnabled = auditLogEnabled;
         this.filter = filter != null ? filter : Filter.DEFAULT;
         this.compliance = compliance != null ? compliance : ComplianceConfig.DEFAULT;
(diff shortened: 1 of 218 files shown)


reta (approved):

src/main/java/org/opensearch/security/auditlog/impl/AuditCategory.java
```diff
@@ -33,13 +33,8 @@ public enum AuditCategory {
     COMPLIANCE_INTERNAL_CONFIG_WRITE;
 
     public static Set<AuditCategory> parse(final Collection<String> categories) {
-        if (categories.isEmpty())
-            return Collections.emptySet();
+        if (categories.isEmpty()) return Collections.emptySet();
```
peternied, line 36: 🫠 I would prefer we always have curly brackets on if statements - can this be tweaked?

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
```diff
@@ -400,58 +444,65 @@ public void log(Logger logger) {
...
-                    ", isTransportApiAuditEnabled=" + isTransportApiAuditEnabled +
-                    ", disabledTransportCategories=" + disabledTransportCategories +
-                    ", resolveBulkRequests=" + resolveBulkRequests +
-                    ", logRequestBody=" + logRequestBody +
-                    ", resolveIndices=" + resolveIndices +
-                    ", excludeSensitiveHeaders=" + excludeSensitiveHeaders +
-                    ", ignoredAuditUsers=" + ignoredAuditUsersMatcher +
-                    ", ignoreAuditRequests=" + ignoredAuditRequestsMatcher +
-                    '}';
+            return "Filter{"
```
peternied, line 447: (Comment on the code) This is a good case for a string builder with format appends for variable pairs. This is terrible performance and heap utilization to allocation each temp substring.

peternied (approved):
While I've got nit this is great let's merge and we can tackle the small stuff in spectate PRs - great work

src/main/java/org/opensearch/security/auditlog/impl/AuditCategory.java
scrawfor99: Yeah, I can look into modifying the formatting file to include that!

scrawfor99: Let me know if you think of any other style guides you would like.

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
scrawfor99: Should we create an issue here?

peternied: I don't think so. While there are significant drawbacks to its implementation - it's working code. 

Worth revisiting if they were reports of customer impact

PR #2681, opened 2023-04-14
Authorize system index permissions in SecurityIndexAccessEvaluator

samuelcostae:
### Description

SecurityIndexAccessEvaluator was modified  to check  if an accounting trying to access a system index specifically has the "system:admin/system_index" permission for the index, and block the access if it doesn't.

A Permission needs to be set explicitly and a "*" permission will not allow access to the sistem indices.

* Category: Enhancement

* Why these changes are required?
Extensions do not have the ability to elevate privileges, which means they cannot access system indexes. To overcome this limitation, extensions will require a way to access and store protected metadata.

* What is the old behavior before changes and new behavior after changes?
     - Before: Plugins would elevate privileges to access system indices
     - After: An account needs to have the additional index permission "system:admin/system_index"  in order to modify indices defined as "System Indices".

### Issues Resolved
#2553 

### Testing
Unit Testing - Added tests to SystemIndexTests.java
Draft PR - Existing tests still need to be updated to account for new beahaviour

### Check List
- [x] New functionality includes testing
- [ ] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #2845, opened 2023-06-08
Add configuration flag to disable system index permissions


peternied:
#### Problem:
System index permissions allow access to data on clusters that was previously not accessible.  For services that operate managed OpenSearch clusters they could want to restriction based on custom index / plugins configuration.

#### Proposed solution:
Check for a value in the security config something like "system index permissions disabled".
- For existing clusters that migrate to the latest version, the features should be disabled by default.
- For new clusters, the feature should be enabled by default.

scrawfor99:
[Triage] Hi @peternied, this sounds like a good request considering the discussions around system indices. I think that @cwperks will have some suggestions on what may fall under this as well and may want to follow up here.

PR #2887, opened 2023-06-20
System Index permission check

samuelcostae:
**Description**

This PR is based on the work done on https://github.com/opensearch-project/security/pull/2681/ .  As the branch got stale and there were some issues with the sign-off of the commits, I created a new clean branch  and generating a new PR

SecurityIndexAccessEvaluator was modified to check if an accounting trying to access a system index specifically has the "system:admin/system_index" permission for the index, and block the access if it doesn't.

A Permission needs to be set explicitly and a "*" permission will not allow access to the sistem indices.

Category: Enhancement

Extensions do not have the ability to elevate privileges, which means they cannot access system indexes. To overcome this limitation, extensions will require a way to access and store protected metadata.

What is the old behavior before changes and new behavior after changes?

Before: Plugins would elevate privileges to access system indices
After: An account needs to have the additional index permission "system:admin/system_index" in order to modify indices defined as "System Indices".
Issues Resolved
https://github.com/opensearch-project/security/issues/2553

**Testing**
Unit Testing - Added tests to SystemIndexTests.java

**Check List**
 

- [ ]  New functionality has been documented

- [x]  Commits are signed per the DCO using --signoff


src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -105,9 +113,22 @@ public PrivilegesEvaluatorResponse evaluate(
...
+        ConfigModelV7.SecurityRoles securityRoles
     ) {
+
         final boolean isDebugEnabled = log.isDebugEnabled();
+
+        if (matchAnySystemIndices(requestedResolved) && !checkSystemIndexPermissionsForUser(securityRoles)) {
+            log.warn(
+                "An account without the {} permission  is trying to access a System Index. Related indexes: {}",
+                ConfigConstants.SYSTEM_INDEX_PERMISSION,
+                getProtectedIndexes(requestedResolved).stream().collect(Collectors.joining(", "))
```
peternied, line 126: Guard the info message in `log.isInfoEnabled()` so we don't take the perf hit assembling the message even if it isn't used

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -122,13 +143,13 @@ public PrivilegesEvaluatorResponse evaluate(
...
-                    presponse.allowed = false;
-                    return presponse.markComplete();
                 }
-            } else if (matchAnySystemIndices(requestedResolved)) {
+                auditLog.logSecurityIndexAttempt(request, action, task);
+                log.warn("{} for '_all' indices is not allowed for a regular user", action);
+                presponse.allowed = false;
+                return presponse.markComplete();
+            }
+            if (matchAnySystemIndices(requestedResolved) && !checkSystemIndexPermissionsForUser(securityRoles)) {
```
peternied, line 152: Seems like this check won't be hit now, the condition is the same as on 122, no?

src/test/resources/roles_system_indices.yml
```diff
@@ -12,3 +12,25 @@ opendistro_security_all_access:
...
+extensions_role:
+  description: "ExtensionsRole"
+  cluster_permissions:
+    - '*'
+  index_permissions:
+    - index_patterns:
+        - '*'
+      allowed_actions:
+        - '*'
+        - 'system:admin/system_index'
```
peternied, line 25: If the index pattern was `.k*` would only system indices that start with the letter k be allowed, so `.alerting` would be denied access, but `.knn` would be allowed?  Can we get a test that confirms this?

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -175,6 +196,21 @@ public PrivilegesEvaluatorResponse evaluate(
         return presponse;
     }
 
+    private boolean checkSystemIndexPermissionsForUser(ConfigModelV7.SecurityRoles securityRoles) {
+        Set<WildcardMatcher> userPermMatchers = securityRoles.getRoles()
+            .stream()
+            .flatMap(role -> role.getIpatterns().stream())
+            .map(ConfigModelV7.IndexPattern::getNonWildCardPerms)
```
peternied, line 203: Comment would be good for how this mapping should affect the results

peternied (changes requested):
Thanks for the contribution.  I've added several structural comments and a broader question about how the permission works with index pattern matching.

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -175,6 +196,21 @@ public PrivilegesEvaluatorResponse evaluate(
         return presponse;
     }
 
+    private boolean checkSystemIndexPermissionsForUser(ConfigModelV7.SecurityRoles securityRoles) {
+        Set<WildcardMatcher> userPermMatchers = securityRoles.getRoles()
```
scrawfor99, line 200: You can rename this variable to something like "UserPermissions" to avoid needing to add a comment.

src/test/java/org/opensearch/security/system_indices/SystemIndicesTests.java
```diff
@@ -269,29 +311,45 @@ public void testDelete() throws Exception {
     }
 
     @Test
-    public void testDeleteWithSystemIndices() throws Exception {
+    public void testDeleteDocWithSystemIndicesShouldSucceedAsSuperAdmin() throws Exception {
```
scrawfor99, line 314: NIT: I like the informative name but I think this will be a bit long for consoling logging. You will get the full path `src/test/java/org/opensearch/security/system_indices/SystemIndicesTests.testDeleteDocWithSystemIndicesShouldSucceedAsSuperAdmin` which is pretty long...

src/test/java/org/opensearch/security/system_indices/SystemIndicesTests.java
```diff
@@ -269,29 +311,45 @@ public void testDelete() throws Exception {
...
             RestHelper.HttpResponse responseDoc = keyStoreRestHelper.executeDeleteRequest(index + "/_doc/document1");
             assertEquals(RestStatus.OK.getStatus(), responseDoc.getStatusCode());
 
             RestHelper.HttpResponse responseIndex = keyStoreRestHelper.executeDeleteRequest(index);
             assertEquals(RestStatus.OK.getStatus(), responseIndex.getStatusCode());
         }
+    }
+
+    @Test
+    public void testDeleteDocWithSystemIndicesShouldFailsAsAdmin() throws Exception {
```
scrawfor99, line 329: Same as above.

scrawfor99 (commented):
I think this is good to go once Peter's comments are addressed. I did not see anything that he did not comment  on that looked blocking.

DarshitChanpura (dismissed):
FYI for maintainers: This is undergoing security review and will be merged once the review is complete.

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -105,9 +113,22 @@ public PrivilegesEvaluatorResponse evaluate(
...
         final Resolved requestedResolved,
-        final PrivilegesEvaluatorResponse presponse
+        final PrivilegesEvaluatorResponse presponse,
+        ConfigModelV7.SecurityRoles securityRoles
     ) {
+
         final boolean isDebugEnabled = log.isDebugEnabled();
+
+        if (matchAnySystemIndices(requestedResolved) && !checkSystemIndexPermissionsForUser(securityRoles)) {
+            log.warn(
```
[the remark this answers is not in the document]
samuelcostae: Cool.

Other similar error logs in the SecurityIndexAccessEvaluator are also set to _warn_ level, i will update them as well

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
samuelcostae: Again a result of conflicts/ reapplying old commits. Thanks for pointing it out

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -308,7 +309,14 @@ public PrivilegesEvaluatorResponse evaluate(
...
 
         // Security index access
-        if (securityIndexAccessEvaluator.evaluate(request, task, action0, requestedResolved, presponse).isComplete()) {
+        if (securityIndexAccessEvaluator.evaluate(
+            request,
+            task,
+            action0,
+            requestedResolved,
+            presponse,
+            (ConfigModelV7.SecurityRoles) securityRoles
```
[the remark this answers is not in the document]
samuelcostae: I did a push with those changes. I've renamed the ConfigModelV6's and ConfigModelV7 inner classes to include V6/V7 in their name to make it clearer and avoid having to fully qualify them to differentiate from the interface.

The interface implementation is a bit rough still and could do with some cleanup, so i appreciate any feedback on that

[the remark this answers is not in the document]
samuelcostae: Just adding comment from today's call here, Nils said that  the 'V6' model refered to Elastic Search 6 and should'nt really be being used by anybody anymore, and suggested to deprecate it.

src/test/java/org/opensearch/security/system_indices/SystemIndicesTests.java
```diff
@@ -326,27 +384,61 @@ public void testCloseOpen() throws Exception {
...
+            Assert.assertTrue(
+                responseClose.getBody()
+                    .contains(
+                        "{\"type\":\"security_exception\",\"reason\":\"no permissions for [] and User [name=admin_all_access, backend_roles=[], requestedTenant=null]\"}"
+                    )
+            );
 
             RestHelper.HttpResponse responseOpen = sslRestHelper.executePostRequest(index + "/_open", "", allAccessUserHeader);
+            assertEquals(RestStatus.FORBIDDEN.getStatus(), responseOpen.getStatusCode());
+            Assert.assertTrue(
```
peternied, line 421: Replace with `assertThat` using `containsString`, this will print the error message if it didn't match.

Alternatively, use `assertThat(responseOpen.getBody().findValueInJson(".reason"), equalTo("no permissions for [] and User [name=admin_all_access, backend_roles=[], requestedTenant=null]"))` so the test is more readable.

Note; occurs in several places, assertTrue/assertFalse are an anti-pattern

src/test/java/org/opensearch/security/system_indices/SystemIndicesTests.java
```diff
@@ -549,21 +690,61 @@ public void testSnapshotWithSystemIndices() throws Exception {
                 sslRestHelper.executeGetRequest("_snapshot/" + index + "/" + index + "_1", allAccessUserHeader).getStatusCode()
             );
             assertEquals(
-                HttpStatus.SC_OK,
+                HttpStatus.SC_FORBIDDEN,
```
peternied, line 693: Why did this change? Isn't this backwards incompatible

src/test/java/org/opensearch/security/system_indices/SystemIndicesTests.java
samuelcostae: **assertTrue/assertFalse are an anti-pattern**

 Very true! I think I was awake for too long when writing those tests  xD

DarshitChanpura:
QQ: Are we preventing access to security index? If not should there be a follow-up that prevents access to security-index with this permission scheme?

Also @samuelcostae Could you please address CI failures?

cwperks:
> QQ: Are we preventing access to security index? If not should there be a follow-up that prevents access to security-index with this permission scheme?

IMO Security index should only ever be limited to requests from the admin certificate. The admin certificate is configured in `opensearch.yml` and is not dependent on the security index at all. i.e. If a user connecting with admin certificate deletes the security index, they can connect again in a following request to create the security index.

Any other user would not be able to reconnect after the security index is nuked because their permissions rely on the security index.

DarshitChanpura:
Another question @samuelcostae, are we toggling this feature with a feature flag or is this enabled by default?

samuelcostae:
> 

Its on by default, I had added a feature flag in the past, but i got asked  to remove it

samuelcostae:
> QQ: Are we preventing access to security index? If not should there be a follow-up that prevents access to security-index with this permission scheme?
> 
> Also @samuelcostae Could you please address CI failures?



> QQ: Are we preventing access to security index? If not should there be a follow-up that prevents access to security-index with this permission scheme?
> 
> Also @samuelcostae Could you please address CI failures?

I've made no special exceptions for the security index, so if it is marked as a system index (which i believe it is by default), an account should only be able to access it **if it has the security index explicitly added in its list of indices and the "system:admin/system_index" permission**.

DarshitChanpura:
> Its on by default, I had added a feature flag in the past, but it was suggested to be removed

There is a follow-up issue created to address this: https://github.com/opensearch-project/security/issues/2845
I'm not sure if the discussion was done before creating this issue or after. But IMO we should address it in this PR. @peternied @cwperks @scrawfor99 @RyanL1997 thoughts?

src/main/java/org/opensearch/security/securityconf/SecurityRole.java
```diff
@@ -0,0 +1,56 @@
+/*
```
davidlago, line 1: Same comment about headers

src/main/java/org/opensearch/security/securityconf/IndexPattern.java
```diff
@@ -0,0 +1,69 @@
+/*
```
davidlago, line 1: Same comment about headers

davidlago (dismissed):

samuelcostae:
@scrawfor99  @DarshitChanpura  @peternied  i was thinking... and considering  https://github.com/opensearch-project/security/issues/3038 we did a bit of a turn around here.

 "System Indices"(though a customisable list) was basically used to protect the _.opendistro_security index_  

We decided to use the "system indices" for the extensions, keeping  it generic so it could continue be used for the _.opendistro_security index_  .

Now we are creating a new type "denylist index" to separate  the _.opendistro_security  from the system indices. So we end up with 2 different type of indices with higher protection anyway"

Considering the above, it seems to me it would make sense to leave the system_indices as they are currently, which would avoid having to modify existing tests and use a new list (Either called "extensions indices" or something else)  for the extensions.

Does that make sense?

peternied:
@samuelcostae Sorry I think I might have missed some context, it sounds like are proposing changes to the end to end scenarios that are supported, previously this was:

> - The new permissions applies to the indexes that match its index pattern, no other system indexes.
> - The new permissions allows for read and write operations on the permitted index(es)
> - (Should already exist, if not add) Make sure that '*' allowed actions doesn't automatically give '{NEW PERMISSION R/W SYSTEM INDEX}' access

As this has been discussed back and forth it sounds like there are additional scenario(s):

- The new permissions when applied to the security index does not allow access, the users still gets permissions denied.

Are there other scenarios that should be added, or should we modify the previous set?

build.gradle
```diff
@@ -262,7 +262,8 @@ def setCommonTestConfig(Test task) {
                 "sun.security.jgss.*",
                 "sun.security.pkcs11.*",
                 "sun.security.smartcardio.*",
-                "sun.util.resources.provider.*"
+                "sun.util.resources.provider.*",
+                "org/opensearch/security/securityconf/impl/v6.*"
```
peternied, line 266: Remove this exclusion, there is lots of code that could be the source of serious bugs if not tested properly.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1783,6 +1783,23 @@ public List<Setting<?>> getSettings() {
...
+                Setting.boolSetting(
+                    ConfigConstants.SECURITY_SYSTEM_INDICES_ADDITIONAL_CONTROL_ENABLED_KEY,
+                    ConfigConstants.SECURITY_SYSTEM_INDICES_ADDITIONAL_CONTROL_ENABLED_DEFAULT,
+                    Property.NodeScope,
+                    Property.Filtered
+                )
+            );
+            settings.add(
+                Setting.listSetting(
+                    ConfigConstants.SECURITY_INDICES_DENYLIST_KEY,
```
peternied, line 1796: This should not be a configurable setting, lets delete pulling this from configuration

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -71,11 +75,13 @@ public SecurityIndexAccessEvaluator(final Settings settings, AuditLog auditLog,
         this.systemIndexMatcher = WildcardMatcher.from(
             settings.getAsList(ConfigConstants.SECURITY_SYSTEM_INDICES_KEY, ConfigConstants.SECURITY_SYSTEM_INDICES_DEFAULT)
         );
+        this.denylistIndexMatcher = WildcardMatcher.from(
+            settings.getAsList(ConfigConstants.SECURITY_INDICES_DENYLIST_KEY, ConfigConstants.SECURITY_INDICES_DENYLIST_KEY_DEFAULT)
+        );
```
peternied, line 80: RE: The deny list should not be configurable
We can keep the deny list index matcher, but populate it with only the value for the security index.

src/test/java/org/opensearch/security/auditlog/compliance/ComplianceAuditlogTest.java
```diff
@@ -56,6 +56,7 @@ public void testSourceFilter() throws Exception {
         Settings additionalSettings = Settings.builder()
             .put("plugins.security.audit.type", TestAuditlogImpl.class.getName())
             .put(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_TRANSPORT, true)
+            .put(ConfigConstants.SECURITY_SYSTEM_INDICES_ADDITIONAL_CONTROL_ENABLED_KEY, false)
```
peternied, line 59: Looks like many tests were updated, like this one, why this change?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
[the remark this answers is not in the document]
peternied: We should not diverge the behavior if it can be avoided, switching from Generic -> Specific Impl is a big code change and its harder to review.

Lets not try to solve the V6 vs V7 issue on this pull request, what do you think about making a separate issue for follow up?

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
samuelcostae: Just to be sure, we are updating the requirements  described on issue https://github.com/opensearch-project/security/issues/3038

src/test/java/org/opensearch/security/auditlog/compliance/ComplianceAuditlogTest.java
samuelcostae: The system already defaults to false, but i thought would be better to be explicit on the tests that are expecting it to be false.

peternied: These tests that were update don't seem to have any new behavior or assertions, why add this change?

peternied:
@samuelcostae It looks like you are continuing to work on this - thanks.  When you are ready for another review please make sure to respond to this [[comment]](https://github.com/opensearch-project/security/pull/2887#issuecomment-1662572660).  Without knowing the specific scenarios you are targeting it will be hard to approve this pull request.

src/test/java/org/opensearch/security/auditlog/compliance/ComplianceAuditlogTest.java
samuelcostae: Ah, I see what you mean now. Those were in tests that i was debugging and made changes before the flag was added but didnt not notice that i had reverted my changes on them.  I removed the referenced lines

samuelcostae:
> * not allow access, the users still gets permissions denied.

No, what i meant was just changing the naming of the 'bucket' containing the indices, just to have less parts of the code touched.

This is a paint drawing of what i meant   

image


But that is no big deal, we can proceed as is, mainly now that the "denylist" is just the security index and not a "open bucket". You can disregard my comment.

peternied:
@samuelcostae While we are on the final stretch reviewing this change, please do not rebase, but instead merge from main, it will protect the comment threads so they are early to see in the files view of the pull request.

src/test/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluatorTest.java
```diff
@@ -54,17 +54,36 @@ public class SecurityIndexAccessEvaluatorTest {
...
     private SecurityIndexAccessEvaluator evaluator;
-
     private static final String UNPROTECTED_ACTION = "indices:data/read";
     private static final String PROTECTED_ACTION = "indices:data/write";
+    @Mock
+    ConfigModelV7 configModelV7;
+    @Mock
+    ConfigModelV7.SecurityRoles securityRoles;// = configModelV7.getSecurityRoles();
+
+    public void setupEvaluatorWithSystemIndicesControl() {
```
peternied, line 65: Nit: setupEvaluatorWithSystemIndicesControl is nearly identical setupEvaluatorWithoutSystemIndicesControl, take a parameter?

src/test/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluatorTest.java
```diff
@@ -77,70 +96,125 @@ public void before() {
...
         // Action
-        evaluator.evaluate(request, null, UNPROTECTED_ACTION, resolved, presponse);
-        evaluator.evaluate(searchRequest, null, UNPROTECTED_ACTION, resolved, presponse);
-        evaluator.evaluate(realtimeRequest, null, UNPROTECTED_ACTION, resolved, presponse);
+        evaluator.evaluate(request, null, UNPROTECTED_ACTION, resolved, presponse, securityRoles);
+        evaluator.evaluate(searchRequest, null, UNPROTECTED_ACTION, resolved, presponse, securityRoles);
+        evaluator.evaluate(realtimeRequest, null, UNPROTECTED_ACTION, resolved, presponse, securityRoles);
 
-        verifyNoInteractions(presponse);
+        // verifyNoInteractions(presponse);
```
peternied, line 136: Why is this commented out?

src/test/java/org/opensearch/security/IntegrationTests.java
```diff
@@ -1001,7 +1001,7 @@ public void testNoDnfof() throws Exception {
 
     @Test
     public void testSecurityIndexSecurity() throws Exception {
-        setup();
+        setup(Settings.builder().put(ConfigConstants.SECURITY_SYSTEM_INDICES_ADDITIONAL_CONTROL_ENABLED_KEY, false).build());
```
peternied, line 1004: Isn't false the default, can we revert this change?

peternied: Same for the other changes in this file

src/test/java/org/opensearch/security/SnapshotRestoreTests.java
```diff
@@ -731,14 +731,12 @@ public void testSnapshotRestore() throws Exception {
...
             + "}";
 
-        Assert.assertEquals(
-            HttpStatus.SC_OK,
-            rh.executePutRequest(
-                "_snapshot/bckrepo/" + putSnapshot.hashCode() + "?wait_for_completion=true&pretty",
-                putSnapshot,
-                encodeBasicHeader("snapresuser", "nagilum")
-            ).getStatusCode()
+        RestHelper.HttpResponse response = rh.executePutRequest(
```
peternied, line 734: Looks unrelated, can you revert?

build.gradle
```diff
@@ -261,7 +261,7 @@ def setCommonTestConfig(Test task) {
                 "sun.security.jgss.*",
                 "sun.security.pkcs11.*",
                 "sun.security.smartcardio.*",
-                "sun.util.resources.provider.*"
```
peternied, line 264: Looks unrelated, please revert

src/test/java/org/opensearch/security/system_indices/DenyIndicesTests.java
```diff
@@ -0,0 +1,166 @@
...
+            );
+        }
+
+        // search all indices
+        RestHelper.HttpResponse response = restHelper.executePostRequest("/_search", matchAllQuery, allAccessUserHeader);
+        assertEquals(RestStatus.OK.getStatus(), response.getStatusCode());
+    }
+
+    @Test
+    public void testSearchWithDenyIndicesShouldSuccedAsAdminNoAdditinalAccesslControl() throws Exception {
```
peternied, line 150: Run a spell checker on all files, e.g. Succed -> Succeed, expectecdHits -> expectedHits

src/test/java/org/opensearch/security/system_indices/DenyIndicesTests.java
```diff
@@ -0,0 +1,166 @@
...
+    private static final List<String> listOfIndexesToTest = Arrays.asList(".opendistro_security");
+    private static final String matchAllQuery = "{\n\"query\": {\"match_all\": {}}}";
+    private static final String allAccessUser = "admin_all_access";
+    private static final Header allAccessUserHeader = encodeBasicHeader(allAccessUser, allAccessUser);
+    private static final String generalErrorMessage = String.format(
+        "no permissions for [] and User [name=%s, backend_roles=[], requestedTenant=null]",
+        allAccessUser
+    );
+
+    private void setupDenyIndicesEnabledWithSsl(Boolean securedIndicesAdditionalControlenable) throws Exception {
```
peternied, line 53:
```suggestion
    private void setup(boolean isSystemIndexPermissionEnabled) throws Exception {
```

src/test/java/org/opensearch/security/system_indices/DenyIndicesTests.java
```diff
@@ -0,0 +1,166 @@
...
+            .createParser(NamedXContentRegistry.EMPTY, LoggingDeprecationHandler.INSTANCE, response.getBody());
+        SearchResponse searchResponse = SearchResponse.fromXContent(xcp);
+        assertEquals(RestStatus.OK, searchResponse.status());
+        assertEquals(expectecdHits, searchResponse.getHits().getHits().length);
+        assertEquals(0, searchResponse.getFailedShards());
+        assertEquals(5, searchResponse.getSuccessfulShards());
+    }
+
+    @Test
+    public void testSearchWithDenyIndicesAsSuperAdmin() throws Exception {
```
peternied, line 113: This name is confusing, seems like it is testSuperAdminAccessSystemIndices is what is being tested.  There should also be a second test that confirms the behavoir is the same with the feature on

src/test/java/org/opensearch/security/system_indices/DenyIndicesTests.java
```diff
@@ -0,0 +1,166 @@
...
+        setupDenyIndicesEnabledWithSsl(true);
+        RestHelper restHelper = keyStoreRestHelper();
+
+        // search system indices
+        for (String index : listOfIndexesToTest) {
+            validateSearchResponse(restHelper.executePostRequest(index + "/_search", matchAllQuery), 10);
+        }
+
+        // search all indices
+        RestHelper.HttpResponse response = restHelper.executePostRequest("/_search", matchAllQuery);
```
peternied, line 123: This is the same query that is run on 119?

src/test/java/org/opensearch/security/system_indices/DenyIndicesTests.java
```diff
@@ -0,0 +1,166 @@
...
+        assertEquals(RestStatus.OK.getStatus(), response.getStatusCode());
+    }
+
+    @Test
+    public void testSearchWithDenyIndicesShouldFailAsAdmin() throws Exception {
```
peternied, line 128: testUserWithStarPermissionsCannotAccessSystemIndices

peternied: Also a second test with the freature disabled

src/test/java/org/opensearch/security/system_indices/DenyIndicesTests.java
```diff
@@ -0,0 +1,166 @@
...
+        );
+    }
+
+    /**
+     * Creates a set of test indices and indexes one document into each index.
+     *
+     * @throws Exception
+     */
+
+    private RestHelper keyStoreRestHelper() {
```
peternied, line 81: rename to useSuperAdminAuthenication?

src/test/java/org/opensearch/security/system_indices/SystemIndicesTests.java
```diff
@@ -58,10 +61,16 @@ public class SystemIndicesTests extends SingleClusterTest {
         allAccessUser
     );
 
+    private static final String extensionUser = "extensions_user";
```
peternied, line 64: Don't name these `extension`, it isn't needed.  this feature isn't limited to extensions

src/test/java/org/opensearch/security/system_indices/SystemIndicesTests.java
```diff
@@ -58,10 +61,16 @@ public class SystemIndicesTests extends SingleClusterTest {
...
+    private static final Header extensionUserHeader = encodeBasicHeader(extensionUser, allAccessUser);
+    private static final String extensionUserC = "extensions_user_c";
+    private static final Header extensionUserCHeader = encodeBasicHeader(extensionUserC, allAccessUser);
+
     private void setupSystemIndicesDisabledWithSsl() throws Exception {
 
         Settings systemIndexSettings = Settings.builder()
             .put(ConfigConstants.SECURITY_SYSTEM_INDICES_ENABLED_KEY, false)
+            .put(ConfigConstants.SECURITY_SYSTEM_INDICES_ADDITIONAL_CONTROL_ENABLED_KEY, false)
```
peternied, line 73: Lets try to change as few of these existing tests as possible, revert the whole file, and make new tests that are specifically oriented towards the new scenarios you've added.

Otherwise it isn't clear if we are changing existing behavoir, which should be the same using the default settings.

src/test/java/org/opensearch/security/system_indices/SystemIndicesTests.java
```diff
@@ -237,26 +255,60 @@ public void testSearchWithSystemIndicesAsAdmin() throws Exception {
         assertEquals(0, searchResponse.getHits().getHits().length);
     }
 
+    @Test
+    public void testSearchInOwnSystemIndicesShouldSucceedAsExtensionUser() throws Exception {     // CS-SUPRESS-ALL: Legacy code to be
```
peternied, line 259: Remove this CS-Sup?

peternied (commented):
Please structure the tests to be as easy to read as possible - decoupled from the existing tests, otherwise its hard to tell that the feature works as intended or didn't regress existing behavoir

peternied:
I am tagging this for v2.10 as it is a candidate, but code freeze is tomorrow EOD

src/test/java/org/opensearch/security/system_indices/DenyIndicesTests.java
samuelcostae: _"This name is confusing, seems like it is testSuperAdminAccessSystemIndices is what is being tested"_

It is testing DenyIndices. The comment  "        // search system indices" inside the method is a copy/past error.


_There should also be a second test that confirms the behavoir is the same with the feature on_

This test is with the feature on. 
I didn't create a separate test with the feature off because SuperAdmin should have access in both scenarios, so i just used the more stringent scenario.

src/test/java/org/opensearch/security/system_indices/SystemIndicesTests.java
samuelcostae: i used extensions for the tests cause is a concrete scenario. What you think i should name them? systemIndexUser?

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -98,16 +102,167 @@ public SecurityIndexAccessEvaluator(final Settings settings, AuditLog auditLog,
...
-        final PrivilegesEvaluatorResponse presponse
+        final PrivilegesEvaluatorResponse presponse,
+        final SecurityRoles securityRoles
     ) {
         final boolean isDebugEnabled = log.isDebugEnabled();
+
+        // As per issue #2845, the legacy access control to indices with additional protection should be kept in place in the meantime and
+        // be the default.
+        if (systemIndicesAdditionalControlFlag) {
+            evaluateNewSecuredIndicesAccess(action, requestedResolved, request, task, presponse, securityRoles, isDebugEnabled);
```
cwperks, line 124: Why `new` or `legacy` here? Can the flag be a param of the method `evaluateSystemIndicesAccess(..., permissionableSystemIndicesEnabled)`?

IMO this is a setting that users can toggle on or of. New/legacy don't really carry that much meaning and I wouldn't consider disabled permission-able system indices (the default behavior) as legacy.

samuelcostae: This is more a featureflag than a settings for users to turn on and off.  

Initially it was supposed to be a "flat modification", but over time  it was decided to include a flag, as we are adding additional constraints, so certain accounts could lose access to those indices.

DarshitChanpura:
To help get this across finish line for 2.10, I'm helping @samuelcostae clean up formatting and rewrite some tests.

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
scrawfor99: I am also not sure I understand why it is being separated--can you clarify Sam? I am sure you did it for a reason but I don't see why there are two methods right now.

src/main/java/org/opensearch/security/securityconf/IndexPattern.java
```diff
@@ -0,0 +1,54 @@
...
+
+    WildcardMatcher getPerms();
+
+    Set<String> getStringPerm();
+
+    WildcardMatcher getNonWildCardPerms();
+
+    String getDlsQuery(User user);
+
+    Set<String> concreteIndexNames(User user, IndexNameExpressionResolver resolver, ClusterService cs);
```
scrawfor99, line 35: Nit: Can we rename this? I am not sure what a "concrete" index name means. I assume that it is the full name?

src/main/java/org/opensearch/security/securityconf/ConfigModelV6.java
```diff
@@ -791,6 +800,11 @@ public String getDlsQuery(User user) {
             return replaceProperties(dlsQuery, user);
         }
 
+        @Override
+        public Set<String> concreteIndexNames(User user, IndexNameExpressionResolver resolver, ClusterService cs) {
```
scrawfor99, line 804: Nit: Same nit as below. Can we swap the inherited method to a different name if this is something added in this PR? I don't understand the concrete terminology here and I think we may be able to find a more intuitive convention.

src/main/java/org/opensearch/security/securityconf/ConfigModelV6.java
```diff
@@ -799,69 +813,39 @@ public Set<String> getMaskedFields() {
...
-            final int prime = 31;
-            int result = 1;
-            result = prime * result + ((perms == null) ? 0 : perms.hashCode());
-            result = prime * result + ((typeMatcher == null) ? 0 : typeMatcher.hashCode());
-            return result;
+        public boolean hasMaskedFields() {
+            return false;
         }
 
-        @Override
```
scrawfor99, line 832: We should either keep this here (with the updated logic) or make sure it is properly inherited since this is a complex object.

src/main/java/org/opensearch/security/securityconf/SecurityRole.java
```diff
@@ -0,0 +1,41 @@
...
+import org.opensearch.security.resolver.IndexResolverReplacer;
+import org.opensearch.security.user.User;
+
+import java.util.Set;
+
+public interface SecurityRole {
+    boolean impliesClusterPermission(String action);
+
+    // get indices which are permitted for the given types and actions
+    // dnfof + opensearchDashboards special only
```
scrawfor99, line 25: Nit: Write out dnfof meaning

scrawfor99 (commented):
Nice job with all the tests @samuelcostae. Just a few comments but otherwise great.

src/main/java/org/opensearch/security/securityconf/SecurityRoles.java
```diff
@@ -85,4 +85,6 @@ Set<String> getAllPermittedIndicesForDashboards(
 
     SecurityRoles filter(Set<String> roles);
 
+    Set<SecurityRole> getRoles();
```
peternied, line 88: [Blocking] This breaks the encapsulation of this SecurityRoles, instead there should be a way to check for explicit permissions, something like, `boolean hasExplicitIndexPermissions(Resolved resolved, User user, String action, ...)`.

This would greatly reduce the amount of code in the SecurityIndexAccessEvaluator, and remove the other casting requirements in these ConfigModel classes.

peternied: I have created a PR [1 ]that can address this, still needs cleanup but it can get you started
- https://github.com/samuelcostae/security/pull/1

It passes the following tests:
- SystemIndexDisabledTests
- SystemIndexPermissionDisabledTests	
- SystemIndexPermissionEnabledTests

peternied: FYI - @DarshitChanpura @samuelcostae

src/test/java/org/opensearch/security/system_indices/AbstractSystemIndicesTests.java
```diff
@@ -0,0 +1,193 @@
...
+import org.opensearch.common.xcontent.XContentType;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.core.rest.RestStatus;
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.security.test.DynamicSecurityConfig;
+import org.opensearch.security.test.SingleClusterTest;
+import org.opensearch.security.test.helper.file.FileHelper;
+import org.opensearch.security.test.helper.rest.RestHelper;
+import static org.junit.Assert.assertEquals;
```
RyanL1997, line 39: nit: should we just import `import org.junit.Assert;` and use `Assert.xxx()`? I saw that the other test cases we did such as `PrivilegesEvaluatorTest.java` is following this.

src/test/java/org/opensearch/security/system_indices/SystemIndexPermissionDisabledTests.java
```diff
@@ -0,0 +1,397 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.opensearch.action.admin.indices.close.CloseIndexRequest;
+import org.opensearch.client.Client;
+import org.opensearch.core.rest.RestStatus;
+import org.opensearch.security.test.helper.rest.RestHelper;
+
+import java.io.IOException;
+
+import static org.junit.Assert.assertEquals;
```
RyanL1997, line 25: same as the above

src/main/java/org/opensearch/security/securityconf/ConfigModelV6.java
```diff
@@ -448,6 +448,17 @@ public EvaluatedDlsFlsConfig getDlsFls(
...
         }
 
+        public boolean hasExplicitIndexPermission(
+            Resolved resolved,
+            User user,
+            String[] actions,
+            IndexNameExpressionResolver resolver,
+            ClusterService cs
+        ) {
+            // TODO: Handle this scenario in V6 config
```
RyanL1997, line 458: Do we need this in v6 btw?

src/test/java/org/opensearch/security/IntegrationTests.java
```diff
@@ -1045,9 +1046,8 @@ public void testSecurityIndexSecurity() throws Exception {
             encodeBasicHeader("nagilum", "nagilum")
         );
         Assert.assertEquals(HttpStatus.SC_FORBIDDEN, res.getStatusCode());
-        // res = rh.executePostRequest(".opendistro_security/_freeze", "",
-        // encodeBasicHeader("nagilum", "nagilum"));
-        // Assert.assertTrue(res.getStatusCode() >= 400);
+        res = rh.executePostRequest(".opendistro_security/_freeze", "", encodeBasicHeader("nagilum", "nagilum"));
+        Assert.assertTrue(res.getStatusCode() >= 400);
```
RyanL1997, line 1050: nit: Is there a specific status code we can get? such as 403?

RyanL1997 (commented):
Hi @samuelcostae and @DarshitChanpura. Thanks for the work of this. In generally this is looking good and I just left some minor comments.

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
peternied: Thes have been combined into a single call, resolving

src/main/java/org/opensearch/security/securityconf/ConfigModelV6.java
peternied: This change has been reverted

src/main/java/org/opensearch/security/securityconf/IndexPattern.java
peternied: This change has been reverted

src/main/java/org/opensearch/security/securityconf/SecurityRole.java
peternied: This change has been reverted

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -71,122 +82,240 @@ public SecurityIndexAccessEvaluator(final Settings settings, AuditLog auditLog,
         this.systemIndexMatcher = WildcardMatcher.from(
             settings.getAsList(ConfigConstants.SECURITY_SYSTEM_INDICES_KEY, ConfigConstants.SECURITY_SYSTEM_INDICES_DEFAULT)
         );
-        this.systemIndexEnabled = settings.getAsBoolean(
+        this.protectedSystemIndexMatcher = WildcardMatcher.from(ConfigConstants.OPENDISTRO_SECURITY_DEFAULT_CONFIG_INDEX);
```
peternied, line 85: This is incorrect, if the security index is renamed by the configuration it will not be used, make the matcher from `this.securityIndex`, or just create the matcher inline as securityIndex is used throughout this file

src/test/java/org/opensearch/security/system_indices/AbstractSystemIndicesTests.java
DarshitChanpura: either is fine.

src/test/java/org/opensearch/security/system_indices/SystemIndexDisabledTests.java
```diff
@@ -0,0 +1,398 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.opensearch.action.admin.indices.close.CloseIndexRequest;
+import org.opensearch.client.Client;
+import org.opensearch.core.rest.RestStatus;
+import org.opensearch.security.test.helper.rest.RestHelper;
+
+import java.io.IOException;
+
+import static org.junit.Assert.assertEquals;
```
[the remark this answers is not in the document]
DarshitChanpura: no need. static imports are a cleaner approach.

src/test/java/org/opensearch/security/IntegrationTests.java
DarshitChanpura: not sure why this was added as >=400 originally,  I will correct this to the expected return code which is 400

DarshitChanpura (approved):

src/main/java/org/opensearch/security/securityconf/ConfigModelV6.java
DarshitChanpura: done. Please take a look.

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -71,122 +82,242 @@ public SecurityIndexAccessEvaluator(final Settings settings, AuditLog auditLog,
...
-
         final boolean restoreSecurityIndexEnabled = settings.getAsBoolean(
             ConfigConstants.SECURITY_UNSUPPORTED_RESTORE_SECURITYINDEX_ENABLED,
             false
         );
 
-        final List<String> securityIndexDeniedActionPatternsList = new ArrayList<String>();
+        final List<String> securityIndexDeniedActionPatternsList = deniedActionPatterns();
+
+        final List<String> securityIndexDeniedActionPatternsListNoSnapshot = new ArrayList<>(securityIndexDeniedActionPatternsList);
```
cwperks, line 97: Nit: This is already in a class called SecurityIndexAccessEvaluator so wdyt about shortening these variable names?

```
deniedActionPatternsList
deniedActionPatternsListNoSnapshot
```

DarshitChanpura:
@cwperks @peternied @RyanL1997 All outstanding comments have been addressed. Please review it once you get a chance.

RyanL1997 (approved):

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -71,122 +82,242 @@ public SecurityIndexAccessEvaluator(final Settings settings, AuditLog auditLog,
         this.systemIndexMatcher = WildcardMatcher.from(
             settings.getAsList(ConfigConstants.SECURITY_SYSTEM_INDICES_KEY, ConfigConstants.SECURITY_SYSTEM_INDICES_DEFAULT)
         );
-        this.systemIndexEnabled = settings.getAsBoolean(
+        this.protectedSystemIndexMatcher = WildcardMatcher.from(this.securityIndex);
```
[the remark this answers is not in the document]
peternied: Want to make a follow up PR to change - I tend to agree with these names could be improved.

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
peternied: Want to make a follow up PR to change - I agree.

peternied (approved):
Thanks for the work on this @DarshitChanpura and @samuelcostae

Issue #2263, opened 2022-11-16
[BUG] java.util.ConcurrentModificationException with plugins.security.cache.ttl_minutes=0


olddanmer:
**What is the bug?**
With option `plugins.security.cache.ttl_minutes: 0` opensearch sometimes raises java.util.ConcurrentModificationException error and logstash get 500 error at that moment. 

I use dashboards with OpenID integration, so option `plugins.security.cache.ttl_minutes: 0` is recommended by documentation. Without `plugins.security.cache.ttl_minutes: 0` dashboards doesn't work properly with OpenID (see bug https://github.com/opensearch-project/security-dashboards-plugin/issues/1138#issuecomment-1279635558)

**How can one reproduce the bug?**
1. Setup opensearch 2.3.0, logstash-oss 7.12.1
2. Set opensearch setting `plugins.security.cache.ttl_minutes=0` 
3. logstash uses bulk write (?)
4. See errors in logs

opensearch logs
```
Nov 15 15:15:07 srvroom-es-logs-1 docker[1142]: [2022-11-15T15:15:07,061][ERROR][o.o.s.f.SecurityFilter   ] [es-logs-1.srvroom.labs.intellij.net] Unexpected exception java.util.ConcurrentModificationException
Nov 15 15:15:07 srvroom-es-logs-1 docker[1142]: java.util.ConcurrentModificationException: null
Nov 15 15:15:07 srvroom-es-logs-1 docker[1142]: at java.util.HashMap$HashIterator.nextNode(HashMap.java:1597) ~[?:?]
Nov 15 15:15:07 srvroom-es-logs-1 docker[1142]: at java.util.HashMap$KeyIterator.next(HashMap.java:1620) ~[?:?]
Nov 15 15:15:07 srvroom-es-logs-1 docker[1142]: at java.util.Collections$UnmodifiableCollection$1.next(Collections.java:1054) ~[?:?]
Nov 15 15:15:07 srvroom-es-logs-1 docker[1142]: at java.util.AbstractCollection.addAll(AbstractCollection.java:335) ~[?:?]
Nov 15 15:15:07 srvroom-es-logs-1 docker[1142]: at java.util.HashSet.<init>(HashSet.java:121) ~[?:?]
[...]

PR #2912, opened 2023-06-28
ConcurrentModificationException issue fix

RyanL1997:
### Description
Fix the `java.util.ConcurrentModificationException` with `plugins.security.cache.ttl_minutes=0` configured.

### Issues Resolved
* Resolve https://github.com/opensearch-project/security/issues/2263

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
@@ -207,7 +207,7 @@ public class PrivilegesEvaluator {
             StringJoiner joiner = new StringJoiner("|");
             joiner.add(user.getName());
             joiner.add(String.join(",", user.getRoles()));
-            joiner.add(String.join(",", Sets.union(user.getSecurityRoles().stream().collect(ImmutableSet.toImmutableSet()), mappedRoles)));
+            joiner.add(String.join(",", Sets.union(user.getSecurityRoles(), mappedRoles)));
             String requestedTenant = user.getRequestedTenant();
             if (!Strings.isNullOrEmpty(requestedTenant)) {
                 joiner.add(requestedTenant);
--- a/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
@@ -262,7 +262,7 @@ public class PrivilegesEvaluator {
 
         setUserInfoInThreadContext(user, mappedRoles);
         // Add the security roles for this user so that they can be used for DLS parameter substitution.
-        user.addSecurityRoles(mappedRoles);
+        user.setSecurityRoles(mappedRoles);
 
         final boolean isDebugEnabled = log.isDebugEnabled();
         if (isDebugEnabled) {
--- a/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
@@ -207,7 +207,7 @@ public class PrivilegesEvaluator {
             StringJoiner joiner = new StringJoiner("|");
             joiner.add(user.getName());
             joiner.add(String.join(",", user.getRoles()));
-            joiner.add(String.join(",", Sets.union(user.getSecurityRoles(), mappedRoles)));
+            joiner.add(String.join(",", Sets.union(user.getSecurityRoles().stream().collect(ImmutableSet.toImmutableSet()), mappedRoles)));
             String requestedTenant = user.getRequestedTenant();
             if (!Strings.isNullOrEmpty(requestedTenant)) {
                 joiner.add(requestedTenant);
--- a/src/main/java/org/opensearch/security/user/User.java
+++ b/src/main/java/org/opensearch/security/user/User.java
@@ -272,13 +272,13 @@ public class User implements Serializable, Writeable, CustomAttributesAware {
         return attributes;
     }
 
-    public final void addSecurityRoles(final Collection<String> securityRoles) {
+    public synchronized final void addSecurityRoles(final Collection<String> securityRoles) {
         if (securityRoles != null && this.securityRoles != null) {
             this.securityRoles.addAll(securityRoles);
         }
     }
 
-    public final Set<String> getSecurityRoles() {
+    public synchronized final Set<String> getSecurityRoles() {
         return this.securityRoles == null
             ? Collections.synchronizedSet(Collections.emptySet())
             : Collections.unmodifiableSet(this.securityRoles);
--- a/src/main/java/org/opensearch/security/user/User.java
+++ b/src/main/java/org/opensearch/security/user/User.java
@@ -33,9 +33,14 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 
 import org.opensearch.common.io.stream.StreamInput;
@@ -73,7 +78,7 @@ public class User implements Serializable, Writeable, CustomAttributesAware {
      * roles == backend_roles
      */
     private final Set<String> roles = Collections.synchronizedSet(new HashSet<String>());
-    private final Set<String> securityRoles = Collections.synchronizedSet(new HashSet<String>());
+    private final AtomicReference<Set<String>> securityRoles = new AtomicReference<>(new HashSet<String>());
     private String requestedTenant;
     private Map<String, String> attributes = Collections.synchronizedMap(new HashMap<>());
     private boolean isInjected = false;
@@ -84,7 +89,7 @@ public class User implements Serializable, Writeable, CustomAttributesAware {
         roles.addAll(in.readList(StreamInput::readString));
         requestedTenant = in.readString();
         attributes = Collections.synchronizedMap(in.readMap(StreamInput::readString, StreamInput::readString));
-        securityRoles.addAll(in.readList(StreamInput::readString));
+        securityRoles.get().addAll(in.readList(StreamInput::readString));
     }
 
     /**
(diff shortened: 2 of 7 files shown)


cwperks (dismissed):
Thank you for figuring this out @RyanL1997! 

`addSecurityRoles` has always sounded like a weird method to me in the User class - I would think it should just have a setter.

If a BulkRequest comes into OpenSearch the mapped roles should be computed only once and then available on all of the threads. Can we ask opensearch-benchmarking how to measure the performance impact of a change? For this PR specifically how it impacts the performance of BulkRequests.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -207,7 +207,7 @@ private void setUserInfoInThreadContext(User user, Set<String> mappedRoles) {
             StringJoiner joiner = new StringJoiner("|");
             joiner.add(user.getName());
             joiner.add(String.join(",", user.getRoles()));
-            joiner.add(String.join(",", Sets.union(user.getSecurityRoles(), mappedRoles)));
+            joiner.add(String.join(",", Sets.union(user.getSecurityRoles().stream().collect(ImmutableSet.toImmutableSet()), mappedRoles)));
```
peternied, line 210: What does this change do?

src/main/java/org/opensearch/security/user/User.java
```diff
@@ -272,13 +272,13 @@ public synchronized final Map<String, String> getCustomAttributesMap() {
         return attributes;
     }
 
-    public final void addSecurityRoles(final Collection<String> securityRoles) {
+    public synchronized final void addSecurityRoles(final Collection<String> securityRoles) {
```
peternied, line 275: ConcurrentModificationException is thrown when iterating through a collection its contents are changed - so the iterator is invalid [1].  In `PrivilegesEvaluator` we are still iterating over the collection - outside of a synchronized block were the iterator could be modified.

To prevent this exception we would need to remove the `addSecurityRoles(...)` function, because this it updates the underlying collection - @cwperks calls this out as well.  

Alternatively - less clean, change addSecurityRoles(...) so it changes the collection after its been updated, ala change the field `securityRoles` to be an ActomicReference<Set<String>>, then do `securityRole.set(Sets.union(securityRoles.get(), securityRoles)`, if any references to the original securityRoles exist in another thread/memory they are different references than the new object that was created


- [1] https://stackoverflow.com/questions/602636/why-is-a-concurrentmodificationexception-thrown-and-how-to-debug-it

peternied (changes requested):
I don't think this change fixes the issue, it should also come with some kind of test to ensure it works as expected

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
cwperks: @peternied Read the conversation from here and below: https://github.com/opensearch-project/security/issues/2263#issuecomment-1603344501

This gets a copy of the set so that its not susceptible to concurrent modification. The better approach is to set the roles once and remove `addSecurityRoles` if its possible.

src/main/java/org/opensearch/security/user/User.java
RyanL1997: @peternied and @cwperks, Yes, I agree. Working on removing this `addSecurityRoles()`

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
peternied: > This gets a copy of the set so that its not susceptible to concurrent modification.

That is not correct, `user.getSecurityRoles().stream().collect(...)` still iterates over each item to make the copy.

peternied: In to strict sense, while this code is running, the iterator is susceptible to concurrent modification exception.

If we look at the Sets.union(...) [[source code](https://github.com/google/guava/blob/f2f4f864d720ad03a045c9b918fe06cb11a65832/guava/src/com/google/common/collect/Sets.java#L713)], it is creating a view over both collections, so this change would use the resulting in memory 'ImmutableSet' and lower the likelihood for this exception.

Sets.union(...) earns a 👎 since on interaction its doing duplicate filtering ever time and doesn't make it obvious it is creating a side effect.

src/main/java/org/opensearch/security/user/User.java
scrawfor99, line 1: Hi @RyanL1997, I am just following up here to see if you need any assistance. I see that you mentioned removing the `addSecurityRoles()` below. If this proves a challenge, you can make use of an AtomicReference as Peter suggested. This will allow you to preserve the logic and prevent any attempts to set the value more than a single time.

RyanL1997:
Some updates:

I have tried the change of `addSecurityRoles( )` so it changes the collection after its been updated with the AtomicReference approach. However, it caused some deserialization problem:
```
...
'caused_by': {'type': 'invalid_class_exception', 'reason': 'Unauthorized deserialization attempt ; java.util.concurrent.atomic.AtomicReference'}}, 'data': 
...
```

So I think I have to go with another approach of removing `addSecurityRoles` and trying to use a setter.

scrawfor99:
Hi @peternied, @cwperks, @RyanL1997, just following up to see if we can work something out to be able to resolve this. I know there was some discussion above but not sure where that ended. Let me know if I can assist.

scrawfor99:
@RyanL1997, just checking in to see if you needed anything with regard to this PR.

RyanL1997:
@scrawfor99 Thanks for the check up. This currently Atomic Reference works now. However, as we discussed on Monday, I think we will remove the `addSecurityRoles()` with a `setSecurityRoles()` and all the role related set up will only happen in REST layer. I will include this change in my next commit soon.

src/main/java/org/opensearch/security/user/User.java
```diff
@@ -272,15 +272,16 @@ public synchronized final Map<String, String> getCustomAttributesMap() {
...
+                this.securityRoles.addAll(securityRoles);
+            }
         }
     }
 
     public final Set<String> getSecurityRoles() {
-        return this.securityRoles == null
-            ? Collections.synchronizedSet(Collections.emptySet())
-            : Collections.unmodifiableSet(this.securityRoles);
+        return this.securityRoles == null ? Collections.synchronizedSet(Collections.emptySet()) : ImmutableSet.copyOf(this.securityRoles);
```
RyanL1997, line 287: I think it is okay we keep this change from the last commit.

src/main/java/org/opensearch/security/user/User.java
```diff
@@ -74,6 +75,7 @@ public class User implements Serializable, Writeable, CustomAttributesAware {
      */
     private final Set<String> roles = Collections.synchronizedSet(new HashSet<String>());
     private final Set<String> securityRoles = Collections.synchronizedSet(new HashSet<String>());
```
RyanL1997, line 77: Switching back from the Atomic Reference

RyanL1997 (commented):
Hi @cwperks , this is the latest commit for changing the approach to `setSecurityRoles()`. I just wanna set up this thread for some of the questions I have.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -262,7 +262,7 @@ public PrivilegesEvaluatorResponse evaluate(
 
         setUserInfoInThreadContext(user, mappedRoles);
         // Add the security roles for this user so that they can be used for DLS parameter substitution.
-        user.addSecurityRoles(mappedRoles);
+        user.setSecurityRoles(mappedRoles);
```
cwperks, line 265: Is it possible to verify that `setSecurityRoles` is only called once per REST Request? IMO it should throw an error if its trying to set a value that's already been set.

src/main/java/org/opensearch/security/user/User.java
cwperks: This should not be returning a new copy every time its called. If `securityRoles` is only set once, then there is no risk of concurrent modification. `Collections.unmodifiableSet(...)` would be better used here since it does not return a copy.

src/main/java/org/opensearch/security/user/User.java
```diff
@@ -272,15 +274,16 @@ public synchronized final Map<String, String> getCustomAttributesMap() {
         return attributes;
     }
 
-    public final void addSecurityRoles(final Collection<String> securityRoles) {
-        if (securityRoles != null && this.securityRoles != null) {
-            this.securityRoles.addAll(securityRoles);
+    public final void setSecurityRoles(final Collection<String> securityRoles) {
```
peternied, line 277: This perpetuates a confusing design, its still a side-effect.  Remove this field entirely from User, and create a subclass UserWithRoles that extends User.  UserWithRoles could require the mapped roles as part of its construction ensuring they aren't modified after the user has been created.

Then in places where `user.getSecurityRoles()` is called we can work backwards on what needs to be fixed so the UserWithRoles object is chained correctly ensuring a single assignment and no possiblity of dynamic update.

IF we see that dynamic updates are taking place we can address them individually

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
scrawfor99: You can make an atomic reference to the roles and then use SetOnce.

scrawfor99 (commented):
Hi @RyanL1997 just checking in on this to see if you need help driving things to a close.

jkrishnarao2003:
I am seeing this issue when running monitors.  Is there a fix for this in future versions or is it something on my side to fix?
OpenSearch version **2.10.0**
Exception trace as follows.

`[2024-05-27T22:35:10,135][ERROR][o.o.a.DocumentLevelMonitorRunner] [omhqp120d5.uprr.com] Failed running Document-level-monitor Monitor1
java.util.ConcurrentModificationException: null
        at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:756) ~[?:?]
        at java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:778) ~[?:?]
        at org.opensearch.alerting.DocumentLevelMonitorRunner.runMonitor(DocumentLevelMonitorRunner.kt:141) [opensearch-alerting-2.10.0.0.jar:2.10.0.0]
        at org.opensearch.alerting.DocumentLevelMonitorRunner$runMonitor$1.invokeSuspend(DocumentLevelMonitorRunner.kt) [opensearch-alerting-2.10.0.0.jar:2.10.0.0]
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33) [kotlin-stdlib-1.8.21.jar:1.8.21-release-380(1.8.21)]
        at kotlinx.coroutines.DispatchedTask.run(Dispatched.kt:233) [kotlinx-coroutines-core-1.1.1.jar:?]
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:594) [kotlinx-coroutines-core-1.1.1.jar:?]
        at kotlinx.coroutines.scheduling.CoroutineScheduler.access$runSafely(CoroutineScheduler.kt:60) [kotlinx-coroutines-core-1.1.1.jar:?]
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:742) [kotlinx-coroutines-core-1.1.1.jar:?]
`

jkrishnarao2003:
Will do.  Thank you.

PR #2936, opened 2023-07-05
[Backport] Parallel test jobs for CI (#2861)

pawel-gudel-eliatra:
Backport https://github.com/opensearch-project/security/commit/e4f4817e0d2d9caf80da0d0cb1da73e4913f62e3 from https://github.com/opensearch-project/security/pull/2861


.github/workflows/ci.yml
```diff
@@ -6,13 +6,34 @@ env:
   GRADLE_OPTS: -Dhttp.keepAlive=false
 
 jobs:
-  build:
-    name: build
+  generate-test-list:
+    runs-on: ubuntu-latest
```
scrawfor99, line 10: If this is only running on Linux, I think this means the tests below which need "generate-test-list" will not run on Windows?

pawel-gudel-eliatra: We are only generating list of tasks which is OS-agnostic. It is already used in generating (and running) tasks on windows machines. Please take a look at #2861 or pretty much every current "CI" workflow: [example](https://github.com/opensearch-project/security/actions/runs/5467098686)

scrawfor99: Hmm, it would appear the internet lied to me haha. Sounds good then!

scrawfor99 (dismissed):

peternied (approved):

cwperks (approved):

peternied:
Interesting, codecov/project is out of date compared to the results generated from all the integration test runs, that is kinda silly

image

PR #2787, opened 2023-05-23
Implements token handlers and tests 

scrawfor99:
### Description
[Describe what this change achieves]

First introduces the feature branch into the main branch. 

This PR introduces both internalUser and Security User (JWT) token handlers inside the security plugin. These handlers are managed by an overlying SecurityTokenManager class which sorts the token interaction requests coming from core. 

These three managers work in concert to provide the required functionality of the Identity Plugin in core for Extensions to be live. 

NOTE: This requires https://github.com/opensearch-project/OpenSearch/pull/7452 to pass. 

### Testing
Adds three new test files with mocked tests for all methods. 

### Check List
- [x] New functionality includes testing
- [x] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2938, opened 2023-07-05
#2616  bwcPlugin extension setting and getTokenManager implementation

### LINKED ISSUE #2616: [Extensions] Add `opensearch.yml` setting to enable backward compatible plugin mode for extensions
A new setting in `opensearch.yml` needs to be added to enable backward compatible plugin mode for extensions. The setting can be:

```
plugins.security.experimental.extensions.bwc_plugin_mode: true
```

**Background**

Many plugins currently rely on user information to be present on the ThreadContext to support plugin use-cases. Examples include:

1. **resource protection** - i.e. sharing detectors in Anomaly Detection via backend roles)
2. **asynchronous operation**  -i.e. running a detector on a schedule and injecting the user's roles from the time of detector creation to evaluate if a user has privileges on an index
 
See: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L198-L210

```
private void setUserInfoInThreadContext(User user, Set<String> mappedRoles) {
    if (threadContext.getTransient(OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT) == null) {
        joiner.add(user.getName());
            joiner.add(requestedTenant);
        }
    }
}
```

In order for extensions to initially be a replacement for plugins and support all current use-cases, they will need the data from the above snippet. (**Note**: specifically the information needed is the `username`, `backend_roles` and `mapped_roles` - mapped roles here refers to the ultimate roles that have been resolved to after all roles mapping - 1. roles explicitly assigned + 2. roles mapped to via backend_roles 3. roles matching the hosts in roles_mapping and 4. And_backend_roles - the user needs all backend roles in the list to be mapped. Full Details: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java#L1206-L1269)

This setting will ultimately determine if the `encryption_key` from https://github.com/opensearch-project/security/issues/2615 is used to encrypt the `roles` and `backend_roles` claims in the payload of the token sent to an extension.

samuelcostae:
### Description
This Draft PR includes the new setting bwcPluginMode (backward compatible plugin mode for extensions )  and an tentaptive/draft  implementation of  OpenSearchSecurityPlugin.java.getTokenManager().

getTokenManager() is a overriden method from the IdentityPlugin Interface and  is called by core in RestSendToExtensionAction.java. 

The issued token will be sent to the extension in the payload and should include the user and its mapped roles if the setting is set to true. 

### Issues Resolved
#2616 

### Testing
-

### Check List


scrawfor99:
Hi Sam, thanks for taking this on. I think we may want to split the extension setting away from the tokenManager. If you look here: https://github.com/opensearch-project/security/pull/2787, you can see some of what I had been working on with the TokenManager but for the time being, we need to wait until the feature branch is at a state appropriate to be merged in main.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1098,6 +1108,21 @@ public Settings additionalSettings() {
         return builder.build();
     }
 
+    @Override
+    public List<Setting<?>> getExtensionSettings() {
```
cwperks, line 1112: This looks fine to me, but I'd also like to see the usage of the setting inside the JWTVendor in `feature/extensions` branch to show how this setting impacts the payload of the JWT created of the obo token for forwarding to an extension.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1876,6 +1901,16 @@ private static String handleKeyword(final String field) {
         return field;
     }
 
+    @Override
+    public Subject getSubject() {
+        return null;
```
cwperks, line 1906: This will also need to be addressed as we implement IdentityPlugin for the security plugin. Subject should never be null and if no subject is present in the current context then its assumed to be unauthenticated.

You can see an example implementation of SecuritySubject in this draft PR: https://github.com/opensearch-project/security/pull/2773/files#diff-f21aa0165f9d0633658e7c99a1375b84258d1ef2beecaad8df286cfe5ba2a7d2R1-R49

cwperks (commented):
Hey Sam, I took an initial pass. I see that you have a usage of the setting in the token manager introduced here, but I think we need to combine this PR with the `feature/extensions` branch which introduces a JWTVendor which was created for issuing JWTs (on behalf of tokens) for an extension. You are on the right track, but `issueToken` should vend a JWT using the vendor and this setting will help determine the payload of that JWT and whether backend roles are included in the payload or not. The purpose of including backend roles in the payload is for plugins that are being rewritten as extensions and currently rely on the knowledge of a user's backend roles.

src/main/java/org/opensearch/security/auth/SecurityTokenManager.java
```diff
@@ -0,0 +1,69 @@
...
+import org.opensearch.security.http.XFFResolver;
+import org.opensearch.security.securityconf.ConfigModel;
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.security.user.User;
+import org.opensearch.threadpool.ThreadPool;
+
+import java.util.Set;
+import java.util.StringJoiner;
+
+public class SecurityTokenManager implements TokenManager {
```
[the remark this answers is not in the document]
samuelcostae: @cwperks  Either way is fine by me, I agree that is beyond the scope originally stated in the issue  description, but I thought you had previously asked me(In the other PR)  to include some usage of the setting. Were you referencing some other part of the implementation?

samuelcostae:
Closing as will create a PR from a  different branch based on feature/extensions

PR #2975, opened 2023-07-10
[Backport 2.8] Resolve CVE-2023-2976 by forcing use of Guava 32.0.1 (#2937)

DarshitChanpura:
Backports #2937 

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -285,6 +285,8 @@ configurations.all {
         force "io.netty:netty-transport:${versions.netty}"
         force "io.netty:netty-transport-native-unix-common:${versions.netty}"
         force "org.apache.bcel:bcel:6.6.0" // This line should be removed once Spotbugs is upgraded to 4.7.4
+        force "com.github.luben:zstd-jni:${versions.zstd}"
+        force "org.xerial.snappy:snappy-java:1.1.10.1"
         force 'com.google.guava:guava:32.0.1-jre'
     }
 }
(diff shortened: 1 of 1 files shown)


scrawfor99 (approved):

willyborankin (dismissed):

build.gradle
```diff
@@ -285,17 +285,20 @@ configurations.all {
         force "io.netty:netty-transport:${versions.netty}"
         force "io.netty:netty-transport-native-unix-common:${versions.netty}"
         force "org.apache.bcel:bcel:6.6.0" // This line should be removed once Spotbugs is upgraded to 4.7.4
+        force "com.github.luben:zstd-jni:${versions.zstd}"
```
peternied, line 288: > Could not resolve com.github.luben:zstd-jni:null.

willyborankin: Hmmm I think for 2.x it needs to be 1.75. Since `versions.zstd` comes with 2.9 version. So after 2.9 has been released we can use it in 2.x.

willyborankin: Ahh sorry for the inconvenience.  I did not notice that it is for 2.8 version. The version of ZSTD should stay the same in this case. 
ZSTD 1.5.5-3 was merged in OS main branch and before that sec plugin uses its own. To avoid regressions lets leave it as is.

davidlago:
No longer needed as 2.9 is about to be released.

Issue #2891, opened 2023-06-22
[FEATURE] Handle the edge cases of On-behalf-of Authentication


RyanL1997:
**Is your feature request related to a problem?**
Since the OBO Token can perform on behalf of another user, it is necessary for us to frame the usages of this token.

* Relate https://github.com/opensearch-project/security/issues/2573

**What solution would you like?**
- OBO Token cannot be used to issue another OBO token
- OBO Token cannot be used to change the passwords

DarshitChanpura:
[Triage] Thanks for filing the issue.

peternied:
This issue was closed without any finding or description of the expectations, reopening.

RyanL1997:
Transfer the conversation here ([comment1](https://github.com/opensearch-project/security/pull/3179#discussion_r1303713840), [comment2](https://github.com/opensearch-project/security/pull/3179#discussion_r1303731097)).

Hi @peternied, I can re-edit the issue if we choose to refactor the logic into a public field. But still I think as we mentioned in the standup before, the main goal of this original issue is having the functionality of block these 2 endpoint, which has been implemented.

PR #3008, opened 2023-07-13
[Feature/Extension] Restrict OBO token's usage for certain endpoints

RyanL1997:
### Description
Edge Cases Of OBO Token Usage: 
- [x] User should not use OBO Token to issue another OBO Token
- [x] User should not use OBO Token to modify users' authinfo, such as resetting password

New feature

### Issues Resolved
* Resolve https://github.com/opensearch-project/security/issues/2891

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
+++ b/src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
@@ -105,12 +105,15 @@ public class JwtVendor {
         List<String> roles,
         List<String> backendRoles
     ) throws Exception {
+        String tokenIdentifier = "OBO";
         long timeMillis = timeProvider.getAsLong();
         Instant now = Instant.ofEpochMilli(timeProvider.getAsLong());
 
         jwtProducer.setSignatureProvider(JwsUtils.getSignatureProvider(signingKey));
         JwtClaims jwtClaims = new JwtClaims();
         JwtToken jwt = new JwtToken(jwtClaims);
+
+        jwtClaims.setProperty("token_identifier", tokenIdentifier);
 
         jwtClaims.setIssuer(issuer);
 
--- a/src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
+++ b/src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
@@ -105,7 +105,7 @@ public class JwtVendor {
         List<String> roles,
         List<String> backendRoles
     ) throws Exception {
-        String tokenIdentifier = "OBO";
+        String tokenIdentifier = "obo";
         long timeMillis = timeProvider.getAsLong();
         Instant now = Instant.ofEpochMilli(timeProvider.getAsLong());
 
@@ -113,7 +113,7 @@ public class JwtVendor {
         JwtClaims jwtClaims = new JwtClaims();
         JwtToken jwt = new JwtToken(jwtClaims);
 
-        jwtClaims.setProperty("token_identifier", tokenIdentifier);
+        jwtClaims.setProperty("typ", tokenIdentifier);
 
         jwtClaims.setIssuer(issuer);
 
--- a/src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
+++ b/src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
@@ -89,6 +89,8 @@ public class SecurityRestFilter {
     private static final String HEALTH_SUFFIX = "health";
     private static final String WHO_AM_I_SUFFIX = "whoami";
 
+    private static final String ON_BEHALF_OF_SUFFIX = "onbehalfof";
+
     private static final String REGEX_PATH_PREFIX = "/(" + LEGACY_OPENDISTRO_PREFIX + "|" + PLUGINS_PREFIX + ")/" + "(.*)";
     private static final Pattern PATTERN_PATH_PREFIX = Pattern.compile(REGEX_PATH_PREFIX);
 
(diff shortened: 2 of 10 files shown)


RyanL1997:
This branch should sync with `main` after the merge of https://github.com/opensearch-project/security/pull/3003

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -259,6 +259,17 @@ private boolean checkAndAuthenticateRequest(RestRequest request, RestChannel cha
             }
         }
 
+        if (HTTPHelper.containsOBOToken(request)) {
+            String OBO_ENDPOINT_PREFIX = "_plugins/_security/api/user/onbehalfof";
+            if (request.method() == Method.POST && OBO_ENDPOINT_PREFIX.equals(suffix)) {
```
cwperks, line 264: Similar to how the suffixes are configured in [SecurityRestFilter](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/filter/SecurityRestFilter.java#L89-L90) to skip authc for endpoints ending in the suffix, this could also just be the last portion of the endpoint `onbehalfof`

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -105,13 +105,16 @@ public String createJwt(
...
     ) throws Exception {
+        String tokenIdentifier = "OBO";
         long timeMillis = timeProvider.getAsLong();
         Instant now = Instant.ofEpochMilli(timeProvider.getAsLong());
 
         jwtProducer.setSignatureProvider(JwsUtils.getSignatureProvider(signingKey));
         JwtClaims jwtClaims = new JwtClaims();
         JwtToken jwt = new JwtToken(jwtClaims);
 
+        jwtClaims.setProperty("token_identifier", tokenIdentifier);
```
cwperks, line 116: Maybe `typ` here? Short for type. https://datatracker.ietf.org/doc/html/rfc7519#section-5.1

From my understanding `typ` is typically used to help the receiver of this token decide what logic to use while handling.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -105,13 +105,16 @@ public String createJwt(
         List<String> roles,
         List<String> backendRoles
     ) throws Exception {
+        String tokenIdentifier = "OBO";
```
cwperks, line 108: IMO OBO should be lower cased here.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -259,6 +259,17 @@ private boolean checkAndAuthenticateRequest(RestRequest request, RestChannel cha
             }
         }
 
+        if (HTTPHelper.containsOBOToken(request)) {
```
cwperks, line 262: This doesn't feel like the correct place to put this logic. Would it make sense to put this in the `OnBehalfOfAuthenticator.extractCredentials0` and fail to authenticate within the authenticator if its using the token for a forbidden endpoint?

Also wdyt about updating the OBOAuthenticator to check for the type claim this PR introduces as well?

cwperks (commented):
Thank you Ryan! I took a first pass and left a few comments around structure.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
RyanL1997: You mean the `SetProperty("typ", xxx)`? I didnt see there is function for `setType()`, but I do see there is something like `setTokenId`. I'm not sure if the id should fit into our case?

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
RyanL1997: Nice catch. I just fixed.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
RyanL1997: Thanks for the reference btw. Yes I have change the key into `typ` with the `setProperty()` still.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
RyanL1997: Done~

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
RyanL1997: > Would it make sense to put this in the OnBehalfOfAuthenticator.extractCredentials0 and fail to authenticate within the authenticator if its using the token for a forbidden endpoint?

I think I get the idea.I see I think I got the idea. Is that fine to leave `containsOBOToken()` check in the `HTTPHelper`?

> Also wdyt about updating the OBOAuthenticator to check for the type claim this PR introduces as well?

Sure thing! I just added the check and test cases.

cwperks: If a request can be authenticated by the OnBehalfOfAuthenticator than there's no need for the `containsOBOToken()` check because the authenticator extracts the Authorization header and verifies the token in the HTTPHeader.

The reason its best to put this check after authentication is that we can log the attempts to change password or request a new token as suspicious behavior.

peternied (changes requested):
I'm struggling to think what the correct criteria is for these restrictions, while the issue has called out the OBO tokens creation API, I think we should articulate the criteria for something to be acceptable vs not acceptable first.  Lets try to build consensus around this idea first before implementing restrictions so we can be consistent.

What if instead of denying permissions to different APIs system, what if instead only let you create OBO tokens that expire at latest when the currently OBO token expires.  This prevents the 'no expiration' in a more 'logical' way than the permissions denial.

Issue #2764, opened 2023-05-11
Determine how auth tokens are forwarded to Extensions


davidlago:
With the security plugin, the auth token to be forwarded to an extension will be generated after successful authentication on a request destined for an extension. The actual request forwarding is performed in the extensions rest handler [RestSendToExtensionAction](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/extensions/rest/RestSendToExtensionAction.java). This handler needs to be able to obtain the generated token to forward to the extension.

scrawfor99:
We can use this: https://github.com/opensearch-project/OpenSearch/pull/7452. The security plugin would implement the interface and then the Extension manager could pass everything into the RestSendToExtensionAction.

cwperks:
[Triage] Assigned to @scrawfor99 . Please update with more details on what the initial interface will look like.

scrawfor99:
Overview of the interface:

PR https://github.com/opensearch-project/OpenSearch/pull/7452, will introduce a new TokenManager interface to be used as a template for what an IdentityPlugin will expect any implementation to support. The TokenManager Interface defines five operations which are supposed to be implemented:

   1. public AuthToken issueToken(): Issues a new AuthToken
   2. public boolean validateToken(AuthToken token): Validates an auth token based on the rules associated with its format
   3. public String getTokenInfo(AuthToken token): Fetch the info from a token
   4. public void revokeToken(AuthToken token): Revokes a token that should no longer be treated as valid
   5. public void resetToken(AuthToken token): Updates a token to be valid for a greater period of time or to have different attributes.
 
Each of the implementing methods are relatively straightforward. The only ones which may require complicated handling are token issuance and token revocation. 

Issuing a token is likely to requiring overloading the method in some implementations of the IdentityPlugin. For example you may need to pass extension information to an IdentityPlugin for a new token to be created for that extension. At the same time, a key-based system could simply generate a random key with no information. 

Revoking a token can also be complicated depending on the implementation. For example, you would not want to create a revocation list for a token that encoded BasicAuth information. Instead, you may need to erase the token to effectively disable it.

scrawfor99:
https://github.com/opensearch-project/OpenSearch/pull/8679

cwperks:
@scrawfor99 Is there a companion security PR?

PR #3034, opened 2023-07-19
[Feature/Extension] Extension token handler

RyanL1997:
### Description
This PR is the refactor/modification based on @scrawfor99's previous PR(#2787) against main branch for Extension token handler. However, the scope of this PR is only for OBO related changes.
New Feature

### Issues Resolved
* Pre-requisite https://github.com/opensearch-project/OpenSearch/pull/8679
* Relate https://github.com/opensearch-project/security/issues/2573
* Resolve https://github.com/opensearch-project/security/issues/2764

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -146,6 +146,7 @@ import org.opensearch.security.filter.SecurityRestFilter;
 import org.opensearch.security.http.SecurityHttpServerTransport;
 import org.opensearch.security.http.SecurityNonSslHttpServerTransport;
 import org.opensearch.security.http.XFFResolver;
+import org.opensearch.security.identity.SecurityTokenManager;
 import org.opensearch.security.privileges.PrivilegesEvaluator;
 import org.opensearch.security.privileges.PrivilegesInterceptor;
 import org.opensearch.security.privileges.RestLayerPrivilegesEvaluator;
@@ -192,9 +193,13 @@ import org.opensearch.core.transport.TransportResponse;
 import org.opensearch.transport.TransportResponseHandler;
 import org.opensearch.transport.TransportService;
 import org.opensearch.watcher.ResourceWatcherService;
+
+import org.opensearch.identity.Subject;
+import org.opensearch.identity.tokens.TokenManager;
+import org.opensearch.plugins.IdentityPlugin;
 // CS-ENFORCE-SINGLE
 
-public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin implements ClusterPlugin, MapperPlugin {
+public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin implements ClusterPlugin, MapperPlugin, IdentityPlugin {
 
     private static final String KEYWORD = ".keyword";
     private static final Logger actionTrace = LogManager.getLogger("opendistro_security_action_trace");
@@ -208,6 +213,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin 
     private volatile SecurityInterceptor si;
     private volatile PrivilegesEvaluator evaluator;
     private volatile UserService userService;
+    private volatile SecurityTokenManager securityTokenManager;
     private volatile RestLayerPrivilegesEvaluator restLayerEvaluator;
     private volatile ThreadPool threadPool;
     private volatile ConfigurationRepository cr;
@@ -996,6 +1002,8 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin 
         cr = ConfigurationRepository.create(settings, this.configPath, threadPool, localClient, clusterService, auditLog);
 
         userService = new UserService(cs, cr, settings, localClient);
+
+        securityTokenManager = new SecurityTokenManager(threadPool, clusterService, cr, localClient, settings, userService);
 
         final XFFResolver xffResolver = new XFFResolver(threadPool);
         backendRegistry = new BackendRegistry(settings, adminDns, xffResolver, auditLog, threadPool);
@@ -1888,6 +1896,24 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin 
             return field.substring(0, field.length() - KEYWORD.length());
         }
         return field;
+    }
+
+    public static DiscoveryNode getLocalNode() {
+        return localNode;
+    }
+
+    public static void setLocalNode(DiscoveryNode node) {
+        localNode = node;
+    }
+
+    @Override
+    public Subject getSubject() {
+        return null;
+    }
+
+    @Override
+    public TokenManager getTokenManager() {
+        return securityTokenManager;
     }
 
     public static class GuiceHolder implements LifecycleComponent {
--- a/src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
+++ b/src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
@@ -13,6 +13,7 @@ package org.opensearch.security.dlic.rest.api;
 
 import java.io.IOException;
 import java.nio.file.Path;
+import java.security.NoSuchAlgorithmException;
 import java.util.List;
 import java.util.Map;
 
@@ -159,6 +160,8 @@ public class InternalUsersApiAction extends PatchableResourceApiAction {
             return;
         } catch (IOException ex) {
             throw new IOException(ex);
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
         }
 
         // for existing users, hash is optional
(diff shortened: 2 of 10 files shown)


RyanL1997:
**Moving some of the communication between me and @scrawfor99 over here**
1. The scope of this PR should only focus on the interface of `issueOnBehalfOfToken()`, reference to [here](https://github.com/opensearch-project/OpenSearch/pull/8679/files#diff-64c81937af16f1033656c0e817a517102ac5cf8f24854d35543ec7156170bfd2R24).
2. For the changes that got carried over from #2787, if the team thinks it is not in the scope of this PR it is ok to remove them. I'm leaving them here for now cuz I think it is easier to continue the service account work from here.

src/main/java/org/opensearch/security/user/UserTokenHandler.java
```diff
@@ -0,0 +1,184 @@
...
+        JwtToken jwt = jwtConsumer.getJwtToken();
+
+        Long iat = (Long) jwt.getClaim("iat");
+        Long exp = (Long) jwt.getClaim("exp");
+        SecurityDynamicConfiguration revokedTokens = load(getRevokedTokensConfigName(), false);
+        Long currentTime = System.currentTimeMillis();
+        return (exp > currentTime && !revokedTokens.exists(bearerAuthToken.getCompleteToken()));
+    }
+
+    public boolean validateJustInTimeToken(AuthToken authToken) {
```
RyanL1997, line 114: The reason I make this validation is that, I think we do have different expectations for obo tokens than the other bearer auth tokens. For obo tokens, according to its feature of just-in-time, we do expect the payload of `iat` and `nbf` have the same value.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -105,13 +105,16 @@ public String createJwt(
...
     ) throws Exception {
+        String tokenIdentifier = "obo";
         long timeMillis = timeProvider.getAsLong();
         Instant now = Instant.ofEpochMilli(timeProvider.getAsLong());
 
         jwtProducer.setSignatureProvider(JwsUtils.getSignatureProvider(signingKey));
         JwtClaims jwtClaims = new JwtClaims();
         JwtToken jwt = new JwtToken(jwtClaims);
 
+        jwtClaims.setProperty("typ", tokenIdentifier);
```
RyanL1997, line 116: This is the idea that can be also applied for the potential approach of handling the edge cases of OBO endpoint (see [comments](https://github.com/opensearch-project/security/pull/3008#discussion_r1263825789)). Since having a token type as the identifier will be easier for us to know if it is a obo token or not. So that we can have different expectations.

If we decide to go with this design, I will also update the threat model documentation

src/main/java/org/opensearch/security/user/UserTokenHandler.java
```diff
@@ -0,0 +1,184 @@
...
+            throw new UserServiceException("The provided token is not a BearerAuthToken.");
+        }
+        BearerAuthToken bearerAuthToken = (BearerAuthToken) authToken;
+        JwsJwtCompactConsumer jwtConsumer = new JwsJwtCompactConsumer(bearerAuthToken.getCompleteToken());
+        JwtToken jwt = jwtConsumer.getJwtToken();
+
+        Long iat = (Long) jwt.getClaim("iat");
+        Long nbf = (Long) jwt.getClaim("nbf");
+        Long exp = (Long) jwt.getClaim("exp");
+        SecurityDynamicConfiguration revokedTokens = load(getRevokedTokensConfigName(), false);
```
RyanL1997, line 125: If we choose to split the validations between obo tokens and other bearer auth tokens in this way, then I don't think we need this line here, since the revocation is not in the scope of obo token. (At least for now)

RyanL1997 (commented):
Adding some additional context here:

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -105,13 +105,16 @@ public String createJwt(
         List<String> roles,
         List<String> backendRoles
     ) throws Exception {
+        String tokenIdentifier = "obo";
```
cwperks, line 108: Can this variable be called `tokenType`?

Sometimes JWTs include a claim called `jti` that is a unique identifier for a token. In my experience, systems use `jti` to invalidate tokens since this identifier uniquely identifies a token.

src/main/java/org/opensearch/security/user/InternalUserTokenHandler.java
```diff
@@ -0,0 +1,100 @@
...
+import org.opensearch.identity.tokens.TokenManager;
+import org.opensearch.security.securityconf.Hashed;
+import org.opensearch.security.securityconf.impl.SecurityDynamicConfiguration;
+
+import java.io.IOException;
+import java.security.NoSuchAlgorithmException;
+
+import static org.opensearch.security.dlic.rest.support.Utils.universalHash;
+
+public class InternalUserTokenHandler implements TokenManager {
```
cwperks, line 27: This class looks like its doing too much. In the PR description it mentions that this PR is for `issueOnBehalfOfToken`. Are all of these methods being utilized in this class?

cwperks (commented):
@RyanL1997 There looks to be a lot more on this PR than an implementation of `issueOnBehalfOfToken`. Is everything in this PR needed for the issuance of obo tokens? Can we narrow the PR to the issuance of obo tokens and have a subsequent PR address the issuance of service account tokens?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1890,6 +1898,24 @@ private static String handleKeyword(final String field) {
         return field;
     }
 
+    public static DiscoveryNode getLocalNode() {
+        return localNode;
+    }
+
+    public static void setLocalNode(DiscoveryNode node) {
+        localNode = node;
+    }
```
DarshitChanpura, line 1907: this should be removed. Check this PR for more details: https://github.com/opensearch-project/security/pull/3066/files

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -159,6 +160,8 @@ protected void handlePut(RestChannel channel, final RestRequest request, final C
             return;
         } catch (IOException ex) {
             throw new IOException(ex);
+        } catch (NoSuchAlgorithmException e) {
```
DarshitChanpura, line 163: Why add this catch block to then throw RuntimeException?

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,221 @@
...
+
+    @Subscribe
+    public void onConfigModelChanged(ConfigModel configModel) {
+        this.configModel = configModel;
+    }
+
+    @Subscribe
+    public void onDynamicConfigModelChanged(DynamicConfigModel dcm) {
+        this.dcm = dcm;
+        if (dcm.getDynamicOnBehalfOfSettings().get("signing_key") != null
```
DarshitChanpura, line 214: we should extract: dcm.getDynamicOnBehalfOfSettings() to a variable

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -214,24 +214,8 @@ private void verifyServiceAccount(SecurityJsonNode securityJsonNode, String acco
...
-            numericCharacterRule,
-            specialCharacterRule
-        );
-        PasswordGenerator passwordGenerator = new PasswordGenerator();
-
-        Random random = Randomness.get();
-
-        return passwordGenerator.generatePassword(random.nextInt(8) + 8, rules);
+    private String generatePassword() {
+        return "superSecurePassword";
```
DarshitChanpura, line 218: why make this change?

src/main/java/org/opensearch/security/user/UserTokenHandler.java
```diff
@@ -0,0 +1,184 @@
...
+import org.opensearch.security.configuration.ConfigurationRepository;
+import org.opensearch.security.securityconf.DynamicConfigFactory;
+import org.opensearch.security.securityconf.impl.CType;
+import org.opensearch.security.securityconf.impl.SecurityDynamicConfiguration;
+import org.opensearch.threadpool.ThreadPool;
+
+import java.io.IOException;
+import java.util.Collections;
+
+public class UserTokenHandler {
```
DarshitChanpura, line 39: Is this class different that InternalUserTokenHandler in the sense that this class deals with actual user vs the other one deals with service accounts?
If so why not combine them into 1 class?

src/test/java/org/opensearch/security/auth/SecurityTokenManagerTests.java
```diff
@@ -0,0 +1,104 @@
...
+
+    UserService userService;
+
+    @Before
+    public void setup() {
+        claims = new HashMap<String, Object>() {
+            {
+                put(JwtConstants.CLAIM_AUDIENCE, "ext_0");
+            }
+        };
```
DarshitChanpura, line 66: can write this as: `Map.of(JwtConstants.CLAIM_AUDIENCE, "ext_0");`

PR #3036, opened 2023-07-19
[Backport 2.x] Update imports for files refactored in core PR #8157 (#3003)

DarshitChanpura:
### Description
Backports https://github.com/opensearch-project/security/commit/37aacdc0ceab4462498c9bdf339e6e0b5d30ef80 from #3003 

Also adds integration tests to 2.x line.

### Check List


DarshitChanpura:
Plugin install seems to be failing due to stale core artifact. Will wait for new artifact to be available before re-running these tasks.

The error below should not be thrown as that is the correct path for [Writeable.Reader](https://github.com/opensearch-project/OpenSearch/blob/83a4b6e9effc0e73ecba67e5dd9246e06b2c3df5/libs/core/src/main/java/org/opensearch/core/common/io/stream/Writeable.java#L166):
```
java.lang.NoClassDefFoundError: org/opensearch/core/common/io/stream/Writeable$Reader
```

DarshitChanpura:
Backwards-compatibility also seems to be failing only for windows due to stale artifact issue too.

src/integrationTest/java/org/opensearch/security/http/JwtAuthenticationTests.java
```diff
@@ -0,0 +1,270 @@
+/*
```
cwperks, line 1: Should the files in `integrationTest` be added to 2.x? @peternied thoughts?

src/main/java/org/opensearch/security/ssl/SecureSSLSettings.java
```diff
@@ -0,0 +1,130 @@
+/*
```
cwperks, line 1: Looks like this class is being added, but not utilized. Can you remove this change and/or include an entire backport of https://github.com/opensearch-project/security/pull/2296?

scripts/integtest.sh
```diff
@@ -101,7 +101,7 @@ then
 fi
 if [ -z "$COMMON_UTILS_VERSION" ]
 then
-  COMMON_UTILS_VERSION="2.9.0.0-SNAPSHOT"
+  COMMON_UTILS_VERSION="3.0.0.0-SNAPSHOT"
```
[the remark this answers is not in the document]
DarshitChanpura: Yes, this was accidental.

src/main/java/org/opensearch/security/ssl/SecureSSLSettings.java
DarshitChanpura: I'll create a separate backport for that and remove it from here.

src/integrationTest/java/org/opensearch/security/http/JwtAuthenticationTests.java
DarshitChanpura: We should have integrationTests for 2.x line.

cwperks: Are you able to run them? iirc there is more to the setup than backporting individual files in `src/integrationTest`?

DarshitChanpura:
Backwards compatibility tests are broken because the CI task checks out 2.x and current HEAD at 2.x doesn;t contain the import changes. (This PR brings those changes)

DarshitChanpura:
We either need to force-merge this to fix bwc tests or temporarily change the reference to current branch to `DarshitChanpura/backport/backport-3003-to-2.x/` and fix that once this PR is merged into 2.x

RyanL1997:
I think this is the similar issue happened on main, and if this is the case I think [Craig's procedure](https://github.com/opensearch-project/security/pull/3003#issuecomment-1638904833) can be also applied to here. But yeah, instead of referencing to the current 2.x, reference to current fix's branch will definitely resolve the issue.

cwperks:
bwc tests need to be revisited on the 2.x branch. I see a change in main that isn't in 2.x that could help the issue: https://github.com/opensearch-project/security/pull/2253

cwperks (approved):

cwperks:
I think the bwc tests in 2.x are doing too much. It should be testing from the 1.3 -> branch of the PR

cwperks:
@DarshitChanpura I believe you can update these lines: https://github.com/opensearch-project/security/blob/2.x/.github/workflows/ci.yml#L92-L93

The new values would be `2.9` and `current_branch` respectively

cwperks (approved):

RyanL1997 (approved):

Issue #3021, opened 2023-07-17
[BUG] [CI] Investigate Flaky test failure for windows CI tasks


DarshitChanpura:
**What is the bug?**

Windows CI has been flaky with recent runs for tasks, specifically  `citest`, `dlicRestApiTest`.

See these runs for example:

https://github.com/opensearch-project/security/actions/runs/5577190458/attempts/1 <-- 3 tasks failed
https://github.com/opensearch-project/security/actions/runs/5577190458/attempts/2 <-- 2 tasks failed (1 prev failed task passed)
https://github.com/opensearch-project/security/actions/runs/5577190458/attempts/3 <-- 1 task failed
https://github.com/opensearch-project/security/actions/runs/5577190458/attempts/5 <-- all tasks passed

<details>
  <summary>Failure Details</summary>

|        | (citest, windows-latest, 11)                                                                                                            | (citest, windows-latest, 17)                                                                                              |
|--------|----------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------|
[...]

PR #3046, opened 2023-07-24
Revert "Adds a check to skip serialization-deserialization if request is for same node (#2765) (#2973)

DarshitChanpura:
An issue was identified https://github.com/opensearch-project/security/issues/3021#issuecomment-1647895898 where un-reproducible test failures were observed, which suggest that #2973 might be the culprit.

We are reverting the commit ddbe517f48a688143d2ab764cfac487f29d4c050 until we have 100% confidence in ensuring that this PR is not the root cause.

### Check List


DarshitChanpura:
Converting it to draft until we identify whether it will have to be reverted.

cwperks:
@DarshitChanpura Can this be closed?

Issue #3055, opened 2023-07-26
[BUG] Create user API giving misleading error messages


derek-ho:
Reproduce:
-  with password validation message, even if password passes validation, and create user is only getting rejected due to weak password it is still showing the password validation error message, which can be frustrating

scrawfor99:
[Triage] @derek-ho just following-up.

PR #3057, opened 2023-07-26
Change password security message

derek-ho:
### Description
Returns the error message instead of the validation message because validation message should already have been shown in above case statements.

### Issues Resolved
Fix: #3055 

### Testing

### Check List


src/main/java/org/opensearch/security/dlic/rest/validation/AbstractConfigurationValidator.java
```diff
@@ -264,10 +264,7 @@ public XContentBuilder errorsAsXContent(RestChannel channel) {
                 case WEAK_PASSWORD:
                 case SIMILAR_PASSWORD:
                     builder.field("status", "error");
-                    builder.field(
-                        "reason",
-                        opensearchSettings.get(ConfigConstants.SECURITY_RESTAPI_PASSWORD_VALIDATION_ERROR_MESSAGE, errorType.message)
-                    );
+                    builder.field("reason", errorType.message);
```
cwperks, line 267: Hey @derek-ho, this change looks good to me. Would it be possible to write an integ test that uses the reset password API and show this change working?

You can look at [DashboardsInfoWithSettingsTest](https://github.com/opensearch-project/security/blob/main/src/integrationTest/java/org/opensearch/security/api/DashboardsInfoWithSettingsTest.java) for an example of setting up a node with different settings and using a REST client to hit security APIs.

This logic should be hit during reset_password, so it may be useful to have integration tests that cover different password scenarios. See https://opensearch.org/docs/latest/security/access-control/api/#change-password

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
```diff
@@ -286,14 +286,6 @@ public XContentBuilder toXContent(XContentBuilder builder, ToXContent.Params par
                         .get(SECURITY_RESTAPI_PASSWORD_VALIDATION_ERROR_MESSAGE, "Password does not match minimum criteria")
                 );
                 break;
-            case WEAK_PASSWORD:
```
derek-ho, line 289: Will be picked up in the default statement

willyborankin: I bit wrong here `ValidationError` has `WEAK_PASSWORD` and `SIMILAR_PASSWORD` errors.
AFAIU you want to get the same error message for all cases, so the solution in this case something like this:
```java

case INVALID_PASSWORD:
case WEAK_PASSWORD:
case SIMILAR_PASSWORD:
    builder.field("status", "error");
                builder.field(
                    "reason",
                    validationContext.settings()
                        .get(SECURITY_RESTAPI_PASSWORD_VALIDATION_ERROR_MESSAGE, "Password does not match minimum criteria")
                );
```
Otherwise you will end up in the `default` case:
which is: 
```java
    default:
                builder.field("status", "error");
                builder.field("reason", validationError.message());
                break;
```

willyborankin:
Checkstyle fails :-)

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
willyborankin: Here you can find different type of passwords to check:
https://github.com/opensearch-project/security/blob/6479513169e92a3781e9dd8da4ff32e0631b1bca/src/test/java/org/opensearch/security/dlic/rest/validation/PasswordValidatorTest.java#L30

willyborankin (dismissed):

willyborankin:
I think test failures are not related to this change.

cwperks (dismissed):

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
```diff
@@ -286,14 +286,6 @@ public XContentBuilder toXContent(XContentBuilder builder, ToXContent.Params par
...
-            case SIMILAR_PASSWORD:
-                builder.field("status", "error");
-                builder.field(
-                    "reason",
-                    validationContext.settings().get(SECURITY_RESTAPI_PASSWORD_VALIDATION_ERROR_MESSAGE, validationError.message())
```
peternied, line 294: Why remove this message, it was useful?

peternied (changes requested):
I think the previous behavior was better for users to act on, why remove the distinct message?

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
derek-ho: @peternied I removed this message because it can be misleading in some situations. For example:
- Set validation message = 'Password must be longer than 5 char'
- Try to create user with password 'aaaaaa'
- API returns 'Password must be longer than 5 char', even though this is already true, and it is actually just weak

By removing this case statement, weak and similar passwords to the username will instead be caught in the default statement, which would return a more specific message in the API, such as "Weak password", or "Similar password"

peternied: OK, I've looked into this - the way validation errors are used with the customizable messages is convoluted.  Lets not try to fix that right now.  Since no existing tests were broken lets move forward and we can come back to this later.

Thanks for looking into this issue

Root problem is that `ValidationError` is an enum with hardcoded messages these custom messages should have an alternative channel (e.g. CustomizableValidationMessage) to be populated at the source rather than that strange toXContent function,

peternied (approved):

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
peternied: > Since no existing tests were broken

And I just checked the CI to see that this wasn't the case, look at https://github.com/opensearch-project/security/actions/runs/5810367794

@derek-ho can you look into these failures and propose how you'd like to move forward?

peternied (changes requested):
CI failures due to broken tests

peternied (approved):
Thanks for cleaning up that code, this makes the documentation align better to how the product behaves.

cwperks (dismissed):

peternied (approved):

cwperks (approved):

DarshitChanpura:
@derek-ho Mind creating the backport manually?

PR #3066, opened 2023-07-28
Remove static local-node reference

DarshitChanpura:
### Description
Remove static reference/initialization of localNode variable inside security plugin, to fix `No user found..` errors caused due to mismatching localNode values in test. 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -212,7 +212,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin 
     private volatile ConfigurationRepository cr;
     private volatile AdminDNs adminDns;
     private volatile ClusterService cs;
-    private static volatile DiscoveryNode localNode;
+    private volatile AtomicReference<DiscoveryNode> localNode = new AtomicReference<>();
     private volatile AuditLog auditLog;
     private volatile BackendRegistry backendRegistry;
     private volatile SslExceptionHandler sslExceptionHandler;
@@ -776,7 +776,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin 
                             TransportRequestOptions options,
                             TransportResponseHandler<T> handler
                         ) {
-                            si.sendRequestDecorate(sender, connection, action, request, options, handler);
+                            si.sendRequestDecorate(sender, connection, action, request, options, handler, localNode.get());
                         }
                     };
                 }
@@ -1806,7 +1806,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin 
         if (!SSLConfig.isSslOnlyMode() && !client && !disabled) {
             cr.initOnNodeStart();
         }
-        this.localNode = localNode;
+        this.localNode.set(localNode);
         final Set<ModuleInfo> securityModules = ReflectionHelper.getModulesLoaded();
         log.info("{} OpenSearch Security modules loaded so far: {}", securityModules.size(), securityModules);
     }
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1806,7 +1805,9 @@ public void onNodeStarted(DiscoveryNode localNode) {
         if (!SSLConfig.isSslOnlyMode() && !client && !disabled) {
             cr.initOnNodeStart();
         }
-        this.localNode = localNode;
+        if (!SSLConfig.isSslOnlyMode() && si != null) {
+            si.setLocalNode(localNode); // set localNode
+        }
```
DarshitChanpura, line 1810: Followed same logic as on line 917 where si [is initialized](https://github.com/opensearch-project/security/blob/f6f2444223721ff3955c8faa720f5e43263a4b93/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L1052) only if the if condition below is false: https://github.com/opensearch-project/security/blob/f6f2444223721ff3955c8faa720f5e43263a4b93/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L917-L931

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1806,7 +1805,9 @@ public void onNodeStarted(DiscoveryNode localNode) {
         if (!SSLConfig.isSslOnlyMode() && !client && !disabled) {
             cr.initOnNodeStart();
```
cwperks, line 1806: Why can't `si.setLocalNode(localNode);` be called in this block?

It looks like the SecurityInterceptor component is created if this condition evaluates to true based on these lines: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L919-L943

DarshitChanpura: I'm not sure about when will client and disabled flags be set to true or false.

From the link that you mentioned, SecurityInterceptor will not be initialized if client or disabled flags are set to true.

but the condition in the present case explicitly asks `&& !client && !disabled` to be true  and so SecurityInterceptor localNode will not be initialized if either of client or disabled is set to true.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
peternied: Switching to AtomicReference fixed this issue, IMO much cleaner, I've pushed this change.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
peternied: We should use an AtomicReference to retrieve/capture this value when it comes in onNodeStarted(), I've pushed a change that captures this.  It still returns null if it wasn't set, but that is OK since the fallback handles this.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -130,7 +130,8 @@ public <T extends TransportResponse> void sendRequestDecorate(
         String action,
         TransportRequest request,
         TransportRequestOptions options,
-        TransportResponseHandler<T> handler
+        TransportResponseHandler<T> handler,
+        DiscoveryNode localNode
```
peternied, line 134: Making this a parameter for `sendRequestDecorate` makes more sense IMO

cwperks (approved):

peternied (approved):

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -165,7 +162,7 @@ public void testSendRequestDecorate() {
         assertEquals(threadPool.getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER), null);
 
         // isSameNodeRequest = false
-        securityInterceptor.sendRequestDecorate(sender, connection2, action, request, options, handler);
+        securityInterceptor.sendRequestDecorate(sender, connection2, action, request, options, handler, otherNode);
```
peternied, line 165: This check is now fully deterministic 👍

Issue #3090, opened 2023-08-02
[BUG] Backports onto 2.x often run into conflicts


peternied:
**What is the bug?**
We are having to manually resolve conflicts when attempting to backport changes from main into 2.x

**How can one reproduce the bug?**
1. Try to backport a change from main into core, such as https://github.com/opensearch-project/security/pull/2926
2. Wait on the pull request
3. See backport failed message

**What is the expected behavior?**
Conflicts should only occur when there is a significant difference - there are many changes that have not been backported that are causing conflicts

**Do you have any additional context?**
Seems like this is due to some large backports into 2.x that did not use the automated tooling and we've got changes that never made it into the 2.x branch

peternied:
Found the following dependency version number changes that need to be backported into 2.x:
- [ ] https://github.com/opensearch-project/security/pull/2926
- [ ] https://github.com/opensearch-project/security/pull/2925
- [ ] https://github.com/opensearch-project/security/pull/3071
- [ ] https://github.com/opensearch-project/security/pull/3072
- [ ] https://github.com/opensearch-project/security/pull/3074

peternied:
- taking this on as part of https://github.com/opensearch-project/security/issues/3028

PR #3092, opened 2023-08-02
[Backport 2.x] Synchronize major changes from main 

peternied:
### Description
Manual backports have been failing for a while, specifically the failure on https://github.com/opensearch-project/security/pull/2926 clued me into the issue, we've been manually resolving these conflicts, but we should be trying to keep everything in sync as much as possible, this set of backports lumped together attempts to handle that.

### Issues Resolved
- Resolves https://github.com/opensearch-project/security/issues/3090

### Testing
Lots of `/gradlew testClasses`, CI checks should verify everything else.

### Leaving Draft Checklist
- [x] Check in on CI status 
- [x] ~GitHub workflows need to migrate?~ Not going to worry about these for the moment
- [x] Are other backports stuck waiting that should be pulled in?
- [x] Clean up commit history so they look like correctly executed backports
- [X] ~Note; this change should NOT be squashed merged, I'll manually push so commit history is preserved~ Actually its been paired down anyways while dealing with other conflicts, just going to squash.

### Check List


build.gradle
```diff
@@ -24,8 +26,12 @@ buildscript {
 
         common_utils_version = System.getProperty("common_utils.version", '2.9.0.0-SNAPSHOT')
         kafka_version  = '3.5.1'
-        jaxb_version = '2.3.8'
+        apache_cxf_version = '4.0.2'
+        open_saml_version = '4.3.0'
```
willyborankin, line 30: Could we we exclude so far OpenSAML 4.3.0 version?

cwperks: +1, OpenSAML has a known issue with permissions that needs to be resolved before backport to 2.x.

There's a PR that will resolve the permissions issue here: https://github.com/opensearch-project/security/pull/2991

peternied: Thanks for catching this

peternied:
@cwperks @willyborankin Could I get another look?

willyborankin:
> @cwperks @willyborankin Could I get another look?

LGTM.

willyborankin (approved):

build.gradle
```diff
@@ -41,6 +47,8 @@ buildscript {
         maven { url "https://plugins.gradle.org/m2/" }
         maven { url "https://aws.oss.sonatype.org/content/repositories/snapshots" }
         maven { url "https://d1nvenhzbhpy0q.cloudfront.net/snapshots/lucene/" }
+        maven { url "https://build.shibboleth.net/nexus/content/groups/public" }
```
cwperks, line 50: The shibboleth repositories aren't needed after the reversion, but no harm in having them either.

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -168,9 +168,7 @@ private AuthTokenProcessorAction.Response handleImpl(
 
         try {
 
-            SamlResponse samlResponse = new SamlResponse(saml2Settings, null);
```
cwperks, line 171: Is this change fine even after reverting back to the original version of opensaml?

PR #3133, opened 2023-08-08
Automatically update GitHub Actions dependencies with dependabot

peternied:
### Description
Automatically update GitHub Actions dependencies with dependabot.  Noticed this was possible when I was reading the notice around CodeQL's v1 deprecation, [[codeql-action-v1-is-now-deprecated]](https://github.blog/changelog/2023-01-18-code-scanning-codeql-action-v1-is-now-deprecated/#can-i-use-dependabot-to-help-me-with-this-upgrade).

### Check List

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,9 @@ updates:
       # For all packages, ignore all major versions to minimize breaking issues
       - dependency-name: "*"
         update-types: ["version-update:semver-major"]
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "dependabot:"


.github/dependabot.yml
```diff
@@ -10,3 +10,9 @@ updates:
       # For all packages, ignore all major versions to minimize breaking issues
       - dependency-name: "*"
         update-types: ["version-update:semver-major"]
+  - package-ecosystem: "github-actions"
```
peternied, line 13: Why not include `version-update:semver-major` like was recently added onto gradle dependencies?

We have fewer (I think?) dependencies in GitHub Actions and I'd like to try it before I see lots of churn.  Easy to make this change as a fast follow if there is concern.

reta (approved):

.github/dependabot.yml
reta: I think for Github actions the update to major is not as breaking as for main codebase, most do major anyway (fe codeql v1 -> v2, checkout v2 -> v3, ...)

cwperks (approved):

PR #3150, opened 2023-08-10
Only run CI on push for 'official' branches

peternied:
### Description
CodeQL is throwing errors when run inside of branches created by dependabot requests.  Also we've had issue with a flood of CI checks that were redudant when inspecting pull request check results.  This should limit the 'doubling' up unless you are making a pull request from a fork's main branch.

image

### Check List

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,12 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+      - 1.*
+      - 2.*
+  pull_request:
 
 env:
   GRADLE_OPTS: -Dhttp.keepAlive=false


peternied:
Build failures are due to an incoming breaking change, need that resolved before this can be merged

- https://github.com/opensearch-project/security/issues/3151

cwperks (approved):

.github/workflows/ci.yml
```diff
@@ -1,6 +1,12 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+      - 1.*
+      - 2.*
+  pull_request:
```
cwperks, line 9: In this valid yaml. Does it need one list element like `- *`?

peternied: Plan to keep it as is, 'Valid YAML!' according to https://www.yamllint.com/

Looks like an alternative expression would be to indicate `null` explicitly.
```yaml
on:
  push:
    branches:
      - main
      - 1.*
      - 2.*
  pull_request: null
env:
```
^ At least that is what a linter did.

RyanL1997 (approved):

RyanL1997:
@peternied backport to 2.x, 1.x/1.3?

PR #3180, opened 2023-08-15
Feature/extensions bwc setting

### LINKED ISSUE #2616: [Extensions] Add `opensearch.yml` setting to enable backward compatible plugin mode for extensions

samuelcostae:
### Description
This Draft PR includes the new setting bwcPluginMode (backward compatible plugin mode for extensions ) 

### Issues Resolved
#2616 

### Testing

### Check List


src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -142,7 +147,10 @@ public String createJwt(
             throw new Exception("Roles cannot be null");
         }
 
-        /* TODO: If the backendRoles is not null and the BWC Mode is on, put them into the "dbr" claim */
+        if (bwcModeEnabled && backendRoles != null) {
```
samuelcostae, line 150: @DarshitChanpura  i wonder if just keeping the format used a few lines above would be ok, or if for the backwards compatibility to work we should keep the exact format currently used when adding to the Context:

`            StringJoiner joiner = new StringJoiner("|");
            joiner.add(user.getName());
            joiner.add(String.join(",", Sets.union(user.getSecurityRoles(), mappedRoles)));`

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -142,7 +145,10 @@ public String createJwt(
...
+            String listOfBackendRoles = String.join(",", backendRoles);
+            jwtClaims.setProperty("br", EncryptionDecryptionUtil.encrypt(claimsEncryptionKey, listOfBackendRoles));
```
cwperks, line 150: If this setting is enabled then send the backend roles unencrypted. This setting will determine whether backend roles (`br`) is included or excluded as a claim in the token.

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -320,6 +320,9 @@ public enum RolesMappingResolution {
     public static final String TENANCY_GLOBAL_TENANT_NAME = "global";
     public static final String TENANCY_GLOBAL_TENANT_DEFAULT_NAME = "";
 
+    public static final String EXTENSIONS_BWC_PLUGIN_MODE = "bwcPluginMode";
```
cwperks, line 323: `BWC_PLUGIN_MODE` may not be the best name to capture what this setting does. In an early draft of security for extensions it was called this because it was not fully known yet what it would take for an extension to be backward compatible with plugins. 

wdyt about calling this setting `EXTENSIONS_INCLUDE_BACKEND_ROLES`?

cwperks:
Thank you @samuelcostae ! This PR looks good to me. Can you add a test in JwtVendorTest to show how this setting affects token issuance?

src/main/java/org/opensearch/security/support/ConfigConstants.java
RyanL1997: I just initiated the run of CI. I doubt the naming including 'EXTENSIONS' gonna pass the lint task, since we do have restrictions of using it.

samuelcostae: `jwtTokenIncludesBackendRoles` ?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1111,6 +1116,21 @@ public Settings additionalSettings() {
         return builder.build();
     }
 
+    @Override
+    public List<Setting<?>> getExtensionSettings() {
+        List<Setting<?>> extentionSettings = new ArrayList<Setting<?>>();
```
DarshitChanpura, line 1121: nit: extensionSettings

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
DarshitChanpura: I think the idea behind this is to send backend roles unencrypted if its in backwards compatibility mode. Can you please elaborate on `keep the exact same format`?

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
DarshitChanpura: +1. Based on this TODO comment:
```
/* TODO: If the backendRoles is not null and the BWC Mode is on, put them into the "dbr" claim */
```
@samuelcostae Please update it to reflect non-encrypted backend role `dbr`

DarshitChanpura (commented):
One generic comment to add tests, and look into codecov warnings for files you touched.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
samuelcostae: The code snipped above that i pasted in the above post is how the back end roles are added to the context (joinined with " | " symbol and  adding the user's name  at the begininng ). 

My question if we should populate using the same format as the plugins might be expecting/parsing that exact format and could fail if the joiner character is different. 

(Im not aware if this is the case or not)

DarshitChanpura: Ah I see, this is only for token parsing so we should be good with the format you have in this PR. @RyanL1997 Can you confirm?

DarshitChanpura (approved):

src/main/java/org/opensearch/security/support/ConfigConstants.java
RyanL1997: I think that should be fine since this mode is targeting for the usage of the extension.

RyanL1997 (approved):

Issue #3137, opened 2023-08-09
[BUG] Unexpected Coverage Changes on SelfRefreshingKeySet, SecurityInterceptor and SecuritySSLNettyTransport


peternied:
**What is the bug?**
There are several classes that are seeing random behavior in code coverage reports.  Test should be added/modified to ensure these code paths are always exercised - keeping codecov results useful.

- [x] src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySet.java
  - Has tests to attempt race conditions, might need to alter the class and tests to be able to exercise the 'races' guaranteed
- [x] src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
  - Needs a tweak, or additional unit test to confirm a local node vs remote node scenario
- [x] src/main/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransport.java
  - Needs unit tests
- [x] src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
  - Needs unit tests that exercise exceptions around ClusterHealthRequest
  
![image](https://github.com/opensearch-project/security/assets/2754967/dcae6001-32ad-4ba7-90b8-e06a22af1f5e)

Codecov [[Report link]](https://app.codecov.io/gh/opensearch-project/security/pull/3133/indirect-changes)
Codecov [[Unexpected Coverage Changes]](https://docs.codecov.com/docs/unexpected-coverage-changes)

**How can one reproduce the bug?**
1. Create a pull request that only changes markdown file
2. Look at the checks for the pull requset
3. See the codecov check
4. See error that coverage has dropped .0X%

**What is the expected behavior?**
Code coverage should not fluctuate, testing should be deterministic and 'exhaustive' 

**Do you have any additional context?**
This is causing code coverage failures on many pull requests and code is being merged without all checks passing regularly.

scrawfor99:
[Triage] Thank you for filing this issue @peternied. Based on the provided checklist, I will label this issue as actionable and triaged.

davidosorno:
@peternied I Started looking into it. Please let me know if you have any useful information that I can use to address this.

peternied:
@davidosorno I'm excited to see these get fixed, In the descriptions 'checklist' is most of my thoughts - we can dig in more with a pull request - I'd recommend to create a draft, happy to review if you've got questions.

scrawfor99:
The second and third check boxes should be all set shortly.

PR #3196, opened 2023-08-16
Add code coverage exclusions on false positives

peternied:
### Description
Adding code coverage exclusions on files that are creating code coverage measurement issues.

### Issues Resolved
- Related https://github.com/opensearch-project/security/issues/3137

### Testing
Will be checking the codecov output on this PR and expect it to come back clean.  Might need to iterate a couple times to make sure the max threshold % amount is correct.

### Check List

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,28 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 0%
+      # Exclusions to prevent false positives triggered
+      # Tracking with https://github.com/opensearch-project/security/issues/3137
+      keyset:
+        paths:
+          - "src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySet.java"
+        target: auto
+        threshold: 10%
+      securityinterceptor:
+        paths:
+          - "src/main/java/org/opensearch/security/transport/SecurityInterceptor.java"
+        target: auto
+        threshold: 10%
+      sslnettytransport:
+        paths:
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -0,0 +1,25 @@
+coverage:
+  status:
+    project:
+      default:
```
peternied, line 4: Hmm, isn't clear that its putting default into one bucket and these files into another, need to look into this...

codecov.yml
```diff
@@ -0,0 +1,25 @@
...
+        target: auto
+        threshold: 0%
+      # Exclusions to prevent false positives triggered
+      # Tracking with https://github.com/opensearch-project/security/issues/3137
+      inconsistant-coverage-files:
```
cwperks, line 9: nit: inconsistent

codecov.yml
peternied: Looks like the exclusions must be manually declared, was able to see this by looking at the coverage configuration at the commit level - https://app.codecov.io/gh/opensearch-project/security/pull/3196/commits

peternied: I've resolved this by seperating the plugin out with its own exclusions which is working as intented

image

DarshitChanpura (approved):

cwperks (approved):
@peternied I'm approving this PR, but also had a question. Could this be achieved using java annotations?

peternied:
> Could this be achieved using java annotations?

Codecov doesn't support any java annotations that I'm aware of.  The only really annotations that I've seen used in the code coverage space are about automatically excluding code that was generated, such as by adding your own @.generated annotation [example](https://www.baeldung.com/jacoco-report-exclude#excluding-with-custom-annotation).

PR #3231, opened 2023-08-22
Clean up sysouts and replace with logger statements where necessary

DarshitChanpura:
### Description
A lot of files (mostly tests) have sysout's which do nothing but degrade the performance. This PR removes all sysouts from tests, replace sysouts with logger in project files.
Also removes all e.printStacktrace(); calls.

- `tools/` package  - no sysouts were removed from here
- HTTPSpnegoAuthenticator.java - a portion was not modified since it enabled debugging on console.

### Check List


DarshitChanpura:
Taking a look at the test failure

src/main/java/org/opensearch/security/tools/AuditConfigMigrater.java
```diff
@@ -43,6 +45,8 @@ public class AuditConfigMigrater {
     private static final HelpFormatter formatter = new HelpFormatter();
     private static final CommandLineParser parser = new DefaultParser();
 
+    static final Logger log = LogManager.getLogger(AuditConfigMigrater.class);
```
peternied, line 48: There aren't any tests on this tool, this could be introducing breaking changes in the log output, what do you think about avoiding this file?

peternied: Same for Mitgrater.java

peternied (commented):
Nice - thanks for keeping the console output cleaner.  Can you see about adding an rule to prevent system.out/system.err usage to keep the codebase programmatically cleaned?

Note; I think adding exceptions is fine for a couple places we do want console output

src/test/java/org/opensearch/security/ConfigTests.java
```diff
@@ -58,30 +58,25 @@ public void testEmptyConfig() throws Exception {
     @Test
     public void testMigrate() throws Exception {
 
+        // TODO add a better way to test it rather than adding sysouts
```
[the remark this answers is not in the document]
DarshitChanpura: hmm.. I'm not planning to resolve them in this PR. I will remove TODO

DarshitChanpura:
> Nice - thanks for keeping the console output cleaner. Can you see about adding an rule to prevent system.out/system.err usage to keep the codebase programmatically cleaned?
> 
> Note; I think adding exceptions is fine for a couple places we do want console output

I was contemplating between adding such check and not adding it, but your comment helped decide :D. I have added a github check.

.github/workflows/code-hygiene.yml
```diff
@@ -83,3 +83,17 @@ jobs:
           fi
         done
         exit $exit_code
+
+  check-println:
```
peternied, line 87: Instead of a new check, what about leveraging checkstyle's RegexpSingleline, below was how the 'extension' filter was added, maybe we could add entries for system.out.print & system.err.print that would be easier to maintain

```xml
  <module name="RegexpSingleline">
    <property name="format" value="extension"/>
    <property name="ignoreCase" value="true"/>
    <property name="message" value="Extension should only be used sparingly to keep implementations as generic as possible" />
    <property name="severity" value="error"/>
  </module>
```

DarshitChanpura: I tried that but didn't have a working solution, I'll invest some more time to see if I can make it work

DarshitChanpura: I was not able to have it running with:
```xml
  <module name="RegexpSingleline">
    <property name="format" value="System\.out\.println" />
    <property name="message" value="Do not use System.out.println" />
    <!-- Exception list of files and directories -->
    <property name="fileExtensions" value="java" />
    <property name="exclude" value="./src/main/java/org/opensearch/security/tools, ./src/main/java/com/amazon/dlic/auth/http/kerberos/HTTPSpnegoAuthenticator.java" />
  </module>
```

DarshitChanpura:
> Taking a look at the test failure

Fixed

.github/workflows/code-hygiene.yml
peternied: Try building off of this, works locally for me
```xml
  <module name="RegexpSingleline">
    <property name="format" value="println"/>
    <property name="ignoreCase" value="true"/>
    <property name="message" value="SYSTEM.OUT.PRINTLN NONONO!" />
    <property name="severity" value="error"/>
  </module>
```

```
./gradlew checkstyleMain 
...
> Task :checkstyleMain
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java:247: SYSTEM.OUT.PRINTLN NONONO! [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java:248: SYSTEM.OUT.PRINTLN NONONO! [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/auditlog/sink/AuditLogSink.java:63: SYSTEM.OUT.PRINTLN NONONO! [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/auditlog/sink/DebugSink.java:30: SYSTEM.OUT.PRINTLN NONONO! [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/ssl/OpenSearchSecuritySSLPlugin.java:161: SYSTEM.OUT.PRINTLN NONONO! [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/ssl/OpenSearchSecuritySSLPlugin.java:162: SYSTEM.OUT.PRINTLN NONONO! [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/ssl/OpenSearchSecuritySSLPlugin.java:228: SYSTEM.OUT.PRINTLN NONONO! [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/ssl/OpenSearchSecuritySSLPlugin.java:229: SYSTEM.OUT.PRINTLN NONONO! [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/tools/AuditConfigMigrater.java:94: SYSTEM.OUT.PRINTLN NONONO! [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/tools/AuditConfigMigrater.java:109: SYSTEM.OUT.PRINTLN NONONO! [RegexpSingleline]
[ant:checkstyle] [ERROR] /local/home/petern/git/security/src/main/java/org/opensearch/security/tools/AuditConfigMigrater.java:112: SYSTEM.OUT.PRINTLN NONONO! [RegexpSingleline]
```

DarshitChanpura: Yes, I the problem is how to exclude. There is no such Property called `exclude`. Do you know of anything of that might work?
Closest thing I found was this: https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/sun_checks.xml#L46-L50
However this applies to whole file, so I might have to end up creating a separate xml for sysout check anyway.

DarshitChanpura: I was able to achieve this by adding a separate file for println checks.

peternied: That too hard - exclusions can be done in place, see https://github.com/opensearch-project/security/blob/main/DEVELOPER_GUIDE.md#checkstyle-violations

willyborankin (approved):

peternied (approved):

Issue #1815, opened 2022-05-02
[BUG] do_not_fail_on_forbidden_empty does not work for cat api


jezsy:
**What is the bug?**
Setting `do_not_fail_on_forbidden` to true does not seem to have any impact on some APIs like cat/_indices and cat/_aliases - the whole operation is rejected even if indices exist for which the user has access to. 

_cat/indices/some-index-* would work, but not _cat/indices because it include internal indices which the user does not permission for. 

Adding permissions under index_patterns: '*' can make them work, but I think this defeats the purpose of having the do_not_fail_on_forbidden parameter.

**How can one reproduce the bug?**
roles.yml
```
my_user:
  cluster_permissions:
    - "cluster_monitor"
    - "cluster_manage_index_templates"
  index_permissions:
  - index_patterns:
    - "some-index-*"
    allowed_actions:
      - "read"
      - "write"
      - "create_index"
      - "manage"
```
config.yml
```
config:
  dynamic:
    do_not_fail_on_forbidden: true
    do_not_fail_on_forbidden_empty: true
    ...
```
Error for GET _cat/indices
```
        "type" : "security_exception",
        "reason" : "no permissions for [indices:monitor/settings/get] and User [name=my_user, backend_roles=[], requestedTenant=null]"

```
Error for GET _cat/aliases
```
        "type" : "security_exception",
        "reason" : "no permissions for [indices:admin/aliases/get] and User [name=my_user, backend_roles=[], requestedTenant=null]"
```
Both of these permissions should be covered under the `manage` default_action_group, but the operations only work when targeted specifically to the index, so do_not_fail_on_forbidden does not seem to be working as intended.

**What is the expected behavior?**
do_not_fail_on_forbidden should work similarly for all APIs, where results are filtered based only on the indices that the  user has permissions for. 

**What is your host/environment?**
 - Version OpenSearch 1.2.3
 - Plugins OpenSearch Security 1.2.3.0

peternied:
Thanks for filing @jezsy, I am having trouble reproducing the issue with `/_search`.  As that data might be sensitive, could you email me directly at petern@amazon.com?  Sending the `security-auditlog-*` log entries, should give me the data I need to dive into this further.

jezsy:
[...]

PR #3236, opened 2023-08-24
Command `cat/indices` will filter results per the Do Not Fail On Forbidden setting

derek-ho:
### Description
This change allows for DNFOF behavior on the _cat/_indices API. It adds the required index permissions into the DNFOF regex to be picked up in the DNFOF code path. Previously it was being skipped/returning 403, since the index permissions were not in the regex.

### Issues Resolved
Fix: #1815

### Testing

### Check List


src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -104,7 +104,7 @@ public class PrivilegesEvaluator {
     private static final WildcardMatcher ACTION_MATCHER = WildcardMatcher.from("indices:data/read/*search*");
 
     private static final Pattern DNFOF_PATTERNS = Pattern.compile(
-        "indices:(data/read/.*|(admin/(mappings/fields/get.*|shards/search_shards|resolve/index)))"
+        "indices:(data/read/.*|(admin/(mappings/fields/get.*|shards/search_shards|resolve/index))|(monitor/((settings/get)|(stats))))"
     );
```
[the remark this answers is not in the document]
derek-ho: Also not sure if this covers all applicable CAT API calls - is there anymore calls we want to enable DNFOF for in this same PR?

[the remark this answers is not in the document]
peternied: Can you add a unit test around the what the pattern matches / doesn't match?

src/integrationTest/java/org/opensearch/security/privileges/PrivilegesEvaluatorDNFOFTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.opensearch.test.framework.cluster.TestRestClient;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class PrivilegesEvaluatorDNFOFTest {
```
peternied, line 20: I see another class, DoNotFailOnForbiddenTests, is there a reason we need a new test class?

derek-ho: Moved to existing testing file in: [92d475b](https://github.com/opensearch-project/security/pull/3236/commits/92d475bf02c966ae6359eeb222f234cd6766a003). My only concern now is that it seems to be an integration test file, should I move the unit/regex matching tests to another file?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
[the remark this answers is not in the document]
derek-ho: Added a unit test in: [92d475b](https://github.com/opensearch-project/security/pull/3236/commits/92d475bf02c966ae6359eeb222f234cd6766a003), only question is if it is ok living in the existing DNFOF file or not.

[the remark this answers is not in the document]
derek-ho: But my original question still stands - do we think that a list would be cleaner/clearer than matching against a regex pattern?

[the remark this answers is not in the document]
peternied: All test under `src/integrationTest` create a fresh cluster with each test (or if not they should move...)  When I say unit test it could be a test that just accesses the regex static, not even making an instance of PrivilegesEvaluator

src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
```diff
@@ -408,4 +434,32 @@ public void shouldPerformStatAggregation_negative() throws IOException {
...
+            assertThat(DNFOF_PATTERNS.matcher(permission).matches(), equalTo(false));
+        }
+    }
+
+    @Test
+    public void testDnfofPermissions_positive() {
+        for (final String permission : allowedDnfof) {
+            assertThat(DNFOF_PATTERNS.matcher(permission).matches(), equalTo(true));
+        }
+    }
```
peternied, line 463: Both testDnfofPermissions_negative and testDnfofPermissions_positive are spinning up a cluster, because of the @Class rule in this file.  Looks like `PrivilegesEvaluatorUnitTest.java` already exists to handle cases like this, can you move these tests over there?

src/test/java/org/opensearch/security/privileges/PrivilegesEvaluatorUnitTest.java
```diff
@@ -8,14 +8,95 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.opensearch.security.privileges.PrivilegesEvaluator.DNFOF_PATTERNS;
 import static org.opensearch.security.privileges.PrivilegesEvaluator.isClusterPerm;
 
 public class PrivilegesEvaluatorUnitTest {
 
+    private static final List<String> allowedDnfof = ImmutableList.of(
```
derek-ho, line 25: I just copied a list of all index permissions found on the documentation website - am I understanding DNFOF correctly in that its only an index level permission filter?

src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
derek-ho: Moved in - [4185872](https://github.com/opensearch-project/security/pull/3236/commits/41858729da8b2af78dd7bde4267613deaf611fd2).

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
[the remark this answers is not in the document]
derek-ho: Moved in [4185872](https://github.com/opensearch-project/security/pull/3236/commits/41858729da8b2af78dd7bde4267613deaf611fd2)

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -103,8 +103,8 @@ public class PrivilegesEvaluator {
 
     private static final WildcardMatcher ACTION_MATCHER = WildcardMatcher.from("indices:data/read/*search*");
 
-    private static final Pattern DNFOF_PATTERNS = Pattern.compile(
-        "indices:(data/read/.*|(admin/(mappings/fields/get.*|shards/search_shards|resolve/index)))"
+    public static final Pattern DNFOF_PATTERNS = Pattern.compile(
+        "indices:(data/read/.*|(admin/(mappings/fields/get.*|shards/search_shards|resolve/index))|(monitor/((settings/get)|(stats))))"
```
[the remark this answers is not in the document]
derek-ho: I'm not familiar with this option - I included all index permissions in a unit test so that should be pretty clear/tell us if something in the regex changed in an unexpected way, but this may be more clear. Any advantage of WildcardMatcher over checking inclusion in a set/something similar?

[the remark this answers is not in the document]
derek-ho: I ended up turning it into a function like the isClusterPerm, I think that is relatively clean, maintainable, and consistent with the code base.

src/test/java/org/opensearch/security/privileges/PrivilegesEvaluatorUnitTest.java
derek-ho: Confirmed that this is the case

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -674,6 +669,15 @@ public static boolean isClusterPerm(String action0) {
         );
     }
 
+    static boolean isDnfofAction(String action0) {
```
willyborankin, line 672: I think `WildcardMatcher` can be used here. Whdyt?

willyborankin, line 672: I think `WildcardMatcher` can be used here. Whdyt?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -674,6 +669,15 @@ public static boolean isClusterPerm(String action0) {
...
+        return (action0.startsWith("indices:data/read/")
+            || (action0.startsWith("indices:admin/mappings/fields/get")
+                || action0.equals("indices:admin/shards/search_shards")
+                || action0.equals("indices:admin/resolve/index"))
+            || action0.equals("indices:monitor/settings/get")
+            || action0.equals("indices:monitor/stats"));
+    }
```
peternied, line 679: Not sure why there was nesting around some of those predicates, they all seem mutually exclusive

```suggestion
        return (action0.startsWith("indices:data/read/")
            || action0.startsWith("indices:admin/mappings/fields/get")
            || action0.equals("indices:admin/shards/search_shards")
            || action0.equals("indices:admin/resolve/index")
            || action0.equals("indices:monitor/settings/get")
            || action0.equals("indices:monitor/stats"));
    }
```

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
derek-ho: Used in : [c12333c](https://github.com/opensearch-project/security/pull/3236/commits/c12333ce5e1442bdd206d23664a5b71e43787203)

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
derek-ho: Might have been a typo/spotless issue. I changed the implementation to wildcardMatcher in: [c12333c](https://github.com/opensearch-project/security/pull/3236/commits/c12333ce5e1442bdd206d23664a5b71e43787203), so I think this no longer applies.

peternied (approved):

RyanL1997 (approved):

peternied:
Backport likely failed due to the integration tests not being backported

Issue #1822, opened 2022-05-04
[Refactor] Rename `.opendistro_security` index to `.opensearch_security`


DarshitChanpura:
This change is part of the renaming effort last year for the project from OpenDistro to OpenSearch. As a part of this effort:

- [ ] Look up `opendistro_security` and rename it to `opensearch_security`
- [ ] All the test cases should pass after this change
- [ ] Sanity test by installing this plugin in OpenSearch

cliu123:
The plan is not to rename indices. 
Related discussion: https://github.com/opensearch-project/opensearch-plugins/issues/17#issuecomment-847204447

peternied:
We have a couple of conversations in flight about how/where security data is stored.  As we move forward in these efforts we can shift the index name.

PR #3237, opened 2023-08-24
Adds .opendistro-job-scheduler-lock to list of protected indices

joshpalis:
### Description
Companion PR (Need to merge first) : https://github.com/opensearch-project/job-scheduler/pull/474

In addition to converting the `.opendistro-job-scheduler-lock` index to a system index, adding the index name to the security plugin's list of protected indices within the linux/windows installation scripts.

### Issues Resolved
https://github.com/opensearch-project/job-scheduler/issues/466
https://github.com/opensearch-project/job-scheduler/issues/305

### Check List


willyborankin (approved):

tools/install_demo_configuration.bat
```diff
@@ -315,7 +315,7 @@ echo plugins.security.enable_snapshot_restore_privilege: true >> "%OPENSEARCH_CO
...
+echo plugins.security.system_indices.indices: [".plugins-ml-config", ".plugins-ml-connector", ".plugins-ml-model-group", ".plugins-ml-model", ".plugins-ml-task", ".opendistro-alerting-config", ".opendistro-alerting-alert*", ".opendistro-anomaly-results*", ".opendistro-anomaly-detector*", ".opendistro-anomaly-checkpoints", ".opendistro-anomaly-detection-state", ".opendistro-reports-*", ".opensearch-notifications-*", ".opensearch-notebooks", ".opensearch-observability", ".ql-datasources", ".opendistro-asynchronous-search-response*", ".replication-metadata-store", ".opensearch-knn-models", ".geospatial-ip2geo-data*", ".opendistro-job-scheduler-lock"] >> "%OPENSEARCH_CONF_FILE%"
```
peternied, line 318: Is it possible to rename this to opensearch, its been a long time since the project was opendistro

cwperks: @peternied is there general migration guidance on renaming system indices? The security index is `.opendistro_security`

joshpalis: While I agree with the name change, I'm concerned how this may affect backwards compatibility/upgrades.

peternied: > I'm concerned how this may affect backwards compatibility/upgrades.

Isn't the transition from non-system to system index going to force these same concerns?  If so - I'd like use to take the chance to rename.

joshpalis: Not necessarily, there's no direct user interaction with the `.opendistro-job-scheduler-lock` index so converting this into a system index just adds guardrails so a user does not accidentally delete it. But regarding the index name change there is clear [documentation](https://github.com/opensearch-project/opensearch-plugins/blob/main/UPGRADING.md#indices) against renaming indices ([reference issue](https://github.com/opensearch-project/opensearch-plugins/issues/17#issuecomment-847204447)).

joshpalis: I see the same justification was used to decide against renaming the security system index : https://github.com/opensearch-project/security/issues/1822

peternied (approved):

PR #3263, opened 2023-08-29
Add triage agenda items for waiting on quorum and member requests

peternied:
### Description
During the triage meeting on Aug 27th, based on time constraints member requests were opened up earlier than previously scheduled on the agenda.  This pull request captures this change to ensure meeting members can more quickly discuss an issue/PR.  The opening phase of the triage meeting was not captured, and has been included to document the informal and open tone of the meeting.

### Check List


TRIAGING.md
peternied, line 1: Recommended: Use the rich diff view
image

TRIAGING.md
```diff
@@ -20,13 +20,16 @@ If you have an issue you'd like to bring forth please consider getting a link to
 
 ### Is there an agenda for each week?
 
-Meetings are lightly structured as follows:
+Meetings 60 minutes and structured as follows:
+
+1. Wait for quorum: As we gather and waiting for at least three maintainers to join, feel free to turn on video and have informal and open-to-all conversation.  Once quorum is reached, a volunteer will share their screen and proceed with the agenda.
```
davidlago, line 25: What if quorum is never reached? maybe we only have 2 maintainers join. Are we scrapping the meeting? or are there just specific sections that we want to have with enough quorum?

peternied: I think whoever is in the meeting can make a call what to do or not.  I wouldn't expect the same structure to be honored if this was the case.   Not sure I'd want to capture that in this agenda lines, or do you have a recommendation?

willyborankin (approved):

TRIAGING.md
davidlago: I'm not sure we should put anything about quorum at all if we could start w/o reaching one. If this ends up reading "wait for quorum, but if quorum is not reached don't wait", then it's not saying much. If the intent behind this point is more of a "let's give folks a few mins to join", I'd either say that or just remove the paragraph altogether.

TRIAGING.md
```diff
@@ -20,13 +20,16 @@ If you have an issue you'd like to bring forth please consider getting a link to
 
 ### Is there an agenda for each week?
 
-Meetings are lightly structured as follows:
+Meetings is 60 minutes and structured as follows:
```
davidlago, line 23: `s/is/are`

davidlago (approved):

peternied:
@willyborankin Care to take another look?

willyborankin (approved):

reta (approved):

Issue #3174, opened 2023-08-14
[BUG] The node certificate in the demo certificates does not include the IPv6 loopback address `::1`


cwperks:
The demo `esnode.pem` certificate (defined [here](https://github.com/opensearch-project/security/blob/main/tools/install_demo_configuration.sh#L260-L289)) does not include `::1` as a subject alternate name which has caused issues running the backwards compatibility tests with the security plugin installed.

```
> openssl x509 -noout -text -in 'config/esnode.pem'

X509v3 Subject Alternative Name:
                Registered ID:1.2.3.4.5.5, DNS:node-0.example.com, DNS:localhost, IP Address:127.0.0.1
```

The IPv6 loopback address can be added in an ext file using openssl like this:

```
openssl genrsa -out esnode-key-temp.pem 2048
openssl pkcs8 -inform PEM -outform PEM -in esnode-key-temp.pem -topk8 -nocrypt -v1 PBE-SHA1-3DES -out esnode-key.pem
openssl req -new -key esnode-key.pem -subj "/C=de/L=test/O=node/OU=node/CN=node-0.example.com" -out esnode.csr
openssl x509 -req -days 3650 -extfile <(printf "subjectAltName=DNS:node-0.example.com,DNS:localhost,IP:::1,IP:127.0.0.1,RID:1.2.3.4.5.5") -in esnode.csr -out esnode.pem -CA root-ca.pem -CAkey root-ca-key.pem
```

scrawfor99:
[Triage] The outcome of this would require generating a new set of certificates. The steps to generate new certificates can be found online and in past PRs to generate certs.

cwperks:
FYI if the Demo Certificates are replaced then new hashes need to be added here: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L305-L319

cwperks:
@DarshitChanpura Can you reply back with the commands to generate the cert checksums when you figure it out?

DarshitChanpura:
## Solution Approach:

NOTE: Since this required adding a new value to SAN for node certificates, I had to end up generating a new set of demo certs.

1. Generate a new root certificate as Signing authority while ensuring all X509 extensions are preserved.
2. Generate a new node certificate + key signed with root-ca while ensuring all X509 extensions are preserved.
3. Generate a new admin certificate + key signed with root-ca while preserving all X509 extensions.
[...]

PR #3268, opened 2023-08-29
Generate new demo certs with IPv6 loopback added to SAN in node certificate

DarshitChanpura:
### Description
This PR solves bug where backwards compatibility tests would fail for IPv6 loopback address (`::1`) with:
`No subject alternative names matching IP address ::1 found`

* Category: Maintenance

#### Exit criteria:
- [x] BWC tests should no longer fail with the above error.

### Issues Resolved
- Resolves: https://github.com/opensearch-project/security/issues/3174

### Testing
Manual testing

### Check List


DarshitChanpura:
Looking into plugin-install failures now. Seems like we need to generate a new JKS file for sanity tests since the certificates were updated.

src/main/java/org/opensearch/security/tools/ChecksumCalculator.java
```diff
@@ -0,0 +1,71 @@
+/*
```
peternied, line 1: Why do we need a tool for this? There are many command line tools available for this kind of purpose, e.g. https://stackoverflow.com/questions/3358420/generating-a-sha-256-hash-from-the-linux-command-line

DarshitChanpura: Thought it'd be of value to add capability to generate shas of all files in a folder with a given extension. Otherwise they'd have to cat each file manually and run sha256sum. i.e. `cd <cert-folder>;  cat cert.pem | sha256sum`

But the main purpose was to seek feedback on PR whether we should ship with this or not. I felt it might be useful down the line so I added it.

peternied (changes requested):
We should not ship ChecksumCalcuator as part of the plugin distribution, its purpose is to automate making changes in the java code, not something a consumer of the plugin would need.

src/main/java/org/opensearch/security/tools/ChecksumCalculator.java
peternied: If instead this tool that generated the certs and then updated all of the relevant locations (including creating the checksums) - I would be onboard.

DarshitChanpura:
> We should not ship ChecksumCalcuator as part of the plugin distribution, its purpose is to automate making changes in the java code, not something a consumer of the plugin would need.

Agreed. I have added a section in Developer guide on steps to refresh demo certs.

cwperks:
@DarshitChanpura Is there any way to 'diff' the certs?

i.e. I'd like to verify that the only differences are:

1. Issue date
2. Expiry date
3. the SAN in the updated certs is mostly the same, but in the new certs it contains the IPv6 loopback address (::1)

peternied (approved):

peternied:
@cwperks What do you think is missing?  By virtue of BWC tests passing, and Plugin Install passing - I am assuming they work.

cwperks:
@peternied Just trying to cover all the bases. This PR looks good to me.

cwperks (approved):

xfournet:
Unfortunately this has generated an already expired root CA certificate : valid from 29/08/2023 to 28/09/2023

PR #3305, opened 2023-09-05
Commands GET `cat/aliases` and POST `_rollover` will work on specified indices

### LINKED ISSUE #1861: [BUG] Certain index APIs fail with permission denied when user role is restricted based on index pattern
**What is the bug?**
Certain index APIs fail with permission denied when user role is restricted based on index pattern

1. Create an internal role like so
`
curl -XPUT https://localhost:9200/_plugins/_security/api/roles/sem-role -H 'Content-Type: application/json' -d' { "cluster_permissions": [ "cluster_monitor" ], "index_permissions": [{ "index_patterns": [ "sem*" ], "allowed_actions" : ["*"] }] }
`
2. Create a user with this role
`
curl -XPUT https://localhost:9200/_plugins/_security/api/internalusers/sem-user -H 'Content-Type: application/json' -d' { "password": "######", "opendistro_security_roles": ["sem-role"] } '
`
3. Following APIs fail 
- Get aliases/alias, e.g. `curl https://localhost:9200/_cat/aliases/sem*?pretty -u sem-user` fails with _no permissions for [indices:admin/aliases/get]_
- Index rollover, e.g. `# curl -u sem-user -X POST "https://localhost:9200/sem-source1/_rollover?pretty"` fails with _no permissions for [indices:admin/rollover]_ 

These APIs work fine when the index pattern on the role is changed to "*" instead of "sem*"

- Get aliases/alias - should fetch aliases for index patterns the user has access to instead of throwing error
- Rollover should work fine given the user has permission to access the source index and create the target index based on the allowed index patterns for the user.

OpenSearch 1.2.4 with opensearch-security plugin enabled on OEL7

NA

- https://forum.opensearch.org/t/unable-to-set-permissions-for-put-index-template/4863/12
- https://forum.opensearch.org/t/internal-user-not-able-to-perform-operations-based-on-internal-role/9529 and https://forum.opensearch.org/t/internal-user-not-able-to-perform-operations-based-on-internal-role/9529/8

derek-ho:
### Description
Updates Get Aliases Request and Rollover to correctly look at the indices specified for these requests. This means that they will now work when a user has access to some specified set of indices, whereas before they would fail unless a user had permissions on all indices.

### Issues Resolved
#1861 

### Testing

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
+++ b/src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
@@ -116,7 +116,7 @@ public class DoNotFailOnForbiddenTests {
                 "indices:data/read/scroll",
                 "indices:monitor/settings/get",
                 "indices:monitor/stats",
-                    "indices:admin/aliases/get"
+                "indices:admin/aliases/get"
             )
             .on(MARVELOUS_SONGS)
     );
@@ -435,7 +435,7 @@ public class DoNotFailOnForbiddenTests {
             // High level client doesn't support _cat/_indices API
             Response getAliasesResponse = restHighLevelClient.getLowLevelClient().performRequest(getAliasesRequest);
             List<String> aliases = new BufferedReader(new InputStreamReader(getAliasesResponse.getEntity().getContent())).lines()
-                    .collect(Collectors.toList());
+                .collect(Collectors.toList());
 
             // Does not fail on forbidden, but alias response only contains index which user has access to
             assertThat(getAliasesResponse.getStatusLine().getStatusCode(), equalTo(200));
--- a/src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
+++ b/src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
@@ -115,7 +115,8 @@ public class DoNotFailOnForbiddenTests {
                 "indices:data/read/msearch",
                 "indices:data/read/scroll",
                 "indices:monitor/settings/get",
-                "indices:monitor/stats"
+                "indices:monitor/stats",
+                    "indices:admin/aliases/get"
             )
             .on(MARVELOUS_SONGS)
     );
@@ -424,6 +425,24 @@ public class DoNotFailOnForbiddenTests {
 
             assertThat(indexes.size(), equalTo(1));
             assertThat(indexes.get(0), containsString("marvelous_songs"));
+        }
+    }
+
+    @Test
+    public void shouldPerformCatAliases_positive() throws IOException {
+        try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(LIMITED_USER)) {
+            Request getAliasesRequest = new Request("GET", "/_cat/aliases");
+            // High level client doesn't support _cat/_indices API
+            Response getAliasesResponse = restHighLevelClient.getLowLevelClient().performRequest(getAliasesRequest);
+            List<String> aliases = new BufferedReader(new InputStreamReader(getAliasesResponse.getEntity().getContent())).lines()
+                    .collect(Collectors.toList());
+
+            // Does not fail on forbidden, but alias response only contains index which user has access to
+            assertThat(getAliasesResponse.getStatusLine().getStatusCode(), equalTo(200));
+            assertThat(aliases.size(), equalTo(1));
+            assertThat(aliases.get(0), containsString("marvelous_songs"));
+            assertThat(aliases.get(0), not(containsString("horrible_songs")));
+
         }
     }
 
--- a/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
@@ -108,7 +108,8 @@ public class PrivilegesEvaluator {
             "indices:admin/shards/search_shards",
             "indices:admin/resolve/index",
             "indices:monitor/settings/get",
-            "indices:monitor/stats"
+            "indices:monitor/stats",
+            "indices:admin/aliases/get"
         )
     );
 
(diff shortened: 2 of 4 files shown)


src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -108,7 +108,8 @@ public class PrivilegesEvaluator {
             "indices:admin/shards/search_shards",
             "indices:admin/resolve/index",
             "indices:monitor/settings/get",
-            "indices:monitor/stats"
+            "indices:monitor/stats",
+            "indices:admin/aliases/get"
```
derek-ho, line 112: It seems to me that the get aliases should also be a DNFOF like behavior - should I keep this in the PR and add a test for it?

src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
```diff
@@ -427,4 +428,22 @@ public void shouldPerformCatIndices_positive() throws IOException {
...
+            // High level client doesn't support _cat/_indices API
+            Response getAliasesResponse = restHighLevelClient.getLowLevelClient().performRequest(getAliasesRequest);
+            List<String> aliases = new BufferedReader(new InputStreamReader(getAliasesResponse.getEntity().getContent())).lines()
+                .collect(Collectors.toList());
+
+            // Does not fail on forbidden, but alias response only contains index which user has access to
+            assertThat(getAliasesResponse.getStatusLine().getStatusCode(), equalTo(200));
+            assertThat(aliases.size(), equalTo(1));
+            assertThat(aliases.get(0), containsString("marvelous_songs"));
+            assertThat(aliases.get(0), not(containsString("horrible_songs")));
```
derek-ho, line 444: @peternied @cwperks can you folks confirm this is the expected behavior (DNFOF returns alias even when user doesn't have index permissions on that alias when that alias contains an index that that user has access to (but the alias doesn't reveal any information about indices the user doesn't have access to))?

src/main/java/org/opensearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -750,11 +752,16 @@ private boolean getOrReplaceAllIndices(final Object request, final IndicesProvid
             }
             ((IndexRequest) request).index(newIndices.length != 1 ? null : newIndices[0]);
         } else if (request instanceof Replaceable) {
-            String[] newIndices = provider.provide(((Replaceable) request).indices(), request, true);
-            if (checkIndices(request, newIndices, false, allowEmptyIndices) == false) {
+            String[] newIndicesOrAliases;
+            if (request instanceof GetAliasesRequest) {
```
cwperks, line 756: I see that [GetAliasesRequest implements AliasesRequest](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/admin/indices/alias/get/GetAliasesRequest.java#L49). Can this use the more generic AliasesRequest? What other AliasesRequests are there and would they be impacted by the same issue?

derek-ho: Looks like the only other AliasesRequest I could find is within the IndicesAliasRequest: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/admin/indices/alias/IndicesAliasesRequest.java#L100. From the description it is: `A request to add/remove aliases for one or more indices.`. IMO we can change this to `instaceof AliasesRequest`, wdyt?

cwperks: `AliasesRequests` extends `IndicesRequest.Replaceable` which means an `AliasesRequest` is also an `IndicesRequest`. In that case would the `provider.provide` call need to consider both `.aliases()` and `.indices()` from the request object?

Is it possible that a GetAliasesRequest can contain `.aliases()` and `.indices()` at the same time?

src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
```diff
@@ -427,4 +428,22 @@ public void shouldPerformCatIndices_positive() throws IOException {
         }
     }
 
+    @Test
+    public void shouldPerformCatAliases_positive() throws IOException {
+        try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(LIMITED_USER)) {
+            Request getAliasesRequest = new Request("GET", "/_cat/aliases");
+            // High level client doesn't support _cat/_indices API
```
peternied, line 435: This comment isn't correct, let's remove it.

If you feel strongly that it is 'weird' this API isn't on the high-level client API, I'd recommend opening an issue.

src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
peternied: Index names themselves are considered protected by the access control systems, so an alias that exposes an index name would violate the access control checks.  Therefore; Users see aliases for indices I they have access to see, and the inverse if the user does not have access to an index they should not see any aliases associated with it.

src/main/java/org/opensearch/security/resolver/IndexResolverReplacer.java
DarshitChanpura: +1 . Will there be a case where indices() and aliases() are present at the same time. Which should be given higher priority in that case or should they both be considered ?

src/main/java/org/opensearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -781,6 +788,8 @@ private boolean getOrReplaceAllIndices(final Object request, final IndicesProvid
         } else if (request instanceof ReindexRequest) {
             result = getOrReplaceAllIndices(((ReindexRequest) request).getDestination(), provider, false) && result;
             result = getOrReplaceAllIndices(((ReindexRequest) request).getSearchRequest(), provider, false) && result;
+        } else if (request instanceof RolloverRequest) {
```
DarshitChanpura, line 791: should admins be allowed to rollover all indices with `*` permission or should these be index pattern specific? I'm not aware of any risks in allowing admin to rollover all indices, but would like to confirm.

peternied:
@derek-ho looks like this PR has stalled, would you like to keep working on it, or would you prefer to close it out?

peternied:
Oops meant to hit comment, not close.  I've reopened - let us know what you are thinking - thanks!

derek-ho:
@peternied @DarshitChanpura closing this until I get more bandwidth to see it through to completion. Will keep the branch in its current state for now

Issue #3309, opened 2023-09-05
Security Plugin cannot startup due to AccessControlException: access denied


peternied:
The security plugin cannot be loaded because of an issue with the bouncy castle dependancy coming from OpenSearch core.

```
 java.lang.IllegalStateException: failed to load plugin class [org.opensearch.security.OpenSearchSecurityPlugin]
�  Likely root cause: java.lang.InternalError: cannot create instance of org.bouncycastle.jcajce.provider.digest.GOST3411$Mappings : java.security.AccessControlException: access denied ("java.security.SecurityPermission" "putProviderProperty.BC")
�  	at org.bouncycastle.jce.provider.BouncyCastleProvider.loadServiceClass(Unknown Source)
�  	at org.bouncycastle.jce.provider.BouncyCastleProvider.loadAlgorithms(Unknown Source)
�  	at org.bouncycastle.jce.provider.BouncyCastleProvider.setup(Unknown Source)
�  	at org.bouncycastle.jce.provider.BouncyCastleProvider.access$000(Unknown Source)
�  	at org.bouncycastle.jce.provider.BouncyCastleProvider$1.run(Unknown Source)
�  	at java.base/java.security.AccessController.doPrivileged(Native Method)
�  	at org.bouncycastle.jce.provider.BouncyCastleProvider.<init>(Unknown Source)
�  	at org.opensearch.security.OpenSearchSecurityPlugin$2.run(OpenSearchSecurityPlugin.java:323)
...
```

### Additional Context
- Plugin install failure https://github.com/opensearch-project/security/actions/runs/6086574945/job/16513238110?pr=3307
- BWC test startup failure https://github.com/opensearch-project/security/actions/runs/6086574929/job/16513238774?pr=3307
- SecurityManager permissions issue with other plugins https://github.com/opensearch-project/security/issues/3213

peternied:
Root cause change in main [1] & backport to 2.x [2]

- [1] https://github.com/opensearch-project/OpenSearch/pull/9289/
- [2] https://github.com/opensearch-project/OpenSearch/pull/9736

peternied:
@cwperks Can you help drive this from the security plugin?

FYI - @reta @willyborankin

reta:
@cwperks let me know if you need any help

cwperks:
@reta all eyes are appreciated on this issue. This is release blocking for 2.10.0.

reta:
Sure, I am taking it, @cwperks!

[...]

PR #3317, opened 2023-09-05
Security Plugin cannot startup due to AccessControlException: access denied

reta:
### Description
Fixing AccessControlException after https://github.com/opensearch-project/OpenSearch/pull/9289, we have 2 options here:
 1. since the `bcprov` moved to core, it will be loaded by core, the plugin should not try to enforce it
 2. plugin may try to enforce it but the core security policy should be extended to allow that

```
grant codeBase "${codebase.bcprov-jdk15to18}" {
  permission java.security.SecurityPermission "putProviderProperty.BC";
  permission java.security.SecurityPermission "insertProvider.BC";
  permission java.security.SecurityPermission "removeProviderProperty.BC";
};
```

### Issues Resolved
Closes https://github.com/opensearch-project/security/issues/3309

### Testing
Covered by existing tests

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -35,7 +35,6 @@ import java.nio.file.attribute.PosixFilePermission;
 import java.security.AccessController;
 import java.security.MessageDigest;
 import java.security.PrivilegedAction;
-import java.security.Security;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -60,7 +59,6 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.lucene.search.QueryCachingPolicy;
 import org.apache.lucene.search.Weight;
-import org.bouncycastle.jce.provider.BouncyCastleProvider;
 
 import org.opensearch.OpenSearchException;
 import org.opensearch.OpenSearchSecurityException;
@@ -330,16 +328,6 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
         if (sm != null) {
             sm.checkPermission(new SpecialPermission());
         }
-
-        AccessController.doPrivileged(new PrivilegedAction<Object>() {
-            @Override
-            public Object run() {
-                if (Security.getProvider("BC") == null) {
-                    Security.addProvider(new BouncyCastleProvider());
-                }
-                return null;
-            }
-        });
 
         final String advancedModulesEnabledKey = ConfigConstants.SECURITY_ADVANCED_MODULES_ENABLED;
         if (settings.hasValue(advancedModulesEnabledKey)) {
--- a/plugin-security.policy
+++ b/plugin-security.policy
@@ -56,11 +56,6 @@ grant {
   permission java.net.NetPermission "getProxySelector";
   permission java.net.SocketPermission "*", "connect,accept,resolve";
 
-  // BouncyCastle permissions
-  permission java.security.SecurityPermission "putProviderProperty.BC";
-  permission java.security.SecurityPermission "insertProvider.BC";
-  permission java.security.SecurityPermission "removeProviderProperty.BC";
-
   permission java.lang.RuntimePermission "accessUserInformation";
   
   permission java.security.SecurityPermission "org.apache.xml.security.register";


willyborankin:
@reta we had more settings:
```
  permission java.security.SecurityPermission "putProviderProperty.BC";
  permission java.security.SecurityPermission "insertProvider.BC";
  permission java.security.SecurityPermission "removeProviderProperty.BC";
  permission java.util.PropertyPermission "*", "write";

```
add @cwperks aded this fix: https://github.com/opensearch-project/security/pull/3289

reta:
> permission java.util.PropertyPermission "*", "write";

It seems like this one have to go to core :( I am still looking hence this pull request is in draft state

willyborankin:
> > permission java.util.PropertyPermission "*", "write";
> 
> It seems like this one have to go to core :( I am still looking hence this pull request is in draft state
ehhh  "read,write" :-(

willyborankin:
@peternied, @cwperks and @reta I think this solution is ok for 2.10 but is not good in the future.

willyborankin:
@reta im wrong we have such login in the policy.
```
grant codeBase "${codebase.netty-common}" {
  permission java.lang.RuntimePermission "loadLibrary.*";
};

```

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -331,16 +331,6 @@ public OpenSearchSecurityPlugin(final Settings settings, final Path configPath)
             sm.checkPermission(new SpecialPermission());
         }
 
-        AccessController.doPrivileged(new PrivilegedAction<Object>() {
```
reta, line 334: @peternied @cwperks any reasons we need to explicitly add `BouncyCastleProvider`?

willyborankin: Ahhh it is historical. It was added this way.

reta:
> @reta im wrong we have such login in the policy.
> 
> ```
> grant codeBase "${codebase.netty-common}" {
>   permission java.lang.RuntimePermission "loadLibrary.*";
> };
> ```

@willyborankin the ${codebase} placeholder is resolved relatively to plugin: `/..../plugins/opensearch-security`, there is no way to point to $OPENSEARCH/lib from there

willyborankin:
> > @reta im wrong we have such login in the policy.
> > ```
> > grant codeBase "${codebase.netty-common}" {
> >   permission java.lang.RuntimePermission "loadLibrary.*";
> > };
> > ```
> 
> @willyborankin the ${codebase} placeholder is resolved relatively to plugin: `/..../plugins/opensearch-security`, there is no way to point to $OPENSEARCH/lib from there

got it

cwperks:
> add @cwperks aded this fix: https://github.com/opensearch-project/security/pull/3289. So all of them need to be added to the the SDK now? If yes this is a partial fix :-(

@willyborankin you are right that the permissions added in that PR would also need to be added to core since the dependency is now coming from core. I have a PR in core to add the existing bouncy castle permissions that the security plugin has into core https://github.com/opensearch-project/OpenSearch/pull/9770 (Option #2 from this PR's description)

I am trying to figure out now the reasons why the security plugin calls on `Security.addProvider`, but from my own understanding bouncycastle is used to setup TLS (reading certs and keys) and is used for its implementation of BCrypt

reta:
> I am trying to figure out now the reasons why the security plugin calls on `Security.addProvider`, but from my own understanding bouncycastle is used to setup TLS (reading certs and keys) and is used for its implementation of BCrypt

We definitely need to add the BC provider, but since it is in core now, it should be done by core, that would also solve any issues with other plugins trying to add/remove BC.

cwperks:
@reta From the javadoc for `java.security.Security` it is safe to call `addProvider` multiple times: https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/security/Security.html#addProvider(java.security.Provider)

> the preference position in which the provider was added, or -1 if the provider was not added because it is already installed.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
cwperks: From the javadoc for `java.security.Security` it is safe to call `addProvider` multiple times: https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/security/Security.html#addProvider(java.security.Provider)

> the preference position in which the provider was added, or -1 if the provider was not added because it is already installed.

reta:
> @reta From the javadoc for `java.security.Security` it is safe to call `addProvider` multiple times: https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/security/Security.html#addProvider(java.security.Provider)

The problem is not how many times, the problem is who is adds it because that would impact the security policy to be taken into account.

willyborankin:
> > I am trying to figure out now the reasons why the security plugin calls on `Security.addProvider`, but from my own understanding bouncycastle is used to setup TLS (reading certs and keys) and is used for its implementation of BCrypt
> 
> We definitely need to add the BC provider, but since it is in core now, it should be done by core, that would also solve any issues with other plugins trying to add/remove BC.

If you set any securoty plugin this way it means that in the list of sec providers JDK knows about it is in the first place.So you do not need to lest say create a cipher this way: 
```java
Cipher cipher = Cipher.getInstance("DESede/CBC/NoPadding");
```
otherwise you need add a provider name:
```java
Cipher cipher = Cipher.getInstance("DESede/CBC/NoPadding", BouncyCastleProvider.PROVIDER_NAME)
```
We an change it but it is not so fast as you could think.

reta:
Closing in favor  of https://github.com/opensearch-project/OpenSearch/pull/9779

Issue #3204, opened 2023-08-18
[BUG] TransportConfigUpdateAction causing a race condition during node bootstrap leading to 403/500 errors


shikharj05:
**What is the bug?**
Partial configuration is loaded/cached in security plugin objects/memory in case of a race condition during OpenSearch process bootstrap on a node. For e.g. this can lead to missing role mappings in-memory causing invalid 403s to the requests landing on the buggy node. 
Details:
During a node bootstrap, [initOnNodeStart](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java#L309) is kicked-off after Rest layer is ready. However, a node joins the cluster and replica shard for security index is assigned after transport channel is ready and cluster manager is discovered. Hence, if a [TransportConfigUpdateAction](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java#L127) triggers an update for a particular config (let's say securityconfig), DynamicConfigFactory object gets initialized and [this](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java#L218) condition is never true leading to partial config in memory.

**How can one reproduce the bug?**
1. Create a multi-node cluster. From a healthy node, execute something like below in a continuous loop
``` 
curl -XPATCH "http://localhost:9200/_plugins/_security/api/securityconfig" -H 'Content-Type: application/json' -d'
[{
    "op": "replace", "path": "/config/dynamic/authc/basic_internal_auth_domain/transport_enabled", "value": "true"
  }
]'
```
2. Restart OS process on a different node, (you can enable debug logs on [ConfigurationLoaderSecurity7](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/ConfigurationLoaderSecurity7.java#L125) to check the config being pushed.
3. Since step 1 is running in a continuous loop, chances are transport channel action initializes the object with [Config Type](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/impl/CType.java#L55) only and hence, other configs will be skipped.

A temporary fix here is to either flush the cache using ``flush`` API OR send a dummy config update request.

**What is the expected behavior?**
Do not partially initialize objects. 

**What is your host/environment?**
 - OS: Linux
 - Version - All
 - Plugins - Security plugin

**Do you have any screenshots?**
[...]

PR #3356, opened 2023-09-13
Replace DCF isInitialized with a countdown latch and waitForInit

DarshitChanpura:
### Description
- Resolves https://github.com/opensearch-project/security/issues/3204

### Testing

### Check List


DarshitChanpura:
@shikharj05 @anijain-Amazon Would you please review this?

src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java
```diff
@@ -73,8 +73,9 @@
 import org.opensearch.security.support.WildcardMatcher;
 import org.opensearch.threadpool.ThreadPool;
 
-public class DynamicConfigFactory implements Initializable, ConfigurationChangeListener {
+public class DynamicConfigFactory implements ConfigurationChangeListener {
 
+    private final CountDownLatch initializationLatch = new CountDownLatch(1);
```
peternied, line 78: I don't think this latch is working correctly. There are multiple countDowns and the latch is configured to be triggered after a single countDown?

peternied (commented):
Lets translate the repro from the issue into a test case that we can use to confirm the bug and evaulate this fix.

src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java
DarshitChanpura: agreed, this latch countdown should be removed from onChange. Which is what i intended to get out of discussion on this PR. I have already proposed this as a fix on the original issue: https://github.com/opensearch-project/security/issues/3204#issuecomment-1717993502

DarshitChanpura:
>Lets translate the repro from the issue into a test case that we can use to confirm the bug and evaulate this fix.

Testing race conditions is always a little tricky. I'm currently working on coming up with tests that can reliably reproduce the race conditions to confirm the fix.

DarshitChanpura:
Closing this for now. Will re-open once I resume working on this upon feedback from @shikharj05 .

PR #3387, opened 2023-09-21
Re-enables gradle build cache

### LINKED ISSUE #3185: Reenable gradle build cache
The gradle build cache is used between github actions to speed up build times, but if the wrong cache sources are being used this can result in unexpected failures that are hard to troubleshoot.

We are disabling the remote gradle cache to while stabilizing the build infrastructure.  We should revisit what the impact of not having this cache.

Task 1: Determine the best path forward for re-enabling the build cache. Should we re-instate it the same way or in a different way? 
Task 2: Re-enable cache in best way.

DarshitChanpura:
### Description
Uses https://github.com/marketplace/actions/gradle-cache to replace `gradle-build-action`.

### Issues Resolved
- Resolves #3185 

### Testing
Passing CI

### Check List

### DIFF
--- a/.github/workflows/plugin_install.yml
+++ b/.github/workflows/plugin_install.yml
@@ -26,9 +26,8 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Assemble target plugin
-        uses: gradle/gradle-build-action@v2
+        uses: burrunan/gradle-cache-action@v1
         with:
-          cache-disabled: true
           arguments: assemble
 
       # Move and rename the plugin for installation
@@ -59,7 +58,6 @@ jobs:
           setup-script-name: setup
 
       - name: Run sanity tests
-        uses: gradle/gradle-build-action@v2
+        uses: burrunan/gradle-cache-action@v1
         with:
-          cache-disabled: true
           arguments: integTestRemote -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="opensearch" -Dhttps=true -Duser=admin -Dpassword=admin
--- a/.github/actions/create-bwc-build/action.yaml
+++ b/.github/actions/create-bwc-build/action.yaml
@@ -33,9 +33,8 @@ runs:
         path: ${{ inputs.plugin-branch }}
 
     - name: Build
-      uses: gradle/gradle-build-action@v2
+      uses: burrunan/gradle-cache-action@v1
       with:
-        cache-disabled: true
         arguments: assemble
         build-root-directory: ${{ inputs.plugin-branch }}
 
--- a/.github/actions/run-bwc-suite/action.yaml
+++ b/.github/actions/run-bwc-suite/action.yaml
@@ -37,9 +37,8 @@ runs:
         plugin-branch: ${{ inputs.plugin-next-branch }}
 
     - name: Run BWC tests
-      uses: gradle/gradle-build-action@v2
+      uses: burrunan/gradle-cache-action@v1
       with:
-        cache-disabled: true
         arguments: |
           bwcTestSuite
           -Dtests.security.manager=false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,9 +53,8 @@ jobs:
       uses: actions/checkout@v4
 
     - name: Build and Test
-      uses: gradle/gradle-build-action@v2
+      uses: burrunan/gradle-cache-action@v1
       with:
-        cache-disabled: true
         arguments: |
           ${{ matrix.gradle_task }} -Dbuild.snapshot=false
 
@@ -101,12 +100,10 @@ jobs:
       uses: actions/checkout@v4
 
     - name: Build and Test
-      uses: gradle/gradle-build-action@v2
+      uses: burrunan/gradle-cache-action@v1
       continue-on-error: true # Until retries are enable do not fail the workflow https://github.com/opensearch-project/security/issues/2184
       with:
-        cache-disabled: true
-        arguments: |
-          integrationTest -Dbuild.snapshot=false
+        arguments: integrationTest -Dbuild.snapshot=false
 
   backward-compatibility:
     strategy:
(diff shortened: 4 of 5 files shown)


peternied:
Thanks for looking into this, could you explain why you think this will work better/different than removing the flag on the exiting gradle-build-action?

cwperks:
What are the tradeoffs of cache enabled or disabled?

I was having difficulty when the cache was enabled bc there were unpredictable builds. Any time caches needed to be removed, there was no bulk way of doing it. The caches could only be removed 1-by-1 and there are caches for each GHA check which in this repo is a lot since the tests are split into parallel test groups. IMO I haven't seen degradation of the GHA runtimes after the cache was disabled so I'd be in favor of having it remain disabled unless there are good reasons to re-enable it.

DarshitChanpura:
> Thanks for looking into this, could you explain why you think this will work better/different than removing the flag on the exiting gradle-build-action?

Based on the description on the marketplace, the cache action uses a "small" index and a proxy server which redirects requests to @actions/cache API and the cache receives only the relevant updates. The increased granularity enables GitHub to evict entries better (it removes unused entries automatically).
Another advantage: `gradle-cache-action` creates a layered cache, and it uses a small "index" cache to identify the required layers. If only a small fraction of files changes, then the action reuses the existing caches, and it adds a layer on top of it. That enables to save cache space (GitHub has a default limit of 5 GiB), and it reduces upload time as only the cache receives only the updated files.

> What are the tradeoffs of cache enabled or disabled?

Enabled gradle build cache leads to: Faster builds . Reduced network usage. Cache can be shared between CI agents when remote cache is enabled.

peternied:
@DarshitChanpura I think I am missing something, what about `gradle-build-action` that was incorrect that is fixed with `gradle-cache-action`?

To Craig's concern, any caching system would be faster, but it wasn't accurate; how do we know this will be more accurate?

DarshitChanpura:
AFAIU, gradle-cache-action is faster and saves memory by creating a layered cache, and it uses a small "index" cache to identify the required layers. If only a small fraction of files changes, then the action reuses the existing caches, and it adds a layer on top of it.
Here is how it is better than gradle-build-action in their own words:
```
🚀 Gradle remote build cache backend (pulls only the needed entries from GitHub cache)
🎉 Support multiple remote caches via [gradle-multi-cache](https://github.com/burrunan/gradle-multi-cache) (e.g. GitHub Actions + S3)
👋 Simplified configuration (action name + gradle command is enough for most case)
👾 Less space usage (GitHub imposes overall 5GiB limit by default, so cache space matters)
🔗 Link to Build Scan in build results
💡 Gradle build failure markers added to the diff view (e.g. compileJava or compileKotlin markers right in the commit diff)
```

I understand Craig's concern, however without trying this solution we wouldn't know for sure whether cache would work for our CI. If it doesn't fit our CI, we can always revert this PR and permanently disable the cache. Wdyt?

.github/workflows/plugin_install.yml
```diff
@@ -63,7 +62,6 @@ jobs:
           admin-password: ${{ steps.random-password.outputs.generated_name }}
 
       - name: Run sanity tests
-        uses: gradle/gradle-build-action@v2
+        uses: burrunan/gradle-cache-action@v1
         with:
-          cache-disabled: true
-          arguments: integTestRemote -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="opensearch" -Dhttps=true -Duser=admin -Dpassword=${{ steps.random-password.outputs.generated_name }} -i
+          arguments: integTestRemote -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="opensearch" -Dhttps=true -Duser=admin -Dpassword=${{ steps.random-password.outputs.generated_name }}
```
DarshitChanpura, line 67: Was the `-i` intentional here? My understanding is that it was not, and hence i removed it to prevent from printing extra logs, but if it was then I can revert this.

RyanL1997: Will the log still be shown after the removal of hte `-i`?

DarshitChanpura: `-i` enables printing info logs. Unless we absolutely need them, these will just use up extra network and memory for each run, and so there is an opportunity to save that by not adding -i

peternied (changes requested):
> what about gradle-build-action that was incorrect that is fixed with gradle-cache-action?

Without data to justify this change I don't think we should move forward.

DarshitChanpura:
> > what about gradle-build-action that was incorrect that is fixed with gradle-cache-action?
> 
> Without data to justify this change I don't think we should move forward.

Would you suggest some steps in gathering the data to verify whether there was an improvement?

peternied:
I would first recommending asking yourself what you would expect reviewers of this PR to weight in choosing to accept it or not, its very rare that changes are accepted without some form of validation, no?

If you are trying to find an approach I might suggest, starting by running a number of builds with different conditions in your fork, then you can use those results from the GitHub action to make claims with justify of it being faster, or uses less resources.

As Craig mentioned accuracy I would suggest seeing if you can test how the cache invalidation happens, and force different conditions.

These experiments might be with the full build or a streamlined part that is easier to measure/control.  Please let us know the path you are taking and the results you generate so we can weight our options.

DarshitChanpura:
> I would first recommending asking yourself what you would expect reviewers of this PR to weight in choosing to accept it or not, its very rare that changes are accepted without some form of validation, no?
> 
> If you are trying to find an approach I might suggest, starting by running a number of builds with different conditions in your fork, then you can use those results from the GitHub action to make claims with justify of it being faster, or uses less resources.
> 
> As Craig mentioned accuracy I would suggest seeing if you can test how the cache invalidation happens, and force different conditions.
> 
> These experiments might be with the full build or a streamlined part that is easier to measure/control. Please let us know the path you are taking and the results you generate so we can weight our options.

makes sense... I will update this thread with my results once I have them.

peternied:
Lets close this out since there hasn't been progress in 2 weeks - please reopen if you pick this back up

Issue #2944, opened 2023-07-05
[META] Service Accounts


scrawfor99:
This tracking issue describes a feature where applications (plugins/extensions) make use of service accounts when completing actions outside the context of an impersonated user. 

Service Accounts will officially become a context where there is enforcement of permissions and it is possible to determine the context an application is operating under. 

### Design
- [x] https://github.com/opensearch-project/security/issues/2596
- [x] #3200

### Implementation
- [x] #2939
- [x] https://github.com/opensearch-project/security/issues/2704

### Integration
- [x] https://github.com/opensearch-project/security/issues/2611
- [x] https://github.com/opensearch-project/security/issues/2646
- [x] #2942
  - [x] https://github.com/opensearch-project/security/issues/3176
  - [x] https://github.com/opensearch-project/OpenSearch/pull/8526

### Release criteria
- [x] #3201
- [x] #3557 
- [x] https://github.com/opensearch-project/security/issues/3199

### Follow up
- [x] #2941 
- [x] #2943

scrawfor99:
[Triage] This is a tracking issue for work being done in core around Identity.

PR #3416, opened 2023-09-27
[2.x] Adds On-Behalf-Of authentication mechanism and service account capability

DarshitChanpura:
### Description
Backport On-behalf-of Authentication and Service account into the 2.x branch.

Category: New Feature

### Issues Resolved
* Related to #2944 
* Related to https://github.com/opensearch-project/security/issues/2573
* Pre-requisite PR: https://github.com/opensearch-project/OpenSearch/pull/10258

### Testing
Integration testing

### Check List


RyanL1997:
The manual testing with the build of https://github.com/opensearch-project/OpenSearch/pull/10258 succeed:
```
➜  ~ curl -XPOST "https://localhost:9200/_plugins/_security/api/generateonbehalfoftoken" --insecure -u 'admin:admin' --data '
{
  "reason":"Testing",
  "service":"extension123",
  "durationSeconds":"600"
}' -H 'Content-Type: application/json' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   462  100   382  100    80    758    158 --:--:-- --:--:-- --:--:--   920
{
  "user": "admin",
  "authenticationToken": "eyJhbGciOiJIUzUxMiJ9.eyJpc3MiOiJvcGVuc2VhcmNoIiwiaWF0IjoxNjk1OTIzNzU1LCJzdWIiOiJhZG1pbiIsImF1ZCI6ImV4dGVuc2lvbjEyMyIsIm5iZiI6MTY5NTkyMzc1NSwiZXhwIjoxNjk1OTI0MzU1LCJlciI6IjczU1JBZjlueStqR3AxczFLSjk4d1VOeXJDckY0TWcydW1pSU4vN0c5YkE9In0.op-l1zhpwz93zvYYzLEuGXFLH5ZGf_5wvMbd4SNtAtWClTmDrP7tepfC7X5yxuymnirZwHtYNk0eams_NQOJew",
  "durationSeconds": 600
}

**Comment by scrawfor99** [comment]
I would recommend splitting this into two PRs: one that does OBO changes and one that does service account changes. 

I am not as familiar with the OBO changes since I did not write them originally but here is what is required for service accounts: 

1. Implement `IdentityPlugin` in `OpenSearchSecurityPlugin`
2. Create an `SecurityPluginTokenManager` or something similar that implements the `TokenManager` interface.
    This should be able to issue a service account token and an on behalf of token 
3. Connect the `IdentityPlugin` method to the Token Manager so that when a new extension is registered and the transport action registering its security settings is called, we also issue it a new service account token. 
4. Inside the token manager, the service account issuance process should register a new internal user account for the extension if one does not already exist. 
5. When an extension attempts to use a service account token it will be using it as the authentication header for its request so all the changes relating to using it should go through the normal auth process of an internal user account. 
    The only differences are that a `enabled` equals `false` service account should fail to perform _any_ requests. 
    As part of the threat model mitigation we also need to block any actions which are not. being explicitly performed on the system index associated with the service account being used. This requires issuance of system indices etc. for extensions which is not currently being done. I think we can wait on this but I would ask @peternied and @cwperks. Since they should know the decision there. 
7. Test service accounts are issued properly by using the fake rest handler with a service account token generated in a mocked scenario. We don't need to test the transport deliverance inside the security plugin because that is outside of the scope.

File: src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java [hunk 1]
```diff
@@ -1110,6 +1144,15 @@ public Settings additionalSettings() {
         return builder.build();
     }
 
+    @Override
```
scrawfor99, line 1147: So the way Craig set this up, the idea is to register a set of extension security settings when an extension is installed/registered as part of the core initialization process. To make this work with service accounts the general idea would be to make sure that a service account is only able to operate on the indices listed in the reserved_indices list and affiliated with themselves--this should use a mapping or some other mechanism.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1110,6 +1144,15 @@ public Settings additionalSettings() {
...
+    public List<Setting<?>> getExtensionSettings() {
+        List<Setting<?>> settings = new ArrayList<Setting<?>>();
+        settings.add(RESERVED_INDICES_SETTING);
+        settings.add(SEND_BACKEND_ROLES_SETTING);
```
scrawfor99, line 1151: I don't think extensions should have backend roles so this can be removed.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -0,0 +1,179 @@
...
+    private final ClusterService clusterService;
+
+    private ConfigModel configModel;
+
+    private DynamicConfigModel dcm;
+
+    public static final Integer OBO_DEFAULT_EXPIRY_SECONDS = 5 * 60;
+    public static final Integer OBO_MAX_EXPIRY_SECONDS = 10 * 60;
+
+    public static final String DEFAULT_SERVICE = "self-issued";
```
scrawfor99, line 62: I recommend renaming this since `DEFAULT_SERVICE` does not tell you what the actual purpose of the constant is.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -0,0 +1,179 @@
...
+            @Override
+            public void accept(RestChannel channel) throws Exception {
+                final XContentBuilder builder = channel.newBuilder();
+                BytesRestResponse response;
+                try {
+                    if (vendor == null) {
+                        channel.sendResponse(
+                            new BytesRestResponse(
+                                RestStatus.SERVICE_UNAVAILABLE,
+                                "The OnBehalfOf token generating API has been disabled, see {link to doc} for more information on this feature." /* TODO: Update the link to the documentation website */
```
scrawfor99, line 122: I think @RyanL1997 is doing this?

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -0,0 +1,179 @@
...
+                    final Map<String, Object> requestBody = request.contentOrSourceParamParser().map();
+                    final String description = (String) requestBody.getOrDefault("description", null);
+
+                    final Integer tokenDuration = Optional.ofNullable(requestBody.get("durationSeconds"))
+                        .map(value -> (String) value)
+                        .map(Integer::parseInt)
+                        .map(value -> Math.min(value, OBO_MAX_EXPIRY_SECONDS)) // Max duration seconds are 600
+                        .orElse(OBO_DEFAULT_EXPIRY_SECONDS); // Fallback to default
+
+                    final Boolean roleSecurityMode = Optional.ofNullable(requestBody.get("roleSecurityMode"))
```
scrawfor99, line 139: I would recommend renaming this boolean as the name is not super informative imo.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -0,0 +1,179 @@
...
+                        .orElse(true); // Default to false if null
+
+                    final String service = (String) requestBody.getOrDefault("service", DEFAULT_SERVICE);
+                    final User user = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+                    Set<String> mappedRoles = mapRoles(user);
+
+                    builder.startObject();
+                    builder.field("user", user.getName());
+
+                    final String token = vendor.createJwt(
```
scrawfor99, line 150: Nit: Standardize terminology across the framework--core expects a token manager to issue an on behalf of token or service account token. I would suggest avoiding referring to the obo token provider as a vendor while we refer to it as a token manager elsewhere--this can be done via renaming or raising the reference up a level to be pointing to the issuance call in the manager.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -605,7 +606,12 @@ private User impersonate(final RestRequest request, final User originalUser) thr
             final boolean isDebugEnabled = log.isDebugEnabled();
             // loop over all http/rest auth domains
             for (final AuthDomain authDomain : restAuthDomains) {
+
                 final AuthenticationBackend authenticationBackend = authDomain.getBackend();
+                // Skip over the OnBehalfOfAuthenticator since it is not compatible for user impersonation
+                if (authDomain.getHttpAuthenticator() instanceof OnBehalfOfAuthenticator) {
```
scrawfor99, line 612: I don't follow this comment--I would clarify with Ryan or Craig what was meant here.

src/main/java/org/opensearch/security/authtoken/jwt/EncryptionDecryptionUtil.java
```diff
@@ -0,0 +1,75 @@
...
+        DECRYPT(Cipher.DECRYPT_MODE);
+
+        private final int opmode;
+
+        private CipherMode(final int opmode) {
+            this.opmode = opmode;
+        }
+    }
+
+    // TODO: RYAN PUT YOUR LOGIC HERE TO CONNECT WITH
```
scrawfor99, line 73: Looks like this is still missing code? @RyanL1997

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -0,0 +1,171 @@
...
+        expirySeconds = (expirySeconds == null) ? defaultExpirySeconds : Math.min(expirySeconds, maxExpirySeconds);
+        if (expirySeconds <= 0) {
+            throw new Exception("The expiration time should be a positive integer");
+        }
+        long expiryTime = timeProvider.getAsLong() + expirySeconds;
+        jwtClaims.setExpiryTime(expiryTime);
+
+        if (roles != null) {
+            String listOfRoles = String.join(",", roles);
+            jwtClaims.setProperty("er", encryptionDecryptionUtil.encrypt(listOfRoles));
```
scrawfor99, line 146: Nit: I don't think `er` and `br` are super clear. I know `br` is backend roles but still don't know what `er` means. Can we use full term or more complete abbreviation? I don't think the small size difference should be an issue

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,108 @@
...
+        .put(
+            "signing_key",
+            Base64.getEncoder()
+                .encodeToString(
+                    "This is my super secret that no one in the universe will ever be able to guess in a bajillion years".getBytes(
+                        StandardCharsets.UTF_8
+                    )
+                )
+        )
+        .put("encryption_key", Base64.getEncoder().encodeToString("encryptionKey".getBytes(StandardCharsets.UTF_8)))
```
scrawfor99, line 39: We will want to make all of these UrlEncoders

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -199,6 +213,24 @@ private List<String> getAllProtectedSystemIndices(final Resolved requestedResolv
...
     }
 
+    /**
+     * Checks whether the request contains any of the indices reserved by extension
+     * @param requestedResolved request to be checked for
+     * @param reservedIndices the list of indices reserved by extension
+     * @return true if it contains reserved index or if system indices are disabled,
+     * false if request doesn't contain any reserved index
+     */
+    private boolean matchAllReservedIndices(final Resolved requestedResolved, final Set<String> reservedIndices) {
```
scrawfor99, line 223: This is meant for the service account logic I believe. I could be mistaken but if this is for the service account logic, we can probably make this more straightforward--just check whether the index being targeted with a request is the one which matches the system index associated with the extension.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
scrawfor99, line 1: We can remove these changes around the ephemeral account stuff. This is not required for service accounts and I don't believe Obo tokens use this logic either.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
```diff
@@ -563,6 +600,53 @@ private boolean containsDlsFlsConfig() {
         }
     }
 
+    public static class OnBehalfOfSettings {
```
scrawfor99, line 603: Should keep this part.

src/main/java/org/opensearch/security/securityconf/Role.java
```diff
@@ -0,0 +1,13 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+package org.opensearch.security.securityconf;
+
+public interface Role {}
```
scrawfor99, line 13: Not sure whether this is required--I think we can avoid using this.

src/main/java/org/opensearch/security/securityconf/SecurityRoles.java
```diff
@@ -55,6 +55,8 @@ boolean hasExplicitIndexPermission(
 
     Set<String> getRoleNames();
 
+    void addRole(Role role);
```
scrawfor99, line 58: This can be removed assuming `Role.java` is dropped.

src/main/java/org/opensearch/security/securityconf/impl/v7/RoleV7.java
```diff
@@ -38,10 +38,11 @@
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 import org.opensearch.security.securityconf.Hideable;
+import org.opensearch.security.securityconf.Role;
 import org.opensearch.security.securityconf.StaticDefinable;
 import org.opensearch.security.securityconf.impl.v6.RoleV6;
 
-public class RoleV7 implements Hideable, StaticDefinable {
+public class RoleV7 implements Hideable, StaticDefinable, Role {
```
scrawfor99, line 45: See above--can be removed.

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -134,7 +134,7 @@ public SecurityDynamicConfiguration<?> createOrUpdateAccount(ObjectNode contentA
         if (!attributeNode.get("service").isNull() && attributeNode.get("service").asString().equalsIgnoreCase("true")) { // If this is a
                                                                                                                           // service account
             verifyServiceAccount(securityJsonNode, accountName);
-            String password = generatePassword();
```
scrawfor99, line 137: Swap this back to generate password. This was just for the pen test env.

RyanL1997:
Thanks @scrawfor99! Alright then, I will do the backport of OBO separately.

scrawfor99 (commented):
Left comments on the original changes to give some direction.

src/main/java/org/opensearch/security/securityconf/SecurityRoles.java
DarshitChanpura: This is required since getSecurityRoles method calls addRole method internally

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
DarshitChanpura: @RyanL1997 this is required correct?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: I have removed BACKEND_ROLES setting for now. In SecurityIndexAccessEvaluator we check by looking up extension settings by username (should match extension name), and then matching the requested index with extension's reserved index to determine whether access should be granted

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
RyanL1997: I'm fine to redirect that to the homepage of our documentation website, at least for now. So it will be https://opensearch.org/docs/latest/

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
RyanL1997: yes the er is encrypted roles. I think this is the original design we came up with in our main branch, so lets keep using it.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
RyanL1997: We should keep this for sure, but I just checked the main branch this should be located in ConfigV7.java. Should we also move that?

RyanL1997: I have refactored it.

src/main/java/org/opensearch/security/authtoken/jwt/EncryptionDecryptionUtil.java
RyanL1997: Nope. Thanks for double checking it tho. I have removed this comment and also added the tests for this class.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
RyanL1997: Yes, DC is correct. The obo auth backend should not be used for impersonate a user.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
RyanL1997: Or should we wait until the doc team publish the work so that we can use aa more accurate url?

DarshitChanpura:
@RyanL1997 @scrawfor99 Should this be closed as there is a separate effort on-going for this?

DarshitChanpura:
I'll be closing this out in lieu of ongoing efforts to merge Service accounts and OBO related changes separately. Please re-open if this is picked back up.

Issue #3267, opened 2023-08-29
Consider switching JWT library implementations


peternied:

scrawfor99:
[Triage] This issue seems to have clearly outlined closure criteria and is the result of guidance around how to improve our overall security posture. We will need this to be backwards compatible so that the a token made with the CXF implementation can then go into the Nimbus version.

Given the prototype Peter made, this seems like something we can mark good first issue.

MaciejMierzwa:
Started looking into it

MaciejMierzwa:
Hi, here's the branch and draft PR I'm pushing my changes to: https://github.com/opensearch-project/security/pull/3421
So far so good although there are more differences between library implementations then I hoped for. Also, I noticed that nimbus library doesn't accept shorter keys for signing JWT, seems like cxf was adding padding if secret key was too short, nimbus on the other hand forces user to use proper length secret (512 bits for HMAC512 etc)

MaciejMierzwa:
Hi, for visibility I'll write here instead of PR. 
There are few differences between how both libraries dealt with implementing specification which caused issues. 
1. For HMAC keys it's suggested that the key is at least as long as defined by the hash function. Shorter keys make it significantly to break hashing. The specification says that shorter keys should be padded: https://www.rfc-editor.org/rfc/rfc2104#section-2
CXF was apparently padding the keys, nibus on the other hand throws an error if key is too short. I added util method that pads key with "\0" which seems to do the trick. One thing I'm thinking about is introducing a utility class and moving the method there. https://github.com/MaciejMierzwa/security/blob/7e2c6ca4b9a4fa755e4ca9c81ff156e7f2eab3f1/src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java#L108C26-L108C35
2. There is similar difference in implementation for audience claims. [Specification](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.3) suggests that audience claims is array of strings, in special cases it can be single string. Nimbus is parsing audience claims as array, no way around it: [relevant part of library code](https://bitbucket.org/connect2id/nimbus-jose-jwt/src/81c7f24cc8a49f0f87c530e50d750bb1db22b4a8/src/main/java/com/nimbusds/jwt/JWTClaimsSet.java#lines-955). There are a lot of changes in tests because of that. 
3. I removed cxf-rt-rs-security-jose package from dependencies. There are 3 other cxf packages that would be ok to remove. Was intention to remove all of those?

[...]

PR #3421, opened 2023-09-28
Switch JWT library implementations from cxf to nimbus

MaciejMierzwa:
### Description
Switch from org.apache.cxf.rs.security.jose to com.nimbusds.jose.jwk.

### Issues Resolved
https://github.com/opensearch-project/security/issues/3267

### Testing

### Check List
- [x]  Add a checkstyle rule to detect including references to org.apache.cxf.rs.security.jose
- [x] Remove all references to org.apache.cxf.rs.security.jose [[15]](https://github.com/search?q=repo%3Aopensearch-project%2Fsecurity+org.apache.cxf.rs.security.jose&type=code) files impacted
- [x] Replace all unfulfilled references with com.nimbusds.jose.* components
- [x] Add backwards combability tests for any features that have previously shipped


src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -286,55 +286,60 @@ JsonWebKey createJwkFromSettings(Settings settings, Settings jwtSettings) throws
                 );
             }
 
-            JsonWebKey jwk = new JsonWebKey();
+            //TODO nimbus library doesn't accept generic key creation. Decide between RSAKey, OCTET, ECKey and create key from raw settings
```
peternied, line 289: I'd recommend to keep with Octet for the moment

src/test/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticatorTest.java
```diff
@@ -127,7 +128,7 @@ public void basicTest() throws Exception {
             .put(IDP_METADATA_URL, mockSamlIdpServer.getMetadataUri())
             .put("kibana_url", "http://wherever")
             .put("idp.entity_id", mockSamlIdpServer.getIdpEntityId())
-            .put("exchange_key", "abc")
+            .put("exchange_key", "secret_key_length_should_be_at_least_256_bits_or_86_characters_in_lengthabcdefghabcdef")
```
peternied, line 131: This is a strange requirement, I'm guessing that CXF RS was padding the key - can you see if we can automatically pad the keys in the same was as was being done with CXF, otherwise we wouldn't be able to migrate folks that are already using 'undersized' keys?

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
MaciejMierzwa: Currently there is support for for HMAC and RSA. This part of code was taking raw settings and putting it into generic cxf JWK. Nimbus library doesn't allow that generic implementation, so I'll add in-code check for RSA/HMAC algo and build object using proper constuctor.

build.gradle
```diff
@@ -488,6 +484,7 @@ dependencies {
     implementation 'commons-cli:commons-cli:1.5.0'
     implementation "org.bouncycastle:bcprov-jdk15to18:${versions.bouncycastle}"
     implementation 'org.ldaptive:ldaptive:1.2.3'
+    implementation 'com.nimbusds:nimbus-jose-jwt:9.31'
```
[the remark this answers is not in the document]
MaciejMierzwa: Should be at this point

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/JwtVerifier.java
```diff
@@ -43,31 +46,29 @@ public JwtVerifier(KeyProvider keyProvider, int clockSkewToleranceSeconds, Strin
...
             }
-            JsonWebKey key = keyProvider.getKey(kid);
+             JWK key = keyProvider.getKey(kid);
 
-            // Algorithm is not mandatory for the key material, so we set it to the same as the JWT
-            if (key.getAlgorithm() == null && key.getPublicKeyUse() == PublicKeyUse.SIGN && key.getKeyType() == KeyType.RSA) {
-                key.setAlgorithm(jwt.getJwsHeaders().getAlgorithm());
+            // TODO algorithm is final in jose implementation. Algorithm is not mandatory for the key material, so we set it to the same as the JWT, check if it's even necessary
+            if (key.getAlgorithm() == null && key.getKeyUse() == KeyUse.SIGNATURE && key.getKeyType() == KeyType.RSA) {
+//                key.setAlgorithm(jwt.getJwsHeaders().getAlgorithm());
```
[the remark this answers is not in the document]
MaciejMierzwa: Not really, removed this line alltogether

checkstyle/checkstyle.xml
```diff
@@ -219,6 +219,13 @@
     <property name="severity" value="error"/>
   </module>
 
+  <module name="RegexpSingleline">
+    <property name="format" value="println"/>
+    <property name="ignoreCase" value="true"/>
+    <property name="message" value="SYSTEM.OUT.PRINTLN NONONO!" />
+    <property name="severity" value="error"/>
+  </module>
```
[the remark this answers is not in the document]
MaciejMierzwa: This comes from first commit, done by @peternied

src/test/java/org/opensearch/security/authtoken/jwt/JwtVendorTest.java
```diff
@@ -152,7 +172,9 @@ public void testCreateJwtWithBadEncryptionKey() {
         List<String> roles = List.of("admin");
         Integer expirySeconds = 300;
 
-        Settings settings = Settings.builder().put("signing_key", "abc123").build();
+        Settings settings = Settings.builder()
+            .put("signing_key", "abc123-1234567812345678123456781234567812345678123456781234567812345678")
```
RyanL1997, line 176: Hi @MaciejMierzwa, Is there any key format restriction for this library? What is the story behind this test "signing_key" change?

src/test/java/org/opensearch/security/authtoken/jwt/JwtVendorTest.java
```diff
@@ -62,25 +70,32 @@ public void testCreateJwtWithRoles() throws Exception {
...
+        assertThat(signedJWT.getJWTClaimsSet().getClaims().get("exp"), is(notNullValue()));
         EncryptionDecryptionUtil encryptionUtil = new EncryptionDecryptionUtil(claimsEncryptionKey);
-        Assert.assertEquals(expectedRoles, encryptionUtil.decrypt(jwt.getClaim("er").toString()));
-        Assert.assertNull(jwt.getClaim("br"));
+        assertThat(encryptionUtil.decrypt(signedJWT.getJWTClaimsSet().getClaims().get("er").toString()), equalTo(expectedRoles));
+
+        assertThat(
+            encodedJwt,
+            equalTo(
+                "eyJhbGciOiJIUzUxMiJ9.eyJzdWIiOiJhZG1pbiIsImF1ZCI6ImF1ZGllbmNlXzAiLCJuYmYiOjAsImlzcyI6ImNsdXN0ZXJfMCIsImV4cCI6MCwiaWF0IjowLCJlciI6IjJ0UUd1TzczMXNYMmhraFk4OTJYUEE9PSJ9.s12TCp0aRfKLO3NZQR-Rq9UhTYIDFzGvXEonXjnqLdLxW72-J6_Q2s8RaQZBDpAWNTHzd_cxcvIjiFNddp6tPA"
```
RyanL1997, line 95: This is for debugging right?

src/test/java/org/opensearch/security/authtoken/jwt/JwtVendorTest.java
```diff
@@ -95,32 +110,34 @@ public void testCreateJwtWithRoleSecurityMode() throws Exception {
...
+        assertThat(signedJWT.getJWTClaimsSet().getClaims().get("br").toString(), equalTo(expectedBackendRoles));
+        EncryptionDecryptionUtil encryptionUtil = new EncryptionDecryptionUtil(claimsEncryptionKey);
+        assertThat(encryptionUtil.decrypt(signedJWT.getJWTClaimsSet().getClaims().get("er").toString()), equalTo(expectedRoles));
 
-        JwsJwtCompactConsumer jwtConsumer = new JwsJwtCompactConsumer(encodedJwt);
-        JwtToken jwt = jwtConsumer.getJwtToken();
+        assertThat(
+            encodedJwt,
+            equalTo(
+                "eyJhbGciOiJIUzUxMiJ9.eyJiciI6IlNhbGVzLFN1cHBvcnQiLCJzdWIiOiJhZG1pbiIsImF1ZCI6ImF1ZGllbmNlXzAiLCJuYmYiOjAsImlzcyI6ImNsdXN0ZXJfMCIsImV4cCI6MCwiaWF0IjowLCJlciI6IjJ0UUd1TzczMXNYMmhraFk4OTJYUEE9PSJ9.MlwFL1ZgcKcFoqM_7ZQEXHvSlYkmvdKflbkvjfLpmV980wd-tPwa-lMA5q1UupvCy5WGV3phsX_BIfHC5CVnyg"
```
RyanL1997, line 137: so is there any way we can get those specific claims? Like `iss`, `er` ... etc.

RyanL1997 (commented):
Hi @MaciejMierzwa, thanks for working on this. I know the pr is still in draft, and I just left some comment for now.

src/test/java/org/opensearch/security/authtoken/jwt/JwtVendorTest.java
MaciejMierzwa: Yes, it turns out that by default library forces us to use keys of certain length in this case it's 256 bit's. Otherwise tests throw:

```
OpenSearchException[com.nimbusds.jose.KeyLengthException: The secret length must be at least 256 bits
]; nested: KeyLengthException[The secret length must be at least 256 bits];
	at org.opensearch.security.authtoken.jwt.JwtVendor.createJwkFromSettings(JwtVendor.java:83)
	at org.opensearch.security.authtoken.jwt.JwtVendorTest.testCreateJwkFromSettings(JwtVendorTest.java:42)
...
Caused by: com.nimbusds.jose.KeyLengthException: The secret length must be at least 256 bits
	at com.nimbusds.jose.crypto.impl.MACProvider.<init>(MACProvider.java:118)
	at com.nimbusds.jose.crypto.MACSigner.<init>(MACSigner.java:130)
	at com.nimbusds.jose.crypto.MACSigner.<init>(MACSigner.java:178)
	at org.opensearch.security.authtoken.jwt.JwtVendor.createJwkFromSettings(JwtVendor.java:81)
	... 28 more
```

Currently looking for a solution, either padding the keys or some switch that would allow shorter ones

src/test/java/org/opensearch/security/authtoken/jwt/JwtVendorTest.java
MaciejMierzwa: Yes, removing that

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySet.java
```diff
@@ -119,11 +119,11 @@ private JsonWebKey getKeyWithKeyId(String kid) throws AuthenticatorUnavailableEx
         return result;
     }
 
-    private synchronized JsonWebKey getKeyWithRefresh(String kid) throws AuthenticatorUnavailableException {
+    private synchronized JWK getKeyWithRefresh(String kid) throws AuthenticatorUnavailableException {
```
willyborankin, line 122: why the method is `synchronized`? 
is it a global object? if it creates per request there is no need to synchronize it

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySet.java
```diff
@@ -143,11 +143,11 @@ private JsonWebKey getKeySimple(String kid) {
             }
 
         } else {
-            return jsonWebKeys.getKey(kid);
+            return jsonWebKeys.getKeyByKeyId(kid);
         }
     }
 
-    private synchronized JsonWebKey waitForRefreshToFinish(String kid) {
+    private synchronized JWK waitForRefreshToFinish(String kid) {
```
willyborankin, line 150: same here? is it a global object? if it creates per request there is no need to synchronize it

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -241,80 +240,68 @@ private BytesRestResponse handleLowLevel(RestRequest restRequest) throws SamlCon
...
-            return jwk;
+            return new OctetSequenceKey.Builder(exchangeKey.getBytes(StandardCharsets.UTF_8)).algorithm(JWSAlgorithm.HS512)
+                .keyUse(KeyUse.SIGNATURE)
+                .build();
         } else {
-
             Settings jwkSettings = jwtSettings.getAsSettings("key");
 
-            if (jwkSettings.isEmpty()) {
+            if (jwkSettings.isEmpty() || jwkSettings.get("k") == null || jwkSettings.get("k").isBlank()) {
```
willyborankin, line 259: `jwkSettings.hasValue("k") && !Strings.isEmptyOrNull(jwkSettings.get("k"))`

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -241,80 +240,68 @@ private BytesRestResponse handleLowLevel(RestRequest restRequest) throws SamlCon
...
         if (token_log.isDebugEnabled()) {
-            token_log.debug(
-                "Created JWT: "
-                    + encodedJwt
-                    + "\n"
-                    + jsonMapReaderWriter.toJson(jwt.getJwsHeaders())
-                    + "\n"
-                    + JwtUtils.claimsToJson(jwt.getClaims())
-            );
+            token_log.debug("Created JWT: " + encodedJwt + "\n" + jwt.getHeader().toString() + "\n" + jwt.getJWTClaimsSet().toString());
```
willyborankin, line 304:
```java
if (token_log.isDebugEnbaled()) {
 token_log.debug("Created JWT: " + encodedJwt + "\n" + jwt.getHeader().toString() + "\n" +   jwt.getJWTClaimsSet().toString());
}
```

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -324,20 +311,20 @@ private long getJwtExpiration(SamlResponse samlResponse) throws Exception {
         DateTime sessionNotOnOrAfter = samlResponse.getSessionNotOnOrAfter();
 
         if (this.expiryBaseValue == ExpiryBaseValue.NOW) {
-            return System.currentTimeMillis() / 1000 + this.expiryOffset;
+            return System.currentTimeMillis() + this.expiryOffset * 1000;
```
willyborankin, line 314: PLease use `TimeUnit` here

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -324,20 +311,20 @@ private long getJwtExpiration(SamlResponse samlResponse) throws Exception {
...
         } else if (this.expiryBaseValue == ExpiryBaseValue.SESSION) {
             if (sessionNotOnOrAfter != null) {
-                return sessionNotOnOrAfter.getMillis() / 1000 + this.expiryOffset;
+                return sessionNotOnOrAfter.getMillis() + this.expiryOffset * 1000;
```
willyborankin, line 317: same `TimeUnit`

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -324,20 +311,20 @@ private long getJwtExpiration(SamlResponse samlResponse) throws Exception {
...
             }
         } else {
             // AUTO
 
             if (sessionNotOnOrAfter != null) {
-                return sessionNotOnOrAfter.getMillis() / 1000;
+                return sessionNotOnOrAfter.getMillis();
             } else {
-                return System.currentTimeMillis() / 1000 + (this.expiryOffset > 0 ? this.expiryOffset : 60 * 60);
+                return System.currentTimeMillis() + (this.expiryOffset > 0 ? this.expiryOffset * 1000 : 60 * 60_000);
```
willyborankin, line 327: same `TimeUnit`

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -113,59 +126,48 @@ public String createJwt(
...
-        jwtClaims.setSubject(subject);
-
-        jwtClaims.setAudience(audience);
-
-        jwtClaims.setNotBefore(nowAsMillis);
-
-        if (expirySeconds > maxExpirySeconds) {
-            throw new Exception("The provided expiration time exceeds the maximum allowed duration of " + maxExpirySeconds + " seconds");
+        if (expirySeconds > MAX_EXPIRY_SECONDS) {
+            throw new Exception("The provided expiration time exceeds the maximum allowed duration of " + MAX_EXPIRY_SECONDS + " seconds");
```
willyborankin, line 139: `Exception` is too broad

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
```diff
@@ -119,25 +120,30 @@ private AuthCredentials extractCredentials0(final RestRequest request) throws Op
             return null;
         }
 
-        JwtClaims claims = jwt.getClaims();
+        JWTClaimsSet claimsSet;
```
willyborankin, line 123: Could you please combine all this calls in the one `try/catch` block?

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/JwtVerifier.java
```diff
@@ -97,38 +93,44 @@ private void validateSignatureAlgorithm(JsonWebKey key, JwtToken jwt) throws Bad
...
 
-    private void validateRequiredAudienceAndIssuer(JwtClaims claims) {
-        String audience = claims.getAudience();
+    private void validateRequiredAudienceAndIssuer(JWTClaimsSet claims) throws BadJWTException {
+        List<String> audience = claims.getAudience();
         String issuer = claims.getIssuer();
 
-        if (!Strings.isNullOrEmpty(requiredAudience) && !requiredAudience.equals(audience)) {
-            throw new JwtException("Invalid audience");
+        if (!Strings.isNullOrEmpty(requiredAudience) && !requiredAudience.equals(audience.stream().findFirst().orElse(""))) {
```
willyborankin, line 128: Could you please extract it to a variable `audience.stream().findFirst().orElse("")`?

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySet.java
```diff
@@ -177,7 +177,7 @@ private synchronized JsonWebKey waitForRefreshToFinish(String kid) {
         }
     }
 
-    private synchronized JsonWebKey performRefresh(String kid) {
+    private synchronized JWK performRefresh(String kid) {
```
willyborankin, line 180: same about synchronized.

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -241,80 +240,68 @@ private BytesRestResponse handleLowLevel(RestRequest restRequest) throws SamlCon
...
     private String createJwt(SamlResponse samlResponse) throws Exception {
-        JwtClaims jwtClaims = new JwtClaims();
-        JwtToken jwt = new JwtToken(jwtClaims);
-
-        jwtClaims.setNotBefore(System.currentTimeMillis() / 1000);
-        jwtClaims.setExpiryTime(getJwtExpiration(samlResponse));
-
-        jwtClaims.setProperty(this.jwtSubjectKey, this.extractSubject(samlResponse));
+        JWTClaimsSet.Builder jwtClaimsBuilder = new JWTClaimsSet.Builder().notBeforeTime(
+            new Date(new Timestamp(System.currentTimeMillis()).getTime())
```
willyborankin, line 274: please use `TimeUnit` or `ChronoUnit`

willyborankin (commented):
Overall LGTM except some comments I left. 
I think we need use `LocalDateTime` or `DateTime` or `TimeUnit` or `ChronoUnit`. To calculate expiration time for tokens.

willyborankin (dismissed):
LGTM. But I would prefer to use `java.time` API, but it is a matter of another PR .

checkstyle/checkstyle.xml
```diff
@@ -120,6 +120,7 @@
     </module>
     <module name="IllegalImport"> <!-- defaults to sun.* packages -->
       <property name="severity" value="error"/>
+      <property name="illegalPkgs" value="org.apache.cxf.rs.security.jose"/>
```
peternied, line 123: 🔥

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/JwtVerifier.java
```diff
@@ -97,38 +92,44 @@ private void validateSignatureAlgorithm(JsonWebKey key, JwtToken jwt) throws Bad
...
+            claimsVerifier.setMaxClockSkew(clockSkewToleranceSeconds);
+            claimsVerifier.verify(claims, null);
             validateRequiredAudienceAndIssuer(claims);
         }
     }
 
-    private void validateRequiredAudienceAndIssuer(JwtClaims claims) {
-        String audience = claims.getAudience();
+    private void validateRequiredAudienceAndIssuer(JWTClaimsSet claims) throws BadJWTException {
+        String audience = claims.getAudience().stream().findFirst().orElse("");
```
cwperks, line 124: If there are multiple audiences in the token claims, what should the behavior be?

Looks like cxf assumed this field was singular, but nimbusds returns the `audience` claim as a list?

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -326,20 +310,20 @@ private long getJwtExpiration(SamlResponse samlResponse) throws Exception {
         DateTime sessionNotOnOrAfter = samlResponse.getSessionNotOnOrAfter();
 
         if (this.expiryBaseValue == ExpiryBaseValue.NOW) {
-            return System.currentTimeMillis() / 1000 + this.expiryOffset;
```
cwperks, line 329: Curious, why change times in the JWT from seconds to milliseconds?

cwperks (commented):
Overall this looks good to me. I left a couple of comments.

cwperks (dismissed):
Looks good to me. Thank you @MaciejMierzwa !

Edit: Can you please resolve the conflict?

peternied (approved):
Thanks for switching us over - code changes looks good.

Lets get those conflicts resolved so this change can be merged - thanks!

peternied:
@MaciejMierzwa Not sure if you are working on this, but there are tests that are failing on both [integration-tests](https://github.com/opensearch-project/security/actions/runs/6549679224/job/17787117809?pr=3421#logs) jobs:

```
Tests with failures:
 - org.opensearch.security.http.OnBehalfOfJwtAuthenticationTest.shouldNotIncludeRolesFromHostMappingInOBOToken
 - org.opensearch.security.http.OnBehalfOfJwtAuthenticationTest.shouldAuthenticateForNonAdminUserWithOBOPermission
 - org.opensearch.security.http.OnBehalfOfJwtAuthenticationTest.shouldAuthenticateWithOBOTokenEndPoint
``` 

Just wrote this command out, it or a varient should work to run the failing tests locally `./gradlew integrationTests --tests org.opensearch.security.http.OnBehalfOfJwtAuthenticationTest`

DarshitChanpura (commented):
@MaciejMierzwa looks good. Seems like some Integ-test failures need to be addressed.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -112,60 +126,53 @@ public String createJwt(
...
 
-        expirySeconds = (expirySeconds == null) ? defaultExpirySeconds : Math.min(expirySeconds, maxExpirySeconds);
+        expirySeconds = (expirySeconds == null) ? DEFAULT_EXPIRY_SECONDS : Math.min(expirySeconds, MAX_EXPIRY_SECONDS);
         if (expirySeconds <= 0) {
-            throw new Exception("The expiration time should be a positive integer");
+            throw new IllegalArgumentException("The expiration time should be a positive integer");
         }
-        long expiryTime = timeProvider.getAsLong() + expirySeconds;
-        jwtClaims.setExpiryTime(expiryTime);
+        final Date expiryTime = new Date(timeProvider.getAsLong() + expirySeconds * 1000);
```
RyanL1997, line 149: This one right here is not in millisecond: `timeProvider.getAsLong()`. 

You have to change L69:
From `this.timeProvider = () -> System.currentTimeMillis() / 1000;` 
into `this.timeProvider = () -> System.currentTimeMillis();`

RyanL1997 (changes requested):
Hi @MaciejMierzwa, thanks for resolving all the conflicts. In generally, this is really good. However, I just found out a some issues. Could you please also rebase with main and fix that integration test case? I suspected that the failure of the integration test is related to the following finding:

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
MaciejMierzwa: Hi I'm looking into integration tests

RyanL1997:
Adding backport label since #3563 has been merged.

RyanL1997:
Hi @MaciejMierzwa, I was pushing directly to your forked branch, so please do not force push to overwrite the work. Thanks!

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -73,35 +76,46 @@ public JwtVendor(final Settings settings, final Optional<LongSupplier> timeProvi
...
-            }
+        try {
+            return new Tuple<>(key, new MACSigner(key));
+        } catch (KeyLengthException kle) {
+            throw new OpenSearchException(kle);
+        }
+    }
 
-            return jwk;
+    public static String padSecret(String signingKey, JWSAlgorithm jwsAlgorithm) {
```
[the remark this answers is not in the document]
RyanL1997: I have removed this btw. The original purpose of the padding function is to trim the keys when they are not meeting the length requirement of bits. However, this should be the scenario of throwing the exception and let users know the key they are providing is invalid.

RyanL1997:
Everything has been fixed.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -106,18 +101,6 @@ static Tuple<JWK, JWSSigner> createJwkFromSettings(Settings settings) {
...
-        try {
-            requiredSecretLength = getMinRequiredSecretLength(jwsAlgorithm);
-        } catch (JOSEException e) {
-            throw new RuntimeException(e);
-        }
-        int requiredByteLength = ByteUtils.byteLength(requiredSecretLength);
-        // padding the signing key with 0s to meet the minimum required length
-        return StringUtils.rightPad(signingKey, requiredByteLength, "\0");
-    }
-
```
MaciejMierzwa: Hi, @RyanL1997 regarding this part of code. It was added for the backwards compatibility. If user uses signing secret that's shorter then 64bytes nimbus would throw errors like that:
image
Overall I think that it would be a good idea to force users to use longer, safer secret key. On the other hand, in previous discussion, we agreed that users should not be surprised by the sudden change. This padding would also allow for JWTs generated by node running cxf implementation, to get parsed by node running with nimbus implementation.

RyanL1997: Hi @MaciejMierzwa , thanks for the info! Yes, I think if this is the case, lets bring back the padding function. I'm on it.

MaciejMierzwa: Hi, take a look at this: https://github.com/opensearch-project/security/commit/eb4533766aeabbe0be3838de7f7976364c562054. I did bring it back with slight modifications in a different branch. As I understand the config for signing secret is base64url encoded? Padding I added again is taking this into consideration. So it's reading config in base64 -> decoding it -> padding -> creating jwk for signing jwt.

src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java
```diff
@@ -111,6 +111,65 @@ public void testInvalid() throws Exception {
         Assert.assertNull(credentials);
     }
 
+    @Test
+    public void testParsePrevGeneratedJwt() {
```
peternied, line 115: I don't normally suggestion including commented out code, but I think this is a very good case - could you include the code used to generate the previous token and a link to where you injected it?

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -117,6 +121,18 @@ class AuthTokenProcessorHandler {
         this.jwsHeader = this.createJwsHeaderFromSettings();
     }
 
+    public static String padSecret(String signingKey, JWSAlgorithm jwsAlgorithm) {
```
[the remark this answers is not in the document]
RyanL1997: I just moved this into an util and added a simple unit tests

src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java
RyanL1997: I just added

RyanL1997:
looking into the failure of ci

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/JwtVerifier.java
```diff
@@ -43,31 +46,24 @@ public JwtVerifier(KeyProvider keyProvider, int clockSkewToleranceSeconds, Strin
...
 
-            boolean signatureValid = jwtConsumer.verifySignatureWith(signatureVerifier);
+            JWSVerifier signatureVerifier = getInitializedSignatureVerifier(key, jwt);
+            boolean signatureValid = jwt.verify(signatureVerifier);
 
             if (!signatureValid && Strings.isNullOrEmpty(kid)) {
                 key = keyProvider.getKeyAfterRefresh(null);
                 signatureVerifier = getInitializedSignatureVerifier(key, jwt);
-                signatureValid = jwtConsumer.verifySignatureWith(signatureVerifier);
+                signatureValid = jwt.verify(signatureVerifier);
```
peternied, line 66: Looks like we don't have a code path hitting this section anymore, I'm forgetting what purpose it serves, can we remove?

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -326,20 +314,20 @@ private long getJwtExpiration(SamlResponse samlResponse) throws Exception {
         DateTime sessionNotOnOrAfter = samlResponse.getSessionNotOnOrAfter();
 
         if (this.expiryBaseValue == ExpiryBaseValue.NOW) {
-            return System.currentTimeMillis() / 1000 + this.expiryOffset;
+            return System.currentTimeMillis() + this.expiryOffset * 1000;
```
peternied, line 317: Umm, this looks like it could be a bug - we should have test that confirms this behavior.

src/test/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticatorTest.java
```diff
@@ -132,7 +177,7 @@ public void basicTest() throws Exception {
             .put(IDP_METADATA_URL, mockSamlIdpServer.getMetadataUri())
             .put("kibana_url", "http://wherever")
             .put("idp.entity_id", mockSamlIdpServer.getIdpEntityId())
-            .put("exchange_key", "abc")
+            .put("exchange_key", signingKeyB64Encoded)
```
peternied, line 180: Why did we need to change this key? Seems like it should be the same as it was before the change?

peternied (changes requested):

src/test/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticatorTest.java
RyanL1997: Fixed. I just reverted back. I was changing this cuz I was removing the padding function, so the key was forced to be at least 256 bits long. However, we brought back the padding function, so that this should be fine now.

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
RyanL1997: No, this isn't a bug, since the new library is taking the time as a `Date` type, which is in milliseconds. When we set up the expiry, it takes something like this:
```
/**
 * Sets the expiration time ({@code exp}) claim.
 *
 * @param exp The expiration time, {@code null} if not
 *            specified.
 *
 * @return This builder.
 */

public Builder expirationTime(final Date exp) {

  claims.put(JWTClaimNames.EXPIRATION_TIME, exp);
  return this;
}
```
So all the tests with Authentication should cover this change, since we do need the correct expiry to authenticate.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/JwtVerifier.java
RyanL1997: We do have this one in the abstract of `HTTPJwtAuthenticator` for the verification of the token.

RyanL1997: Here: https://github.com/opensearch-project/security/pull/3421/files/10045958a33fc496470c75dd8cebf9ac2cfd71cb#diff-47ab5d4b1ee445b7a9fedb95d77b06fd263414d55adc7add00400cd32ec30c2aL118

peternied (approved):
Thanks for the answers, looks great @RyanL1997 & @MaciejMierzwa

DarshitChanpura (approved):
LGTM!

RyanL1997:
Manual backport has been created: https://github.com/opensearch-project/security/pull/3595

PR #3430, opened 2023-10-01
Expanding Authentication with SecurityRequest Abstraction

peternied:
### Description
Introduced a new abstraction, SecurityRequest, to streamline and secure the authentication process in the OpenSearch Security plugin. By isolating the essential request components needed for authentication, we minimize potential risks associated with previous designs and provide a more maintainable architecture.

```mermaid
graph TD
    A[SecurityRequestFactory] -->|Creates| B[SecurityRequest]
    C[Netty Channel Handler] -->|Provides data to| A
    D[OpenSearch Rest Controller] -->|Provides data to| A
    B --> |Used in| SecurityFilter
    B --> |Used in| BackendRegistry
    B --> |Used in| HttpAuthenticator

    style A fill:#f9d,stroke:#333,stroke-width:2px;
    style B fill:#afd,stroke:#333,stroke-width:2px;
    style C fill:#fea,stroke:#333,stroke-width:2px;
    style D fill:#fea,stroke:#333,stroke-width:2px;

```

#### Draft Status
For this pull request to transition out of draft mode, it's imperative that we validate the compatibility and robustness of the new changes, especially with flows originating from a Netty source - as we have solid coverage from OpenSearch flows. Key tasks include:

1. Experimentation with how actions on the RestChannel/NettyChannel are applied or abstracted.
1. Integrating and wiring the new flow to ensure seamless operation around channel related callbacks.
1. Implementing comprehensive testing to address potential edge cases.
1. Reviewing and possibly enhancing existing tests to confirm their relevance and accuracy.

Upon fulfilling these conditions can we be confident in the stability and effectiveness of the proposed modifications.

### Issues Resolved
- Related https://github.com/opensearch-project/OpenSearch/issues/10260

### Check List


peternied:
@cwperks I had previously included a `AuthenticatorVerifier` that would use this shared request interface as part of the updated Netty flow, instead I'm going to pull that into an isolated change which is more consumable since most of this change is security infrastructure / refactoring

peternied:
After this change is merged, https://github.com/peternied/security/pull/13/files is the delta needed to add the AuthenticationVerifier.

willyborankin (dismissed):

src/main/java/org/opensearch/security/dlic/rest/api/RestApiPrivilegesEvaluator.java
```diff
@@ -446,7 +448,11 @@ private String checkAdminCertBasedAccessPermissions(RestRequest request) throws
         }
 
         // Certificate based access, Check if we have an admin TLS certificate
-        SSLRequestHelper.SSLInfo sslInfo = SSLRequestHelper.getSSLInfo(settings, configPath, request, principalExtractor);
+        final SecurityRequestChannel securityRequest = SecurityRequestFactory.from(
+            request,
+            /* OK to be null so long as a response isn't sent */ null
```
DarshitChanpura, line 453: Can you elaborate on this comment?

src/main/java/org/opensearch/security/filter/OpenSearchRequest.java
```diff
@@ -0,0 +1,78 @@
...
+            || underlyingRequest.getHttpChannel() == null
+            || !(underlyingRequest.getHttpChannel() instanceof Netty4HttpChannel)) {
+            return null;
+        }
+
+        final Netty4HttpChannel httpChannel = (Netty4HttpChannel) underlyingRequest.getHttpChannel();
+        SslHandler sslhandler = (SslHandler) httpChannel.getNettyChannel().pipeline().get("ssl_http");
+        if (sslhandler == null && httpChannel.inboundPipeline() != null) {
+            sslhandler = (SslHandler) httpChannel.inboundPipeline().get("ssl_http");
+        }
```
DarshitChanpura, line 44: What do you think about adding a brief comment here that explains these 4 lines of code:
```java
// We look for Ssl_handler called `ssl_http` in the outbound pipeline of Netty channel first, and if its not
// present we look for it in inbound channel. If its present in neither we return null, else we return the sslHandler.
```

src/main/java/org/opensearch/security/filter/SecurityRequestFactory.java
```diff
@@ -0,0 +1,20 @@
...
+
+import org.opensearch.rest.RestChannel;
+import org.opensearch.rest.RestRequest;
+
+/**
+ * Generates wrapped versions of requests for use in the security plugin
+ */
+public class SecurityRequestFactory {
+
+    /** Creates a security requset from a RestRequest */
```
DarshitChanpura, line 11: nit: request

src/main/java/org/opensearch/security/filter/SecurityRequestChannel.java
```diff
@@ -0,0 +1,17 @@
+package org.opensearch.security.filter;
+
+import java.util.Map;
+
+/**
+ * When a request is recieved by the security plugin this governs getting information about the request as well as a way to complet
```
DarshitChanpura, line 6: nit: received & complete

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -131,15 +128,32 @@ public SecurityRestFilter(
...
+            final SecurityRequestChannel requestChannel = SecurityRequestFactory.from(request, channel);
+
+            // Authenticate request
+            checkAndAuthenticateRequest(requestChannel);
+            if (requestChannel.hasCompleted()) {
+                // Unable to authenticate the caller
+                return;
+            }
+
+            // Authorize Requset
```
DarshitChanpura, line 140: nit: Request

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -131,15 +128,32 @@ public SecurityRestFilter(
...
-            if (!checkAndAuthenticateRequest(request, channel)) {
-                User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
-                boolean isSuperAdminUser = userIsSuperAdmin(user, adminDNs);
-                if (isSuperAdminUser
-                    || (whitelistingSettings.checkRequestIsAllowed(request, channel, client)
-                        && allowlistingSettings.checkRequestIsAllowed(request, channel, client))) {
-                    if (isSuperAdminUser || authorizeRequest(original, request, channel, user)) {
-                        original.handleRequest(request, channel, client);
-                    }
+            final SecurityRequestChannel requestChannel = SecurityRequestFactory.from(request, channel);
```
DarshitChanpura, line 131: 👏 Much cleaner impl

src/main/java/org/opensearch/security/ssl/http/netty/SecuritySSLNettyHttpServerTransport.java
```diff
@@ -94,6 +94,12 @@ public void onException(HttpChannel channel, Exception cause0) {
         super.onException(channel, cause0);
     }
 
+    // TODO: Disable since changes in core aren't avaliable yet
```
DarshitChanpura, line 97: Is there a PR associated with this change in core?

src/test/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticatorTest.java
```diff
@@ -908,10 +852,7 @@ public void initialConnectionFailureTest() throws Exception {
 
     private AuthenticateHeaders getAutenticateHeaders(HTTPSamlAuthenticator samlAuthenticator) {
         RestRequest restRequest = new FakeRestRequest(ImmutableMap.of(), new HashMap<String, String>());
-        TestRestChannel restChannel = new TestRestChannel(restRequest);
-
-        final BytesRestResponse authenticatorResponse = samlAuthenticator.reRequestAuthentication(restRequest, null);
-        restChannel.sendResponse(authenticatorResponse);
+        TestRestChannel restChannel = sendToAuthenticator(samlAuthenticator, restRequest);
 
         List<String> wwwAuthenticateHeaders = restChannel.response.getHeaders().get("WWW-Authenticate");
```
DarshitChanpura, line 857: 👏 nicely done

DarshitChanpura (commented):
Great stuff @peternied. I like the approach of SecurityRequest and SecurityRequestChannel. It is much cleaner and easier to follow. I've left some comments to understand some changes better and some nits for spelling

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
cwperks, line 131: @peternied Do you think it makes sense in the long-term to aim for the security plugin to not have access to the channel at all? I'd like to eventually have a clean interface in core that the security plugin can provide an `earlyResponse` something similar that can then be used to send the response before the original handler is called.

src/main/java/com/amazon/dlic/auth/http/kerberos/HTTPSpnegoAuthenticator.java
```diff
@@ -369,7 +368,7 @@ private static String getUsernameFromGSSContext(final GSSContext gssContext, fin
         return null;
     }
 
-    private XContentBuilder getNegotiateResponseBody() {
+    private String getNegotiateResponseBody() {
```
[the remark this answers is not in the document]
peternied: We weren't using the 'XContent''ness other than for serializing the JSON, easier to keep it away from the interfaces so they are clean when netty is involved.

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -118,18 +122,19 @@ class AuthTokenProcessorHandler {
     }
 
     @SuppressWarnings("removal")
-    BytesRestResponse handle(RestRequest restRequest) throws Exception {
+    boolean handle(RestRequest restRequest, RestChannel restChannel) throws Exception {
```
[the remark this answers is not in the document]
peternied: I still need to do a deeper dive into your changes here, I might have reverted them too quickly.

This class it was written in such a way that it wouldn't require authentication to process the request which is the ACS endpoint request for SAML, but frankly it should be registered as our other http handlers with an allowance not to require authentication, it would greatly clean up these classes if this was 'unhooked' from the reauth flow.  I tried to avoid doing too many refactors at once.

src/main/java/org/opensearch/security/dlic/rest/api/RestApiPrivilegesEvaluator.java
peternied: Thanks for catching, I created SecurityRequest & SecurityRequestChannel to handle this kind of case - I just need to clean this up to use SecurityRequest.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
peternied: I think the concept of an 'early' response shouldn't be part of the RestHandler ecosystem at all - can you get more than one, what is the order they are used ? is a good example of a problem to avoid

However, in the context of the HeaderValidator and the classes inside of core that use it, it might make sense to check for an early response.

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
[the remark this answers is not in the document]
peternied: I've reincluded your changes, but I changed the format to `SecurityResponse` to keep using a more generic interface

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
peternied: 🙇

src/main/java/org/opensearch/security/ssl/http/netty/SecuritySSLNettyHttpServerTransport.java
peternied: Oh, I should pull that commented out section from here, you can see the core changes [1] on PR 10261

- [1] https://github.com/opensearch-project/OpenSearch/pull/10261

peternied:
@cwperks I'm going to refrain from anymore force pushes, and try to keep the renames/refactors very limited - feel free to start merging from this change

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -131,17 +128,37 @@ public SecurityRestFilter(
...
+
+            // Authorize Request
+            final User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+            if (userIsSuperAdmin(user, adminDNs)) {
+                // Super admins are always authorized
+                original.handleRequest(request, channel, client);
+                return;
+            }
+
+            if (!(whitelistingSettings.checkRequestIsAllowed(request, channel, client)
```
cwperks, line 148: nit: wdyt about filing an issue to refactor these to return a Response object and then call on `channel.sendResponse` in here instead of relying on `checkRequestIsAllowed` to send the response as a side-effect of the call?

src/main/java/org/opensearch/security/dlic/rest/api/RestApiPrivilegesEvaluator.java
peternied: Just to be clear - this confusing comment and the null value have been removed.

src/main/java/org/opensearch/security/ssl/util/SSLRequestHelper.java
```diff
@@ -121,25 +119,14 @@ public String toString() {
...
-        }
-
-        final Netty4HttpChannel httpChannel = (Netty4HttpChannel) request.getHttpChannel();
-        SslHandler sslhandler = (SslHandler) httpChannel.getNettyChannel().pipeline().get("ssl_http");
-        if (sslhandler == null && httpChannel.inboundPipeline() != null) {
-            sslhandler = (SslHandler) httpChannel.inboundPipeline().get("ssl_http");
-        }
-
-        if (sslhandler == null) {
+        final SSLEngine engine = request.getSSLEngine();
```
RyanL1997, line 125: nit: sslEngine?

src/main/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticator.java
```diff
@@ -397,7 +409,7 @@ String buildLogoutUrl(AuthCredentials authCredentials) {
 
     }
 
-    private void initLogoutUrl(RestRequest restRequest, ThreadContext threadContext, AuthCredentials authCredentials) {
+    private void initLogoutUrl(ThreadContext threadContext, AuthCredentials authCredentials) {
```
RyanL1997, line 412: Why don't we need the request parameter here?

RyanL1997 (commented):
Hi @peternied, great change. I just left some comments

src/main/java/org/opensearch/security/http/XFFResolver.java
```diff
@@ -53,44 +55,45 @@ public XFFResolver(final ThreadPool threadPool) {
         this.threadContext = threadPool.getThreadContext();
     }
 
-    public TransportAddress resolve(final RestRequest request) throws OpenSearchSecurityException {
+    public TransportAddress resolve(final SecurityRequest request) throws OpenSearchSecurityException {
         final boolean isTraceEnabled = log.isTraceEnabled();
         if (isTraceEnabled) {
-            log.trace("resolve {}", request.getHttpChannel().getRemoteAddress());
+            log.trace("resolve {}", request.getRemoteAddress().orElse(null));
```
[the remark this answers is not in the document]
peternied: Previous null was written to the output, by using Optional<> it makes this apparent how you are seeing it now.

The behavior is identical to how it worked before this change

src/main/java/org/opensearch/security/ssl/util/SSLRequestHelper.java
peternied: I'll keep it as is, thanks for the feedback

src/main/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticator.java
peternied: We never used it, there are a couple of places where I found unused parameters and removed them rather than refactoring RestRequest -> SecurityRequest

src/main/java/org/opensearch/security/http/RemoteIpDetector.java
```diff
@@ -196,7 +203,7 @@ String detect(RestRequest request, ThreadContext threadContext) {
                 log.trace(
                     "Skip RemoteIpDetector for request {} with originalRemoteAddr '{}' cause no internal proxy matches",
                     request.uri(),
-                    request.getHttpChannel().getRemoteAddress()
+                    request.getRemoteAddress().orElse(null)
```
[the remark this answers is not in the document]
peternied: The core of this change is around allowing different request implementations through the authentication systems. Not all requests use the same format for the IP Address by wrapping the value in a interface we can make them map to a uniform style.  

There is a possibility of no ip address on some request flows, by using optional it forces deal with cases where the value is undefined

RyanL1997 (approved):

src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java
```diff
@@ -11,17 +11,21 @@
...
 import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
 import java.util.Map.Entry;
 import java.util.regex.Pattern;
 
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.cxf.rs.security.jose.jwt.JwtClaims;
 import org.apache.cxf.rs.security.jose.jwt.JwtToken;
-import org.apache.hc.core5.http.HttpHeaders;
```
cwperks, line 24: nit: Use this version of HttpHeaders for importing the AUTHORIZATION constant

peternied: Good catch, I'll fast follow fix this

src/main/java/org/opensearch/security/filter/SecurityRequestChannel.java
```diff
@@ -0,0 +1,29 @@
...
+public interface SecurityRequestChannel extends SecurityRequest {
+
+    /** Associate a response with this channel */
+    public void queueForSending(final SecurityResponse response);
+
+    /** Acess the queued response */
+    public Optional<SecurityResponse> getQueuedResponse();
+
+    /** Send the response through the channel */
+    public boolean sendResponse();
```
cwperks, line 28: FYI when I was implementing this interface in https://github.com/opensearch-project/security/pull/3418 I ran into an issue where this method needs to consume a RestChannel because a RestChannel would not be available in the `header_verifier` which needs to create an instance of `SecurityRequestChannel` to send to `checkAndAuthenticate`

cwperks (approved):

Issue #1832, opened 2022-05-09
[Test Refactoring] Replace `Assert.assertEquals()` with `assertThat()`


cliu123:
coming from a [PR review comment](https://github.com/opensearch-project/security/pull/1808/files#r864082852)
Use `assertThat(...)` to replace `Assert.assertEquals(...)`, as it will provide a more verbose error message.

cwperks:
Closing this as completed with https://github.com/opensearch-project/security/pull/4544

PR #3443, opened 2023-10-03
Response to Issue #1832 [Replace Assert.assertEquals() with assertThat()]

EduardoCorazon:
### Description
This change is a response to opensearch-project/security issue #1832 whereby it's recommended to replace Assert.assertEquals(...) with assertThat(...) for a greater degree of verbosity.
* Refactoring
* Using the Hamcrest matchers with assertThat() yields clearer failure messages for testing.
* The old behavior uses assertEquals() which is part of the Assert class in JUnit meanwhile the new method, assertThat(), belongs to the Hamcrest class which tends to be more human readable. 

### Issues Resolved
Addressed security issue #1832 
[Test Refactoring] Replace Assert.assertEquals() with assertThat()
https://github.com/opensearch-project/security/issues/1832 

### Testing
Verified code coherence and compatibility by integrating with OpenSearch for comprehensive testing. While I did not encounter any errors a detailed unit test is recommended for more detailed testing and full compatibility.

### Check List


cwperks:
Thank you @EduardoCorazon , I have approved the CI to be run.

This change looks good to me, but let's see if the CI checks uncover anything.

cwperks (approved):
LGTM! 🚢

DarshitChanpura (commented):
Thank you for this contribution @EduardoCorazon. I've found few other test files [1] that are using assertEquals. How do you feel about making those changes and including them here?

[1]: https://github.com/search?q=repo%3Aopensearch-project%2Fsecurity+assertEquals&type=code

EduardoCorazon:
@DarshitChanpura Certainly! I'll get right to it.

src/test/java/org/opensearch/security/dlic/rest/api/AllowlistApiTest.java
```diff
@@ -97,7 +97,9 @@ public void testResponseDoesNotContainMetaHeader() throws Exception {
 
         rh.sendAdminCertificate = true;
         RestHelper.HttpResponse response = rh.executeGetRequest(ENDPOINT);
-        Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
+
+        // Replaced Assert.asserEquals to assertThat
```
peternied, line 101: Lets remove these comments; `git diff` `git blame` and `git log -- src/test/java/org/opensearch/security/dlic/rest/api/AllowlistApiTest.java` are enough to track the were and why of a given change.

src/test/java/org/opensearch/security/securityconf/impl/v7/ConfigV7Test.java
```diff
@@ -13,10 +13,11 @@
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.ImmutableList;
-import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
```
peternied, line 20: Pretty sure there is a CI failure because `*` imports shouldn't be allowed, please expand this out.

src/test/java/org/opensearch/security/securityconf/impl/v7/ConfigV7Test.java
peternied, line 1: I don't think this is a useful test file, at all.  I think it would be worthwhile to delete it entirely rather than spend time on cleaning it up.  See this stackoverflow question [How Should I Unit Test A Data Transfer Object?](https://softwareengineering.stackexchange.com/questions/408837/how-should-i-unit-test-a-data-transfer-object) that I think applies here.

What do you think?

peternied (changes requested):

src/test/java/org/opensearch/security/securityconf/impl/v7/ConfigV7Test.java
willyborankin: `gradle checkstyleMain checkstyleTest` and `gradle spotlessCheck spotlessApply` commends help to solve such problems :-)

EduardoCorazon:
I apologize for the delayed response on this pr, I'll get right back to work!

src/test/java/org/opensearch/security/securityconf/impl/v7/ConfigV7Test.java
EduardoCorazon: While I think it ultimately depends on the thoughts of the other contributors as well, I believe it might be good to keep this test file. While it does increase maintenance effort it might be good to have a little safety net for regressions.

src/test/java/org/opensearch/security/securityconf/impl/v7/ConfigV7Test.java
EduardoCorazon: Thank you ! :)

PR #3457, opened 2023-10-04
Get OpenSearch version from gradle properties

davidosorno:
### Description
* Replaced the asterisk symbol to prevent potential confusion when naming folders. In a later section of the document, the asterisk symbol is reintroduced and should be used accordingly.
* Added `-p` to `mkdir` command to allow it to create subfolders if needed.

### Check List

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -30,9 +30,9 @@ To get started, follow the [getting started section](https://github.com/opensear
 
 Next, run the following commands to copy the built code (snapshot) to a new folder in a different location. (This where you'll be running the OpenSearch service). Run this from the base directory of the OpenSearch fork you cloned above:
 ```bash
-export OPENSEARCH_HOME=~/<your-folder-location>/opensearch-*
-export OPENSEARCH_BUILD=distribution/archives/darwin-tar/build/install/opensearch-*
-cp -Rf $OPENSEARCH_BUILD $OPENSEARCH_HOME
+export OPENSEARCH_HOME=~/<your-folder-location>/opensearch-{VERSION-NUMBER}
+export OPENSEARCH_BUILD=distribution/archives/darwin-tar/build/install/opensearch-{VERSION-NUMBER}-SNAPSHOT
+cp -Rf $OPENSEARCH_BUILD/* $OPENSEARCH_HOME
 ```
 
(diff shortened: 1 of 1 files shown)


DEVELOPER_GUIDE.md
```diff
@@ -30,9 +30,9 @@ To get started, follow the [getting started section](https://github.com/opensear
 
 Next, run the following commands to copy the built code (snapshot) to a new folder in a different location. (This where you'll be running the OpenSearch service). Run this from the base directory of the OpenSearch fork you cloned above:
 ```bash
-export OPENSEARCH_HOME=~/<your-folder-location>/opensearch-*
-export OPENSEARCH_BUILD=distribution/archives/darwin-tar/build/install/opensearch-*
-cp -Rf $OPENSEARCH_BUILD $OPENSEARCH_HOME
+export OPENSEARCH_HOME=~/<your-folder-location>/opensearch-{VERSION-NUMBER}
```
peternied, line 33: On this line there is another kind of substitution parameter `<your-folder-location>`, I think we should pick a single format.  However, I think it might be even more worthwhile to see about reframing this section to not require developers to think about the version number, how does that sound?

Maybe we can pull version numbers from a gradle command, or build up a script that will do the 'painful' stuff so that you don't have to when you want to get the instance up and running.

https://github.com/opensearch-project/security/blob/87520741566eaa4ea609441a1f53f3cef4a9674c/.github/workflows/ci.yml#L180-L182

peternied (changes requested):
Thanks for this change - the `-p` options are a really good idea to make that part of the script more robust.  I think there are improvements that can also be done to the top section that would reduce the amount of intervention for substitution.

DEVELOPER_GUIDE.md
davidosorno: That's a great idea. Let me work on it. Thanks @peternied.

willyborankin (approved):

DarshitChanpura (approved):

DarshitChanpura:
@peternied Would you please re-review this? Seems like your changes have been addressed, but I chose not to dismiss your review to wait for your input.

peternied (approved):

DarshitChanpura:
@davidosorno Mind creating backports?

davidosorno:
> @davidosorno Mind creating backports? 

Not at all but, @DarshitChanpura can you please clarify what you meant with backports?  I read the backport definition but I don't quite understand how I can apply it with this document.

PR #3509, opened 2023-10-09
Enhance SelfRefreshingKeySet test coverage

### LINKED ISSUE #3137: [BUG] Unexpected Coverage Changes on SelfRefreshingKeySet, SecurityInterceptor and SecuritySSLNettyTransport

davidosorno:
### Description
Added unit testing to address the  random behavior in code coverage reports.

### Issues Resolved
- #3137 

### Testing
* Added unit testing to SelfRefreshingKeySet.
* Improve the race condition test for SelfRefreshingKeySet using synchronized key work and a lock object to ensure exclusive access by a single thread to this block of code. Used multi threading calls to add a substantial number of  concurrent tasks for token refreshing, so it enqueue multiple calls. At the end of the test the queue must be empty.
* Added unit testing to SecuritySSLNettyTransport.

### Check List


davidosorno:
@peternied can I get some feedback about this? I would like to know if I'm in the right way, before continuing with the other 2 files. Thanks.

src/test/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransportTests.java
```diff
@@ -0,0 +1,101 @@
...
+
+    ChannelHandler channelHandler = securitySSLNettyTransport.getServerChannelInitializer("test-server-channel");
+
+    Assert.assertNotNull(channelHandler);
+    Assert.assertTrue(channelHandler instanceof SecuritySSLNettyTransport.SSLServerChannelInitializer);
+
+  }
+
+  @Test
+  public void getClientChannelInitializer_shouldReturnValidClienteChannel() {
```
cwperks, line 92:
```suggestion
  public void getClientChannelInitializer_shouldReturnValidClientChannel() {
```

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
```diff
@@ -11,108 +11,103 @@
...
 
-        Future<JsonWebKey> f2 = executorService.submit(() -> selfRefreshingKeySet.getKey("kid/b"));
+		Assert.assertThrows(BadCredentialsException.class, () -> selfRefreshingKeySet.getKey("kid/X"));
+	}
 
-        while (selfRefreshingKeySet.getQueuedGetCount() == 0) {
-            Thread.sleep(10);
-        }
+	@Test
+	public void getKeyAfterRefresh_queuedGetCountVariableShouldBeZeroWhenFinishWithAllKeyRefreshes()
```
cwperks, line 65: Nice! Do you know if this class has tests that validate scenarios where key refresh fails?

Looks like OIDC Authenticator might use a SelfRefreshingKeySet to retrieve keys from the well_known endpoint (`.../well-known/openid-configuration`) which can fail.

davidosorno: I created tests to validate the exceptions that the methods throw when they get null or invalid key, but I'm not sure if this is enough:
- getKey_withNullKidShouldThrowAuthenticatorUnavailableException
- getKey_withInvalidDataShouldReturnBadCredentialException
- getKeyAfterRefresh_withNullKidShouldThrowBadCredentialsException
- getKeyAfterRefresh_withInvalidDataShouldReturnBadCredential

@cwperks  do you think I should cover other scenarios?

cwperks: I think that covers a wide range of scenarios. Nice work on these tests!

Are you still working on this PR or can it be flipped to in review?

src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
```diff
@@ -0,0 +1,132 @@
...
+    when(configRepository.getConfiguration(CType.ROLES)).thenReturn(SecurityDynamicConfiguration
+        .fromJson(objectMapper.writeValueAsString(objectNode), CType.ROLES, 2, 1, 1));
+
+    SecurityDynamicConfiguration<?> config2 = configRepository.getConfiguration(CType.ROLES);
+
+    assertThat(config2, is(notNullValue()));
+    assertThat(config2, instanceOf(SecurityDynamicConfiguration.class));
+    assertThat(config2.getCEntries().size(), is(greaterThan(0)));
+  }
+}
```
scrawfor99, line 132: Nit: missing new line at end of file

src/test/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransportTests.java
```diff
@@ -0,0 +1,106 @@
...
+  @Test
+  public void getClientChannelInitializer_shouldReturnValidClientChannel() {
+
+    ChannelHandler channelHandler = securitySSLNettyTransport.getClientChannelInitializer(discoveryNode);
+    
+    assertThat(channelHandler, is(notNullValue()));
+    assertThat(channelHandler, instanceOf(SSLClientChannelInitializer.class));
+  }
+
+}
```
scrawfor99, line 106: Nit: Missing new line

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
```diff
@@ -11,108 +11,103 @@
...
+					} catch (BadCredentialsException e) {
+					}
+				}
+			});
+		}
 
-        Assert.assertEquals(1, selfRefreshingKeySet.getRefreshCount());
-        Assert.assertEquals(1, selfRefreshingKeySet.getQueuedGetCount());
+		executor.shutdown();
+		executor.awaitTermination(1, TimeUnit.MINUTES);
```
scrawfor99, line 85: Did you find 1 minute from somewhere? I don't know that it is too long or a bad number, but trying to gauge whether we need to keep this so high. In general, I hesitate to add timeouts more then 10 seconds since we end up running tests so often that 1 minute waiting can add up....

Was this from something else or a commonly used value?

scrawfor99 (commented):
Just a few comments, but really great job!

cwperks:
@davidosorno The attached codecov report shows the coverage of SelfRefreshingKeySet going down ~3.5%. I know the attached issue was on unpredictability of the coverage amount for the referenced classes. Does this PR help make codecov more predictable so on consecutive CI runs it will return the same coverage percentage?

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
davidosorno: Sorry, typo here. What I wanted to use was milliseconds.
Now, I realized that it doesn't work with 1 millisecond, so I use 1 second to wait until it finishes.

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
```diff
@@ -11,108 +11,100 @@
...
-    static class BlockingMockKeySetProvider extends MockKeySetProvider {
-        private boolean blocked = true;
-        private boolean called = false;
+        executor.shutdown();
+        executor.awaitTermination(1, TimeUnit.SECONDS);
 
-        @Override
-        public synchronized JsonWebKeys get() throws AuthenticatorUnavailableException {
+        Assert.assertEquals(10, selfRefreshingKeySet.getRefreshCount());
+        Assert.assertEquals(0, selfRefreshingKeySet.getQueuedGetCount());
```
DarshitChanpura, line 85: Why does the count for queue become 0 here compared to 1 in the previous test?

src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
```diff
@@ -0,0 +1,122 @@
+package org.opensearch.security.configuration;
```
DarshitChanpura, line 1: nit: License header.
Thank you for adding these tests.

src/test/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransportTests.java
```diff
@@ -0,0 +1,105 @@
+package org.opensearch.security.ssl.transport;
```
DarshitChanpura, line 1: nit: license header

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
davidosorno: The previous test used Future class to call the getKey(...) method twice and the queue increments when detects a refresh Key in progress, that is why  `while (selfRefreshingKeySet.getQueuedGetCount() == 0) { Thread.sleep(10); }` was used for. So, when the system detects that the queue increments, it resumes the test and passes the assertion with number 1.
On the other hand, when I was working on improving the race condition test, there is a note in SelfRefreshingKeySetc class `// Always re-check within synchronized to handle any races` so I thought it would be more precisely if I use more threads with a lock object to access the refresh method and make sure only one thread can access the key, and then ensure all calls are executed correctly, this is why the queue must be empty at the end of the test.
@DarshitChanpura please, let me know if I answered your question correctly, and thanks for helping me to clarify my approach.

DarshitChanpura: aah, thank you for the explanation @davidosorno.

DarshitChanpura:
@davidosorno Could you merge latest changes from main? This would solve CI issues related JsonWebKey as there was a recent library change.

src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+        when(configRepository.getConfiguration(CType.ROLES)).thenReturn(
+            SecurityDynamicConfiguration.fromJson(objectMapper.writeValueAsString(objectNode), CType.ROLES, 2, 1, 1)
+        );
+
+        SecurityDynamicConfiguration<?> config2 = configRepository.getConfiguration(CType.ROLES);
+
+        assertThat(config2, is(notNullValue()));
+        assertThat(config2, instanceOf(SecurityDynamicConfiguration.class));
+        assertThat(config2.getCEntries().size(), is(greaterThan(0)));
```
DarshitChanpura, line 132: Can you add a test here to ensure that it is actually a clone and not the same copy?

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
```diff
@@ -11,109 +11,104 @@
...
+
+        for (int i = 0; i < numThreads; i++) {
+            executor.submit(() -> {
+                synchronized (lock) {
+                    try {
+                        selfRefreshingKeySet.getKeyAfterRefresh("kid/a");
+                    } catch (AuthenticatorUnavailableException e) {} catch (BadCredentialsException e) {}
+                }
+            });
+        }
```
peternied, line 84: I'm not sure how this works or what it tests, can you provide more details.

Also of note;
- Synchronize block at @78 prevents the threads from executing at the same time
- the method `getKeyAfterRefresh(...)` is synchronized, so it cannot be executed in parallel, even without the sync block

davidosorno: @peternied you're correct. The lock object is preventing the threads to execute at the same time. However, my intention was to make sure that getRefreshCount() has the same amount of calls that we have in numThreads and getQueuedGetCount() has zero pending calls.
I created another test without lock and the assertions are quite different since there is not lock object, so the main thread just keep executing. I'll push it and see what you think.

src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
davidosorno: @DarshitChanpura thanks to you comment I realized that this test is not working as I expected and I spent a lot of time trying to make it work but I couldn't. This line `SecurityDynamicConfiguration<?> conf = configCache.getIfPresent(configurationType);`  always returns null since `configCache` is empty.
I tried mocking up configCache and using `when(configCache.getIfPresent...` but it didn't work either.
I would appreciate any help here.

src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
```diff
@@ -0,0 +1,146 @@
...
+            CType.ROLES,
+            SecurityDynamicConfiguration.fromJson(objectMapper.writeValueAsString(objectNode), CType.ROLES, 2, 1, 1)
+        );
+
+        SecurityDynamicConfiguration<?> config = configRepository.getConfiguration(CType.ROLES);
+        SecurityDynamicConfiguration<?> deepCloneConfig = configRepository.getConfiguration(CType.ROLES);
+
+        assertThat(config, is(notNullValue()));
+        assertThat(config, instanceOf(SecurityDynamicConfiguration.class));
+        // assertThat(config.getCEntries().size(), is(greaterThan(0)));
```
DarshitChanpura, line 143: nit: can remove this commented line

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
```diff
@@ -11,65 +11,98 @@
...
-        Assert.assertEquals(1, selfRefreshingKeySet.getQueuedGetCount());
+    @Test
+    public void getKeyAfterRefresh_withNullKidShouldThrowBadCredentialsException() throws AuthenticatorUnavailableException,
+        BadCredentialsException {
 
+        Assert.assertThrows(BadCredentialsException.class, () -> selfRefreshingKeySet.getKeyAfterRefresh(null));
+    }
+
+    @Test
+    public void getKeyAfterRefresh_withInvalidDataShouldReturnBadCredentialException() throws AuthenticatorUnavailableException,
```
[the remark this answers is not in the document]
davidosorno: `public JWK getKey(String kid) throws AuthenticatorUnavailableException, BadCredentialsException`
and
`public synchronized JWK getKeyAfterRefresh(String kid) throws AuthenticatorUnavailableException, BadCredentialsException`
@DarshitChanpura these methods are different, aren't they?
What I know is that when testing we must test all public methods, so that's what I was trying to do. 
Any guidance here to understand better Test concept?

src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
davidosorno: @DarshitChanpura  I let this comment with the purpose of showing where the test is failing according to this [comment](https://github.com/opensearch-project/security/pull/3509#discussion_r1383470751). I would appreciate any help here.

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
davidosorno: @peternied, while working on this code I found out that this test runs as synchronous code, so all of that code doesn't do much. I'll remove it since it doesn't have any insights at all :disappointed:.
Thanks a lot for the feedback, it helped me to understand better how asynchronous calls work, and also to discover that testing race condition is really, really hard.

peternied: > I'll remove it since it doesn't have any insights at all 😞. what do you think?

I'd rather remove it - that way no one is under the false impression that something is tested better than it really is.

> testing race condition is really, really hard.

🤣 Indeed!

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
```diff
@@ -11,65 +11,98 @@
...
+    public void getKeyAfterRefresh_withKidShouldReturnKey() throws AuthenticatorUnavailableException, BadCredentialsException {
+
+        OctetSequenceKey key = (OctetSequenceKey) selfRefreshingKeySet.getKeyAfterRefresh("kid/b");
+        assertThat(TestJwk.OCT_2_K, is(equalTo(key.getKeyValue().decodeToString())));
+    }
 
-        while (selfRefreshingKeySet.getQueuedGetCount() == 0) {
-            Thread.sleep(10);
+    @Test
+    public void getKeyAfterRefresh_queuedGetCountVariableShouldBeZeroWhenFinishWithAllKeyRefreshes() throws InterruptedException,
```
peternied, line 70: Looks like this test was one of the ones that's reporting failures based on how fast the threads process in the GHA vs a local machine.

What I'd recommend is create a SelfRefreshingKeySet with your own KeySetProvider that you can control how quickly before there is a response.  This way, you can have it sit and wait, make some test assertions, then have it return a key when you are ready and test different assertions when it should be done.  This will remove the 'random' behavior because you'll know exactly when the KeySetProvider is refreshing vs queuing vs completed.

peternied:
@davidosorno Looks like there are some conflicts with recently merged changes, could you address those?  Also please let the maintainers and reviewers know if you are still working on something or awaiting for another round of reviews.

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
davidosorno: Another thing is, this failed could prove that the code is not executed synchronously as I mentioned in this [comment](https://github.com/opensearch-project/security/pull/3509#discussion_r1383685122); If so, it would have been 10 when it reached the assertion, am I wrong?

peternied: Currently the tests use `MockKeySetProvider`, which immediately returns a JWK.  This is not very useful for this async testing - there is no time for anything interesting to happen, instead create a new class like the following that you can control if it waits or doesn't wait.  By using a countdownlatch you can control when to allow execution to continue or stop.  There are more advanced concurrency privatives that you can use for more advanced conditions, I'd recommend starting with [[java.util.concurrency.CountDownLatch]](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/concurrent/CountDownLatch.html).

```java
public class LatchedTestKeySetProvider implements KeySetProvider {

    private final CountDownLatch latch;
    public LatchTestKeySetProvider(CountDownLatch latch) {
        this.latch = latch;
    }

    @Override
    public JWKSet get() throws AuthenticatorUnavailableException {
        try {
            // Wait until the latch is counted down
            latch.await();
            return TestJwk.OCT_1_2_3
        } catch (final InterruptedException e) {
            throw new AuthenticatorUnavailableException("Thread interrupted", e);
        }
    }
}
```

davidosorno: Perfect, I'll try it. Thanks for the guidance.

davidosorno:
@peternied any tips on this line in ConfigurationRepositoryTest
`SecurityDynamicConfiguration<?> conf = configCache.getIfPresent(configurationType);` always returns null since `configCache` is empty.
I tried mocking up configCache and using `when(configCache.getIfPresent...` but it didn't work either.

src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
peternied: The issue here is that the cache you've created isn't connected to the cache used internally be the ConfigurationResponsitory - there is no access to it.  Rather than attempt to push that functionality in, I'd recommend wrapping `configRepository` with a spy, and then substituting your own invocation for a function call that would populate the cache, maybe `getConfigurationsFromIndex(...)` would be a good choice.

davidosorno: Thanks for the tip @peternied. I'll try.

peternied:
@davidosorno did you mean to close out this PR?

davidosorno:
> @davidosorno did you mean to close out this PR?

@peternied Sorry, my bad. I didn't have too much time to work on it last week, so I thought it was better to close it until I had something.

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
davidosorno: This is why I learned working with CountDownLatch.
@peternied I created the class using the idea you suggested, then I had the problem that the `latch.await()` didn't make the main Thread to wait, so I got random responses every time I run the test.
Then I found out that there is another thread created inside the synchronized method `private synchronized JWK performRefresh(String kid) {` so, my threads (using latch) didn't wait for that thread to finish and that was the root cause of my random responses, since sometimes this new thread was slow, and sometimes was fast, so my assertions didn't work for the answers I was looking for.
I decided to create a test that just make a lot of asynchronous calls to the method, so the refreshCount and the queueCount get increased. I couldn't make it work with specific numbers because of the random responses I got, so I went for greaterThan 0.
I hope I explained myself, if not, please let me know.

src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
davidosorno: I decided to test the `deepClone` method directly, so I remove the test from the ConfigRepository, and now both worked.
is this a good approach?
should I add more tests to `SecurityDynamicConfiguration<T>` class?  I didn't find any tests for this class, so I would like to know if it's worth it.
Please let me know your thoughts. @peternied

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
davidosorno: Also, when running tests I got randomly errors for this line where `queuedGetCount` is always 0 even thought you use CountDownLatch.
https://github.com/opensearch-project/security/pull/3509/files#diff-e852cf3e1cb5c1b2226df1b2b3cea2a06ba861823c634c52ebd074ffd41cd727L62

image

peternied:
We do have metrics that track Open PRs and Ages of PRs - and I appreciation your consideration of those metrics; however, I those are guide to help us understand the healthy as a project.  Feel free to be selfish about what you are doing.

If you aren't making any headway feel free to leave it alone, or close it as you prefer - just wanted to make sure to know if you were abandoning the effort vs delayed.

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
```diff
@@ -11,66 +11,91 @@
...
+            // task to complete before continuing the loop, so queuedGetCount would not have pending tasks.
+            executor.execute(() -> {
+                try {
+                    int indexKey = (int) (Math.random() * 2);
+                    String keyToCompare = indexKey == 0 ? keyForKidA : keyForKidB;
+                    OctetSequenceKey key = (OctetSequenceKey) selfRefreshingKeySet.getKeyAfterRefresh(keys[indexKey]);
+
+                    assertThat(key, is(notNullValue()));
+                    assertThat(keyToCompare, is(equalTo(key.getKeyValue().decodeToString())));
+                } catch (AuthenticatorUnavailableException e) {} catch (BadCredentialsException e) {}
```
DarshitChanpura, line 83: we should add fail statements here

catch (AuthenticatorUnavailableException | BadCredentialsException e) {
    fail("No exception was expected but found: " + e.getMessage());
}

peternied (approved):

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
peternied: Did you manage to get to the bottom of this?

davidosorno: This line [`Future<?> future = threadPoolExecutor.submit(new Runnable() {`](https://github.com/opensearch-project/security/blob/07833294e750be097f75c8ccd7d40b882147112f/src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySet.java#L208)  is what makes the tests to get random responses, because the method is synchronized but it creates another thread inside, so it doesn't wait for this new thread to finish its job.

Here is a picture of the random calls made to each thread, and the test failed because sometimes it doesn't reach the point to count before the test ends. As you can see in the picture there is a number `9` when expected to be `10`, and I got a different values every time I run it.  

I removed the line I mentioned, then when I run the test I got good responses., they were the same every time, so I think this is something that needs to be investigated further.

Screenshot from 2023-12-07 09-08-56

peternied: I'd like to see the majority of this change merged - what do you think about creating a specific follow up issue and removing anything that doesn't behave as you'd expect so we can merge?

scrawfor99 (approved):
Great job @davidosorno Looks good to me!

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/SelfRefreshingKeySetTest.java
peternied: @davidosorno Please follow up if you feel there is anything else lingering I'm going to resolve this comment and merge.

PR #3388, opened 2023-09-21
[Backport 2.x] Integration tests framework

peternied:
### Description
Test framework that allows for declaring within the test class all
related users, roles, and other security primitives.  Emphasis on
hamcrest Matchers for straight forward to read, author, and
troubleshoot test cases.

#### Special Note
This is a _giant_ pull request - I would highly recommend using the github.dev [[view]](https://github.com/opensearch-project/security/pull/3388) for reviewing and providing feedback.  I'll also provide comments for the initial reviewers where I call out any deviations I've taken from what is in main

### Issues Resolved
- Resolves https://github.com/opensearch-project/security/issues/3230

### Check List
- [X] New functionality includes testing
- [ ] New functionality has been documented
- [X] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3512, opened 2023-10-09
Forward port integration tests

peternied:
### Description
Forward ports integration test changes from 2.11 and 2.x branches

### Issues Resolved
- Related https://github.com/opensearch-project/security/pull/3388

### Check List


cwperks:
@peternied The version increment PR in this repo is currently blocked because the PR to add the extension points for the header verifier has not been merged into 2.x yet: https://github.com/opensearch-project/OpenSearch/pull/10442

.github/workflows/ci.yml
```diff
@@ -87,8 +87,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        jdk: [17]
-        platform: [ubuntu-latest, windows-latest]
+        jdk: [11, 17]
+        platform: [ubuntu-latest] # Removing windows temporarily
```
DarshitChanpura, line 91: qq, do we know why did we not run integ-tests with 11 before this?

src/integrationTest/java/org/opensearch/security/http/BasicAuthWithoutChallengeTests.java
```diff
@@ -10,7 +10,7 @@
 package org.opensearch.security.http;
 
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
-import org.apache.hc.core5.http.HttpHeaders;
```
DarshitChanpura, line 13: This change should not be ported to main.

DarshitChanpura (changes requested):
This change looks good. I have request changes related to HTTPClient 4 as the main is already on hc 5.

src/integrationTest/java/org/opensearch/security/http/BasicAuthWithoutChallengeTests.java
peternied: This was intentionally - I think we should put this into main.

HttpHeaders are not specific to the the Apache Client5 - they still exist in a common namespace, so this reduces the total number of differences between the 2.x and 3.x branches.  As a future update - I think we'd want to abstract the way 'headers' are declared so only the test client file is different whereas the tests never have conflicts on imports - but that is out of scope for this change.

build.gradle
```diff
@@ -432,6 +432,14 @@ configurations {
             force "com.github.luben:zstd-jni:${versions.zstd}"
             force "org.xerial.snappy:snappy-java:1.1.10.5"
             force "com.google.guava:guava:${guava_version}"
+
+            // TODO: Seems like this should be removable
```
peternied, line 436: I will remove this comment - we need thee changes.

.github/workflows/ci.yml
peternied: We never used them reliably before - notice that `continue-on-error:true` a couple lines down this prevent the tests from surfacing failures until they were enabled on the 2.x line.

RyanL1997 (approved):

Issue #3464, opened 2023-10-05
[BUG]


Gaganjuneja:
**What is the bug?**
Possibly due to race condition windows backward compatibility test is failing - 

� ERROR][o.o.b.OpenSearchUncaughtExceptionHandler] [securityBwcCluster1-0] uncaught exception in thread [main]
�  org.opensearch.bootstrap.StartupException: java.lang.IllegalStateException: Unsupported transport.type [org.opensearch.security.ssl.http.netty.SecuritySSLNettyTransport]
�  	at org.opensearch.bootstrap.OpenSearch.init(OpenSearch.java:184) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
�  	at org.opensearch.bootstrap.OpenSearch.execute(OpenSearch.java:171) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
�  	at org.opensearch.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:104) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
�  	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138) ~[opensearch-cli-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
�  	at org.opensearch.cli.Command.main(Command.java:101) ~[opensearch-cli-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
�  	at org.opensearch.bootstrap.OpenSearch.main(OpenSearch.java:137) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
�  	at org.opensearch.bootstrap.OpenSearch.main(OpenSearch.java:103) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
�  Caused by: java.lang.IllegalStateException: Unsupported transport.type [org.opensearch.security.ssl.http.netty.SecuritySSLNettyTransport]
�  	at org.opensearch.common.network.NetworkModule.getTransportSupplier(NetworkModule.java:257) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
�  	at org.opensearch.node.Node.<init>(Node.java:897) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
[...]

PR #3514, opened 2023-10-10
Fix TraceableTcpTransportChannel classcast exception

### LINKED ISSUE #3515: [BUG]TraceableTcpTransportChannel class cast exception,
**What is the bug?**
When the experimental feature flag (opensearch.experimental.feature.telemetry.enabled) is set to true and the security plugin installed. There is a bootstrap error.

Caused by: org.opensearch.core.common.io.stream.NotSerializableExceptionWrapper: class_cast_exception: class org.opensearch.telemetry.tracing.channels.TraceableTcpTransportChannel cannot be cast to class org.opensearch.transport.TcpTransportChannel (org.opensearch.telemetry.tracing.channels.TraceableTcpTransportChannel and org.opensearch.transport.TcpTransportChannel are in unnamed module of loader 'app')
	at org.opensearch.security.ssl.transport.SecuritySSLRequestHandler.messageReceived(SecuritySSLRequestHandler.java:116) ~[?:?]
	at org.opensearch.security.OpenSearchSecurityPlugin$6$1.messageReceived(OpenSearchSecurityPlugin.java:790) ~[?:?]
	at org.opensearch.transport.RequestHandlerRegistry.processMessageReceived(RequestHandlerRegistry.java:106) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
	at org.opensearch.transport.InboundHandler.handleRequest(InboundHandler.java:262) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
	at org.opensearch.transport.InboundHandler.messageReceived(InboundHandler.java:140) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
	at org.opensearch.transport.InboundHandler.inboundMessage(InboundHandler.java:123) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
	at org.opensearch.transport.TcpTransport.inboundMessage(TcpTransport.java:770) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
	at org.opensearch.transport.InboundPipeline.forwardFragments(InboundPipeline.java:175) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
	at org.opensearch.transport.InboundPipeline.doHandleBytes(InboundPipeline.java:150) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
	at org.opensearch.transport.InboundPipeline.handleBytes(InboundPipeline.java:115) ~[opensearch-3.0.0-SNAPSHOT.jar:3.0.0-SNAPSHOT]
	at org.opensearch.transport.netty4.Netty4MessageChannelHandler.channelRead(Netty4MessageChannelHandler.java:95) ~[?:?]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:442) ~[?:?]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420) ~[?:?]
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412) ~[?:?]
	at io.netty.handler.logging.LoggingHandler.channelRead(LoggingHandler.java:280) ~[?:?]

 - Plugins

Gaganjuneja:
### Description
Fix the TraceableTcpTransportChannel class cast exception when enabling the feature opensearch.experimental.feature.telemetry.enabled. Issue reported here - https://github.com/opensearch-project/security/issues/3464#issuecomment-1753641505
### Issues Resolved
#3515

Reference Chnage - https://github.com/opensearch-project/OpenSearch/blob/8e5e54e87f6f86a2425a5c767e208767ba3df14c/server/src/main/java/org/opensearch/transport/RequestHandlerRegistry.java#L101

### Testing

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
@@ -86,7 +86,6 @@ import org.opensearch.client.indices.ResizeResponse;
 import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.cluster.metadata.IndexTemplateMetadata;
 import org.opensearch.common.settings.Settings;
-import org.opensearch.common.util.FeatureFlags;
 import org.opensearch.index.query.BoolQueryBuilder;
 import org.opensearch.index.query.MatchQueryBuilder;
 import org.opensearch.index.query.QueryBuilders;
@@ -95,7 +94,6 @@ import org.opensearch.index.reindex.ReindexRequest;
 import org.opensearch.repositories.RepositoryMissingException;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.search.builder.SearchSourceBuilder;
-import org.opensearch.telemetry.TelemetrySettings;
 import org.opensearch.test.framework.AuditCompliance;
 import org.opensearch.test.framework.AuditConfiguration;
 import org.opensearch.test.framework.AuditFilters;
@@ -372,16 +370,6 @@ public class SearchOperationTest {
             UPDATE_DELETE_USER,
             USER_ALLOWED_TO_PERFORM_INDEX_OPERATIONS_ON_SELECTED_INDICES,
             USER_ALLOWED_TO_CREATE_INDEX
-        )
-        .nodeSettings(
-            Map.of(
-                FeatureFlags.TELEMETRY_SETTING.getKey(),
-                true,
-                TelemetrySettings.TRACER_FEATURE_ENABLED_SETTING.getKey(),
-                true,
-                TelemetrySettings.METRICS_FEATURE_ENABLED_SETTING.getKey(),
-                true
-            )
         )
         .audit(
             new AuditConfiguration(true).compliance(new AuditCompliance().enabled(true))
--- a/src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
@@ -374,7 +374,14 @@ public class SearchOperationTest {
             USER_ALLOWED_TO_CREATE_INDEX
         )
         .nodeSettings(
-            Map.of(FeatureFlags.TELEMETRY_SETTING.getKey(), true, TelemetrySettings.TRACER_FEATURE_ENABLED_SETTING.getKey(), true)
+            Map.of(
+                FeatureFlags.TELEMETRY_SETTING.getKey(),
+                true,
+                TelemetrySettings.TRACER_FEATURE_ENABLED_SETTING.getKey(),
+                true,
+                TelemetrySettings.METRICS_FEATURE_ENABLED_SETTING,
+                true
+            )
         )
         .audit(
             new AuditConfiguration(true).compliance(new AuditCompliance().enabled(true))
--- a/src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
@@ -379,7 +379,7 @@ public class SearchOperationTest {
                 true,
                 TelemetrySettings.TRACER_FEATURE_ENABLED_SETTING.getKey(),
                 true,
-                TelemetrySettings.METRICS_FEATURE_ENABLED_SETTING,
+                TelemetrySettings.METRICS_FEATURE_ENABLED_SETTING.getKey(),
                 true
             )
         )
--- a/src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
@@ -86,6 +86,7 @@ import org.opensearch.client.indices.ResizeResponse;
 import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.cluster.metadata.IndexTemplateMetadata;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.util.FeatureFlags;
 import org.opensearch.index.query.BoolQueryBuilder;
 import org.opensearch.index.query.MatchQueryBuilder;
 import org.opensearch.index.query.QueryBuilders;
@@ -94,6 +95,7 @@ import org.opensearch.index.reindex.ReindexRequest;
 import org.opensearch.repositories.RepositoryMissingException;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.telemetry.TelemetrySettings;
 import org.opensearch.test.framework.AuditCompliance;
 import org.opensearch.test.framework.AuditConfiguration;
 import org.opensearch.test.framework.AuditFilters;
@@ -370,6 +372,9 @@ public class SearchOperationTest {
             UPDATE_DELETE_USER,
             USER_ALLOWED_TO_PERFORM_INDEX_OPERATIONS_ON_SELECTED_INDICES,
             USER_ALLOWED_TO_CREATE_INDEX
+        )
+        .nodeSettings(
+            Map.of(FeatureFlags.TELEMETRY_SETTING.getKey(), true, TelemetrySettings.TRACER_FEATURE_ENABLED_SETTING.getKey(), true)
         )
         .audit(
             new AuditConfiguration(true).compliance(new AuditCompliance().enabled(true))
--- a/src/main/java/org/opensearch/security/ssl/transport/SecuritySSLRequestHandler.java
+++ b/src/main/java/org/opensearch/security/ssl/transport/SecuritySSLRequestHandler.java
@@ -37,9 +37,9 @@ import org.opensearch.security.ssl.util.SSLRequestHelper;
 import org.opensearch.security.support.ConfigConstants;
 import org.opensearch.tasks.Task;
 import org.opensearch.threadpool.ThreadPool;
+import org.opensearch.transport.BaseTcpTransportChannel;
 import org.opensearch.transport.TaskTransportChannel;
 import org.opensearch.transport.TcpChannel;
-import org.opensearch.transport.TcpTransportChannel;
 import org.opensearch.transport.TransportChannel;
 import org.opensearch.transport.TransportRequest;
 import org.opensearch.transport.TransportRequestHandler;
@@ -113,9 +113,9 @@ public class SecuritySSLRequestHandler<T extends TransportRequest> implements Tr
 
             if (channel instanceof TaskTransportChannel) {
                 final TransportChannel inner = ((TaskTransportChannel) channel).getChannel();
-                nettyChannel = (Netty4TcpChannel) ((TcpTransportChannel) inner).getChannel();
-            } else if (channel instanceof TcpTransportChannel) {
-                final TcpChannel inner = ((TcpTransportChannel) channel).getChannel();
+                nettyChannel = (Netty4TcpChannel) ((BaseTcpTransportChannel) inner).getChannel();
+            } else if (channel instanceof BaseTcpTransportChannel) {
+                final TcpChannel inner = ((BaseTcpTransportChannel) channel).getChannel();
                 nettyChannel = (Netty4TcpChannel) inner;
             } else {
                 throw new Exception("Invalid channel of type " + channel.getClass() + " (" + channel.getChannelType() + ")");
--- a/src/main/java/org/opensearch/security/ssl/transport/SecuritySSLRequestHandler.java
+++ b/src/main/java/org/opensearch/security/ssl/transport/SecuritySSLRequestHandler.java
@@ -37,13 +37,9 @@ import org.opensearch.security.ssl.util.SSLRequestHelper;
 import org.opensearch.security.support.ConfigConstants;
 import org.opensearch.tasks.Task;
 import org.opensearch.threadpool.ThreadPool;
-import org.opensearch.transport.BaseTcpTransportChannel;
-import org.opensearch.transport.TaskTransportChannel;
-import org.opensearch.transport.TcpChannel;
 import org.opensearch.transport.TransportChannel;
 import org.opensearch.transport.TransportRequest;
 import org.opensearch.transport.TransportRequestHandler;
-import org.opensearch.transport.netty4.Netty4TcpChannel;
 
 public class SecuritySSLRequestHandler<T extends TransportRequest> implements TransportRequestHandler<T> {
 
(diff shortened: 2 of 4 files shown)


Gaganjuneja:
@reta, @peternied Please take a look at it. We need to merge this fix in 2.11

src/main/java/org/opensearch/security/ssl/transport/SecuritySSLRequestHandler.java
```diff
@@ -113,9 +113,9 @@ public final void messageReceived(T request, TransportChannel channel, Task task
 
             if (channel instanceof TaskTransportChannel) {
                 final TransportChannel inner = ((TaskTransportChannel) channel).getChannel();
-                nettyChannel = (Netty4TcpChannel) ((TcpTransportChannel) inner).getChannel();
-            } else if (channel instanceof TcpTransportChannel) {
-                final TcpChannel inner = ((TcpTransportChannel) channel).getChannel();
+                nettyChannel = (Netty4TcpChannel) ((BaseTcpTransportChannel) inner).getChannel();
+            } else if (channel instanceof BaseTcpTransportChannel) {
+                final TcpChannel inner = ((BaseTcpTransportChannel) channel).getChannel();
```
peternied, line 118: Lets add a test(s) that confirms this fix is resolved.

If you are looking for an example of how these tests could be written, take a look at https://github.com/opensearch-project/security/blob/main/src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java that might provide some guidance on cluster configuration and a basic setup.

peternied (changes requested):
@Gaganjuneja Thanks for creating this PR - if this is breaking it seems like we are missing some tests for these scenarios.

Gaganjuneja:
> @Gaganjuneja There is another issue with the Telemetry experimental feature and the security plugin. The way that the TraceableHttpChannel wraps the underlying channel [here](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/http/AbstractHttpServerTransport.java#L367) is problematic. Security needed to recently refactor where auth was performed from the last step of the netty pipeline to further up the pipeline and needs access to the netty channel.
> 
> You can see an example here of how the security plugin would access the channel: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/filter/SecurityRestFilter.java#L135

@reta, This seems to be new changes which will be bypassed for TraceableHttpChannel always. I am thinking, if we can add a new method to the HttpChannel::getType and let all the implementations implement it.

```
default String getType() {
    return ChannelType.DEFAULT;
}
```

For Netty4

```
public String getType() {
    return ChannelType.NETTY4;
}
```

reta:
> @reta, This seems to be new changes which will be bypassed for TraceableHttpChannel always. I am thinking, if we can add a new method to the HttpChannel::getType and let all the implementations implement it.

@Gaganjuneja  I sadly cannot look at the issue today, travelling back home, but it does not look right - the HTTP transport should have no dependency on the engine behind it, if any, I will take a look tomorrow morning

Gaganjuneja:
> > @reta, This seems to be new changes which will be bypassed for TraceableHttpChannel always. I am thinking, if we can add a new method to the HttpChannel::getType and let all the implementations implement it.
> 
> @Gaganjuneja I sadly cannot look at the issue today, travelling back home, but it does not look right - the HTTP transport should have no dependency on the engine behind it, if any, I will take a look tomorrow morning

Sure, sorry for bothering you!

src/main/java/org/opensearch/security/ssl/transport/SecuritySSLRequestHandler.java
Gaganjuneja: @peternied, I have written a Integ test case and its failing for 2 more issues. Where RestRequest::getHttpChannel is being type casted to Netty4HttpChannel. At one place this is already been reported by @cwperks and in addition to that one check is here https://github.com/opensearch-project/security/blob/e69013651189238f3c93d57415d37134c8bfd6df/src/main/java/org/opensearch/security/filter/OpenSearchRequest.java#L47

Gaganjuneja:
> > @reta, This seems to be new changes which will be bypassed for TraceableHttpChannel always. I am thinking, if we can add a new method to the HttpChannel::getType and let all the implementations implement it.
> 
> @Gaganjuneja I sadly cannot look at the issue today, travelling back home, but it does not look right - the HTTP transport should have no dependency on the engine behind it, if any, I will take a look tomorrow morning

Other simplest option could be is to have TraceableHttpChannel::getDelegateChannel method and inside the security plugin we can put an additional check 

```
HttpChannel channel = restRequest.getChannel();
if(channel instanceof TraceableHttpChannel){
   channel = ((TraceableHttpChannel)channel).getDelegateChannel();
}

//Now the Netty4 check could proceed.
```

src/main/java/org/opensearch/security/ssl/transport/SecuritySSLRequestHandler.java
reta: @Gaganjuneja @peternied I see at least 2 issues with this code:
 - it very coupled with Netty, that should not be the case
 - the issue will come back when we would instrument `TaskTransportChannel`

Would I would like to suggest: add `get` (and possibly `put`, have to look into it) methods to `HttpChannel` and `TcpChannel`:

```
public T get(String attribute, Class<T> clazz);
```

With ytat, we could remove `|| !(underlyingRequest.getHttpChannel() instanceof Netty4HttpChannel)) {` from `OpenSearchRequest` since it will be `null` value anyway.

Gaganjuneja: Thanks @reta, Yes it would be more cleaner approach.

Gaganjuneja: There is one more change where we are trying to get the SSLHandler from the inbound and outbound pipeline

https://github.com/opensearch-project/security/blob/e69013651189238f3c93d57415d37134c8bfd6df/src/main/java/org/opensearch/security/filter/OpenSearchRequest.java#L54

reta: Ir does not matter, it is all covered, will send the pull request shortly

reta: https://github.com/opensearch-project/OpenSearch/pull/10562

Gaganjuneja: One clarification, Please pardon me for the naive question. How will we access the channel attributes (NettyAttribute::proFrom) for authentication? Or the old implementation should be suffice.

reta: > One clarification, Please pardon me for the naive question. How will we access the channel attributes (NettyAttribute::popFrom) for authentication? Or the old implementation should be suffice.

@Gaganjuneja chatted with Craig (@cwperks) offline, it seems like we may try to pursue a different path for early termination of the the request, I would suggest we do nothing for channel attributes (NettyAttribute::*) now since it is likely to go away (sadly no chances to fix that `2.11`). It should not be an issue, tracing is an experimental feature that is expected to have hiccups.

reta: @Gaganjuneja I've updated the change with respect to core changes

cwperks (approved):

willyborankin (approved):

peternied (commented):
Thanks - this change looks much cleaner.  Happy to re-review after an integration test has been added.

reta:
> Thanks - this change looks much cleaner. Happy to re-review after an integration test has been added.

@Gaganjuneja are you working on that?

Gaganjuneja:
> > Thanks - this change looks much cleaner. Happy to re-review after an integration test has been added.
> 
> @Gaganjuneja are you working on that?

@reta, I will pick this up very first thing next week.

Gaganjuneja:
> > > Thanks - this change looks much cleaner. Happy to re-review after an integration test has been added.
> > 
> > 
> > @Gaganjuneja are you working on that?
> 
> @reta, I will pick this up very first thing next week.

There is nothing we can test about telemetry. Only thing is to enable it and see if cluster is coming up fine as far as the integration with security plugin is concerned. I have enable it for SearchOperationTest.java and now it should be able to catch if anything is breaking due to change in the Telemetry side.

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
```diff
@@ -371,6 +373,16 @@ public class SearchOperationTest {
             USER_ALLOWED_TO_PERFORM_INDEX_OPERATIONS_ON_SELECTED_INDICES,
             USER_ALLOWED_TO_CREATE_INDEX
         )
+        .nodeSettings(
```
reta, line 376: This flag is noop if plugin is not installed, right? I think what we could do is to enhance (or add new) the GA which does full plugin install test to have security and opentelemetry plugin installed together, @peternied does it sound right to you?

Gaganjuneja: yes @reta you are right. but this would atl east create the Traceable wappers.

reta: > yes @reta you are right. but this would at least create the Traceable wrappers.

how? it will use NOOP tracer

Gaganjuneja: Oh, yes. After this recent change tracer:isTracingEnabled, it wont.

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
```diff
@@ -371,6 +373,9 @@ public class SearchOperationTest {
...
+            Map.of(FeatureFlags.TELEMETRY_SETTING.getKey(), true, TelemetrySettings.TRACER_FEATURE_ENABLED_SETTING.getKey(), true)
```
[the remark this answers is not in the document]
Gaganjuneja: @peternied, As mentioned by @reta these settings are not going to work and tracing would be noop in this scenario. In case of NoopTracer we tried to implement in a way that its shouldn't alter the execution pattern by unnecessarily introducing the tracer related core changes (Tracing being an experimental feature eg. is TraceableHttpChannel here. If tracer is noop we are not wrapping the HttpChannel hence its working with FeatureFlag disabled).

Now to test the telemetry plugin and security plugin we need to install them together and execute the tests.

To validate the telemetry working we need to have some search/indexing requests and validate if traces are being generated and which is not very straight forward.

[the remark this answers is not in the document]
peternied: >To validate the telemetry working we need to have some search/indexing requests and validate if traces are being generated and which is not very straight forward.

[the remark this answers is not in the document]
Gaganjuneja: How are we testing security with other plugins today?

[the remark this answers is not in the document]
Gaganjuneja: One more thing, Even if we write the integration test, it's still going to fail bcauase the issue is partially fixed and the check inside NettyAttribute.java is still there - 

```
public static <T> Optional<T> popFrom(final RestRequest request, final AttributeKey<T> attribute) {
        if (request.getHttpChannel() instanceof Netty4HttpChannel) {
            Channel nettyChannel = ((Netty4HttpChannel) request.getHttpChannel()).getNettyChannel();
            return Optional.ofNullable(nettyChannel.attr(attribute).getAndSet(null));
        }
        return Optional.empty();
    }
```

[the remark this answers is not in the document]
reta: Ah, correct, this part is on me, thanks for reminding @Gaganjuneja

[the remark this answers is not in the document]
cwperks: @Gaganjuneja You can see an example of how anomaly-detection tests with security here: https://github.com/opensearch-project/anomaly-detection/blob/main/.github/workflows/test_security.yml

They run their full integ test suite in a cluster w/ the security plugin installed.

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
peternied: I've moved this flag to its own class - but I was not able to reproduce a failure with the other changes in this CR reverted.  

Please pull the latest changes from your fork and test locally until the test can fail if you'd reverted the other changes - its possible I've got the feature flag configured incorrectly.

src/main/java/org/opensearch/security/ssl/transport/SecuritySSLRequestHandler.java
```diff
@@ -108,21 +104,7 @@ public final void messageReceived(T request, TransportChannel channel, Task task
         }
 
         try {
-
-            Netty4TcpChannel nettyChannel = null;
-
```
DarshitChanpura, line 113: curious to understand why these changes were made?

reta: To get rid of type casting

DarshitChanpura:
@Gaganjuneja would you need any helping getting this across?

DarshitChanpura:
@Gaganjuneja are you still working on this?

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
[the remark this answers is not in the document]
Gaganjuneja: > One more thing, Even if we write the integration test, it's still going to fail bcauase the issue is partially fixed and the check inside NettyAttribute.java is still there -
> 
> ```
> public static <T> Optional<T> popFrom(final RestRequest request, final AttributeKey<T> attribute) {
>         if (request.getHttpChannel() instanceof Netty4HttpChannel) {
>             Channel nettyChannel = ((Netty4HttpChannel) request.getHttpChannel()).getNettyChannel();
>             return Optional.ofNullable(nettyChannel.attr(attribute).getAndSet(null));
>         }
>         return Optional.empty();
>     }
> ```



> Ah, correct, this part is on me, thanks for reminding @Gaganjuneja

@reta, What is your opinion, this should be addressed here on in another PR. 

@DarshitChanpura, Once the above issue is addressed, I will write the integration test.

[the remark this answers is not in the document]
reta: > @reta, What is your opinion, this should be addressed here on in another PR.

Up to you @Gaganjuneja

peternied:
Doesn't look like this is being worked on anymore, going to close this out. Please feel free to reopen if you re-engage on this @Gaganjuneja

Issue #3176, opened 2023-08-14
Issue and ferry a Service Account Token to an Extension on bootstrap


cwperks:
As part of [[META] System Indices access for Extensions](https://github.com/opensearch-project/security/issues/2529), the plan is for an extension that reserves indices to receive a token on bootstrap that the extension can utilize to interact with its reserved indices. This token will be referred to as a Service Account Token because it represents the identity of the extension and can be used by the extension to act as itself when making requests on the extension's reserved indices.

I'm filing an issue to track the work to create an interface within the [`IdentityPlugin` extension point](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/plugins/IdentityPlugin.java) of core that can be used to trigger the security plugin's implementation of issuing a service account token. After issuance, the token also needs to be ferried to an extension.

I see 2 possible options for ferrying this token to an extension:

1. As part of the [`InitializeExtensionRequest`](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/extensions/ExtensionsManager.java#L398-L407) which is initiated from core and used to initialize an extension. The handler on the SDK side can be found [here](https://github.com/opensearch-project/opensearch-sdk-java/blob/main/src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java#L48-L96).

2. As a separate Transport action that is sent after successful initialization. I recommend this option. 

You can see a sample implementation of a separate transport action for ferrying service account tokens in [here](https://github.com/cwperks/OpenSearch/pull/94/files#diff-c9a4accc046eb92a0e6a025970b3eacacc330f40bb55b02859df05a79710c8c8R416). (There's an open PR on the SDK repo to address token handling on the SDK side that provided implementations of ferrying a service account token and obo token: https://github.com/opensearch-project/opensearch-sdk-java/pull/892)

Sending this as a separate Transport Request potentially allows the transport action to be re-used for token rotation, whereas an initialization request is only meant to be performed once. 

Acceptance Criteria:

- [x]  Define an interface in core on the IdentityPlugin that can be called to issue a service account token given an extension's unique ID.
- [x] Create a new TransportAction that runs after successful InitializeExtensionRequest to issue and ferry the service account token to an extension
- [ ] Implement the interface in the Security plugin

Future followup:
[...]

PR #3538, opened 2023-10-13
Implement IdentityPlugin in Security plugin

scrawfor99:
### Description
This change implements the IdentityPlugin interface inside the Security Plugin codebase. 
This change adds the methods `getSubject` and `getTokenManager`. It also adds a new class `SecurityTokenManager` which handles the issuance of Service Account tokens and OBO tokens. This is done by having the token manager wrap the appropriate methods inside the `UserService` (for service accounts) and `JwtVendor` (for OBO tokens). 

### Issues Resolved
This issue addresses the final check box for: https://github.com/opensearch-project/security/issues/3176

### Testing
This change adds tests for issuing the tokens using the SecurityTokenManager. The other functionality has already been tested by the supporting classes. 

### Check List


src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -152,8 +152,8 @@ public void accept(RestChannel channel) throws Exception {
                         user.getName(),
                         service,
                         tokenDuration,
-                        mappedRoles.stream().collect(Collectors.toList()),
-                        user.getRoles().stream().collect(Collectors.toList()),
+                        mappedRoles.stream().collect(Collectors.toSet()),
```
scrawfor99, line 155: Had to swap to set in order to avoid needing to change other classes more extensively.

src/main/java/org/opensearch/security/identity/SecuritySubject.java
```diff
@@ -0,0 +1,48 @@
...
+		}
+		final User user = (User) threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+		if (user == null) {
+			return NamedPrincipal.UNAUTHENTICATED;
+		}
+		return new NamedPrincipal(user.getName());
+	}
+
+	@Override
+	public void authenticate(AuthToken authToken) {
```
scrawfor99, line 45: Not sure how to handle this

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
willyborankin: You can use `Set.copyOf()` instead of  `.strem().collect()`

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -152,8 +152,8 @@ public void accept(RestChannel channel) throws Exception {
...
+                        user.getRoles().stream().collect(Collectors.toSet()),
```
willyborankin, line 156: Same `Set.copyOf()`

src/main/java/org/opensearch/security/identity/SecuritySubject.java
```diff
@@ -0,0 +1,48 @@
...
+	public void setThreadContext(ThreadContext threadContext) {
+		this.threadContext = threadContext;
+	}
+
+	@Override
+	public Principal getPrincipal() {
+		if (threadContext == null) {
+			return NamedPrincipal.UNAUTHENTICATED;
+		}
+		final User user = (User) threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
```
willyborankin, line 37: `getTransient` is a generic method no need to explicitly convert it to `User`

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,97 @@
...
+		this.cs = cs;
+		this.threadPool = threadPool;
+		this.userService = userService;
+	}
+
+	private JwtVendor jwtVendor = new JwtVendor(DEMO_SETTINGS, Optional.empty());
+
+	@Override
+	public AuthToken issueOnBehalfOfToken(Subject subject, OnBehalfOfClaims claims) {
+		User user = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
```
willyborankin, line 56: wdyt to use `Optional.ofNullable().map().orElseGet(() -> thorw)`???

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
scrawfor99: Will do!

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
scrawfor99: Will do!

src/main/java/org/opensearch/security/identity/SecuritySubject.java
scrawfor99: Oh let me fix that.

scrawfor99:
Blocked by: https://github.com/opensearch-project/OpenSearch/pull/10614

src/main/java/org/opensearch/security/identity/SecuritySubject.java
scrawfor99: @peternied With regard to our other chat about the level where this authenticate method should be, what did you have in mind for this? 

I am not sure how we are planning on handling this call but it is basically dead for the time being. I would like this to inject the new user into the thread context based on the resolution of the auth credentials. 

The issue is that the tokens should not be hitting this. I would expect authc/z to take place with the normal mechanisms for handling basic auth & jwts. In turn, there should not be any difference to the security plugin whether a on behalf of token or service account token are being used vs. a normal request with a Basic or Bearer auth header. 

So I will need to make the injection via this method standard for all requests not just those involving OBO or Service Accounts.

src/main/java/org/opensearch/security/identity/SecuritySubject.java
```diff
@@ -0,0 +1,51 @@
...
+    public SecuritySubject() {}
+
+    public void setThreadContext(ThreadContext threadContext) {
+        this.threadContext = threadContext;
+    }
+
+    @Override
+    public Principal getPrincipal() {
+        if (threadContext == null) {
+            System.out.println("Thread context is null");
```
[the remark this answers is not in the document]
willyborankin: There are failures for the checkstyle:
```
Error: eckstyle] [ERROR] /home/runner/work/security/security/src/main/java/org/opensearch/security/identity/SecuritySubject.java:35: Do not use System.out.println [RegexpSinglelineJava]
Error: eckstyle] [ERROR] /home/runner/work/security/security/src/main/java/org/opensearch/security/identity/SecuritySubject.java:40: Do not use System.out.println [RegexpSinglelineJava]
Error: eckstyle] [ERROR] /home/runner/work/security/security/src/main/java/org/opensearch/security/identity/SecuritySubject.java:43: Do not use System.out.println [RegexpSinglelineJava]
```

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,99 @@
...
+
+        Set<String> mappedRoles = mapRoles(user, caller);
+
+        return userOptional.map(u -> {
+            try {
+                return jwtVendor.createJwt(
+                    cs.getClusterName().value(),
+                    u.getName(),
+                    claims.getAudience(),
+                    300,
```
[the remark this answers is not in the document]
scrawfor99: I can take care of it. This is stuff I dragged from the previous changes around JWT stuff.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -109,8 +106,8 @@ public String createJwt(
         String subject,
         String audience,
         Integer expirySeconds,
-        List<String> roles,
-        List<String> backendRoles,
+        Set<String> roles,
+        Set<String> backendRoles,
         boolean roleSecurityMode
     ) throws Exception {
```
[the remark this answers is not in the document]
scrawfor99: Yeah no problem

src/main/java/org/opensearch/security/identity/SecuritySubject.java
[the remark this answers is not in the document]
scrawfor99: Woops

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,99 @@
...
+            "signing_key",
+            Base64.getEncoder()
+                .encodeToString(
+                    "This is my super secret that no one in the universe will ever be able to guess in a bajillion years".getBytes(
+                        StandardCharsets.UTF_8
+                    )
+                )
+        )
+        .put("encryption_key", Base64.getEncoder().encodeToString("encryptionKey".getBytes(StandardCharsets.UTF_8)))
+        .build();
```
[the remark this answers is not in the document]
scrawfor99: Sounds good

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,99 @@
...
+        try {
+            return new BasicAuthToken(this.userService.generateAuthToken(extensionUniqueId));
+        } catch (Exception e) {
+            throw new OpenSearchSecurityException(String.valueOf(e));
+        }
+    }
+
+    public Set<String> mapRoles(final User user, final TransportAddress caller) {
+        return this.configModel.mapSecurityRoles(user, caller);
+    }
```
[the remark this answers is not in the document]
scrawfor99: I  just grabbed this from the JWT stuff and figured I should not mess with it.

If I am fine to make more significant changes to it then I will go ahead and do that.

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,99 @@
...
+                    claims.getAudience(),
+                    300,
+                    mappedRoles,
+                    u.getRoles(),
+                    false
+                );
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        }).map(BearerAuthToken::new).orElseThrow(() -> new OpenSearchSecurityException("JWT creation failed."));
```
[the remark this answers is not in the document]
scrawfor99: Instead of mapping you mean? 

I made this swap at a previous suggestion but can go back

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public AuthToken issueServiceAccountToken(String extensionUniqueId) throws OpenSearchSecurityException {
+        try {
+            return new BasicAuthToken(this.userService.generateAuthToken(extensionUniqueId));
+        } catch (Exception e) {
+            throw new OpenSearchSecurityException(String.valueOf(e));
+        }
```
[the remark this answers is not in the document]
scrawfor99: Let me take a look

scrawfor99:
Need to make this change: https://github.com/opensearch-project/OpenSearch/pull/10664

We don't want people to have to provide absolute times in JSON form. Right now we are running into an issue like this where the calculated time expects a Second value to be provided in the case of a JSON is used to request the token settings. But then it auto-propagates the fields with a millisecond absolute time value.

We want our times to be in similar magnitude not way out of scale like this: 
Max expiry time is: 1697562172
Provided expiry time is: 1697561872

src/main/java/org/opensearch/security/identity/SecuritySubject.java
peternied: Lets no-op this for the time being, we still haven't fully aligned what should happen where inside of core vs plugins as far as authentication goes.

src/main/java/org/opensearch/security/identity/SecuritySubject.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.security.Principal;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.identity.NamedPrincipal;
+import org.opensearch.identity.Subject;
+import org.opensearch.identity.tokens.AuthToken;
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.security.user.User;
+
+public class SecuritySubject implements Subject {
```
peternied, line 22: Lets return a new instance of the NoopSubject for the time being, we are close, but not close enough to declare how this should really look IMO.

scrawfor99:
I want to make the Security Plugin instantiate its token manager on bootstrap i.e. put this in the `OpenSearchSecurityPlugin` class: `tokenManager = new SecurityTokenManager(cs, threadPool, userService, null, settings);`

However, we run into issues since the OnBehalfOf token logic requires us to have:

```

    /**
     * The constructor for the SecurityTokenManager
     * @param cs The cluster service for the token manager to use
     * @param threadPool The thread pool for the token manager to use
     * @param userService The global instance of the user service that should be used
     * @param timeProvider An optional time provider that yields the current time in SECONDS
     * @param settings Any settings. It should always include those used for creating JWTs
     */
    public SecurityTokenManager(ClusterService cs, ThreadPool threadPool, UserService userService, final Optional<LongSupplier> timeProvider, final Settings settings) {
        this.cs = cs;
        this.threadPool = threadPool;
        this.userService = userService;
        setKeySettings(timeProvider, settings);
    }

    /**
     * This method allows for configuration of the time provider and settings values
     * This should only be used for testing
     * @param timeProvider An optional time provider. Any provided should be in SECONDS
     * @param settings Any settings to use for creating the JWTs
     */
    public void setKeySettings(final Optional<LongSupplier> timeProvider, final Settings settings) {
        JoseJwtProducer jwtProducer = new JoseJwtProducer();
        try {
            this.signingKey = createJwkFromSettings(settings);
        } catch (Exception e) {
            throw ExceptionUtils.createJwkCreationException(e);
        }
        this.jwtProducer = jwtProducer;
        if (isKeyNull(settings, "encryption_key")) {
            throw new IllegalArgumentException("encryption_key cannot be null");
        } else {
            this.claimsEncryptionKey = settings.get("encryption_key");
            this.encryptionDecryptionUtil = new EncryptionDecryptionUtil(claimsEncryptionKey);
        }
        if (timeProvider != null && timeProvider.isPresent()) {  // Check if timeProvider is not null and then if it's present
            this.timeProvider = timeProvider.get();
        } else {
            this.timeProvider = () -> System.currentTimeMillis() / 1000;
        }
    }
```

Which includes a `signing_key`... This is going to be the same type of problem we are trying to fix with `admin:admin`. I don't want to put a default `signing_key` but don't know what alternative we reasonably have.. 

@RyanL1997  @peternied @davidlago thoughts?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1035,6 +1043,8 @@ public Collection<Object> createComponents(
 
         final XFFResolver xffResolver = new XFFResolver(threadPool);
         backendRegistry = new BackendRegistry(settings, adminDns, xffResolver, auditLog, threadPool);
+        subject.setThreadContext(threadPool.getThreadContext());
```
cwperks, line 1046: This looks like an implementation detail of a Subject in this repo. Would it make sense to make this a constructor arg?

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -1,171 +0,0 @@
-/*
```
cwperks, line 1: Why delete the JWTVendor? Can the token manager use an instance of a JWTVendor to issue tokens for its needs? I think it makes sense to have a generic, reusable JwtVendor.

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,261 @@
...
+    public String createJwt(
+            String issuer,
+            String subject,
+            String audience,
+            Long expirySeconds,
+            Set<String> roles,
+            Set<String> backendRoles,
+            boolean roleSecurityMode
+    ) {
+        final long nowAsSecs = timeProvider.getAsLong();
```
cwperks, line 207: This may collide with the https://github.com/opensearch-project/security/pull/3421 PR where its being switched to epoch millis. We will need to stick to one convention to avoid collisions. Personally, I think seconds are a bit easier to reason with then millis but we shouldn't have a mix in different places.

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,261 @@
+package org.opensearch.security.identity;
```
cwperks, line 1: The SecurityTokenManager should use an instance of a JWTVendor to issue JWTs. There are now multiple places where this repo issues JWTs and in each instance they would use a different JWK (signing key) when signing the JWT. 

1. OBO and Service Account Tokens for extensions would use a JWTVendor to issue a JWT.
2. The SAML auth flow could be refactored to utilize a JWTVendor.

cwperks (commented):
Thank you @scrawfor99 . I took a first pass at the PR and left a few comments.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
scrawfor99: Peter's request: https://github.com/opensearch-project/security/pull/3538#discussion_r1362149896

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
scrawfor99: Using milliseconds is less than ideal because of how the expiration time is handled. 

If you look at the way the OnBehalfOf token requests are set up, we want people to be able to provide an expiration time that makes sense. Unfortunately, no one is going to want to provide an expiration time of 179XXXXX. So I swapped this over to be reasonable for configuration. 

I agree we will have to make sure everything is compatible.

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
scrawfor99: Peter's request: https://github.com/opensearch-project/security/pull/3538#discussion_r1362149896

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
peternied: > Can we remove the JwtVendor and replace it with the SecurityTokenManager?

My intention was to removed the JwtVendor from being passed to the Handler and replaced with an instance of the SecurityTokenManager.  That said, I don't mind if we delete the JwtVendor and replace it with the SecurityTokenManager, thats more about naming which I'm indifferent to in this case.

I do have some recommended requirements if you'd like to consider them @scrawfor99:
- Generation and validation for JWTs used by OBO should be the same instance in memory.  There should be only one place where this component is 'new'ed' up, and that single instance is somehow shared everywhere it is needed.
- Dynamic configuration changes can alter the secrets associated with JWTs, this should be allowed.
- The SecurityTokenManager should be able to communicate that it was unable generate a token with an exception.

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
RyanL1997: Hi @scrawfor99 @cwperks, if the seconds is the case, we can always *1000 in the #3421 right?

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -87,15 +91,18 @@ public void onDynamicConfigModelChanged(DynamicConfigModel dcm) {
...
         }
     }
 
-    public CreateOnBehalfOfTokenAction(final Settings settings, final ThreadPool threadPool, final ClusterService clusterService) {
+    public CreateOnBehalfOfTokenAction(final Settings settings, final ThreadPool threadPool, final ClusterService clusterService, Optional<LongSupplier> longSupplier, UserService userService) {
         this.threadPool = threadPool;
         this.clusterService = clusterService;
+        this.userService = userService;
+        this.settings = settings;
+        this.longSupplier = longSupplier;
```
DarshitChanpura, line 105: is this variable being used?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: Why duplicated the whole thread context here?

src/test/java/org/opensearch/security/identity/SecurityTokenManagerTest.java
```diff
@@ -1,16 +1,11 @@
...
- * The OpenSearch Contributors require contributions made to
- * this file be licensed under the Apache-2.0 license or a
- * compatible open source license.
- *
- * Modifications Copyright OpenSearch Contributors. See
- * GitHub history for details.
- */
-
-package org.opensearch.security.authtoken.jwt;
+package org.opensearch.security.identity;
```
DarshitChanpura, line 1: nit: license header

peternied:
I was think to keep reusing this PR, there has been good discussion here I wouldn't want to drop, @scrawfor99 does that work for you?

scrawfor99:
Hi @peternied, my bad I misunderstood. That is just fine.

peternied:
Finished with the architectural changes, `./gradlew integrationTest --tests org.opensearch.security.http.OnBehalfOfJwtAuthenticationTest` is passing again.

Next step, will patch up test scenarios and make sure config update scenario is correctly tested, afterward will take out of draft.  Feel free to review in the interim, or wait till later.

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
peternied: SecurityTokenManager uses an instance of JwtVendor again

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
peternied: I've made this change

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
peternied: I've swapped this for the NoopSubject for the moment, until we are using it lets not insert anymore context than necessary into this the IdentityPlugin implementation

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,136 @@
...
+    public void onConfigModelChanged(final ConfigModel configModel) {
+        this.configModel = configModel;
+    }
+
+    @Subscribe
+    public void onDynamicConfigModelChanged(final DynamicConfigModel dcm) {
+        final Settings oboSettings = dcm.getDynamicOnBehalfOfSettings();
+        final Boolean enabled = oboSettings.getAsBoolean("enabled", false);
+        if (enabled) {
+            jwtVendor = new JwtVendor(oboSettings, Optional.empty());
```
cwperks, line 75: What happens if an error is thrown here?

src/main/java/org/opensearch/security/authtoken/jwt/ExpiringBearerAuthToken.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public String getSubject() {
+        return subject;
+    }
+
+    public Date getExpiry() {
+        return expiry;
+    }
+
+    public long getExpiresInSeconds() {
```
DarshitChanpura, line 37: qq, what is the diff between `expiry` and `expiresInSeconds`?

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,136 @@
...
+    private boolean oboNotSupported() {
+        return jwtVendor == null;
+    }
+
+    @Override
+    public ExpiringBearerAuthToken issueOnBehalfOfToken(final Subject subject, final OnBehalfOfClaims claims) {
+        if (oboNotSupported()) {
+            // TODO: link that doc!
+            throw new OpenSearchSecurityException(
+                "The OnBehalfOf token generate is not enabled, see {link to doc} for more information on this feature."
```
DarshitChanpura, line 90: nit: generation (probably a relic from the original PR)

src/main/java/org/opensearch/security/authtoken/jwt/ExpiringBearerAuthToken.java
peternied: One in the date time of the expiration, one of them is the time in seconds.

Its a little strange to only provide the expiry, which is the value on the token - because when you passed the parameter you were indicating the value in seconds - however, if you are downstream client you'd want to know when you should stop using the token so you'd want the computed DateTime value as well.  Hence the two values.  This is an internal only representation, we can iterate on what the different interfaces are that are exposed at OpenSearch to make this clearer for consumers.

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
peternied: Good call, I'll try/catch this to make sure it doesn't bleed out crashing something unexpected, and log the error.

peternied (approved):

peternied:
Thanks @cwperks & @DarshitChanpura I've responded to all feedback and resolved all threads - let me know what you think.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -116,7 +80,7 @@ public void accept(final RestChannel channel) throws Exception {
                 final XContentBuilder builder = channel.newBuilder();
                 BytesRestResponse response;
                 try {
-                    if (vendor == null) {
+                    if (securityTokenManager == null) {
```
cwperks, line 83: Will the securityTokenManager ever be null? It looks like it will always be present, but the internal jwtVendor is nullable.

wdyt about having an `isInitialized()` on the `securityTokenManager` and using that here?

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -72,45 +67,46 @@ public JwtVendor(final Settings settings, final Optional<LongSupplier> timeProvi
...
-        String audience,
-        Integer expirySeconds,
-        List<String> roles,
-        List<String> backendRoles,
-        boolean includeBackendRoles
+    public ExpiringBearerAuthToken createJwt(
+        final String issuer,
+        final String subject,
+        final String audience,
+        final long requestedExpirySeconds,
```
cwperks, line 103: wdyt about `requestedDurationSeconds`?

Initially when I read this variable I thought it may be an epoch time in seconds.

cwperks: Duration is the length of the window of the token from issuance to expiry. i.e. expiry minus issued_at = duration

Expiry is an exact Date where the token is expired

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
peternied: Good catch, this is not possible to have a null tokenManager.  I've renamed and opened up `issueOnBehalfOfTokenAllowed()` to handle this scenario since Service Accounts & OBO tokens can enabled / disabled seperately.

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,141 @@
...
+        if (Strings.isNullOrEmpty(claims.getAudience())) {
+            throw new IllegalArgumentException("Claims must be supplied with an audience value");
+        }
+
+        final User user = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+        if (user == null) {
+            throw new OpenSearchSecurityException("Unsupported user to generate OnBehalfOfToken");
+        }
+
+        final TransportAddress callerAddress = null; /* OBO tokens must not roles based on location from network address */
```
DarshitChanpura, line 113: nit: the comment seems to be incomplete, probly a relic from previous changes? apologies if I didn't spot it earlier

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
peternied: I'm not going to fix this nit; we can do a future pass where we align the claims language and the downstream systems, as these names are spanning OpenSearch core + security plugin.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -461,7 +461,7 @@ public Map<CType, SecurityDynamicConfiguration<?>> getConfigurationsFromIndex(
             throw new OpenSearchException(e);
         }
 
-        if (logComplianceEvent && auditLog.getComplianceConfig().isEnabled()) {
+        if (logComplianceEvent && auditLog.getComplianceConfig() != null && auditLog.getComplianceConfig().isEnabled()) {
```
DarshitChanpura, line 464: Curious to know why this change was required?

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
peternied: This was intentional - the address should be null

peternied: Hmm, this is a pretty minor issue, I think it will be ok to leave it for now - whereas I've got fully passing CI

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
peternied: During the integration test when I was patching the security configuration to enable/disable OBO it errored out

peternied (approved):

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
```diff
@@ -43,12 +42,14 @@
...
-import static org.opensearch.security.OpenSearchSecurityPlugin.PLUGINS_PREFIX;
-import static org.opensearch.security.util.AuthTokenUtils.isAccessToRestrictedEndpoints;
+import io.jsonwebtoken.Claims;
+import io.jsonwebtoken.JwtParser;
+import io.jsonwebtoken.JwtParserBuilder;
+import io.jsonwebtoken.security.WeakKeyException;
 
 public class OnBehalfOfAuthenticator implements HTTPAuthenticator {
 
+    private static final int MINIMUM_SIGNING_KEY_BIT_LENGTH = 512;
```
scrawfor99, line 52: Do we have a reference from this number? It seems fine but don't know where it came from

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,141 @@
...
+        final Boolean enabled = oboSettings.getAsBoolean("enabled", false);
+        if (enabled) {
+            jwtVendor = createJwtVendor(oboSettings);
+        } else {
+            jwtVendor = null;
+        }
+    }
+
+    /** For testing */
+    JwtVendor createJwtVendor(final Settings settings) {
```
scrawfor99, line 77: Should this not also set the vendor for the instance's attribute?

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,141 @@
...
+        final Set<String> mappedRoles = configModel.mapSecurityRoles(user, callerAddress);
+
+        try {
+            return jwtVendor.createJwt(
+                cs.getClusterName().value(),
+                user.getName(),
+                claims.getAudience(),
+                claims.getExpiration(),
+                mappedRoles.stream().collect(Collectors.toList()),
+                user.getRoles().stream().collect(Collectors.toList()),
```
cwperks, line 123: Can this lookup the audience (extension)'s setting to see if this token should include the backend roles in the payload?

cwperks (commented):
I have one more question around backend_roles, but otherwise this PR looks good to me.

scrawfor99 (commented):
Looks good just a few comments but otherwise seems all set!

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
peternied: Lets look into this later as a separate change.

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -0,0 +1,141 @@
...
+    @Override
+    public ExpiringBearerAuthToken issueOnBehalfOfToken(final Subject subject, final OnBehalfOfClaims claims) {
+        if (oboNotSupported()) {
+            // TODO: link that doc!
+            throw new OpenSearchSecurityException(
+                "The OnBehalfOf token generation is not enabled, see {link to doc} for more information on this feature."
+            );
+        }
+
+        if (subject != null && !(subject instanceof NoopSubject)) {
```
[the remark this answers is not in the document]
peternied: I think they will be really one accepted type (or a hierarchy defined in the security plugin), which is subject that represents a security plugin user.  Any other kind of subject - we wouldn't have the authority to create an OBO token for it.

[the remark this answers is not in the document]
peternied: To your point, we shouldn't have to invest in managing this list over time, so I don't think there should be an approved list

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
peternied: Good question, I had a little bit of a hard time articulating where this value came from other than I copied it from another error message that said the value was associated with the HS512 algorithm.

_ChatGPT helped me answer this with:_
> "The 512 value aligns with the HS512 algorithm's requirement for a secure hash-based message authentication code"

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
peternied: The jwt vendor is only created when the dynamic config is triggered, so it starts null and until the lifecycle of the security plugin has published the dynamic config update in the process it should be unset.

peternied:
@cwperks @DarshitChanpura Anything holding up approval of this PR?

cwperks (approved):

PR #3574, opened 2023-10-18
Allow adding URL to the cluster

### LINKED ISSUE #3321: [FEATURE] Support URLs instead of IPs on external_opensearch option for audit logs
I want to centralise the audit logging of all our clusters into one, using the `external_opensearch` option of audit logs, targeting an external OpenSearch cluster.

As explained in the official [documentation](https://opensearch.org/docs/latest/security/audit-logs/storage-types/#external-opensearch), I need to specify a comma-separated list of hosts/IP addresses and the REST port, like the following:
```
plugins.security.audit.config.http_endpoints: [192.168.178.1:9200,192.168.178.2:9200]
```
As expected, things like `[localhost:9200,localhost:9201]` also work.

What doesn't work is setting an actual URL there, like `https://my-opensearch-cluster.company.com:9200` or `https://my-opensearch-cluster.company.com:443/opensearch`. This will persist even if the hosts behind the cluster change. 
At the moment, if a cluster changes hosts, I need to adapt and restart all clusters, to pick up the new IPs.

Enable the possibility of using an actual URL that corresponds to an OpenSearch cluster on `plugins.security.audit.config.http_endpoints` (or similar setting)

None that I can think of.

davidosorno:
### Description
Allow adding URLs to the OpenSearch cluster with support for both http:// and https:// for external logs.

### Issues Resolved
- #3321 

### Testing
Tested using *RouterTest.java* and manual testing.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/httpclient/HttpClient.java
+++ b/src/main/java/org/opensearch/security/httpclient/HttpClient.java
@@ -34,7 +34,6 @@ import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLParameters;
 
-import com.google.common.collect.Lists;
 import org.apache.hc.client5.http.config.RequestConfig;
 import org.apache.hc.client5.http.impl.async.HttpAsyncClientBuilder;
 import org.apache.hc.client5.http.impl.nio.PoolingAsyncClientConnectionManagerBuilder;
@@ -54,7 +53,6 @@ import org.apache.hc.core5.ssl.SSLContextBuilder;
 import org.apache.hc.core5.ssl.SSLContexts;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-
 import org.opensearch.action.index.IndexRequest;
 import org.opensearch.action.index.IndexResponse;
 import org.opensearch.action.support.WriteRequest.RefreshPolicy;
@@ -64,6 +62,8 @@ import org.opensearch.client.RestClient;
 import org.opensearch.client.RestClientBuilder;
 import org.opensearch.client.RestHighLevelClient;
 import org.opensearch.common.xcontent.XContentType;
+
+import com.google.common.collect.Lists;
 
 public class HttpClient implements Closeable {
 
@@ -179,30 +179,8 @@ public class HttpClient implements Closeable {
         this.supportedCipherSuites = supportedCipherSuites;
         this.keystoreAlias = keystoreAlias;
 
-        /*
-        * pattern helps to get the port number after last colon.
-        * \d matches a  digit equivalent to [0-9].
-        */
-        Pattern pattern = Pattern.compile(":(\\d+)");
-
-        HttpHost[] hosts = Arrays.stream(servers).map(server -> {
-            int lastIndexColon = server.lastIndexOf(':');
-            String url = server.substring(0, lastIndexColon);
-            String uri = server.substring(lastIndexColon);
-            Matcher matcher = pattern.matcher(uri);
-            if (matcher.find()) {
-                int port = Integer.parseInt(matcher.group(1));
-                return new HttpHost(ssl ? "https" : "http", url, port);
-            } else {
-                return null;
-            }
-        })
-            .filter(Objects::nonNull) // Filter out null values
-            .collect(Collectors.toList())
-            .toArray(HttpHost[]::new);
-
+        HttpHost[] hosts = createHosts(servers);
         RestClientBuilder builder = RestClient.builder(hosts);
-        // builder.setMaxRetryTimeoutMillis(10000);
 
         builder.setFailureListener(new RestClient.FailureListener() {
             @Override
@@ -225,6 +203,30 @@ public class HttpClient implements Closeable {
         });
 
         rclient = new RestHighLevelClient(builder);
+    }
+
+    private HttpHost[] createHosts(String[] servers) {
+        /*
+         * pattern helps to get the port number.
+         * \d matches a digit equivalent to [0-9].
+         * $ determines the end of the URL.
+         * So the port number has to be at the end of the URL,
+         * if it is not there, it means the URL contains a
+         * specific path for OpenSearch.
+         */
+        Pattern pattern = Pattern.compile(":(\\d+)$");
+
+        return Arrays.stream(servers).map(server -> {
+            server = server.replaceAll("https://|http://", "");
+            Matcher matcher = pattern.matcher(server);
+            if (matcher.find()) {
+                String url = server.split(pattern.toString())[0];
+                int port = Integer.parseInt(matcher.group(1));
+                return new HttpHost(ssl ? "https" : "http", url, port);
+            } else {
+                return new HttpHost(ssl ? "https" : "http", server);
+            }
+        }).collect(Collectors.toList()).toArray(HttpHost[]::new);
     }
 
     public boolean index(final String content, final String index, final String type, final boolean refresh) {
--- a/src/main/java/org/opensearch/security/httpclient/HttpClient.java
+++ b/src/main/java/org/opensearch/security/httpclient/HttpClient.java
@@ -178,6 +178,7 @@ public class HttpClient implements Closeable {
         this.keystoreAlias = keystoreAlias;
 
         HttpHost[] hosts = Arrays.stream(servers)
+            .map(s -> s.replaceAll("https://|http://", ""))
             .map(s -> s.split(":"))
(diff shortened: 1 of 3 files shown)


peternied (changes requested):
Thanks @davidosorno - could you add a tests that would verify that `127.0.0.1` or similar address would work to validate this change?

davidosorno:
> Thanks @davidosorno - could you add a tests that would verify that `127.0.0.1` or similar address would work to validate this change?

@peternied is it ok if use the `RouteTest.java` file that is already validating URLs or should I create a new test file?

peternied:
> ok if use the RouterTest.java file that is already validating URLs

@davidosorno 💯 Reuse is great!

davidosorno (commented):
Added URLs to `routing/configuration_valid.yml` to validate their format when testing with `RouterTest.java`

peternied (approved):

scrawfor99 (approved):

peternied:
@davidosorno Looks like you've got an issue that is causing the CI to fail - let us know if you need a hand

davidosorno:
Oh I'm sorry @peternied, I didn't realized about the errors cause. I reviewed them now, and it seems my solution didn't work as expected, so I will work on it right away. I'll let you know if I have any questions. Thanks!!!

src/test/resources/auditlog/endpoints/routing/configuration_valid.yml
```diff
@@ -15,7 +15,18 @@ plugins.security:
...
         config:
-          http_endpoints: ['localhost:9200','localhost:9201','localhost:9202']
+          http_endpoints: [
+            'localhost:9200',
+            'localhost:9201',
+            'localhost:9202',
+            'localhost:9202/opensearch',
+            '127.0.0.1',
+            '127.0.0.1:9200',
+            '127.0.0.1:9200/opensearch',
```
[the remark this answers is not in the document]
davidosorno: Hello @DarshitChanpura, good point.
I updated the code and included IPv6 formats, and it worked, however, when testing [::1] it takes like 1 is the port number, but it still passes all tests, so I removed it from tests.
Is this something we should worry about?

peternied:
@davidosorno Still have some spotless issues - https://github.com/opensearch-project/security/actions/runs/6644692249?pr=3574

image

You can also go to the `checks` tab at the top of the pull request to inspect these further.

peternied (approved):

src/test/resources/auditlog/endpoints/routing/configuration_valid.yml
```diff
@@ -15,7 +15,20 @@ plugins.security:
...
+          http_endpoints: [
+            'localhost:9200',
+            'localhost:9201',
+            'localhost:9202',
+            'localhost:9202/opensearch',
+            '127.0.0.1:9200',
+            '127.0.0.1:9200/opensearch',
+            'my-opensearch-cluster.company.com:9200',
+            'my-opensearch-cluster.company.com:9200/opensearch',
+            'https://my-opensearch-cluster.company.com:9200',
```
cwperks, line 27: Is it valid to have the protocol here? For the code that extracts the url w/o port in this PR

```
int lastIndexColon = server.lastIndexOf(':');
String url = server.substring(0, lastIndexColon);
```

would that extract the value of URL as `https://my-opensearch-cluster.company.com`?

If a protocol is configurable here should it be used passed in as the scheme when the instantiating the HttpHost?

i.e.

```
if (urlContainsScheme) {
    return new HttpHost(schemeFromUrl, url, port);
} else {
    return new HttpHost(ssl ? "https" : "http", url, port);
}
```

davidosorno: > Is it valid to have the protocol here? For the code that extracts the url w/o port in this PR

Yes, it works; it extracted the URL correctly and passed the test. 
 
> would that extract the value of URL as `https://my-opensearch-cluster.company.com`?

Yes, that's the result when the URL is like `https://my-opensearch-cluster.company.com:9200`.
However, if there isn't port number, it won't work. I didn't consider URLs without port number, since this is a basic configuration that users should add in order to have the full address for their OpenSearch.

@cwperks should I leave the code as it is or should I update it to make it work even when there are not port number?

davidosorno: Oh, I just realized that even though all the tests passed, it didn't give me a right answer because when it's creating the hosts it adds another http or https to the URL.
I'll fix it.

src/main/java/org/opensearch/security/httpclient/HttpClient.java
```diff
@@ -209,6 +205,30 @@ public HttpAsyncClientBuilder customizeHttpClient(HttpAsyncClientBuilder httpCli
         rclient = new RestHighLevelClient(builder);
     }
 
+    private HttpHost[] createHosts(String[] servers) {
+        /*
```
peternied, line 209: Can we reuse an existing URL/URI parser, and then contruct the HttpHost via that process?

davidosorno: @peternied I tried using URL and URI, but the tests didn't pass when URLs are like `127.0.0.1` or `[::1]:9200`.
Any suggestions?

src/test/resources/auditlog/endpoints/routing/configuration_valid.yml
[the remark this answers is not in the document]
DarshitChanpura: If it covers all scenarios then including it should be fine , correct?

src/main/java/org/opensearch/security/httpclient/HttpClient.java
peternied: ChatGPT produced this [sample program](https://ideone.com/LkL4dF), and it looks like it works.  What didn't pass?

davidosorno: Thanks a lot @peternied, here is what happened:
When creating the URI it needs the scheme, as you can see in the example you linked, so I fixed it, and it worked.
Then I removed the `ssl ? "https://" : "http://"` statement when creating the Host because it was already in the URL and it failed with `java.lang.AssertionError`.
Finally, I created the Host using URI parameters and it worked, however, when creating Hosts it doesn't add the path to the host `HttpHost(uri.getScheme(), uri.getHost(), uri.getPort());`, so 
`https://my-opensearch-cluster.company.com:9200/opensearch` is added as 
`https://my-opensearch-cluster.company.com:9200`
I couldn't find a way to add the path to the host, any ideas on this?

peternied: Hmm, a path on a 'httphost' is a contradiction, then its not a `host` its an endpoint, or a host + a basepath.   Semantics aside, if you print out the list of transformed http hosts using the previous process do that have the path component?  Following up on that, are there tests that exercise that http client.  Because so long as those tests are still passing its not using that path segment a base path, so this approach would be safe.

If it is use the path as base path, then the data structure is wrong, then don't use http host, but create them as Node objects using the normalized url that was parsed.

Note; Also this seems more complex than originally thought so we might need more tests that properly validate this end to end kind of workflow where the client makes a network calls and we are sure they resolve correctly.

Thanks for digging in on this - let me know if you need help.

davidosorno: Host class does not have _path_ component, but URI does.
There are not tests for URL with path, I was just trying to make it work based on this [comment](https://github.com/opensearch-project/security/issues/3321#issuecomment-1721290097).
I'll keep working on it to see what else I can find.

DarshitChanpura: From what I understand, the issue at hand seems to be of supporting urls, and the setting `plugins.security.audit.config.http_endpoints:  ` intuitively suggests setting the endpoint, i.e. `https://localhost:9200` and not host `localhost:9200`,  however it seems to be expecting a host `localhost:9200`. If that is the problem being solved, to support urls with custom opensearch endpoints, this change seems comprehensible to me. But, if that is not the case, then this might be more complicated then what it seems to be,

peternied (approved):

src/main/java/org/opensearch/security/httpclient/HttpClient.java
davidosorno: Using the feedback I got, I created the last commit where it can create hosts from URLs and IP addresses, and I think it solved the main issue described in #3321

cwperks (approved):
Thank you @davidosorno ! These changes look good to me 🚢

cwperks:
@davidosorno The automatic backport failed. Could you raise a manual backport for this change to 2.x?

Issue #2932, opened 2023-07-04
[Maintenance] Review use of SamlHTTPMetadataResolver


willyborankin:
**Is your feature request related to a problem?**
During review of https://github.com/opensearch-project/security/pull/2927, @cwperks matinied that the main reason of why we still use HTTP commons 4 is `SamlHTTPMetadataResolver` which reads IDP metadata file and track the state of it.
The new version 4.3.0 still use HTTP commons 4 though. 

**What solution would you like?**
To exclude HTTP commons 4 completely  there are 2 options exists:
1. Implement our own solution which will use HTTP common 5. For that we need to extend `org.opensaml.saml.metadata.resolver.impl.AbstractReloadingMetadataResolver` and implement almost the same logic as `org.opensaml.saml.metadata.resolver.impl.HTTPMetadataResolver` which `SamlHTTPMetadataResolver` extends.  
2. Try to prepare PR in `Shibboleth` repository with new version (do not know what is the process since repo is not GiHub one)

scrawfor99:
[Triage] Hi @willyborankin, thank you for opening this issue. It sounds like you have done some research and that version 5 is in the works. Please follow up on what you recommend for moving forward (wait for v5 or not).

MaciejMierzwa:
Hi, looking into this one

MaciejMierzwa:
Hi @willyborankin I created draft pr for the task: https://github.com/opensearch-project/security/pull/3894
Feel free to take a look and make suggestions

PR #3580, opened 2023-10-20
[BUG-1402] Return HTTP 409 on version confict when creating & updating tenants - added relevant unit test

### LINKED ISSUE #1402: [BUG] Return HTTP 409 on version confict when creating & updating tenants
**Describe the bug**

Hello, I opened https://github.com/opensearch-project/security/issues/1095 earlier, which has recently been fixed (thanks!), but it looks like a HTTP 500 (instead of 409) is being returned for the tenant endpoint as well. This is related to https://github.com/phillbaker/terraform-provider-elasticsearch/issues/206.

When OpenDistro receives parallel requests to create or update tenants (`/_opendistro/_security/api/tenants/{name}`), it returns an HTTP 500 response.

Similar to Elasticsearch itself, these conflict exceptions should return a http 409 instead of a 500 to signal that this is an expect error case and a retry is expected.

**To Reproduce**
1. Issue concurrent PUT requests to the tenant endpoint, e.g. `/_opendistro/_security/api/tenants/tenant_1`, `/_opendistro/_security/api/tenants/tenant_2`
2. See 500

**Expected behavior**
A HTTP 409 instead of HTTP 500.

**Plugins**
Opendistro

**Screenshots**
N/A

**Host/Environment (please complete the following information):**
 - OS: Ubuntu 
 - Version 20.04.2

**Additional context**
N/A

prabhask5:
### Description
Added new TenantsActionApiTest file to put api unit tests for the /api/tenants api call- additionally wrote new unit test to test for parallel PUT /tenant/{name} calls, that they successfully return a 409 call and retry the call.

Test fix

There was no test that checked for the behavior mentioned in the linked ticket, so previously we didn't know if the issue still existed or not (it apparently was fixed before), to make sure this doesn't happen again for this particular case, I decided to write a concrete test for it.

Test was added, there's no behavior change?

### Issues Resolved
- #1402 - added unit test to cover bug in this issue

### Testing
Added new tenant action unit test to test parallel tenant api calls.

### Check List


prabhask5:
@cwperks This PR was created to address the comment you left under #1402

src/test/java/org/opensearch/security/dlic/rest/api/TenantsActionTest.java
```diff
@@ -0,0 +1,137 @@
...
+                    Assert.assertEquals(HttpStatus.SC_CONFLICT, sc);
+                    break;
+            }
+        }
+    }
+
+    private HttpResponse[] executeMultipleAsyncPutRequest(final String[] requests,
+            final String[] bodies) throws Exception {
+        final int numOfRequests = Math.min(requests.length, bodies.length);
+        final ExecutorService executorService = Executors.newFixedThreadPool(numOfRequests);
```
peternied, line 117: None of the async request logic is needed, after the first `PUT` all future requests will fail with conflict because the tenant has already been created.  You can verify this by setting the threadpool size to 1.

src/test/java/org/opensearch/security/dlic/rest/api/TenantsActionTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+package org.opensearch.security.dlic.rest.api;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+import org.apache.hc.core5.http.HttpStatus;
```
peternied, line 20: Switch to `import org.apache.http.HttpStatus;`, core5 version isn't needed.

peternied: Note; I've filed [1] to have this be caught by automation

- [1] https://github.com/opensearch-project/security/issues/3581

src/test/java/org/opensearch/security/dlic/rest/api/TenantsActionTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Test
+    public void testParallelPutRequests() throws Exception {
+
+        setup();
+
+        rh.keystore = "restapi/kirk-keystore.jks";
+        rh.sendAdminCertificate = true;
+
+        String[] caseOneRequests = { ENDPOINT + "/tenant1", ENDPOINT + "/tenant1" };
```
peternied, line 47: These endpoint urls are always the same, no need to have different copies

src/test/java/org/opensearch/security/dlic/rest/api/TenantsActionTest.java
```diff
@@ -0,0 +1,137 @@
...
+        String[] caseOneBodies = { "{\"description\":\"create tenant 1\"}", "{\"description\":\"create tenant 2\"}" };
```
peternied, line 48: These request bodies aren't verified, I would suggest adding a get on the tenant and then verify the expected description matches.

peternied: Hint; use `assertThat(response.findValueInJson("description"), equalTo("create tenant 1"))` to easily and verbosely make sure the result matches.

src/test/java/org/opensearch/security/dlic/rest/api/TenantsActionTest.java
```diff
@@ -0,0 +1,137 @@
...
+                    break;
+                case HttpStatus.SC_OK:
+                    break;
+                default:
+                    Assert.assertEquals(HttpStatus.SC_CONFLICT, sc);
+                    break;
+            }
+        }
+
+        String[] caseTwoRequests = { ENDPOINT + "/tenant1", ENDPOINT + "/tenant1", ENDPOINT + "/tenant1",
```
peternied, line 66: Further tests are unneeded, so long as the 1 PUT is accepted with CREATED, and the 2nd PUT is rejected with CONFLICT this is working as indented.

src/test/java/org/opensearch/security/dlic/rest/api/TenantsActionTest.java
peternied, line 1: This test should be added to `TenantInfoActionTest.java`, not be in its own file

peternied (changes requested):

peternied:
@prabhask5 Thanks for the contribution of the test - much appreciated! I've left feedback please address and let me know and I'll take another look.

src/test/java/org/opensearch/security/dlic/rest/api/TenantsActionTest.java
cwperks: @peternied Should `org.apache.http.HttpStatus` be used in the integrationTest package to make backporting to 2.x easier? The main branch has already upgraded to org.apache.HttpComponents v5 and v4 is used in 2.x. `main` still has a dependency on v4 because its needed by OpenSAML. See details in https://github.com/opensearch-project/security/issues/2932

cwperks: Let's update the import to `org.apache.http.HttpStatus` which will ensure that the auto backport to 2.x will not have merge conflicts around the package name for these classes.

src/test/java/org/opensearch/security/dlic/rest/api/TenantsActionTest.java
prabhask5: Do you mean even if they're not synchronous? This is interesting because when running `curl -X PUT https://localhost:9200/_plugins/_security/api/tenants/tenant_1 -u 'admin:admin' --insecure -d '{"description": "create tenant 1"}' -H 'Content-Type: application/json'` in my terminal twice in a row I get 20X responses, with the first one giving a "created" response, and the second one giving a "updated" response.

peternied: To be honest, I'm not sure - I just changed the setting associated with this line to `1` and noticed that nothing failed - which would indicate the tests aren't verifying something correctly if you'd expect them to fail.  Its possible there is a bug in that isn't causing the failure state to be noticed in the code, maybe the issue is because cause of how that validate loop works:

```
                case HttpStatus.SC_OK:
                    break;
```

However, in general test around 'racing' for a conflict at an API level are going to be flaky typically, so I would generally discourage test of this nature unless there is a strongly deterministic way to get a reproduction, such as being able to control the timing across threads, or use C/ETags to ensure processing order.

prabhask5:
@peternied Alright I refactored the added test to test only two parallel requests and I'm getting a consistent one 409 and one 201 codes, which should be good. The only thing that changes from test run to test run is which request is the "conflict" one, but I think I account for this when checking the description with the get request. Let me know if you need me to make any other changes! TY!

src/test/java/org/opensearch/security/dlic/rest/api/TenantInfoActionTest.java
```diff
@@ -72,17 +79,78 @@ public void testTenantInfoAPIUpdate() throws Exception {
...
+
+    private HttpResponse[] executeMultipleAsyncPutRequest(final String[] names,
+            final String[] descriptions) throws Exception {
+        final int numOfRequests = Math.min(names.length, descriptions.length);
+        final String TENANTS_ENDPOINT = BASE_ENDPOINT + "/api/tenants/";
+        final ExecutorService executorService = Executors.newFixedThreadPool(numOfRequests);
+        try {
+            List<Future<HttpResponse>> futures = new ArrayList<>(numOfRequests);
+            for (int i_ = 0; i_ < numOfRequests; i_++) {
+                final int i = i_;
```
DarshitChanpura, line 138: why need a final declaration here?

DarshitChanpura (commented):
Thank you for adding this test @prabhask5.

src/test/java/org/opensearch/security/dlic/rest/api/TenantInfoActionTest.java
```diff
@@ -12,20 +12,27 @@
 package org.opensearch.security.dlic.rest.api;
 
 import org.apache.hc.core5.http.Header;
-import org.apache.hc.core5.http.HttpStatus;
+import org.apache.http.HttpStatus;
```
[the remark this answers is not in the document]
peternied: This was a suggestion, see the context https://github.com/opensearch-project/security/pull/3580#discussion_r1366912993  What do you think?

src/test/java/org/opensearch/security/dlic/rest/api/TenantInfoActionTest.java
```diff
@@ -72,17 +79,78 @@ public void testTenantInfoAPIUpdate() throws Exception {
...
+        String[] descriptions = { "create tenant 1", "create tenant 2" };
+        HttpResponse[] responses = executeMultipleAsyncPutRequest(tenantNames, descriptions);
+        int numCreatedResponses = 0;
+        int numConflictResponses = 0;
+        Integer requestIndexCreated = null;
+        for (int i = 0; i < responses.length; i++) {
+            HttpResponse response = responses[i];
+            if (response.getStatusCode() == HttpStatus.SC_CREATED) {
+                numCreatedResponses++;
+                requestIndexCreated = i;
```
cwperks, line 115: wdyt about moving the request and assertion from L124-126 to here to avoid having to keep track of the index of the request with the SC_CREATED response?

src/test/java/org/opensearch/security/dlic/rest/api/TenantInfoActionTest.java
```diff
@@ -72,17 +79,78 @@ public void testTenantInfoAPIUpdate() throws Exception {
...
+            } else if (response.getStatusCode() == HttpStatus.SC_CONFLICT) {
+                numConflictResponses++;
+            }
+        }
+        Assert.assertEquals(1, numCreatedResponses); // check that no more than 1 created status is read
+        Assert.assertEquals(1, numConflictResponses); // check that no more than 1 conflict status is read
+        Assert.assertNotNull(requestIndexCreated); // check that a created status is read
```
peternied, line 122: Look into using streams and groupings to solve this problem, you can condense this code into ~4 lines which would also be much more readable.

src/test/java/org/opensearch/security/dlic/rest/api/TenantInfoActionTest.java
```diff
@@ -72,17 +79,78 @@ public void testTenantInfoAPIUpdate() throws Exception {
...
+                new Header[0]);
         Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
 
         rh.sendAdminCertificate = false;
         response = rh.executeGetRequest(ENDPOINT);
         Assert.assertEquals(HttpStatus.SC_OK, response.getStatusCode());
     }
+
+    @Test
+    public void testParallelPutRequests() throws Exception {
```
scrawfor99, line 98: Please rewrite this test case. See slack for context.

scrawfor99 (changes requested):

peternied:
Async tests are on my mind today, see this PR [1] for an example of how I've implemented async tests + validation.  Those utilities might be useful, feel free to merge my branch onto this change if it helps your development.

- [1] https://github.com/opensearch-project/security/pull/3601

src/test/java/org/opensearch/security/dlic/rest/api/TenantInfoActionTest.java
prabhask5: Is this about the generative AI tools? This is confusing because I definitely did not use any AI tools to write this code. I did use code from UserApiTest.java for  the async request test helper functions, but I think that's fine to do so (please let me know if it's not). I just want to make it clear that I did not use any generative AI tools and will continue to not do so going forward. Thanks!

src/test/java/org/opensearch/security/dlic/rest/api/TenantInfoActionTest.java
prabhask5: I previously got a weird java warning asking me to use that, but it looks like it's not required with the current version of this function, resolved!

src/test/java/org/opensearch/security/dlic/rest/api/TenantInfoActionTest.java
scrawfor99: Hi @prabhask5, thanks for the clarification. I will remove the block on the PR.

prabhask5:
I restructured the test using streams and the AsyncActions.java file from #3601 - let me know of other changes, ty!

DarshitChanpura:
We can wait until #3601 merge to ensure any conflicts are resolved.

DarshitChanpura (dismissed):
Ty for adding the test @prabhask5.

peternied:
@prabhask5 please take a look at the test failures, I'm going to hold off on another review until those are passing. Please let us know if you need help with an issue.

davidlago:
@prabhask5 if you are not ready to continue work on this PR, we can go ahead and close it, and re-open at a later time. Please let us know!

prabhask5:
> @prabhask5 please take a look at the test failures, I'm going to hold off on another review until those are passing. Please let us know if you need help with an issue.

@peternied The errors that I was getting in the test were related to java not being able to find import org.opensearch.test.framework.AsyncActions; when it does exist (which is used in your PR as well), I merged main into this PR to try to solve this error, but if it does not fix it, do you have any fixes going forward?

peternied:
@prabhask5 Checkout this guide [1] on rebasing it might help you get unblocked.

- [1] https://akrabat.com/the-beginners-guide-to-rebasing-your-pr/

prabhask5:
@peternied I think I rebased so feel free to unblock the tests. TY!

prabhask5:
I'm going to make a new PR to see if that fixes things.

peternied:
@prabhask5 Could you link the new PR?

prabhask5:
> @prabhask5 Could you link the new PR?

Here: #3673

PR #3593, opened 2023-10-24
OnBehalfOf tokens odds and ends

peternied:
### Description
I was doing some inspection around the OBO feature and noticed some items to clean up.

### Testing
Unit + Integration tests

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
+++ b/src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
@@ -58,18 +58,12 @@ public class CreateOnBehalfOfTokenAction extends BaseRestHandler {
 
     private ConfigModel configModel;
 
-    private DynamicConfigModel dcm;
-
     public static final Integer OBO_DEFAULT_EXPIRY_SECONDS = 5 * 60;
     public static final Integer OBO_MAX_EXPIRY_SECONDS = 10 * 60;
 
     public static final String DEFAULT_SERVICE = "self-issued";
 
     protected final Logger log = LogManager.getLogger(this.getClass());
-
-    private static final Set<String> RECOGNIZED_PARAMS = new HashSet<>(
-        Arrays.asList("durationSeconds", "description", "roleSecurityMode", "service")
-    );
 
     @Subscribe
     public void onConfigModelChanged(ConfigModel configModel) {
@@ -78,8 +72,6 @@ public class CreateOnBehalfOfTokenAction extends BaseRestHandler {
 
     @Subscribe
     public void onDynamicConfigModelChanged(DynamicConfigModel dcm) {
-        this.dcm = dcm;
-
         Settings settings = dcm.getDynamicOnBehalfOfSettings();
 
         Boolean enabled = Boolean.parseBoolean(settings.get("enabled"));
@@ -146,10 +138,6 @@ public class CreateOnBehalfOfTokenAction extends BaseRestHandler {
 
                     final String description = (String) requestBody.getOrDefault("description", null);
 
-                    final Boolean roleSecurityMode = Optional.ofNullable(requestBody.get("roleSecurityMode"))
-                        .map(value -> (Boolean) value)
-                        .orElse(true); // Default to false if null
-
                     final String service = (String) requestBody.getOrDefault("service", DEFAULT_SERVICE);
                     final User user = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
                     Set<String> mappedRoles = mapRoles(user);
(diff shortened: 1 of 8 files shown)


src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -146,10 +138,6 @@ public void accept(RestChannel channel) throws Exception {
 
                     final String description = (String) requestBody.getOrDefault("description", null);
 
-                    final Boolean roleSecurityMode = Optional.ofNullable(requestBody.get("roleSecurityMode"))
```
peternied, line 149: This is the most substantive change - we should not allow external systems to generate these tokens with backend roles encoded in them.  For plugins that could these tokens, they should instead use the internal API where this is possible.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -111,7 +111,7 @@ public String createJwt(
         Integer expirySeconds,
         List<String> roles,
         List<String> backendRoles,
-        boolean roleSecurityMode
+        boolean includeBackendRoles
```
peternied, line 114: Renamed to make this clearer

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
RyanL1997: Transferring some of the internal conversation between me and @peternied here:

This change is only targeting on the REST API endpoint for the generation of the obo token. For the usage of the `JwtVendor` we can always disable this by setting up the parameter into true. However, we do need to change our documentation issue tho, since this was documented as an API parameter option.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -164,7 +151,7 @@ public void accept(RestChannel channel) throws Exception {
                         tokenDuration,
                         mappedRoles.stream().collect(Collectors.toList()),
                         user.getRoles().stream().collect(Collectors.toList()),
-                        roleSecurityMode
+                        false
```
RyanL1997, line 154: If im understanding this correct, we need this into `true` to have encrypted role claims.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
RyanL1997: Nice.

RyanL1997 (commented):
Hi @peternied , thanks for making these changes. Here are some comments:

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
DarshitChanpura: 1. What will be the default value for this? 
2. Would the API recognize this parameter and accept it or fail ? (My understanding is that it will be the latter based on modifications to `validateRequestParameters`)

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
```diff
@@ -58,28 +57,20 @@ public class CreateOnBehalfOfTokenAction extends BaseRestHandler {
...
-    private DynamicConfigModel dcm;
-
     public static final Integer OBO_DEFAULT_EXPIRY_SECONDS = 5 * 60;
     public static final Integer OBO_MAX_EXPIRY_SECONDS = 10 * 60;
 
     public static final String DEFAULT_SERVICE = "self-issued";
 
     protected final Logger log = LogManager.getLogger(this.getClass());
 
-    private static final Set<String> RECOGNIZED_PARAMS = new HashSet<>(
```
RyanL1997, line 70: Hi @peternied, why we are removing this?

RyanL1997: oh nvm I saw that you transferred that into a `enum` :)

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
peternied: >What will be the default value for this?

Default value for `includeBackendRoles` is defined farther down this file as `false`.

>Would the API recognize this parameter and accept it or fail?

If someone passed it as a parameter it will fail the request.

src/main/java/org/opensearch/security/action/onbehalf/CreateOnBehalfOfTokenAction.java
peternied: I've changed the parameter to be called `includeBackendRoles`, this inverted the previous (IMO confusing) true/false statements.  With a value of `false` the backend roles will not be included in the JWT.

peternied:
FYI - I've got some merge conflicts and test failures to look into.

scrawfor99 (dismissed):

cwperks (approved):

DarshitChanpura (approved):
LGTM!

PR #3579, opened 2023-10-19
[Forwardport main] Switch to `supportsImpersonation` check for http auth backend and add Privileged Action for `JwtParserBuilder `

RyanL1997:
### Description
Switch to `supportsImpersonation` check for http auth backend + wrap JwtParserBuilder with `doPrivileged`

Reference to @cwperks's [comment](https://github.com/opensearch-project/security/pull/3563#discussion_r1365539239):
>As a default implementation the authDomain could have:
>
>```
>default boolean supportsImpersonation() { return true; }
>```
>
>and any authDomain that does not support it can override:
>
>```
>@Override
>public boolean supportsImpersonation() { return false; }
>```

* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation)
Enhancement

### Issues Resolved
* Related https://github.com/opensearch-project/security/pull/3563

Is this a backport? If so, please add backport PR # and/or commits #
It has already been included in `2.x` 

### Check List
- [ ] New functionality includes testing
- [ ] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3594, opened 2023-10-24
Wraps HTTPJwtAuthenticator jwtParserBuilder.build call with access controller to prevent AccessControlException

DarshitChanpura:
### Description
HTTPJwtAuthenticator fails initialization of JwtParser due to InvocationTargetException when calling jwtParserBuilder.build().
This is causing API calls with JWT token to fail.

It is root cause of failures identified in https://github.com/opensearch-project/security-dashboards-plugin/pull/1617 and only exists in 2.12.0

I verified manually by running `jwt_auth.test.ts` and making curl calls that this behavior works as expected on `main` and `2.11`

### Issues Resolved
- Unblocks https://github.com/opensearch-project/security-dashboards-plugin/pull/1617

### Testing
Manual Testing.

Spin up a 2.x cluster without this change & run jwt_auth.test.ts. They would fail with ```TimeoutError: Waiting for at least one element to be located By(xpath, //*[@id="osdOverviewPageHeader__title"])
    Wait timed out after 15011msError: Waiting for at least one element to be located By(xpath, //*[@id="osdOverviewPageHeader__title"])
    Wait timed out after 15011ms```
  
After this change, these tests should pass.

A similar change was made in https://github.com/opensearch-project/security/pull/3579 to unblock OBO Authenticator.

### Check List

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
+++ b/src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
@@ -81,7 +81,17 @@ public class HTTPJwtAuthenticator implements HTTPAuthenticator {
                 jwtParserBuilder = jwtParserBuilder.require("iss", requireIssuer);
             }
 
-            jwtParser = jwtParserBuilder.build();
+            final SecurityManager sm = System.getSecurityManager();
+            if (sm != null) {
+                sm.checkPermission(new SpecialPermission());
+            }
+            JwtParserBuilder finalJwtParserBuilder = jwtParserBuilder;
+            jwtParser = AccessController.doPrivileged(new PrivilegedAction<JwtParser>() {
+                @Override
+                public JwtParser run() {
+                    return finalJwtParserBuilder.build();
+                }
+            });
         }
     }
 
--- a/src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
+++ b/src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
@@ -69,29 +69,23 @@ public class HTTPJwtAuthenticator implements HTTPAuthenticator {
         requireAudience = settings.get("required_audience");
         requireIssuer = settings.get("required_issuer");
 
-        JwtParserBuilder jwtParserBuilder = KeyUtils.createJwtParserBuilderFromSigningKey(signingKey, log);
+        final JwtParserBuilder jwtParserBuilder = KeyUtils.createJwtParserBuilderFromSigningKey(signingKey, log);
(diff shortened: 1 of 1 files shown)


RyanL1997 (dismissed):

RyanL1997:
Just for some context. the error msg looks like this:
```
[2023-10-24T16:37:20,820][ERROR][o.o.s.s.DynamicConfigModelV7] [3c06300b34da.ant.amazon.com] Unable to initialize auth domain jwt_auth_domain=AuthcDomain [http_enabled=true, transport_enabled=false, order=5, http_authenticator=HttpAuthenticator [challenge=true, type=jwt, config={signing_key=VGhpcyBpcyBhIHZlcnkgc2VjdXJlIHNlY3JldC4gTm8gb25lIHdpbGwgZXZlciBiZSBhYmxlIHRvIGd1ZXNzIGl0IQ==, jwt_header=Authorization, jwt_url_parameter=token, subject_key=sub, roles_key=roles}], authentication_backend=AuthcBackend [type=noop, config={}], description=null] due to OpenSearchException[java.lang.reflect.InvocationTargetException]; 
nested: InvocationTargetException; 
nested: AccessControlException[access denied ("java.lang.RuntimePermission" "getClassLoader")];
org.opensearch.OpenSearchException: java.lang.reflect.InvocationTargetException
...
```

cwperks:
@RyanL1997 Do you have the full stack trace? Where is the authenticator instantiated?

The code to `newInstance` in `DynamicConfigModelV7` is wrapped in a doPrivileged call on main, but not on 2.x: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java#L438-L444

If I remember correctly, that call is what instantiates the auth backends in the `authc:` section of the security configuration.

cwperks:
Ok I see the full stacktrace here: https://github.com/opensearch-project/security-dashboards-plugin/pull/1617#issuecomment-1778079379

Not sure where it makes the most sense to wrap in `doPrivileged`. Would it make sense to backport the change on main where `DynamicSecurityModelV7.newInstance` has the wrapped code?

DarshitChanpura:
@cwperks here is the full stack trace: https://github.com/opensearch-project/security-dashboards-plugin/pull/1617#issuecomment-1778079379

DarshitChanpura:
I think we should to keep it in sync. Let me test with that wrap to verify whether that approach works.

DarshitChanpura:
Upon digging further it looks like this change was present in 2.x for a bit before it was removed accidentally via this commit: https://github.com/opensearch-project/security/commit/587d747dd87a5f75a4d95d618ef606114e7fe00f

cwperks (approved):

RyanL1997 (approved):

PR #3596, opened 2023-10-25
[Feature-2188] AuditMessage validation in tests should be more useful / removed

### LINKED ISSUE #2188: [FEATURE] AuditMessage validation in tests should be more useful / removed
Test code was written to verify the format of audit log messages.  This code is clunky as in tests it's written as `Assert.assertTrue(validateMsgs(messages));`.  It would be far more useful if it emitted errors about why the message(s) are invalid rather than asserting false.

Furthermore; It is not clear that the message format level validation in the integration tests is useful, so we should investigate under what conditions we are testing for/against.

Code in question
https://github.com/opensearch-project/security/blob/b836040c577894a08db9377d46a3d4eaca426fb8/src/test/java/org/opensearch/security/auditlog/AbstractAuditlogiUnitTest.java#L88

We should have assertions that clearly indicate what is wrong so the failure message can be instructive to developers diagnosing the tests, or we should remove this functionality if it isn't providing valid to our test cases.

prabhask5:
### Description
Changed validateMsgs in AbstractAuditlogUnitTest.java to throw exceptions with descriptions instead of just returning true or false. Not sure if the message format level validation in the integration tests is useful yet, so feel free to help me think about the usefullness of that.

Enhancement

Previous version of the tests were unclear in what is erroring, can be hard for developers to debug.

validateMsgs and traceback previously returned boolean true or false values without any errors, now they throw Exceptions with specific descriptions of what went wrong.

### Issues Resolved
#2188 

### Testing
Updated tests.

### Check List


prabhask5:
These helper functions are a bit confusing since the only function that throws its own exceptions is validateJson- so I added a NullPointerException for if the json is empty or null, and I assume the other code block may throw its own exceptions since it was previously in a try catch block. For all the other level functions (validateMsg, validateMsgs) I'm not really sure how to add more specific exceptions since in the original code they just appear to take Exceptions from validateJson and throw their own exceptions- this is why I added a specific message for the exception.

prabhask5:
Not sure how to handle catching the JsonMappingExceptions and making them more specific up the stack trace since apparently the constructor for it is deprecated. Would like some guidance on that if possible

src/test/java/org/opensearch/security/auditlog/AbstractAuditlogiUnitTest.java
```diff
@@ -83,34 +84,46 @@ protected void setupStarfleetIndex() {
...
     }
 
-    protected boolean validateJson(final String json) {
-
-        if (json == null || json.isEmpty()) {
-            return false;
+    protected void validateMsg(final AuditMessage msg) throws Exception {
+        try {
+            validateJson(msg.toJson());
+        } catch (IllegalArgumentException e) {
```
peternied, line 103: You can remove these catch blocks since the method throws `Exception` and you end up rethrowing the exception anyways

prabhask5: I wanted to build on to the exception message as it travels up the method stack trace, do you think this is required?

prabhask5: Or if this is good, how can I do that without a catch block?

src/test/java/org/opensearch/security/auditlog/AbstractAuditlogiUnitTest.java
```diff
@@ -83,34 +83,27 @@ protected void setupStarfleetIndex() {
...
-    }
 
-    protected boolean validateMsg(final AuditMessage msg) {
-        return validateJson(msg.toJson()) && validateJson(msg.toPrettyString());
     }
 
-    protected boolean validateJson(final String json) {
+    protected void validateMsg(final AuditMessage msg) throws Exception {
+        validateJson(msg.toJson());
+        validateJson(msg.toPrettyString());
```
DarshitChanpura, line 95: why do we need to validate twice?

prabhask5: I'm not sure what toPrettyString does so I'm assuming this tests if both versions of the msg are valid. This was in the previous implementation, do we need this? @DarshitChanpura

prabhask5: @DarshitChanpura Following up on this, please also allow the tests to run so I can see what errors pop up. Thanks!

peternied (approved):

src/test/java/org/opensearch/security/auditlog/AbstractAuditlogiUnitTest.java
```diff
@@ -83,34 +83,27 @@ protected void setupStarfleetIndex() {
         rh.keystore = keystore;
     }
 
-    protected boolean validateMsgs(final Collection<AuditMessage> msgs) {
-        boolean valid = true;
+    protected void validateMsgs(final Collection<AuditMessage> msgs) throws Exception {
         for (AuditMessage msg : msgs) {
-            valid = validateMsg(msg) && valid;
+            validateMsg(msg);
```
scrawfor99, line 88: Nice!

scrawfor99 (approved):
Looks good to me

peternied:
@prabhask5 Please look into the failing CI checks, it looks like there are some outstanding issues that need to be resolved

prabhask5:
@peternied Most of the errors from before where not related to changes that I made, so I'm pretty confused on how to fix them. I merged main, so hopefully(?) it fixes things. If not, I'd like some guidance on how to proceed. Thanks!

DarshitChanpura (approved):

peternied (approved):

Issue #3426, opened 2023-09-29
[Test] SearchOperationTest is flaky


peternied:
### Description
- There are many tests that use audit log validation of scenarios and it seems either the number of requests is not always consistent
- There are many tests that use reindexing and it looks like the operation isn't being confirmed as completed before the tests continue
- There are many tests that use alias creation / clean up and it seems resolution on these actions is not completing before the tests continue causing issues.

### Repo
1. Go to `src/integrationTest/java/org/opensearch/security/SearchOperationTest.java`
2. Find the `@Ignore` many tests in this class
3. Delete that attribute
4. Run the tests with `./gradlew integrationTest --tests org.opensearch.security.SearchOperationTest`

### Exit Criteria
- [ ] Find the root cause(s) of the audit log mismatches, all audit log data should expect exact counts, or have a deterministic function to calculate the count based on number of requests, etc.
- [ ] Find the root cause of the reindexing issues, add a deterministic process to be sure this has completed, note; this could involve a product fix
- [ ] Find the root cause of the aliases issues, add a deterministic process to be sure this has completed, note; this could involve a product fix
- [ ] Run the test at least 10 times and see no failures - [how to do this](https://github.com/opensearch-project/security/pull/3388/commits/a714cf5cd07c0fa624965ab0b1a7c34e5143b75e)

MaciejMierzwa:
Looking into it

MaciejMierzwa:
This issue is linked to this one: https://github.com/opensearch-project/security/issues/2141
Looking into both

MaciejMierzwa:
Hi, here is backport PR: https://github.com/opensearch-project/security/pull/3862
@scrawfor99 @peternied @cwperks @davidlago @DarshitChanpura

cwperks:
Closing this issue as associated PRs have been merged.

Issue #2169, opened 2022-10-14
[BUG] An inconsistent number of audit logs is created when a snapshot is taken.


lukasz-soszynski-eliatra:
**What is the bug?**
A variable number of audit logs is created when a snapshot is taken:

1. One or two audit logs are created with the category `AUTHENTICATED` and request path `/_snapshot/test-snapshot-repository/snapshot-positive-test`
2. Two or four audit logs are created with the category `GRANTED_PRIVILEGES` and request type `GetSnapshotsRequest`

**How can one reproduce the bug?**
The bug was found during work on integration tests for snapshot features. The easiest way to reproduce the problem is to run the test `org.opensearch.security.SearchOperationTest#shouldCreateSnapshot_positive`, https://github.com/opensearch-project/security/pull/2153/files#diff-a407f817ee7d41d20775ea3fec31ed0a1fb81230503909a94266e78fe73a07d0R1385

The number of audit logs depends on the test execution order. When only the above test is run from the IDE then always 2 audits `AUTHENTICATED/_snapshot/test-snapshot-repository/snapshot-positive-test` and 4 `GRANTED_PRIVILEGES/GetSnapshotsRequest` are created. But when all tests from the class `SearchOperationTest` are run then depending on the execution order only 1 audits `AUTHENTICATED/_snapshot/test-snapshot-repository/snapshot-positive-test` and 2 `GRANTED_PRIVILEGES/GetSnapshotsRequest` are created during test `shouldCreateSnapshot_positive`.

It is also worth mentioning that after each test snapshot repository is removed if exists.

**What is the expected behaviour?**
The consistent number of audit logs is created

**What is your host/environment?**
 - Integration tests

**Do you have any screenshots?**
no

**Do you have any additional context?**
To easily tracks the number of audit logs created during tests it is worth assigning `DEBUG` level to logger `org.opensearch.test.framework.audit.AuditLogsRule`

MaciejMierzwa:
Hi, it seems like the problem was caused by Awaitility.await() method used in [SnapshotSteps.waitForSnapshotCreation()](https://github.com/opensearch-project/security/pull/2153/files#diff-776e1c7c6f6b07f8ab4a3fdca127458389f1ae9bf385dfcc7fcb8bc9954a9b5f)
Default poll interval is 100 miliseconds, if the call wasn't succesfull in this time it would perform another request, increasing number of audit logs.
This task should be solved by: https://github.com/opensearch-project/security/pull/3602
[...]

Issue #2141, opened 2022-10-05
[BUG] Privileges are incorrectly evaluated in case of shrink, clone and split index operations


kt-eliatra:
**What is the bug?**
A forbidden response is returned for shrink, clone and split operations even when user has all required permissions to the source index. 
In addition, it seems that permissions to create the target index are not verified at all.

**How can one reproduce the bug?**
1. Define user, role, and role mapping, e.g.:
```
internal_users.yml
    test-user:
      hash: ...
roles.yml
    test-user-role:
      index_permissions:
        - index_patterns:
            - 'source-index'
          allowed_actions:
            - 'indices:admin/resize'
            - 'indices:monitor/stats'
roles_mapping.yml
    test-user-role:
      reserved: false
      users:
      - "test-user"
```
2. Upload configuration
3. Create source index, e.g.:
```
curl --location --request PUT 'https://localhost:9200/source-index' \
--header 'Authorization: Basic ...' \
--header 'Content-Type: application/json' \
--data-raw '{
  "settings": {
    "index": {
      "blocks.write": true
    }
  }
}'
```
4. Try to execute shrink/split/clone operation as the user created in previous step, e.g.: 
```
curl --location --request PUT 'https://localhost:9200/source-index/_clone/target-index' \
--header 'Authorization: Basic ...'
```
5. Response status code is equal to 403, body says that `no permissions for [indices:admin/resize] and User ....`
6. Update role setup & upload configuration, e.g.
```
roles.yml
    test-user-role:
      index_permissions:
        - index_patterns:
            - '*'
          allowed_actions:
            - 'indices:admin/resize'
        - index_patterns:
            - 'source-index'
          allowed_actions:
            - 'indices:monitor/stats'
```
7. Try to execute shrink/split/clone operation as the user created in previous step, e.g.: 
```
curl --location --request PUT 'https://localhost:9200/source-index/_clone/target-index' \
--header 'Authorization: Basic ...'
```
8. Response status code is equal to 200, target index exists

**What is the expected behavior?**
It should be possible to grant the 'indices:admin/resize' permission to selected index/indices.
It should be verified whether the user has the right to create the target index or not.

**What is your host/environment?**
 - OS: Linux Mint 
 - Version: 20.3
 - Plugins: Security Plugin 2.3.0.0

**Do you have any screenshots?**

**Do you have any additional context?**
The problem was noticed during writing integration tests for index operations.

vinayak15:
[...]

PR #3602, opened 2023-10-26
Search operation test flakiness fix

MaciejMierzwa:
### Description
test PR 
### Issues Resolved
- https://github.com/opensearch-project/security/issues/3426
- https://github.com/opensearch-project/security/issues/2141
- https://github.com/opensearch-project/security/issues/2169

Similar to this task: https://github.com/opensearch-project/security/issues/1917
From what I've noticed some logs are duplicated on faster machines. During test creation audit logging results were added based on actual results produced by tests. Now if there are slower moments where logging produces non-duplicated logs, those were marked as failed. 


src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
```diff
@@ -999,8 +1003,8 @@ public void shouldIndexDocumentInBulkRequest_partiallyPositive() throws IOExcept
         auditLogsRule.assertExactlyOne(userAuthenticated(LIMITED_WRITE_USER).withRestRequest(POST, "/_bulk"));
         auditLogsRule.assertExactlyOne(grantedPrivilege(LIMITED_WRITE_USER, "BulkRequest"));
         auditLogsRule.assertExactly(2, grantedPrivilege(LIMITED_WRITE_USER, "CreateIndexRequest"));
-        auditLogsRule.assertExactly(6, auditPredicate(INDEX_EVENT).withEffectiveUser(LIMITED_WRITE_USER));
```
peternied, line 1002: To me its a bug if we don't have a consistent count.  If there is some kind of retry or reissuing or other dynamic contribution to the test, we should be able to measure it and verify that its behaving consistently with these assertions.

For example we might want to change `waitForSnapshotCreation` so it returns the number of calls issued, this might be causing the inconsistency, then we can use that value to indicate how many events are in the audit log for those specific scenarios.

src/main/java/org/opensearch/security/resolver/IndexResolverReplacer.java
```diff
@@ -776,6 +777,18 @@ private boolean getOrReplaceAllIndices(final Object request, final IndicesProvid
                 return false;
             }
             ((CreateIndexRequest) request).index(newIndices.length != 1 ? null : newIndices[0]);
+        } else if (request instanceof ResizeRequest) {
+            // clone or shrink operations
```
DarshitChanpura, line 781: why is this change needed? Is it only for testing?

src/integrationTest/java/org/opensearch/test/framework/cluster/LocalCluster.java
```diff
@@ -531,4 +533,7 @@ public TestCertificates getTestCertificates() {
         return testCertificates;
     }
 
+    public void setPrimaryNode(String primaryNode) {
```
DarshitChanpura, line 536: why need primary Node?

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
DarshitChanpura: +1. Is there a way we can make this more concrete?

src/main/java/org/opensearch/security/resolver/IndexResolverReplacer.java
MaciejMierzwa: I noticed that test shrinkIndex_positive()  and others that dealt with shrinking index were ignored from the start. The request that was created in test wasn't analyzed in any way in indexResolver. After this change request is checked, authorized against index and if user has permissions, shrink operation takes place. 
I'm not sure yet if this is the case in manual testing, have to try it

MaciejMierzwa:
Two things cause the flakiness.
1. The way requests are performed on the cluster. When HighLevelRestClient is created depending on availability, nodes are fetched in random order. Then during https calls available nodes fetched in the first step are rotated using round robin. Because of that rotation with each run different nodes might get request which causes different amount of traffic between the nodes.  
2. Another thing causing a lot of internal traffic was internal shard allocation within the cluster there might be different amount of transport messages. For some cases what seems to be helpful is to tightly control index settings and to use fewer shards. Tests containing bulk requests are example of that.

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
MaciejMierzwa: Seems like issue in this test was caused by transport requests between the cluster nodes. Forcing rest client to communicate through one of the managers node helps to stabilize audit logs count. (cluster.setPrimaryNode() is used for that) Other helpful thing was setting up index to use less shards.

src/integrationTest/java/org/opensearch/test/framework/cluster/LocalCluster.java
MaciejMierzwa: During creation of high level rest client available nodes were probed, cluster managers would have highest priority, but data nodes were also checked. On top of that nodes were rotated which caused inconsistency amount of transport audit logs. This method is used to force one node to be used as rest client

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
```diff
@@ -1843,28 +1943,30 @@ public void shouldRestoreSnapshot_positive() throws IOException {
...
+        auditLogsRule.assertExactly(
+            snapshotGetCount,
             userAuthenticated(LIMITED_WRITE_USER).withRestRequest(GET, "/_snapshot/test-snapshot-repository/restore-snapshot-positive")
         );
-        auditLogsRule.assertExactly(2, grantedPrivilege(LIMITED_WRITE_USER, "PutRepositoryRequest"));
-        auditLogsRule.assertExactly(2, grantedPrivilege(LIMITED_WRITE_USER, "CreateSnapshotRequest"));
+        auditLogsRule.assertExactly(1, grantedPrivilege(LIMITED_WRITE_USER, "PutRepositoryRequest"));
+        auditLogsRule.assertExactly(1, grantedPrivilege(LIMITED_WRITE_USER, "CreateSnapshotRequest"));
         auditLogsRule.assertExactly(2, grantedPrivilege(LIMITED_WRITE_USER, "BulkRequest"));
-        auditLogsRule.assertExactly(2, grantedPrivilege(LIMITED_WRITE_USER, "CreateIndexRequest"));
```
scrawfor99, line 1855: It seems like this check is dropped? Is there a reason we don't need it here? It looks like you kept it above but then stopped using it

MaciejMierzwa: Yes, it was dropped. I used IndexOperationsHelper.createIndex(cluster, WRITE_SONG_INDEX_NAME, indexSettings); to create index instead of creating it via rest API. It was causing problems with different amount of audit logs

src/main/java/org/opensearch/security/resolver/IndexResolverReplacer.java
DarshitChanpura: okay, you can post the results of your manual testing here. I know you were planning to address this separately. Maybe file an issue and a follow-up PR?

src/integrationTest/java/org/opensearch/security/SnapshotSteps.java
```diff
@@ -58,13 +60,21 @@ public CreateSnapshotResponse createSnapshot(String repositoryName, String snaps
...
+    public int waitForSnapshotCreation(String repositoryName, String snapshotName) {
+        AtomicInteger count = new AtomicInteger();
         GetSnapshotsRequest getSnapshotsRequest = new GetSnapshotsRequest(repositoryName, new String[] { snapshotName });
-        Awaitility.await().alias("wait for snapshot creation").ignoreExceptions().until(() -> {
-            GetSnapshotsResponse snapshotsResponse = snapshotClient.get(getSnapshotsRequest, DEFAULT);
-            SnapshotInfo snapshotInfo = snapshotsResponse.getSnapshots().get(0);
-            return SnapshotState.SUCCESS.equals(snapshotInfo.state());
-        });
+        Awaitility.await()
+            .pollDelay(250, TimeUnit.MILLISECONDS)
```
DarshitChanpura, line 67: will these values hold or be stable?

src/main/java/org/opensearch/security/resolver/IndexResolverReplacer.java
MaciejMierzwa: I was able to reprocude the error by manual testing, also found old task documenting the issue. Here's the task: https://github.com/opensearch-project/security/issues/2141
Here's PR where I cherry picked fix changes from this branch: https://github.com/opensearch-project/security/pull/3716
All shrink operation/resize tests are un-ignored and are passing, I'll cleanup this branch to not push same changes 2 times.

src/integrationTest/java/org/opensearch/security/SnapshotSteps.java
MaciejMierzwa: I put this part in (pollDelay()) on earlier iteration of fixing tests dealing with snapshot creation. I found that introducing a little delay before performing HTTP check call is beneficial. At this point it shouldn't really matter, I refactored the method to return a number of calls performed, so after snapshot is verified it doesn't matter if state was checked once or couple times. Asserts check returned number of calls from this method against audit messages logged.
I could remove it and check again does it passes ci correctly without these changes (it should). The default poll delay and poll interval is 100 milliseconds.
Let me know what do you think.

.github/workflows/ci.yml
```diff
@@ -110,6 +110,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        number: [1,2,3,4,5,6,7,8,9,10]
```
MaciejMierzwa, line 113: //TODO
It needs to be reversed before a merge

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
```diff
@@ -1843,28 +1869,30 @@ public void shouldRestoreSnapshot_positive() throws IOException {
                 "/_snapshot/test-snapshot-repository/restore-snapshot-positive/_restore"
             )
         );
-        auditLogsRule.assertExactlyOne(userAuthenticated(LIMITED_WRITE_USER).withRestRequest(POST, "/restored_write_song_index/_count"));
+        auditLogsRule.assertAtLeast(1, userAuthenticated(LIMITED_WRITE_USER).withRestRequest(POST, "/restored_write_song_index/_count"));
```
scrawfor99, line 1872: Can we not assert an exact for this?

scrawfor99 (commented):
Overall looks good to me just one comment on a test assertion that I hope we can swap to an exact.

src/integrationTest/java/org/opensearch/security/SnapshotSteps.java
DarshitChanpura: If the tests seem to be stable with this delay, I don't mind keeping it. I would like to be sure that this is not bound to change inconsistently in future as it is a magic number being introduced here.

MaciejMierzwa: It will be consistent thanks to the number of calls returned.

src/main/java/org/opensearch/security/httpclient/HttpClient.java
```diff
@@ -33,6 +33,7 @@
 import javax.net.ssl.SSLEngine;
```
peternied, line 33: Please revert these changes

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
```diff
@@ -1138,14 +1147,35 @@ public void shouldDeleteDocumentInBulk_positive() throws IOException {
...
         auditLogsRule.assertExactly(2, grantedPrivilege(LIMITED_WRITE_USER, "BulkRequest"));
-        auditLogsRule.assertExactly(2, grantedPrivilege(LIMITED_WRITE_USER, "CreateIndexRequest"));
-        auditLogsRule.assertExactly(4, grantedPrivilege(LIMITED_WRITE_USER, "PutMappingRequest"));
-        auditLogsRule.assertExactly(6, auditPredicate(INDEX_EVENT).withEffectiveUser(LIMITED_WRITE_USER));
+        auditLogsRule.assertExactly(4, auditPredicate(null).withLayer(AuditLog.Origin.TRANSPORT));
+        auditLogsRule.assertAtLeastTransportMessages(4, grantedPrivilege(LIMITED_WRITE_USER, "PutMappingRequest"));
+        auditLogsRule.assertAtLeastTransportMessages(4, auditPredicate(INDEX_EVENT).withEffectiveUser(LIMITED_WRITE_USER));
+    }
+
+    private static void getAndSetPrimaryNode() throws IOException {
```
peternied, line 1155: We should randomly issue requests on primary and non-primary nodes.  Instead of doing this work, lets disable audit logging of transport requests - IMO they aren't important.  Then we should always have the same deterministic number of requests.

Note; if these were `audit log test` I'd have other thoughts, but that isn't the case.

peternied (changes requested):

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
MaciejMierzwa: What do you think about removing this part and starting to rotate nodes, but also leaving assertAtLeastTransportMessages() part of logs verification?

peternied (approved):

src/integrationTest/java/org/opensearch/test/framework/audit/AuditLogsRule.java
```diff
@@ -122,16 +135,35 @@ private void whenTimeoutOccurs(String methodName) {
...
         TestRuleAuditLogSink.registerListener(this);
     }
 
     public void onAuditMessage(AuditMessage auditMessage) {
-        currentTestAuditMessages.add(auditMessage);
-        log.debug("New audit message received '{}', total number of audit messages '{}'.", auditMessage, currentTestAuditMessages.size());
+        if (auditMessage.getAsMap().keySet().contains("audit_transport_headers")) {
+            if (log.isDebugEnabled()) {
+                log.debug(
+                    "New transport audit message received '{}', total number of transport audit messages '{}'.",
```
cwperks, line 152: Will the updated tests still pass with different cluster configurations? i.e. more data nodes added to the cluster?

I like separating out the message types here which can differentiate types of audit messages.

src/integrationTest/java/org/opensearch/test/framework/audit/AuditLogsRule.java
```diff
@@ -122,16 +135,35 @@ private void whenTimeoutOccurs(String methodName) {
...
+                    auditMessage,
+                    currentTransportTestAuditMessages.size()
+                );
+            }
+            currentTransportTestAuditMessages.add(auditMessage);
+        } else {
```
DarshitChanpura, line 158: nit: the only diff I see in this if-else block is the word `transport` in the log message. Can we extract and modify to pass message depending on the condition?

MaciejMierzwa: @DarshitChanpura there is also a different list used in each parts of if-else statement
I could refactor it into something like this:
```
public void onAuditMessage(AuditMessage auditMessage) {
    List<AuditMessage> targetList;
    String logMessage;

    if (auditMessage.getAsMap().containsKey("audit_transport_headers")) {
        targetList = currentTransportTestAuditMessages;
        logMessage = "New transport audit message received '{}', total number of transport audit messages '{}'.";
    } else {
        targetList = currentTestAuditMessages;
        logMessage = "New audit message received '{}', total number of audit messages '{}'.";
    }

    if (log.isDebugEnabled()) {
        log.debug(logMessage, auditMessage, targetList.size());
    }

    targetList.add(auditMessage);
}
``` 
It's little bit harder to read/follow what's happening so not sure about this one :/
Let me know what do you think

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
```diff
@@ -1525,12 +1515,12 @@ public void shouldUpdateTemplate_positive() throws IOException {
...
         auditLogsRule.assertExactly(2, userAuthenticated(LIMITED_WRITE_USER).withRestRequest(PUT, "/_template/musical-index-template"));
         auditLogsRule.assertExactlyOne(userAuthenticated(LIMITED_WRITE_USER).withRestRequest(PUT, "/song-transcription-jazz/_doc/000one"));
-        auditLogsRule.assertExactly(4, grantedPrivilege(LIMITED_WRITE_USER, "PutIndexTemplateRequest"));
+        auditLogsRule.assertExactly(2, grantedPrivilege(LIMITED_WRITE_USER, "PutIndexTemplateRequest"));
         auditLogsRule.assertExactlyOne(grantedPrivilege(LIMITED_WRITE_USER, "IndexRequest"));
         auditLogsRule.assertExactlyOne(grantedPrivilege(LIMITED_WRITE_USER, "BulkRequest"));
-        auditLogsRule.assertExactly(2, grantedPrivilege(LIMITED_WRITE_USER, "CreateIndexRequest"));
-        auditLogsRule.assertExactly(4, grantedPrivilege(LIMITED_WRITE_USER, "PutMappingRequest"));
-        auditLogsRule.assertExactly(10, auditPredicate(INDEX_EVENT).withEffectiveUser(LIMITED_WRITE_USER));
+        auditLogsRule.assertExactlyOne(grantedPrivilege(LIMITED_WRITE_USER, "CreateIndexRequest"));
```
scrawfor99, line 1521: Nit: How come we use `assertExactly(1...` above but `assertExactlyOne` here? Can we use only a single of these patterns?

scrawfor99 (commented):
Is there a reason we cannot get a hard count for the transport messages and still have to use the "at Least" version? If so, please add it as documentation somewhere in the change otherwise please swap to the exact versions. Otherwise looks good!

peternied (approved):

peternied:
@scrawfor99 @DarshitChanpura @cwperks Want to take another look?

scrawfor99 (approved):
Looks good to me! Thanks Maciej!

peternied:
@MaciejMierzwa Could you make the backport to 2.x for this change?

Issue #3604, opened 2023-10-26
[BUG] SamlAuthenticator expects `exchange_key` to be Base64 encoded


DarshitChanpura:
**What is the bug?**

The new library switch from cxf to nimbus introduced a change in behavior of the format a key is accepted in. Prior to this change, the key didn't need to be base 64 encoded but with this change.

in 2.11 we don’t expect it to be encoded: https://github.com/opensearch-project/security/blob/2.11/src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java#L255
in main/2.x we expect it to be encoded: https://github.com/opensearch-project/security/blob/2.x/src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java#L255

This implies the library switch PR introduced a breaking change.

**How can one reproduce the bug?**
0. Start Opensearch and Opensearch Dashboards with security plugin installed.
1. Run [saml_auth.test.ts](https://github.com/opensearch-project/security-dashboards-plugin/blob/main/test/jest_integration/saml_auth.test.ts#L151) without base64 encoded exchange key
2. You'll see an exception in opensearch logs: `nested: IllegalArgumentException[Illegal base64 character 2d];` Full error logs [here](https://github.com/opensearch-project/security-dashboards-plugin/pull/1617#issuecomment-1781327698).

**What is the expected behavior?**
Should work without base-64 encoded exchange key.

Exit Criteria:

- [ ] Fix the [breaking change](https://github.com/opensearch-project/security/blob/2.x/src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java#L255)
OR
- [ ] Document the breaking change

PR #3605, opened 2023-10-26
Remove unneeded decoding from AuthTokenProcessorHandler

DarshitChanpura:
### Description
Fixes bug where the SAMLAuthenticator expects the provided key to be base64 encoded otherwise it would cause saml-authenticator initialization to fail.

Front-end tests pass exchange-key in URL encoded format. Another solution to this would be to remove `getDecoder()` completely. This works and I've tested it locally as well.

I proceed with the approach to remove decoding as that was the state prior to library change.

The tests didn't catch this because they use `abc` as exchange_key. I've updated it to reflect the exchange key used in front-end tests: `6aff3042-1327-4f3d-82f0-40a157ac4464`. 

Backport of this PR will unblock frontend CI

### Issues Resolved

- Resolves: https://github.com/opensearch-project/security/issues/3604

### Testing
Existing tests.

### Check List

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
+++ b/src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
@@ -18,6 +18,7 @@ import java.nio.charset.StandardCharsets;
 import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
+import java.util.Base64;
 import java.util.Date;
 import java.util.List;
 import java.util.Optional;
@@ -242,7 +243,7 @@ class AuthTokenProcessorHandler {
         String exchangeKey = settings.get("exchange_key");
 
         if (!Strings.isNullOrEmpty(exchangeKey)) {
-            exchangeKey = padSecret(exchangeKey, JWSAlgorithm.HS512);
+            exchangeKey = padSecret(new String(Base64.getUrlDecoder().decode(exchangeKey), StandardCharsets.UTF_8), JWSAlgorithm.HS512);
 
             return new OctetSequenceKey.Builder(exchangeKey.getBytes(StandardCharsets.UTF_8)).algorithm(JWSAlgorithm.HS512)
                 .keyUse(KeyUse.SIGNATURE)
@@ -256,7 +257,7 @@ class AuthTokenProcessorHandler {
                 );
             }
 
-            String k = padSecret(jwkSettings.get("k"), JWSAlgorithm.HS512);
+            String k = padSecret(new String(Base64.getUrlDecoder().decode(jwkSettings.get("k")), StandardCharsets.UTF_8), JWSAlgorithm.HS512);
 
             return new OctetSequenceKey.Builder(k.getBytes(StandardCharsets.UTF_8)).algorithm(JWSAlgorithm.HS512)
                 .keyUse(KeyUse.SIGNATURE)
--- a/src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
+++ b/src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
@@ -257,7 +257,10 @@ class AuthTokenProcessorHandler {
                 );
             }
 
-            String k = padSecret(new String(Base64.getUrlDecoder().decode(jwkSettings.get("k")), StandardCharsets.UTF_8), JWSAlgorithm.HS512);
+            String k = padSecret(
+                new String(Base64.getUrlDecoder().decode(jwkSettings.get("k")), StandardCharsets.UTF_8),
+                JWSAlgorithm.HS512
+            );
 
             return new OctetSequenceKey.Builder(k.getBytes(StandardCharsets.UTF_8)).algorithm(JWSAlgorithm.HS512)
                 .keyUse(KeyUse.SIGNATURE)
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -99,21 +98,17 @@ class AuthTokenProcessorHandler {
...
 
-        if (samlSubjectKey == null || samlSubjectKey.length() == 0) {
+        if (samlSubjectKey == null || samlSubjectKey.isEmpty()) {
             // If subjectKey == null, get subject from the NameID element.
             // Thus, this is a valid configuration.
             samlSubjectKey = null;
         }
 
-        if (samlRolesSeparator == null || samlRolesSeparator.length() == 0) {
-            samlRolesSeparator = null;
```
DarshitChanpura, line 114: this assignment was unused.

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -128,12 +123,7 @@ Optional<SecurityResponse> handle(RestRequest restRequest) throws Exception {
                 sm.checkPermission(new SpecialPermission());
             }
 
-            return AccessController.doPrivileged(new PrivilegedExceptionAction<Optional<SecurityResponse>>() {
-                @Override
-                public Optional<SecurityResponse> run() throws SamlConfigException, IOException {
-                    return handleLowLevel(restRequest);
-                }
-            });
+            return AccessController.doPrivileged((PrivilegedExceptionAction<Optional<SecurityResponse>>) () -> handleLowLevel(restRequest));
```
DarshitChanpura, line 126: seems like a quick code cleanup and so included in this PR. Same pattern can be found in other files but that is out of scope of this PR.

src/test/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticatorTest.java
```diff
@@ -171,7 +171,7 @@ public void basicTest() throws Exception {
             .put(IDP_METADATA_URL, mockSamlIdpServer.getMetadataUri())
             .put("kibana_url", "http://wherever")
             .put("idp.entity_id", mockSamlIdpServer.getIdpEntityId())
-            .put("exchange_key", "abc")
+            .put("exchange_key", "6aff3042-1327-4f3d-82f0-40a157ac4464")
```
RyanL1997, line 174: why we need this btw?

DarshitChanpura: The tests didn't catch the error because they use `abc` as exchange_key. I've updated it to reflect the exchange key used in front-end tests: `6aff3042-1327-4f3d-82f0-40a157ac4464`

src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -252,7 +242,7 @@ JWK createJwkFromSettings(Settings settings, Settings jwtSettings) throws Except
         String exchangeKey = settings.get("exchange_key");
 
         if (!Strings.isNullOrEmpty(exchangeKey)) {
```
cwperks, line 244: Before the JWT library change, the CXF library assumed that the exchangeKey was Base64 encoded:

1. https://github.com/opensearch-project/security/blob/2.11/src/main/java/com/amazon/dlic/auth/http/saml/AuthTokenProcessorHandler.java#L257
2. https://github.com/apache/cxf/blob/main/rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jws/JwsUtils.java#L109-L112
3. https://github.com/apache/cxf/blob/main/rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/common/JoseUtils.java#L112-L114
4. https://github.com/apache/cxf/blob/main/rt/security/src/main/java/org/apache/cxf/rt/security/crypto/CryptoUtils.java#L669-L675

With the change to nimbus, the configured value should also be assumed to be base64 encoded. The nimbus call:

```
new OctetSequenceKey.Builder(exchangeKey.getBytes(StandardCharsets.UTF_8))
```

will perform Base64 encoded as part of the [constructor of the builder](https://github.com/felx/nimbus-jose-jwt/blob/master/src/main/java/com/nimbusds/jose/jwk/OctetSequenceKey.java#L179-L186) so the code above decodes the configured value, pads it (for backwards compatibility?) and then encodes again. 

The code to remove the Base64 decoding here would be needed for backwards compatibility so it cannot be removed.

DarshitChanpura: > The code to remove the Base64 decoding here would be needed for backwards compatibility so it cannot be removed.

I'm slightly confused. Are you suggesting we move forward with getUrlDecoder in place of getDecoder?

cwperks: Yes, `getUrlDecoder()` would be in line with the previous JWT functionality using cxf.

DarshitChanpura: I've also replaced with url decoder to keep it backwards compatible.

src/test/java/com/amazon/dlic/auth/http/saml/HTTPSamlAuthenticatorTest.java
cwperks, line 174: Can there be 2 separate tests to confirm key behavior in both scenarios?

DarshitChanpura: done

cwperks (approved):

RyanL1997 (approved):

Issue #3201, opened 2023-08-16
Limit Service Account permissions


peternied:
For the initial release permissions for service accounts are going to be locked down so they only can be used with explicitly granted system indexes

### Exit Criteria
- [x] modify authz workflow to filter out all permissions other than index permissions with system index grant
- [x] add/modify test case that confirms cluster-wide permissions are not accessible
- [x] add/modify test case where index permissions without system index grant is filtered out
- [x] add/modify test case where index permissions with system index grant is allowed (happy path)

scrawfor99:
Looks good to me!

scrawfor99:
[Triage] Thank you for filing this issue @peternied. This issue has clear expectations for action items (completing the check list) so going to mark as triaged.

cwperks:
> modify authz workflow to filter out all permissions other than index permissions with system index grant

@peternied What if this did not rely on a role at all and instead extension initialization included a `reserved_indices` setting that can be used on initialization for extension to reserve indices.

With this setting, the ExtensionsManager will have a mapping from `<extension_unique_id>` -> `reserved set of indices`.

Can the service account not be mapped to any roles at all and only carry the identity of the extension?

Within the SystemIndexAccessEvaluator it can lookup if a request is coming from an extension's service account and determine if the extension is making the request on only its set of reserved indices and no other indices or else fail the request.

The implementation of service account tokens I did in https://github.com/opensearch-project/opensearch-sdk-java/pull/892 to show the SDK integrated with security for extensions does this if you'd like to see an example.

peternied:
This task has already been triaged, pointed, and scheduled in the current sprint, I'm not sure what you are asking?

cwperks:
@peternied I'm concerned that it could be confusing to have a role like

```
extension_a_svc_account_role:
  reserved: true
  hidden: false
  static: true
  description: "Allow full access to all indices and all cluster APIs"
  cluster_permissions:
    - "*"
  index_permissions:
    - index_patterns:
        - "*"
        - ".opensearch-observability"
      allowed_actions:
        - "*"
        - "system:admin/system_index"
  tenant_permissions:
    - tenant_patterns:
        - "*"
      allowed_actions:
        - "kibana_all_write"
```

but effectively we introduce logic to whittle it down to 

```
extension_a_svc_account_role:
  reserved: true
  hidden: false
  static: true
  description: "Allow full access to all indices and all cluster APIs"
  index_permissions:
    - index_patterns:
[...]

PR #3607, opened 2023-10-26
Limit service account permission

RyanL1997:
### Description
Limit service account permission

Enhancement

### Issues Resolved
* Resolve https://github.com/opensearch-project/security/issues/3201

### Testing

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
@@ -128,6 +128,18 @@ public class ServiceAccountAuthenticationTest {
         }
     }
 
+    @Test
+    public void testReadBothWithServiceAccountCred() {
+        try (TestRestClient client = cluster.getRestClient(SERVICE_ACCOUNT_USER_NAME, DEFAULT_PASSWORD)) {
+            client.confirmCorrectCredentials(SERVICE_ACCOUNT_USER_NAME);
+            TestRestClient.HttpResponse response = client.get("test-non-sys-index,test-sys-index");
+            response.assertStatusCode(HttpStatus.SC_OK);
+            // TODO: CHECK IT ONLY READ SYSTEM INDEX AND FILTERED OUT REGULAR INDEX
+            // TODO: REMOVE THIS AND PARSING/CHECKING THE RESPONSE
+            System.out.println(response);
+        }
+    }
+
     // TODO: REMOVE THIS DEBUGGING TEST CASE
     @Test
     public void testReadNonSysIndexWithAdminCred() {
--- a/src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
@@ -1,0 +1,98 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.http;
+
+import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.hc.core5.http.HttpStatus;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.opensearch.test.framework.TestSecurityConfig;
+import org.opensearch.test.framework.cluster.ClusterManager;
+import org.opensearch.test.framework.cluster.LocalCluster;
+import org.opensearch.test.framework.cluster.TestRestClient;
+
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.opensearch.security.support.ConfigConstants.*;
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+import static org.opensearch.test.framework.TestSecurityConfig.Role.ALL_ACCESS;
+
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class ServiceAccountAuthenticationTest {
+
+    public static final String DEFAULT_PASSWORD = "secret";
+
+    static final TestSecurityConfig.User ADMIN_USER = new TestSecurityConfig.User("admin").roles(ALL_ACCESS);
+
+    public static final String SERVICE_ACCOUNT_USER_NAME = "admin-extension";
+
+    private static final TestSecurityConfig.Role SERVICE_ACCOUNT_ADMIN_ROLE = new TestSecurityConfig.Role("admin-extension-role")
+        .clusterPermissions("*")
+        .indexPermissions("*", "system:admin/system_index")
+        .on("*");
+
+    static final TestSecurityConfig.User SERVICE_ACCOUNT_ADMIN_USER = new TestSecurityConfig.User(SERVICE_ACCOUNT_USER_NAME).roles(
+        SERVICE_ACCOUNT_ADMIN_ROLE
+    ).attr("service", true);
+
+    @ClassRule
+    public static final LocalCluster cluster = new LocalCluster.Builder().clusterManager(ClusterManager.SINGLENODE)
+        .anonymousAuth(false)
+        .users(ADMIN_USER, SERVICE_ACCOUNT_ADMIN_USER)
+        .nodeSettings(
+            Map.of(
+                SECURITY_SYSTEM_INDICES_PERMISSIONS_ENABLED_KEY,
+                true,
+                SECURITY_ALLOW_DEFAULT_INIT_SECURITYINDEX,
+                true,
+                SECURITY_RESTAPI_ROLES_ENABLED,
+                List.of("user_admin__all_access")
+            )
+        )
+        .authc(AUTHC_HTTPBASIC_INTERNAL)
+        .build();
+
+    // TODO: REMOVE THIS DEBUGGING TEST CASE
+    @Test
+    public void testClusterHealthWithAdminCred() {
+        try (TestRestClient client = cluster.getRestClient("admin", DEFAULT_PASSWORD)) {
+            client.confirmCorrectCredentials("admin");
+            TestRestClient.HttpResponse response = client.get("_cluster/health");
+            response.assertStatusCode(HttpStatus.SC_OK);
+            System.out.println(response);
+        }
+    }
+
+    @Test
+    public void testClusterHealthWithServiceAccountCred() throws JsonProcessingException {
+        try (TestRestClient client = cluster.getRestClient("admin-extension", DEFAULT_PASSWORD)) {
+            client.confirmCorrectCredentials("admin-extension");
+            TestRestClient.HttpResponse response = client.get("_cluster/health");
+            response.assertStatusCode(HttpStatus.SC_FORBIDDEN);
+
+            String responseBody = response.getBody();
+            assertNotNull("Response body should not be null", responseBody);
+
+            ObjectMapper objectMapper = new ObjectMapper();
+            String typeField = objectMapper.readTree(responseBody).at("/error/root_cause/0/type").asText();
+
+            assertEquals("security_exception", typeField);
+        }
+    }
+}
--- a/src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
@@ -48,15 +48,16 @@ public class ServiceAccountAuthenticationTest {
     // CS-SUPPRESS-SINGLE: RegexpSingleline get Extensions Settings
     public static final String SERVICE_ACCOUNT_USER_NAME = "admin-extension";
     // CS-ENFORCE-SINGLE
-    private static final TestSecurityConfig.Role SERVICE_ACCOUNT_ADMIN_ROLE = new TestSecurityConfig.Role("admin-extension-role")
-        .clusterPermissions("*")
-        .indexPermissions("*", "system:admin/system_index")
-        .on("*");
 
     static final TestSecurityConfig.User SERVICE_ACCOUNT_ADMIN_USER = new TestSecurityConfig.User(SERVICE_ACCOUNT_USER_NAME).attr(
         SERVICE_ATTRIBUTE,
         "true"
-    ).roles(SERVICE_ACCOUNT_ADMIN_ROLE);
+    )
+        .roles(
+            new TestSecurityConfig.Role("admin-extension-role").clusterPermissions("*")
+                .indexPermissions("*", "system:admin/system_index")
+                .on("*")
+        );
 
     private static final TestIndex TEST_NON_SYS_INDEX = TestIndex.name("test-non-sys-index")
         .setting("index.number_of_shards", 1)
@@ -134,7 +135,7 @@ public class ServiceAccountAuthenticationTest {
     public void testReadBothWithServiceAccountCred() {
         TestRestClient client = cluster.getRestClient(SERVICE_ACCOUNT_USER_NAME, DEFAULT_PASSWORD);
         client.confirmCorrectCredentials(SERVICE_ACCOUNT_USER_NAME);
-        TestRestClient.HttpResponse response = client.get("test-non-sys-index,test-sys-index");
+        TestRestClient.HttpResponse response = client.get((TEST_SYS_INDEX.getName() + "," + TEST_NON_SYS_INDEX.getName()));
         response.assertStatusCode(HttpStatus.SC_OK);
 
         String responseBody = response.getBody();
--- a/src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
@@ -38,24 +38,18 @@ import static org.opensearch.test.framework.TestSecurityConfig.Role.ALL_ACCESS;
 @ThreadLeakScope(ThreadLeakScope.Scope.NONE)
 public class ServiceAccountAuthenticationTest {
 
-    public static final String DEFAULT_PASSWORD = "secret";
-
     public static final String SERVICE_ATTRIBUTE = "service";
 
     static final TestSecurityConfig.User ADMIN_USER = new TestSecurityConfig.User("admin").roles(ALL_ACCESS);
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -569,6 +576,11 @@ public PrivilegesEvaluatorResponse evaluate(
 
     }
 
+    public Boolean isServiceAccount(final User user) {
+        Map<String, String> userAttributesMap = user.getCustomAttributesMap();
+        return userAttributesMap != null && "true".equals(userAttributesMap.get("attr.internal.service"));
+    }
```
peternied, line 582: This seems like a function that should be on the User class.

RyanL1997: makes sense! I will refactor it into it after some the testing work im doing.

src/test/java/org/opensearch/security/IntegrationTests.java
```diff
@@ -331,7 +331,7 @@ public void testSpecialUsernames() throws Exception {
         setup();
         RestHelper rh = nonSslRestHelper();
 
-        Assert.assertEquals(HttpStatus.SC_OK, rh.executeGetRequest("", encodeBasicHeader("bug.99", "nagilum")).getStatusCode());
```
RyanL1997, line 334: This bug.99 was a service account btw, so we have to replace it with a non-service account, since service account is not allowed anymore

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -233,8 +233,18 @@ private void evaluateSystemIndicesAccess(
     ) {
         // Perform access check is system index permissions are enabled
         boolean containsSystemIndex = requestContainsAnySystemIndices(requestedResolved);
+        boolean serviceAccountUser = user.isServiceAccount();
 
         if (isSystemIndexPermissionEnabled) {
+            if (serviceAccountUser && !containsSystemIndex) {
+                auditLog.logSecurityIndexAttempt(request, action, task);
```
cwperks, line 240: @RyanL1997 How would this PR handle the following scenario?

Extension uses its service account token to search on 2 indices: 1 regular, 1 system index. i.e.

```
GET /regular-index,.system-index/_search
{
  "query": {
    "match": {
      "foo": "bar"
    }
  }
}
```

Its my understanding in the design from @scrawfor99 and @peternied that a service account would be conventionally tied to a role that is limited to only system indices and may look something like:

```
ext1role:
  description: "Extension1 Role"
  index_permissions:
    - index_patterns:
        - '.ext1-*'
        - '.system-index'
      allowed_actions:
        - "system:admin/system_index"
```

but what if the role was defined with regular indexes like:

```
ext1role:
  description: "Extension1 Role"
  index_permissions:
    - index_patterns:
        - '.ext1-*'
        - '.system-index'
        - 'regular-index*'
      allowed_actions:
        - "system:admin/system_index"
        - "read"
```

In the case of the second role would it be possible for the account to search on indices matching `regular-index*`?

RyanL1997: > In the case of the second role would it be possible for the account to search on indices matching regular-index*?

Hi @cwperks , thanks for the detailed feedback. I think for the above question, the answer is no. As long as this account is a service account, it wont be able to access to non-system index for this implementation.

Yesterday I did a manual testing with @DarshitChanpura yesterday, and the procedure was captured in this [issue in my fork](https://github.com/RyanL1997/security/issues/2). I think the procedure of what we did also covered this case since the role we were using for testing is having a `*` wild card index permission.

src/test/java/org/opensearch/security/IntegrationTests.java
```diff
@@ -1074,4 +1074,16 @@ public void testMonitorHealth() throws Exception {
         RestHelper rh = nonSslRestHelper();
         Assert.assertEquals(HttpStatus.SC_OK, rh.executeGetRequest("_cat/health", encodeBasicHeader("picard", "picard")).getStatusCode());
     }
+
+    // TODO: SOMETHING LIKE THIS
```
peternied, line 1078: I'd recommend creating the new tests inside of the the new integration test framework in a new file 'ServiceAccountTests` side by side with `OnBehalfOfJwtAuthenticationTest`, what do you think?

RyanL1997: Yes! This is what im doing haha. I was just leaving this comment for self-reminder

RyanL1997: Because some of the test setup here is not what we want.

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
```diff
@@ -0,0 +1,98 @@
...
+    public static final String SERVICE_ACCOUNT_USER_NAME = "admin-extension";
+
+    private static final TestSecurityConfig.Role SERVICE_ACCOUNT_ADMIN_ROLE = new TestSecurityConfig.Role("admin-extension-role")
+        .clusterPermissions("*")
+        .indexPermissions("*", "system:admin/system_index")
+        .on("*");
+
+    static final TestSecurityConfig.User SERVICE_ACCOUNT_ADMIN_USER = new TestSecurityConfig.User(SERVICE_ACCOUNT_USER_NAME).roles(
+        SERVICE_ACCOUNT_ADMIN_ROLE
+    ).attr("service", true);
```
peternied, line 52: Nit; Add a separate `Service` object as part of the configuration that handles this.

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
```diff
@@ -0,0 +1,98 @@
...
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class ServiceAccountAuthenticationTest {
+
+    public static final String DEFAULT_PASSWORD = "secret";
+
+    static final TestSecurityConfig.User ADMIN_USER = new TestSecurityConfig.User("admin").roles(ALL_ACCESS);
+
+    public static final String SERVICE_ACCOUNT_USER_NAME = "admin-extension";
+
+    private static final TestSecurityConfig.Role SERVICE_ACCOUNT_ADMIN_ROLE = new TestSecurityConfig.Role("admin-extension-role")
```
peternied, line 45: Nit: this role can be in-lined in the account creation

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
```diff
@@ -0,0 +1,98 @@
...
+                List.of("user_admin__all_access")
+            )
+        )
+        .authc(AUTHC_HTTPBASIC_INTERNAL)
+        .build();
+
+    // TODO: REMOVE THIS DEBUGGING TEST CASE
+    @Test
+    public void testClusterHealthWithAdminCred() {
+        try (TestRestClient client = cluster.getRestClient("admin", DEFAULT_PASSWORD)) {
```
peternied, line 74: ! You can pass the `SERVICE_ACCOUNT_ADMIN_USER` in directly

RyanL1997: Yes I was just using this example to test something, and this will be removed in future commit.

RyanL1997: Removed

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
RyanL1997: Actually you were correct @cwperks, and it didnt handle the case you provided. I was only check if the request `Boolean containSystemIndex`, so that if it is a mixed the request with both non-system index and system-index it will still be considered as true for the above condition. According to that, I have pushed another [commit](https://github.com/opensearch-project/security/pull/3607/commits/ac9416ee018c75b986f87233608055d98d3b12c9) to handle this 'mixed' case. So now, if there is a mixed request, it will filter out the regular index from the request and then only do the system index only. Thanks for pointing this out btw. I also added a test case for this, and the detailed parsing to prove it only do the system index request will be added later.

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
RyanL1997: Done

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
```diff
@@ -0,0 +1,162 @@
...
+        response.assertStatusCode(HttpStatus.SC_OK);
+
+        ObjectMapper objectMapper = new ObjectMapper();
+        JsonNode jsonResponse = objectMapper.readTree(response.getBody());
+
+        boolean hasTestSysIndex = jsonResponse.has("test-sys-index");
+        boolean hasTestNonSysIndex = jsonResponse.has("test-non-sys-index");
+
+        assertTrue("Response should contain 'test-sys-index'", hasTestSysIndex);
+        assertFalse("Response should not contain 'test-non-sys-index'", hasTestNonSysIndex);
```
DarshitChanpura, line 159: 👏

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+            assertEquals("security_exception", typeField);
+        }
+    }
+
+    @Test
+    public void testReadSysIndexWithServiceAccountCred() throws JsonProcessingException {
+        try (TestRestClient client = cluster.getRestClient(SERVICE_ACCOUNT_USER_NAME, DEFAULT_PASSWORD)) {
+            client.confirmCorrectCredentials(SERVICE_ACCOUNT_USER_NAME);
+            TestRestClient.HttpResponse response = client.get("test-sys-index");
```
DarshitChanpura, line 116: nit: you can use TEST_SYS_INDEX.getName() and TEST_NON_SYS_INDEX.getName() for all hard-coded instances

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -353,7 +353,15 @@ public PrivilegesEvaluatorResponse evaluate(
             namedXContentRegistry
         );
 
+        final boolean serviceAccountUser = user.isServiceAccount();
         if (isClusterPerm(action0)) {
+            if (serviceAccountUser) {
+                presponse.missingPrivileges.add(action0);
+                presponse.allowed = false;
+                log.info("{} is a service account which has no access to cluster level permission of {}.", user, action0);
```
DarshitChanpura, line 361: nit: `{} is a service account which doesn't have access to cluster level permission: {}` Thoughts?

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -199,6 +199,29 @@ private List<String> getAllProtectedSystemIndices(final Resolved requestedResolv
...
+        List<String> allProtectedSystemIndices = getAllProtectedSystemIndices(requestedResolved);
+
+        for (String index : allIndices) {
+            // Check if the index is not in the list of system or protected system indices
+            if (!allSystemIndices.contains(index) && !allProtectedSystemIndices.contains(index)) {
+                return true;
+            }
+        }
+
+        return false;
```
DarshitChanpura, line 222: another option:
```java
return allIndices.stream()
        .noneMatch(index -> allSystemIndices.contains(index) || allProtectedSystemIndices.contains(index));
```

DarshitChanpura (commented):
Looks good @RyanL1997. I've left some nits.

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
```diff
@@ -0,0 +1,162 @@
...
+            response.assertStatusCode(HttpStatus.SC_OK);
+
+            ObjectMapper objectMapper = new ObjectMapper();
+            JsonNode jsonResponse = objectMapper.readTree(response.getBody());
+
+            boolean hasTestSysIndex = jsonResponse.has("test-sys-index");
+            assertTrue("Response should contain 'test-sys-index'", hasTestSysIndex);
```
peternied, line 123: Please do not use ObjectMapper in test cases, instead use `getTextFromJsonBody` or related functions to author `assertThat` statements.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
RyanL1997: fixed

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
RyanL1997: applied

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
RyanL1997: Fixed

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
RyanL1997: Applied

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
RyanL1997: Thanks for the advice, but I think I will leave like this so that to have the same pattern for `admin` user creation.

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
```diff
@@ -0,0 +1,147 @@
...
+    public static final String SERVICE_ATTRIBUTE = "service";
+
+    static final TestSecurityConfig.User ADMIN_USER = new TestSecurityConfig.User("admin").roles(ALL_ACCESS);
+
+    // CS-SUPPRESS-SINGLE: RegexpSingleline get Extensions Settings
+    public static final String SERVICE_ACCOUNT_USER_NAME = "admin-extension";
+    // CS-ENFORCE-SINGLE
+    private static final TestSecurityConfig.Role SERVICE_ACCOUNT_ADMIN_ROLE = new TestSecurityConfig.Role("admin-extension-role")
+        .clusterPermissions("*")
+        .indexPermissions("*", "system:admin/system_index")
```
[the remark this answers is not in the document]
RyanL1997: I think, before this change, Service account was behaving like a regular user that its privilege is fully depending on its roles configuration/mapping. According to this, I think it will be up to a cluster admin to configure the role correctly.

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -233,8 +249,42 @@ private void evaluateSystemIndicesAccess(
...
+                    }
+                    presponse.allowed = false;
+                    presponse.markComplete();
+                    return;
+                } else if (containsSystemIndex && containsRegularIndex) {
+                    // Filter out regular indices
+                    List<String> systemIndices = getAllSystemIndices(requestedResolved);
+                    irr.replace(request, false, systemIndices.toArray(new String[0]));
+                    if (log.isDebugEnabled()) {
+                        log.debug("Filtered out regular indices, resulting list is {}", systemIndices);
```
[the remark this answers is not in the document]
RyanL1997: I was thinking about the scenario of the requester may not know the which index is/is not a system-index.(I may be overthinking about this btw)

What do you think about this one? Should we just reject? @peternied @scrawfor99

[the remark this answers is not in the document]
peternied: I will always vote for the most predictable behavior - which I believe in this case is failing the request.

When(TBD) we 'loosen' the permissions requirements, suddenly requests would be running with more access than they had before - which isn't clear to the administrator why the query changed.  Where as a request that was getting a 403 will start getting another return code seems much easier to understand IMO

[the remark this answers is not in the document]
RyanL1997: Ok that sounds good. I will switch this into the case that if there is a request contains mixed indices of system and non-system, it will be failed.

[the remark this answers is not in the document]
RyanL1997: Applied

src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java
```diff
@@ -233,8 +249,35 @@ private void evaluateSystemIndicesAccess(
...
+                                    && !getAllProtectedSystemIndices(requestedResolved).contains(index)
+                            )
+                            .collect(Collectors.toList());
+                        log.debug("Service account cannot access regular indices: {}", regularIndices);
+                    }
+                    presponse.allowed = false;
+                    presponse.markComplete();
+                    return;
+                }
+            }
```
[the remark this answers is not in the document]
RyanL1997: Fixed

DarshitChanpura (dismissed):
LGTM!

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+    public static final String DEFAULT_PASSWORD = "secret";
+
+    public static final String SERVICE_ATTRIBUTE = "service";
+
+    static final TestSecurityConfig.User ADMIN_USER = new TestSecurityConfig.User("admin").roles(ALL_ACCESS);
+
+    // CS-SUPPRESS-SINGLE: RegexpSingleline get Extensions Settings
+    public static final String SERVICE_ACCOUNT_USER_NAME = "admin-extension";
+    // CS-ENFORCE-SINGLE
```
peternied, line 49: Nit; No need to suppress if you remove the work extension, what do you think?

src/integrationTest/java/org/opensearch/security/http/ServiceAccountAuthenticationTest.java
```diff
@@ -0,0 +1,148 @@
...
+                List.of(TEST_SYS_INDEX.getName())
+            )
+        )
+        .authc(AUTHC_HTTPBASIC_INTERNAL)
+        .indices(TEST_NON_SYS_INDEX, TEST_SYS_INDEX)
+        .build();
+
+    @Test
+    public void testClusterHealthWithServiceAccountCred() {
+        try (TestRestClient client = cluster.getRestClient(SERVICE_ACCOUNT_USER_NAME, DEFAULT_PASSWORD)) {
```
peternied, line 95: There are additional locations as well, then you can delete the username/password variables.
```suggestion
        try (TestRestClient client = cluster.getRestClient(ADMIN_USER)) {
```

RyanL1997: done

DarshitChanpura (approved):

Issue #3581, opened 2023-10-20
[BUG] [Code Style] HttpStatus & Headers should not be imported from `org.apache.hc.core5.http`


peternied:
### What is the bug?
In this codebase on main, we are using a newer version of the apache http client; however, in our currently supported major version, 2.x we don't use this version of the http client and cannot due to dependency conflicts.  Any changes that reference `org.apache.hc.core5.http.HttpStatus` or `org.apache.hc.core5.http.Header` cannot be backported without manual conflict resolution.

### What is the expected behavior?
Our code hygiene workflows should catch usages of these namespace and prevent them from being checked into the codebase. It would be a bonus to fixing the issue by switching to `org.apache.http.HttpStatus` & `org.apache.http.Header` which will work in both the 2.x and mainline branches or provide the specific recommendation.

### Exit Criteria
- [ ] Add a new checkstyle or spotless rule that prevents the use of the `org.apache.hc.core5.http` package for HttpStatus and Header classes.
- [ ] Fix all existing references to use `org.apache.http.HttpStatus` & `org.apache.http.Header`
- [ ] [Optional] Update `spotlessApply` so that it can switch the references to the preferred replacement OR have the error message indicate how to resolve the issue by suggesting the correct replacement class.

cwperks:
@peternied Should` org.apache.http.HttpStatus` be used in the integrationTest package to make backporting to 2.x easier? The main branch has already upgraded to org.apache.HttpComponents v5 and v4 is used in 2.x. main still has a dependency on v4 because its needed by OpenSAML. See details in https://github.com/opensearch-project/security/issues/2932.

The library was updated in https://github.com/opensearch-project/security/pull/2166

cwperks:
With 3.0.0 around the corner and no future releases on the 2.x branch, I think this issue can be closed.

PR #3663, opened 2023-11-07
[BUG-3581] [Code Style] HttpStatus & Headers should not be imported from `org.apache.hc.core5.http`

prabhask5:
### Description

Accounts for refactor of Header and HttpStatus classes from hc.cor5.http to just http by adding checkstyle test, and manually refactoring all current occurrences of the illegal import. Additionally updated spotlessApply to make the refactor automatic on future cases of illegal import.

Bug fix/refactor

Refer to #3581 

Ideally nothing.

### Issues Resolved
- #3581 

No.

### Testing
Tests have been modified and should work before merging.

### Check List


prabhask5:
@cwperks @scrawfor99 @DarshitChanpura @peternied Through updating the illegal import `org.apache.hc.core5.http.Header`, to `org.apache.http.Header` I'm getting a lot of type compilation errors that try unsuccessfully to cast `org.apache.hc.core5.http.Header` to `org.apache.http.Header`. Here is an example of one such error: /Users/prabhask/Documents/Internships/opensearch-security/src/test/java/org/opensearch/security/test/helper/rest/RestHelper.java:238: error: incompatible types: org.apache.http.Header cannot be converted to org.apache.hc.core5.http.Header

To solve this problem I can either manually add a cast to `org.apache.hc.core5.http.Header` whenever this error appears or change the imports of all the dependent classes to also exclude core5 or client5, both of which seem like bad solutions. Is there another way to solve this problem?

reta:
> @cwperks @scrawfor99 @DarshitChanpura @peternied Through updating the illegal import `org.apache.hc.core5.http.Header`,

@prabhask5 why  `org.apache.hc.core5.http.Header` is illegal? The `main` branch uses Apache HttpClient 5 and should import these packages, the `2.x` uses Apache HttpClient 4.x and should import `org.apache.http.Header`, there should be no attempts to do any kind of type casting.

prabhask5:
> > @cwperks @scrawfor99 @DarshitChanpura @peternied Through updating the illegal import `org.apache.hc.core5.http.Header`,
> 
> @prabhask5 why `org.apache.hc.core5.http.Header` is illegal? The `main` branch uses Apache HttpClient 5 and should import these packages, the `2.x` uses Apache HttpClient 4.x and should import `org.apache.http.Header`, there should be no attempts to do any kind of type casting.

I'm referencing #3581 , if its not illegal I might be misunderstanding what the issue is saying.

reta:
> I'm referencing #3581 , if its not illegal I might be misunderstanding what the issue is saying.

Me neither, the `security` plugin uses `RestClient` which directly exports some of the Apache HttpClient 5 or 4.x APIs (depending on the branch).

prabhask5:
> @prabhask5 thanks for being proactive on this.
> 
> Similar to @reta:
> 
> I am not sure what the original issue was referencing. I don't think that it will be possible to standardize the versions without _significant_ refactoring. You will need to swap to the v4 equivalents for everything in main.
> 
> Eventually we will move to 3.0 and likely swap completely to v5 when we do.
> 
> I don't think it makes sense to spend too much time converting the versions back and forth.

So is this issue irrelevant then? @peternied Do you have some insight on this?

peternied:
Illegal has weighty connotations - I'd prefer to think of direct imports of `...hc.core5...` as _not advisable_.  The rational is that it makes our great auto-backporting tool not work for nearly any case where we are authoring test cases (which I hope is often!).  There is more context in the issue [1].

@reta What do you think about this rational? After we release 3.0.0, we will still be supporting 2.x builds for a least a year, anything that helps make the backport process result in less merge conflicts is a solid win to me.  

- [1] https://github.com/opensearch-project/security/issues/3581

peternied:
@prabhask5 Looks like there are many merge conflicts that you'll need to resolve - I'd suggest a `git merge origin/main` YMMV

prabhask5:
> @prabhask5 Looks like there are many merge conflicts that you'll need to resolve - I'd suggest a `git merge origin/main` YMMV

@peternied Sounds good. How do you recommend fixing the errors I'm getting? I have them copied below for convenience.

Through updating the illegal import org.apache.hc.core5.http.Header, to org.apache.http.Header I'm getting a lot of type compilation errors that try unsuccessfully to cast org.apache.hc.core5.http.Header to org.apache.http.Header. Here is an example of one such error: /Users/prabhask/Documents/Internships/opensearch-security/src/test/java/org/opensearch/security/test/helper/rest/RestHelper.java:238: error: incompatible types: org.apache.http.Header cannot be converted to org.apache.hc.core5.http.Header

To solve this problem I can either manually add a cast to org.apache.hc.core5.http.Header whenever this error appears or change the imports of all the dependent classes to also exclude core5 or client5, both of which seem like bad solutions. Is there another way to solve this problem?

peternied:
`org.apache.hc.core5.http.Header` and `org.apache.http.Header` are different implementations, they won't translate between one another.  There are a couple of options to handle this:
1) Don't attempt to convert these objects.  While it isn't idea,l if the majority of the time we are using status codes and header constants ARE converted, that solves the majority problem.
2) Create a shared interface / use another object at object declartions.  Headers aren't very complex, it would be relatively easy to create or use another object that represented this key value pairing, then convert it in the RestHelper where it needs to be type specific.
3) An even better solution some invents.

prabhask5:
> `org.apache.hc.core5.http.Header` and `org.apache.http.Header` are different implementations, they won't translate between one another. There are a couple of options to handle this:
> 
> 1. Don't attempt to convert these objects.  While it isn't idea,l if the majority of the time we are using status codes and header constants ARE converted, that solves the majority problem.
> 2. Create a shared interface / use another object at object declartions.  Headers aren't very complex, it would be relatively easy to create or use another object that represented this key value pairing, then convert it in the RestHelper where it needs to be type specific.
> 3. An even better solution some invents.

@peternied I'm pretty confused on these solutions. The majority of the errors I'm getting are either functions where the new Header is not an applicable type for the parameter, but I can't edit these functions since they are in the Apache codebase. How should I fix these errors? The only solution I'm seeing is to manually cast to the new type of Header, which again I don't think is advisable. A more specific solution would help a lot, ty!

peternied:
@prabhask5 I might recommend going with the first option, remove changes for `Headers` for this PR.  After the status codes have been aligned take a single file and see what it takes to get it working with a draft PR.

This change is already massive and providing feedback is non-trivial if you are running into build/runtime issues.

prabhask5:
> @prabhask5 I might recommend going with the first option, remove changes for `Headers` for this PR. After the status codes have been aligned take a single file and see what it takes to get it working with a draft PR.
> 
> This change is already massive and providing feedback is non-trivial if you are running into build/runtime issues.

Done! Feel free to unblock tests.

src/integrationTest/java/org/opensearch/security/http/JwtAuthenticationTests.java
```diff
@@ -16,6 +16,8 @@
 import java.util.Map;
 
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
+import io.jsonwebtoken.SignatureAlgorithm;
+import io.jsonwebtoken.security.Keys;
 import org.apache.hc.core5.http.Header;
 import org.apache.hc.core5.http.message.BasicHeader;
```
DarshitChanpura, line 22: this seems to be unrelated. The original import for hc5 still remains as is

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
```diff
@@ -21,8 +21,12 @@
 import javax.crypto.SecretKey;
 
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
+import io.jsonwebtoken.Claims;
+import io.jsonwebtoken.Jwts;
+import io.jsonwebtoken.security.Keys;
+
 import org.apache.hc.core5.http.Header;
```
DarshitChanpura, line 28: why only 1 of these 2 changed?

src/integrationTest/java/org/opensearch/security/rest/CompressionTests.java
```diff
@@ -20,7 +20,7 @@
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
 import org.apache.hc.client5.http.classic.methods.HttpPost;
 import org.apache.hc.core5.http.ContentType;
-import org.apache.hc.core5.http.HttpStatus;
+import org.apache.http.HttpStatus;
```
DarshitChanpura, line 23: should this be a mixed usage between 4 & 5?

src/integrationTest/java/org/opensearch/security/rest/WhoAmITests.java
```diff
@@ -22,7 +22,8 @@
 import java.util.stream.Collectors;
 
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
-import org.apache.hc.core5.http.HttpStatus;
+import joptsimple.internal.Strings;
```
DarshitChanpura, line 25: what is this import? is it being used?

src/test/java/org/opensearch/security/system_indices/AbstractSystemIndicesTests.java
```diff
@@ -15,6 +15,7 @@
 import java.util.List;
 
 import org.apache.hc.core5.http.Header;
+
```
DarshitChanpura, line 18: nit: seems like this import was not replaced

src/test/java/org/opensearch/security/system_indices/SystemIndexDisabledTests.java
```diff
@@ -11,10 +11,8 @@
 
 package org.opensearch.security.system_indices;
 
-import java.io.IOException;
-
 import org.apache.hc.core5.http.Header;
```
DarshitChanpura, line 14: should this be org.apache.http.HttpStatus

DarshitChanpura (commented):
I see mixed usages of 4 & 5. Also some classes still import Header from 5 vs 4, is that expected?
@reta thoughts?

prabhask5:
Closed PR due to git issues new PR here: https://github.com/opensearch-project/security/pull/3741

Issue #3633, opened 2023-10-31
Extracting `install_demo_configuration` script to a standalone Java Tool


DarshitChanpura:
**Description:**
This issue is created to track the process of extracting all the logic from the `install_demo_configuration` script and transferring it to a standalone Java tool.

**What's Happening with the Script?**
The script will serve as a wrapper to call the Java tool. This approach will ensure minimal changes to the current behavior when running the demo script. Additionally, it will allow the script to use the bundled JDK (similar to the OpenSearch process) if no JAVA_HOME is set, thereby preventing script failures. Without this wrapper, the newly created Java tool would require JAVA_HOME to be set, potentially causing issues in certain distributions (such as rpm/deb).

**What Will the New Tool Look Like?**
The behavior of new tool would be consistent with the script. The new tool will contain all of the current individual components, including generating certificates, updating the opensearch.yml file, and modifying the admin password in internal_users.yml, among others. These components will serve as the foundation for future enhancements, particularly focusing on addressing static secrets.

**Why a Standalone Java Tool?**
1. **Reusability:** The requirement to generate a new random password can be fulfilled using tools like passay, which is currently employed for the service-account capability. This tool can be leveraged to generate a new password for the admin user. Similarly, the existing WeakPasswordValidation framework can be used to validate the provided password to ensure it is not weak.
2. **Avoiding Duplication:** Currently, the logic for installing the demo configuration exists in both .sh and .bat files, leading to redundancy. Creating a Java tool will serve as a single source of truth for the demo config setup, eliminating this redundancy.

**Acceptance Criteria:**

- [x] A new password is generated if not provided, and the generated password is logged in the executing terminal (stdout).
- [x] The script should accept a new parameter environment `-t` to reflect that this script is going to be used in `test` environment. Default value is set to `demo`
- [x] When the environment is set to `demo` (default), and a custom password is provided:
    - [x] It must be validated to ensure it is not a weak password:
        - [x] The password should be at least 8 characters long.
        - [x] The password should not contain the username.
- [x] When the environment is set to `test`, the password validation should be skipped.

scrawfor99:
[Triage] Hi @DarshitChanpura, thank you for filing this issue. This issue has all the right info to be marked triaged :)

PR #3449, opened 2023-10-04
[Bug Fix] Fix the demo configuration script and remove the admin credential from `internal_user.yml` 

RyanL1997:
### Description
By working with @DarshitChanpura for addressing the following issue: 
> https://github.com/opensearch-project/opensearch-build/issues/4094#issuecomment-1745785846

We need the initiation of the cluster to be failed once we encountered a failure of our demo configuration script. This PR introduces the method that if we failed to set up a admin credential during the set up of demo configuration script, it will lead to an absent of admin credential section in `internal_users.yml`, so that the initiation of the cluster will be failed.

* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation)
Bug fix
* What is the old behavior before changes and new behavior after changes?
The cluster initiation will be succeed and keep using the `admin:admin` credential even if the there is an hard exit on the set up on demo configuration script.

### Issues Resolved
* Relate https://github.com/opensearch-project/opensearch-build/issues/4094
* Relate https://github.com/opensearch-project/security/pull/3329

### Testing
CI

### Check List
- [x] New functionality includes testing
- [ ] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3669, opened 2023-11-08
Extracts demo configuration script to a java tool

DarshitChanpura:
### Description

This change extracts the core demo configuration install script to java tool to allow re-usability while leveraging that OpenSearch itself is JAVA process. All other tools shipped by security are java tools, and now with this PR, demo configuration setup will also be a java tool.

* Category : Enhancement, Refactoring

The need to implement these changes arose from the requirement to replace default admin credentials with a custom provided admin password.

The script execution behavior has changed slightly. Here's how:
1. Script now also validates custom provided password, and fails if it doesn't meet minimum requirements
2. Script now accepts a new option `-t` to skip password validation for test environments.
3. Previously, the script would fail if no password was provided. Now, the script would generate and set a random password instead.

### Issues Resolved
- Resolves https://github.com/opensearch-project/security/issues/3633

### Testing

Automated testing will be implemented in a follow-up PR.

Manual testing.

Tests:

<details>
<summary>1. Works without any options: </summary>

```console
➜  tools ./install_demo_configuration.sh
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
Install demo certificates? [y/N] y
Initialize Security Modules? [y/N] y
Cluster mode requires maybe additional setup of:
  - Virtual memory (vm.max_map_count)

Enable cluster mode? [y/N] N
OpenSearch install type: .tar.gz on Mac OS X 13.6 aarch64
OpenSearch config dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/
OpenSearch config file: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch.yml
OpenSearch bin dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/bin/
OpenSearch plugins dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/
OpenSearch lib dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/lib/
Detected OpenSearch Version: x-content-3.0.0-SNAPSHOT
Detected OpenSearch Security Version: 3.0.0.0-SNAPSHOT
No custom admin password found. Generating a new password now.
	***************************************************
		ADMIN PASSWORD SET TO: I:0―₽″h9¢
	***************************************************
### Success
### Execute this script now on all your nodes and then start all nodes
### OpenSearch Security will be automatically initialized.
### If you like to change the runtime configuration
### change the files in ../../../config/opensearch-security and execute:
sudo "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/opensearch-security/tools/securityadmin.sh" -cd "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch-security" -icl -key "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk-key.pem" -cert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk.pem" -cacert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/root-ca.pem" -nhnv
### or run ./securityadmin_demo.sh
### To use the Security Plugin ConfigurationGUI
### To access your secured cluster open https://<hostname>:<HTTP port> and log in with admin/admin.
### (Ignore the SSL certificate warning because we installed self-signed demo certificates)
➜  tools
```
</details>

2. Works as before with existing options supplied
<details> 
<summary>a. `-h`</summary>

```console
➜  tools ./install_demo_configuration.sh -h
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
install_demo_configuration.sh [-y] [-i] [-c]
  -h show help
  -y confirm all installation dialogues automatically
  -i initialize Security plugin with default configuration (default is to ask if -y is not given)
  -c enable cluster mode by binding to all network interfaces (default is to ask if -y is not given)
  -s skip updates if config is already applied to opensearch.yml
  -t set the execution environment to `test` to skip password validation. Should be used only for testing. (default is set to `demo`)
```
</details>

<details> 
<summary>b. `-y`</summary>

```console
➜  tools ./install_demo_configuration.sh -y
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
OpenSearch install type: .tar.gz on Mac OS X 13.6 aarch64
OpenSearch config dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/
OpenSearch config file: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch.yml
OpenSearch bin dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/bin/
OpenSearch plugins dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/
OpenSearch lib dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/lib/
Detected OpenSearch Version: x-content-3.0.0-SNAPSHOT
Detected OpenSearch Security Version: 3.0.0.0-SNAPSHOT
No custom admin password found. Generating a new password now.
	***************************************************
		ADMIN PASSWORD SET TO: ‛5ij¼₢₣Wvcq
	***************************************************
### Success
### Execute this script now on all your nodes and then start all nodes
### After the whole cluster is up execute:
sudo "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/opensearch-security/tools/securityadmin.sh" -cd "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch-security" -icl -key "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk-key.pem" -cert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk.pem" -cacert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/root-ca.pem" -nhnv
### or run ./securityadmin_demo.sh
### After that you can also use the Security Plugin ConfigurationGUI
### To access your secured cluster open https://<hostname>:<HTTP port> and log in with admin/admin.
### (Ignore the SSL certificate warning because we installed self-signed demo certificates)
```
</details>

<details> 
<summary>c. `-i`</summary>

```console
➜  tools ./install_demo_configuration.sh -i
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
Install demo certificates? [y/N] y
Cluster mode requires maybe additional setup of:
  - Virtual memory (vm.max_map_count)

Enable cluster mode? [y/N] y
OpenSearch install type: .tar.gz on Mac OS X 13.6 aarch64
OpenSearch config dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/
OpenSearch config file: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch.yml
OpenSearch bin dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/bin/
OpenSearch plugins dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/
OpenSearch lib dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/lib/
Detected OpenSearch Version: x-content-3.0.0-SNAPSHOT
Detected OpenSearch Security Version: 3.0.0.0-SNAPSHOT
No custom admin password found. Generating a new password now.
	***************************************************
		ADMIN PASSWORD SET TO: ½1S\‰…†²°9lI¿‡b
	***************************************************
### Success
### Execute this script now on all your nodes and then start all nodes
### OpenSearch Security will be automatically initialized.
### If you like to change the runtime configuration
### change the files in ../../../config/opensearch-security and execute:
sudo "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/opensearch-security/tools/securityadmin.sh" -cd "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch-security" -icl -key "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk-key.pem" -cert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk.pem" -cacert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/root-ca.pem" -nhnv
### or run ./securityadmin_demo.sh
### To use the Security Plugin ConfigurationGUI
### To access your secured cluster open https://<hostname>:<HTTP port> and log in with admin/admin.
### (Ignore the SSL certificate warning because we installed self-signed demo certificates)
➜  tools
```
</details>

<details> 
<summary>d. `-c`</summary>

```console
➜  tools ./install_demo_configuration.sh -c
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
Install demo certificates? [y/N] y
Initialize Security Modules? [y/N] y
OpenSearch install type: .tar.gz on Mac OS X 13.6 aarch64
OpenSearch config dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/
OpenSearch config file: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch.yml
OpenSearch bin dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/bin/
OpenSearch plugins dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/
OpenSearch lib dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/lib/
Detected OpenSearch Version: x-content-3.0.0-SNAPSHOT
Detected OpenSearch Security Version: 3.0.0.0-SNAPSHOT
No custom admin password found. Generating a new password now.
	***************************************************
		ADMIN PASSWORD SET TO: 2»jb'”1″BK
	***************************************************
### Success
### Execute this script now on all your nodes and then start all nodes
### OpenSearch Security will be automatically initialized.
### If you like to change the runtime configuration
### change the files in ../../../config/opensearch-security and execute:
sudo "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/opensearch-security/tools/securityadmin.sh" -cd "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch-security" -icl -key "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk-key.pem" -cert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk.pem" -cacert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/root-ca.pem" -nhnv
### or run ./securityadmin_demo.sh
### To use the Security Plugin ConfigurationGUI
### To access your secured cluster open https://<hostname>:<HTTP port> and log in with admin/admin.
### (Ignore the SSL certificate warning because we installed self-signed demo certificates)
➜  tools
```
</details>

<details> 
<summary>e. `-s`</summary>

```console
➜  tools ./install_demo_configuration.sh -s
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
Install demo certificates? [y/N] y
Initialize Security Modules? [y/N] y
Cluster mode requires maybe additional setup of:
  - Virtual memory (vm.max_map_count)

Enable cluster mode? [y/N] N
OpenSearch install type: .tar.gz on Mac OS X 13.6 aarch64
OpenSearch config dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/
OpenSearch config file: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch.yml
OpenSearch bin dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/bin/
OpenSearch plugins dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/
OpenSearch lib dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/lib/
Detected OpenSearch Version: x-content-3.0.0-SNAPSHOT
Detected OpenSearch Security Version: 3.0.0.0-SNAPSHOT
/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch.yml seems to be already configured for Security. Quit.
```
</details>

<details>
<summary>3. Sets the custom admin password, if one is supplied</summary>

```console
➜  tools export initialAdminPassword=43a5knwZKGmv
➜  tools ./install_demo_configuration.sh
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
Install demo certificates? [y/N] y
Initialize Security Modules? [y/N] y
Cluster mode requires maybe additional setup of:
  - Virtual memory (vm.max_map_count)

Enable cluster mode? [y/N] N
OpenSearch install type: .tar.gz on Mac OS X 13.6 aarch64
OpenSearch config dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/
OpenSearch config file: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch.yml
OpenSearch bin dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/bin/
OpenSearch plugins dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/
OpenSearch lib dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/lib/
Detected OpenSearch Version: x-content-3.0.0-SNAPSHOT
Detected OpenSearch Security Version: 3.0.0.0-SNAPSHOT
	***************************************************
		ADMIN PASSWORD SET TO: 43a5knwZKGmv
	***************************************************
### Success
### Execute this script now on all your nodes and then start all nodes
### OpenSearch Security will be automatically initialized.
### If you like to change the runtime configuration
### change the files in ../../../config/opensearch-security and execute:
sudo "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/opensearch-security/tools/securityadmin.sh" -cd "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch-security" -icl -key "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk-key.pem" -cert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk.pem" -cacert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/root-ca.pem" -nhnv
### or run ./securityadmin_demo.sh
### To use the Security Plugin ConfigurationGUI
### To access your secured cluster open https://<hostname>:<HTTP port> and log in with admin/admin.
### (Ignore the SSL certificate warning because we installed self-signed demo certificates)
➜  tools export initialAdminPassword=admin
➜  tools
➜  tools
➜  tools
➜  tools
➜  tools
➜  tools
➜  tools export initialAdminPassword=admin
➜  tools ./install_demo_configuration.sh
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
Install demo certificates? [y/N] y
Initialize Security Modules? [y/N] y
Cluster mode requires maybe additional setup of:
  - Virtual memory (vm.max_map_count)

Enable cluster mode? [y/N] N
OpenSearch install type: .tar.gz on Mac OS X 13.6 aarch64
OpenSearch config dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/
OpenSearch config file: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch.yml
OpenSearch bin dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/bin/
OpenSearch plugins dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/
OpenSearch lib dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/lib/
Detected OpenSearch Version: x-content-3.0.0-SNAPSHOT
Detected OpenSearch Security Version: 3.0.0.0-SNAPSHOT
Password admin is weak. Please re-try with a stronger password.
```
</details>

<details>
<summary> 4. Generates a random password, if one is not supplied</summary>

```console
➜  tools ./install_demo_configuration.sh
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
Install demo certificates? [y/N] y
Initialize Security Modules? [y/N] y
Cluster mode requires maybe additional setup of:
  - Virtual memory (vm.max_map_count)

Enable cluster mode? [y/N] N
OpenSearch install type: .tar.gz on Mac OS X 13.6 aarch64
OpenSearch config dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/
OpenSearch config file: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch.yml
OpenSearch bin dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/bin/
OpenSearch plugins dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/
OpenSearch lib dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/lib/
Detected OpenSearch Version: x-content-3.0.0-SNAPSHOT
Detected OpenSearch Security Version: 3.0.0.0-SNAPSHOT
No custom admin password found. Generating a new password now.
	***************************************************
		ADMIN PASSWORD SET TO: YW¢+(7₦a₸
	***************************************************
### Success
### Execute this script now on all your nodes and then start all nodes
### OpenSearch Security will be automatically initialized.
### If you like to change the runtime configuration
### change the files in ../../../config/opensearch-security and execute:
sudo "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/opensearch-security/tools/securityadmin.sh" -cd "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch-security" -icl -key "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk-key.pem" -cert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk.pem" -cacert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/root-ca.pem" -nhnv
### or run ./securityadmin_demo.sh
### To use the Security Plugin ConfigurationGUI
### To access your secured cluster open https://<hostname>:<HTTP port> and log in with admin/admin.
### (Ignore the SSL certificate warning because we installed self-signed demo certificates)
➜  tools
```
</details>

<details>
<summary>5. Accepts a new option `-t` to disable password validation for test environment</summary>

```console
➜  tools export initialAdminPassword=admin
➜  tools ./install_demo_configuration.sh -y
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
OpenSearch install type: .tar.gz on Mac OS X 13.6 aarch64
OpenSearch config dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/
OpenSearch config file: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch.yml
OpenSearch bin dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/bin/
OpenSearch plugins dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/
OpenSearch lib dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/lib/
Detected OpenSearch Version: x-content-3.0.0-SNAPSHOT
Detected OpenSearch Security Version: 3.0.0.0-SNAPSHOT
Password admin is weak. Please re-try with a stronger password.
➜  tools ./install_demo_configuration.sh -y -t
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
OpenSearch install type: .tar.gz on Mac OS X 13.6 aarch64
OpenSearch config dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/
OpenSearch config file: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch.yml
OpenSearch bin dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/bin/
OpenSearch plugins dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/
OpenSearch lib dir: /Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/lib/
Detected OpenSearch Version: x-content-3.0.0-SNAPSHOT
Detected OpenSearch Security Version: 3.0.0.0-SNAPSHOT
	***************************************************
		ADMIN PASSWORD SET TO: admin
	***************************************************
### Success
### Execute this script now on all your nodes and then start all nodes
### After the whole cluster is up execute:
sudo "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/plugins/opensearch-security/tools/securityadmin.sh" -cd "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/opensearch-security" -icl -key "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk-key.pem" -cert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/kirk.pem" -cacert "/Users/dchanp/Documents/amazon/testing-playground/test/binaries/opensearch-3.0.0-SNAPSHOT/config/root-ca.pem" -nhnv
### or run ./securityadmin_demo.sh
### After that you can also use the Security Plugin ConfigurationGUI
### To access your secured cluster open https://<hostname>:<HTTP port> and log in with admin/admin.
### (Ignore the SSL certificate warning because we installed self-signed demo certificates)
➜  tools
```
</details>

### Check List


DarshitChanpura:
CI `plugin-install` [failed](https://github.com/opensearch-project/security/actions/runs/6801420762/job/18492232096?pr=3669) with:
```


OpenSearch Security Demo Installer
** Warning: Do not use on production or public reachable systems **
Install demo certificates? [y/N] Initialize Security Modules? [y/N] Cluster mode requires maybe additional setup of:
  - Virtual memory (vm.max_map_count)

Enable cluster mode? [y/N] OpenSearch install type: .tar.gz on Linux 6.2.0-1015-azure amd64
OpenSearch config dir: /home/runner/work/security/security/opensearch-3.0.0-SNAPSHOT/config/
OpenSearch config file: /home/runner/work/security/security/opensearch-3.0.0-SNAPSHOT/config/opensearch.yml
OpenSearch bin dir: /home/runner/work/security/security/opensearch-3.0.0-SNAPSHOT/bin/
OpenSearch plugins dir: /home/runner/work/security/security/opensearch-3.0.0-SNAPSHOT/plugins/
OpenSearch lib dir: /home/runner/work/security/security/opensearch-3.0.0-SNAPSHOT/lib/
Detected OpenSearch Version: compress-3.0.0-SNAPSHOT
Detected OpenSearch Security Version: 3.0.0.0-SNAPSHOT
Password continuous-silver is weak. Please re-try with a stronger password.
```
Script is doing what it is supposed to. I will update the CI task to pass option `-t` and that should skip password-validation

src/main/java/org/opensearch/security/tools/InstallDemoConfiguration.java
```diff
@@ -0,0 +1,757 @@
...
+        printScriptHeaders();
+        readArguments(args);
+        gatherUserInputs();
+        initializeVariables();
+        printVariables();
+        checkIfSecurityPluginIsAlreadyConfigured();
+        setAdminPassword();
+        createDemoCertificates();
+        writeSecurityConfigToOpenSearchYML();
+        runSecurityAdminCommands();
```
peternied, line 79: What do you think about break these responsibilities into logical components and passing data between components.  I think changes like this could make authoring tests easier, what do you think?

DarshitChanpura: i'm not sure I understand. These made more sense to me as they use some variables that are declared at class-level and initialize them in initializeVariables(). For tests, I can instantiate those variables in each individual tests as needed. Do you have an example in mind for breaking these down?

peternied: Here is a good stack overflow [Q&A [1]](https://softwareengineering.stackexchange.com/questions/17100/clarify-the-single-responsibility-principle) about Single Responsibility Principle that I think applies here.

- [1] https://softwareengineering.stackexchange.com/questions/17100/clarify-the-single-responsibility-principle

DarshitChanpura:
Spotbugs gradle check will be fixed once https://github.com/opensearch-project/security/pull/3676 is merged

src/main/java/org/opensearch/security/tools/InstallDemoConfiguration.java
scrawfor99: In short, I think Peter is suggesting you break each of these operations out further into their own classes. While the original script had all of these functions in it, I think that was more of a result of the medium (a shell script) vs. best option.

Unless you are making a static methods class for a functional programming design, I think the idea would be something along the lines of:

```
Class passwordSetter() {

1. read in user password from env
2. read in from file 
3. generate random 
4. getPassword
}

Class certificateGenerator() {
1. read in settings
2. generate certs
3. getCerts
}

etc.

```


The idea is basically that each class has only a single thing that would cause it to be rewritten. For the certificateGenerator, you can imagine that it may need to be rewritten if we used a different format of certificates. But otherwise it would probably stay the same. For the passwordSetter, maybe it would be rewritten if we supported more ways of getting a password. Otherwise it probably will stay the same.

I don't necessarily buy into all of these programming philosophies and gospels but I do think that some like this can be helpful in moderation. 

TL;DR split the code into classes where each class does 1 thing and is only likely to be rewritten in 1 scenario

src/main/java/org/opensearch/security/tools/InstallDemoConfiguration.java
```diff
@@ -0,0 +1,858 @@
...
+        printVariables();
+        checkIfSecurityPluginIsAlreadyConfigured();
+        setAdminPassword();
+        createDemoCertificates();
+        writeSecurityConfigToOpenSearchYML();
+        finishScriptExecution();
+    }
+
+    /**
+     * Prints deprecation warning and other headers for the scrip
```
scrawfor99, line 92:
```suggestion
     * Prints deprecation warning and other headers for the script
```

src/main/java/org/opensearch/security/tools/InstallDemoConfiguration.java
```diff
@@ -0,0 +1,858 @@
...
+                if (!confirmAction(scanner, "Install demo certificates?")) {
+                    System.exit(0);
+                }
+
+                if (!initsecurity) {
+                    initsecurity = confirmAction(scanner, "Initialize Security Modules?");
+                }
+
+                if (!cluster_mode) {
+                    System.out.println("Cluster mode requires maybe additional setup of:");
```
scrawfor99, line 172: Nit: Remove "maybe"

src/main/java/org/opensearch/security/tools/InstallDemoConfiguration.java
DarshitChanpura: The original idea behind this implementation was to pick-up the script as is and translate this to java-tool. I had originally followed this philosophy, however I have made some modifications to make it modular.

DarshitChanpura:
@cwperks @RyanL1997 @scrawfor99 @peternied Could I get some reviews on this ?

tools/install_demo_configuration.sh
```diff
@@ -21,455 +16,15 @@ else
...
+elif [ ! -z "$JAVA_HOME" ]; then
+    BIN_PATH="$JAVA_HOME/bin/java"
 else
-    echo "### OpenSearch Security will be automatically initialized."
-    echo "### If you like to change the runtime configuration "
-    echo "### change the files in ../../../config/opensearch-security and execute: "
-	$SUDO_CMD cat securityadmin_demo.sh | tail -1
-	echo "### or run ./securityadmin_demo.sh"
-	echo "### To use the Security Plugin ConfigurationGUI"
+    echo "WARNING: nor OPENSEARCH_JAVA_HOME nor JAVA_HOME is set, will use $(which $BIN_PATH)"
```
peterzhuamazon, line 27: This should just be clear and show `which java` instead of `which $BIN_PATH` as it is `java` anyway.
You should also fail the script if you reach this `else` branch and still could not find `java` with `command -v`.

peterzhuamazon: Use `if command -v java >/dev/null` to see if java even exist in this else branch, if not, fail.

DarshitChanpura: Would we run into problems with rpm/deb by failing script?

tools/install_demo_configuration.sh
```diff
@@ -21,455 +16,15 @@ else
...
-$SUDO_CMD chmod +x securityadmin_demo.sh
+BIN_PATH="java"
 
-if [ "$initsecurity" == 0 ]; then
-	echo "### After the whole cluster is up execute: "
-	$SUDO_CMD cat securityadmin_demo.sh | tail -1
-	echo "### or run ./securityadmin_demo.sh"
-    echo "### After that you can also use the Security Plugin ConfigurationGUI"
+# now set the path to java: first OPENSEARCH_JAVA_HOME, then JAVA_HOME
+if [ ! -z "$OPENSEARCH_JAVA_HOME" ]; then
```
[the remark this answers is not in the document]
DarshitChanpura: I followed the same pattern as other scripts in this tools folder. I can definitely update it if you think its a better solution since you are the expert in scripts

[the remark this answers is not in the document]
peterzhuamazon: Yes, they both work but we should definitely use `-n` since `-z` is checking if it is zero and `-n` is checking if it is non-zero so no need to add a `!` just to check opposite.

tools/install_demo_configuration.sh
peterzhuamazon: I would rather let it fail hard if there is no `java` found in path at all.
Also when you reach that third branch it means no OS_JAVA_HOME or JAVA_HOME anyway.

On deb/rpm there will always be a OS_JAVA_HOME there, so it will always goes to the 1st branch.
If we failed on the third branch we know this is an issue with core not bundling jdk, not your script.

In my opinion that is a even better scenario.
Thanks.

DarshitChanpura: ahh, I see. I will update the script. Ty!

tools/install_demo_configuration.sh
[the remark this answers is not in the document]
DarshitChanpura: okay.. I'll push that change.

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
```diff
@@ -0,0 +1,377 @@
...
+     */
+    static void showHelp() {
+        System.out.println("install_demo_configuration.sh [-y] [-i] [-c]");
+        System.out.println("  -h show help");
+        System.out.println("  -y confirm all installation dialogues automatically");
+        System.out.println("  -i initialize Security plugin with default configuration (default is to ask if -y is not given)");
+        System.out.println("  -c enable cluster mode by binding to all network interfaces (default is to ask if -y is not given)");
+        System.out.println("  -s skip updates if config is already applied to opensearch.yml");
+        System.out.println(
+            "  -t set the execution environment to `test` to skip password validation. Should be used only for testing. (default is set to `demo`)"
```
RyanL1997, line 134: nice

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -0,0 +1,316 @@
...
+                    .build()
+            );
+
+            // Read custom password
+            if (initialAdminPassword != null && !initialAdminPassword.isEmpty()) {
+                ADMIN_PASSWORD = initialAdminPassword;
+            } else {
+                File adminPasswordFile = new File(ADMIN_PASSWORD_FILE_PATH);
+                if (adminPasswordFile.exists() && adminPasswordFile.length() > 0) {
+                    try (BufferedReader br = new BufferedReader(new FileReader(ADMIN_PASSWORD_FILE_PATH, StandardCharsets.UTF_8))) {
```
RyanL1997, line 74: Nit: Do we need to add a catch block for this and log the error that there is something wrong with the file reading?

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -0,0 +1,316 @@
...
+                        ADMIN_PASSWORD = br.readLine();
+                    }
+                }
+            }
+
+            // If script execution environment is set to demo, validate custom password, else if set to test, skip validation
+            if (shouldValidatePassword
+                && !ADMIN_PASSWORD.isEmpty()
+                && passwordValidator.validate("admin", ADMIN_PASSWORD) != RequestContentValidator.ValidationError.NONE) {
+                System.out.println("Password " + ADMIN_PASSWORD + " is weak. Please re-try with a stronger password.");
```
RyanL1997, line 84: Nit: we should also list out the rules for a stronger password (e.g. at least 1 symbol...etc.).

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -0,0 +1,316 @@
...
+        Path internalUsersPath = Paths.get(internalUsersFile);
+
+        try (
+            BufferedReader reader = new BufferedReader(new FileReader(internalUsersFile, StandardCharsets.UTF_8));
+            BufferedWriter writer = new BufferedWriter(new FileWriter(tempFilePath.toFile(), StandardCharsets.UTF_8))
+        ) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                if (line.matches(" *hash: *\"\\$2a\\$12\\$VcCDgh2NDk07JGN0rjGbM.Ad41qVR/YFJcgHp0UGns5JDymv..TOG\"")) {
+                    line = line.replace(
```
RyanL1997, line 135: I assume that this is the hash original admin password, which is "admin" at the moment. I know that we discussed this before, and I still think that if we are deciding to implement this entire new user flow for admin password setup, we should remove the current default admin credential from the `internal_users.yml`. Wdyt? (Btw even if we decide to remove it, I think this can be resolved as a followup).

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    /**
+     * Helper method to check if network.host config is present
+     * @param filePath path to opensearch.yml
+     * @return true is present, false otherwise
+     */
+    static boolean isNetworkHostAlreadyPresent(String filePath) {
+        try {
+            String searchString = "^network.host";
```
RyanL1997, line 240: Just for my knowledge, is the `^` sign escaping of the comment? I know that we had this line as a default comment in the config file.

RyanL1997 (commented):
Hi @DarshitChanpura , thanks for putting this together. In generally, this is java tool is well structured and clear, and I just left some comments.

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
```diff
@@ -0,0 +1,377 @@
...
+    static String OPENSEARCH_VERSION;
+    static String SECURITY_VERSION;
+
+    static ExecutionEnvironment environment = ExecutionEnvironment.demo;
+
+    static final String OS = System.getProperty("os.name") + " " + System.getProperty("os.version") + " " + System.getProperty("os.arch");
+
+    static final String FILE_EXTENSION = OS.toLowerCase().contains("win") ? ".bat" : ".sh";
+
+    static final String SYSTEM_INDICES = ".plugins-ml-config, .plugins-ml-connector, .plugins-ml-model-group, .plugins-ml-model, "
```
scrawfor99, line 52: What do you think about finding a way to standardize this to a single list somewhere and accessing it that way? 

If I remember, every time another plugin wants to add a system index they have to register it with us and we have to have them add it in multiple places. Since this is in Java now, it should be pretty straightforward to have a single global list which all operations relating to system indices can reference.

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
```diff
@@ -0,0 +1,377 @@
...
+                    System.out.println("Invalid option: " + options[i]);
+            }
+        }
+    }
+
+    /**
+     * Prints the help menu when -h option is passed
+     */
+    static void showHelp() {
+        System.out.println("install_demo_configuration.sh [-y] [-i] [-c]");
```
cwperks, line 127: As soon as the min version of Java supported is >= 15 we will start being able to use [text blocks](https://docs.oracle.com/en/java/javase/21/text-blocks/index.html).

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
```diff
@@ -0,0 +1,377 @@
...
+        securitySettingsConfigurer.configureSecuritySettings();
+        certificateGenerator.createDemoCertificates();
+        finishScriptExecution();
+    }
+
+    /**
+     * Prints deprecation warning and other headers for the script
+     */
+    static void printScriptHeaders() {
+        System.out.println("**************************************************************************");
```
scrawfor99, line 79: Is this true anymore?

cwperks (commented):
Took a first pass. This change looks good to me. 

What tests should be added for this PR? I'd love to have a bwc test where the old script is run in 2.11 w/ the bash script and another with the Java version of the script and show that the output is identical.

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
```diff
@@ -0,0 +1,377 @@
...
+        System.out.println("### ** Warning: Do not use on production or public reachable systems **");
+    }
+
+    /**
+     * Reads the options passed to the script
+     * @param options an array of strings containing options passed to the script
+     */
+    static void readOptions(String[] options) {
+        // set script execution dir
+        SCRIPT_DIR = options[0];
```
scrawfor99, line 94: Should we specify that we need to pass the script directory now? I don't see it mentioned in the help, but I don't recall having to pass the path before. This seems like something we should make very explicit unless you have a way of autopopulating it (I have not read all the files yet).

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
DarshitChanpura: this can be done in future iterations. The purpose of this iteration was to translate it to a java tool. I can definitely add it in a follow-up PR.

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
```diff
@@ -0,0 +1,377 @@
...
+        }
+
+        if (!(new File(OPENSEARCH_PLUGINS_DIR).exists())) {
+            System.out.println("Unable to determine OpenSearch plugins directory. Quit.");
+            System.exit(-1);
+        }
+
+        if (!(new File(OPENSEARCH_MODULES_DIR).exists())) {
+            System.out.println("Unable to determine OpenSearch modules directory. Quit.");
+            // System.exit(-1);
```
scrawfor99, line 231: This looks like it should not be commented out.

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
scrawfor99: I think we have unit tests which require the old default password be valid. If there is someway to replace this or this is not the case, I agree with Ryan.

src/main/java/org/opensearch/security/tools/democonfig/ExecutionEnvironment.java
```diff
@@ -0,0 +1,9 @@
+package org.opensearch.security.tools.democonfig;
+
+/**
+ * The environment in which the demo config installation script is being executed
+ */
+public enum ExecutionEnvironment {
+    demo, // default value
```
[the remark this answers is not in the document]
DarshitChanpura: I'll change them to all-caps to comply with syntax.

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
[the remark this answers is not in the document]
DarshitChanpura: I wasn't sure if I should or shouldn't so I kept it. I will remove it now.

scrawfor99 (commented):
Looks very good to me, just  a few comments but overall great job with this. 


Can we also unit tests each of these parts now?

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
DarshitChanpura: BASE_DIR always used script path. Take a look here: https://github.com/opensearch-project/security/blob/b7f47b1f516bee03a7dfaeec3a9133e4eafa3a73/tools/install_demo_configuration.sh#L101

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
DarshitChanpura: I will remove deprecation notice

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
DarshitChanpura: I don't think we use script to initiate configuration in unit tests.  This will only be used when spinning up a distribution and installing demo configuration.

Also, for tests I have added -t option which skips password validation.

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
DarshitChanpura: good catch. I originally added it as try-with-resources. but I'll add a catch block.

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
DarshitChanpura: Thoughts on addressing this in a follow-up PR? The purpose of this list currently is to write the same to opensearch.yml. If someone were to update the list they would have to modify opensearch.yml, in which case, having a static constant wouldn't matter.

But, we can figure out a way to add it as dynamic setting in `config.yml` however that is out of scope of this PR.

scrawfor99: Sounds good :)

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
DarshitChanpura: I translated the script as is. But upon digging further I see that this constant is never used after assignment. I'll remove it completely

DarshitChanpura:
> Can we also unit tests each of these parts now?

I have a follow-up PR that I'm working on to specifically add tests for this.

DarshitChanpura:
@RyanL1997 @cwperks @scrawfor99 Thank you for the detailed feedback!!
I've addressed all comments and pushed the changes. Feel free to re-review/resolve the conversations.

scrawfor99 (approved):
Looks good to me ! Thanks DC :)

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
RyanL1997: Ah, thats good to know.

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
RyanL1997: Yes I agree, and lets do this in the future iterations.

RyanL1997 (approved):
LGTM! Nice work @DarshitChanpura !

RyanL1997:
>I think we have unit tests which require the old default password be valid. If there is someway to replace this or this is not the case, I agree with Ryan.

Hi @scrawfor99 and @DarshitChanpura. Since I previously resolved the comment, I will reply it here for future reference. I think the unit tests has its own test configuration and that will be fine, but it does impact the integration test. But we can have a workaround in CI. You can reference to the experimental work I did here:
https://github.com/opensearch-project/security/pull/3449/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR65

And the impact tests is in this comment: https://github.com/opensearch-project/security/pull/3449/files#r1346368658

cwperks (approved):

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
camerondurham: Was there a reason we couldn't include the validation failure enum message here directly?

From (https://github.com/opensearch-project/security/pull/3057/files#diff-4b39b8c38e1e191dbba2db25c2f52a7588c6b6087a5d0ff2795d97bd0b2bc776), it looks like the validation rules from `RequestContentValidator.ValidationError` can provide some details on why the password was "bad" (e.g. too short, too similar to username, too weak per zxcvbn checks) and that info could be helpful to the users. 

Will make PR to add some more context to address #4048 anyways.

PR #3674, opened 2023-11-09
[Maintenance-2930] Review use of JSON Flattener

### LINKED ISSUE #2930: [Maintenance] Review use of JSON Flattener
During his review of #2926, @cwperks brought up a good point about potentially removing the dependency on the JSON-flattener.  

Craig mentioned: "If that is all we use the json-flattener for than I would be in favor of less dependencies. If I am understanding it correctly, it is recursively going through a JSON structure and flattening it so all.entries.are.dot.separated and there is no object nesting?" 

I am making this issue to unblock the original JSON bump PR but make sure that Craig's point is not forgotten. 

To resolve this issue, we will want to review the use of the library and see whether it can be removed.

prabhask5:
### Description

Implement JsonFlattener helper class as written in #2926 to deprecate the use of the unnecessary JsonFlattener third party module.

Maintenance

The JsonFlattener module was being utilized in only one place for one specific purpose, so these functions can be implemented as part of the OpenSearch codebase instead of importing an unnecessary third party module.

Hopefully nothing.

### Issues Resolved
#2930 

No

### Testing
Tests checked to make sure functions are not broken.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/support/JsonFlattener.java
+++ b/src/main/java/org/opensearch/security/support/JsonFlattener.java
@@ -1,0 +1,55 @@
+package org.opensearch.security.support;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import org.opensearch.core.common.Strings;
+import org.opensearch.security.DefaultObjectMapper;
+
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class JsonFlattener {
+
+    public static Map<String, Object> flattenAsMap(String jsonString) {
+        try {
+            final byte[] bytes = jsonString.getBytes("utf-8");
+            final TypeReference<Map<String, Object>> typeReference = new TypeReference<>() {
+            };
+            final Map<String, Object> jsonMap = DefaultObjectMapper.objectMapper.readValue(bytes, typeReference);
+            final Map<String, Object> flattenMap = new LinkedHashMap<>();
+            flattenEntries("", jsonMap.entrySet(), flattenMap);
+            return flattenMap;
+        } catch (final IOException ioe) {
+            throw new IllegalArgumentException("Unparseable json", ioe);
+        }
+    }
+
+    private static void flattenEntries(String prefix, final Iterable<Map.Entry<String, Object>> entries, final Map<String, Object> result) {
+        if (!Strings.isNullOrEmpty(prefix)) {
+            prefix += ".";
+        }
+
+        for (final Map.Entry<String, Object> e : entries) {
+            flattenElement(prefix.concat(e.getKey()), e.getValue(), result);
+        }
+    }
+
+    private static void flattenElement(String prefix, final Object source, final Map<String, Object> result) {
+        if (source instanceof Iterable) {
+            flattenCollection(prefix, (Iterable<Object>) source, result);
+        }
+        if (source instanceof Map) {
+            flattenEntries(prefix, ((Map<String, Object>) source).entrySet(), result);
+        }
+        result.put(prefix, source);
+    }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/security/support/JsonFlattener.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class JsonFlattener {
+
+    public static Map<String, Object> flattenAsMap(String jsonString) {
+        try {
+            final byte[] bytes = jsonString.getBytes("utf-8");
+            final TypeReference<Map<String, Object>> typeReference = new TypeReference<>() {
+            };
+            final Map<String, Object> jsonMap = DefaultObjectMapper.objectMapper.readValue(bytes, typeReference);
+            final Map<String, Object> flattenMap = new LinkedHashMap<>();
```
willyborankin, line 19: `DefaultObjectMapper.objectMapper.readValue` supports strings no need to take bytes Jackson does the same.

willyborankin (changes requested):
Hi @prabhask5, thank you for the fix. 
Could you please:
 - add a Unit test for the class
 - the library and all it dependencies needs to be excluded as well form `build.gradle`
 -  fix all java compiler warnings, this is the reason why build did not pass

prabhask5:
@willyborankin I'm getting a type safety warning that's causing the java builds to fail. I could not find a way to fix it while still using a generic cast, so I'm manually suppressing it as a temp fix. Please let me know if you know any better fix. Thanks!

src/test/java/org/opensearch/security/support/JsonFlattenerTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.junit.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+
+public class JsonFlattenerTest {
+    @Test
+    public void testFlattenAsMapBasic() {
+        Map<String, Object> flattenedMap = JsonFlattener.flattenAsMap("{\"key\": {\"nested\": 1}, \"another.key\": [\"one\", \"two\"] }");
```
willyborankin, line 15: Ltes add more simple and complex JSON tests.

src/main/java/org/opensearch/security/support/JsonFlattener.java
```diff
@@ -0,0 +1,56 @@
+package org.opensearch.security.support;
```
DarshitChanpura, line 1: nit: license header & java-doc for the class

src/test/java/org/opensearch/security/support/JsonFlattenerTest.java
```diff
@@ -0,0 +1,22 @@
+package org.opensearch.security.support;
```
DarshitChanpura, line 1: nit: license header

src/test/java/org/opensearch/security/support/JsonFlattenerTest.java
DarshitChanpura: +1. Can we add a comprehensive test suite verifying all methods of the new JsonFlattener?

prabhask5: @DarshitChanpura I can add more complex tests for the overall flattenAsMap function but I don't think making tests for the other functions is needed/would work since there is no return value and every other function is used within the workflow of flattenAsMap. Is there a particular benefit to writing smaller tests for these helper functions?

prabhask5:
@willyborankin @DarshitChanpura I added more complex tests- lmk if I should write some for the helper functions. Ty!

willyborankin (approved):

cwperks (approved):

Issue #3466, opened 2023-10-05
[FEATURE] Certificate hot-reload support for SSL only mode clusters.


shubhmkr-amazon:
**Is your feature request related to a problem?**
Yes, Currently for SSL only mode `reloadcerts` and `certs info` apis are not supported. 

**What solution would you like?**
SSL certificate reload and certificate info api support also for SSL only mode clusters.

**What alternatives have you considered?**
Don't have any other alternative in existing code to support certificate hot-reload feature in ssl mode cluster.(Currently node reboot needed to rotate certificates.)

**Do you have any additional context?**
In existing code we are not registering certificate related handlers for SSL Only Mode.
Code ref:- https://github.com/opensearch-project/security/blob/2.10/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L496

scrawfor99:
[Triage] Hi @shubhmkr-amazon, thank you for filing this issue. The first item here should be making sure this is in fact not supported. If it is not, this seems like a good request for support.

PR #3693, opened 2023-11-14
Hot-reload support in SSL only mode

shubhmkr-amazon:
### Description
* Category (Enhancement)
Currently hot-reload feature was not supported with SSL only mode.
Old behaviour:- SSL certificate renewal was not supported in SSL only mode. To update certificate we need to bring cluster down to pick latest certificate. 

New behaviour:- Added certificate renewal for SSL only mode. Now certificate will be reloaded by using `reloadcerts` api.

### Issues Resolved
https://github.com/opensearch-project/security/issues/3466

### Testing
unit testing
##### Manual testing

```
opensearch curl https://localhost:9200/_plugins/_security/api/ssl/certs\?pretty -k
{
  "http_certificates_list" : [
    {
      "issuer_dn" : "CN=Example Com Inc. Root CA,OU=Example Com Inc. Root CA,O=Example Com Inc.,DC=example,DC=com",
      "subject_dn" : "CN=node-0.example.com,OU=node,O=node,L=test,C=de",
      "san" : "[[8, 1.2.3.4.5.5], [2, node-0.example.com], [2, localhost], [7, 0:0:0:0:0:0:0:1], [7, 127.0.0.1]]",
      "not_before" : "2023-08-29T04:23:12Z",
      "not_after" : "2033-08-26T04:23:12Z"
    }
  ],
  "transport_certificates_list" : [
    {
      "issuer_dn" : "CN=Example Com Inc. Root CA,OU=Example Com Inc. Root CA,O=Example Com Inc.,DC=example,DC=com",
      "subject_dn" : "CN=node-0.example.com,OU=node,O=node,L=test,C=de",
      "san" : "[[8, 1.2.3.4.5.5], [2, node-0.example.com], [2, localhost], [7, 0:0:0:0:0:0:0:1], [7, 127.0.0.1]]",
      "not_before" : "2023-08-29T04:23:12Z",
      "not_after" : "2033-08-26T04:23:12Z"
    }
  ]
}
```
```
opensearch curl -sX PUT https://localhost:9200/_plugins/_security/api/ssl/transport/reloadcerts -k
{"message":"updated transport certs"}%

```

### Check List
- [ X ] New functionality includes testing
- [ X ] Commits are signed per the DCO using --signoff

### DIFF
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -310,7 +310,6 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
         }
 
         if (SSLConfig.isSslOnlyMode()) {
-            this.sslCertReloadEnabled = false;
             log.warn("OpenSearch Security plugin run in ssl only mode. No authentication or authorization is performed");
             return;
         }
@@ -588,6 +587,25 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
                     )
                 );
                 log.debug("Added {} rest handler(s)", handlers.size());
+            } else {
+                handlers.addAll(
+                    SecurityRestApiActions.getHandler(
+                        settings,
+                        configPath,
+                        restController,
+                        localClient,
+                        adminDns,
+                        cr,
+                        cs,
+                        principalExtractor,
+                        evaluator,
+                        threadPool,
+                        auditLog,
+                        userService,
+                        sks,
+                        sslCertReloadEnabled
+                    )
+                );
             }
         }
 
@@ -962,6 +980,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
     ) {
 
         SSLConfig.registerClusterSettingsChangeListener(clusterService.getClusterSettings());
+        this.threadPool = threadPool;
         if (SSLConfig.isSslOnlyMode()) {
             return super.createComponents(
                 localClient,
@@ -1160,6 +1179,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
         settings.addAll(super.getSettings());
 
         settings.add(Setting.boolSetting(ConfigConstants.SECURITY_SSL_ONLY, false, Property.NodeScope, Property.Filtered));
+        settings.add(Setting.boolSetting(ConfigConstants.SECURITY_SSL_CERT_RELOAD_ENABLED, false, Property.NodeScope, Property.Filtered));
 
         // currently dual mode is supported only when ssl_only is enabled, but this stance would change in future
         settings.add(SecuritySettings.SSL_DUAL_MODE_SETTING);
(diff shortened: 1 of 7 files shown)


willyborankin:
AFAIU this PR just opens access to reloadcerts and cert info endpoints without security check. I'm afraid that this is not a right solution. Since in this case any user without proper access can reload SSL certificates on HTTP and Transport layers and cluster as result will be compromised. I think the right solution would to use `WatchService` and if cert files for one of the cluster node have been changed reload them for the particular node, in other words each node in the cluster should track cert files names and reload them automatically.

shubhmkr-amazon:
@cwperks @willyborankin Thanks for looking into this PR. 
> How is the reloadcerts endpoint protected in SSL only mode?

It's not protected, as we don't have the concept of an admin (privileged) user for API authorization in SSL only mode.

> Can a requestor provide a secret to ensure that only requestors that know the secret are able to use the endpoint?

For giving a secret key as an option, we will have to modify the API contract, which does not seems like a good option. Also, storing and validating secret keys will be more challenging and have some other limitations too.

> Since in this case any user without proper access can reload SSL certificates on HTTP and Transport layers and cluster as result will be compromised.

With this logic all user will only able to see the certificate details, To compromise the cluster they will need to get access of node to modify certs path in opensearch.yaml file and bounce the node.

 > I think the right solution would to use WatchService
 
Yes, it seems to be one of the options to reload the SSL certificate, but the only small problem I see is constant polling for the cert file change, which doesn't fit for the one-time process of certificate renewal.

cwperks:
@shubhmkr-amazon Do you think its possible to think about SSL Only mode as the same as FGAC enabled, but 0 authenticators configured? I'm wondering if we can extend the notion of connecting with admin certificate to SSL Only mode and have the authorization of this endpoint the same for both scenarios. 

Admin cert authentication is done in the [BackendRegistry](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L213-L218) and this component is not created for SSL only mode so there would need to be some re-work to create that component and enabled admin cert for SSL only mode. 

Another issue I can see in SSL Only mode is that system indices are not given special protections. A benefit of differentiating between calls with the admin cert vs other calls in SSL only mode is that the security plugin could also start protecting system indices in SSL only mode.

DarshitChanpura (commented):
@shubhmkr-amazon What is the use-case for enabling it in SSL only mode? Will certificates ever be updated to a malicious cert? What is the security risk of enabling this vs the blocker for not enabling this?

peternied:
@shubhmkr-amazon I'm rejecting this pull request as @willyborankin pointed out this change is creating a serious security vulnerability in the posture of the cluster.

Before we can consider a change of this approach we will need a detailed design and thread model a security review.

shubhmkr-amazon:
> @shubhmkr-amazon What is the use-case for enabling it in SSL only mode? 

-> Hi @DarshitChanpura , I went for this pull request because hot-renewal in SSL only mode is not supported, and if someone creates a cluster in SSL only mode, then he should be given the option of certificate hot-reload support for certificate renewal similar to that of advance security option.

-> Will certificates ever be updated to a malicious cert?
It will update only if the actor who is using the `reloadcerts` API has cert file path access; without path access, it's impossible to change a certificate to a malicious certificate by just using the API.

 -> What is the security risk of enabling this vs the blocker for not enabling this?
I don't think we have any security risks from enabling these two APIs for SSL-only mode clusters, as they don't have any concept of admin users.



@peternied, I am open for discussion on security risks. Here in the `reloadcerts` API, I am not seeing any issues, whereas for the `certs info` API, I will check whether we will need any modifications for ssl-only mode or not.

peternied:
@shubhmkr-amazon Good to hear - thanks.  Please update us when you have the design proposal for this feature so we can reviewed and the risks discussed and weighted.

PR #3732, opened 2023-11-16
Removing flaky test case testAutoInit

### LINKED ISSUE #3730: [Flaky Test] org.opensearch.security.auditlog.compliance.RestApiComplianceAuditlogTest.testAutoInit is flaky
Seen in https://github.com/opensearch-project/security/actions/runs/6896058512/job/18761442974?pr=3727

peternied:
### Description
Considering what is tested - the auto log events that are creating when a test cluster starts up.  This isn't a useful test case, removing.

### Issues Resolved
- Resolves #3730

### Check List

### DIFF
--- a/src/test/java/org/opensearch/security/auditlog/compliance/RestApiComplianceAuditlogTest.java
+++ b/src/test/java/org/opensearch/security/auditlog/compliance/RestApiComplianceAuditlogTest.java
@@ -126,31 +126,6 @@ public class RestApiComplianceAuditlogTest extends AbstractAuditlogiUnitTest {
     }
 
     @Test
-    public void testAutoInit() throws Exception {
-
-        Settings additionalSettings = Settings.builder()
-            .put("plugins.security.audit.type", TestAuditlogImpl.class.getName())
-            .put(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_TRANSPORT, true)
-            .put(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_ENABLE_REST, true)
-            .put(ConfigConstants.OPENDISTRO_SECURITY_AUDIT_RESOLVE_BULK_REQUESTS, false)
-            .put(ConfigConstants.OPENDISTRO_SECURITY_COMPLIANCE_HISTORY_EXTERNAL_CONFIG_ENABLED, true)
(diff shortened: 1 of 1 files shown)


willyborankin (approved):

cwperks (changes requested):
I'm not convinced yet that this is a test that should be removed.

I took a glance at this test suite and it looks like it needs to be refactored to use [TestAuditLogImpl.doThenWaitForMessages](https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/auditlog/integration/TestAuditlogImpl.java#L60)

As it is, this test is useless because its not idempotent. The audit log is not cleared before this test runs so it relies on being at a certain location in the test suite to run properly.

peternied:
@cwperks if we had an issue to create a test that verifies the audit actions during the test cluster setup process - I think it would need to be authored differently to properly measure what happens.  Behavior such as calling the cluster health API once instead of twice, the test would need to be written to interpret those changes.  I think we'd need to rebuild the test to accommodate - more than just use the better concurrency safe capture method.

In terms of risk of not having _this test_ is pretty low on my list.  I'd rather spend time building better support for the new credentials generator or tackling some incoming bugs we have then try to rehab this test case.

peternied:
> There will be changes that destabilize or block contributions. The impact of these changes will be localized on the repository or even the entire OpenSearch project. We should bias towards keeping contributions unblocked by immediately reverting impacting changes, these reverts will be done by a maintainer. After the change has been reverted, an issue will be openned to re-merge the change and callout the elements of the contribution that need extra examination such as additional tests or even pull request workflows.
>
> Exceptional, instead of immediately reverting, if a contributor knows how and will resolve the issue in an hour or less we should fix-forward to reduce overhead.

_From the maintainers [guidance](https://github.com/opensearch-project/security/blob/main/MAINTAINERS.md#reverting-commits)_

@cwperks Our builds are constantly failing due to this test.  And I think it will take more than an hour to fix - with that in mind do you think we could merge as is? When the rewritten version that is great we add it back in, but until that happens we will have builds that are more stable.

DarshitChanpura (approved):

cwperks:
@cwperks I've lifted the Requested Changes, but my preference would be to merge the [test fix](https://github.com/opensearch-project/security/pull/3733) for this instead of removal. This whole suite of tests needs to be revisited. Our audit log tests could be a lot stronger than what they are today.

cwperks:
@peternied Are you planning to open a backport for this PR?

PR #3746, opened 2023-11-20
[BUG-2556] Add new DLS filtering test

### LINKED ISSUE #2556: [BUG] Can't use multiple roles with DLS permissions on the same Index
**What is the bug?**
When a user is bound to two roles using DLS permissions on the same index, only one DLS policy is taken into account.
Each role with DLS permissions works _independently_ as expected, but they won't combine in the typical additive fashion.

1: create a "global" role with READ on index pattern * and the following DLS:
  ```
  {
    "bool": {
      "must_not": {
        "match": {
          "sensitive": true
        }
      }
    }
  }
  ```
2: create a higher privileged role on any indices (example-*) with some catch-all DLS policy:
  ```
  {
    "bool": {
      "must": {
        "match_all": {}
      }
    }
  }
  ```
3: Query that example-* index pattern using the a User bound to both roles. Observe that you will not have any logs that contains `"sensitive": true`

The policies to be additive (combine each DLS statement in an OR query)

 - Provider: AWS Opensearch
 - Version: OpenSearch 2.3
 - Plugins: unsure

The codebase seems to indicate selection of a single arbitrary DLS policy for each index:
https://github.com/opensearch-project/security/blob/ca4d7528851ea94d936181f86428a25f0c59c8c1/src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java#L405

https://github.com/opensearch-project/security/blob/ca4d7528851ea94d936181f86428a25f0c59c8c1/src/main/java/org/opensearch/security/support/SecurityUtils.java#L98

Same issue [described here](https://forum.opensearch.org/t/how-is-dls-applied-when-user-has-multiple-roles/2946)

prabhask5:
### Description
This PR adds a new DLS filtering test that checks for the union of different dls settings related to different roles attached to the same user.

Enhancement

These tests were previously not included, which makes testing bugs including the one in #2556 difficult. Adding this test makes it easy to recreate these bugs if they appear in the future.

A new test should increase code coverage.

### Issues Resolved
#2556 

No

### Testing
Added new test

### Check List

### DIFF
--- a/src/test/java/org/opensearch/security/dlic/rest/api/DlsRoleSearchFilteringTest.java
+++ b/src/test/java/org/opensearch/security/dlic/rest/api/DlsRoleSearchFilteringTest.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ */
+
+package org.opensearch.security.dlic.rest.api;
+
+import java.util.stream.IntStream;
+
+import org.apache.hc.core5.http.HttpStatus;
+import org.junit.Test;
+
+import org.opensearch.security.test.AbstractSecurityUnitTest;
+import org.opensearch.security.test.helper.rest.RestHelper;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.opensearch.security.OpenSearchSecurityPlugin.PLUGINS_PREFIX;
+
+public class DlsRoleSearchFilteringTest extends AbstractRestApiUnitTest {
+    private final String SECURITY_ENDPOINT;
+
+    public DlsRoleSearchFilteringTest() {
+        SECURITY_ENDPOINT = PLUGINS_PREFIX + "/api";
+    }
+
+    @Test
+    public void testRoleUnionSearchFiltering() throws Exception {
+        setup();
+
+        rh.keystore = "restapi/kirk-keystore.jks";
+        rh.sendAdminCertificate = true;
+
+        // create roles
+        final String ROLE_ENDPOINT = SECURITY_ENDPOINT + "/roles";
+        RestHelper.HttpResponse testRoleOneResponse = rh.executePutRequest(
+            ROLE_ENDPOINT + "/test_role_1",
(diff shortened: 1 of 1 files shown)


prabhask5:
@scrawfor99 @sbeginCoveo @peternied I'm trying to write a new test to check for correct dls filtering functionality, including the union of different roles attached to the same user. I have most of this test written (including API calls to make a new user, new roles, new index, and populate that index with documents), but I'm having difficulty having the _search api call work properly. Even when I'm trying to get all the documents, only 6 of the 10 make show up, and this number is inconsistent, with the number of documents ranging from 0 to 6. Can I get some guidance on this, and why this problem is happening?

src/test/java/org/opensearch/security/dlic/rest/api/DlsRoleSearchFilteringTest.java
```diff
@@ -0,0 +1,93 @@
...
+         IntStream.range(1, documents.length + 1).forEach(i -> {
+            RestHelper.HttpResponse createDocumentResponse = rh.executePutRequest(
+                INDEX_ENDPOINT + "/_doc/" + i,
+                documents[i - 1],
+                AbstractSecurityUnitTest.encodeBasicHeader("admin", "admin")
+            );
+            assertThat(createDocumentResponse.getStatusCode(), equalTo(HttpStatus.SC_CREATED));
+        });
+
+        RestHelper.HttpResponse testResponse = rh.executeGetRequest(INDEX_ENDPOINT + "/_search", "{\"query\":{\"match_all\":{}}}");
```
peternied, line 90: I'd recommend doing this request as admin to confirm you've created what you expect, and then make another request as users with restricted roles so you can side-by-side comparison them.

peternied (commented):
This test is written with the `src/test` framework, whereas the `src/integrationTest` framework is the more mature starting point - give that a try you might find it much easier to deal with setup/configuration issues since everything is declared in once place.

Further note, its much easier to followed a comment `thread` if you've got questions about a specific bit of code (and they can be resolved/reopened) whereas comments are generally good for setting broad expectations around the pull request process.

src/test/java/org/opensearch/security/dlic/rest/api/DlsRoleSearchFilteringTest.java
```diff
@@ -0,0 +1,93 @@
...
+    public void testRoleUnionSearchFiltering() throws Exception {
+        setup();
+
+        rh.keystore = "restapi/kirk-keystore.jks";
+        rh.sendAdminCertificate = true;
+
+        // create roles
+        final String ROLE_ENDPOINT = SECURITY_ENDPOINT + "/roles";
+        RestHelper.HttpResponse testRoleOneResponse = rh.executePutRequest(
+            ROLE_ENDPOINT + "/test_role_1",
```
cwperks, line 42: Can these roles be shortened to only the permissions needed for the test. i.e. only index permissions on the indices needed. It looks like the `tenant_permissions` portions of these roles are not necessary for the test.

Its hard to read the permissions in the test since the JSON is escaped and on a single line.

cwperks:
@prabhask5 I'm not positive on why you get a different number of documents with the search query in this test, but according to the [index document documentation](https://opensearch.org/docs/latest/api-reference/document-apis/index-document/) the index request may need `?refresh=true` to ensure that the newly indexed documents are available for searching.

peternied:
Doesn't look like this is being worked on anymore, going to close this out. Please feel free to reopen if you re-engage on this @prabhask5

PR #3754, opened 2023-11-20
[Enhancement-3191] `transport_enabled` setting on an auth domain and authorizer may be unnecessary after transport client removal 

### LINKED ISSUE #3191: `transport_enabled` setting on an auth domain and authorizer may be unnecessary after transport client removal
Since OpenSearch 2.0, support for the transport client has been removed. The transport client was removed from the security plugin in this PR: https://github.com/opensearch-project/security/pull/1701

My understanding of `transport_enabled` is that it is used to enable an auth domain for authentication on the transport layer. There is a separate setting called `http_enabled` to enable the auth domain on the REST layer. This setting also appears to be applicable to authorizers in the `authz` section.

Below is an example of a basic auth entry in the `authc:` section of the security plugin's `config.yml` file from the demo configuration:

```
basic_internal_auth_domain:
    description: "Authenticate via HTTP Basic against internal users database"
    http_enabled: true
    transport_enabled: true
    order: 4
    http_authenticator:
      type: basic
      challenge: true
    authentication_backend:
      type: intern
```

The `transport_enabled` setting may not be needed here after the TransportClient's removal.

First: Determine if it is safe to remove values from settings. 
Second: If safe to remove, remove. If not, update documentation.

prabhask5:
### Description
Remove the transport_enabled setting in config.yml and its related code in DynamicConfigModel.

Enhancement

#3191 states that since the transport client has been removed, there is no need for the transport_enabled config and its related code to exist in the codebase, so this ticket removes that unnecessary code.

Nothing.

### Issues Resolved
#3191 

No

### Testing
No tests were made, all tests should run since the behavior should not change.

### Check List

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -14,9 +14,8 @@
 # After authentication authorization (authz) will be applied. There can be zero or more authorizers which collect
 # the roles from a given backend for the authenticated user.
 #
-# Both, authc and auth can be enabled/disabled separately for REST and TRANSPORT layer. Default is true for both.
+# Both, authc and auth can be enabled/disabled separately for REST layer. Default is true for both.
 #        http_enabled: true
-#        transport_enabled: true
 #
 # For HTTP it is possible to allow anonymous authentication. If that is the case then the HTTP authenticators try to
 # find user credentials in the HTTP request. If credentials are found then the user gets regularly authenticated.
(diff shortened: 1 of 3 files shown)


cwperks:
@prabhask5 This change looks good to me, but is it backwards compatible. If someone uses the same config that includes `transport_enabled` as a setting will it read the config and ignore the value or will it fail to startup since there is an unknown config value? Can we add any logging in on startup to log that a removed setting is still being referenced in a config file and that it can be safely removed?

scrawfor99:
Closing this for inaction. Please open a new PR if you choose to pursue this change further.

PR #3756, opened 2023-11-21
[BUG-2548] LDAP fails when the security.ssl.transport.truststore_filepath is missing

### LINKED ISSUE #2548: [BUG] LDAP fails when the `security.ssl.transport.truststore_filepath` is missing
**What is the bug?**
This is a continuation of the issue [1866](https://github.com/opensearch-project/security/issues/1866). The LDAP authentication process fails when the `security.ssl.transport.truststore_filepath` is missing from the `opensearch.yml` file, leading to an unresolved bug from a previous issue. The steps to reproduce the bug are documented in the Opensearch forum thread, along with the host/environment details.

All the steps and config files are published in https://forum.opensearch.org/t/empty-file-path-for-plugins-security-ssl-transport-truststore-filepath-ldap-authentication/12991 thread. 

The LDAP authentication process should function as expected when the `security.ssl.transport.truststore_filepath` is missing from the `opensearch.yml` file, and the `pemtrustedcas_filepath:` is included in the `config.yml` file

All the information is provided in the Opesearch forum thread 

It appears that the Java code responsible for the logic of how the CA certificates are read is located at [this location](https://github.com/opensearch-project/security/blob/main/src/main/java/com/amazon/dlic/util/SettingsBasedSSLConfigurator.java#L181). We are unsure if the documentation is incorrect, as the `security.ssl.transport.truststore_filepath` option is listed as optional, while it appears to be required for authorization. We are also considering the possibility of changing the logic for searching the IdP RootCA certificate

prabhask5:
### Description
Refactored SECURITY_SSL_TRANSPORT_TRUSTSTORE_FILEPATH to SECURITY_SSL_HTTP_TRUSTSTORE_FILEPATH in LDAP authentication and related functions. The reason for doing so is that according to the documentation: https://opensearch.org/docs/latest/security/authentication-backends/ldap/, TRANSPORT is not used for LDAP, HTTP is. So, whenever SECURITY_SSL_TRANSPORT_TRUSTSTORE_FILEPATH is removed from `opensearch.yml`, LDAP authentication fails.

Bug fix

SECURITY_SSL_TRANSPORT_TRUSTSTORE_FILEPATH not used in LDAP authentication according to docs, causing bug #2548 

LDAP uses HTTP instead of TRANSPORT, so now SECURITY_SSL_HTTP_TRUSTSTORE_FILEPATH and related HTTP settings are used in LDAP authentication.

### Issues Resolved
#2548 

No

### Testing
At end of process, all LDAP related tests will also use HTTP instead of TRANSPORT.

### Check List

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/ldap/backend/LDAPAuthorizationBackend.java
+++ b/src/main/java/com/amazon/dlic/auth/ldap/backend/LDAPAuthorizationBackend.java
@@ -82,8 +82,8 @@ import org.ldaptive.ssl.CredentialConfigFactory;
 import org.ldaptive.ssl.SslConfig;
 import org.ldaptive.ssl.ThreadLocalTLSSocketFactory;
 
-import static org.opensearch.security.ssl.SecureSSLSettings.SSLSetting.SECURITY_SSL_TRANSPORT_KEYSTORE_PASSWORD;
-import static org.opensearch.security.ssl.SecureSSLSettings.SSLSetting.SECURITY_SSL_TRANSPORT_TRUSTSTORE_PASSWORD;
+import static org.opensearch.security.ssl.SecureSSLSettings.SSLSetting.SECURITY_SSL_HTTP_KEYSTORE_PASSWORD;
+import static org.opensearch.security.ssl.SecureSSLSettings.SSLSetting.SECURITY_SSL_HTTP_TRUSTSTORE_PASSWORD;
 
 public class LDAPAuthorizationBackend implements AuthorizationBackend {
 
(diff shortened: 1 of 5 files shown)


prabhask5:
@peternied @scrawfor99 @cwperks To fix #2548 , I'm refactoring SECURITY_SSL_TRANSPORT_TRUSTSTORE_FILEPATH to SECURITY_SSL_HTTP_TRUSTSTORE_FILEPATH in the LDAP backend and related files. This is a major refactor, and I will have to refactor this for all the LDAP tests as well. Before I do that, I want to check whether this change is actually correct in this scenario because I'm just going off what the documentation suggests: https://opensearch.org/docs/latest/security/authentication-backends/ldap/, which states that SECURITY_SSL_HTTP_TRUSTSTORE_FILEPATH is used for LDAP, not SECURITY_SSL_TRANSPORT_TRUSTSTORE_FILEPATH.

scrawfor99:
Hey @prabhask5, LDAP can be enabled for both transport and HTTP. That setting is mentioned here: https://opensearch.org/docs/latest/security/authentication-backends/ldap/#certificate-validation.

There is not a mention of the transport truststore setting explicitly but it is the same as the HTTP version and required for use of SSL on the transport layer. 

The issue looks to be around checking for the truststore when a pemcert is provided. We should only enforce one or the other not both. So I think you will want to look in that direction.

scrawfor99:
Closing this for inaction. Please open a new PR if you choose to pursue this change further.

PR #3777, opened 2023-11-30
Updates demo script to execute with bundled jdk

DarshitChanpura:
### Description
This PR adds changes to the tool to allow execution with bundled jdk. It also refactors usage of static variables.
It also addresses a small bug in `isStringAlreadyPresentFile` to modify from regex match to "startsWith". 

Tests are added as fast-follow in #3807 

### Check List
~- [ ] New functionality includes testing~ <- fast-follow via #3807 


src/main/java/org/opensearch/security/tools/democonfig/Installer.java
```diff
@@ -205,23 +205,28 @@ static void setOpenSearchVariables() {
         OPENSEARCH_LIB_PATH = BASE_DIR + "lib" + File.separator;
         OPENSEARCH_INSTALL_TYPE = determineInstallType();
 
+        boolean shouldExit = false;
         if (!(new File(OPENSEARCH_CONF_FILE).exists())) {
-            System.out.println("Unable to determine OpenSearch config directory. Quit.");
-            System.exit(-1);
+            System.out.println("Unable to determine OpenSearch config file. Quit.");
+            shouldExit = true;
```
peternied, line 211: Nit: better control flow by using a collection to captures validation error, and then if you have any errors exiting

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
```diff
@@ -205,23 +205,28 @@ static void setOpenSearchVariables() {
         OPENSEARCH_LIB_PATH = BASE_DIR + "lib" + File.separator;
         OPENSEARCH_INSTALL_TYPE = determineInstallType();
 
+        boolean shouldExit = false;
         if (!(new File(OPENSEARCH_CONF_FILE).exists())) {
-            System.out.println("Unable to determine OpenSearch config directory. Quit.");
-            System.exit(-1);
+            System.out.println("Unable to determine OpenSearch config file. Quit.");
```
peternied, line 210: Are we should be running the same checkstyle + spotless rules on these files as the rest of the codebase?

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
```diff
@@ -356,11 +361,37 @@ static void finishScriptExecution() {
...
             System.out.println(e.getMessage());
         }
     }
+
+    /**
+     * FOR TESTS ONLY
+     * Resets the class-level variables to their original values
+     * Is needed for tests since the variables are declared static
+     */
+    static void resetState() {
```
peternied, line 381: If Installer was a class in the `static ... main()` method, you could just make an instance of the class, this would allow your tests to create fresh instances without any worry of static state management.

Maintaining this list as new stateful values are added and change behavior will be hard to check

src/test/java/org/opensearch/security/tools/democonfig/CertificateGeneratorTests.java
```diff
@@ -0,0 +1,86 @@
...
+            String fileContents = null;
+            try {
+                fileContents = new String(Files.readAllBytes(Path.of(certFilePath)));
+            } catch (Exception e) {
+                fail("Expected the test to pass.");
+            }
+
+            assertThat(fileContents.isEmpty(), not(true));
+            assertThat(fileContents, containsString("---BEGIN"));
+            assertThat(fileContents, containsString("---END"));
```
peternied, line 69: Feels like we should load and verify items like; expiry time (10 years?), cert chain is consistent, etc ?

src/test/java/org/opensearch/security/tools/democonfig/InstallerTests.java
```diff
@@ -0,0 +1,512 @@
...
+        assertEquals("/scriptDir", SCRIPT_DIR);
+        assertThat(assumeyes, is(true));
+        assertThat(initsecurity, is(true));
+        assertThat(cluster_mode, is(true));
+        assertEquals(0, skip_updates);
+        assertEquals(ExecutionEnvironment.TEST, environment);
+    }
+
+    @Test
+    public void testReadOptions_help() {
```
peternied, line 115: every time we add/change help text a unit test should break?  Can we verify this checking structured output instead of the process output?

peternied (commented):
These tests look pretty brittle since they are handling output written to the console and files, I'd recommend breaking up how the parts of the tool work so you can have easier to maintain test cases.

Something to consider if the tool gets a new option, say <10 lines of code.  In the tests you would only really need two new test cases (pass/fail) and no unrelated tests would break.

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
DarshitChanpura: I see that you disagreed. Other files in tools package use the same format of using sysout to print information. To address this, we added an except to checkstyle [here](https://github.com/opensearch-project/security/blob/main/checkstyle/checkstyle.xml#L41)

DarshitChanpura:
it is quite perplexing, that the test that are failing for windows only fail for 11 and 17 distributions of java. They work perfectly fine with 21.

I have reproduced these successfully on command prompt on windows machine. The interesting part is when I run the test on debugger in VS code on the same windows machine (using java 11), the test pass, but when I run the same on command prompt using java 11 the tests fail.

DarshitChanpura:
the issue was with comparing characters when generating random password.

cwperks (commented):
Took a first pass and left a few suggestions.

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
```diff
@@ -45,7 +45,7 @@ public class Installer {
 
     static ExecutionEnvironment environment = ExecutionEnvironment.DEMO;
 
-    static final String OS = System.getProperty("os.name") + " " + System.getProperty("os.version") + " " + System.getProperty("os.arch");
+    static String OS = System.getProperty("os.name") + " " + System.getProperty("os.version") + " " + System.getProperty("os.arch");
```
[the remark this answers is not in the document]
DarshitChanpura: yes, for testing. I modify these for unit tests (testing output when OS is Windows, something else, etc) and hence is required to be non-final

src/test/java/org/opensearch/security/sanity/tests/SecurityRestTestCase.java
```diff
@@ -45,6 +45,8 @@ public class SecurityRestTestCase extends OpenSearchRestTestCase {
 
     private static final String CERT_FILE_DIRECTORY = "sanity-tests/";
 
+    static boolean useAdminAsPassword = false;
```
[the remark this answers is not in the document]
DarshitChanpura: yes, It is required by buildclient to establish what password to be used for building an admin user client. Since build client is invoked before test-setup, the only way to access it is via static variable (modified via @BeforeClass)

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
DarshitChanpura: okie. I have rewrote the tool to follow singleton pattern.

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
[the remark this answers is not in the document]
DarshitChanpura: The tool now follows singleton pattern

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -193,46 +194,68 @@ static void writeSecurityConfigToOpenSearchYML() {
...
-        securityConfigLines.append("plugins.security.enable_snapshot_restore_privilege:  true\n");
-        securityConfigLines.append("plugins.security.check_snapshot_restore_write_privileges:  true\n");
-        securityConfigLines.append("plugins.security.restapi.roles_enabled:  [\"all_access\", \"security_rest_api_access\"]\n");
-
-        securityConfigLines.append("plugins.security.system_indices.enabled: true\n");
-        securityConfigLines.append("plugins.security.system_indices.indices: [").append(SYSTEM_INDICES).append("]\n");
+        securityConfigLines.append(
+            "plugins.security.authcz.admin_dn:"
+                + System.lineSeparator()
+                + "  - CN=kirk,OU=client,O=client,L=test, C=de"
```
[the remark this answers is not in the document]
DarshitChanpura: yep..updated

peternied:
> These tests look pretty brittle since they are handling output written to the console and files, I'd recommend breaking up how the parts of the tool work so you can have easier to maintain test cases.
> 
> Something to consider if the tool gets a new option, say <10 lines of code. In the tests you would only really need two new test cases (pass/fail) and no unrelated tests would break.

This comment isn't addressed.

src/test/java/org/opensearch/security/sanity/tests/SingleClusterSanityIT.java
```diff
@@ -34,6 +38,13 @@ public void testSecurityPluginInstallation() throws Exception {
         verifyPluginInstallationOnAllNodes();
     }
 
+    @Test
+    public void testAdminCredentials_customStrongPassword_shouldSucceed() throws Exception {
```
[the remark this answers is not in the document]
DarshitChanpura: performRequest() requires an exception to be handled or thrown. I chose to not handle it since the test should fail if performRequest didn't work as expected.

src/test/java/org/opensearch/security/sanity/tests/InvalidAdminPasswordIT.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @BeforeClass
+    public static void setUpInvalidAdminCredentials() {
+        useAdminAsPassword = true;
+    }
+
+    @AfterClass
+    public static void restoreValidAdminCredentials() {
+        useAdminAsPassword = false;
+    }
```
derek-ho, line 36: Why is this necessary? Does the useAdminAsPassword carry over to other files?

DarshitChanpura: this is to test that the request is not successful when the password is set to `admin`.

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
```diff
@@ -356,11 +383,19 @@ static void finishScriptExecution() {
...
+                    + "."
+            );
             System.out.println("### (Ignore the SSL certificate warning because we installed self-signed demo certificates)");
 
         } catch (Exception e) {
             System.out.println(e.getMessage());
         }
     }
+
+    public static void resetInstance() {
```
cwperks, line 398: If this isn't used anywhere can it be removed?

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -151,15 +187,15 @@ static void writePasswordToInternalUsersFile(String adminPassword, String intern
     /**
      * Checks if security plugin is already configured. If so, the script execution will not continue.
      */
-    static void checkIfSecurityPluginIsAlreadyConfigured() {
+    void checkIfSecurityPluginIsAlreadyConfigured() {
         // Check if the configuration file contains the 'plugins.security' string
-        if (OPENSEARCH_CONF_FILE != null && new File(OPENSEARCH_CONF_FILE).exists()) {
-            try (BufferedReader br = new BufferedReader(new FileReader(OPENSEARCH_CONF_FILE, StandardCharsets.UTF_8))) {
+        if (this.installer.OPENSEARCH_CONF_FILE != null && new File(this.installer.OPENSEARCH_CONF_FILE).exists()) {
```
cwperks, line 192: nit: is `this` required everywhere here?

Would it make sense to move this method to the Installer or consider making getters for the values being accessed here?

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -175,66 +211,75 @@ static void checkIfSecurityPluginIsAlreadyConfigured() {
...
+        String configFooter = "######## End OpenSearch Security Demo Configuration ########" + System.lineSeparator();
+
+        Map<String, Object> securityConfigAsMap = buildSecurityConfigMap();
+
+        try (FileWriter writer = new FileWriter(this.installer.OPENSEARCH_CONF_FILE, StandardCharsets.UTF_8, true)) {
+            writer.write(configHeader);
+            Yaml yaml = new Yaml();
+            DumperOptions options = new DumperOptions();
+            options.setDefaultFlowStyle(DumperOptions.FlowStyle.BLOCK);
+            String yamlString = yaml.dump(securityConfigAsMap);
```
cwperks, line 230: Nice!

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -151,15 +187,15 @@ static void writePasswordToInternalUsersFile(String adminPassword, String intern
...
-            try (BufferedReader br = new BufferedReader(new FileReader(OPENSEARCH_CONF_FILE, StandardCharsets.UTF_8))) {
+        if (this.installer.OPENSEARCH_CONF_FILE != null && new File(this.installer.OPENSEARCH_CONF_FILE).exists()) {
+            try (BufferedReader br = new BufferedReader(new FileReader(this.installer.OPENSEARCH_CONF_FILE, StandardCharsets.UTF_8))) {
                 String line;
                 while ((line = br.readLine()) != null) {
                     if (line.toLowerCase().contains("plugins.security")) {
-                        System.out.println(OPENSEARCH_CONF_FILE + " seems to be already configured for Security. Quit.");
-                        System.exit(skip_updates);
+                        System.out.println(this.installer.OPENSEARCH_CONF_FILE + " seems to be already configured for Security. Quit.");
+                        System.exit(this.installer.skip_updates);
```
cwperks, line 198: Is `skip_updated` ever changed from -1? Why have a variable?

cwperks (commented):
Took another pass. Nice work getting SnakeYaml to write to opensearch.yml!

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
DarshitChanpura: It is used in tests

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
DarshitChanpura: We do need to keep the method here since we split the installer class to be slightly modular. This class handles security related config settings. Given that we are using singleton instance, installer might not need this since all instances of SecuritySettingsConfigurer will still refer to same Installer instance. Now that I think out loud, I'll remove this.

src/test/java/org/opensearch/security/tools/democonfig/InstallerTests.java
DarshitChanpura: I'm moving this discussion to #3807 and resolving it here.

DarshitChanpura:
Code Coverage will be enhanced in #3807

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
```diff
@@ -82,60 +142,45 @@ static void printScriptHeaders() {
...
+            CommandLine line = parser.parse(options, args);
+
+            if (line.hasOption("h")) {
+                showHelp();
+                return;
             }
+            assumeyes = line.hasOption("y");
+            initsecurity = line.hasOption("i");
+            cluster_mode = line.hasOption("c");
+            skip_updates = line.hasOption("s") ? 0 : -1;
```
scrawfor99, line 162: Why is this the only boolean redefined into a number?

DarshitChanpura: this is to defined exit code used later on line 196 in SecuritySettingsConfigurer.java

src/main/java/org/opensearch/security/tools/democonfig/Installer.java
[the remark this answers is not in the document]
DarshitChanpura: see comment below.

src/main/java/org/opensearch/security/tools/democonfig/Certificates.java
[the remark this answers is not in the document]
DarshitChanpura: yes.

derek-ho (approved):

derek-ho:
Ship it 🚢

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -258,7 +300,7 @@ static boolean isNetworkHostAlreadyPresent(String filePath) {
      */
     static boolean isNodeMaxLocalStorageNodesAlreadyPresent(String filePath) {
         try {
-            String searchString = "^node.max_local_storage_nodes";
+            String searchString = "node.max_local_storage_nodes:";
             return isStringAlreadyPresentInFile(filePath, searchString);
```
cwperks, line 304: Is it possible to read the file in as YAML and check if the YAML object has the key you are searching for?

cwperks (approved):
This PR LGTM. 

One additional comment: Should the SecuritySettingsConfigurer get the values it needs in its constructor instead of the entire `installer` object?

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
DarshitChanpura: Thank you for catching this. I've updated it.

DarshitChanpura:
> One additional comment: Should the SecuritySettingsConfigurer get the values it needs in its constructor instead of the entire installer object?

I started out with that but thought it might be better to pass just one reference `installer` rather than multiple local copies of variables inside Installer class. IMO, I'd prefer to keep it as is.

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -258,30 +304,29 @@ static boolean isNetworkHostAlreadyPresent(String filePath) {
...
      */
-    static boolean isStringAlreadyPresentInFile(String filePath, String searchString) throws IOException {
-        try (BufferedReader reader = new BufferedReader(new FileReader(filePath, StandardCharsets.UTF_8))) {
-            String line;
-            while ((line = reader.readLine()) != null) {
-                if (line.matches(searchString)) {
-                    return true;
-                }
-            }
+    static boolean isStringAlreadyPresentInYMLFile(String filePath, String fieldName) throws IOException {
```
cwperks, line 321:
```suggestion
    static boolean isKeyPresentInYMLFile(String filePath, String fieldName) throws IOException {
```

cwperks (approved):

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
DarshitChanpura: done.

cwperks (approved):

scrawfor99 (approved):
Looks good to me! Thanks DC

Issue #3776, opened 2023-11-30
[BUG] DLS performance has regressed with new serialization format


peternied:
**What is the bug?**
Some users of OpenSearch have seen a performance decrease associated with DLS queries.  Larger DLS queries (more characters) 
have a larger impact.

**How can one reproduce the bug?**
1. Checkout my fork with the repro `git clone https://github.com/peternied/security.git`
2. Checkout the most recent security plugin version with the new tests `git checkout dls-perf`
3. Execute the new tests `./gradlew integrationTest --tests org.opensearch.security.DlsTests.testDlsLargerQueryScenarios -x jacocoTestReport`
4. Open the test report `./build/reports/tests/integrationTest/classes/org.opensearch.security.DlsTests.html`
5. View the standard output
```
Creating 5 indices with 1 document
User, Count, Avg, Max, Min, Std ms reader, 100, 11.07, 17, 8, 1.67
Attached READER with role DLS_ONLY_LONG_VALUE
User, Count, Avg, Max, Min, Std ms reader, 100, 8.45, 17, 6, 1.64
Finished checks in 5282ms
Creating 50 indices with 1 document
User, Count, Avg, Max, Min, Std ms reader, 100, 13.68, 21, 10, 2.27
Attached READER with role DLS_ONLY_LONG_VALUE
User, Count, Avg, Max, Min, Std ms reader, 100, 252.04, 538, 228, 31.20
Finished checks in 30905ms
Creating 100 indices with 1 document
User, Count, Avg, Max, Min, Std ms reader, 100, 12.56, 19, 10, 1.78
Attached READER with role DLS_ONLY_LONG_VALUE
User, Count, Avg, Max, Min, Std ms reader, 100, 984.57, 1027, 916, 21.51
Finished checks in 107909ms
```

**What is the expected behavior?**
After the `DLS_ONLY_LONG_VALUE` is added, the AVG should not jump up so much.

**Do you have any additional context?**
You can collect measures from the 2.3 build by running `git checkout dls-perf` from that branch to collect numbers from that version
- https://github.com/opensearch-project/security/compare/main...peternied:security:dls-perf
- https://github.com/opensearch-project/security/compare/2.3...peternied:security:2.3-dls-perf

This issue was introduced in the following PR which did improve serialization in many scenarios, but seems to be impacting DLS queries as stored in headers.
- https://github.com/opensearch-project/security/pull/2802

peternied:
[...]

PR #3789, opened 2023-12-04
Disable custom serialization

parasjain1:
### Description
Disables custom serialization for the `2.11.1` patch release.

Needs to be backported to `2.x`.

Below is the approach. 

* For **requests sent** from the `2.11.1` nodes to -
    * Case 1 : a node running OS version < `2.11`
        * always use JDK to serialize the headers [this check was already present]
    * Case 2: a node running *OS* `2.11.0`
        * serialize via custom serialization
    * Case 3:  a node running *OS >=* `2.11.1`  
        * serialize via JDK serialization
* For **requests received** on the `2.11.1` nodes from -
    * Case 1 : a node running OS version < `2.11`
        * always use JDK to deserialize the headers [this check was already present]
    * Case 2: a node running OS  *`2.11.0`*
        * deserialize via custom serialization
    * Case 3: a node running OS >=  `2.11.1` 
        *  deserialize via JDK serialization

### Related Issues
- Related #3776 

### Testing 
BWCs should cover most of the scenarios, while still carry out a manual cluster upgrade test.

### DIFF
--- a/src/main/java/org/opensearch/security/support/Base64Helper.java
+++ b/src/main/java/org/opensearch/security/support/Base64Helper.java
@@ -28,6 +28,8 @@ package org.opensearch.security.support;
 
 import java.io.Serializable;
 
+import org.opensearch.Version;
+
 public class Base64Helper {
 
     public static String serializeObject(final Serializable object, final boolean useJDKSerialization) {
@@ -35,11 +37,11 @@ public class Base64Helper {
     }
 
     public static String serializeObject(final Serializable object) {
-        return serializeObject(object, false);
+        return serializeObject(object, true);
     }
 
     public static Serializable deserializeObject(final String string) {
-        return deserializeObject(string, false);
+        return deserializeObject(string, true);
     }
 
     public static Serializable deserializeObject(final String string, final boolean useJDKDeserialization) {
@@ -68,5 +70,33 @@ public class Base64Helper {
         }
         // If we see an exception now, we want the caller to see it -
         return Base64Helper.serializeObject(serializable, true);
+    }
+
+    /**
+     * Ensures that the returned string is custom serialized.
+     *
+     * If the supplied string is a JDK serialized representation, will deserialize it and further serialize using
+     * custom, otherwise returns the string as is.
+     *
+     * @param string original string, can be JDK or custom serialized
+     * @return custom serialized string
+     */
+    public static String ensureCustomSerialized(final String string) {
+        Serializable serializable;
+        try {
+            serializable = Base64Helper.deserializeObject(string, true);
+        } catch (Exception e) {
+            // We received an exception when de-serializing the given string. It is probably custom serialized.
+            // Try to deserialize using custom
+            Base64Helper.deserializeObject(string, false);
+            // Since we could deserialize the object using custom, the string is already custom serialized, return as is
+            return string;
+        }
+        // If we see an exception now, we want the caller to see it -
+        return Base64Helper.serializeObject(serializable, false);
+    }
+
+    public static boolean shouldUseJDKSerialization(Version remoteVersion) {
+        return !remoteVersion.equals(ConfigConstants.FIRST_CUSTOM_SERIALIZATION_SUPPORTED_OS_VERSION);
     }
 }
--- a/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
+++ b/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
@@ -72,6 +72,7 @@ import org.opensearch.transport.TransportRequestOptions;
 import org.opensearch.transport.TransportResponseHandler;
 
 import static org.opensearch.security.OpenSearchSecurityPlugin.isActionTraceEnabled;
+import static org.opensearch.security.support.Base64Helper.shouldUseJDKSerialization;
 
 public class SecurityInterceptor {
 
@@ -148,7 +149,7 @@ public class SecurityInterceptor {
         final String origCCSTransientMf = getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_MASKED_FIELD_CCS);
 
         final boolean isDebugEnabled = log.isDebugEnabled();
-        final boolean useJDKSerialization = connection.getVersion().before(ConfigConstants.FIRST_CUSTOM_SERIALIZATION_SUPPORTED_OS_VERSION);
+        final boolean useJDKSerialization = shouldUseJDKSerialization(connection.getVersion());
         final boolean isSameNodeRequest = localNode != null && localNode.equals(connection.getNode());
 
         try (ThreadContext.StoredContext stashedContext = getThreadContext().stashContext()) {
@@ -226,13 +227,13 @@ public class SecurityInterceptor {
                 );
             }
 
-            if (useJDKSerialization) {
-                Map<String, String> jdkSerializedHeaders = new HashMap<>();
+            if (!useJDKSerialization) {
+                Map<String, String> customSerializedHeaders = new HashMap<>();
                 HeaderHelper.getAllSerializedHeaderNames()
                     .stream()
                     .filter(k -> headerMap.get(k) != null)
-                    .forEach(k -> jdkSerializedHeaders.put(k, Base64Helper.ensureJDKSerialized(headerMap.get(k))));
-                headerMap.putAll(jdkSerializedHeaders);
+                    .forEach(k -> customSerializedHeaders.put(k, Base64Helper.ensureCustomSerialized(headerMap.get(k))));
+                headerMap.putAll(customSerializedHeaders);
             }
 
             getThreadContext().putHeader(headerMap);
--- a/src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
+++ b/src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
@@ -21,6 +21,7 @@ import org.opensearch.cluster.ClusterName;
 import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.core.common.transport.TransportAddress;
 import org.opensearch.core.transport.TransportResponse;
 import org.opensearch.extensions.ExtensionsManager;
@@ -108,8 +109,7 @@ public class SecurityInterceptorTests {
         );
     }
 
-    private void testSendRequestDecorate(Version remoteNodeVersion) {
-        boolean useJDKSerialization = remoteNodeVersion.before(ConfigConstants.FIRST_CUSTOM_SERIALIZATION_SUPPORTED_OS_VERSION);
+    private void testSendRequestDecorate(DiscoveryNode localNode, DiscoveryNode otherNode, boolean shouldUseJDKSerialization) {
         ClusterName clusterName = ClusterName.DEFAULT;
         when(clusterService.getClusterName()).thenReturn(clusterName);
 
@@ -143,17 +143,7 @@ public class SecurityInterceptorTests {
         @SuppressWarnings("unchecked")
         TransportResponseHandler<TransportResponse> handler = mock(TransportResponseHandler.class);
 
-        InetAddress localAddress = null;
-        try {
-            localAddress = InetAddress.getByName("0.0.0.0");
-        } catch (final UnknownHostException uhe) {
-            throw new RuntimeException(uhe);
-        }
-
-        DiscoveryNode localNode = new DiscoveryNode("local-node", new TransportAddress(localAddress, 1234), Version.CURRENT);
         Connection connection1 = transportService.getConnection(localNode);
-
-        DiscoveryNode otherNode = new DiscoveryNode("remote-node", new TransportAddress(localAddress, 4321), remoteNodeVersion);
         Connection connection2 = transportService.getConnection(otherNode);
 
         // from thread context inside sendRequestDecorate
@@ -189,7 +179,7 @@ public class SecurityInterceptorTests {
                 TransportResponseHandler<T> handler
             ) {
                 String serializedUserHeader = threadPool.getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER);
-                assertEquals(serializedUserHeader, Base64Helper.serializeObject(user, useJDKSerialization));
+                assertEquals(serializedUserHeader, Base64Helper.serializeObject(user, shouldUseJDKSerialization));
             }
         };
         // isSameNodeRequest = false
@@ -201,17 +191,49 @@ public class SecurityInterceptorTests {
         assertEquals(threadPool.getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER), null);
     }
 
+    /**
+     * Tests the scenario when remote node is on same OS version
+     */
     @Test
     public void testSendRequestDecorate() {
-        testSendRequestDecorate(Version.CURRENT);
+        DiscoveryNode localNode = new DiscoveryNode("local-node", new TransportAddress(getLocalAddress(), 1234), Version.CURRENT);
+        DiscoveryNode otherNode = new DiscoveryNode("other-node", new TransportAddress(getLocalAddress(), 3456), Version.CURRENT);
+        testSendRequestDecorate(localNode, otherNode, true);
     }
 
     /**
-     * Tests the scenario when remote node does not implement custom serialization protocol and uses JDK serialization
+     * Tests the scenarios for mixed node versions
      */
     @Test
-    public void testSendRequestDecorateWhenRemoteNodeUsesJDKSerde() {
-        testSendRequestDecorate(Version.V_2_0_0);
+    public void testSendRequestDecorateWithMixedNodeVersions() {
+
+        // local on latest version, remote on 2.11.0 - should use custom
+
+        try (ThreadContext.StoredContext ignore = threadPool.getThreadContext().stashContext()) {
+            DiscoveryNode localNode = new DiscoveryNode("local-node", new TransportAddress(getLocalAddress(), 1234), Version.CURRENT);
+            DiscoveryNode otherNode = new DiscoveryNode(
+                "other-node",
+                new TransportAddress(getLocalAddress(), 3456),
+                ConfigConstants.FIRST_CUSTOM_SERIALIZATION_SUPPORTED_OS_VERSION
+            );
+            testSendRequestDecorate(localNode, otherNode, false);
+        }
+
+        // remote node is on a version > 2.11.1 while local node is on version 2.11.1 - should use JDK
+        try (ThreadContext.StoredContext ignore = threadPool.getThreadContext().stashContext()) {
+            DiscoveryNode localNode = new DiscoveryNode("local-node", new TransportAddress(getLocalAddress(), 1234), Version.CURRENT);
+            DiscoveryNode otherNode = new DiscoveryNode("other-node", new TransportAddress(getLocalAddress(), 3456), Version.V_2_11_1);
+            testSendRequestDecorate(localNode, otherNode, true);
+        }
+
+    }
+
+    private static InetAddress getLocalAddress() {
+        try {
+            return InetAddress.getByName("0.0.0.0");
+        } catch (final UnknownHostException uhe) {
+            throw new RuntimeException(uhe);
+        }
     }
 
 }
--- a/src/test/java/org/opensearch/security/transport/SecuritySSLRequestHandlerTests.java
+++ b/src/test/java/org/opensearch/security/transport/SecuritySSLRequestHandlerTests.java
@@ -89,14 +89,15 @@ public class SecuritySSLRequestHandlerTests {
         Assert.assertTrue(threadPool.getThreadContext().getTransient(ConfigConstants.USE_JDK_SERIALIZATION));
 
         threadPool.getThreadContext().stashContext();
-        when(transportChannel.getVersion()).thenReturn(Version.V_2_11_0);
+        when(transportChannel.getVersion()).thenReturn(ConfigConstants.FIRST_CUSTOM_SERIALIZATION_SUPPORTED_OS_VERSION);
         Assert.assertThrows(Exception.class, () -> securitySSLRequestHandler.messageReceived(transportRequest, transportChannel, task));
         Assert.assertFalse(threadPool.getThreadContext().getTransient(ConfigConstants.USE_JDK_SERIALIZATION));
 
         threadPool.getThreadContext().stashContext();
-        when(transportChannel.getVersion()).thenReturn(Version.V_3_0_0);
+        when(transportChannel.getVersion()).thenReturn(Version.CURRENT);
         Assert.assertThrows(Exception.class, () -> securitySSLRequestHandler.messageReceived(transportRequest, transportChannel, task));
-        Assert.assertFalse(threadPool.getThreadContext().getTransient(ConfigConstants.USE_JDK_SERIALIZATION));
+        Assert.assertTrue(threadPool.getThreadContext().getTransient(ConfigConstants.USE_JDK_SERIALIZATION));
+
     }
 
     @Test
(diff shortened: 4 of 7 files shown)


src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -201,17 +195,65 @@ public <T extends TransportResponse> void sendRequest(
         assertEquals(threadPool.getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER), null);
     }
 
+
+    /**
+     * Tests the scenario when remote node is on same OS version and has custom serialization enabled in node attrs
```
backslasht, line 200:
```suggestion
     * Tests the scenario when remote node is on same OS version and has custom serialization not present in node attrs
```

backslasht (changes requested):

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1151,9 +1152,19 @@ public Settings additionalSettings() {
...
         return builder.build();
     }
 
+    private Settings getCustomSerializationNodeAttribute(Settings settings) {
+        return Settings.builder().put(
+                Node.NODE_ATTRIBUTES.getKey() + ConfigConstants.CUSTOM_SERIALIZATION_ENABLED,
+                SecuritySettings.CUSTOM_SERIALIZATION_ENABLED.get(settings)
+        ).build();
+    }
+
```
Bukhtawar, line 1167: Should we do a patch release instead for this to avoid identification through node attributes?

src/main/java/org/opensearch/security/support/SecuritySettings.java
```diff
@@ -28,4 +28,11 @@ public class SecuritySettings {
         Setting.Property.Dynamic
     ); // Not filtered
 
+    public static final Setting<Boolean> CUSTOM_SERIALIZATION_ENABLED = Setting.boolSetting(
+            ConfigConstants.CUSTOM_SERIALIZATION_ENABLED,
+            false,
+            Setting.Property.NodeScope,
+            Setting.Property.Dynamic
+    ); // Not filtered
+
```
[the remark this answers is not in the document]
parasjain1: The dynamic cluster setting is consumed into a node attribute during OS startup. 

My understanding about node attributes was that these may not be altered (please let me know if this is incorrect). The idea of keeping the cluster setting as dynamic was to allow user to change the setting cluster setting followed by a rolling restart which would update the node attribute to reflect the latest cluster setting.

Further, the serde protocol handling has a check on the node attribute so during update of this setting despite nodes having different setting value, the cluster will continue to work seamlessly.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -408,4 +413,19 @@ public String executor() {
         }
     }
 
+    private boolean shouldUseJDKSerialization(Connection connection) {
+        if(connection.getVersion().before(ConfigConstants.FIRST_CUSTOM_SERIALIZATION_SUPPORTED_OS_VERSION)) {
+            return true;
+        } else {
+            String customSerializationEnabledNodeAttr = connection.getNode().getAttributes().get(
```
cwperks, line 420: Can `getAttributes()` ever return `null`?

src/main/java/org/opensearch/security/support/SecuritySettings.java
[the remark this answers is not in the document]
backslasht: I agree with @Bukhtawar, it makes it much easier to completely disable it for now and re-enable once the performance of custom serialization is better (in future versions).

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
parasjain1: Done

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
parasjain1: We've removed the usage of node property, so this is no longer applicable.

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -201,17 +195,65 @@ public <T extends TransportResponse> void sendRequest(
...
-    public void testSendRequestDecorateWhenRemoteNodeUsesJDKSerde() {
-        testSendRequestDecorate(Version.V_2_0_0);
+    public void testSendRequestDecorateWhenRemoteNodeOnOlderVersion() {
+        DiscoveryNode localNode = new DiscoveryNode("local-node", new TransportAddress(getLocalAddress(), 1234), Version.CURRENT);
+        DiscoveryNode otherNode = new DiscoveryNode("local-node", new TransportAddress(getLocalAddress(), 1234), Version.V_2_0_0);
+        testSendRequestDecorate(localNode, otherNode, true);
     }
 
+    /**
+     * Tests the scenario when remote node is on same OS version and has custom serialization enabled in node attrs
```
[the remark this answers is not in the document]
parasjain1: Updated the tests as per the revised logic that relies on the patch version check.

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
parasjain1: Tests updated as per the revised logic that relies upon the patch version check.

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -143,17 +144,10 @@ private void testSendRequestDecorate(Version remoteNodeVersion) {
...
-        } catch (final UnknownHostException uhe) {
-            throw new RuntimeException(uhe);
-        }
-
-        DiscoveryNode localNode = new DiscoveryNode("local-node", new TransportAddress(localAddress, 1234), Version.CURRENT);
+//        DiscoveryNode localNode = new DiscoveryNode("local-node", new TransportAddress(localAddress, 1234), Version.CURRENT);
         Connection connection1 = transportService.getConnection(localNode);
 
-        DiscoveryNode otherNode = new DiscoveryNode("remote-node", new TransportAddress(localAddress, 4321), remoteNodeVersion);
+//        DiscoveryNode otherNode = new DiscoveryNode("remote-node", new TransportAddress(localAddress, 4321), remoteNodeVersion);
```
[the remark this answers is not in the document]
parasjain1: Done

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -143,17 +144,10 @@ private void testSendRequestDecorate(Version remoteNodeVersion) {
...
 
-        InetAddress localAddress = null;
-        try {
-            localAddress = InetAddress.getByName("0.0.0.0");
-        } catch (final UnknownHostException uhe) {
-            throw new RuntimeException(uhe);
-        }
-
-        DiscoveryNode localNode = new DiscoveryNode("local-node", new TransportAddress(localAddress, 1234), Version.CURRENT);
+//        DiscoveryNode localNode = new DiscoveryNode("local-node", new TransportAddress(localAddress, 1234), Version.CURRENT);
```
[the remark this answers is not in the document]
parasjain1: Done

src/test/java/org/opensearch/security/transport/SecuritySSLRequestHandlerTests.java
```diff
@@ -97,6 +97,21 @@ public void testUseJDKSerializationHeaderIsSetOnMessageReceived() throws Excepti
         when(transportChannel.getVersion()).thenReturn(Version.V_3_0_0);
         Assert.assertThrows(Exception.class, () -> securitySSLRequestHandler.messageReceived(transportRequest, transportChannel, task));
         Assert.assertFalse(threadPool.getThreadContext().getTransient(ConfigConstants.USE_JDK_SERIALIZATION));
+
+        // test when remote node has sent custom serialization enabled header
+        threadPool.getThreadContext().stashContext();
+        threadPool.getThreadContext().putHeader(ConfigConstants.CUSTOM_SERIALIZATION_ENABLED, "true");
```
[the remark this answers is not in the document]
parasjain1: Not needed as per the revised logic that relies upon the patch version check.

parasjain1:
Updated the PR. We will be keeping custom serialization disabled for the `2.11.1` patch release and re-enable later after the regression is fixed.

src/main/java/org/opensearch/security/support/Base64Helper.java
```diff
@@ -69,4 +71,33 @@ public static String ensureJDKSerialized(final String string) {
...
+            // Since we could deserialize the object using custom, the string is already custom serialized, return as is
+            return string;
+        }
+        // If we see an exception now, we want the caller to see it -
+        return Base64Helper.serializeObject(serializable, false);
+    }
+
+    public static boolean shouldUseJDKSerialization(Version remoteVersion) {
+        return remoteVersion.before(ConfigConstants.FIRST_CUSTOM_SERIALIZATION_SUPPORTED_OS_VERSION) ||
+                remoteVersion.onOrAfter(ConfigConstants.FIRST_CUSTOM_SERIALIZATION_DISABLED_OS_VERSION);
```
backslasht, line 101: Given that it was supported in only one version 2.11.0, can this be an exact match of that version?

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -226,12 +227,12 @@ && getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROL
                 );
             }
 
-            if (useJDKSerialization) {
+            if (!useJDKSerialization) {
                 Map<String, String> jdkSerializedHeaders = new HashMap<>();
```
backslasht, line 231:
```suggestion
                Map<String, String> customSerializedHeaders = new HashMap<>();
```

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -226,12 +227,12 @@ && getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROL
...
                 HeaderHelper.getAllSerializedHeaderNames()
                     .stream()
                     .filter(k -> headerMap.get(k) != null)
-                    .forEach(k -> jdkSerializedHeaders.put(k, Base64Helper.ensureJDKSerialized(headerMap.get(k))));
+                    .forEach(k -> jdkSerializedHeaders.put(k, Base64Helper.ensureCustomSerialized(headerMap.get(k))));
```
backslasht, line 235: if the version is 2.11.0, why it would not be custom serialized?

backslasht (changes requested):

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
parasjain1: We by default will serialize all headers via JDK and if during send we evaluate that the remote node understands custom, we invoke `ensureCustomSerialized` on the intermediate header map.

2.11.0 will receive custom serialized headers only.

src/main/java/org/opensearch/security/support/Base64Helper.java
parasjain1: Sure!

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
parasjain1: Done

backslasht (approved):

src/main/java/org/opensearch/security/support/Base64Helper.java
```diff
@@ -69,4 +71,32 @@ public static String ensureJDKSerialized(final String string) {
...
+     *
+     * @param string original string, can be JDK or custom serialized
+     * @return custom serialized string
+     */
+    public static String ensureCustomSerialized(final String string) {
+        Serializable serializable;
+        try {
+            serializable = Base64Helper.deserializeObject(string, true);
+        } catch (Exception e) {
+            // We received an exception when de-serializing the given string. It is probably custom serialized.
```
DarshitChanpura, line 89: Can we have a check from the exception message to confirm that the cause of the exception was indeed deserialization failure using Custom serialization (this would allow to skip the check on line 91 and avoid unnecessary deserialization)

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -334,6 +334,7 @@ public enum RolesMappingResolution {
     public static final boolean EXTENSIONS_BWC_PLUGIN_MODE_DEFAULT = false;
     // CS-ENFORCE-SINGLE
 
+
```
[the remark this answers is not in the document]
DarshitChanpura: +1.

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -201,17 +191,48 @@ public <T extends TransportResponse> void sendRequest(
         assertEquals(threadPool.getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER), null);
     }
 
+
+    /**
+     * Tests the scenario when remote node is on same OS version
```
DarshitChanpura, line 196: If I'm interpreting this correctly, all versions will use JDK serialization (except 2.11.0)?
Can we add a test to verify what should happen if shouldUseJDKSerialization is set to false when version is current for both?

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -201,17 +191,48 @@ public <T extends TransportResponse> void sendRequest(
...
+        try (ThreadContext.StoredContext ignore = threadPool.getThreadContext().stashContext()) {
+            DiscoveryNode localNode = new DiscoveryNode("local-node", new TransportAddress(getLocalAddress(), 1234), Version.CURRENT);
+            DiscoveryNode otherNode = new DiscoveryNode("other-node", new TransportAddress(getLocalAddress(), 3456), ConfigConstants.FIRST_CUSTOM_SERIALIZATION_SUPPORTED_OS_VERSION);
+            testSendRequestDecorate(localNode, otherNode, false);
+        }
+
+        // remote node is on a version > 2.11.1 while local node is on version 2.11.1 - should use JDK
+        try (ThreadContext.StoredContext ignore = threadPool.getThreadContext().stashContext()) {
+            DiscoveryNode localNode = new DiscoveryNode("local-node", new TransportAddress(getLocalAddress(), 1234), Version.CURRENT);
+            DiscoveryNode otherNode = new DiscoveryNode("other-node", new TransportAddress(getLocalAddress(), 3456), Version.V_2_11_1);
```
DarshitChanpura, line 222: Is it possible enable custom serialization for 2.11.1? Can we add test for that?

DarshitChanpura (commented):
ty @parasjain1 for adding this. A general question, will there be a follow-up PR to re-enable it for future versions?

peternied (approved):
Minor comments, thanks for this contribution

peternied (changes requested):
Looks like test cases are failing due to serialization issues, please look into the failures on the integrationTest test-citests, and test dlicDlsflsTest


```
  2> java.lang.AssertionError: 
    Expected: Successful multi get response
         but: "Get an item from index: ""first-test-index" failed: "OpenSearch exception [type=exception, reason=java.lang.IllegalArgumentException: -84 is not a valid id]"
        at __randomizedtesting.SeedInfo.seed([52523052604917F5:E52F053CD1B7F4D]:0)
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
        at org.opensearch.security.FlsAndFieldMaskingTests.multiGetDocuments(FlsAndFieldMaskingTests.java:722)
  2> java.lang.AssertionError: 
    Expected: Successful multi search response
         but: "Get an item failed: ""OpenSearch exception [type=exception, reason=java.lang.IllegalArgumentException: -84 is not a valid id]"
        at __randomizedtesting.SeedInfo.seed([52523052604917F5:ECF7F2CDE87C8BDC]:0)
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
        at org.opensearch.security.FlsAndFieldMaskingTests.multiSearchDocuments(FlsAndFieldMaskingTests.java:769)
```

parasjain1:
A change was needed at SecurityFilter:187, from -

```
if (threadContext.getTransient(ConfigConstants.USE_JDK_SERIALIZATION) == null) {
                threadContext.putTransient(ConfigConstants.USE_JDK_SERIALIZATION, false);
}
```

to -

```
if (threadContext.getTransient(ConfigConstants.USE_JDK_SERIALIZATION) == null) {
                threadContext.putTransient(ConfigConstants.USE_JDK_SERIALIZATION, true);
}
```

as we now are defaulting to JDK serialization instead of custom. 

Have amended the commit to include this change. Will monitor the CI checks.

parasjain1:
FYI backward compatibility tests are expected to fail as they're running against v2.12.0 vs v3.0.0 where-in both OS versions use custom serialization, but our change assumes that any version > v2.11.0 will be using JDK serialization for now.

Not sure how can we run the suite with previous version 2.11.1. The `2.x` branch is bringing in 2.12 already.

cwperks (approved):

cwperks:
> Not sure how can we run the suite with previous version 2.11.1. The 2.x branch is bringing in 2.12 already.

@parasjain1 One way we can fix the checks is by first merging this change into 2.x, running the CI again and then merging this into main.

parasjain1:
> One way we can fix the checks is by first merging this change into 2.x, running the CI again and then merging this into main.

Got it. Should I modify this PR to set the destination branch to `2.x` ?

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
parasjain1: We are keeping custom serialization completely disabled for `2.11.1`+ until we come up with a long term fix for the DLS regression issue.

cwperks:
> Got it. Should I modify this PR to set the destination branch to 2.x

Can you keep this PR open and open another manual backport against 2.x?

peternied:
@parasjain1 Please look into the other failures all the BWC are failing - and it looks like there are spotless code style changes needed

src/main/java/org/opensearch/security/support/Base64Helper.java
parasjain1: We're trying to deserialize via JDK here. Depending on the exception message doesn't look like a good idea as it may vary across versions of JDK or a change in the underlying implementation of the `Base64JDKHelper` in future.

parasjain1:
>@parasjain1 Please look into the other failures all the BWC are failing - and it looks like there are spotless code style changes needed

As mentioned above, BWCs are expected to fail. @cwperks suggested to open a Backport to 2.x PR which can be merged first. Working on that now.

parasjain1:
Have fixed the spotless scan failures.

One of the plugin install tests are failing due to the following error which looks unrelated to the code change.

```
 > Could not resolve org.opensearch.client:opensearch-rest-high-level-client:3.0.0-SNAPSHOT.
     Required by:
         project :
      > Could not resolve org.opensearch.client:opensearch-rest-high-level-client:3.0.0-SNAPSHOT.
         > Unable to load Maven meta-data from https://d1nvenhzbhpy0q.cloudfront.net/snapshots/lucene/org/opensearch/client/opensearch-rest-high-level-client/3.0.0-SNAPSHOT/maven-metadata.xml.
            > Could not GET 'https://d1nvenhzbhpy0q.cloudfront.net/snapshots/lucene/org/opensearch/client/opensearch-rest-high-level-client/3.0.0-SNAPSHOT/maven-metadata.xml'. Received status code 403 from server: Forbidden
```

parasjain1:
All other CI tests except plugin install (mentioned in the above comment) and BWCs are passing.

parasjain1:
Tried running BWCs on the backport PR #3826 by setting the previous branch to `2.11`, but it's building OS version `2.11.1` instead of `2.11.0` leading to BWC failure yet again.

I'm highlighting below the steps that we need to carry out in the mentioned order to successfully test for backward compatibility -
* Setup explicit CI BWC test for `2.11.0` to `2.11.1`. This does not seem to be possible out of the box per how the BWC test suite and CI is setup.
* Raise and merge PR against 2.11 - this will test BWC from `2.10.x` to `2.11.1`
* Re-run BWCs on the backport PR #3826 - this will test BWC from `2.11.1` to `2.12.0` as I've added an additional commit to temporarily set the previous branch to `2.11` in `ci.yml`. We'll need to remove this commit and re-run with previous branch `2.10` before final merge.
* Re-run BWCs on this PR and merge.

peternied:
@parasjain1  This change has be lingering and there is still complexity around targeted validation.  I'm starting to thing that we should fix the root cause rather than disable the feature.  What is preventing us from fixing the root issue?

parasjain1:
> @parasjain1 This change has be lingering and there is still complexity around targeted validation. I'm starting to thing that we should fix the root cause rather than disable the feature. What is preventing us from fixing the root issue?


To reflect on the root cause, JDK de-dupes object references using back references for the objects that it has already written on the stream, while OpenSearch's StreamInput/StreamOutput implementations do not offer such a support.

The ideal fix should come as an optimization in how OpenSearch's `StreamInput` / `StreamOutput` implementations (de)serialize the Java Map and Collections. This will take some time to research and come up with a proper solution. While this happens, we need to decide whether to keep custom serde enabled in security - keeping it enabled will negatively impact users with heavy usage of DLS/FLS. 


Requesting @Bukhtawar / @backslasht to comment with their opinions regarding this.

backslasht:
I agree with @parasjain1, given that it is negatively impacting the performance to an extent that the cluster is unusable we should disable it for the time being till the fix is available. I would have preferred to go with the fix if it is straight forward, but considering the complexities involved, it would take few weeks to a month to vet out a clean workable solution and we shouldn't be having this regression for so long.

krishna-ggk:
Just wanted to throw in couple thoughts.

It sounds like the rabbit is now out of the hat - but is it still possible to introduce a feature flag and control it? That typically gives us best mechanism to bake & experiment and roll out post confidence.

Could we do something like introduce feature flag in addition to above logic? That should atleast provide a workaround for users who run into this until we have the appropriate long term fix?

mgodwan:
Sorry for chiming in late on this.

> is it still possible to introduce a feature flag and control it? 

I agree with this approach adding more checks than version. I think it will also help to control the rollout of this and drive faster development to actually gain more confidence, while helping the customers/operators who may get impacted due to this. 

The current implementation of feature flags is generally confined to node level, we can modify the logic to rely on the node attributes which are available amongst all the node and can help any node sending a request to take decision whether it should use the custom serialization or jdk one. The value of the node attribute can be populated based on the feature flag. This should help us control the roll-out across versions if needed as well.

@krishna-ggk @peternied @scrawfor99 Let me know your thoughts if this is something we can pursue for this.

mgodwan:
Sorry for chiming in late on this.

> is it still possible to introduce a feature flag and control it? 

I agree with this approach adding more checks than version. I think it will also help to control the rollout of this and drive faster development to actually gain more confidence, while helping the customers/operators who may get impacted due to this. 

The current implementation of feature flags is generally confined to node level, we can modify the logic to rely on the node attributes which are available amongst all the node and can help any node sending a request to take decision whether it should use the custom serialization or jdk one. The value of the node attribute can be populated based on the feature flag. This should help us control the roll-out across versions if needed as well.

@krishna-ggk @peternied @scrawfor99 Let me know your thoughts if this is something we can pursue for solving the problem at hand.

peternied:
@mgodwan @krishna-ggk Thank you for the thoughts and participation in this discussion. The intention behind the serialization changes were to improve performance for every node-to-node request - a huge success.

We know every customer’s needs are important, but we have to think about what’s best for everyone. Right now, turning off this feature would slow things down for a many customers, and we don’t have a good reason to do that yet.

We’re committed to providing the best balance for all our customers and have carefully considered this decision. At this time, we will not be pursuing further changes or discussions on feature flag for serialization.

Please keep me informed if there are any pull requests to address the root causes and I'll be glad to give them my attention.

Issue #3636, opened 2023-11-01
Implement tests to verify changes to default admin credentials


DarshitChanpura:
## Description:
With the completion of the changes to extract the demo configuration script into a Java tool, it is crucial to add tests to validate and ensure the accuracy of these changes. This GitHub issue is dedicated to the task of incorporating tests that will guarantee the new functionality works as expected.

## Acceptance Criteria

- [x] Add tests to ensure that demo certificate files are created
- [x] Add tests to ensure that the script accepts a custom password
- [x] Add tests to ensure that the script generates a new password if one is not provided
- [x] Add test to ensure that a call to the cluster with admin:admin fails
- [x] Add test to ensure existing options behave as is :`hyics`

By implementing these tests, the aim is to provide comprehensive coverage of the admin:admin feature, validating its functionality under various scenarios and ensuring the reliability and security of the new Java tool.

PR #3807, opened 2023-12-07
Adds tests for demo configuration java tool

DarshitChanpura:
### Description
This PR adds tests associated with the demo-configuration tool.

## Issues Resolved:
 - Resolves https://github.com/opensearch-project/security/issues/3636

### Testing
- Unit Testing
- Integration Testing

### Check List

### DIFF
--- a/src/test/java/org/opensearch/security/sanity/tests/InvalidAdminPasswordIT.java
+++ b/src/test/java/org/opensearch/security/sanity/tests/InvalidAdminPasswordIT.java
@@ -25,14 +25,16 @@ import static org.hamcrest.Matchers.is;
 
 public class InvalidAdminPasswordIT extends SecurityRestTestCase {
 
+    static String currentPasswordVariable = System.getProperty("password");
+
     @BeforeClass
     public static void setUpInvalidAdminCredentials() {
-        useAdminAsPassword = true;
+        System.setProperty("password", "admin");
     }
 
     @AfterClass
     public static void restoreValidAdminCredentials() {
-        useAdminAsPassword = false;
+        System.setProperty("password", currentPasswordVariable);
     }
 
     @Test
--- a/src/test/java/org/opensearch/security/sanity/tests/InvalidAdminPasswordIT.java
+++ b/src/test/java/org/opensearch/security/sanity/tests/InvalidAdminPasswordIT.java
@@ -1,0 +1,48 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.sanity.tests;
+
+import org.hamcrest.MatcherAssert;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import org.opensearch.client.Request;
+import org.opensearch.client.Response;
+import org.opensearch.client.ResponseException;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+public class InvalidAdminPasswordIT extends SecurityRestTestCase {
+
+    @BeforeClass
+    public static void setUpInvalidAdminCredentials() {
+        useAdminAsPassword = true;
+    }
+
+    @AfterClass
+    public static void restoreValidAdminCredentials() {
+        useAdminAsPassword = false;
+    }
+
+    @Test
+    public void testAdminCredentials_adminPassword_shouldFail() throws Exception {
+        try {
+            client().performRequest(new Request("GET", ""));
+        } catch (ResponseException e) {
+            Response res = e.getResponse();
+            MatcherAssert.assertThat(res.getStatusLine().getStatusCode(), is(equalTo(401)));
+            MatcherAssert.assertThat(res.getStatusLine().getReasonPhrase(), is(equalTo("Unauthorized")));
+        }
+    }
+}
--- a/src/test/java/org/opensearch/security/sanity/tests/InvalidAdminPasswordIT.java
+++ b/src/test/java/org/opensearch/security/sanity/tests/InvalidAdminPasswordIT.java
@@ -28,17 +28,17 @@ public class InvalidAdminPasswordIT extends SecurityRestTestCase {
     static String currentPasswordVariable = System.getProperty("password");
 
     @BeforeClass
-    public static void setUpInvalidAdminCredentials() {
+    public static void setUpAdminAsPasswordVariable() {
         System.setProperty("password", "admin");
     }
 
     @AfterClass
-    public static void restoreValidAdminCredentials() {
+    public static void restorePasswordProperty() {
         System.setProperty("password", currentPasswordVariable);
     }
 
     @Test
-    public void testAdminCredentials_adminPassword_shouldFail() throws Exception {
+    public void testAdminCredentials_adminAsPassword_shouldFail() throws Exception {
         try {
             client().performRequest(new Request("GET", ""));
         } catch (ResponseException e) {
--- a/src/test/java/org/opensearch/security/sanity/tests/SecurityRestTestCase.java
+++ b/src/test/java/org/opensearch/security/sanity/tests/SecurityRestTestCase.java
@@ -45,6 +45,8 @@ public class SecurityRestTestCase extends OpenSearchRestTestCase {
 
     private static final String CERT_FILE_DIRECTORY = "sanity-tests/";
 
+    static boolean useAdminAsPassword = false;
+
     private boolean isHttps() {
         return System.getProperty("https").equals("true");
     }
@@ -88,7 +90,7 @@ public class SecurityRestTestCase extends OpenSearchRestTestCase {
 
             // create client with passed user
             String userName = System.getProperty("user");
-            String password = System.getProperty("password");
+            String password = useAdminAsPassword ? "admin" : System.getProperty("password");
 
             return new SecureRestClientBuilder(hosts, isHttps(), userName, password).setSocketTimeout(60000)
                 .setConnectionRequestTimeout(180000)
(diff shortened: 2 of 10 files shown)


DarshitChanpura:
continuing the discussion here from #3777 :

> You've got an methods showHelp() and readOptions() both of these do not share any common code, there are many libraries such as those used by the [SecurityAdmin](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/tools/SecurityAdmin.java#L200) that handle these use cases. Then rather than checking a string output you could verify the collection - or frankly not even verify that since its just configuration at that point.

Tests should verify something that can break the API / user contract. Consider the usage of the code and work backward for what should be tested.

src/test/java/org/opensearch/security/sanity/tests/SecurityRestTestCase.java
```diff
@@ -45,6 +45,8 @@ public class SecurityRestTestCase extends OpenSearchRestTestCase {
 
     private static final String CERT_FILE_DIRECTORY = "sanity-tests/";
 
+    static boolean useAdminAsPassword = false;
```
peternied, line 48: Adding static override is a code smell, instead could the test set the property of `password` to be admin?

src/test/java/org/opensearch/security/sanity/tests/InvalidAdminPasswordIT.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @BeforeClass
+    public static void setUpInvalidAdminCredentials() {
+        useAdminAsPassword = true;
+    }
+
+    @AfterClass
+    public static void restoreValidAdminCredentials() {
+        useAdminAsPassword = false;
+    }
```
peternied, line 36: This is hard to follow, instead could we encode the username/password as a basic auth header in the way that gets a 401 and then the other where it works as expected?

src/test/java/org/opensearch/security/sanity/tests/SecurityRestTestCase.java
peternied: We should delete this

src/test/java/org/opensearch/security/sanity/tests/SingleClusterSanityIT.java
```diff
@@ -34,6 +38,13 @@ public void testSecurityPluginInstallation() throws Exception {
         verifyPluginInstallationOnAllNodes();
     }
 
+    @Test
+    public void testAdminCredentials_customStrongPassword_shouldSucceed() throws Exception {
+        Response response = client().performRequest(new Request("GET", ""));
```
peternied, line 43: This test doesn't set or verify the strength of the password, can we change the name or make the test assert something about the state of the system?

src/test/java/org/opensearch/security/sanity/tests/SingleClusterSanityIT.java
```diff
@@ -34,6 +38,13 @@ public void testSecurityPluginInstallation() throws Exception {
...
+        MatcherAssert.assertThat(response.getStatusLine().getStatusCode(), is(equalTo(200)));
+        MatcherAssert.assertThat(response.getStatusLine().getReasonPhrase(), is(equalTo("OK")));
```
peternied, line 45: nit: use static import of `assertThat(...)`

src/test/java/org/opensearch/security/tools/democonfig/CertificateGeneratorTests.java
```diff
@@ -0,0 +1,155 @@
...
+                Instant expiry = expiryDate.toInstant();
+
+                assertThat(isExpiryAtLeastAYearLater(expiry), is(equalTo(true)));
+                assertThat(x509Certificate.getSigAlgName(), is(equalTo("SHA256withRSA")));
+            } else {
+                fail("Certificate is invalid. Expected X.509 certificate.");
+            }
+        } catch (Exception e) {
+            fail("Error checking certificate validity: " + e.getMessage());
+        }
```
peternied, line 116: nit: no need to catch, `throws Exception` in the method declaration

src/test/java/org/opensearch/security/tools/democonfig/CertificateGeneratorTests.java
```diff
@@ -0,0 +1,155 @@
...
+            Certificate certificate = cf.generateCertificate(certInputStream);
+
+            if (certificate instanceof X509Certificate) {
+                X509Certificate x509Certificate = (X509Certificate) certificate;
+                x509Certificate.checkValidity();
+
+                Date expiryDate = x509Certificate.getNotAfter();
+                Instant expiry = expiryDate.toInstant();
+
+                assertThat(isExpiryAtLeastAYearLater(expiry), is(equalTo(true)));
```
peternied, line 109: nit: you can use `org.hamcrest.Matchers.greaterThan` to verify the time, then you'll get a meaningful error if the check fails.  Checking `equalTo(true)` doesn't produce a useful output

src/test/java/org/opensearch/security/tools/democonfig/CertificateGeneratorTests.java
```diff
@@ -0,0 +1,155 @@
...
+        try {
+            String pemContent = readPEMFile(keyPath);
+
+            String base64Data = pemContent.replaceAll("-----BEGIN PRIVATE KEY-----|-----END PRIVATE KEY-----", "").replaceAll("\\s", "");
+
+            byte[] keyBytes = Base64.getDecoder().decode(base64Data);
+            KeyFactory kf = KeyFactory.getInstance("RSA");
+            PrivateKey key = kf.generatePrivate(new PKCS8EncodedKeySpec(keyBytes));
+            assertThat(key.getFormat(), is(equalTo("PKCS#8")));
+            assertThat(key.getAlgorithm(), is(equalTo("RSA")));
```
peternied, line 138: 🔥

src/test/java/org/opensearch/security/tools/democonfig/CertificateGeneratorTests.java
```diff
@@ -0,0 +1,155 @@
...
+
+        for (Certificates cert : certificatesArray) {
+            String certFilePath = installer.OPENSEARCH_CONF_DIR + File.separator + cert.getFileName();
+            File certFile = new File(certFilePath);
+            assertThat(certFile.exists(), is(equalTo(true)));
+            assertThat(certFile.canRead(), is(equalTo(true)));
+
+            if (certFilePath.endsWith("-key.pem")) {
+                checkPrivateKeyValidity(certFilePath);
+                return;
```
peternied, line 77: This can exit early before everything is verified.  IMO there should be an expected number of cert generated and check should be asserted against (no loop)

src/test/java/org/opensearch/security/sanity/tests/SingleClusterSanityIT.java
DarshitChanpura: for some unknown reason, assertThat without inline import (with static import) always refers to JUnit assertThat instead of MatcherAssert's assertThat. That is why I had to resort to inline imports

src/test/java/org/opensearch/security/sanity/tests/SecurityRestTestCase.java
DarshitChanpura: refactored.

src/test/java/org/opensearch/security/sanity/tests/InvalidAdminPasswordIT.java
DarshitChanpura: since the client is build before the class is invoked, the only way to build client with invalid credentials is using @BeforeClass and then setting invalid admin password.

src/test/java/org/opensearch/security/sanity/tests/SingleClusterSanityIT.java
DarshitChanpura: the purpose of this test is to verify that admin is called with correct credentials and those correct credentials are assumed to have strong password which was used during cluster formation.

src/test/java/org/opensearch/security/tools/democonfig/CertificateGeneratorTests.java
DarshitChanpura: done

src/test/java/org/opensearch/security/tools/democonfig/CertificateGeneratorTests.java
DarshitChanpura: agreed and done.

src/test/java/org/opensearch/security/tools/democonfig/CertificateGeneratorTests.java
DarshitChanpura: refactored to verify correct behavior. I think I meant to write break instead of return here. Also added a check for expected number of certificate files.
I've kept it a loop so that any changes to Certificates.java would not trigger re-write this completely. Rather the test would fail if number of certificates files was now greater or less than 5.

DarshitChanpura:
ty @peternied for the feedback. I've addressed all comments and have refactored some tests.

src/test/java/org/opensearch/security/sanity/tests/SingleClusterSanityIT.java
peternied: Aside: Can you dig into this outside of this PR? Having unknown behavior in codebase seems like a "Bad Thing" (tm)

src/test/java/org/opensearch/security/sanity/tests/InvalidAdminPasswordIT.java
peternied: This seems to be due to the construction of the rest client, could you create a follow up issue? Not being able to change credentials during a test seems like it will hamper many commonly tested scenarios in our codebase.

peternied (approved):

src/test/java/org/opensearch/security/sanity/tests/SingleClusterSanityIT.java
DarshitChanpura: sounds good. Created an issue: https://github.com/opensearch-project/security/issues/3832

scrawfor99 (approved):
Looks good to me! Thanks DC!

Issue #2189, opened 2022-10-19
[BUG] unable to Hide unmapped backend roles


anubisg1:
**What is the bug?**
I have linked opensearch to our active directory domain.
I also mapped some AD groups to the proper roles…

everything works but i noticed that Dashboards exposes too much information to the users…
specifically when a user clicks on “view roles and identities” the user can see every Active Directory group the user name is linked to (as backend roles) even if that group is never used in OpenSearch…

this is information i would like to hide as it shows details the user shouldn’t know.
Everything i tried doesn't seem to work

this is the closest thing i found
https://opensearch.org/docs/latest/security-plugin/configuration/ldap/#advanced-exclude-roles-from-nested-role-lookups

my steps in configuration are:

1. Reducing the scope by looking for roles/AD Groups in a "sub" OU.

```
{
  "enable_ssl": false,
  "hosts": [
    "192.168.0.20:389",
    "192.168.0.21:389"
  ],
  "bind_dn": "CN=binduser,OU=Bind Users,OU=UsersOU,DC=customer,DC=company,DC=com",
  "password": "xxxxxx",
  "userbase": "OU=UsersOU,DC=customer,DC=company,DC=com",
  "usersearch": "(sAMAccountName={0})",
  "username_attribute": "sAMAccountName",
  "rolebase": "OU=OpenSearch,OU=User Groups - Customer,DC=customer,DC=company,DC=com",
  "rolesearch": "(uniqueMember={0})"
}
```
this doesn't work as i see the roles also from

`OU=User Groups - Customer,DC=customer,DC=company,DC=com` 

2. configured nested role filter

i added the following configuration
```
                    resolve_nested_roles: true
                    nested_role_filter:
                      - '^(?!.*OpenSearch.*$)'
```
that should match (and therefore filterout) everything except groups which have “OpenSearch” in them

![image](https://user-images.githubusercontent.com/4518931/196626901-a3c1e3c4-5667-42a8-8dfb-c01d42fe2553.png)

but also doesn't work as i still see all the roles.

**What is the expected behavior?**
Only roles used by OpenSearch should be visible to the user as backend-roles
Also, filtering of roles, should work as expected 

**What is your host/environment?**
 - OS: Azure AKS
 - Version - 2.3.0
 - Plugins - nothing manually installed, only default plugins installed by the k8s operator

**Do you have any screenshots?**
[...]

PR #3809, opened 2023-12-08
Hide ldap roles

MaciejMierzwa:
### Description

* Category: Enhancement, New feature
To allow admin to have more control on what backend roles user sees. 
Old behavior is that all user's roles are fetched from ldap/ad. Even if role is not used in opensearch use can view all assigned roles.
New behavior is that admin can configure what roles will be visible to the user. If flag is not configured cluster will behave the same. With new config admin is able to hide part of roles from users using simple regex.
Example config:
```
    authz:
      ldap_roles:
        description: "Authorize using LDAP"
        http_enabled: true
        transport_enabled: true
        authorization_backend:
          type: ldap
          config:
            enable_ssl: false
            enable_start_tls: false
            enable_ssl_client_auth: false
            verify_hostnames: false
            hosts:
            - 127.0.1.1:390
            bind_dn: cn=readonly,dc=example,dc=org
            password: changethistoo
            userbase: ou=People,dc=example,dc=org
            usersearch: (cn={0})
            username_attribute: cn
            skip_users:
              - admin
              - kibanaserver
            rolebase: ou=Groups,dc=example,dc=org
            rolesearch: (uniqueMember={0})
            userroleattribute: null
            userrolename: disabled
            rolename: cn
            resolve_nested_roles: true
  new ->    exclude_roles: "(this_role_will_be_hidden.*)"

```

### Issues Resolved
https://github.com/opensearch-project/security/issues/2189

### Testing
Unit testing and manual testing performed

### Check List

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/ldap/backend/LDAPAuthorizationBackend.java
+++ b/src/main/java/com/amazon/dlic/auth/ldap/backend/LDAPAuthorizationBackend.java
@@ -964,8 +964,7 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
                 for (final LdapName roleLdapName : nestedReturn) {
                     final String role = getRoleFromEntry(connection, roleLdapName, roleName);
 
-                    if ((Strings.isNullOrEmpty(excludeRoles) && !Strings.isNullOrEmpty(role))
-                        || (!Strings.isNullOrEmpty(excludeRoles) && !Strings.isNullOrEmpty(role) && !Pattern.matches(excludeRoles, role))) {
+                    if (filterRole(excludeRoles, role)) {
                         user.addRole(role);
                     } else {
                         log.warn("Role not allowed or empty, attribute: '{}' for entry: {}", roleName, roleLdapName);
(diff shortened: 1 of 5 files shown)


cwperks:
@MaciejMierzwa Can `exclude_roles` support a list as well?

MaciejMierzwa:
@cwperks 
Current implementation accepts regex. In case of multiple roles to filter out user can do it as following:
```
exclude_roles: "(excluded_role1|excluded_role2|excluded_role1|another_one.*)"
```
to exclude roles:
- excluded_role1
- excluded_role2
- excluded_role1
- another_one<any any string>

scrawfor99:
Hey @MaciejMierzwa just checking in to see if you need help driving this home? Thanks for all your help :)

scrawfor99:
Hi @MaciejMierzwa, you will want to file an issue over on the documentation website repo explaining the change and then someone there should be able to help you.

cwperks (commented):
@MaciejMierzwa This looks good to me, but one comment I have is that its inconsistent with the `skip_users` setting which has a list of users (I believe it accepts wildcards too):

```
skip_users:
    - admin
    - kibanaserver
```

Do you think it makes sense to name the setting `exclude_roles_regex` to differentiate it as a regex instead of a list or do you think documentation will suffice?

MaciejMierzwa:
@cwperks I guess we could do either of those options, or even circle back to your suggestion from https://github.com/opensearch-project/security/pull/3809#issuecomment-1847451658. Let me know what do you think

scrawfor99:
Hi @MaciejMierzwa and @cwperks to keep this moving forward lets just keep things as is for now and add some documentation we can always refactor this if need be but I don't want this to stall.

cwperks:
IMO I think supporting this as a list would be more consistent with other LDAP settings like `skip_users`.

DarshitChanpura:
Hi @MaciejMierzwa , IMO we should support wildcard-entries. Please let us know if you would need anything from us to keep this moving forward.

cwperks:
Hey @MaciejMierzwa , I modified this PR on my fork take in a list of globs for `exclude_roles` (https://github.com/cwperks/security/pull/20). A list of globs would be consistent with other configuration settings for the LDAP Authorization backend like `skip_users`. Let me know what you think.

DarshitChanpura:
@cwperks should this PR be closed in lieu of #4025

PR #3825, opened 2023-12-12
Add do not fail on forbidden test cases around the stats API

peternied:
### Description
There was some question of what permissions around DNFOF applied to scenarios with the stats API, adding in test cases that confirm these APIs are behaving as expected.

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
+++ b/src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
@@ -445,6 +445,7 @@ public class DoNotFailOnForbiddenTests {
 
     @Test
     public void checkStatsApi() {
+        // As admin creates 2 documents in different indices, can find both indices in search, cat indice & stats APIs
         try (final TestRestClient client = cluster.getRestClient(ADMIN_USER.getName(), ADMIN_USER.getPassword())) {
             final HttpResponse createDoc1 = client.postJson("hi1/_doc?refresh=true", "{\"hi\":\"Hello1\"}");
             createDoc1.assertStatusCode(SC_CREATED);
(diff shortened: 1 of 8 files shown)


src/integrationTest/java/org/opensearch/test/framework/cluster/TestRestClient.java
```diff
@@ -109,17 +105,8 @@ public TestRestClient(InetSocketAddress nodeHttpAddress, List<Header> headers, S
         this.sourceInetAddress = sourceInetAddress;
     }
 
-    public HttpResponse get(String path, List<NameValuePair> queryParameters, Header... headers) {
-        try {
-            URI uri = new URIBuilder(getHttpServerUri()).setPath(path).addParameters(queryParameters).build();
```
peternied, line 114: `URIBuilder(...).setPath(path)` has much different behavior for how URLs are constructed that was not aligning well with how `post/delete/...` paths were constructed, URLs would effectively have different rules.  Removing this alternative approach cleans up an inconsistancy in how these test cases work so all REST invocations follow the same path'ing rules.

peternied: Also, this parameterized constructor wasn't used.  Note; we can always add support for URIBuilder type construction in the future if tests warrant it.

willyborankin (approved):

cwperks (approved):

DarshitChanpura (approved):

src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
```diff
@@ -434,4 +443,39 @@ public void shouldPerformCatIndices_positive() throws IOException {
...
+    public void checkStatsApi() {
+        // As admin creates 2 documents in different indices, can find both indices in search, cat indice & stats APIs
+        try (final TestRestClient client = cluster.getRestClient(ADMIN_USER.getName(), ADMIN_USER.getPassword())) {
+            final HttpResponse createDoc1 = client.postJson("hi1/_doc?refresh=true", "{\"hi\":\"Hello1\"}");
+            createDoc1.assertStatusCode(SC_CREATED);
+            final HttpResponse createDoc2 = client.postJson("hi2/_doc?refresh=true", "{\"hi\":\"Hello2\"}");
+            createDoc2.assertStatusCode(SC_CREATED);
+
+            final HttpResponse search = client.postJson("hi*/_search", "{}");
+            assertThat("Unexpected document results in search:" + search.getBody(), search.getBody(), containsString("2"));
```
derek-ho, line 456: Would it be better to assert something more specific here? I am assuming the 2 you want here is coming from the count of documents, but it oculd also be coming from hi2/hello2?

peternied: That's a good callout to improve the test - this could be replaced with `            assertThat(search.getIntFromJsonBody("/hits/total/value"), equalTo(2));`  However, I think we can live with the current version, and if we expand on these tests that is a good area to invest in

Issue #3847, opened 2023-12-14
[FEATURE] Clear text OPENSEARCH_INITIAL_ADMIN_PASSWORD and OPENSEARCH_INITIAL_ADMIN_PASSWORD_TXT are insecure


dblock:
**Is your feature request related to a problem?**

Storing passwords in the clear is insecure. The implementation that was chosen in #3560 seems to ask for the admin password to be provided either via file (permanently stored) or via environment.

**What solution would you like?**

1. Reduce the insecure surface by removing one of the two ways to specify the password, I recommend getting rid of the file.txt support.
2. Require that the password be a password hash (generated using the equivalent of `./plugins/opensearch-security/tools/hash.sh`).

**What alternatives have you considered?**

Do nothing.

**Do you have any additional context?**

https://code.dblock.org/2023/08/08/changing-the-default-admin-password-in-opensearch.html

derek-ho:
Closing as it was done in #3850

PR #3850, opened 2023-12-14
Removes support for txt file for specifying initial admin password and prevents password from being printed in the logs

DarshitChanpura:
### Description
Storing an admin password in a text file that can be available to anyone that has access to config folder at any time could be potentially vulnerable. Although, originally added with the perception to support autonomous process to easily read from the config folder, we now only allow passing the admin password through environment variable as it is session-scoped. Further, we don't print the password to the logs to avoid any traces.

- Related to https://github.com/opensearch-project/security/issues/3847

### Testing
- automated tests

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
+++ b/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
@@ -132,7 +132,7 @@ public class SecuritySettingsConfigurer {
 
             // Read custom password from environment variable
             String initialAdminPassword = System.getenv().get(ConfigConstants.OPENSEARCH_INITIAL_ADMIN_PASSWORD);
-            if (initialAdminPassword != null && !initialAdminPassword.isEmpty()) {
+            if (!Strings.isNullOrEmpty(initialAdminPassword)) {
                 ADMIN_PASSWORD = initialAdminPassword;
             }
 
--- a/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
+++ b/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
@@ -93,6 +93,30 @@ public class SecuritySettingsConfigurer {
     }
 
     /**
+     * Checks if security plugin is already configured. If so, the script execution will exit.
+     */
+    void checkIfSecurityPluginIsAlreadyConfigured() {
+        // Check if the configuration file contains the 'plugins.security' string
+        if (installer.OPENSEARCH_CONF_FILE != null && new File(installer.OPENSEARCH_CONF_FILE).exists()) {
+            try (BufferedReader br = new BufferedReader(new FileReader(installer.OPENSEARCH_CONF_FILE, StandardCharsets.UTF_8))) {
+                String line;
+                while ((line = br.readLine()) != null) {
+                    if (line.toLowerCase().contains("plugins.security")) {
+                        System.out.println(installer.OPENSEARCH_CONF_FILE + " seems to be already configured for Security. Quit.");
+                        System.exit(installer.skip_updates);
+                    }
+                }
+            } catch (IOException e) {
+                System.err.println("Error reading configuration file.");
+                System.exit(-1);
+            }
+        } else {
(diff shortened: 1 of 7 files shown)


DarshitChanpura:
I'm updating the plugin install workflow rn, since it relied on text file being generated for passing admin password

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -34,6 +34,7 @@
 import org.opensearch.security.support.ConfigConstants;
 import org.opensearch.security.tools.Hasher;
 
+import joptsimple.internal.Strings;
```
cwperks, line 37:
```suggestion
import org.opensearch.core.common.Strings;
```

derek-ho: 👍

DarshitChanpura: ah..good catch!

src/test/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurerTests.java
```diff
@@ -79,42 +76,25 @@ public void tearDown() throws NoSuchFieldException, IllegalAccessException {
...
-        securitySettingsConfigurer.updateAdminPassword();
-
-        assertEquals(customPassword, SecuritySettingsConfigurer.ADMIN_PASSWORD);
-        verifyStdOutContainsString("ADMIN PASSWORD SET TO: " + customPassword);
+        verifyStdOutContainsString("Admin password set successfully.");
     }
 
     @Test
     public void testUpdateAdminPassword_noPasswordSupplied() {
+        SecuritySettingsConfigurer.ADMIN_PASSWORD = ""; // to ensure 0 flaky-ness
```
derek-ho, line 97: Why is this necessary?

DarshitChanpura: flkay-ness caused probly due to static variable.

DarshitChanpura: in my local setup, I saw this error once in about 5 runs. After adding this fix I was able to verify that it is no longer flaky. (ran about 20 times locally)

DarshitChanpura: In actual run, this static variable will be assigned only once, but for testing it maybe assigned multiple times which can cause flaky-ness.

derek-ho (approved):
LGTM

src/test/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurerTests.java
DarshitChanpura: here is an example run: https://github.com/opensearch-project/security/actions/runs/7213760630/job/19656615873?pr=3844

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -107,21 +130,10 @@ void updateAdminPassword() {
                     .build()
             );
 
-            // Read custom password
+            // Read custom password from environment variable
+            String initialAdminPassword = System.getenv().get(ConfigConstants.OPENSEARCH_INITIAL_ADMIN_PASSWORD);
             if (initialAdminPassword != null && !initialAdminPassword.isEmpty()) {
```
cwperks, line 135: nit: Use `Strings.inNullOrEmpty` here.

cwperks (approved):

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
DarshitChanpura: done.

cwperks (approved):

peternied (approved):

PR #3868, opened 2023-12-18
Ensure Github workflow runs on docker image used by Production Distribution Build

reta:
### Description
Ensure Github workflow runs on docker image used by Production Distribution Build

### Issues Resolved
Closes https://github.com/opensearch-project/security/issues/3494

### Testing
Tested as Github Action worflows

### Check List

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,11 @@ env:
   CI_ENVIRONMENT: normal
 
 jobs:
+  Get-CI-Image-Tag:
+    uses: opensearch-project/opensearch-build/.github/workflows/get-ci-image-tag.yml@main
+    with:
+      product: opensearch
+
   generate-test-list:
     runs-on: ubuntu-latest
     outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -63,13 +102,13 @@ jobs:
     - uses: alehechka/upload-tartifact@v2
       if: always()
       with:
-        name: ${{ matrix.platform }}-JDK${{ matrix.jdk }}-${{ matrix.gradle_task }}-reports
+        name: windows-latest-JDK${{ matrix.jdk }}-${{ matrix.gradle_task }}-reports
         path: |
           ./build/reports/
 
   report-coverage:
     needs:
```
peternied, line 110: Pretty sure we need `test-linux` too, otherwise it won't incorporate the run data from the linux runners

.github/workflows/ci.yml
```diff
@@ -13,6 +13,11 @@ env:
   CI_ENVIRONMENT: normal
 
 jobs:
+  Get-CI-Image-Tag:
+    uses: opensearch-project/opensearch-build/.github/workflows/get-ci-image-tag.yml@main
```
peternied, line 17: Lets use a published task not a direct repo reference

.github/workflows/ci.yml
```diff
@@ -32,16 +37,50 @@ jobs:
...
+      image: ${{ needs.Get-CI-Image-Tag.outputs.ci-image-version-linux }}
+      # need to switch to root so that github actions can install runner binary on container without permission issues.
+      options: --user root
+
+    steps:
+    - uses: actions/checkout@v3
+      name: Checkout security
+    - name: Build and Test
+      run: |
+        # https://github.com/opensearch-project/opensearch-build/issues/4191
```
peternied, line 64: This issue is marked as fixed, can this be cleaned up?

peternied (commented):
Thanks for getting a PR out here to review, can you do a runtime comparison between main and this change? I'm curious what the impact is to our PR checks time with this change.

.github/workflows/ci.yml
reta: > Pretty sure we need test-linux too, otherwise it won't incorporate the run data from the linux runners

Absolutely,  I am trying to understand why this job fails (no artifacts)

.github/workflows/ci.yml
reta: Hm... may you share the example? (PS: this step is replicated to all repos)

peternied: I don't like tightly coupling these repositories together; security depends on opensearch-build, opensearch build depends on security - this is a formula to get into a broken state and have to start bypassing CI.

peternied: For example, here is a GitHub action that allows for multi-platform file download [1] it shouldn't be part of the security project even though it was originally written to support this project. 

I have trouble adopting this level of coupling when we still have problems doing normal builds after version bumps [2]

- [1] https://github.com/peternied/download-file
- [2] https://github.com/opensearch-project/security/pull/3829

reta: I am still unsure what you are suggesting, my apologies. If this is not the way we should share the workflows, no objections to change it, but to what model?

reta:
> Thanks for getting a PR out here to review, can you do a runtime comparison between main and this change? I'm curious what the impact is to our PR checks time with this change.

The only change should be an usage of our (OpenSearch) image for Linux runner (instead of ubuntu-latest), there  only price to be paid is for pulling the image .

.github/workflows/ci.yml
peternied: We should publish the workflows as github actions e.g. `opensearch-project/opensearch-ci-image-by-tag@v1`?

reta: Oh I see, thanks @peternied , @peterzhuamazon fyi ^^

reta:
@peternied opening it up for review, if you prefer to have Githib Action (instead of referencing the workflow directly), I have no objection - please feel to close this one (and preferably open an issue for `opensearch-build` repo to work on one), thank you!

peternied:
Thanks @reta I'm created an issue for OpenSearch-build to publish a GitHub action [1] that should be how we adopt that job.  After this has been created, we'll be in a position to adopt this change in this repo.

- [1] https://github.com/opensearch-project/opensearch-build/issues/4296

PR #3869, opened 2023-12-19
Dynamic sign in options

davidosorno:
### Description
**New Feature**
Allow admins to define the sign-in options that will be displayed on OpenSearch Dashboard login page. There are couple of sign-in options defined in [Security documentation](https://opensearch.org/docs/latest/security/configuration/multi-auth/), and theses options must be available in security _config.yml_ file to be able to change them dynamically in Security Dashboard.

*Old Behavior*
Admins have to update _opensearch_dashboards.yml_ adding or removing sign-in options, and then restart Dashboards to be able to log in using other sign-in option.

*New Behavior*
Admins can change sign-in options dynamically without having to restart the Dashboards, and the changes are applied immediately. Users just need to logout in order to see the sign-in options available.

### Issues Resolved
https://github.com/opensearch-project/security-dashboards-plugin/issues/1573

### Testing
Unit Testing, Integration Testing, and Manual Testing.

### Check List


src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
```diff
@@ -177,6 +183,19 @@ private void updateAndValidatesValues(final ConfigV7 config, final JsonNode json
         if (Objects.nonNull(jsonContent.findValue(MULTITENANCY_ENABLED_JSON_PROPERTY))) {
             config.dynamic.kibana.multitenancy_enabled = jsonContent.findValue(MULTITENANCY_ENABLED_JSON_PROPERTY).asBoolean();
         }
+        if (Objects.nonNull(jsonContent.findValue(DASHBOARD_SIGNIN_OPTIONS))
```
willyborankin, line 186: I think `jsonContent.hasNonNull` is better here.

src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
```diff
@@ -177,6 +183,19 @@ private void updateAndValidatesValues(final ConfigV7 config, final JsonNode json
...
+            && jsonContent.findValue(DASHBOARD_SIGNIN_OPTIONS).size() > 0) {
```
willyborankin, line 187: Same: `jsonContent.findValue(DASHBOARD_SIGNIN_OPTIONS).isEmpty()`

src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
```diff
@@ -33,6 +34,7 @@
 import org.opensearch.security.dlic.rest.validation.RequestContentValidator;
 import org.opensearch.security.dlic.rest.validation.RequestContentValidator.DataType;
 import org.opensearch.security.securityconf.impl.CType;
+import org.opensearch.security.securityconf.impl.DashboardSignInOption;
```
cwperks, line 37: wdyt about renaming this class to `DashboardsConfigApiAction`?

src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
```diff
@@ -177,6 +183,19 @@ private void updateAndValidatesValues(final ConfigV7 config, final JsonNode json
         if (Objects.nonNull(jsonContent.findValue(MULTITENANCY_ENABLED_JSON_PROPERTY))) {
             config.dynamic.kibana.multitenancy_enabled = jsonContent.findValue(MULTITENANCY_ENABLED_JSON_PROPERTY).asBoolean();
         }
+        if (Objects.nonNull(jsonContent.findValue(DASHBOARD_SIGNIN_OPTIONS))
+            && jsonContent.findValue(DASHBOARD_SIGNIN_OPTIONS).size() > 0) {
+            JsonNode newOptions = jsonContent.findValue(DASHBOARD_SIGNIN_OPTIONS);
```
cwperks, line 188: wdyt about taking advantage of streams?

i.e.

```
List<String> newOptions = jsonContent.findValuesAsText(DASHBOARD_SIGNIN_OPTIONS);
List<DashboardSignInOption> options = newOptions.stream().map(DashboardSignInOption::valueOf).collect(Collectors.toList());
```

cwperks (commented):
Thank you @davidosorno and @EduardoCorazon. This PR looks good to me. Left a couple of suggestions.

src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
davidosorno: This class is just an enum to match Sign In values, so I think `DashboardsConfigApiAction` is not an appropriate name because it doesn't do anything else.

src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
davidosorno: Sure, thanks for the feedback.

src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
cwperks: Woops, I added my comment on this line because it was the closest to the top of the changes in this file. The comment is not refereing to `DashboardsSigninOptions`, but rather `MultiTenancyConfigApiAction` (the name of the class of the file). wdyt about change the name from `MultiTenancyConfigApiAction` to `DashboardsConfigApiAction`?

cwperks: I may file an issue to make this class more generic in general which would need to be done in tandem with the frontend since the routes in this class begin with `/tenancy/...`.

willyborankin (dismissed):

peternied (changes requested):
Nice job plumbing this through, I think we need a couple more test cases around different positive scenarios.

Before we merge lets make the validation clearer:
- Is the response is when Enum::valueOf doesn't have a match actionable? I hope their is a clear error message
- What should happen in the authc configuration only saml is supported?  I would think the other sign in types be invalid, how do you think this should be handled?

src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
davidosorno: @cwperks  I tried, however `findValuesAsText` returns empty.
Also, the returned values using `JsonNode newOptions = jsonContent.findValue(DASHBOARD_SIGNIN_OPTIONS);` are json nodes, it needs to access to each node by its index, so I had to go back to the previous version.

src/test/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiTest.java
```diff
@@ -148,6 +152,18 @@ private void verifyTenantUpdateFailed(final Header... header) throws Exception {
             setRandomStringAsDefaultTenant.findValueInJson("error.reason"),
             containsString("Default tenant should be selected from one of the available tenants.")
         );
+
+        final HttpResponse signInOptionsNonArrayValue = rh.executePutRequest(
+            "/_plugins/_security/api/tenancy/config",
+            "{\"dashboard_signin_options\": \"BASIC\"}",
+            header
+        );
+        assertThat(signInOptionsNonArrayValue.getStatusCode(), equalTo(HttpStatus.SC_BAD_REQUEST));
```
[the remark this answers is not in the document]
davidosorno: I already added some test when updating sign in options.
> * What should happen in the authc configuration only saml is supported?  I would think the other sign in types be invalid, how do you think this should be handled?

I agree that the other options should be invalid. I'm still working on it.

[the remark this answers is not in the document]
davidosorno: I'm thinking about adding a function like `boolean isAllowedOption()` to [public interface HTTPAuthenticator](https://github.com/opensearch-project/security/blob/31f1625742792b7033fa8a24781b95695ef2e1a2/src/main/java/org/opensearch/security/auth/HTTPAuthenticator.java#L54), so any class that implements it must validate the option before let the user sign in.
What do you think? any other ideas?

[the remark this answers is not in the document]
peternied: I'm not sure why an authenticator would know if it was allowed or not, maybe push a change and make a comment so we can look at it?

cwperks (dismissed):
Thank you for adding all of the tests @davidosorno. This PR looks good to me.

I would like to rename the `MultiTenancyConfigApiAction` to a more general `DashboardsConfigApiAction`, but that can be done on a future PR.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -171,6 +172,7 @@ public void onDynamicConfigModelChanged(DynamicConfigModel dcm) {
 
         invalidateCache();
         anonymousAuthEnabled = dcm.isAnonymousAuthenticationEnabled()// config.dynamic.http.anonymous_auth_enabled
+            && dcm.getDashboardSignInOptions().contains(DashboardSignInOption.ANONYMOUS)
```
peternied, line 175: This is the opposite behavior to what I would expect in the comment thread around validation [1].  If you had HttpBasic and Anonymous auth enabled in the authc config section of the security plugin, if a request set the sign-in option to be SAML, I would expect an error - because SAML isn't enabled.  It seems like the options that would be accepted would be either one of the enabled authenticators, both of them, or none - how does that behavior seem to you?

- [1] https://github.com/opensearch-project/security/pull/3869#discussion_r1442982969

davidosorno: This condition is just to ensure that Anonymous sign-in option is enabled in the options available, so, after enabling Anonymous authentication in `config.yml` and in `opensearch_dashboards.yml` the admin can enable or disable this option in Dashboard Sign In Options.
Until now if both options SAML and Anonymous are enabled in this panel, Anonymous option will take precedence, so SAML authentication will fail. Furthermore, if Anonymous options is not enabled, SAML option will work fine.
Finally, there is not need for modifying the the yml files anymore to enable or disable these options.

image

cwperks: @peternied We uncovered a bug where having Anonymous Login + SAML as dashboards sign in options does not work. There is logic on the backend where the frontend is unable to get to redirect info it needs to go to the external IdP because anonymous is enabled.

I will write up an issue this afternoon describe the existing bug. 

Thank you for pointing out this line, this line should be removed from this PR and the underlying bug should be addressed so that SAML + Anonymous can work together as dashboards sign in options.

cwperks: Filed an issue describing the existing bug with SAML and anonymous: https://github.com/opensearch-project/security-dashboards-plugin/issues/1731

davidosorno:
> Hi @davidosorno, something wrong with your PR. Your branch has commits from 2.x branch.

@willyborankin  Sorry, my very bad, somehow I got confused with the branch I was working on, and since I needed to undo the last commit, I did the `reset --soft`, then I made sure the last change I made was in Staged mode and I removed it. Then I just push the branch to the origin with `-f` option, and here is where everything went wrong.
I realized immediately about the error I made, but I wasn't sure where, so I started investigating, and finally I figured it out and fixed it.
Thanks for checking.

src/integrationTest/java/org/opensearch/security/api/DashboardsInfoTest.java
```diff
@@ -53,4 +55,17 @@ public void testDashboardsInfoValidationMessage() throws Exception {
...
     }
+
+    @Test
+    public void testDashboardsInfoContainsSignInOptions() throws Exception {
+
+        try (TestRestClient client = cluster.getRestClient(DASHBOARDS_USER)) {
+            TestRestClient.HttpResponse response = client.get("_plugins/_security/dashboardsinfo");
+            assertThat(response.getStatusCode(), equalTo(HttpStatus.SC_OK));
+            assertThat(
+                response.getTextArrayFromJsonBody("/dashboard_signin_options").contains(DashboardSignInOption.BASIC.toString()),
```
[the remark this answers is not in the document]
davidosorno: There should be just 1 dashboard sign in option enable when Security plugin is installed, then the admin can enable the other options using the dashboard security panel.
Also, there is another test where it updates these sign in options.

src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
davidosorno: @cwperks, I agree about the class name change and as you mentioned, it can be another issue so it aligns with the work that needs to be done.

src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
```diff
@@ -119,7 +122,9 @@ public Map<String, DataType> allowedKeys() {
                             PRIVATE_TENANT_ENABLED_JSON_PROPERTY,
                             DataType.BOOLEAN,
                             MULTITENANCY_ENABLED_JSON_PROPERTY,
-                            DataType.BOOLEAN
+                            DataType.BOOLEAN,
+                            DASHBOARD_SIGNIN_OPTIONS,
```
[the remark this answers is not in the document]
davidosorno: I used spotlessApply to format the file and this is the result.

src/test/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiTest.java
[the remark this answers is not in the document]
davidosorno: @peternied The last commit allows to enable/disable sign in options ability when admin updates them using the dashboard security panel.
I'll add some tests, but I push it so I can get some feedback.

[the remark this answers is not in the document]
peternied: I think you are missing two tests cases that roughly should look like:

### Valid 
```java
// Validate starting state
final HttpResponse originalConfig = rh.executeGetRequset("/_plugins/_security/api/tenancy/config");
assertThat(originalConfig.getTextArrayFromJsonBody("/dashboard_signin_options"), equalTo(List.of("SAML")))

// Update config
final HttpResponse updateSignInOption = rh.executePutRequest(
    "/_plugins/_security/api/tenancy/config",
    "{\"dashboard_signin_options\": [\"BASIC\"]}",
    header
);
updateSignInOption.assertStatusCode(HttpStatus.SC_OK);

// Validate starting state
final HttpResponse updatedConfig = rh.executeGetRequset("/_plugins/_security/api/tenancy/config");
assertThat(updatedConfig.getTextArrayFromJsonBody("/dashboard_signin_options"), equalTo(List.of("BASIC"))) 
```

### Unavailable SignIn Option 
```java
// ...Configure cluster to only support BASIC AUTHC provider...

// Update config
final HttpResponse updateUnavailableSignInOption = rh.executePutRequest(
    "/_plugins/_security/api/tenancy/config",
    "{\"dashboard_signin_options\": [\"SAML\"]}",
    header
);
updateUnavailableSignInOption.assertStatusCode(HttpStatus.SC_BAD_REQUEST);

// Validate starting state
final HttpResponse updatedConfig = rh.executeGetRequset("/_plugins/_security/api/tenancy/config");
assertThat(invalidSignInOption.findValueInJson("error.reason"), equalTo("Validation failure; no SAML authentication provider are available for this cluster"));
```

DarshitChanpura:
Hi @davidosorno, the changes looks good to me. Would you please address the conflicts?

src/test/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiTest.java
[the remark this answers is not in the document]
davidosorno: Thank you very much @peternied for the guidance.

src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
```diff
@@ -202,4 +221,30 @@ private void updateAndValidatesValues(final ConfigV7 config, final JsonNode json
...
+    }
+
+    /**
+     * It enables/disables the authentication options available just for security dashboard sign-in options.
+     */
+    private void toggleAuthcDashboardSignInAbility(Authc authc, List<DashboardSignInOption> options, boolean enable) {
+        for (DashboardSignInOption option : options) {
+            for (String key : authc.getDomains().keySet()) {
+                if (key.contains(option.toString().toLowerCase())) {
+                    authc.getDomains().get(key).http_enabled = enable;
```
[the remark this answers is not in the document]
davidosorno: Hi @cwperks, the assumption I made to change `http_enabled` is :
If an admin is using the Security Dashboard to enable or disable sign in options, and let's say the BASIC authentication is removed, other users still could get information from the API using an http requests.
In the image attached there are 3 request and you can see the following:

1. I made a request using admin and password and I got the information since I was authenticated correctly.
2. I disabled the BASIC authentication using Security Dashboard, tried to get the same information, and I got authentication failed since the option was disabled (isn't it the expected behavior?).
3. I removed the code you commented and as you can see even though the BASIC authentication is not among the available options I still got the information. isn't that bad? 

If I'm using Security Dashboard and I don't want the users authenticate using BASIC authentication, why should be the option still available?

image

Also, there is already an implementation in Security Dashboard where the option using to sign-in is validated before sending the request.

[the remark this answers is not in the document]
cwperks: > I removed the code you commented and as you can see even though the BASIC authentication is not among the available options I still got the information. isn't that bad?

Its not bad and its actually quite common. Some clusters may have SAML configured for sign in to dashboards, but also have basic auth where they can connect programmatically with an opensearch client. You're right that it is often the case that the authenticators enabled on the backend === the sign in options, but its not necessarily always the case and we should support that. 

This feature will exclusively control the sign in options that are displayed in OSD, but not enable or disable the authenticators on the backend. If a cluster admin wants to enable or disable an authenticator on the backend they would need to use the [security admin tool](https://opensearch.org/docs/latest/security/configuration/security-admin/) and connect with the admin certificate to apply changes.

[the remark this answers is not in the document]
davidosorno: Oh, I understand now. @cwperks Thank you very much for the explanation.

src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
```diff
@@ -177,6 +183,12 @@ private void updateAndValidatesValues(final ConfigV7 config, final JsonNode json
         if (Objects.nonNull(jsonContent.findValue(MULTITENANCY_ENABLED_JSON_PROPERTY))) {
             config.dynamic.kibana.multitenancy_enabled = jsonContent.findValue(MULTITENANCY_ENABLED_JSON_PROPERTY).asBoolean();
         }
+        if (jsonContent.hasNonNull(DASHBOARD_SIGNIN_OPTIONS) && jsonContent.findValue(DASHBOARD_SIGNIN_OPTIONS).isEmpty() == false) {
+            JsonNode newOptions = jsonContent.findValue(DASHBOARD_SIGNIN_OPTIONS);
+            List<DashboardSignInOption> options = getNewDashboardSignInOptions(newOptions);
+            config.dynamic.kibana.dashboardSignInOptions = options;
```
cwperks, line 189: The standard for the security config is lower_snake_case. Can this be updated to `dashboards_sign_in_options` or shortened to `sign_in_options`.

This PR looks good to me otherwise!

DarshitChanpura (commented):
Looks good to me. I agree with Craig here about using snake_case for naming dashboard variables. Mind addressing that change @davidosorno ?

src/test/resources/restapi/securityconfig_nondefault.json
```diff
@@ -9,7 +9,8 @@
       "private_tenant_enabled" : true,
       "default_tenant" : "",
       "server_username" : "kibanaserver",
-      "index" : ".kibana"
+      "index" : ".kibana",
+      "dashboardSignInOptions":["BASIC"]
```
DarshitChanpura, line 13: this should be `sign_in_options`

src/test/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiTest.java
```diff
@@ -54,9 +57,22 @@ private void verifyTenantUpdate(final Header... header) throws Exception {
             setPrivateTenantAsDefaultResponse.getStatusCode(),
             equalTo(HttpStatus.SC_OK)
         );
+        assertThat(getDashboardsinfoResponse.findArrayInJson("sign_in_options"), hasItem(DashboardSignInOption.BASIC.toString()));
+        assertThat(getDashboardsinfoResponse.findArrayInJson("sign_in_options"), not(hasItem(DashboardSignInOption.SAML.toString())));
+
+        final HttpResponse updateDashboardSignInOptions = rh.executePutRequest(
```
DarshitChanpura, line 63: nit: missed renaming to `signInOptions`

DarshitChanpura (commented):
Looks good to me @davidosorno . Once the below comments are addressed the CI should pass.

cwperks (dismissed):

DarshitChanpura (approved):
LGTM. Thanks @davidosorno for addressing all comments!

DarshitChanpura:
@peternied This PR is currently blocked by your requested change. Would you please re-review this?

peternied:
This PR uses the security's backend for data storage with no reasoning about the correctness of the values set.  Until this is addressed I don't think this change should be merged.  If we don't want to perform this reasoning I would advocate that this scenario is supported by using `.kibana` index to dynamically config this setting post boot.

@DarshitChanpura What do you think?

- Outstanding comment with requested test cases https://github.com/opensearch-project/security/pull/3869#discussion_r1433176251

src/test/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiTest.java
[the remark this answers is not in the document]
DarshitChanpura: +1 Let's add tests to ensure that out of the given 4 dashboard sign-in options, only the ones configured are supported. Putting it in other way, what would happen if say OPENID option is set but no open-id is configured?

DarshitChanpura (commented):
My understanding is that this is a security config related update like multi-tenancy toggle which we are trying to make dynamically change-able, and all security config related updates have been part of security backend. I agree with adding couple more test cases to ensure that correct data is being displayed, and I think that using security backend along with tests that cover all bases should suffice to add this change. @davidosorno Would you mind addressing the outstanding comment about adding tests for all scenarios?

willyborankin (approved):

scrawfor99:
@cwperks what do you want to do with this? I know @davidosorno is not going to be able to finish this so should we close it or do you have plans to take this over?

davidosorno:
Hi @scrawfor99, I can fix anything to this PR if needed, I'm just waiting @peternied unblock it and the PR gets more reviews.

src/test/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiTest.java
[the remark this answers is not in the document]
peternied: @davidosorno Can you follow up on this thread?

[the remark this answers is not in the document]
davidosorno: Hi @peternied, [please review this tests](https://github.com/davidosorno/security/blob/8c7343bb37f9559520a19d5959df3c05e0b60a0a/src/test/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiTest.java#L61), isn't it covering the test case @DarshitChanpura mentioned? 
It ensures that SAML option is not configured yet, then it updates the sign in options and the SAML options is found.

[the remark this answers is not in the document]
peternied: @davidosorno Thanks for checking back in.  There is class of scenarios missing from these tests, when the option is valid - but the security configuration doesn't support this state, such as the following example.  This is separate from invalid input which I see covered.

```java
// ...Configure cluster to only support BASIC AUTHC provider...

// Update config
final HttpResponse updateUnavailableSignInOption = rh.executePutRequest(
    "/_plugins/_security/api/tenancy/config",
    "{\"dashboard_signin_options\": [\"SAML\"]}",
    header
);
updateUnavailableSignInOption.assertStatusCode(HttpStatus.SC_BAD_REQUEST);
assertThat(updateUnavailableSignInOption.findValueInJson("error.reason"), equalTo("Validation failure; no SAML authentication provider are available for this cluster"));
```

peternied:
@davidosorno The data storage is being used to save a value for use on the front end - it isn't usable by the backend or other clients.  With the value being specific to the front end, it seems stored it in the .kibana index would be an easier alternative with less coupling, no?

src/test/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiTest.java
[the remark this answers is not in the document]
davidosorno: Oh, I see. Thanks for the guidance. I will work on it.

cwperks:
> The data storage is being used to save a value for use on the front end - it isn't usable by the backend or other clients. With the value being specific to the front end, it seems stored it in the .kibana index would be an easier alternative with less coupling, no?

@peternied The convention for [dynamic dashboards settings](https://github.com/opensearch-project/security/blob/main/config/config.yml#L68-L74) has been to store them in the security index. Putting the sign-in options in the same place is consistent with the other dynamic dashboards settings for the security dashboards plugin.

DarshitChanpura (commented):
+1 to Craig's comments. This should be stored in security index since it is about the ability to update sign-in options and from what i understand .kibana index stores information about dashboard objects and not about security related configuration. Maybe I'm misunderstanding the question here?

src/test/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiTest.java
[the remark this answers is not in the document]
davidosorno: I added a validation to ensure the sign-in option is configured at back end before modifying the sign in options. Thanks for the guidance @peternied and @DarshitChanpura.

src/test/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiTest.java
```diff
@@ -54,9 +57,43 @@ private void verifyTenantUpdate(final Header... header) throws Exception {
...
+
+        assertThat(getDashboardsinfoResponse.findArrayInJson("sign_in_options"), hasItem((DashboardSignInOption.BASIC.toString())));
+        assertThat(getDashboardsinfoResponse.findArrayInJson("sign_in_options"), hasItem((DashboardSignInOption.OPENID.toString())));
+
+        final HttpResponse updateUnavailableSignInOption = rh.executePutRequest(
+            "/_plugins/_security/api/tenancy/config",
+            "{\"sign_in_options\": [\"BASIC\", \"SAML\"]}",
+            header
+        );
+        assertThat(updateUnavailableSignInOption.getStatusCode(), equalTo(HttpStatus.SC_BAD_REQUEST));
```
DarshitChanpura, line 83: this should probly return a 2xx instead of 400 (BAD_REQUEST) since partial update was successful. This return code is mis-leading. This could be covered as a separate PR since this code-change doesn't touch the return code for the associated API.

davidosorno: @DarshitChanpura if it doesn't find the option it throws an exception and it doesn't update the sign in options. I created an assertion after the bad request to make sure the sign in options were not updated. 
I chose to return 400, because if that happens, it means there must be some problem in the interface because that option should not be available for selection.
When the admin goes to the Security Dashboard sign-in options, the options available to select are taken from the authc options available in backend [Getting sign-in options in security dashboard](https://github.com/davidosorno/security-dashboards-plugin/blob/33c1c4381148a16dcc75e2caa1d734dcd94fbdde/public/apps/configuration/panels/auth-view/dashboard-signin-options.tsx#L81)

DarshitChanpura: I see. so no options will be updated even if 1 succeeded and other failed

src/main/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiAction.java
```diff
@@ -202,4 +215,29 @@ private void updateAndValidatesValues(final ConfigV7 config, final JsonNode json
...
+    private boolean isOptionConfiguredAtBackEnd(Authc authc, String option) {
+        for (String key : authc.getDomains().keySet()) {
+            if (key.contains(option.toLowerCase())) {
+                return true;
+            }
+        }
+        throw new IllegalArgumentException(
+            "Validation failure: " + option.toUpperCase() + " authentication provider is not available for this cluster."
+        );
+    }
```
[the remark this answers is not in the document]
davidosorno: I remember I had tried an approach like this before thanks to @cwperks suggestions, but I couldn't make it work.
Thanks @DarshitChanpura this way worked great.

DarshitChanpura (approved):

src/test/java/org/opensearch/security/dlic/rest/api/MultiTenancyConfigApiTest.java
peternied: Returning 400 if any part of the request is invalid seems correct to me- I don't think there should be any changes here.  Good work @davidosorno

peternied (approved):

cwperks (approved):

willyborankin (approved):

davidosorno:
Finally, it was merged.

I want to Thank @cwperks, @peternied, and @DarshitChanpura for your guidance, and helping me to develop this feature. It was really nice working with you all.

PR #3845, opened 2023-12-13
[2.x] Extracts demo configuration setup into a java tool, adds support for Bundled JDK for this tool and updates DEVELOPER_GUIDE.md

DarshitChanpura:
### Description
Backports following commits related to demo configuration tool from main to 2.x:

- [x] [17748b9](https://github.com/opensearch-project/security/commit/17748b9e7edb143ff108d45483e2333f70ab15df) from #3669 
- [x] [4496440](https://github.com/opensearch-project/security/commit/4496440de11e1d1435ff5803942bbdad0c6be513) from #3734 
- [x] [06d8c29](https://github.com/opensearch-project/security/commit/06d8c29eea9a648c463e6063542283222e5b1805) from #3777 
- [x] [e698315](https://github.com/opensearch-project/security/commit/e69831505a13fc79dab14c0743c8a12864cdb94d) from #3807 
- [x] [9d11524](https://github.com/opensearch-project/security/commit/9d11524a2f86fb964671a7496b42b9aa6f9cc5fc) from #3843 
- [x] [62aed21](https://github.com/opensearch-project/security/commit/62aed210be664492783a317f6996185e3bb7487b) from https://github.com/opensearch-project/security/pull/3850
- [x] [ceabe13](https://github.com/opensearch-project/security/commit/ceabe13b94dcd6dfd583d95cd43badeaad39be51) from #3844 

### Issues Resolved
- Related to https://github.com/opensearch-project/security/issues/3827

### Testing
- automated tests

### Check List
- [x] New functionality includes testing
- [x] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3874, opened 2023-12-19
Revert "[2.x] Extracts demo configuration setup into a java tool, adds support for Bundled JDK for this tool and updates DEVELOPER_GUIDE.md"

derek-ho:
Reverts opensearch-project/security#3845

Preparing this revert PR in case we need to unblock folks who get main/2.x CI failures due to this change


peternied:
@derek-ho I don't think we need this and its out of date - I'm going to close this and we can reopen if we need it

PR #3879, opened 2023-12-20
Improve messaging on how to set initial admin password

derek-ho:
### Description
When the demo configuration script fails when no initial admin password is set, it does not provide helpful output telling the user how to set the initial admin password. This PR provides some helpful output indicating what variable the user should modify in order for the setup to succeed.

### Issues Resolved

### Testing
Changed test asserting output 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
+++ b/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
@@ -149,7 +149,9 @@ public class SecuritySettingsConfigurer {
 
             // if ADMIN_PASSWORD is still an empty string, it implies no custom password was provided. We exit the setup.
             if (Strings.isNullOrEmpty(ADMIN_PASSWORD)) {
-                System.out.println("No custom admin password found. Please provide a password via the environment variable OPENSEARCH_INITIAL_ADMIN_PASSWORD.");
+                System.out.println(
+                    "No custom admin password found. Please provide a password via the environment variable OPENSEARCH_INITIAL_ADMIN_PASSWORD."
+                );
                 System.exit(-1);
             }
 
--- a/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
+++ b/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
@@ -150,7 +150,10 @@ public class SecuritySettingsConfigurer {
             // if ADMIN_PASSWORD is still an empty string, it implies no custom password was provided. We exit the setup.
             if (Strings.isNullOrEmpty(ADMIN_PASSWORD)) {
                 System.out.println(
-                    "No custom admin password found. Please provide a password via the environment variable OPENSEARCH_INITIAL_ADMIN_PASSWORD."
+                    String.format(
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -149,7 +149,7 @@ void updateAdminPassword() {
 
             // if ADMIN_PASSWORD is still an empty string, it implies no custom password was provided. We exit the setup.
             if (Strings.isNullOrEmpty(ADMIN_PASSWORD)) {
-                System.out.println("No custom admin password found. Please provide a password.");
+                System.out.println("No custom admin password found. Please provide a password via the environment variable OPENSEARCH_INITIAL_ADMIN_PASSWORD.");
```
cwperks, line 152:
```suggestion
                System.out.println("No custom admin password found. Please provide a password by setting the environment variable OPENSEARCH_INITIAL_ADMIN_PASSWORD. Example: export OPENSEARCH_INITIAL_ADMIN_PASSWORD=<password>");
```

derek-ho: I'm ok with suggesting something like this, but do we need to account for possibility of running on windows/different distributions?

peternied:
@derek-ho It doesn't look like you've updated the PR description, could you please do so?  Here is a good example: https://github.com/opensearch-project/security/pull/3850

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
cwperks: Good point. Leave it generic.

derek-ho:
> @derek-ho It doesn't look like you've updated the PR description, could you please do so?  Here is a good example: https://github.com/opensearch-project/security/pull/3850

@peternied done 👍

DarshitChanpura (approved):

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -149,7 +149,9 @@ void updateAdminPassword() {
 
             // if ADMIN_PASSWORD is still an empty string, it implies no custom password was provided. We exit the setup.
             if (Strings.isNullOrEmpty(ADMIN_PASSWORD)) {
-                System.out.println("No custom admin password found. Please provide a password.");
+                System.out.println(
+                    "No custom admin password found. Please provide a password via the environment variable OPENSEARCH_INITIAL_ADMIN_PASSWORD."
```
peternied, line 153: Please use the `ConfigConstants.OPENSEARCH_INITIAL_ADMIN_PASSWORD` constant instead of hardcoding the value in the string

derek-ho: 👍 good idea, I did so in: [163bf49](https://github.com/opensearch-project/security/pull/3879/commits/163bf494a7ebc623a154ccfb2ce587fab346a7b4)

peternied (approved):

cwperks (approved):

PR #3741, opened 2023-11-19
HttpStatus is only imported from `org.apache.http`

prabhask5:
### Description

Accounts for refactor of HttpStatus classes from hc.core5.http to just http by adding checkstyle test, and manually refactoring all current occurrences of the illegal import. Additionally updated spotlessApply to make the refactor automatic on future cases of illegal import.

* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation)

Bug fix/refactor

* Why these changes are required?

- Refer to #3581 

* What is the old behavior before changes and new behavior after changes?

Ideally nothing.

### Issues Resolved
- Related #3581 

Is this a backport? If so, please add backport PR # and/or commits #
No.

### Testing
Tests have been modified and should work before merging.

### Check List
- [x] New functionality includes testing
- [x] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3894, opened 2023-12-22
Saml http metadata resolver refactor

MaciejMierzwa:
### Description
Refactor SamlHTTPMetadataResolver to remove of HTTP commons 4 lib
* Category (Enhancement, Refactoring, Maintenance)
* Behavior should stay the same

### Issues Resolved
- Resolves https://github.com/opensearch-project/security/issues/2932

### Testing
No functionality changes, tests refactored a little.

### Check List

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/http/saml/HTTPMetadataResolver.java
+++ b/src/main/java/com/amazon/dlic/auth/http/saml/HTTPMetadataResolver.java
@@ -1,0 +1,157 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package com.amazon.dlic.auth.http.saml;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Timer;
+
+import org.apache.hc.client5.http.classic.HttpClient;
+import org.apache.hc.client5.http.classic.methods.HttpGet;
+import org.apache.hc.client5.http.protocol.HttpClientContext;
+import org.apache.hc.core5.http.ClassicHttpResponse;
+import org.apache.hc.core5.http.Header;
+import org.apache.hc.core5.http.HttpException;
+import org.apache.hc.core5.http.HttpHeaders;
+import org.apache.hc.core5.http.io.entity.EntityUtils;
+import org.apache.hc.core5.http.HttpStatus;
+
+import net.shibboleth.utilities.java.support.resolver.ResolverException;
+import org.opensaml.saml.metadata.resolver.impl.AbstractReloadingMetadataResolver;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class HTTPMetadataResolver extends AbstractReloadingMetadataResolver {
+    private final Logger log = LoggerFactory.getLogger(HTTPMetadataResolver.class);
+    private HttpClient httpClient;
+    private URI metadataURI;
+    private String cachedMetadataETag;
+    private String cachedMetadataLastModified;
+
+    public HTTPMetadataResolver(final HttpClient client, final String metadataURL) throws ResolverException {
+        this(null, client, metadataURL);
+    }
+
+    public HTTPMetadataResolver(final Timer backgroundTaskTimer, final HttpClient client, final String metadataURL)
+        throws ResolverException {
+        super(backgroundTaskTimer);
+
+        if (client == null) {
+            throw new ResolverException("HTTP client may not be null");
+        }
+        httpClient = client;
+
+        try {
+            metadataURI = new URI(metadataURL);
+        } catch (final URISyntaxException e) {
+            throw new ResolverException("Illegal URL syntax", e);
+        }
+    }
+
+    public String getMetadataURI() {
+        return metadataURI.toASCIIString();
+    }
+
+    @Override
+    protected void doDestroy() {
+        if (httpClient instanceof AutoCloseable) {
+            try {
+                ((AutoCloseable) httpClient).close();
+            } catch (final Exception e) {
+                log.error("Error closing HTTP client", e);
+            }
+        }
+        httpClient = null;
+        metadataURI = null;
+        cachedMetadataETag = null;
+        cachedMetadataLastModified = null;
+
+        super.doDestroy();
+    }
+
+    @Override
+    protected String getMetadataIdentifier() {
+        return metadataURI.toString();
+    }
+
+    @Override
+    protected byte[] fetchMetadata() throws ResolverException {
+        final HttpGet httpGet = buildHttpGet();
+        final HttpClientContext context = HttpClientContext.create();
+
+        try {
+            log.debug("{} Attempting to fetch metadata document from '{}'", getLogPrefix(), metadataURI);
+            return httpClient.execute(httpGet, context, response -> {
+                final int httpStatusCode = response.getCode();
+                if (httpStatusCode == HttpStatus.SC_NOT_MODIFIED) {
+                    log.debug("{} Metadata document from '{}' has not changed since last retrieval", getLogPrefix(), getMetadataURI());
+                    return null;
+                }
+                if (httpStatusCode != HttpStatus.SC_OK) {
+                    final String errMsg = "Non-ok status code " + httpStatusCode + " returned from remote metadata source " + metadataURI;
+                    log.error("{} " + errMsg, getLogPrefix());
+                    throw new HttpException(errMsg);
+                }
+
+                processConditionalRetrievalHeaders(response);
+                try {
+                    return getMetadataBytesFromResponse(response);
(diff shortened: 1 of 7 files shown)


src/main/java/com/amazon/dlic/auth/http/saml/HTTPMetadataResolver.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    @Override
+    protected byte[] fetchMetadata() throws ResolverException {
+        final HttpGet httpGet = buildHttpGet();
+        final HttpClientContext context = HttpClientContext.create();
+
+        try {
+            log.debug("{} Attempting to fetch metadata document from '{}'", getLogPrefix(), metadataURI);
+            return httpClient.execute(httpGet, context, response1 -> {
```
willyborankin, line 91: `response1` is misleading :-) lets call it differently

src/main/java/com/amazon/dlic/auth/http/saml/HTTPMetadataResolver.java
```diff
@@ -0,0 +1,156 @@
...
+            log.error("{} {}: {}", getLogPrefix(), errMsg, e.getMessage());
+            throw new ResolverException(errMsg, e);
+        }
+    }
+
+    protected HttpGet buildHttpGet() {
+        final HttpGet getMethod = new HttpGet(getMetadataURI());
+
+        if (cachedMetadataETag != null) {
+            getMethod.setHeader("If-None-Match", cachedMetadataETag);
```
willyborankin, line 122: I think better to use per-defined constants for headers. e.g. `org.apache.http.HttpHeaders`

src/main/java/com/amazon/dlic/auth/http/saml/HTTPMetadataResolver.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+        if (cachedMetadataLastModified != null) {
+            getMethod.setHeader("If-Modified-Since", cachedMetadataLastModified);
```
willyborankin, line 125: Same better to use constants

src/main/java/com/amazon/dlic/auth/http/saml/HTTPMetadataResolver.java
```diff
@@ -0,0 +1,156 @@
...
+        return getMethod;
+    }
+
+    protected void processConditionalRetrievalHeaders(final ClassicHttpResponse response) {
+        Header httpHeader = response.getFirstHeader("ETag");
+        if (httpHeader != null) {
+            cachedMetadataETag = httpHeader.getValue();
+        }
+
+        httpHeader = response.getFirstHeader("Last-Modified");
```
willyborankin, line 137: same

willyborankin (commented):
Thank you @MaciejMierzwa.  Added some comments.

src/main/java/com/amazon/dlic/auth/http/saml/HTTPMetadataResolver.java
MaciejMierzwa: Oh yea, it's leftover from refactoring from the deprecated method. Renamed to just "response"

peternied:
@MaciejMierzwa Can this be backported into 2.x - can you take a look if this is possible?  I'm concerned that the Apache Http Client 5 will create some issues in that codebase

cwperks:
@MaciejMierzwa With this PR can these 3 dependencies (https://github.com/opensearch-project/security/blob/main/build.gradle#L579-L581) be changed to `integrationTestImplementation`?

MaciejMierzwa:
@peternied I think it can be backported into 2.x. Currently 2.x doesn't contain dependencies on any of http5 libraries. If the library is allowed I think it's little bit of additional work.

@cwperks [question](https://github.com/opensearch-project/security/pull/3894#issuecomment-1883926286):
I mostly focused on SamlHTTPMetadataResolver class and the surrounding ones. There is still quite a bit of http4 used in implementation, mostly static headers and HTTP codes are imported from http4 library.

peternied:
> @peternied I think it can be backported into 2.x. Currently 2.x doesn't contain dependencies on any of http5 libraries. If the library is allowed I think it's little bit of additional work.

@MaciejMierzwa We can accept these changes in the main branch, we would be in striking distance to remove all http client 4 references.  There might be other issue aside just importing the library [1], but I'd rather move forward with what you have, just want you to be aware the backport might be a little gross.

- https://github.com/opensearch-project/OpenSearch/issues/3351

src/main/java/com/amazon/dlic/util/SettingsBasedSSLConfiguratorV5.java
```diff
@@ -51,7 +46,7 @@
 import static org.opensearch.security.ssl.SecureSSLSettings.SSLSetting.SECURITY_SSL_TRANSPORT_KEYSTORE_PASSWORD;
 import static org.opensearch.security.ssl.SecureSSLSettings.SSLSetting.SECURITY_SSL_TRANSPORT_TRUSTSTORE_PASSWORD;
 
-public class SettingsBasedSSLConfiguratorV4 {
+public class SettingsBasedSSLConfiguratorV5 {
```
peternied, line 49: Lets not change the name as I don't think the version number was associated with the Apache Http Client.  What do you think?

src/main/java/com/amazon/dlic/auth/http/saml/HTTPMetadataResolver.java
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.hc.core5.http.HttpHeaders;
+import org.apache.hc.core5.http.io.entity.EntityUtils;
+import org.apache.hc.core5.http.HttpStatus;
+
+import net.shibboleth.utilities.java.support.resolver.ResolverException;
+import org.opensaml.saml.metadata.resolver.impl.AbstractReloadingMetadataResolver;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class HTTPMetadataResolver extends AbstractReloadingMetadataResolver {
```
peternied, line 32: I see code patterns around the logging that are unique and distinct from the security codebase, was this class's contents copied from another implementation?

peternied (changes requested):

src/main/java/com/amazon/dlic/auth/http/saml/HTTPMetadataResolver.java
MaciejMierzwa: You're right, I followed suggestion no. 1 from https://github.com/opensearch-project/security/issues/2932#issue-1788296478
Switched to LogManager.getLogger() for uniformity although i noticed package org.slf4j.LoggerFactory is used in one more place: CleanerFactory.class

DarshitChanpura:
> Hi I noticed that checkstyle fails for http5 library. Looks like this whole PR could be reversed
https://github.com/opensearch-project/security/pull/3741/files

If http 5 is necessary for this PR to resolve, we may need to revert back to http 5 but that would leave this [bug](https://github.com/opensearch-project/security/issues/3581) unresolved. If the only reason we are not moving to http 5 is due to the saml metadata resolver, we can revert the mentioned PR to move back to http 5 since this PR migrates the resolver to http 5 as well.

peternied:
> Hi I noticed that checkstyle fails for http5 library. Looks like this whole PR could be reversed https://github.com/opensearch-project/security/pull/3741/files

The checkstyle issue is associated with the use of client5 specific status codes, you can use the broader ones
```
-import·org.apache.hc.core5.http.HttpStatus;
+import·org.apache.http.HttpStatus;
```

DarshitChanpura:
> The checkstyle issue is associated with the use of client5 specific status codes, you can use the broader ones

Curious, because the changes introduced here refactor the usage from http 4 to http 5 and I believe they are necessary for this PR. Correct me if I'm wrong here @MaciejMierzwa . If they are indeed required to be http5 what should be the path forward?

DarshitChanpura:
@MaciejMierzwa seems like the progress on this PR has stalled. Do you still intend to work on this? Maintainers will be closing this PR out soon if there is no activity.

MaciejMierzwa:
Hi @DarshitChanpura sorry, I see that I got mail notifications, must've missed it. Regarding comments from https://github.com/opensearch-project/security/pull/3894#issuecomment-1894057689 onward - I think the idea behind this ticket was to move towards removal of http4 library from the codebase. @cwperks also highlighted in this comment https://github.com/opensearch-project/security/issues/3581#issuecomment-1772716656 that http4 usage was due to opensaml library, causing conflicts.
For me, it looks like if this PR gets backported to 2.x the bug mentioned by Peter in https://github.com/opensearch-project/security/issues/3581 get's resolved. What do you think?

peternied:
> I see code patterns around the logging that are unique and distinct from the security codebase, was this class's contents copied from another implementation?

@MaciejMierzwa We do not merge unattributed code copied from other code bases, this violates the [DCO](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin) - developer certificate of origin.  I am closing out this PR.

Issue #1833, opened 2022-05-10
Question: what to do with the stale dependencies in THIRD_PARTY.txt


duhang:
**What is the bug?**
A clear and concise description of the bug.

The THIRD_PARTY.txt in this project hasn't been updated since 2019.

We are facing challenges to attribute properly to those 69 stale 3rd-party dependencies listed in THIRD_PARTY.txt.

My questions are:

1. Currently, is there any action/plan to update this THIRD_PARTY.txt and more importantly, to keep it updated as code base evolves?

2. If the answer to the above question is No, shall we ignore this THIRD_PARTY.txt for now and attribute all current 3rd-party dependencies gathered from pom.xml?

Many many thanks!
 
**How can one reproduce the bug?**

Go to https://github.com/opensearch-project/security/blob/main/THIRD-PARTY.txt

**What is the expected behavior?**

Keep THIRD-PARTY.txt up-to-date 

**What is your host/environment?**
N/A

**Do you have any screenshots?**
N/A

**Do you have any additional context?**
Add any other context about the problem.

peternied:
Thanks for opening @duhang, we do not have a timeline to update the THIRD_PARTY.txt.  We would be willing to accept pull requests to update it if you were so inclined. I think you'd want to pull from the [build.gradle](https://github.com/opensearch-project/security/blob/main/build.gradle#L87) dependencies if you were looking to recreate this list.

peternied:
@opensearch-project/plugins What do you think about the suggestion? I am in favor of option 1.

saratvemulapalli:
> @opensearch-project/plugins What do you think about the suggestion? I am in favor of option 1.

We do have repositories which maintain Third Party notices([1] OpenSearch-SDK). 
Im not a legal expert but Option 1 seems like a cleaner approach if we're not ready to maintain it. 
@hyandell @dblock  any thoughts here?

[1] https://github.com/opensearch-project/opensearch-sdk/blob/main/THIRD-PARTY-NOTICES

scrawfor99:
[Triage] @scrawfor99 will you follow up on whether we can remove this.

PR #3907, opened 2024-01-02
Remove Third-party.txt file

scrawfor99:
### Description
The Security plugin is one of only two plugins (alongside the SQL plugin) which has its own THIRD-PARTY.txt file. This file is not automatically generated and serves no obvious use. This changes follows the suggestions on the linked issue (below) and removes the file.

I requested a legal review for this change just in case and will update this PR with a comment once I have confirmation it is fine to merge. 

### Issues Resolved
- resolves https://github.com/opensearch-project/security/issues/1833

### Check List


derek-ho (approved):
LGTM

peternied (approved):

cwperks (approved):

DarshitChanpura (changes requested):
> I requested a legal review for this change just in case and will update this PR with a comment once I have confirmation it is fine to merge.

Waiting for this to be addressed so it doesn't get merged accidentally. Feel free to dismiss this review once the comment has been addressed.

cwperks:
@scrawfor99 wdyt about enabling the [dependencyLicenses check](https://github.com/opensearch-project/security/blob/main/build.gradle#L160)?

scrawfor99:
Hi @cwperks, do you mean in general or instead of removing the 3rd party audit file? 

This is the search for the existing use of dependencyLicenses across the project. To be honest, I am not positive how it is being used everwhere but it does not look to me that it is actually related to the 3rd party audit file. For instance opensearch-hadoop uses it (https://github.com/opensearch-project/opensearch-hadoop/blob/73b0745d2c247139d4073146cecb3503196a2c08/buildSrc/src/main/java/org/opensearch/hadoop/gradle/buildtools/UpdateShasTask.java#L59) but they do not have the third-party file. 

Anyways, I think more checks is better than fewer but am not sure if that is related to the third-party file.

cwperks:
iirc `dependencyLicenses` enforces that whenever a repo adds dependencies, that they also keep track of the LICENSE and NOTICE for the dependency. With the check enabled, the security plugin would need to keep track of a [licenses/](https://github.com/opensearch-project/OpenSearch/tree/main/server/licenses) directory that collects the licenses of the dependencies. I believe it can be used in place of Third-party.txt and is the standard for how OpenSearch tracks dependency licenses.

peternied:
@cwperks Can you create an issue for enabling that task?

scrawfor99:
Update 1/12/2024:

After legal consult I was told: 

"The goal of a THIRD-PARTY.txt file like that is to provide appropriate attribution of third-party software packages from which portions, in source code or binary form, have been copied or otherwise integrated into the project. This is a compliance requirement of most Open Source software licenses.
 
Importantly, any third-party dependencies that are declared as required to build, test or run the project, but which have not been copied into the project and should instead be supplied by users, do not need to be listed in a THIRD-PARTY.txt file — except maybe as a courtesy to project users who want a quick summary of project dependencies."

This would mean we have to keep the file but I did point out:

"The THIRD-PARTY.txt (https://github.com/opensearch-project/security/blob/main/THIRD-PARTY.txt) file does contain dependencies incorporated in the Security plugin. These are the same Version 2 Apache Licensed dependencies used across OpenSearch as a whole and come from various outside projects such as Apache or Hamcrest. 

The fact that we import their libraries does not seem to me like it would require us to keep the file however. Looking at the central OpenSearch, here is an example of where Hamcrest is used (https://github.com/opensearch-project/OpenSearch/blob/a90b6b64d5fbde8c13b6acca076584e8060c599a/server/src/internalClusterTest/java/org/opensearch/action/RejectionActionIT.java#L51). There is no THIRD-PARTY.txt file however. 

OpenSearch does make use of license files which I imagine cover the same type of documentation as a THIRD-PARTY listing file would. However, other plugins such as the KNN plugin (https://github.com/opensearch-project/k-NN/blob/45282e056a9eb875aa53b08f7ee5865b277e651b/src/test/java/org/opensearch/knn/KNNTestCase.java#L27C2-L27C2) has neither a THIRD-PARTY.txt file or license files despite also using third party libraries. "

I am of the opinion that we should be safe to remove the file but I am also in no way qualified to know what is/is not okay. 

I wanted to share this update while I wait for further input from a legal perspective.

peternied:
@scrawfor99 @DarshitChanpura Is there something we are waiting on to merge this change?

DarshitChanpura:
> I wanted to share this update while I wait for further input from a legal perspective.

From what I understand we are waiting for more inputs from legal. I'm okay with removing third party text files if we are sure that any portion of code has not been copied into, and legal gives green light.

scrawfor99:
Hi @peternied and @DarshitChanpura, thanks for checking in. 

So with the last update, I had pushed back that we needed to keep the file. I did not get a response though which is why I did not confirm/deny. I am of the opinion it should be fine but I did not get any further input which is why this is still here...

DarshitChanpura:
@scrawfor99 Feel free to take a judgement call here and merge it.

peternied:
@scrawfor99 I don't think this needs to be backported, up to you.

PR #3908, opened 2024-01-02
[BUG-2556] Add new DLS filtering test

### LINKED ISSUE #2556: [BUG] Can't use multiple roles with DLS permissions on the same Index

prabhask5:
### Description
This PR adds a new DLS filtering test that checks for the union of different dls settings related to different roles attached to the same user.

Enhancement

These tests were previously not included, which makes testing bugs including the one in #2556 difficult. Adding this test makes it easy to recreate these bugs if they appear in the future. Additionally the data in these test cases matches the examples provided in the bug case so these tests were also used to determine if the bug in the ticket is actually happening. Since these two tests pass, I personally don't think this bug is reproducible.

A new test should increase code coverage.

### Issues Resolved
#2556 

No

### Testing
Added new test

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/DlsIntegrationTests.java
+++ b/src/integrationTest/java/org/opensearch/security/DlsIntegrationTests.java
@@ -167,52 +167,52 @@ public class DlsIntegrationTests {
     /**
      * Test role 1 for DLS filtering with two (non)overlapping roles. This role imposes a filter where the user can only access documents where the sensitive field is false. This role is applied at a higher level for all index patterns.
      */
-    static final TestSecurityConfig.Role TEST_ROLE_ONE = new TestSecurityConfig.Role("test_role_1").clusterPermissions(
+    static final TestSecurityConfig.Role ROLE_NON_SENSITIVE_ONLY = new TestSecurityConfig.Role("test_role_1").clusterPermissions(
         "cluster_composite_ops_ro"
     ).indexPermissions("read").dls("{\"match\":{\"sensitive\":false}}").on("*");
 
     /**
      * Test role 2 for DLS filtering with two overlapping roles. This role does not impose any filter, and combined with TEST_ROLE_ONE should yield a union that does not impose any filter. This role is applied at a lower level for index patterns my_index*.
      */
-    static final TestSecurityConfig.Role TEST_ROLE_TWO = new TestSecurityConfig.Role("test_role_2").clusterPermissions(
+    static final TestSecurityConfig.Role ROLE_ALLOW_ALL = new TestSecurityConfig.Role("test_role_2").clusterPermissions(
         "cluster_composite_ops_ro"
     ).indexPermissions("read").dls("{\"match_all\": {}}").on("my_index*");
 
     /**
-     * Test role 3 for DLS filtering with two nonoverlapping roles. This role imposes a filter wher ethe user can only access documents where the genre field is History, and combined with TEST_ROLE_ONE should yield a union that allows the user to access every document except the one with genre Science and sensitive true. This role is applied at a lower level for index patterns my_index*.
+     * Test role 3 for DLS filtering with two nonoverlapping roles. This role imposes a filter where the user can only access documents where the genre field is History, and combined with TEST_ROLE_ONE should yield a union that allows the user to access every document except the one with genre Science and sensitive true. This role is applied at a lower level for index patterns my_index*.
      */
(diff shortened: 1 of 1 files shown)


prabhask5:
@peternied @scrawfor99 @cwperks Ready for review!

src/integrationTest/java/org/opensearch/security/DlsIntegrationTests.java
```diff
@@ -158,6 +164,56 @@ public class DlsIntegrationTests {
                 .on("*")
         );
 
+    /**
+     * Test role 1 for DLS filtering with two (non)overlapping roles. This role imposes a filter where the user can only access documents where the sensitive field is false. This role is applied at a higher level for all index patterns.
+     */
+    static final TestSecurityConfig.Role TEST_ROLE_ONE = new TestSecurityConfig.Role("test_role_1").clusterPermissions(
```
scrawfor99, line 170: Nit: What do you think about renaming the roles so that they say what they do similar to below with the music tests? I.e ROLE_NON_SENSITIVE_ONLY

scrawfor99 (commented):
Really awesome work here. Just one comment about potential naming and then a typo I caught. Thanks for all your work

src/integrationTest/java/org/opensearch/security/DlsIntegrationTests.java
```diff
@@ -517,4 +597,110 @@ public void testAggregateAndComputeStarRatings() throws IOException {
...
+                )
+            );
+        }
+
+        try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(TEST_UNION_OF_NONOVERLAPPING_ROLES_USER)) {
+            SearchRequest searchRequest = new SearchRequest(UNION_TEST_INDEX_NAME);
+            SearchResponse searchResponse = restHighLevelClient.search(searchRequest, DEFAULT);
+
+            assertThat(searchResponse, isSuccessfulSearchResponse());
+            assertThat(searchResponse, numberOfTotalHitsIsEqualTo(9));
```
DarshitChanpura, line 689: This code can be cleaned up to allow re-usability.

```suggestion
            assertSearchResponseHitsEqualTo(9);
```

and a function:
```
private void assertSearchResponseHitsEqualTo(int hits) {
        SearchRequest searchRequest = new SearchRequest(UNION_TEST_INDEX_NAME);
        SearchResponse searchResponse = restHighLevelClient.search(searchRequest, DEFAULT);

        assertThat(searchResponse, isSuccessfulSearchResponse());
        assertThat(searchResponse, numberOfTotalHitsIsEqualTo(hits));
}
```

prabhask5: searchResponse is used in the two checks below this so I'll keep them in the main function but I'll make that helper function to be used in future tests

prabhask5:
@scrawfor99 @DarshitChanpura Ready for review! I think I added all the requested changes.

scrawfor99 (approved):
Looks great to me! Thanks a bunch @prabhask5

DarshitChanpura (approved):
LGTM!

PR #3909, opened 2024-01-02
Validate 409s occur when multiple config updates happen simultaneously

prabhask5:
### Description
Added new TenantAsyncActionTest.java file to put api unit tests for the /api/tenants api call- additionally wrote new unit test to test for parallel PUT /tenant/{name} calls, that they successfully return a 409 call and retry the call.

Test fix

There was no test that checked for the behavior mentioned in the linked ticket, so previously we didn't know if the issue still existed or not (it apparently was fixed before), to make sure this doesn't happen again for this particular case, I decided to write a concrete test for it.

Test was added, there's no behavior change?

### Issues Resolved
- #1402 - added unit test to cover bug in this issue

### Testing
Added new tenant action unit test to test parallel tenant api calls.

### Check List


src/integrationTest/java/org/opensearch/security/api/TenantAsyncActionTest.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
+import static org.opensearch.security.OpenSearchSecurityPlugin.PLUGINS_PREFIX;
+import static org.opensearch.security.support.ConfigConstants.SECURITY_RESTAPI_ROLES_ENABLED;
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+import static org.opensearch.test.framework.TestSecurityConfig.Role.ALL_ACCESS;
+
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class TenantAsyncActionTest {
```
peternied, line 43: Please rename to TenantTest so it can be reused for any further tests around tenancy

src/integrationTest/java/org/opensearch/security/api/TenantAsyncActionTest.java
```diff
@@ -0,0 +1,82 @@
...
+    public void testParallelPutRequests() throws Exception {
+        try (TestRestClient client = cluster.getRestClient(USER_ADMIN)) {
+
+            final CountDownLatch countDownLatch = new CountDownLatch(1);
+            final List<CompletableFuture<TestRestClient.HttpResponse>> conflictingRequests = AsyncActions.generate(() -> {
+                return client.putJson(TENANT_ENDPOINT, TENANT_BODY);
+            }, 4, 4);
+
+            // Make sure all requests start at the same time
+            countDownLatch.countDown();
```
peternied, line 69: This isn't being used to pause execution - runnables at 65 need to wait on the countDownLatch

src/integrationTest/java/org/opensearch/security/api/TenantAsyncActionTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @ClassRule
+    public static LocalCluster cluster = new LocalCluster.Builder().clusterManager(ClusterManager.THREE_CLUSTER_MANAGERS)
+        .authc(AUTHC_HTTPBASIC_INTERNAL)
+        .users(USER_ADMIN)
+        .anonymousAuth(false)
+        .nodeSettings(Map.of(SECURITY_RESTAPI_ROLES_ENABLED, List.of("user_" + USER_ADMIN.getName() + "__" + ALL_ACCESS.getName())))
+        .build();
+
+    @Test
+    public void testParallelPutRequests() throws Exception {
```
peternied, line 60: This test should be moved into SecurityConfigurationTests, it doesn't need unique setup

src/integrationTest/java/org/opensearch/security/api/TenantAsyncActionTest.java
```diff
@@ -0,0 +1,82 @@
...
+            AtomicInteger numCreatedResponses = new AtomicInteger();
+            AsyncActions.getAll(conflictingRequests, 1, TimeUnit.SECONDS).forEach((response) -> {
+                assertThat(response.getStatusCode(), anyOf(equalTo(HttpStatus.SC_CREATED), equalTo(HttpStatus.SC_CONFLICT)));
+                if (response.getStatusCode() == HttpStatus.SC_CREATED) numCreatedResponses.getAndIncrement();
+            });
+            assertThat(numCreatedResponses.get(), equalTo(1)); // should only be one 201
+
+            TestRestClient.HttpResponse getResponse = client.get(TENANT_ENDPOINT); // make sure the one 201 works
+            assertThat(getResponse.getBody(), containsString("create new tenant"));
+        }
```
peternied, line 80: Make another call to the update tenants config here to confirm that the 409 is not longer an issue.

peternied (changes requested):

src/integrationTest/java/org/opensearch/security/api/TenantAsyncActionTest.java
prabhask5: just to clarify, should these tests not be in their own file? I made a new file since these tests seemed different than the ones in the files that already existed

src/integrationTest/java/org/opensearch/security/api/TenantAsyncActionTest.java
peternied: Disregard in favor of https://github.com/opensearch-project/security/pull/3909#discussion_r1450814138

src/integrationTest/java/org/opensearch/security/api/TenantAsyncActionTest.java
peternied: The system that is throwing the 409 doesn't apply to only tenants, it applies to any security configuration.  When writing tests for features group them together makes it easier to see what has coverage - yup move this into that file.

peternied (approved):

cwperks (approved):
LGTM!

PR #3910, opened 2024-01-02
[Enhancement-3139] Make FLS, DLS and Field Masking Unit Tests more extensive by testing more document retrieval APIs

### LINKED ISSUE #3139: Make FLS, DLS and Field Masking Unit Tests more extensive by testing more document retrieval APIs
There was an [issue filed recently](https://github.com/opensearch-project/security/issues/3088) where a user in the public slack mentioned that document retrieval was broken for a user that was mapped to a role with a DLS terms query. This bug revealed a gap in testing where our current [DlsTest Suite](https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/dlic/dlsfls/DlsTest.java) did not cover this scenario so it was not caught during CI. 

I'm filing an issue to track what the current Test Suite covers and identify opportunities to make the FLS, DLS and Field Masking tests more extensive. 

Ideally, all permutations of DLS queries and document retrieval actions have a test.

[ ] Identify all OpenSearch actions that retrieve documents from an index or set of indices
[ ] Write a test for the action with a user that has a role with DLS restrictions containing a `bool` query
[ ] Write a test for the action with a user that has a role with DLS restrictions containing a `terms` query
[ ] Write a test for the action with a user that has a role with FLS restrictions
[ ] Write a test for the action with a user that has a role with Field Masking restrictions

prabhask5:
### Description
Add new cases to DlsIntegrationTests.java, and FlsAndFieldMaskingTests.java to account for all permutations of cases.

Enhancement

Previously these tests weren't comprehensive, therefore leading to uncatched bugs in deployment.

Tests are more comprehensive.

### Issues Resolved
#3139 

No

### Testing
Added new test cases.

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/DlsIntegrationTests.java
+++ b/src/integrationTest/java/org/opensearch/security/DlsIntegrationTests.java
@@ -19,6 +19,7 @@ import java.util.stream.Collectors;
 
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.cxf.common.i18n.Exception;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


prabhask5:
@peternied @scrawfor99 @cwperks Ready for review!

src/integrationTest/java/org/opensearch/security/FlsAndFieldMaskingTests.java
```diff
@@ -206,6 +209,36 @@ public class FlsAndFieldMaskingTests {
...
+        new TestSecurityConfig.Role("example_inclusive_fls").clusterPermissions("cluster_composite_ops_ro")
+            .indexPermissions("read")
+            .fls(FIELD_TITLE)
+            .on(FIRST_INDEX_NAME)
+    );
+
+    /**
+     * Example user with fls filter in which the user can see every field but the {@link Song#FIELD_TITLE} field.
+     */
+    static final TestSecurityConfig.User EXCLUSIVE_FLS_USER = new TestSecurityConfig.User("exclusive_fls_user").roles(
```
scrawfor99, line 225: Nit: Similar to the other PR maybe we can swap the names so they illustrate what the roles/users do? For example USER_NO_FIELD_Title or something similar.

src/integrationTest/java/org/opensearch/security/FlsAndFieldMaskingTests.java
```diff
@@ -811,4 +847,287 @@ public void getFieldCapabilities() throws IOException {
         }
     }
 
+    @Test
+    public void testGetDocumentWithFLSRestrictions() throws Exception {
```
scrawfor99, line 851: Nit: Another naming thing; Can we make this illustrate what the test should do? For example testGetDocumentFLSNoTitles or something similar?

scrawfor99 (commented):
What do you think about adding a couple tests to try out the combinations of the new roles you added? 

Otherwise looks really good, just a couple naming things so that in shorthand logs it is easier to tell what is failing.

src/integrationTest/java/org/opensearch/security/DlsIntegrationTests.java
```diff
@@ -517,4 +552,92 @@ public void testAggregateAndComputeStarRatings() throws IOException {
...
+        }
+
+        try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(TERM_USER)) {
+            GetResponse termFoundGetResponse = restHighLevelClient.get(findExistingDoc, DEFAULT);
+            assertThat(termFoundGetResponse, containDocument(FIRST_INDEX_NAME, FIRST_INDEX_ID_SONG_1));
+
+            GetResponse termNotFoundGetResponse = restHighLevelClient.get(findNonExistingDoc, DEFAULT);
+            assertThat(termNotFoundGetResponse.isExists(), equalTo(false));
+        }
+    }
```
DarshitChanpura, line 576:
```suggestion
public void testGetDocumentWithDLSRestrictions() throws Exception {
        GetRequest findExistingDoc = new GetRequest(FIRST_INDEX_NAME, FIRST_INDEX_ID_SONG_1);
        GetRequest findNonExistingDoc = new GetRequest(FIRST_INDEX_NAME, "RANDOM_INDEX");

        try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(BOOL_USER)) {
            GetResponse boolFoundGetResponse = restHighLevelClient.get(findExistingDoc, DEFAULT);
            assertThat(boolFoundGetResponse, containDocument(FIRST_INDEX_NAME, FIRST_INDEX_ID_SONG_1));

            GetResponse boolNotFoundGetResponse = restHighLevelClient.get(findNonExistingDoc, DEFAULT);
            assertThat(boolNotFoundGetResponse.isExists(), equalTo(false));
        }

        try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(TERM_USER)) {
            
        }
    }
    
    private void assertNotFoundGetResponse() {
            GetResponse response = restHighLevelClient.get(findExistingDoc, DEFAULT);
            assertThat(response, containDocument(FIRST_INDEX_NAME, FIRST_INDEX_ID_SONG_1));

            response = restHighLevelClient.get(findNonExistingDoc, DEFAULT);
            assertThat(response.isExists(), equalTo(false));
    }
```

This suggestion can be applied to all tests below for better readability and cleaner code.

scrawfor99:
Hi @prabhask5, thanks for making those changes above. Looking at some of the definitions you have, it seems like the issue is because of the contradiction in some of the roles and the ordering of how the rules are applied. I can see where you tried to combine a role which can only see "titles" with a role which cannot see "titles." I think you thought that this would result in having a user which could not see any "titles" since that is more restrictive then the other role. I can definitely see where you are coming from...

Unfortunately, things don't quite work like that. When DLS/FLS is computed, it is done so in an additive way. Here are some related discussions around DLS/FLS that may be of use for moving forward: https://github.com/opensearch-project/security/issues/2556. Basically, we end up taking a union of the roles and this can lead to weird behavior when the ordering of the rules is swapped. In this case, I suspect the read title option is being applied after the read none option meaning the final result is that the user can get the titles. I would have to step through your tests to be sure but from what I remember this is my best guess. You can try reordering the rules or redefining them to be more consistent.

DarshitChanpura:
@prabhask5 seems like the progress on this has been stalled. Would you like any help from maintainers to bring this across finish line?

peternied:
This PR has stalled, and based on the initial code coverage numbers there is no large jump I am closing out this PR.  @prabhask5 Please let us know if you'd like this reopened thanks.

PR #3920, opened 2024-01-04
Add additional sendRequestDecorate cases

scrawfor99:
### Description

This change is intended to increase the code coverage of the SecurityInterceptor class. Specifically, it adds two new test cases to the `testSendRequestDecorate` scenario. The first case added is an instance where we check that after previously sending a local request, we can still issue a second local request without making use of the serialization. This case was covered in part previously but now we check that following the original localnode request we still do not run into issues with a second node and connection.

The other case added creates a completely different transport address and confirms the serialization is consistent when the remote request is across a different transport address. 

The original issue mentioned needing coverage for remote to local so I am hoping this is what was missing (it is hard to tell from the code coverage report). 

### Issues Resolved
Should resolve the second point of #3137

### Check List


peternied:
@scrawfor99 Codecov report does not show addition coverage was added in SecurityIntercepter, did you see increased code coverage when looking at `jacocoTestReport` from gradle locally?

scrawfor99:
Hi @peternied, hmm I am not sure what the deal is but let me try to fix it. I'll add the coverage report too.

DarshitChanpura (commented):
Thank you @scrawfor99 for adding this test. Can we split it into smaller tests or helper methods for easier readability?

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -247,7 +249,7 @@ && getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROL
                 useJDKSerialization
             );
 
-            if (isActionTraceEnabled()) {
+            if (isActionTraceEnabled() || actionTraceForTesting) {
```
peternied, line 252: Lets change how `isActionTraceEnabled` is declared so we can inject it - so its controllable in test code.  I'd recommend using a 'Supplier<Boolean>' in class construction.  This allows the removal of `actionTraceForTesting`, what do you think?

peternied (changes requested):
Lets use dependency injection in the production code to keep it cleaner.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
scrawfor99: Sure I will swap it over.

scrawfor99:
Hi @DarshitChanpura and @peternied thanks for your feedback. @DarshitChanpura, I tried to split the tests a bit so that there is a more standard structure and you can tell what it is going on based on the name. I also moved as much of the code into the setup as I could so we can reuse it.

@peternied, I swapped over to a supplier and just made the test return true. Let me know if that is not what you meant.

Otherwise should be all set. Thank you both!

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -128,7 +132,8 @@ public void setup() {
             clusterService,
             sslExceptionHandler,
             clusterInfoHolder,
-            sslConfig
+            sslConfig,
+            this::testActionTraceSupplier
```
peternied, line 136: Super tiny nit, you can inline this function if you aren't testing the alternative case
```suggestion
            () -> true
```

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -201,17 +292,306 @@ public <T extends TransportResponse> void sendRequest(
...
+                Connection connection,
+                String action,
+                TransportRequest request,
+                TransportRequestOptions options,
+                TransportResponseHandler<T> handler
+            ) {
+                String serializedUserHeader = threadPool.getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER);
+                assertEquals(serializedUserHeader, Base64Helper.serializeObject(user, useJDKSerialization));
+            }
+        };
```
peternied, line 346: Lets make this more readable by creating a method that generates AsyncSender that takes in a function that could look like:

```java
final AsyncSender verifyUserInThreadContext = createAsyncSender((user) -> {
   final String serializedUserHeader = threadPool.getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER);
   assertEquals(serializedUserHeader, Base64Helper.serializeObject(user, useJDKSerialization));
}); 
```

Also it might be worthwhile to make a helper function for what seems like an often repeated pattern `threadPool.getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_USER_HEADER)` -> getUserInThreadContext() to condense the tests

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -201,17 +292,306 @@ public <T extends TransportResponse> void sendRequest(
...
+
+        // this is a request where the local node is null; have to use the remote connection since the serialization will fail
+        securityInterceptor.sendRequestDecorate(sender, connection2, action, request, options, handler, null);
```
peternied, line 349: I'm not sure we know sender has executed, lets make sure this is confirmed by tying a completable future or countdownlatch to wait on before proceeding.

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -201,17 +292,306 @@ public <T extends TransportResponse> void sendRequest(
...
+                TransportRequestOptions options,
+                TransportResponseHandler<T> handler
+            ) {
+                User transientUser = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+                assertEquals(transientUser, user);
+            }
+        };
+
+        // Having null connection will cause useJDKSerialization to throw npe
+        AsyncSender finalSender = sender; // Have to do this because of being a runnable
```
peternied, line 374: Seems like you are creating a sender, then calling sendRequestDecorate(...) many times in this same method, seems like this single test could be refactored into multiple tests where it is clearer what is being verified so there is a cleaner pattern setup/act/assert.

What do you think?

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -201,17 +292,306 @@ public <T extends TransportResponse> void sendRequest(
...
+     * method in the Security Interceptor.
      */
     @Test
-    public void testSendRequestDecorateWhenRemoteNodeUsesJDKSerde() {
-        testSendRequestDecorate(Version.V_2_0_0);
-    }
+    public void testSendRequestDecorateWithFakeHeaders() {
+        Version remoteNodeVersion = Version.CURRENT;
+        // Add a fake header to show it does not get misinterpreted
+        threadPool.getThreadContext().putHeader("FAKE_HEADER", "fake_value");
```
peternied, line 494: For this test it seems like it would be fair easier to mock the thread pool and thread context, what do you think?

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
peternied: Note: this is the only comment that I'd feel is a barrier to merging this change

DarshitChanpura (commented):
Thank you for breaking the tests into smaller chunks. There is an opportunity to make it slightly more readable. I see this pattern repeated and this could be extracted into its own method:
```
 // This is a different way to get the same result which exercises the origin0 = null logic of ensureCorrectHeaders
        securityInterceptor.sendRequestDecorate(finalSender, connection1, action, request, options, handler, localNode);

        User transientUser2 = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
        assertEquals(transientUser2, user);

        // Make the remote address null should cause the ensureCorrectHeaders to keep the TransportAddress as null ultimately causing local
        // logic to occur
        threadPool.getThreadContext().putHeader(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS, null);

        sender = new AsyncSender() {
            @Override
            public <T extends TransportResponse> void sendRequest(
                Connection connection,
                String action,
                TransportRequest request,
                TransportRequestOptions options,
                TransportResponseHandler<T> handler
            ) {
                User transientUser = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
                assertEquals(transientUser, user);
            }
        };
```

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
scrawfor99: Yeah sounds good; I was just keeping the original formatting as much as possible but I can swap it.

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
scrawfor99: Sure will do. I think that the sender validation should theoretically check it but I will add this in case I am wrong.

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
scrawfor99: Yeah good idea.

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
scrawfor99: I am not sure if that will trigger the coverage in the same way. In a couple places the interceptor checks against specific headers in a big conditional. In order to get full coverage it wants the negative and positive for each of these which is why I was adding each one like this. 

I can try swapping over and keep the changes if it seems like that will result in the same coverage.

DarshitChanpura (dismissed):
LGTM. Thanks @scrawfor99 !

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -236,47 +237,54 @@ final void verifyOriginalContext(User user) {
...
+        TransportRequest request,
+        TransportRequestOptions options,
+        TransportResponseHandler handler,
+        DiscoveryNode localNode
+    ) {
+        CompletableFuture<TransportResponse> future = new CompletableFuture<>();
+        securityInterceptor.sendRequestDecorate(sender, connection, action, request, options, handler, localNode);
+        future.thenRun(() -> { verifyOriginalContext(user); });
+        future.complete(null);
+        return future;
```
peternied, line 254: Pretty sure this code doesn't work, the future needs to be completed by the asynchronous action in the 'sender', here is code that should handle the scenario for the happy path and report any exceptions raised that are catchable by the caller.  

```suggestion
        final var sendRequestFuture = new CompletableFuture<>();
        final var sender = new AsyncSender() {
            @Override
            public <T extends TransportResponse> void sendRequest(
                Connection connection,
                String action,
                TransportRequest request,
                TransportRequestOptions options,
                TransportResponseHandler<T> handler
            ) {
                // This is a Callable passed to this function - or any type that can execute and throw exceptions
                try {
                   verificationAction.call();
                   sendRequestFuture.complete(null);
                } catch (final Exception e) {
                   sendRequestFuture.completeExceptionally(e);
                }
            }
        };
        securityInterceptor.sendRequestDecorate(sender, connection, action, request, options, handler, localNode);
        sendRequestFuture.join()
```

scrawfor99: I am not sure I follow. ThenRun should not call verifyOriginalContext until the the future is completed. This is the documentation for thenRun: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#thenRun-java.lang.Runnable-

Why do you think this will not work?

scrawfor99: I am going to preemptively swap over to the form you suggested but still curious^

peternied: >> Can you tell me if this code will / wont work as written?
>
> This Java code snippet involves using `CompletableFuture` and an asynchronous request. Let's analyze it for potential issues:
> 
> 1. **Creation of `CompletableFuture`**: The `CompletableFuture` is correctly instantiated. This part should work as expected.
> 
> 2. **Asynchronous Request Call**: The `securityInterceptor.sendRequestDecorate` method is called. This seems to be an asynchronous operation. However, it's not clear from this snippet whether this method is supposed to complete the `future` or not. If it's supposed to complete the `future`, it needs a reference to the `future` object to complete it, which is not passed in the snippet.
> 
> 3. **Chaining with `thenRun`**: The `thenRun` method is used to add an action (`verifyOriginalContext(user)`) that is triggered upon the future's completion. This is syntactically correct. However, it's important to note that `thenRun` runs in the same thread that completes the future, unless you explicitly use `thenRunAsync`.
> 
> 4. **Completing the Future**: The `future.complete(null)` immediately completes the future with a `null` value. This is a potential issue because it's called right after the asynchronous operation is initiated. If the intention is for `verifyOriginalContext(user)` to run after `sendRequestDecorate` completes its process, this immediate completion contradicts that. It effectively renders the asynchronous operation irrelevant, as the future is completed before the operation can potentially affect its outcome.
> 
> 5. **Return Statement**: The method correctly returns the `CompletableFuture`.
> 
> From this analysis, the main issue is with the premature completion of the future (`future.complete(null)`). This call should ideally be inside the callback of the asynchronous operation (`sendRequestDecorate`), assuming that method should dictate when the future is completed. If `sendRequestDecorate` does not interact with the `future`, then `future.complete(null)` might be correct, but in that case, the use of `CompletableFuture` might not be serving its intended purpose of managing asynchronous operations. 
> 
> In summary, the code will work syntactically, but it may not work as intended due to the premature completion of the `CompletableFuture`. The behavior depends on the internal workings of `sendRequestDecorate` and how it's supposed to interact with the future.

Sorry - stuck between meetings, I asked ChatGPT and I agree with its analysis

DarshitChanpura:
@scrawfor99 @peternied Anything blocking from this being merged?

peternied:
Thanks for the prompt - I'll take another look

scrawfor99:
@DarshitChanpura @peternied I am blocking this. I have not gotten a chance to swap this over to the fixed completable formatting.

peternied (approved):

DarshitChanpura (approved):

Issue #2402, opened 2023-01-13
[BUG] install_demo_configuration.[sh|bat] assumeyes behavior not working as expected


cwperks:
**What is the bug?**

The `install_demo_configuration` scripts are not actually setting all prompts to `y` when using the `-y` option as you would expect. This issue was initially uncovered in this PR: https://github.com/opensearch-project/security/pull/2308

In addition to the behavior of `-y`, there is no command line option available to only install the demo certificates and leave `initsecurity` and `cluster_mode` as false. 

The install_demo_configuration scripts should be updated with a `-y` option that actually accepts all prompts and a new option should be added to only install the demo certificates.

scrawfor99:
It looks like opensearch build makes use of the -y command in [opensearch-docker-entrypoint.sh](https://github.com/opensearch-project/opensearch-build/blob/56cf1cb017ae38ed7b1916ff328e94c3ab40cfee/docker/release/config/opensearch/opensearch-docker-entrypoint.sh#L38) and also [opensearch-tar-install](https://github.com/opensearch-project/opensearch-build/blob/main/scripts/startup/tar/linux/opensearch-tar-install.sh#L12).

peternied:
Bugs like this are a good reason to migrate away from our current command line scripts. 

- https://github.com/opensearch-project/security/issues/1755

scrawfor99:
[Triage] @scrawfor99 to follow up with build team on the use of `-y` & implement changes accordingly.

davidlago:
@cwperks updates on this issue?

cwperks:
`assumeyes` is being handled in the updated installer: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/tools/democonfig/Installer.java#L198-L218

Closing this issue.

DarshitChanpura:
This bug fix touches a number of different repositories as they already use -i with -y option.

## Proposed Fix:
- Introduce a new option to generate demo certs : `-d`
- Modify the tool to assume `-d`, `-i`, `-c` as yes when `-y` is passed. Else individual options need to be passed
- Modify different distributions and integ tests across different repositories.

PR #3935, opened 2024-01-10
Fix Bug with Install demo configuration running in cluster mode with -y

derek-ho:
### Description
Previous behavior prior to 2.11 would be that init security and cluster mode would only be set if they were explicitly passed in. There is a small bug that is setting them to be true if -y is passed in. This is causing a few failures with duplicate keys in integration tests when -y is passed in. 
https://github.com/opensearch-project/security/blob/2.11/tools/install_demo_configuration.sh#L73-L98

### Issues Resolved

### Testing

### Check List


derek-ho:
Will fix #3775 #3876

cwperks:
Related issue: https://github.com/opensearch-project/security/issues/2402

cwperks:
@derek-ho @DarshitChanpura What should the behavior of `-y` be for `install_demo_configuration` script? Generally, `-y` is an option in CLIs to automatically respond `yes` to all user prompts. There's a longstanding bug in the `install_demo_configuration` script with the `-y` option where it was not actually answering yes to all prompts. 

It looks like the recent switch to implementing the demo configuration installer to Java fixed the bug, but it then introduced issues in the distribution build.

cwperks (dismissed):
Approving as this is inline with previous behavior.

derek-ho:
> Approving as this is inline with previous behavior.

I added another test to verify this behavior.

src/test/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurerTests.java
```diff
@@ -241,6 +241,18 @@ public void testIsStringAlreadyPresentInFile_isPresent() throws IOException {
...
+    @Test
+    public void testAssumeYesDoesNotInitializeClusterMode() throws IOException {
+        String str1 = "node.name"; // cluster_mode
+        String str2 = "plugins.security.allow_default_init_securityindex"; // init_security
+
+        installer.assumeyes = true;
+        securitySettingsConfigurer.writeSecurityConfigToOpenSearchYML();
+
+        assertThat(isKeyPresentInYMLFile(installer.OPENSEARCH_CONF_FILE, str1), is(equalTo(false)));
+        assertThat(isKeyPresentInYMLFile(installer.OPENSEARCH_CONF_FILE, str2), is(equalTo(false)));
```
cwperks, line 253:
```suggestion
        assertThat(isKeyPresentInYMLFile(installer.OPENSEARCH_CONF_FILE, str1), is(false));
        assertThat(isKeyPresentInYMLFile(installer.OPENSEARCH_CONF_FILE, str2), is(false));
```

src/test/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurerTests.java
```diff
@@ -241,6 +241,18 @@ public void testIsStringAlreadyPresentInFile_isPresent() throws IOException {
         assertThat(isKeyPresentInYMLFile(installer.OPENSEARCH_CONF_FILE, str2), is(equalTo(false)));
     }
 
+    @Test
+    public void testAssumeYesDoesNotInitializeClusterMode() throws IOException {
+        String str1 = "node.name"; // cluster_mode
```
cwperks, line 246: nit: These variable names could be more related to their values or placed inline in the `isKeyPresentInYMLFile` call.

derek-ho: Applied suggestion in: [3b5d54e](https://github.com/opensearch-project/security/pull/3935/commits/3b5d54ea76658e84ca9214fa5212c61c7d763941)

src/test/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurerTests.java
derek-ho: Done in: [60769b3](https://github.com/opensearch-project/security/pull/3935/commits/60769b3e2461377e9d0c459e519ecc9ad0a35762)

DarshitChanpura:
> @derek-ho @DarshitChanpura What should the behavior of -y be for install_demo_configuration script? Generally, -y is an option in CLIs to automatically respond yes to all user prompts. There's a longstanding bug in the install_demo_configuration script with the -y option where it was not actually answering yes to all prompts.

This what I assumed to when writing the Java code. But turns out that the original script did nothing to `cluster_mode` and `init_security` when -y was passed which was not the case when I wrote the Java code.

> It looks like the recent switch to implementing the demo configuration installer to Java fixed the bug, but it then introduced issues in the distribution build.

Agreed that -y should inherently mark other options two options as  yes, but since it is breaking builds rn, I'm happy to revert that bug fix and address it later. After this change is merged, -y will only be reduced to choosing yes for installing demo certificates and the other two prompts will be assumed as no.


An ideal usage would be to either use -y ( which automatically assumes yes for demo certs, cluster_mode and init_security)

OR pass individual options : -d for demo certs (new option), -c for cluster mode and -i for init security modules.

where -y will always take priority. But maybe that is for another release.

cwperks (approved):

DarshitChanpura (approved):
Looks good. This existing bug which was addressed in original PR and is being reverted here, should be addressed in future.

PR #3939, opened 2024-01-11
[Enhancement-3191] `transport_enabled` setting on an auth domain and authorizer may be unnecessary after transport client removal 

### LINKED ISSUE #3191: `transport_enabled` setting on an auth domain and authorizer may be unnecessary after transport client removal

prabhask5:
### Description
Remove the transport_enabled setting in config.yml and its related code in DynamicConfigModelV6/7, ConfigV6/7, associated tests. Additionally add handlers in Config files to handle when transport_enabled is still used within the config.yml file while alerting the user with a message in startup that this setting can be safely removed from the config.

Enhancement

#3191 states that since the transport client has been removed, there is no need for the transport_enabled config and its related code to exist in the codebase, so this ticket removes that unnecessary code.

Nothing.

### Issues Resolved
#3191 

No

### Testing
No tests were made, all tests should run since the behavior should not change.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/securityconf/impl/v6/ConfigV6.java
+++ b/src/main/java/org/opensearch/security/securityconf/impl/v6/ConfigV6.java
@@ -38,11 +38,15 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import org.opensearch.security.DefaultObjectMapper;
 import org.opensearch.security.auth.internal.InternalAuthenticationBackend;
 
 public class ConfigV6 {
+
+    private static final Logger LOGGER = LogManager.getLogger(ConfigV6.class);
 
     public Dynamic dynamic;
 
@@ -244,6 +248,18 @@ public class ConfigV6 {
                 + "]";
         }
 
+        @JsonAnySetter
+        public void unknownPropertiesHandler(String name, Object value) {
+            if (name.equals("transport_enabled")) {
+                LOGGER.info(
+                    "Detected transport_enabled setting in config.yml file. Since the transport client has been removed, this setting is now unnecessary/unsupported and therefore can be safely removed."
+                );
+                System.out.println(
+                    "Detected transport_enabled setting in config.yml file. Since the transport client has been removed, this setting is now unnecessary/unsupported and therefore can be safely removed."
+                );
+            }
+        }
+
     }
 
     public static class HttpAuthenticator {
@@ -345,6 +361,18 @@ public class ConfigV6 {
                 + ", authorization_backend="
                 + authorization_backend
                 + "]";
+        }
+
+        @JsonAnySetter
+        public void unknownPropertiesHandler(String name, Object value) {
+            if (name.equals("transport_enabled")) {
+                LOGGER.info(
+                    "Detected transport_enabled setting in config.yml file. Since the transport client has been removed, this setting is now unnecessary/unsupported and therefore can be safely removed."
+                );
+                System.out.println(
+                    "Detected transport_enabled setting in config.yml file. Since the transport client has been removed, this setting is now unnecessary/unsupported and therefore can be safely removed."
+                );
+            }
         }
 
     }
--- a/src/main/java/org/opensearch/security/securityconf/impl/v6/ConfigV6.java
+++ b/src/main/java/org/opensearch/security/securityconf/impl/v6/ConfigV6.java
@@ -38,15 +38,14 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonProcessingException;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
 
 import org.opensearch.security.DefaultObjectMapper;
 import org.opensearch.security.auth.internal.InternalAuthenticationBackend;
+import org.opensearch.security.setting.DeprecatedSettings;
 
 public class ConfigV6 {
-
-    private static final Logger LOGGER = LogManager.getLogger(ConfigV6.class);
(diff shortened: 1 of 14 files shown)


src/main/java/org/opensearch/security/securityconf/impl/v6/ConfigV6.java
```diff
@@ -38,12 +38,16 @@
...
 import com.fasterxml.jackson.core.JsonProcessingException;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import org.opensearch.security.DefaultObjectMapper;
 import org.opensearch.security.auth.internal.InternalAuthenticationBackend;
 
 public class ConfigV6 {
 
+    private static final Logger LOGGER = LogManager.getLogger(ConfigV6.class);
```
peternied, line 49: Lets reuse and update the existing DeprecatedSettings class to support cases this cases, there are many spots with duplicated code in this class and others.

src/main/java/org/opensearch/security/securityconf/impl/v6/ConfigV6.java
```diff
@@ -248,6 +248,15 @@ public String toString() {
                 + "]";
         }
 
+        @JsonAnySetter
+        public void unknownPropertiesHandler(String name, Object value) {
```
peternied, line 252: If the property name doesn't match an error is not thrown?  Lets add/modify tests to 1) make sure the previous behavior is unchanged for unknown properties, and 2) the new behavior logs that message.

Example test case for capturing the deprecation log message:
https://github.com/opensearch-project/security/blob/b905999f81093fc96db8371df5f96badabb9ff38/src/test/java/org/opensearch/security/setting/DeprecatedSettingsTest.java#L30-L31

src/main/java/org/opensearch/security/securityconf/impl/v6/ConfigV6.java
```diff
@@ -248,6 +248,15 @@ public String toString() {
...
+            if (name.equals("transport_enabled")) {
+                LOGGER.warn(
+                    "Detected transport_enabled setting in config.yml file under an AuthcDomain. Since the transport client has been removed, this setting is now unnecessary/unsupported and therefore can be safely removed."
```
peternied, line 255: Cleaned up this text, it would also be an improvement if the message could state which authenticator is associated with the value.
```suggestion
                    "In OpenSearch v2.0.0+ the setting 'transport_enabled' is deprecated, it should be removed from the authenticator named '{}'."
```

scrawfor99 (commented):
Hi @prabhask5 thanks for making these changes. Great job keeping things up to date! Other than Peter's suggestions, things look good to me. Let me know when you address those and I can give a final pass through and approve (assuming nothing else comes up etc.)

DarshitChanpura (commented):
Changes look good to me. Have we analyzed the blast radius for removing these files? Will there be downstream plugins that will have broken tests?

prabhask5:
> Changes look good to me. Have we analyzed the blast radius for removing these files? Will there be downstream plugins that will have broken tests?

I haven't checked for any other repository tests being impacted by this change, though I can confirm that the only errors that were caused by removing this setting being a parsing error from Jackson which I resolved using the unknownPropertiesHandler. If this error would have occurred in downstream plugins it should not.

scrawfor99 (dismissed):
Hi @prabhask5, thanks for your work with this! Looks good to me :)

prabhask5:
@scrawfor99 Oops accidentally re-requested a review from you, can you approve again?

DarshitChanpura:
> @DarshitChanpura I haven't checked for any other repository tests being impacted by this change, though I can confirm that the only errors that were caused by removing this setting being a parsing error from Jackson which I resolved using the unknownPropertiesHandler. If this error would have occurred in downstream plugins it should not. I'm new to contributing, can you let me know of other plugins to check tests for?

I'm not aware of any. We can merge this and if there are any downstream failures we can revert.

DarshitChanpura (approved):

src/main/java/org/opensearch/security/setting/DeprecatedSettings.java
```diff
@@ -28,4 +28,16 @@ public static void checkForDeprecatedSetting(final Settings settings, final Stri
...
         }
     }
+
+    /**
+     * Logs that a specific setting is deprecated, including a specific supplemental message parameter containing information that details where this setting can be removed from. Should be used in cases where a setting is not supported by the codebase and processing it would introduce errors on setup.
+     */
+    public static void logCustomDeprecationMessage(final String deprecationLocationInformation, final String deprecatedSettingKey) {
+        DEPRECATION_LOGGER.deprecate(
+            deprecatedSettingKey,
+            "In OpenSearch v2.0.0+ the setting '{}' is deprecated, it should be removed from the relevant config file using the following location information: "
```
cwperks, line 38:
```suggestion
            "The setting '{}' is deprecated, it should be removed from the relevant config file using the following location information: "
```

Do you think we need to include a version in this log message? If the version is needed here can it be obtained using the [Version](https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/Version.java) class from core?

prabhask5: Not sure if it's necessary I was going with the requested changes from before, should version be used?

prabhask5: I think giving more information might be better so I'll leave it in for now, I'll use the version class to keep it up to date

peternied (approved):

src/main/java/org/opensearch/security/setting/DeprecatedSettings.java
peternied: Nit; I'd drop the version number completely, the version when a setting was deprecated != the current version.  However, this is pretty minor IMO

cwperks (approved):

PR #3953, opened 2024-01-16
Improve code coverage for SSLNettyTransport class

scrawfor99:
### Description
This change increases code coverage for the SecuritySSLNettyTransport class. In the middle of 12/23, a few unit tests were added to give coverage to different parts of the class. This change builds on these existing changes. 

### Issues Resolved
Box three of https://github.com/opensearch-project/security/issues/3137

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransport.java
+++ b/src/main/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransport.java
@@ -39,6 +39,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import org.opensearch.ExceptionsHelper;
+import org.opensearch.OpenSearchSecurityException;
 import org.opensearch.Version;
 import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.common.network.NetworkService;
@@ -104,7 +105,7 @@ public class SecuritySSLNettyTransport extends Netty4Transport {
     }
 
     // This allows for testing log messages
-    public Logger getLogger() {
+     Logger getLogger() {
         return logger;
     }
 
@@ -120,6 +121,9 @@ public class SecuritySSLNettyTransport extends Netty4Transport {
         errorHandler.logError(cause, false);
         getLogger().error("Exception during establishing a SSL connection: " + cause, cause);
 
+        if (channel == null || !channel.isOpen()) {
+            throw new OpenSearchSecurityException("The provided TCP channel is invalid.", e);
+        }
         super.onException(channel, e);
     }
 
--- a/src/main/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransport.java
+++ b/src/main/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransport.java
@@ -103,6 +103,11 @@ public class SecuritySSLNettyTransport extends Netty4Transport {
         this.SSLConfig = SSLConfig;
     }
 
+    // This allows for testing log messages
+    public Logger getLogger() {
+        return logger;
+    }
+
     @Override
     public void onException(TcpChannel channel, Exception e) {
 
@@ -113,7 +118,7 @@ public class SecuritySSLNettyTransport extends Netty4Transport {
         }
 
         errorHandler.logError(cause, false);
-        logger.error("Exception during establishing a SSL connection: " + cause, cause);
+        getLogger().error("Exception during establishing a SSL connection: " + cause, cause);
 
         super.onException(channel, e);
     }
@@ -156,7 +161,7 @@ public class SecuritySSLNettyTransport extends Netty4Transport {
             }
 
             errorHandler.logError(cause, false);
-            logger.error("Exception during establishing a SSL connection: " + cause, cause);
+            getLogger().error("Exception during establishing a SSL connection: " + cause, cause);
 
             super.exceptionCaught(ctx, cause);
         }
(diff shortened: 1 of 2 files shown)


scrawfor99:
Note: This is taking a bit longer than expected because I found some unexpected behavior in Mockito.

Specifically:

Mocking the SSLExceptionHandler like:

```
 @Mock
    private SslExceptionHandler sslExceptionHandler;
``` 

Results in the following test errors:

```

Tests with failures:
 - org.opensearch.security.ssl.transport.SecuritySSLNettyTransportTests.OnException_withNullExceptionShouldSucceed
 - org.opensearch.security.ssl.transport.SecuritySSLNettyTransportTests.OnException_withDecoderExceptionShouldGetCause
```

However mocking the same class as:

```
    sslExceptionHandler = mock(SslExceptionHandler.class);
```

Leads to a single error: 


```

Tests with failures:
 - org.opensearch.security.ssl.transport.SecuritySSLNettyTransportTests.OnException_withNullChannelShouldThrowException

```

I don't see any reason for this behavior to be different. The second version lets us pass more tests which is nice but causes no error to be thrown and I am not sure why.

src/test/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransportTests.java
```diff
@@ -58,42 +72,61 @@ public class SecuritySSLNettyTransportTests {
...
+                trace
+        ));
     }
 
     @Test
     public void OnException_withNullChannelShouldThrowException() {
 
         NullPointerException exception = new NullPointerException("Test Exception");
-
         Assert.assertThrows(NullPointerException.class, () -> securitySSLNettyTransport.onException(null, exception));
```
peternied, line 111: I see the following failure, to me this indicates that there is a resource binding that needs to be freed up in another test.  Seeing as there is a doStart method in TcpTransport, maybe there is a counterpart that can stop using or free bound resources.

```
org.opensearch.security.ssl.transport.SecuritySSLNettyTransportTests > OnException_withNullChannelShouldThrowException FAILED
    java.lang.IllegalStateException: availableProcessors is already set to [4], rejecting [4]
        at io.netty.util.NettyRuntime$AvailableProcessorsHolder.setAvailableProcessors(NettyRuntime.java:51)
        at io.netty.util.NettyRuntime.setAvailableProcessors(NettyRuntime.java:87)
        at org.opensearch.transport.netty4.Netty4Utils.setAvailableProcessors(Netty4Utils.java:75)
        at org.opensearch.transport.netty4.Netty4Transport.<init>(Netty4Transport.java:139)
        at org.opensearch.security.ssl.transport.SecuritySSLNettyTransport.<init>(SecuritySSLNettyTransport.java:89)
        at org.opensearch.security.ssl.transport.SecuritySSLNettyTransportTests.setup(SecuritySSLNettyTransportTests.java:91)
```

peternied (commented):
Note; I think you should be able to keep using the @Mock declaration, I'm pretty sure there is some resource that needs to be freed up

src/test/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransportTests.java
scrawfor99: HI @peternied thanks for the advice. I will take a look down that path today. I just wanted to open this as a draft to record what I had done yesterday before heading home. I agree that there should be consistent behavior between the two so something weird is going on...

scrawfor99: Update: The issue is the mock use, the existing tests don't actually initialize any of the mocks they declare. That is why the behavior is different. The existing tests happen by coincidence since you cannot actually mock some of the objects i.e. NetworkService which is final. 

I will go ahead and start working on rewriting this class.

src/main/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransport.java
```diff
@@ -103,6 +103,11 @@ public SecuritySSLNettyTransport(
         this.SSLConfig = SSLConfig;
     }
 
+    // This allows for testing log messages
+    public Logger getLogger() {
```
DarshitChanpura, line 107: maybe we can make the logger package-private since its static and is final , we don't need encapsulation.

src/test/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransportTests.java
DarshitChanpura: You may explicitly need to unset the number of available processors. A [quick google search](https://www.google.com/search?q=java.lang.IllegalStateException%3A+availableProcessors+is+already+set+to+%5B4%5D%2C+rejecting+%5B4%5D&rlz=1C5GCEM_enUS989US990&oq=java.lang.IllegalStateException%3A+availableProcessors+is+already+set+to+%5B4%5D%2C+rejecting+%5B4%5D&gs_lcrp=EgZjaHJvbWUyBggAEEUYOdIBBzI2M2owajeoAgCwAgA&sourceid=chrome&ie=UTF-8) led me to this [stack overflow post](https://stackoverflow.com/questions/59352973/elasticsearchjava-lang-illegalstateexception-availableprocessors-is-already-se).

scrawfor99: Hi @DarshitChanpura thanks for your suggestion. I am still working on these changes and will go ahead and mark this as ready for review shortly. 

For the overflow post you shared that looks similar but I was able to resolve the issue we were encountering by fixing the mocks. The prior setup was not instantiating things like I mentioned. You need to trigger the mock instantiations one way or another and the prior version was not actually doing that. The changes should be ready shortly and should correct the coverage gaps.

src/main/java/org/opensearch/security/ssl/transport/SecuritySSLNettyTransport.java
scrawfor99: Sure will do.

peternied (approved):

PR #4003, opened 2024-01-30
Added integration test for telemetry Plugin

### LINKED ISSUE #3961: [Integration Test] Add integration test for telemetry plugin
We have seen a compatibility issues between telemetry plugin and the security plugin. So we should have a integration test in the security plugin repo to make sure there is no conflict.

Write an integration test.

We can use this integ test as a reference that might provide some guidance on cluster configuration and a basic setup.
https://github.com/opensearch-project/security/blob/main/src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java

atharvasharma61:
### Description
Added an integration test to check if the cluster comes up healthy with both the telemetry and security plugin installed.

### Issues Resolved
#3961

### Testing
Both plugins were loaded into the local cluster and test succeeded
```
2024-01-30 17:17:25 SUITE-TelemetryPluginIntegrationTests-seed#[E0B6B58EF43BC0AC] INFO  PluginsService:233 - loaded plugin [org.opensearch.security.OpenSearchSecurityPlugin]
2024-01-30 17:17:25 SUITE-TelemetryPluginIntegrationTests-seed#[E0B6B58EF43BC0AC] INFO  PluginsService:233 - loaded plugin [org.opensearch.telemetry.OTelTelemetryPlugin]
```

### Check List


build.gradle
```diff
@@ -574,6 +574,7 @@ check.dependsOn integrationTest
 
 dependencies {
     implementation "org.opensearch.plugin:transport-netty4-client:${opensearch_version}"
+    implementation "org.opensearch.plugin:telemetry-otel-client:${opensearch_version}"
```
reta, line 577: We should never use this artifact , it is an sad oversight we publish it (@Gaganjuneja fyi), the plugins should only be imported and used as ZIP archives.

src/integrationTest/java/org/opensearch/security/TelemetryPluginIntegrationTests.java
reta, line 1: @peternied the `security` plugin seems to have own ways to spin up local clusters, that has a lot of overlapping with what OpenSearch provides, usually other plugins:
  - use test clusters OpenSearch plugin to bootstrap distribution 
  - instruct test clusters OpenSearch plugin to install additional plugins (using standard ZIP distribution)
  - run integration tests

DarshitChanpura (commented):
Is this PR just adding a sanity check for successful installation of Telemetry plugin with a secure cluster?

src/integrationTest/java/org/opensearch/security/TelemetryPluginIntegrationTests.java
```diff
@@ -0,0 +1,39 @@
...
+import org.opensearch.test.framework.cluster.TestRestClient;
+
+import org.opensearch.telemetry.OTelTelemetryPlugin;
+import static org.apache.http.HttpStatus.SC_OK;
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.*;
+import static org.opensearch.test.framework.TestSecurityConfig.Role.ALL_ACCESS;
+
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class TelemetryPluginIntegrationTests {
```
scrawfor99, line 21: Can you please provide info about why this needs to be in the Security plugin? 

I understand that the Telemetry plugin was added to core but this means that a change in the Telemetry plugin could cause a test in Security to fail--it makes it upstream of Security. 

I don't think we should have any plugins upstream of Security if we can help it and I am not sure why this test needs to be housed in the Security repo instead of using the same framework other core-internal plugins (S3 for example) do.

src/integrationTest/java/org/opensearch/security/TelemetryPluginIntegrationTests.java
peternied: Yes - OpenSearch test cases make many assumptions that work well in happy path scenarios, such as using the Java client or assuming the cluster forms.   For better or worse, security often tests things that are 'strange' for a client to do to in order to make sure the failure cases are handled correctly.

If you feel that we are overlapping specific functionality that is duplicated I'm happy to dig in - want to create an issue?

reta: > If you feel that we are overlapping specific functionality that is duplicated I'm happy to dig in - want to create an issue?

Sure, I will pick it up shortly, thanks Peter!

reta: https://github.com/opensearch-project/security/issues/4010

src/integrationTest/java/org/opensearch/security/TelemetryPluginIntegrationTests.java
atharvasharma61: Tagging @Gaganjuneja to answer this since he created the issue and suggested to write the test in security repo.

build.gradle
DarshitChanpura: +1. If it is absolutely needed, can we have it as a testImplementation dependency?

src/integrationTest/java/org/opensearch/security/TelemetryPluginIntegrationTests.java
DarshitChanpura: @Gaganjuneja @atharvasharma61 If the intent is to test the plugin behavior with security, I'd suggest to write a sanity test inside telemetry plugin repo with security plugin enabled

peternied (changes requested):
@atharvasharma61 Thanks for this contribution.  There are several failing checks - please get those to green.

We need to resolve the comment thread around how the telemetry plugin is referenced and used in these test cases.  Please follow up on https://github.com/opensearch-project/security/pull/4003/files#r1471334295

scrawfor99:
Hi @atharvasharma61, are there still plans to try to move forward with this change or is the Telemetry plugin heading in a different direction for testing with security?

atharvasharma61:
@Gaganjuneja to take the call.

Gaganjuneja:
One thing needs to be clarified here is that there is no dependency between telemetry and security plugins. Telemetry plugin is an implementation of telemetry lib in the core. So once the telemetry feature is enabled through plugin we are instrumenting the network layer which creates some wrapper objects. So the ask here is to bring up the OpenSearch cluster with security plugin installed and telemetry feature enabled.

If we have a mechanism to enable the security plugin in the OpenSearch integ testcases then it can easily be tested there.

build.gradle
Gaganjuneja: You are right. Till the time https://github.com/opensearch-project/security/issues/4010 is not closed we can go ahead with having it as a testImplementation dependency.

Gaganjuneja:
> > Thanks for following up @Gaganjuneja
> > 
> > 1. Could you please run `./gradlew spotlessApply` to fix hygiene check failures.
> > 2. Could you add your comment on this thread: [Added integration test for telemetry Plugin #4003 (comment)](https://github.com/opensearch-project/security/pull/4003#discussion_r1471334295)
> 
> Thanks, I will update on this.

If we are ok having it as a testImplementation dependency. We can cleanup this PR and bring it to the ready state for merge.

Gaganjuneja:
@reta, What do you suggest, should we hold on this till https://github.com/opensearch-project/security/issues/4010 or we can go ahead with adding the plugin dependency testImplementation?

peternied:
@reta Did you have an update from conversations with the build team?

reta:
> @reta Did you have an update from conversations with the build team?

Yes, I did actually! The good news - we could leverage the existing integTest scripts to run post distribution smoke tests,  I am learning the way it works now to open up the detailed issue against opensearch-build repo, just need a bit more time, thanks @peternied

peternied:
This repository is not the best place to include these test cases, please work with the opensearch-build's test workflow [1] to create and onboard a telemetry plugin test suite.  Closing out this pull request.

- [1] https://github.com/opensearch-project/opensearch-build/tree/main/src/test_workflow

Issue #2569, opened 2023-03-19
[FEATURE] Score-based password strength verification


willyborankin:
**What solution would you like?**
The current password verification based on pattern marching is not so good as it could be. 
To improve it I suggest to use `zxcvbn4j`  which is a port dropbox `zxcvbn` library which measures 
the strength of the password. 

To implement it such settings need to be added to the plugin configuration:
   
- `plugins.security.restapi.password_min_length` - minimum password length, default and minimum is `8` 
- `plugins.security.restapi.password_score_based_validation_strength` - the strength of the valid password
   Possible values:
     - `fair` - very guessable password: protection from throttled online attacks
     - `good` - somewhat guessable password: protection from unthrottled online attacks
     - `strong` - safely unguessable password: moderate protection from offline slow-hash scenario
     - `very_strong` - very unguessable password: strong protection from offline slow-hash scenario

By default the plugin always checks strength of the password and its minimal length together with the regular expression if its set.
The current implementation of checking username similarity will be changed in favor of `zxcvbn` similarity by adding `username` to the `user_inputs` dictionary which means that usernames and passwords like: 
 - `Andrey_Pleskach`  `Andrey_Pleskach_asdsadas!2e23`  
- `andrey_pleskach`  `asdsadas!2e23-Andrey_Pleskach`

Additional notes: as @peternied the performance of the library `~5-20ms` which is a good trade off since update and set a new user password is not so common operation compare to get list of roles, roles mapping etc.
The calculation time for passwords around `100` characters is `~100ms` as result to avoid of performance degradation for big passwords I suggest to set max length of the password to `100`.

scrawfor99:
[Triage] Hi @willyborankin, thank you for filing this issue. This seems like a good issue and you can find some more feedback on the linked PR. Thank you!

Hannes-Keunen:
Hi, since this issue is still open, I'd like to add some feedback here.

While this is a useful feature, it is a bit problematic for our use case.
We are currently pushing user accounts from our own system to opensearch via the REST API. We have our own password verification method so we do not want opensearch to validate passwords again, because there are cases where verification in opensearch fails where our own method does not, even with `plugins.security.restapi.password_score_based_validation_strength: fair`. I was wondering if a setting exists/could be added to disable this feature completely?

davidlago:
Closing as #2557 merged/released. As to disabling this, if those 2 config values are not provided, this check should not come into play (@willyborankin please confirm)

PR #4071, opened 2024-02-25
Log password requirement details in demo environment

### LINKED ISSUE #4048: [FEATURE] Log more information about Password strength feature
It was brought up that we log out that a password is weak, with no steps as to how to determine the strength of a password. Maybe linking to this: https://lowe.github.io/tryzxcvbn/, or at least logging out that we are using zxcvbn could help.

camerondurham:
### Description

Logs details about the password validations done when demo clusters are created. The current message "Password $ADMIN_PASSWORD is weak. Please re-try with a stronger password." doesn't provide enough information communicate what the user needs to set the password to.

* Category: Enhancement
   * See #4048 - to improve UX when creating test OpenSearch clusters by logging details on password requirements
    * Before: OS 2.12 fails to start with message `Password sample@1029 is weak. Please re-try with a stronger password.`
    * After: security plugin fails to start with message `Password weakpassword failed validation with error "Weak password". Please re-try with a minimum 8 character password that is "strong" per zxcvbn validation.`

### Issues Resolved

Potentially can resolve #4048, though would need issue author @derek-ho to confirm this is the intended change.

### Testing

Updated existing unit test that tests "password validation failure". 

### Check List
- [ ] New functionality has been documented <-- In a separate PR, I also want to update relevant docs, including [docker getting started docs](https://opensearch.org/docs/latest/install-and-configure/install-opensearch/docker/#run-opensearch-in-a-docker-container) docs with perhaps the zxcvbn "validation" tool.

### DIFF
--- a/src/main/java/org/opensearch/security/support/ConfigConstants.java
+++ b/src/main/java/org/opensearch/security/support/ConfigConstants.java
@@ -273,11 +273,9 @@ public class ConfigConstants {
     public static final String SECURITY_RESTAPI_ADMIN_ENABLED = "plugins.security.restapi.admin.enabled";
     public static final String SECURITY_RESTAPI_ENDPOINTS_DISABLED = "plugins.security.restapi.endpoints_disabled";
     public static final String SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX = "plugins.security.restapi.password_validation_regex";
-    public static final String SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX_DEFAULT = "(?=.*[A-Z])(?=.*[^a-zA-Z\\\\d])(?=.*[0-9])(?=.*[a-z]).{8,}";
     public static final String SECURITY_RESTAPI_PASSWORD_VALIDATION_ERROR_MESSAGE =
         "plugins.security.restapi.password_validation_error_message";
     public static final String SECURITY_RESTAPI_PASSWORD_MIN_LENGTH = "plugins.security.restapi.password_min_length";
-    public static final Integer SECURITY_RESTAPI_PASSWORD_MIN_LENGTH_DEFAULT = 8;
     public static final String SECURITY_RESTAPI_PASSWORD_SCORE_BASED_VALIDATION_STRENGTH =
         "plugins.security.restapi.password_score_based_validation_strength";
     // Illegal Opcodes from here on
--- a/src/main/java/org/opensearch/security/support/ConfigConstants.java
+++ b/src/main/java/org/opensearch/security/support/ConfigConstants.java
@@ -273,9 +273,11 @@ public class ConfigConstants {
     public static final String SECURITY_RESTAPI_ADMIN_ENABLED = "plugins.security.restapi.admin.enabled";
     public static final String SECURITY_RESTAPI_ENDPOINTS_DISABLED = "plugins.security.restapi.endpoints_disabled";
     public static final String SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX = "plugins.security.restapi.password_validation_regex";
+    public static final String SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX_DEFAULT = "(?=.*[A-Z])(?=.*[^a-zA-Z\\\\d])(?=.*[0-9])(?=.*[a-z]).{8,}";
     public static final String SECURITY_RESTAPI_PASSWORD_VALIDATION_ERROR_MESSAGE =
         "plugins.security.restapi.password_validation_error_message";
     public static final String SECURITY_RESTAPI_PASSWORD_MIN_LENGTH = "plugins.security.restapi.password_min_length";
+    public static final Integer SECURITY_RESTAPI_PASSWORD_MIN_LENGTH_DEFAULT = 8;
     public static final String SECURITY_RESTAPI_PASSWORD_SCORE_BASED_VALIDATION_STRENGTH =
         "plugins.security.restapi.password_score_based_validation_strength";
     // Illegal Opcodes from here on
--- a/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
+++ b/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
@@ -39,9 +39,7 @@ import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
 
 import static org.opensearch.security.support.ConfigConstants.SECURITY_RESTAPI_PASSWORD_MIN_LENGTH;
-import static org.opensearch.security.support.ConfigConstants.SECURITY_RESTAPI_PASSWORD_MIN_LENGTH_DEFAULT;
 import static org.opensearch.security.support.ConfigConstants.SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX;
-import static org.opensearch.security.support.ConfigConstants.SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX_DEFAULT;
 
 /**
  * This class updates the security related configuration, as needed.
@@ -78,6 +76,7 @@ public class SecuritySettingsConfigurer {
         ".plugins-flow-framework-templates",
         ".plugins-flow-framework-state"
     );
+    static final Integer DEFAULT_PASSWORD_MIN_LENGTH = 8;
     static String ADMIN_PASSWORD = "";
     static String ADMIN_USERNAME = "admin";
 
@@ -132,8 +131,8 @@ public class SecuritySettingsConfigurer {
         try {
             final PasswordValidator passwordValidator = PasswordValidator.of(
                 Settings.builder()
-                    .put(SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX, SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX_DEFAULT)
-                    .put(SECURITY_RESTAPI_PASSWORD_MIN_LENGTH, SECURITY_RESTAPI_PASSWORD_MIN_LENGTH_DEFAULT)
+                    .put(SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX, "(?=.*[A-Z])(?=.*[^a-zA-Z\\\\d])(?=.*[0-9])(?=.*[a-z]).{8,}")
+                    .put(SECURITY_RESTAPI_PASSWORD_MIN_LENGTH, DEFAULT_PASSWORD_MIN_LENGTH)
                     .build()
             );
 
@@ -148,10 +147,12 @@ public class SecuritySettingsConfigurer {
                 RequestContentValidator.ValidationError response = passwordValidator.validate(ADMIN_USERNAME, ADMIN_PASSWORD);
                 if (!RequestContentValidator.ValidationError.NONE.equals(response)) {
                     System.out.println(
-                            String.format(
-                                    "Password %s failed validation: \"%s\". Please re-try with a minimum %d character password that is strong per zxcvbn validation."
-                                    , ADMIN_PASSWORD, response.message(), SECURITY_RESTAPI_PASSWORD_MIN_LENGTH_DEFAULT
-                            )
+                        String.format(
+                            "Password %s failed validation: \"%s\". Please re-try with a minimum %d character password and must contain at least one uppercase letter, one lowercase letter, one digit, and one special character that is strong. Password strength can be tested here: https://lowe.github.io/tryzxcvbn",
+                            ADMIN_PASSWORD,
+                            response.message(),
+                            DEFAULT_PASSWORD_MIN_LENGTH
+                        )
                     );
                     System.exit(-1);
                 }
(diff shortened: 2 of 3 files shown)


src/test/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurerTests.java
```diff
@@ -125,7 +127,13 @@ public void testUpdateAdminPasswordWithWeakPassword() throws NoSuchFieldExceptio
             System.setSecurityManager(null);
         }
 
-        verifyStdOutContainsString("Password weakpassword is weak. Please re-try with a stronger password.");
+        verifyStdOutContainsString(
+                String.format(
+                        "Password weakpassword failed validation: \"%s\". Please re-try with a minimum %d character password that is strong per zxcvbn validation.",
+                        INVALID_PASSWORD_INVALID_REGEX.message(),
```
camerondurham, line 133: Should I include more tests for each of the other validation failures? (e.g. too short password, weak password)

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -273,9 +273,11 @@ public enum RolesMappingResolution {
     public static final String SECURITY_RESTAPI_ADMIN_ENABLED = "plugins.security.restapi.admin.enabled";
     public static final String SECURITY_RESTAPI_ENDPOINTS_DISABLED = "plugins.security.restapi.endpoints_disabled";
     public static final String SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX = "plugins.security.restapi.password_validation_regex";
+    public static final String SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX_DEFAULT = "(?=.*[A-Z])(?=.*[^a-zA-Z\\\\d])(?=.*[0-9])(?=.*[a-z]).{8,}";
```
camerondurham, line 276: I might be using "DEFAULT" incorrectly here, happy to move this back if this isn't the intended use of ConfigConstants.

peternied (commented):
Thanks for the pull request @camerondurham - minor comments

src/test/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurerTests.java
DarshitChanpura: You can add 1 more test around password length validation, otherwise the test-cases should suffice.

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -142,11 +144,17 @@ void updateAdminPassword() {
...
-                && !ADMIN_PASSWORD.isEmpty()
-                && passwordValidator.validate(ADMIN_USERNAME, ADMIN_PASSWORD) != RequestContentValidator.ValidationError.NONE) {
-                System.out.println("Password " + ADMIN_PASSWORD + " is weak. Please re-try with a stronger password.");
-                System.exit(-1);
+            if (shouldValidatePassword && !ADMIN_PASSWORD.isEmpty()) {
+                RequestContentValidator.ValidationError response = passwordValidator.validate(ADMIN_USERNAME, ADMIN_PASSWORD);
+                if (!RequestContentValidator.ValidationError.NONE.equals(response)) {
+                    System.out.println(
+                            String.format(
+                                    "Password %s failed validation: \"%s\". Please re-try with a minimum %d character password that is strong per zxcvbn validation."
```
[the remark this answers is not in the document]
camerondurham: Thanks, will do. I realized these requirements are a bit more clearly spelled out in [these docs](https://opensearch.org/docs/latest/security/configuration/yaml/#password-settings) that I missed, so I'm add something like this:

> Password %s failed validation: "%s". Please re-try with a minimum %d character password and must contain at least one uppercase letter, one lowercase letter, one digit, and one special character that is strong. Password strength can be tested here: https://lowe.github.io/tryzxcvbn

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -131,7 +132,7 @@ void updateAdminPassword() {
             final PasswordValidator passwordValidator = PasswordValidator.of(
                 Settings.builder()
                     .put(SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX, "(?=.*[A-Z])(?=.*[^a-zA-Z\\\\d])(?=.*[0-9])(?=.*[a-z]).{8,}")
-                    .put(SECURITY_RESTAPI_PASSWORD_MIN_LENGTH, 8)
+                    .put(SECURITY_RESTAPI_PASSWORD_MIN_LENGTH, DEFAULT_PASSWORD_MIN_LENGTH)
```
camerondurham, line 135: Taking a closer look at this, the [docs](https://opensearch.org/docs/latest/security/configuration/yaml/#password-settings) say this is configurable via `plugins.security.restapi.password_min_length` but is it actually configurable if it's being set to `8` all the time? I might be totally missing something though.

src/test/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurerTests.java
camerondurham: Thanks! Added a "too short" test case in revision 2.

src/main/java/org/opensearch/security/support/ConfigConstants.java
camerondurham: I see, good point. I've removed `ConfigConstants.java` from the change set and moved things back to where they're used.

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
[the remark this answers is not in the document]
camerondurham: Godd point. I've added the test url and described the validation regex in rev 2.

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
peternied: It would be strange to have different requirements in this tool from the OpenSearch instance that is being setup.

@DarshitChanpura Can you create an issue around this?  The settings configuration should be read from a singular OpenSearch settings, not a separate setting configuration or hardcoded.

derek-ho: +1, not sure but it seems like the regex also has some concept of length, maybe we can combine them?

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -142,11 +143,19 @@ void updateAdminPassword() {
...
-                && !ADMIN_PASSWORD.isEmpty()
-                && passwordValidator.validate(ADMIN_USERNAME, ADMIN_PASSWORD) != RequestContentValidator.ValidationError.NONE) {
-                System.out.println("Password " + ADMIN_PASSWORD + " is weak. Please re-try with a stronger password.");
-                System.exit(-1);
+            if (shouldValidatePassword && !ADMIN_PASSWORD.isEmpty()) {
+                RequestContentValidator.ValidationError response = passwordValidator.validate(ADMIN_USERNAME, ADMIN_PASSWORD);
+                if (!RequestContentValidator.ValidationError.NONE.equals(response)) {
+                    System.out.println(
+                        String.format(
+                            "Password %s failed validation: \"%s\". Please re-try with a minimum %d character password and must contain at least one uppercase letter, one lowercase letter, one digit, and one special character that is strong. Password strength can be tested here: https://lowe.github.io/tryzxcvbn",
```
derek-ho, line 151: I believe that the minimum strength can also be configured with `plugins.security.restapi.password_score_based_validation_strength`, so not sure if we want to hard code strong here? Also not sure how I feel about adding hard coded links into a log message, but maybe its fine, will defer to others here.

derek-ho: More context is the feature was added in this PR: https://github.com/opensearch-project/security/issues/2569 - I pulled this setting from the PR description

derek-ho: Also pretty sure the regex can be changed to not be one uppercase letter, one digit, etc. maybe we should just leave it at "meeting the minimum regex and password strength requirements"? But not sure since I think there will be a decent chunk of people who are using the minimum.

derek-ho:
Thanks for the contribution @camerondurham !

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
DarshitChanpura: `plugins.security.restapi.password_min_length` was introduced in https://github.com/opensearch-project/security/pull/2557 for password validation. 

The password length is not configurable for admin user since the demo script doesn't read settings from opensearch.yml as it is executed before reading the yml file. Remember, admin password is changeable, and the password validation introduced in #2557 should come into play when verifying new password. For changes related to demo script, specifying that the password has to be 8 characters or longer should be enough. 

Let me know if I missed something here.

src/test/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurerTests.java
```diff
@@ -125,7 +128,35 @@ public void testUpdateAdminPasswordWithWeakPassword() throws NoSuchFieldExceptio
...
+            securitySettingsConfigurer.updateAdminPassword();
+        } catch (SecurityException e) {
+            assertThat(e.getMessage(), equalTo("System.exit(-1) blocked to allow print statement testing."));
+        } finally {
+            System.setSecurityManager(null);
+        }
+
+        verifyStdOutContainsString(
+            String.format(
+                "Password short failed validation: \"%s\". Please re-try with a minimum %d character password and must contain at least one uppercase letter, one lowercase letter, one digit, and one special character that is strong. Password strength can be tested here: https://lowe.github.io/tryzxcvbn",
```
DarshitChanpura, line 155: Create a class-level variable:
```java
String PASSWORD_VERIFICATION_FAILURE_MSG = "Password %s failed validation: \"%s\". Please re-try with a minimum %d character password and must contain at least one uppercase letter, one lowercase letter, one digit, and one special character that is strong. Password strength can be tested here: https://lowe.github.io/tryzxcvbn";
```

and then can be used here (and above as):

```java
verifyStdOutContainsString(
            String.format(
                PASSWORD_VERIFICATION_FAILURE_MSG,
                "short",
                INVALID_PASSWORD_TOO_SHORT.message(),
                DEFAULT_PASSWORD_MIN_LENGTH
            )
```

DarshitChanpura (commented):
@camerondurham the settings `plugins.security.restapi.password_min_length` and `plugins.security.restapi.password_score_based_validation_strength` come into play once the security plugin is setup and cluster is ready, which is when all the config settings are read.

However, in case of demo configuration script execution, these settings are not available to be read, as the tools write opensearch.yml with some default settings. 

In case if we try to pass these settings in opensearch.yml, the demo script will stop execution since it thinks that security plugin is already setup because it is able to find `plugins.security.<setting>` already defined in opensearch.yml ([check this out](https://github.com/opensearch-project/security/blob/26b224fa1e2986431f22cf865592c4ce6fa6f49c/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java#L109))

I think we are overthinking this. The regex + password length supplied here are one time use before a fresh cluster start-up. We can leave those settings as is and modify the log message to focus on the scope of this PR. Thoughts?

DarshitChanpura (commented):
Thank you for addressing all the feedback! Nothing more from my end; waiting for CI to be green before approval.

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
derek-ho: Yea got your point. I had a misunderstanding that this was being applied to all user creation/password validation, in which case those custom settings may be present. @DarshitChanpura you are right that that would not be the case, or else they would never make it into this demo script tool. Sorry for the misleading comment.

camerondurham: No worries at all, I had the same misunderstanding. Your comment wasn't misleading :)

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
peternied: >The password length is not configurable for admin user since the demo script doesn't read settings from opensearch.yml as it is executed before reading the yml file. Remember, admin password is changeable, and the password validation introduced in #2557 should come into play when verifying new password. For changes related to demo script, specifying that the password has to be 8 characters or longer should be enough.

I'm not saying we need to implement it, but we should document that the password requirements for this tool that ships with OpenSearch does not use the same settings.  If someone was to implement this functionality I would approve merging it, what do you think @DarshitChanpura ?

peternied (approved):

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
DarshitChanpura: done. https://github.com/opensearch-project/security/issues/4081

DarshitChanpura (approved):

DarshitChanpura:
Thank you for this contribution @camerondurham !

PR #4086, opened 2024-03-02
Fix : Compliance config's readEnabledFieldsCache is not aware of date change

### LINKED ISSUE #3950: [BUG] ComplianceConfig read fields does not expire values from the cache correctly
**What is the bug?**
ComplianceConfig supports auditing for rolling audit logs, such as `audit-logs-YYYY-MM-dd`.  There is a cache for field data, and this caches does not know how to handle data rollover correctly.

1. Create a config with `*` pattern for read fields
2. Use a rolling log config
3. Perform an access with todays config
4. Have the date roll over
5. Perform an access with the same config as 3, however

Easier repro:
1. Go to ComplianceConfigTest, and uncomment the following assertion
https://github.com/opensearch-project/security/blob/c06365ca94c9a1d15e85b578a1ae48168bf0bca7/src/test/java/org/opensearch/security/auditlog/compliance/ComplianceConfigTest.java#L203-L204
2. Run the unit test `./gradlew test --tests org.opensearch.security.auditlog.compliance.ComplianceConfigTest`

The check will be done against the current config, all other configs should be looked up and reported on.

The bug is in ComplianceConfig.java, see how `getFieldsForIndex` is used to build the cache; however, the entry should be invalid after a roll over happens.  How this is determined could be complex and requires investigation.

ksankeerth:
Fix : Compliance config's readEnabledFieldsCache is not aware of date change

### Description
* when the date and audit-log index matches, it is not required to check in cache.
* avoid using readEnabledFieldsCache when the date and audit-log index matches

### Issues Resolved
#3950 

### Testing
* already commented test works after the fix

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/compliance/ComplianceConfig.java
+++ b/src/main/java/org/opensearch/security/compliance/ComplianceConfig.java
@@ -539,6 +539,12 @@ public class ComplianceConfig {
         if (securityIndex.equals(index)) {
             return logInternalConfig;
         }
+        // if the index is used for audit logging (rolling index name), return false
+        if (auditLogPattern != null) {
+            if (index.equalsIgnoreCase(getExpandedIndexName(auditLogPattern, null))) {
+                return false;
+            }
+        }
         try {
             return readEnabledFieldsCache.get(index) != WildcardMatcher.NONE;
         } catch (ExecutionException e) {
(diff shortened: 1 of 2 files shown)


peternied (changes requested):
While this fix addresses the test failure, it prevents the compliance history from being checked against the audit indexes.

Maybe rather than use the index name, the index pattern is used in the cache, what do you think?

scrawfor99:
Hi @ksankeerth, I am just following up to see whether you are still planning on following up on this PR. Please let us know if you need any sort of help.

peternied:
Since there hasn't any updates since this was created, closing this PR out.

Issue #2316, opened 2022-12-09
[FEATURE] New reserved roles are not sourced to security index on an upgrade


cwperks:
**Is your feature request related to a problem?**

New reserved roles are often added between releases when a new plug-in is released and associated roles are created to provision access to features of the new plugin. Reserved roles are added to the security repo `roles.yml` like is done in this PR: https://github.com/opensearch-project/security/pull/2192 which introduces 3 new reserved roles between 2.3.0 and 2.4.0.

When starting up a cluster with the upgraded version of OpenSearch I receive the following line:

```
Index .opendistro_security already contains doc with id roles, skipping update
```

indicating that the new reserved roles in `roles.yml` will need to be manually added either via the API by an admin or via securityadmin tool connecting via clientcert as a super admin.

This becomes additionally complicated when users of OpenSearch add custom roles or modify the security index in any way via the OSD UI or API as it is destructive to run `securityadmin.sh` because it can delete custom configuration. In order to get around this I have:

1. Ran a backup of the security index using security admin: https://opensearch.org/docs/latest/security-plugin/configuration/security-admin/#a-word-of-caution
2. Reconcile the exported `roles.yml` with new reserved roles in the repo's `roles.yml`: https://github.com/opensearch-project/security/blob/main/config/roles.yml
3. Run securityadmin.sh with the coalesced exported `roles.yml` with custom roles and the repo's `roles.yml` of the corresponding version of opensearch

This is not an idea process and there can be a better solution to source new reserved roles during an upgrade

peternied:
@opensearch-project/security-analytics This scenario might be impacting y'all, it might be worthwhile to invest in this scenario.  Here is where I believe there is a lot of overlap, after upgrading a cluster that does not have the new roles needed for the security-analytics plugin when it is 'upgraded-in' on at a later time.

cwperks:
[Triage] In the short-term this can be a documentation fix. 

@peternied to follow-up on this issue.

eirsep:
One possible solution:
**Have 2 security config indices - one for pre-defined roles (internal to Opensearch and its plugins) and one for execlusively for user defined roles**
On every node restart/bootup (or a new admin API invocation), we should reload roles/permissions into the pre-defined roles index from the yml file and leave the user security roles as is.

This would remove some of the static-ness from our design and solve the problem for upgrading domains getting the latest roles and permissions from yml file

peterzhuamazon:
[...]

PR #4027, opened 2024-02-08
v2.12 update roles.yml with new API for experimental alerting plugin feature

AWSHurneyt:
### Description
We added a new API to the alerting plugin to support enhancements to the UI ([link](https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/action/GetRemoteIndexesAction.kt#L13)). The enhancements allow users to configure per query, and per bucket monitors that can query remote clusters via the UI. Previously, such monitors could only be configured using devtools/API commands.

In addition, we added support for configuring cluster metrics monitors which can execute various API ([link to full list of supported API](https://opensearch.org/docs/latest/observing-your-data/alerting/per-cluster-metrics-monitors/#supported-apis)) against remote clusters. The new API is used by the frontend to also support creating these monitors via the UI.

* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation)
  * New experimental feature
  * Enhancement to existing features
* Why these changes are required?
  * To include this API permission in the reserved role that's shipped with the default distribution of OpenSearch.
* What is the old behavior before changes and new behavior after changes?
  * This API was not available before the experimental launch. With the experimental feature enabled, this API can be used to retrieve a list of connect remote clusters, the indexes/aliases on those clusters along with their health statuses, and optionally the mappings for those indexes.

### Issues Resolved
https://github.com/opensearch-project/alerting-dashboards-plugin/issues/796

Is this a backport? If so, please add backport PR # and/or commits #

### Testing
Executed E2E tests with and without the experimental feature enabled to help ensure no regressions.

### Check List
- [x] New functionality includes testing
- [x] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4102, opened 2024-03-08
[Feature] Check for and perform upgrades on security configurations

peternied:
### Description
This adds a new API that allows for checking and updating configurations from the default configurations on disk.  Initial feature supports only Roles.

#### Response when no upgrade is available
```
GET _plugins/_security/api/_upgrade_check
200 {
  "status": "ok",
  "upgradeAvailable" : false
}
```
#### Response when a new role is available
```
GET _plugins/_security/api/_upgrade_check
200 {
  "status": "ok",
  "upgradeAvailable" : true,
  "upgradeActions" : {
    "roles" : {
      "add" : [ "flow_framework_full_access" ]
    }
  }
}
```
#### Response when a new role + existing role were updated
```
GET _plugins/_security/api/_upgrade_check
200 {
  "status": "ok",
  "upgradeAvailable" : true,
  "upgradeActions" : {
    "roles" : {
      "add" : [ "flow_framework_full_access" ],
      "modify" : [ "flow_framework_read_access" ]
    }
  }
}
```
#### Perform an upgrade
```
POST _plugins/_security/api/_upgrade_perform
200 {
  "status" : "OK",
  "upgrades" : {
    "roles" : {
      "add" : [ "flow_framework_full_access" ],
      "modify" : [ "flow_framework_read_access" ]
    }
  }
}
```

#### Perform an upgrade when unneeded
```
POST _plugins/_security/api/_upgrade_perform
400 {
   "status": "BAD_REQUEST",
   "message": "Unable to upgrade, no differences found in 'roles' config"
}
```

### Issues Resolved
- https://github.com/opensearch-project/security/issues/2316

### Testing
New unit test and integration test cases

### Check List


src/main/java/org/opensearch/security/securityconf/impl/CType.java
```diff
@@ -27,14 +27,17 @@
```
peternied, line 27: This file is cherry-picked these from @willyborankin's recent PR that is out in review.
- [1] #4002

src/main/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiAction.java
```diff
@@ -0,0 +1,395 @@
...
+        public ConfigItemChanges(final CType config, final JsonNode differences) {
+            this.config = config;
+            this.itemsGroupedByOperation = classifyChanges(differences);
+        }
+
+        public boolean hasChanges() {
+            return !itemsGroupedByOperation.isEmpty();
+        }
+
+        public void addToNode(final ObjectNode node) {
```
DarshitChanpura, line 368: could you add a brief comment here with an example to visualize what the output would look like for easier readability?
something like this for roles:
```
{
  "roles": {
    "modify": ["role1", "role2"],
    "delete": ["role3"]
  }
}
```

src/main/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiAction.java
```diff
@@ -0,0 +1,395 @@
...
+                }
+            });
+
+            final var itemsGroupedByOperation = items.entrySet()
+                .stream()
+                .collect(Collectors.groupingBy(Map.Entry::getValue, Collectors.mapping(Map.Entry::getKey, Collectors.toList())));
+            return itemsGroupedByOperation;
+        }
+    }
+}
```
DarshitChanpura, line 395: Could you add java-doc comments for all methods? I understand that the names are comprehensive but a brief description should make this class easy to follow.

src/test/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiActionUnitTest.java
```diff
@@ -0,0 +1,291 @@
...
+            tc.updates.accept(updated);
+
+            var sourceAsConfig = SecurityDynamicConfiguration.fromJson(objectMapper.writeValueAsString(source), CType.ROLES, 2, 1, 1);
+
+            doReturn(ValidationResult.success(sourceAsConfig)).when(configUpgradeApiAction)
+                .loadConfiguration(any(), anyBoolean(), anyBoolean());
+            doReturn(updated).when(configUpgradeApiAction).loadConfigFileAsJson(any());
+
+            // Execute
+            var result = configUpgradeApiAction.computeDifferenceToUpdate(CType.ACTIONGROUPS);
```
DarshitChanpura, line 238: should this be CType.ROLES?

src/test/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiActionUnitTest.java
```diff
@@ -0,0 +1,291 @@
...
+                "      \"add\" : [ \"new_role\" ]\n" + //
+                "    }\n" + //
+                "  }\n" + //
+                "}"));
+            return true;
+        }));
+    }
+
+    @Test
+    public void testConfigurationDifferences_OperationBash() throws IOException {
```
DarshitChanpura, line 159: nice 👏

DarshitChanpura (commented):
Thank you @peternied for adding this feature. The code-cov still seems to be complaining about missed lines in test-coverage. Would you please look into that?

src/integrationTest/java/org/opensearch/security/DefaultConfigurationTests.java
```diff
@@ -64,15 +67,79 @@ public static void cleanConfigurationDirectory() throws IOException {
...
             Map<String, Object> users = response.getBodyAs(Map.class);
-            assertThat(users, allOf(aMapWithSize(3), hasKey(ADMIN_USER_NAME), hasKey(NEW_USER), hasKey(LIMITED_USER)));
+            assertThat(
+                users,
+                allOf(aMapWithSize(3), hasKey(ADMIN_USER.getName()), hasKey(NEW_USER.getName()), hasKey(LIMITED_USER.getName()))
+            );
         }
     }
+
+    @Test
```
cwperks, line 85: @peternied Can we also add a test for an existing role that had actions added to it between versions?

Sometimes there are PRs like [this](https://github.com/opensearch-project/security/pull/4027) that add a new action to a role between minor versions.

src/integrationTest/java/org/opensearch/security/DefaultConfigurationTests.java
```diff
@@ -64,15 +67,79 @@ public static void cleanConfigurationDirectory() throws IOException {
...
+                "]").assertStatusCode(200);
+
+            final var upgradeCheckAfterChanges = client.get("_plugins/_security/api/_upgrade_check");
+            upgradeCheckAfterChanges.assertStatusCode(200);
+            assertThat(
+                upgradeCheckAfterChanges.getTextArrayFromJsonBody("/upgradeActions/roles/add"),
+                equalTo(List.of("flow_framework_full_access"))
+            );
+            assertThat(
+                upgradeCheckAfterChanges.getTextArrayFromJsonBody("/upgradeActions/roles/modify"),
```
cwperks, line 125: Ignore the [comment above](https://github.com/opensearch-project/security/pull/4102/files#r1524925380), I see that this test cases encompasses delete and modify.

cwperks (commented):
Thank you @peternied. This PR looks good to me. I left one minor comment on the integration test about strengthening the assertions post upgrade.

src/integrationTest/resources/roles.yml
```diff
@@ -17,3 +17,21 @@ user_limited-user__limited-role:
...
+    - 'cluster:admin/opensearch/flow_framework/*'
+    - 'cluster_monitor'
+  index_permissions:
+    - index_patterns:
+        - '*'
+      allowed_actions:
+        - 'indices:admin/aliases/get'
+        - 'indices:admin/mappings/get'
+        - 'indices_monitor'
+flow_framework_read_access:
```
scrawfor99, line 31: How come we are are modifying this role here? I am not sure how it connects to the rest of the changes outside of being used in the test cases above.

src/test/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiActionUnitTest.java
scrawfor99, line 159: Does "OperationBash" mean something in this context? I don't see any connection to shells below unless I am missing something.

scrawfor99 (approved):
Overall looks good. Just two minor comments

src/integrationTest/resources/roles.yml
peternied: The roles.yml used in this test case are effectively the 'defaults' - in this case I added these two roles they are part of the distribution defaults - without the reserved keyword so they could be altered using conventional REST API json content.

src/main/java/org/opensearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -336,7 +339,7 @@ final void saveOrUpdateConfiguration(
         final SecurityDynamicConfiguration<?> configuration,
         final OnSucessActionListener<IndexResponse> onSucessActionListener
     ) {
-        saveAndUpdateConfigs(securityApiDependencies.securityIndexName(), client, getConfigType(), configuration, onSucessActionListener);
+        saveAndUpdateConfigsAsync(securityApiDependencies, client, getConfigType(), configuration, onSucessActionListener);
```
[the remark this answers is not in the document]
peternied: Yes - the synchronous version is used for the upgrade process, see ConfigUpgradeApiAction.java:142

src/main/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiAction.java
```diff
@@ -0,0 +1,395 @@
...
+public class ConfigUpgradeApiAction extends AbstractApiAction {
+
+    private final static Logger LOGGER = LogManager.getLogger(ConfigUpgradeApiAction.class);
+
+    private final static Set<CType> SUPPORTED_CTYPES = ImmutableSet.of(CType.ROLES);
+
+    private final static String REQUEST_PARAM_CONFIGS_KEY = "configs";
+
+    private static final List<Route> routes = addRoutesPrefix(
+        ImmutableList.of(new Route(Method.GET, "/_upgrade_check"), new Route(Method.POST, "/_upgrade_perform"))
```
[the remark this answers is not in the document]
peternied: I don't see AbstractApiAction using NamedRoutes.  Is there an issue that these should be upgraded?

src/test/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiActionUnitTest.java
peternied: In this unit test any CType is viable - in the future we could support upgrading of any config types in this way.

src/main/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiAction.java
peternied: Could you add comments to specific methods that you think are unclear? I can add comments to them but I'd rather not generically fill them in.

src/test/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiActionUnitTest.java
peternied: I meant bash as in a party or large gathering. as this test is is following a very similar pattern to setup/act/verify different conditions of the differences produced.  Do you have an alternative name suggestion that would work better?

src/main/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiAction.java
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    private JsonNode filterRemoveOperations(final JsonNode diff) {
+        final ArrayNode filteredDiff = JsonNodeFactory.instance.arrayNode();
+        diff.forEach(node -> {
+            if (!isRemoveOperation(node)) {
+                filteredDiff.add(node);
+                return;
+            } else {
+                if (!hasRootLevelPath(node)) {
```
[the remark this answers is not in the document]
peternied: Using `JsonDiff` generates a json patch request that aligns the json source against the target.  The way these are representing is by a number of operations, add, remove, etc...  If there is a removal operation at the 'top-level' this means that its removing an existing entry - like deleting a role that was added in the live configuration.  Removal operations that are in lower level nodes, such as add/deleting an item in an array that is part of an existing role.  

Get more context on patch operations here: https://stackoverflow.com/a/22820994

cwperks (dismissed):
Thank you for adding an additional assertion. This looks good to me.

src/main/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiAction.java
```diff
@@ -0,0 +1,395 @@
...
+        super(Endpoint.CONFIG, clusterService, threadPool, securityApiDependencies);
+        this.requestHandlersBuilder.configureRequestHandlers(rhb -> {
+            rhb.allMethodsNotImplemented().add(Method.GET, this::canUpgrade).add(Method.POST, this::performUpgrade);
+        });
+    }
+
+    void canUpgrade(final RestChannel channel, final RestRequest request, final Client client) throws IOException {
+        getAndValidateConfigurationsToUpgrade(request).map(this::configurationDifferences).valid(differencesList -> {
+            final var allConfigItemChanges = differencesList.stream()
+                .map(kvp -> new ConfigItemChanges(kvp.v1(), kvp.v2()))
```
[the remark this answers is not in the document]
peternied: There isn't a way to know if something is 'not needed' so removal of existing items is not allowed with this version.  Maybe this could be supported in future versions if there is a good case for it.

As written, it will always be an additive process, see the unit test `testConfigurationDifferences_OperationBash` for verification of this behavior in many cases.

src/test/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiActionUnitTest.java
peternied: I'm going to keep this as is, happy to make a small follow up PR for improvements in this space

src/main/java/org/opensearch/security/dlic/rest/api/ConfigUpgradeApiAction.java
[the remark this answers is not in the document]
peternied: @DarshitChanpura I'm going to resolve this thread - please open an issue if you believe there is work that should be done in this space.

peternied:
@DarshitChanpura @cwperks @scrawfor99 I've resolved all comment threads, could you take another look?

cwperks (approved):

DarshitChanpura (approved):

Issue #4118, opened 2024-03-15
[FEATURE] Switch to built-in security transports from core


reta:
**Is your feature request related to a problem?**

As part of [1], the responsibility to provide the secure transports moves to `NetworkPlugin`, whereas the `Plugin` has been extended to provide `SecureSettingsFactory`.

[1] https://github.com/opensearch-project/OpenSearch/issues/12256

**What solution would you like?**
The `security` plugin does not need to provide the  secure transports anymore but `SecureSettingsFactory` so the core transport modules will be able to configure those.

- [ ] Add `SecureSettingsFactory` 
- [ ] Remove SecuritySSLNettyHttpServerTransport  & SecuritySSLNettyTransport
- [ ] Preserve old SSL related settings 
- [ ] Update references in core with permalinks to `2.13.0` branch 

**What alternatives have you considered?**
Keep everything in `security` plugin

**Do you have any additional context?**
See please https://github.com/opensearch-project/OpenSearch/issues/12256

cwperks:
[Triage] Thank you for filing the issue and opening a companion PR in the security plugin to show how the new secure transport modules in core can be configured from the security plugin.

PR #4119, opened 2024-03-15
Switch to built-in security transports from core

reta:
### Description
The security plugin does not need to provide the secure transports anymore but SecureSettingsFactory so the core transport modules will be able to configure those.

### Issues Resolved
Closes https://github.com/opensearch-project/security/issues/4118

### Testing

### Check List


reta:
The work is not done :) but this is a first step towards removing secure transport part from the plugin, there are a few gaps that will be covered by https://github.com/opensearch-project/OpenSearch/issues/12903 shortly

src/main/java/org/opensearch/security/ssl/OpenSearchSecuritySSLPlugin.java
```diff
@@ -664,4 +666,29 @@ public List<String> getSettingsFilter() {
...
+    protected Settings migrateSettings(Settings settings) {
+        final Settings.Builder builder = Settings.builder().put(settings);
+
+        builder.remove(NetworkModule.TRANSPORT_SSL_DUAL_MODE_ENABLED_KEY);
+        builder.remove(NetworkModule.TRANSPORT_SSL_ENFORCE_HOSTNAME_VERIFICATION_RESOLVE_HOST_NAME_KEY);
+        builder.remove(NetworkModule.TRANSPORT_SSL_ENFORCE_HOSTNAME_VERIFICATION_KEY);
+
+        builder.put(NetworkModule.TRANSPORT_SSL_DUAL_MODE_ENABLED_KEY, SecuritySettings.SSL_DUAL_MODE_SETTING.get(settings));
+        builder.put(
+            NetworkModule.TRANSPORT_SSL_ENFORCE_HOSTNAME_VERIFICATION_RESOLVE_HOST_NAME_KEY,
```
peternied, line 684: Isn't this setting from the network module the 'new setting' so it should take precedent over the existing setting from the security plugin?  Alternatively only allowing this setting defined in one place seems like a way to avoid unclear defaults, what do you think?

reta: > Isn't this setting from the network module the 'new setting' so it should take precedent over the existing setting from the security plugin?

Very valid question, at this moment the new setting is not used - the configuration is purely driven by security plugin settings. However the end solution I see would be this: the `security` plugin uses the secure transport as is (this method would be gone), but the we may need to add fallbacks for  `NetworkModule.*` to use `security` plugin settings (I am working towards that in https://github.com/opensearch-project/OpenSearch/issues/12903)

src/main/java/org/opensearch/security/http/SecurityHttpServerTransport.java
```diff
@@ -34,19 +34,21 @@
...
 import org.opensearch.security.ssl.http.netty.ValidatingDispatcher;
 import org.opensearch.telemetry.tracing.Tracer;
 import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.SharedGroupFactory;
 
+import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.util.AttributeKey;
 
-public class SecurityHttpServerTransport extends SecuritySSLNettyHttpServerTransport {
+public class SecurityHttpServerTransport extends SecureNetty4HttpServerTransport {
```
cwperks, line 51: nit: Rename this class to `SecureHttpServerTransport` for consistency?

src/main/java/org/opensearch/security/http/SecurityNonSslHttpServerTransport.java
```diff
@@ -44,7 +46,7 @@
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 
-public class SecurityNonSslHttpServerTransport extends Netty4HttpServerTransport {
+public class SecurityNonSslHttpServerTransport extends SecureNetty4HttpServerTransport {
```
cwperks, line 49: nit: wdyt about renaming this to `NonSslHttpServerTransport`

cwperks (commented):
Thank you @reta! The changes look good to me, I just left 2 small cosmetic comments.

I would be interested in seeing bwc tests for switching to secure transport in core while configured from the security plugin and eventually switching all configuration to core. 

Have you tested this in a mixed cluster where some nodes are using the transport from core and some from the security plugin?

reta:
> Have you tested this in a mixed cluster where some nodes are using the transport from core and some from the security plugin?

Thanks @cwperks ! No, I have not tested it yet but I test migrations for sure, thank you.

src/main/java/org/opensearch/security/ssl/OpenSearchSecuritySSLPlugin.java
peternied: What do you think about throwing an error if the `NetworkModule` is set in the mean time, and we circle back on this when considering pushing more settings into core?  I'd hate for an upgrade to suddenly cause a behavior change after a feature came out of an incubation cycle.

reta: Sure, I will work through it, thanks!

reta: @peternied addressed, curious what do you think, thank you!

peternied: Nice - could you also include a unit test to validate?

reta: @peternied done! thank you!

peternied (approved):

reta:
@cwperks mind please retaking a look? thank you!

cwperks (approved):

cwperks:
Looks good to me!

PR #4151, opened 2024-03-20
Refactor and update existing ml roles

b4sjoo:
### Description
This PR refactor and update exisiting ml roles

### Check List


config/roles.yml
```diff
@@ -272,7 +272,7 @@ cross_cluster_search_remote_full_access:
         - 'indices:data/read/search'
 
 # Allow users to operate query assistant
-ml_query_assistant_access:
+query_assistant_access:
   reserved: true
   cluster_permissions:
     - 'cluster:admin/opensearch/ml/execute'
```
lezzago, line 278: Can we add this permission here: `cluster:admin/opensearch/ml/config/get`
It is needed to fetch the agent id. Also will the `cluster:admin/opensearch/ml/execute` provide the ability to execute agents?

jngz-es: We think the purpose of adding this role allows user-admin to grand an end-user to have permissions to use query assistant, not let end-user to create a new query assistant. We don't think the agent related permission needed.
```
cluster:admin/opensearch/ml/agents/register
cluster:admin/opensearch/ml/agents/get
cluster:admin/opensearch/ml/agents/search
cluster:admin/opensearch/ml/agents/delete
```
@b4sjoo , did you verify with this role, a user can use query assistant?

lezzago: I am also not sure if we need these permissions as there shouldn't be any memory stored for each natural language to query request.
```
cluster:admin/opensearch/ml/memory/conversation/create
cluster:admin/opensearch/ml/memory/interaction/create
```

I believe only the execute agent and get config apis are used for the query assistant feature: https://github.com/opensearch-project/dashboards-observability/blob/2.x/server/routes/query_assist/utils/agents.ts

config/roles.yml
```diff
@@ -284,13 +284,26 @@ ml_query_assistant_access:
 ml_read_access:
   reserved: true
   cluster_permissions:
+    - 'cluster:admin/opensearch/ml/config/get'
     - 'cluster:admin/opensearch/ml/connectors/get'
     - 'cluster:admin/opensearch/ml/connectors/search'
+    - 'cluster:admin/opensearch/ml/controllers/get'
```
ylwu-amzn, line 290: Don't forget to configure in security dashboard plugin https://github.com/opensearch-project/security-dashboards-plugin/pull/1182/files

config/roles.yml
```diff
@@ -284,13 +284,26 @@ ml_query_assistant_access:
 ml_read_access:
```
ylwu-amzn, line 284: Will it cause any BWC issue if user upgrade to 2.13 from an old cluster?

config/roles.yml
b4sjoo: See https://github.com/opensearch-project/security-dashboards-plugin/pull/1838

config/roles.yml
b4sjoo: Will double check this with security team

cwperks:
@b4sjoo @ylwu-amzn @lezzago 

If this PR and associated PR in [security-dashboards-plugin](https://github.com/opensearch-project/security-dashboards-plugin/pull/1838) are intended for 2.13 please mark this approved to signal to the security maintainers that it is good to go and also get signoff from @bbarani for inclusion in the release as this is post code freeze.

config/roles.yml
cwperks: It would cause bwc issues if it were a change in the role name. It is ok to change the permissions in a role. Cluster admins can either use securityadmin or a [new security api being introduced in 2.13](https://github.com/opensearch-project/documentation-website/pull/6634) to source changes to default roles into the security index.

cwperks: FYI changes to the role name are problematic because the role name is used in [`roles_mappings.yml`](https://opensearch.org/docs/latest/security/configuration/yaml/#roles_mappingyml) to map users to a role. If the role name changes, but the corresponding roles_mapping entry does not change, then those users would effectively lose access to functionality that they were able to previously use.

cwperks: Similarly, some plugins utilize roles injection for scheduled jobs to do permissions check on actions for scheduled jobs. If a role name changes, there is a risk that there are orphaned references to roles in disparate system indexes across the plugin ecosystem.

b4sjoo: Thanks for answering!

config/roles.yml
b4sjoo: Minimal set was found through e2e test on test os domain:
```
    "cluster:admin/opensearch/ml/execute",
    "cluster:admin/opensearch/ml/memory/conversation/create",
    "cluster:admin/opensearch/ml/memory/interaction/create",
    "cluster:admin/opensearch/ml/memory/interaction/update",
    "cluster:admin/opensearch/ml/predict",
    "cluster:monitor/state",
    "cluster:monitor/health"
```

config/roles.yml
```diff
@@ -272,25 +272,41 @@ cross_cluster_search_remote_full_access:
         - 'indices:data/read/search'
 
 # Allow users to operate query assistant
-ml_query_assistant_access:
+query_assistant_access:
```
peternied, line 275: The rename of this item can create some confusion.  For new clusters they will have `query_assistant_access`, for older clusters, they will not get this new name/permissions.  So you can have clusters running 2.13 that have either `ml_query_assistant_access` or `query_assistant_access`, even with the new upgrade API [1] admins might have trouble knowing when to use one or the other.

I'd recommend keeping the old name, if the new name is important, duplicate the role with both names to avoid misalignment.  Changing from `ml_query_assistant_access` -> `query_assistant_access` is a manual change for admins.

- [1] https://github.com/opensearch-project/documentation-website/pull/6634

peternied (changes requested):

config/roles.yml
b4sjoo: Hi Peter, but this role has not been released yet, will that matter?

lezzago (approved):

b4sjoo:
@peternied @cwperks Hi security team, just want to let you know we are good to ship both this PR and the [dashboard](https://github.com/opensearch-project/security-dashboards-plugin/pull/1838) one, I am going to notify @bbarani.  In the meantime, you can either merge this one first and wait @bbarani to sign-off the 2.13 backport, or wait until his sign-off to merge it to mainline and backport.

peternied (approved):

cwperks (approved):

ylwu-amzn (approved):

cwperks:
FYI This has already been [merged to 2.13](https://github.com/opensearch-project/security/pull/4157), this will get merged to main and 2.x shortly but the merge to 2.13 was done first.

PR #4152, opened 2024-03-21
Adds saml auth header to differentiate saml requests and prevents auto login as anonymous user when basic authentication fails

DarshitChanpura:
### Description

This PR fixes a bug where SAML and Anonymous auth cannot co-exist. At present, enabling SAML along with anonymous auth causes SAML Login to fail before it hits the IdP because both  SAML and anonymous auth are wired to not have credentials on the first login attempt, causing the authn check against the auth domains to be skipped (see [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L287-L291)). This, eventually, results in setting the default user as anonymous (see [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L389)) since anonymous auth is enabled.

This PR also resolves another bug involving anonymous auth where a user would be automatically logged in as anonymous user upon failed basic authentication if anonymous auth is enabled.

* Category: Bug fix
    * To allow SAML auth to work when anonymous auth is enabled
    * To prevent auto login as anonymous user upon failed basic authentication
 

Security-dashboards companion PR: https://github.com/opensearch-project/security-dashboards-plugin/pull/1839

### Issues Resolved
- Related https://github.com/opensearch-project/security-dashboards-plugin/issues/1731

### Testing
- Automated

### Check List


src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
```diff
@@ -117,7 +117,8 @@ public class DoNotFailOnForbiddenTests {
             "indices:data/read/msearch",
             "indices:data/read/scroll",
             "cluster:monitor/state",
-            "cluster:monitor/health"
+            "cluster:monitor/health",
+            "cluster:monitor/term"
```
DarshitChanpura, line 121: needed to address integration test failures caused by this commit in core: https://github.com/opensearch-project/OpenSearch/commit/7ad3017

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -415,6 +404,22 @@ public boolean authenticate(final SecurityRequestChannel request) {
                 }
             }
 
+            if (authCredentials == null
+                && anonymousAuthEnabled
+                && isRequestForAnonymousLogin(request.params())
+                && !checkIfRequestContainsBasicAuthHeader(request.getHeaders())) {
```
peternied, line 410: I believe some authentication providers can provider information by the query parameter.  IMO I would rather check all authentication provides first, then if none have found the user, then only if `anonymousAuthEnabled` swallow the 401s.

Alternatively, if there was an opt-in header `X-Authorization: Anonymous`, then that would be a viable signal IMO.  What do you think?

DarshitChanpura: > I believe some authentication providers can provider information by the query parameter. IMO I would rather check all authentication provides first, then if none have found the user, then only if anonymousAuthEnabled swallow the 401s.

This is how it behaves today. It loops over all auth domains, and if authCredentials are still null, even after challenging authenticator check, then it assumes anonymous role if no authorization header was present in the first place.

> Alternatively, if there was an opt-in header `X-Authorization: Anonymous`, then that would be a viable signal IMO. What do you think?

I've tried with something similar but the requests failed when a call was made to get dashboards configuration via `GET /.kibana/config`. The reason was that the authorization header could not be passed from `plugin/security/authinfo` to the subsequent call to get kibana config from the dashboard as this happens in a one flow after clicking `Log in as Anonymous`.  https://github.com/opensearch-project/security/pull/4097#issuecomment-2004657061.

IMO this approach addresses both bugs with minimalistic change. Are there any concerns with this approach?

peternied: The check `headers.containsKey(HttpHeaders.AUTHORIZATION)` is not authoritative to indicate that the caller isn't attempting to authenticate.  Query parameters can be used or alternative headers - this is not a reliable with existing auth providers and creates a sharp edge where there could be unexpected behavior with future authentication providers such as SigV4 for AWS' ecosystem or others.

I'm recommend working backwards from if there is a preferred authentication mechanism how can anonymous authentication also be allowed, where existing behavior around receiving 401s needs to be preserved.

IMO there is _no_ way to do this correctly without the client performing explicit negotiation to include a header.

DarshitChanpura: This case only concerns Basic Auth. I'm trying to resolve: https://github.com/opensearch-project/security-dashboards-plugin/issues/1840 where upon passing wrong credentials user was automatically logged in as anonymous. For other types of authorization, like SigV4, if authn fails it won't redirect to opensearch, if authn succeeds then the `authCredentials == null` check will fail and anonymous login check will not be reached.

src/integrationTest/java/org/opensearch/security/http/BasicWithAnonymousAuthTests.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /** No automatic login post anonymous auth request **/
+    @Test
+    public void shouldRespondWith401WhenUserDoesNotExist() {
+        try (TestRestClient client = cluster.getRestClient(NOT_EXISTING_USER, INVALID_PASSWORD)) {
+            HttpResponse response = client.getAuthInfo();
+
+            assertThat(response, is(notNullValue()));
+            response.assertStatusCode(SC_UNAUTHORIZED);
+        }
```
derek-ho, line 59: Would these tests have failed prior to your change? Just curious?

src/integrationTest/java/org/opensearch/security/http/BasicWithAnonymousAuthTests.java
```diff
@@ -0,0 +1,113 @@
...
+
+            HttpResponse response2 = client.getAuthInfo(Map.of("auth_request_type", "anonymous"));
+
+            assertThat(response2, is(notNullValue()));
+            response2.assertStatusCode(SC_OK);
+        }
+    }
+
+    @Test
+    public void testShouldAutomaticallyLoginAsAnonymousIfRequestIsNonAnonymousLogin() {
```
derek-ho, line 100: do you mean shouldNot here?

src/main/java/org/opensearch/security/auth/BackendRegistry.java
derek-ho: @peternied I am trying to understand your perspective here, what is the difference between checking for a `auth_request_type` header, and a `X-Authorization` header? I do agree that the logic should be that anonymous auth should be checked last if possible.

peternied: I just re-read the code - `isRequestForAnonymousLogin(...)` does I was looking sorry for the confusion.

peternied (approved):

src/main/java/org/opensearch/security/auth/BackendRegistry.java
cwperks, line 410: nit: This can be changed to `checkIfRequestContainsAuthHeader`

For proxy requests, JWT and client cert they won't send the credentials through the Authorization header so this is not a universal check to see if the request does not contain credentials

src/integrationTest/java/org/opensearch/security/http/BasicWithAnonymousAuthTests.java
DarshitChanpura: yes

src/main/java/org/opensearch/security/auth/BackendRegistry.java
DarshitChanpura: Since the check is only for basic auth, i named it as such. For other types of auth this method should not be called unless `isRequestForAnonymousLogin(request.params())` succeeds.

src/integrationTest/java/org/opensearch/security/http/BasicWithAnonymousAuthTests.java
DarshitChanpura: updated.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
cwperks: If the request is explicitly for anonymous login (i.e. Has a url param called `auth_request_type` with a value of `anonymous`), then do you think its necessary to check if the request has credentials or is credential-less?

In the scenario where anonymous auth is explicitly requested, should that ignore if credentials were also provided on the request?

DarshitChanpura: Anonymous requests will not contain a basic auth header, and  hence this check will return false which in the if condition check will be negated to true.

DarshitChanpura: I see where the confusion originated from. `isRequestForAnonymousLogin` returns true if no `auth_request_type` param was passed. To address this I've updated the companion front-end PR to avoid confusion here.

cwperks (dismissed):

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -432,6 +433,20 @@ public boolean authenticate(final SecurityRequestChannel request) {
...
 
+    /**
+     * Checks if incoming auth request is from an anonymous user
+     * Defaults all requests to yes, to allow anonymous authentication to succeed
+     * @param params the query parameters passed in this request
+     * @return false only if an explicit `auth_request_type` param is supplied and its value is not anonymous,
+     * otherwise returns true
+     */
+    private boolean isRequestForAnonymousLogin(Map<String, String> params) {
+        if (params.containsKey("auth_request_type")) {
```
peternied, line 444: This is a new public interface - we wouldn't be able to change it after releasing the feature, lets be careful about how we name it, as non-dashboard clients can use this behavior too.

- When using non standard http headers they are preceded with an X
- Header keys are kebab cased
- the word request is implied making it redundant.

What do you think about `X-Auth-Type`?

peternied (changes requested):

src/main/java/org/opensearch/security/auth/BackendRegistry.java
cwperks: @peternied FYI its not required to pass this url param and any request is assumed anonymous unless its explicitly said to not be.

i.e. 

- `curl -XGET https://localhost:9200 -k` is assumed to be anonymous and will give 200 response
- `curl -XGET https://localhost:9200?auth_request_type=saml -k` is explicitly passed to tell the backend auth registry to not allow anonymous login with the request and will give back a 401 Forbidden response if the request is not properly authed.

DarshitChanpura: it is no longer a header. It is a request param. I agree that request is redundant here, so i have rename it to `auth_type`. My understanding is that request params are snake_case. Let me know if otherwise.

peternied (dismissed):

src/integrationTest/java/org/opensearch/security/http/BasicWithAnonymousAuthTests.java
```diff
@@ -0,0 +1,113 @@
...
+        try (TestRestClient client = cluster.getRestClient(NOT_EXISTING_USER, TEST_USER.getPassword())) {
+            HttpResponse response = client.getAuthInfo();
+
+            assertThat(response, is(notNullValue()));
+            response.assertStatusCode(SC_UNAUTHORIZED);
+        }
+    }
+
+    @Test
+    public void shouldRespondWith401WhenPasswordIsIncorrect() {
```
scrawfor99, line 73: nit: To be consistent with naming I think you would want this and the above to be "testShould..."

scrawfor99 (dismissed):
Looks good to me!

DarshitChanpura:
All failures caused by:
```
java.lang.NoSuchMethodError: 'void org.opensearch.transport.TcpTransport.inboundMessage(org.opensearch.transport.TcpChannel, org.opensearch.transport.ProtocolInboundMessage)'
```

Seems like an upstream change: https://github.com/opensearch-project/OpenSearch/pull/12967.

scrawfor99 (approved):
Looks good to me

cwperks (approved):

PR #4170, opened 2024-03-26
Update ci.yml to use upload-artifact@v4 for better build time

nicheexpress:
upload-artifact@v4 is 10x faster (according to GitHub), It has significantly reduced build time

https://github.blog/2024-02-12-get-started-with-v4-of-github-actions-artifacts/

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -67,7 +67,7 @@ jobs:
         asset_content_type: application/zip
 
     - name: Upload Workflow Artifacts
-      uses: actions/upload-artifact@v1
+      uses: actions/upload-artifact@v4
       with:
         name: artifacts
         path: artifacts/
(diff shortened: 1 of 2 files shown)


peternied (changes requested):
Thanks for the contribution, without being paired with the download artifact of the same version this will cause failures.  Please look into a previous pull request [1] that updated this into main.

I recommend creating new pull request following the backport workflow [2] to get this functionality into this 1.3 branch.

- [1] https://github.com/opensearch-project/security/pull/3893
- [2] https://github.com/opensearch-project/security/pull/3893#issuecomment-1867955322

nicheexpress:
Thanks @peternied, I believe that download-artifact does not exists in 1.3 it was introduced later on

peternied:
It was added later, in this repository we use backporting workflows [1] to move code from the main branch out to current major versions and minor versions.  If you'd like to get this change into this branch I'd recommend following that process.

Note; since this branch is maintained for security updates.  We've got a high barrier to entry for changes to it - I'm on the fence if your proposed change meets that bar.  If you are looking to make improvements with more impact, I would suggest looking at the `main` branch.

- [1] https://github.com/opensearch-project/.github/blob/main/RELEASING.md#backporting

nicheexpress:
Thanks @peternied I've learned something new :)
I've opened a new PR by following the guide above

PR #4179, opened 2024-03-27
[FEATURE] Improve built-in secure transports support

reta:
### Description
Incorporate further improvements from https://github.com/opensearch-project/OpenSearch/pull/12907

### Issues Resolved
Part of https://github.com/opensearch-project/OpenSearch/issues/12903

### Testing

### Check List


reta:
Depends on https://github.com/opensearch-project/OpenSearch/pull/12907

src/main/java/org/opensearch/security/ssl/OpenSearchSecuritySSLPlugin.java
```diff
@@ -676,7 +678,7 @@ public List<String> getSettingsFilter() {
 
     @Override
     public Optional<SecureSettingsFactory> getSecureSettingFactory(Settings settings) {
-        return Optional.of(new OpenSearchSecureSettingsFactory(settings, sks, NOOP_SSL_EXCEPTION_HANDLER));
+        return Optional.of(new OpenSearchSecureSettingsFactory(threadPool, sks, NOOP_SSL_EXCEPTION_HANDLER, securityRestHandler));
```
scrawfor99, line 681: Can you quickly explain the movement of the threadpool to being passed here? I am not sure I follow the passing of it to the settings factory. I know you now create a Netty verifier for the HTTP handler but it seems like this was not done before so I am trying to follow how this has changed the factory class.

src/test/java/org/opensearch/security/ssl/OpenSearchSecuritySSLPluginTest.java
scrawfor99, line 1: Is there anything more we can do for this testing? 

I know this was how the class was before, but I just feel like this change modifies a fair amount of logic for the only difference to be some naming conventions in the test. 

Now that you split the factory to produce separate clients is is possible to test the creation of the appropriate one etc.? 

For instance you mentioned on the original issue that there are some providers that only create HTTP or transport etc. so can we verify that somehow? Perhaps that would be better placed in the core change?

scrawfor99 (approved):
A couple comments but clean code as always. Thanks @reta

src/main/java/org/opensearch/security/ssl/OpenSearchSecuritySSLPlugin.java
reta: Yes, sure, the Netty4 header verifier needs `threadpool` as an argument. Historically we have 2 plugins in security (I am not sure why but it is what it is), one inherent from other. So I moved `threadpool` to base plugin since it is needed by both of them now.

src/test/java/org/opensearch/security/ssl/OpenSearchSecuritySSLPluginTest.java
reta: > Now that you split the factory to produce separate clients is is possible to test the creation of the appropriate one etc.?

The full transports are tested in `netty4-transport`:  
- https://github.com/opensearch-project/OpenSearch/blob/main/modules/transport-netty4/src/test/java/org/opensearch/transport/netty4/ssl/SimpleSecureNetty4TransportTests.java
- https://github.com/opensearch-project/OpenSearch/blob/main/modules/transport-netty4/src/test/java/org/opensearch/http/netty4/ssl/SecureNetty4HttpServerTransportTests.java

The plugin tests make sure that:
 - secure transports are configured
 - secure transports are instantiateable

But I have a gap with verifiers and decoders here - I think having these tests will be useful, thank you

reta: Tests added, thanks @scrawfor99 !

reta:
The BWC / plugins check fail, we need distributions being published

src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java
```diff
@@ -11,63 +11,124 @@
...
+                    }
+
+                    @SuppressWarnings("unchecked")
+                    @Override
+                    public <C> Optional<C> create(Settings settings, HttpServerTransport transport, Class<C> adapterClass) {
+                        if (transport instanceof SecureNetty4HttpServerTransport
+                            && ChannelInboundHandlerAdapter.class.isAssignableFrom(adapterClass)) {
+                            return Optional.of((C) new Netty4ConditionalDecompressor());
+                        } else {
+                            return Optional.empty();
```
peternied, line 97: What do you think of throw an exception if we don't have a matching transport instance?

Thinking allowed about the possible states; If ssl disabled off, I don't believe this will be called. If we are in ssl only mode, I could go either way on checking the transport functionality, but would error on the side of throwing an exception, and in full feature ssl we only support Netty based transports.

reta: > What do you think of throw an exception if we don't have a matching transport instance?

So this method (or better to say provider) will be called for any supported transport: Netty4, Reactor-Netty4, Nio. At the moment, only Netty4 is supported, and for others - we do nothing (`Optional.empty()`), we just don't have this implemented. So the next provider could be picked. 

> Thinking allowed about the possible states; If ssl disabled off, I don't believe this will be called

Correct

peternied: Just looked back into the design in OpenSearch.  I suppose that makes sense, its just decoupled a little bit more than I would expect.

As the security plugin I would want to be sure that the secure transport system has the correct relationship between REQUEST_HEADER_VERIFIER and REQUEST_DECOMPRESSOR in place.

reta: > As the security plugin I would want to be sure that the secure transport system has the correct relationship between REQUEST_HEADER_VERIFIER and REQUEST_DECOMPRESSOR in place.

I added test for it in previous round (in response to @scrawfor99 comment). On a general note - if we could somehow formalize that over other transport, we could end up with better solution (I just have not looked into that yet).

reta:
@scrawfor99 may I ask you please to reapprove (new tests were pushed), thank you!

peternied (approved):

reta:
The BWC tests are failing because the change is breaking (on experimental API) and for a clean build, we would need https://github.com/opensearch-project/security/pull/4187 first

willyborankin (approved):

PR #4201, opened 2024-04-02
System index permission grants reading access to documents in the index

### LINKED ISSUE #4188: [BUG] `system:admin/system_index` search results do not include any documents
**What is the bug?**

I've set the permission `system:admin/system_index` on `internal_index_manager` role for reading the custom system index.
The user (binding `internal_index_manager` role) can write documents and the `_search` API works, but it cannot search documents (no hit items).
Only super admin can search documents.
Is this by design? If it was intentional, I wonder the reason.

### Test based reproduction

In SystemIndexPermissionEnabledTests.SystemIndexPermissionEnabledTests should expect 1 document to be returned, edit the call to validateSearchResponse second parameter from 0 -> 1.

https://github.com/opensearch-project/security/blob/4e8297aa4253c675526a803651a9214fe4e02962/src/test/java/org/opensearch/security/system_indices/SystemIndexPermissionEnabledTests.java#L72-L84

---
#### config

```yml
plugins.security.system_indices.enabled: true
plugins.security.system_indices.permission.enabled: true
plugins.security.system_indices.indices: [".my-system-index"]
```

#### role

```yml
internal_index_manager:
  reserved: true
  hidden: true
  description: ""
  index_permissions:
    - index_patterns:
        - ".my-system-index"
      allowed_actions:
        - "*"
        - "system:admin/system_index"
```

#### user

```yml
manager:
  hash: "{hash}"
  reserved: true
  opendistro_security_roles:
    - "internal_index_manager"
  description: ""
```

#### CAT indices

```
health status index                  uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   .my-system-index U-HWdSsKTi2oqcrVNgPqfQ   1   1          1            0     10.4kb         10.4kb
```

#### SEARCH result

```
GET https://{ip}:{port}/.my-system-index/_search
Authorization: Basic {user_token}
```
```json
{
  "took": 3,
  "timed_out": false,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 0,
      "relation": "eq"
    },
    "max_score": null,
    "hits": [] // cannot read documents. no hit item.
  }
}
```

#### expected SEARCH result

```json
{
  "took": 4,
  "timed_out": false,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 1,
      "relation": "eq"
    },
    "max_score": 1.0,
    "hits": [
      {
        "_index": ".my-system-index",
        "_id": "1",
        "_score": 1.0,
        "_source": {
          "field": "value"
         }
       }
     ] // want to get hit items of the custom system index
  }
}
```

```
$ cat /etc/os-release
NAME="Red Hat Enterprise Linux"
VERSION="8.7 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Red Hat Enterprise Linux 8.7 (Ootpa)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:8::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/red_hat_enterprise_linux/8/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_BUGZILLA_PRODUCT_VERSION=8.7
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.7"
```
 - Plugins: 2.11.1

llilyy:
### Description

* Category: **Bug Fix**

* Why these changes are required? **User has `system:admin/system_index permission`, but cannot search documents in the custom system index.**

* What is the old behavior before changes and new behavior after changes? **It was modified to not return `EmptyFilterLeafReader` when the system index permission option is enable.**

### Issues Resolved
- Resolves #4188 

### Testing

- Update `SystemIndexPermissionEnabledTests.testSearchAsNormalUser`

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
+++ b/src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
@@ -32,7 +32,7 @@ import java.util.Set;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.lucene.index.DirectoryReader;
-import org.greenrobot.eventbus.Subscribe;
+
 import org.opensearch.common.CheckedFunction;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.util.concurrent.ThreadContext;
@@ -46,6 +46,8 @@ import org.opensearch.security.support.ConfigConstants;
 import org.opensearch.security.support.HeaderHelper;
 import org.opensearch.security.support.WildcardMatcher;
 import org.opensearch.security.user.User;
+
+import org.greenrobot.eventbus.Subscribe;
 
 public class SecurityIndexSearcherWrapper implements CheckedFunction<DirectoryReader, DirectoryReader, IOException> {
 
--- a/src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
+++ b/src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
@@ -64,7 +64,7 @@ public class SecurityIndexSearcherWrapper implements CheckedFunction<DirectoryRe
     private final Boolean systemIndexEnabled;
     private final WildcardMatcher systemIndexMatcher;
 
-    private final boolean isSystemIndexPermissionEnabled;
+    private final Boolean systemIndexPermissionEnabled;
 
     // constructor is called per index, so avoid costly operations here
     public SecurityIndexSearcherWrapper(
@@ -94,7 +94,7 @@ public class SecurityIndexSearcherWrapper implements CheckedFunction<DirectoryRe
         );
         this.systemIndexMatcher = WildcardMatcher.from(settings.getAsList(ConfigConstants.SECURITY_SYSTEM_INDICES_KEY));
 
-        this.isSystemIndexPermissionEnabled = settings.getAsBoolean(
+        this.systemIndexPermissionEnabled = settings.getAsBoolean(
             ConfigConstants.SECURITY_SYSTEM_INDICES_PERMISSIONS_ENABLED_KEY,
             ConfigConstants.SECURITY_SYSTEM_INDICES_PERMISSIONS_DEFAULT
         );
@@ -115,7 +115,7 @@ public class SecurityIndexSearcherWrapper implements CheckedFunction<DirectoryRe
             return new EmptyFilterLeafReader.EmptyDirectoryReader(reader);
         }
 
-        if (systemIndexEnabled && !isSystemIndexPermissionEnabled && isBlockedSystemIndexRequest() && !isAdminDnOrPluginRequest()) {
+        if (systemIndexEnabled && !systemIndexPermissionEnabled && isBlockedSystemIndexRequest() && !isAdminDnOrPluginRequest()) {
             log.warn("search action for {} is not allowed for a non adminDN user", index.getName());
             return new EmptyFilterLeafReader.EmptyDirectoryReader(reader);
         }
--- a/src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
+++ b/src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
@@ -64,6 +64,8 @@ public class SecurityIndexSearcherWrapper implements CheckedFunction<DirectoryRe
     private final Boolean systemIndexEnabled;
     private final WildcardMatcher systemIndexMatcher;
 
+    private final boolean isSystemIndexPermissionEnabled;
+
     // constructor is called per index, so avoid costly operations here
     public SecurityIndexSearcherWrapper(
         final IndexService indexService,
@@ -91,6 +93,11 @@ public class SecurityIndexSearcherWrapper implements CheckedFunction<DirectoryRe
             ConfigConstants.SECURITY_SYSTEM_INDICES_ENABLED_DEFAULT
         );
         this.systemIndexMatcher = WildcardMatcher.from(settings.getAsList(ConfigConstants.SECURITY_SYSTEM_INDICES_KEY));
+
+        this.isSystemIndexPermissionEnabled = settings.getAsBoolean(
+            ConfigConstants.SECURITY_SYSTEM_INDICES_PERMISSIONS_ENABLED_KEY,
+            ConfigConstants.SECURITY_SYSTEM_INDICES_PERMISSIONS_DEFAULT
+        );
     }
 
     @Subscribe
@@ -108,7 +115,7 @@ public class SecurityIndexSearcherWrapper implements CheckedFunction<DirectoryRe
             return new EmptyFilterLeafReader.EmptyDirectoryReader(reader);
         }
 
-        if (systemIndexEnabled && isBlockedSystemIndexRequest() && !isAdminDnOrPluginRequest()) {
+        if (systemIndexEnabled && !isSystemIndexPermissionEnabled && isBlockedSystemIndexRequest() && !isAdminDnOrPluginRequest()) {
             log.warn("search action for {} is not allowed for a non adminDN user", index.getName());
             return new EmptyFilterLeafReader.EmptyDirectoryReader(reader);
         }
--- a/src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
+++ b/src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
@@ -32,7 +32,7 @@ import java.util.Set;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.lucene.index.DirectoryReader;
-
+import org.greenrobot.eventbus.Subscribe;
 import org.opensearch.common.CheckedFunction;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.util.concurrent.ThreadContext;
@@ -46,8 +46,6 @@ import org.opensearch.security.support.ConfigConstants;
 import org.opensearch.security.support.HeaderHelper;
 import org.opensearch.security.support.WildcardMatcher;
 import org.opensearch.security.user.User;
-
-import org.greenrobot.eventbus.Subscribe;
 
 public class SecurityIndexSearcherWrapper implements CheckedFunction<DirectoryReader, DirectoryReader, IOException> {
 
@@ -116,8 +114,7 @@ public class SecurityIndexSearcherWrapper implements CheckedFunction<DirectoryRe
             return new EmptyFilterLeafReader.EmptyDirectoryReader(reader);
         }
 
-        if (systemIndexEnabled && isBlockedSystemIndexRequest() && !isAdminDnOrPluginRequest()
-            && (!systemIndexPermissionEnabled || !isPermittedOnSystemIndex())) {
+        if (systemIndexEnabled && isBlockedSystemIndexRequest() && !isAdminDnOrPluginRequest()) {
             log.warn("search action for {} is not allowed for a non adminDN user", index.getName());
             return new EmptyFilterLeafReader.EmptyDirectoryReader(reader);
         }
@@ -153,6 +150,17 @@ public class SecurityIndexSearcherWrapper implements CheckedFunction<DirectoryRe
     }
 
     protected final boolean isBlockedSystemIndexRequest() {
+        if (systemIndexPermissionEnabled) {
+            final User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+            if (user == null) {
+                // allow request without user from plugin.
+                return systemIndexMatcher.test(index.getName());
+            }
+            final TransportAddress caller = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
+            final Set<String> mappedRoles = evaluator.mapRoles(user, caller);
+            final SecurityRoles securityRoles = evaluator.getSecurityRoles(mappedRoles);
+            return !securityRoles.isPermittedOnSystemIndex(index.getName());
+        }
         return systemIndexMatcher.test(index.getName());
     }
 
@@ -176,14 +184,5 @@ public class SecurityIndexSearcherWrapper implements CheckedFunction<DirectoryRe
             return true;
         }
         return false;
-    }
-
-    private boolean isPermittedOnSystemIndex() {
-        final User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
-        final TransportAddress caller = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
-        final Set<String> mappedRoles = evaluator.mapRoles(user, caller);
-        final SecurityRoles securityRoles = evaluator.getSecurityRoles(mappedRoles);
-
-        return securityRoles.isPermittedOnSystemIndex(index.getName());
     }
 }
--- a/src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
+++ b/src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
@@ -41,6 +41,7 @@ import org.opensearch.core.index.Index;
 import org.opensearch.index.IndexService;
 import org.opensearch.security.privileges.PrivilegesEvaluator;
 import org.opensearch.security.securityconf.ConfigModel;
+import org.opensearch.security.securityconf.SecurityRoles;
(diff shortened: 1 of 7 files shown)


peternied (changes requested):
@llilyy The security's plugin system index should not be directly accessible - even with the system index permission.  Accessing the data in that index is better served through its data access layer [1].

Can you expand on why direct access to this data should be avaliable?


- [1] https://opensearch.org/docs/latest/security/access-control/api/#configuration

llilyy:
@peternied @DarshitChanpura 

Regarding the system index permission feature, I thought that the system index was divided into two layers as shown below.

1. Accessible only by super admin (ex. `.opendistro_security`)
2. Accessible through `system:admin/system_index` permission

If it does not allow direct access to all system indices regardless of `system:admin/system_index` permission,
could you please explain why only write access is possible while read access is not?

The reason to access the system index is as follows.
During in-house opensearch cloud development, an index that is accessible only to system admin users is required.
Usually, `*` is used to give full access to cloud users.
So I want to use a custom system index that cannot be accessed with `*`.

Here is my configuration.

#### yaml
```
plugins.security.system_indices.enabled: true
plugins.security.system_indices.permission.enabled: true
plugins.security.system_indices.indices: [".my-system-index"]
```

#### role
```
internal_manager:
  reserved: true
  hidden: true
  description: ""
  index_permissions:
    - index_patterns:
        - ".my-system-index"
      allowed_actions:
        - "*"
        - "system:admin/system_index"
```

llilyy:
@peternied 
Thanks for the API information. I will test it.

peternied:
@llilyy I see that you make updates to the test cases, I'd recommend undoing the changes to the test case that is checking the security system index, but add more scenario(s) that test the behavior with different another system index.  I believe that was the take away from the request when the bug was initial filed - which on its face seems OK to me.

llilyy:
> https://opensearch.org/docs/latest/security/access-control/api/#configuration

I tested this API. However, API returns a security configuration. Unable to access custom system index.

>  I see that you make updates to the test cases, I'd recommend undoing the changes to the test case that is checking the security system index, but add more scenario(s) that test the behavior with different another system index.

Is the security system index the `.opendistro_security` and `plugins.security.config_index_name` configuration list?
If correct, the part I modified does not access the security system index. It access a custom system index.

There are three types of system indices tested in `SystemIndexPermissionEnabledTests.testSearchAsNormalUser`.

(1) Accessible only by super admin (ex. `.opendistro_security`)
(2) System index with no associated role permissions (ex. `random_system_index`)
(3) System index with associated role permissions - normal_role, normal_user (ex. `.system_index_1`)

In the test, add these indices to the `plugins.security.system_indices.indices` list.
https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/system_indices/AbstractSystemIndicesTests.java#L86

```java
.putList(ConfigConstants.SECURITY_SYSTEM_INDICES_KEY, SYSTEM_INDICES)
```

The roles and roles_mapping are as follows.
https://github.com/opensearch-project/security/blob/main/src/test/resources/system_indices/roles.yml#L16-L25
https://github.com/opensearch-project/security/blob/main/src/test/resources/system_indices/roles_mapping.yml#L16-L24

```yml
normal_role:
  description: "Normal user Role"
  cluster_permissions:
    - '*'
  index_permissions:
    - index_patterns:
        - '.system*'
      allowed_actions:
        - '*'
        - 'system:admin/system_index'
```

```yml
normal_role:
  reserved: false
  hidden: false
  backend_roles: []
  hosts: []
  users:
    - "normal_user"
  and_backend_roles: []
  description: "System index permissions"
```

In `SystemIndexPermissionEnabledTests.testSearchAsNormalUser` I want to read and write to (3).
https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/system_indices/SystemIndexPermissionEnabledTests.java#L71-L84

```java
for (String index : SYSTEM_INDICES) {
    // security index is only accessible by super-admin
    RestHelper.HttpResponse response = restHelper.executePostRequest(index + "/_search", "", normalUserHeader);
    if (index.equals(ACCESSIBLE_ONLY_BY_SUPER_ADMIN) || index.equals(SYSTEM_INDEX_WITH_NO_ASSOCIATED_ROLE_PERMISSIONS)) {
        validateForbiddenResponse(response, "", normalUser); // (1), (2) should not be accessible.
    } else {
        validateSearchResponse(response, 0); // Current: The document in (3) should not be readable.
        // Request: I want to read the document in (3).
        // So I edited it to make it readable.
        // validateSearchResponse(response, 1);
    }
}
```

Please let me know if I misunderstood.

If it does not allow access to system indices of (3) type regardless of `system:admin/system_index` permission, could you please explain why only write access is possible while read access is not?

llilyy:
@peternied 

> https://opensearch.org/docs/latest/security/access-control/api/#configuration

I tested this API. However, API returns a security configuration. Unable to access custom system index.

>  I see that you make updates to the test cases, I'd recommend undoing the changes to the test case that is checking the security system index, but add more scenario(s) that test the behavior with different another system index.

Is the security system index the `.opendistro_security` and `plugins.security.config_index_name` configuration list?
If correct, the part I modified does not access the security system index. It access a custom system index.

There are three types of system indices tested in `SystemIndexPermissionEnabledTests.testSearchAsNormalUser`.

(1) Accessible only by super admin (ex. `.opendistro_security`)
(2) System index with no associated role permissions (ex. `random_system_index`)
(3) System index with associated role permissions - normal_role, normal_user (ex. `.system_index_1`)

In the test, add these indices to the `plugins.security.system_indices.indices` list.
https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/system_indices/AbstractSystemIndicesTests.java#L86

```java
.putList(ConfigConstants.SECURITY_SYSTEM_INDICES_KEY, SYSTEM_INDICES)
```

The roles and roles_mapping are as follows.
https://github.com/opensearch-project/security/blob/main/src/test/resources/system_indices/roles.yml#L16-L25
https://github.com/opensearch-project/security/blob/main/src/test/resources/system_indices/roles_mapping.yml#L16-L24

```yml
normal_role:
  description: "Normal user Role"
  cluster_permissions:
    - '*'
  index_permissions:
    - index_patterns:
        - '.system*'
      allowed_actions:
        - '*'
        - 'system:admin/system_index'
```

```yml
normal_role:
  reserved: false
  hidden: false
  backend_roles: []
  hosts: []
  users:
    - "normal_user"
  and_backend_roles: []
  description: "System index permissions"
```

In `SystemIndexPermissionEnabledTests.testSearchAsNormalUser` I want to read and write to (3).
https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/system_indices/SystemIndexPermissionEnabledTests.java#L71-L84

```java
for (String index : SYSTEM_INDICES) {
    // security index is only accessible by super-admin
    RestHelper.HttpResponse response = restHelper.executePostRequest(index + "/_search", "", normalUserHeader);
    if (index.equals(ACCESSIBLE_ONLY_BY_SUPER_ADMIN) || index.equals(SYSTEM_INDEX_WITH_NO_ASSOCIATED_ROLE_PERMISSIONS)) {
        validateForbiddenResponse(response, "", normalUser); // (1), (2) should not be accessible.
    } else {
        validateSearchResponse(response, 0); // Current: The document in (3) should not be readable.
        // Request: I want to read the document in (3).
        // So I edited it to make it readable.
        // validateSearchResponse(response, 1);
    }
}
```

Please let me know if I misunderstood.

If it does not allow access to system indices of (3) type regardless of `system:admin/system_index` permission, could you please explain why only write access is possible while read access is not?

llilyy:
@peternied 

`.system_index_1` has a document.

```
{ \"foo\": \"bar\" }
```

But result has empty hits. Cannot read the document.

```
"hits":[] // empty hits
```

Is this by design?

peternied:
@llilyy Thanks for the context - that does seem off - want to update the test plumb the fix through?

llilyy:
@peternied Yes, I want. Is there anything else that needs to be modified?

DarshitChanpura (commented):
Thank you for the explanation @llilyy. It seems off that read is not allowed when it should be allowed with system index permission. Could you please ensure that the tests modified are the ones for custom system index, and the access to `.opendistro_security` index is still only allowed for super admin.

llilyy:
@DarshitChanpura 

Here is the test result I edited.

```
Testing index .system_index_1
	Expecting success response for .system_index_1
		Response status code was 200
		Response body:
		{"took":12,"timed_out":false,"_shards":{"total":5,"successful":5,"skipped":0,"failed":0},"hits":{"total":{"value":1,"relation":"eq"},"max_score":1.0,"hits":[{"_index":".system_index_1","_id":"document1","_score":1.0,"_source":{ "foo": "bar" }}]}}
Testing index random_system_index
	Expecting forbidden response for random_system_index
Testing index .opendistro_security
	Expecting forbidden response for .opendistro_security
```

And I added more logics to check system index permission on `SecurityIndexSearcherWrapper.apply`.

src/main/java/org/opensearch/security/securityconf/ConfigModelV6.java
```diff
@@ -542,6 +542,26 @@ public boolean impliesTypePermGlobal(
...
+
+        @Override
+        public boolean isPermittedOnSystemIndex(String indexName) {
+            boolean isPatternMatched = false;
+            boolean isPermitted = false;
+            for (SecurityRole role : roles) {
+                for (IndexPattern ip : role.getIpatterns()) {
+                    WildcardMatcher wildcardMatcher = WildcardMatcher.from(ip.indexPattern);
+                    if (wildcardMatcher.test(indexName)) {
+                        isPatternMatched = true;
```
DarshitChanpura, line 554: This would match against `*` pattern. Correct me if i'm wrong. we do not want that. We want to explicitly match against an system index pattern. For example, `*` should not be matched but `.system_index*` should be matched.

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
```diff
@@ -563,6 +563,24 @@ private boolean containsDlsFlsConfig() {
...
+
+        @Override
+        public boolean isPermittedOnSystemIndex(String indexName) {
+            boolean isPatternMatched = false;
+            boolean isPermitted = false;
+            for (SecurityRole role : roles) {
+                for (IndexPattern ip : role.getIpatterns()) {
+                    WildcardMatcher wildcardMatcher = WildcardMatcher.from(ip.indexPattern);
+                    if (wildcardMatcher.test(indexName)) {
+                        isPatternMatched = true;
```
DarshitChanpura, line 575: same as above.

src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
```diff
@@ -168,4 +177,13 @@ protected final boolean isPermittedOnIndex() {
...
         return false;
     }
+
+    private boolean isPermittedOnSystemIndex() {
+        final User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+        final TransportAddress caller = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
+        final Set<String> mappedRoles = evaluator.mapRoles(user, caller);
+        final SecurityRoles securityRoles = evaluator.getSecurityRoles(mappedRoles);
+
+        return securityRoles.isPermittedOnSystemIndex(index.getName());
```
DarshitChanpura, line 187: is there a way we can leverage `hasExplicitIndexPermission` here?

src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
```diff
@@ -108,7 +116,8 @@ public final DirectoryReader apply(DirectoryReader reader) throws IOException {
             return new EmptyFilterLeafReader.EmptyDirectoryReader(reader);
         }
 
-        if (systemIndexEnabled && isBlockedSystemIndexRequest() && !isAdminDnOrPluginRequest()) {
+        if (systemIndexEnabled && isBlockedSystemIndexRequest() && !isAdminDnOrPluginRequest()
```
scrawfor99, line 119: Nit: What do you think about either combining isPermittedOnSystemIndex with isBlockedSystemIndexRequest or renaming the new check? From a code-reading perspective, I think it will be confusing to have conditions with such similar names.

src/test/java/org/opensearch/security/system_indices/SystemIndexPermissionEnabledTests.java
scrawfor99, line 1: I think we should add more test cases for this change. I understand it may seem like a lot but when dealing with something like system indices, it is probably best to lean towards more exhaustive testing. Right now, I would minimally look for checks against the different combinations of  `(!systemIndexPermissionEnabled || !isPermittedOnSystemIndex()))`. How does that sound?

src/main/java/org/opensearch/security/securityconf/ConfigModelV6.java
```diff
@@ -542,6 +542,26 @@ public boolean impliesTypePermGlobal(
             roles.stream().forEach(p -> ipatterns.addAll(p.getIpatterns()));
             return ConfigModelV6.impliesTypePerm(ipatterns, resolved, user, actions, resolver, cs);
         }
+
+        @Override
+        public boolean isPermittedOnSystemIndex(String indexName) {
```
scrawfor99, line 547: Do we need this both here and in the newer V7? We reference V6 inside V7 and I don't know how this is planning to be used with the legacy framework.

src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
llilyy: I want it. But I cannot find a way to get the parameters `Resolved, IndexNameExpressionResolver` for `hasExplicitIndexPermission`. Could I get `Resolved, IndexNameExpressionResolver` at that location?

src/main/java/org/opensearch/security/securityconf/ConfigModelV6.java
llilyy: I added `isPermittedOnSystemIndex` to verify system index permission through `SecurityRoles`.
Because I extended the `SecurityRoles` Interface, logic implementation was also required in V6.
https://github.com/opensearch-project/security/pull/4201/files/b3170c2a0a4ba9f1a599b79b7cf66bc66dbeecbf#diff-729d96f84a7b2b9ddced57ed4722f88280fe697df668d9fee80a6979e79742c7R100

Is there another way? Could you explain in more detail, please?

src/test/java/org/opensearch/security/system_indices/SystemIndexPermissionEnabledTests.java
llilyy: Ok, I will add more cases. Thx your review.

src/main/java/org/opensearch/security/securityconf/ConfigModelV6.java
llilyy: It seems that `*` should matched `.system_index1234`. I checked documents and codes.

https://opensearch.org/docs/latest/security/access-control/permissions/#system-index-permissions

> The following example shows a role that grants access to all system indexes:
```
index_permissions:
    - index_patterns:
        - '*'
    - allowed_actions:
        - 'system:admin/system_index'
```

src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
llilyy: I'm combined isPermittedOnSystemIndex with isBlockedSystemIndexRequest.

src/main/java/org/opensearch/security/configuration/SecurityIndexSearcherWrapper.java
DarshitChanpura: It might be tricky to get it there as there are no direct references that you can use.

src/test/java/org/opensearch/security/system_indices/SystemIndexPermissionEnabledTests.java
llilyy: I'm combined `isPermittedOnSystemIndex` and `systemIndexPermissionEnabled` with `isBlockedSystemIndexRequest`. So it seems that could be tested with existing test codes.

peternied (approved):

Issue #3745, opened 2023-11-20
[Feature] Add support for audit log writing to data streams


tmanninger:
I am using opensearch 2.11.

My auditlog config:
```
plugins.security.audit.config.index: opensearch-security-auditlog
plugins.security.audit.type: internal_opensearch
```

"opensearch-security-auditlog" is a datastream.

When opensearch is storing auditlog to the datastream, i get the following error:
```
java.lang.IllegalArgumentException: only write ops with an op_type of create are allowed in data streams
```

How can i change the op_type of the auditlogs?

scrawfor99:
[Triage] Hi @tmanninger, thank you for filing this issue. Right now we do not support the data stream API to allow this use case. Let me mark this as a feature request since it will involve adding this support to the audit logging. 

That being said, sounds like we can close this issue when: 

- [ ] Add data stream audit log support
- [ ] Add associated tests
- [ ] Add documentation 

Are complete.

tmanninger:
Any Feedback, any roadmap available for this future request?

scrawfor99:
Hi @tmanninger, I am not aware of any active efforts to provide this support. If you are eager to see this feature implemented the fastest route would likely be for you to open a pull request adding the feature. Feel free to open a PR with the change and I will try to review it as quickly as I can.

tmanninger:
How should i integrate this feature?
Add an option
`plugins.security.audit.config.is_datastrean: true`
or
`plugins.security.audit.config.op_type: create`
?

tmanninger:
@scrawfor99 i created an PR:
https://github.com/opensearch-project/security/pull/4257

I changed to OP_TYPE to create, because audit logs should never be updated.
It's tested and works in my test environment.
Looks this ok?

cwperks:
Released in 2.18.0: https://github.com/opensearch-project/documentation-website/pull/8356

PR #4257, opened 2024-04-16
Support datastreams as an AuditLog Sink

tmanninger:
### Description

This PR allows writing auditlogs to datastreams.

 auditlog type "internal_opensearch" currently not supporting datatstreams. 
`java.lang.IllegalArgumentException: only write ops with an op_type of create are allowed in data streams
`

This PR changes the op_type to "create" for the "internal_opensearch" auditlogs, because audit logs should never be updated.
Datastreams only supporting "create" as op_type.

### Issues Resolved
- https://github.com/opensearch-project/security/issues/3745

### Testing
Tested manually: Security auditlog with and without datastreams

### DIFF
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -1591,6 +1591,46 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
                 )
             );
 
+            // Internal OpenSearch DataStream
+            settings.add(
+                Setting.simpleString(
+                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_NAME,
+                    Property.NodeScope,
+                    Property.Filtered
+                )
+            );
+            settings.add(
+                Setting.boolSetting(
+                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_MANAGE,
+                    true,
+                    Property.NodeScope,
+                    Property.Filtered
+                )
+            );
+            settings.add(
+                Setting.simpleString(
+                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NAME,
+                    Property.NodeScope,
+                    Property.Filtered
+                )
+            );
+            settings.add(
+                Setting.intSetting(
+                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NUMBER_OF_SHARDS,
+                    1,
+                    Property.NodeScope,
+                    Property.Filtered
+                )
+            );
+            settings.add(
+                Setting.intSetting(
+                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NUMBER_OF_REPLICAS,
+                    0,
+                    Property.NodeScope,
+                    Property.Filtered
+                )
+            );
+
             // External OpenSearch
             settings.add(
                 Setting.listSetting(
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -1617,7 +1617,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
             settings.add(
                 Setting.intSetting(
                     ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NUMBER_OF_SHARDS,
-                    1,
+                    3,
                     Property.NodeScope,
                     Property.Filtered
                 )
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -1601,7 +1601,8 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
             );
             settings.add(
                 Setting.boolSetting(
-                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_MANAGE,
+                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX
+                        + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_MANAGE,
                     true,
                     Property.NodeScope,
                     Property.Filtered
@@ -1609,14 +1610,16 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
             );
             settings.add(
                 Setting.simpleString(
-                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NAME,
+                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX
+                        + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NAME,
                     Property.NodeScope,
                     Property.Filtered
                 )
             );
             settings.add(
                 Setting.intSetting(
-                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NUMBER_OF_SHARDS,
+                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX
+                        + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NUMBER_OF_SHARDS,
                     3,
                     Property.NodeScope,
                     Property.Filtered
@@ -1624,7 +1627,8 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
             );
             settings.add(
                 Setting.intSetting(
-                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NUMBER_OF_REPLICAS,
+                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX
+                        + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NUMBER_OF_REPLICAS,
                     0,
                     Property.NodeScope,
                     Property.Filtered
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -1620,7 +1620,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
                 Setting.intSetting(
                     ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX
                         + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NUMBER_OF_SHARDS,
-                    3,
+                    1,
                     Property.NodeScope,
                     Property.Filtered
                 )
--- a/src/main/java/org/opensearch/security/auditlog/sink/InternalOpenSearchDataStreamSink.java
+++ b/src/main/java/org/opensearch/security/auditlog/sink/InternalOpenSearchDataStreamSink.java
@@ -1,0 +1,135 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.auditlog.sink;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.List;
+
+// CS-SUPPRESS-SINGLE: RegexpSingleline https://github.com/opensearch-project/OpenSearch/issues/3663
+import org.opensearch.action.DocWriteRequest;
+import org.opensearch.action.admin.indices.datastream.CreateDataStreamAction;
+import org.opensearch.action.admin.indices.template.put.PutComposableIndexTemplateAction;
+import org.opensearch.action.support.master.AcknowledgedResponse;
+import org.opensearch.cluster.metadata.ComposableIndexTemplate;
+import org.opensearch.cluster.metadata.Template;
+import org.opensearch.cluster.metadata.DataStream;
+import org.opensearch.client.Client;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.security.auditlog.impl.AuditMessage;
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.threadpool.ThreadPool;
+import org.opensearch.ResourceAlreadyExistsException;
+import org.opensearch.transport.RemoteTransportException;
+
+public final class InternalOpenSearchDataStreamSink extends AbstractInternalOpenSearchSink {
+
+    String dataStreamName;
+    private boolean dataStreamInitialized = false;
+
+    public InternalOpenSearchDataStreamSink(
+        final String name,
+        final Settings settings,
+        final String settingsPrefix,
+        final Path configPath,
+        final Client clientProvider,
+        ThreadPool threadPool,
+        AuditLogSink fallbackSink
+    ) {
+        super(name, settings, settingsPrefix, clientProvider, threadPool, fallbackSink, DocWriteRequest.OpType.CREATE);
+        Settings sinkSettings = getSinkSettings(settingsPrefix);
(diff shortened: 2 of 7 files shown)


peternied:
@tmanninger Thanks for the contribution,  can you please since your commits with the DCO [1]?

- [1] https://github.com/opensearch-project/security/pull/4257/checks?check_run_id=23865229124

src/main/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSink.java
```diff
@@ -85,6 +86,7 @@ public boolean doStore(final AuditMessage msg) {
                     .setRefreshPolicy(RefreshPolicy.IMMEDIATE)
                     .setSource(msg.getAsMap());
                 threadPool.getThreadContext().putHeader(ConfigConstants.OPENDISTRO_SECURITY_CONF_REQUEST_HEADER, "true");
+                irb.setOpType(DocWriteRequest.OpType.CREATE);
```
peternied, line 89: Please add new suite of test cases for audit logs that uses datastream.  I would recommend extending the existing 'basic' test cases [1] and modify them slightly to set a datastream instead of the default index.

- [1] https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/auditlog/integration/BasicAuditlogTest.java

peternied (changes requested):

DarshitChanpura (commented):
+1 to Peter's comments. @tmanninger Could you please +ve/-ve test cases for this change. We would like to really ensure that explicitly marking op_type as CREATE would resolve the question.

willyborankin:
Agree with @peternied and @DarshitChanpura . The data stream can be created via `client.createDataStream`. The setting `OpType.CREATED` for the document can help but currently an audit index is not a data stream. 
To create a data stream first you need:
 - Create an index template since data stream index is different compare to the regular index
 - Create a data stream
 - Create ISM polity to rollover and clean data stream indices
 
Here is an examle for it:
 https://github.com/opensearch-project/opensearch-java/blob/main/samples/src/main/java/org/opensearch/client/samples/DataStreamBasics.java

willyborankin:
I think the right solution would be to extend existing `InternalOpenSearchSink` or crate a new `Sync` call it let's say `internal_opensearch_datastream` or something else.  @peternied, @DarshitChanpura what do you thin about the idea?

tmanninger:
@willyborankin do you mean, the Sink should create the Template, Datastream and ISM policy? 
If i am using the the current internal sink with an index, i also must create the index template with an alias manually.

willyborankin:
> @willyborankin do you mean, the Sink should create the Template, Datastream and ISM policy? If i am using the the current internal sink with an index, i also must create the index template with an alias manually.

@tmanninger yes I think so, the main reason is that if you already has an audit index in your cluster and the cluster was migrated on the new version ... you fix will start to create a datastream ... which is not backward compatible.

tmanninger:
> > @willyborankin do you mean, the Sink should create the Template, Datastream and ISM policy? If i am using the the current internal sink with an index, i also must create the index template with an alias manually.
> 
> @tmanninger yes I think so, the main reason is that if you already has an audit index in your cluster and the cluster was migrated on the new version ... you fix will start to create a datastream ... which is not backward compatible.

Really?
`irb.setOpType(DocWriteRequest.OpType.CREATE);`
Without an datastream template, the fix will not create a datastream. I tested it in my test environment.

DarshitChanpura:
@tmanninger to ensure backwards compatibility for your fix: 
> Really?
irb.setOpType(DocWriteRequest.OpType.CREATE);
Without an datastream template, the fix will not create a datastream. I tested it in my test environment.

Can we add backwards compatibility tests as well?

tmanninger:
I will add some test cases.

tmanninger:
I added a test for datastream and for an index.
For both, i generate some audit logs and check and check the count of documents in the audit log index/datastream.

Thank's ok?

src/test/java/org/opensearch/security/auditlog/integration/BasicAuditlogTest.java
```diff
@@ -551,6 +551,70 @@ public void testIndexPattern() throws Exception {
...
+
+        HttpResponse res = restHelper().executePostRequest("_index_template/auditlog-datastream", templateBody, encodeBasicHeader("admin", "admin"));
+        Assert.assertEquals(res.getStatusCode(), HttpStatus.SC_OK);
+
+        setupStarfleetIndex();
+
+        HttpResponse res2 = rh.executeGetRequest("_data_stream/auditlog-datatream", encodeBasicHeader("admin", "admin"));
+        Assert.assertEquals(res2.getStatusCode(), HttpStatus.SC_OK);
+
+        HttpResponse res3 = rh.executeGetRequest("auditlog-datatream/_stats", encodeBasicHeader("admin", "admin"));
```
peternied, line 588: Getting the broad stats is useful, but we want to be sure the correct auto log entries are being generated, lets query the datastream's contents and be sure we have the expected audit log entries.

If the specific entries are missing the tests should fail, example testing depth [1]

- [1] https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/auditlog/compliance/ComplianceAuditlogTest.java#L309

src/test/java/org/opensearch/security/auditlog/integration/BasicAuditlogTest.java
```diff
@@ -551,6 +551,70 @@ public void testIndexPattern() throws Exception {
         Assert.assertTrue(res.getBody().contains("auditlog-20"));
     }
 
+    @Test
+    public void testDataStreamAsBackend() throws Exception {
```
peternied, line 555: I think we are missing a testing datatream's `_rollover` action - can you add validation for this kind of case where audit entries are created before/during/after and we are sure they have arrived intact?

src/test/java/org/opensearch/security/auditlog/integration/BasicAuditlogTest.java
```diff
@@ -551,6 +551,70 @@ public void testIndexPattern() throws Exception {
...
+        HttpResponse res = restHelper().executePostRequest("_index_template/auditlog-datastream", templateBody, encodeBasicHeader("admin", "admin"));
+        Assert.assertEquals(res.getStatusCode(), HttpStatus.SC_OK);
+
+        setupStarfleetIndex();
+
+        HttpResponse res2 = rh.executeGetRequest("_data_stream/auditlog-datatream", encodeBasicHeader("admin", "admin"));
+        Assert.assertEquals(res2.getStatusCode(), HttpStatus.SC_OK);
+
+        HttpResponse res3 = rh.executeGetRequest("auditlog-datatream/_stats", encodeBasicHeader("admin", "admin"));
+        Assert.assertTrue(Integer.valueOf(res3.getTextFromJsonBody("/_all/primaries/docs/count")) > 0);
```
peternied, line 589: Take this feedback or leave it: `assertThat` provides action error output on failure, especially when authoring tests: 
```suggestion
        var min = 3; var max = 5; // from 
        // ...
        assertThat(Integer.valueOf(res3.getTextFromJsonBody("/_all/primaries/docs/count")), allOf(greaterThan(min),lessThan(max)));
```

This would result in a message like 

```
//  java.lang.AssertionError:
//  Expected: (a value greater than <3> and a value less than <5>)
//       got: <1>
```


- [1] https://stackoverflow.com/a/7956297

peternied (commented):
> I think the right solution would be to extend existing InternalOpenSearchSink or create a new Sync call it let's say internal_opensearch_datastream or something else. @peternied, @DarshitChanpura what do you thin about the idea?

@willyborankin This is a great idea - I think we should do this.

> Add a backward combability test.

I don't think a BWC test is needed in this scenario so long as we are adding a new audit log type.

@tmanninger Let us know if you have questions, while the proposed fix could work.  I'd rather there is consistency.  Having to correctly configure the data stream before its used seems like a way to have a broken audit log and not be aware of it, whereas having a concrete definition ensures that the data stream is setup correctly without user intervention

tmanninger:
I can create a new Sink named internal_opensearch_datastream

- Create index template, if not exists
- Create datastream
- Write audit logs to this datastream

is this way ok?

peternied:
@tmanninger Please look into how the other audit log storage systems [1] work and consider what should be have sensible defaults that is configurable vs required.

As far as sink names go I think `internal_opensearch` and `external_opensearch` would look silly with `external_opensearch_data_stream` side-by-side.  You might need to be creative with how the class is instantiation or create 'inner classes' one for index and one for data_stream

As you mentioned `plugins.security.audit.config.datastream` would effectively communicate the name, and I would recommend that `plugins.security.audit.config.index` is mutually exclusive.  I think the template values should be configurable via a similar mechanism for the index shard/replica count of an index

Looks forward to those changes - thank you.

- [1] https://opensearch.org/docs/latest/security/audit-logs/storage-types/

tmanninger:
How can i create a datastream template?

When i am using this example:
https://github.com/opensearch-project/opensearch-java/blob/main/samples/src/main/java/org/opensearch/client/samples/DataStreamBasics.java.

`error: package org.opensearch.client.opensearch.indices does not exist`

I can use `org.opensearch.client.indices.PutIndexTemplateRequest`, but this class does not support DataStream config.

DarshitChanpura:
@tmanninger Are you still working on this?

tmanninger:
> How can i create a datastream template?
> 
> When i am using this example: https://github.com/opensearch-project/opensearch-java/blob/main/samples/src/main/java/org/opensearch/client/samples/DataStreamBasics.java.
> 
> `error: package org.opensearch.client.opensearch.indices does not exist`
> 
> I can use `org.opensearch.client.indices.PutIndexTemplateRequest`, but this class does not support DataStream config.

@DarshitChanpura see my post above... I find no option to create a IndexTemplate with datastream config...

DarshitChanpura:
@tmanninger Did you try importing java-client? If yes, Can you upload your code set-up here?

DarshitChanpura (commented):
@tmanninger Did you try importing java-client? If yes, Can you upload your code set-up here?

tmanninger:
When i import the java client:

```error: package org.opensearch.client.opensearch.indices does not exist```

**Comment by DarshitChanpura** [comment]
@tmanninger Do you plan to work on this?

**Comment by tmanninger** [comment]
When nobody helps me resolving my issue, no.

Regards,
Thomas

**Comment by DarshitChanpura** [comment]
@tmanninger Can you add this line in build.gradle:
```
implementation('org.opensearch.client:opensearch-java:2.14.0')
```

and run :
```
./gradlew clean build -x test -x integrationTest -x spotbugsIntegrationTest
```

if needed add following lines where other resolutions are forced in build.gradle:
```
force "commons-logging:commons-logging:1.3.4"
force "org.opensearch.client:opensearch-rest-client:3.0.0-SNAPSHOT"
force "jakarta.json:jakarta.json-api:2.1.3"
```

This will resolve import issue.

**Comment by tmanninger** [comment]
@DarshitChanpura thanks for feedback, now i can import it.

I hope, the last question:

My sink provider (InternalOpenSearchDataStreamSink) get passed the client as parameter. This legacy- client cannot create datastreams and datastreams template as shown in the example:
https://github.com/opensearch-project/opensearch-java/blob/main/samples/src/main/java/org/opensearch/client/samples/DataStreamBasics.java

How can i convert the client to a [RestHighLevelClient](https://javadoc.io/static/org.opensearch.client/opensearch-rest-high-level-client/2.1.0/org/opensearch/client/RestHighLevelClient.html) or a IndicesClient or what's the correct way?

Simple Code example:

InternalOpenSearchDataStreamSink.java:
```
import org.opensearch.client.IndicesClient;
import org.opensearch.client.indices.CreateDataStreamRequest;

public final class InternalOpenSearchDataStreamSink extends AuditLogSink {

    public InternalOpenSearchDataStreamSink(
        ...
        final Client clientProvider,
        ...
    ) {
      ...
      CreateDataStreamRequest createDataStreamRequest = new CreateDataStreamRequest(dataStreamName);
      // Now i need an IndicesClient object to call IndicesClient.createDataStream()
      // clientProvider does not have a indices() method.
    )
```

**Comment by cwperks** [comment]
@tmanninger Can you use `clientProvider.admin().indices().createDataStream(Request...)`?

**Comment by tmanninger** [comment]
@DarshitChanpur @peternied 
I commited a new version:

Config options:
```
plugins.security.audit.type: internal_opensearch_data_stream
plugins.security.audit.config.data_stream.name: opensearch-security-auditlog # Datastream name
plugins.security.audit.config.data_stream.template.manage: true # Should the template managed by the plugin?
plugins.security.audit.config.data_stream.template.name: opensearch-security-auditlog # Name of the template
plugins.security.audit.config.data_stream.template.number_of_shards: 1 # Number of shards
plugins.security.audit.config.data_stream.template.number_of_replicas: 0 # Number of repliaces
```

One note: I cannot create the template in the `InternalOpenSearchDataStreamSink` constructor, because at this step, the node is not up and ready.
I moved the creation of the template in the "datastream" funktion.

**Comment by peternied** [comment]
@tmanninger Thanks for circling back to this pull request.  I won't have time to re-review so I've dismissed my review, the other maintainers @cwperks & @DarshitChanpura should be able to help you out.

**Comment by tmanninger** [comment]
Tomorrow i will fix the spotline and checkstyle errors.

File: src/main/java/org/opensearch/security/auditlog/sink/InternalOpenSearchDataStreamSink.java [hunk 5]
```diff
@@ -0,0 +1,190 @@
...
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.core.common.bytes.BytesArray;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.ResourceAlreadyExistsException;
+import org.opensearch.transport.RemoteTransportException;
+
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+
+public final class InternalOpenSearchDataStreamSink extends AuditLogSink {
```
cwperks, line 48: There may be an opportunity to abstract the logic that is common to here and InternalOpenSearchSink into an Abstract class that both classes can inherit. These classes are pretty small, so it won't reduce the code too much but I thought I'd leave a comment  none-the-less.

src/main/java/org/opensearch/security/auditlog/sink/InternalOpenSearchDataStreamSink.java
```diff
@@ -0,0 +1,190 @@
...
+
+        // Create datastream template
+        if (templateManage) {
+        
+            final String templateName = sinkSettings.get(ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NAME, "opensearch-security-auditlog");
+            final Integer numberOfReplicas = sinkSettings.getAsInt(ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NUMBER_OF_REPLICAS, 0);
+            final Integer numberOfShards = sinkSettings.getAsInt(ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NUMBER_OF_SHARDS, 1);
+
+            final String auditlogTemplateJson = "{"
+                + "\"index_patterns\": [ \""+ dataStreamName +"\" ],"
```
cwperks, line 93: Is there a Builder pattern somewhere that can help assemble this payload?

I haven't looked at it in detail, but something like this: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/internalClusterTest/java/org/opensearch/action/admin/indices/datastream/DataStreamTestCase.java#L81-L89

src/main/java/org/opensearch/security/auditlog/sink/InternalOpenSearchDataStreamSink.java
tmanninger: I created an abstracted logic with AbstractInternalOpenSearchSink and shared duplicate code in this class.

src/main/java/org/opensearch/security/auditlog/sink/InternalOpenSearchDataStreamSink.java
tmanninger: Yes, that's really ugly, but i found no other way. Thanks for the link, i fixed it.

tmanninger:
I rebased git in an incorrect way :-/ (and github was auto-closed this MR) ...
I readded all my changes in one commit

tmanninger:
Added MR for documentation: https://github.com/opensearch-project/documentation-website/pull/8356

tmanninger:
I fixed the spotless errors.

tmanninger:
Test
```
    Expected: (a value greater than <0> and a value less than <5>)
         but: a value less than <5> <5> was equal to <5>
        at __randomizedtesting.SeedInfo.seed([BB9BC69071BEFE5B:AF3E69E787FB9BB0]:0)
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:18)
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
        at org.opensearch.security.auditlog.sink.InternalOpensearchDataStreamSinkTest.testTemplate(InternalOpensearchDataStreamSinkTest.java:108)
        at org.opensearch.security.auditlog.sink.InternalOpensearchDataStreamSinkTest.testDefaultSettings(InternalOpensearchDataStreamSinkTest.java:123)
 ```

Sometimes, there is one more audit log in the new datastream backend, maybe the rollover is faster than writing all audit logs generated before the rollover.
I increased the number of maximal audit docs.

@willyborankin pls rerun (hope the last time).

File: src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java [hunk 7]
```diff
@@ -1591,6 +1591,50 @@ public List<Setting<?>> getSettings() {
...
+                        + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NAME,
+                    Property.NodeScope,
+                    Property.Filtered
+                )
+            );
+            settings.add(
+                Setting.intSetting(
+                    ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX
+                        + ConfigConstants.SECURITY_AUDIT_OPENSEARCH_DATASTREAM_TEMPLATE_NUMBER_OF_SHARDS,
+                    3,
```
cwperks, line 1623: Curious, why a default value of 3 for this setting?

cwperks (dismissed):
> Added MR for documentation: [opensearch-project/documentation-website#8356](https://github.com/opensearch-project/documentation-website/pull/8356)

Thank you for creating a documentation PR for this change!

The changes look good to me, I just had one remaining question on the default value for number of shards.

tmanninger:
> The changes look good to me, I just had one remaining question on the default value for number of shards.

ss4o_traces_template and ss4o_metrics_template are using 3 shards as default value on all my opensearch clusters... So i copied the value.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
tmanninger: ss4o_traces_template and ss4o_metrics_template are using 3 shards as default value in opensearch... So i copied the value.

cwperks: wdyt about setting this to `1` by default? 

There was an issue recently filed about the security audit log index coming up as yellow in single node clusters: https://github.com/opensearch-project/security/issues/4739

By setting primary shard count 1 and replica shard count 0 by default (and allowing customization), it will come up as green in single node cluster.

tmanninger: I changed the default shard value to 1

cwperks (approved):

willyborankin (approved):

DarshitChanpura (approved):

Issue #4227, opened 2024-04-09
Introduce Environment variable for choosing serialisation - JDK or custom


Dhruvan1217:
**Is your feature request related to a problem?**
Currently, In some cases, users are having problems (#4085) while upgrading from 2.10 to 2.11+ where it switches the serialisation to custom serialisation technique (Ref: #2780) and have shard allocation issues.

**What solution would you like?**
* We can introduce an environment variable like `USE_JDK_SERIALIZATION=true` (Boolean) to allow the users to choose what serialisation they want to choose until all the corner cases are sorted for using custom serialisation. 
* Maybe default value for `USE_JDK_SERIALIZATION` is **false** if we think the custom serialisation is faster.

cwperks:
Thank you for filing this request @Dhruvan1217. @parasjain1 do you have any concerns with introducing a setting (something like `plugins.security.use_jdk_serialization`) that will explicitly tell a cluster to use JDK serialization? There is a [ConfigConstant](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/ConfigConstants.java#L336) for this, but it is not read in as a setting and used where the serialization format is [determined](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/ssl/transport/SecuritySSLRequestHandler.java#L93-L96).

scrawfor99:
[Triage] Hi @Dhruvan1217 thanks for filing this issue. @parasjain1 could you please provide your input on this request and make a contribution towards resolving the backwards compatibility issue. Otherwise we will have to pull the feature in the next version release.

peternied:
I'll be looking into this, I don't think we should add a feature flag, but I'll address the upgrade problem

PR #2802, opened 2023-05-30
Improve serialization speeds

parasjain1:
### Description
Use custom serialization in security plugin. 

Please refer https://github.com/opensearch-project/security/issues/2780 for details.

### Issues Resolved
- Resolves https://github.com/opensearch-project/security/issues/2780

### Testing
 - [x] Unit tests updated for Base64Helper
 - [x] Test version upgrade scenario
 - [x]  Test version upgrade with DLS/FLS/Masked Fields with active search traffic

### Check List
- [x] New functionality includes testing
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4264, opened 2024-04-17
Use predictable serialization logic for transport headers

peternied:
### Description
This change will prevent new clusters from using the 'custom serialization' format that was causing performance impact with customers in OpenSearch 2.11.

Background: the serialization changes from #2802 introduced issues where for certain serialization headers that were previously very small for over the wire transmission become much larger.  The root cause of this was that the JDK serialization process was able to detect duplicate entries and then use an encoding format to make it compressible.  Adding this logic into the serialization system from OpenSearch is non-trivial and is not being invested in. 

### Issues Resolved
- Related #2802
- Resolves https://github.com/opensearch-project/security/issues/4227
- Resolves https://github.com/opensearch-project/security/issues/3776

### Testing
BWC test should cover the scenarios
Before leaving draft, going to build tests that verify larger thread constructors are serialized in a compatible way.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
+++ b/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
@@ -63,6 +63,7 @@ import org.opensearch.security.ssl.transport.SSLConfig;
 import org.opensearch.security.support.Base64Helper;
 import org.opensearch.security.support.ConfigConstants;
 import org.opensearch.security.support.HeaderHelper;
+import org.opensearch.security.support.SerializationFormat;
 import org.opensearch.security.user.User;
 import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.Transport.Connection;
@@ -152,7 +153,7 @@ public class SecurityInterceptor {
 
         final boolean isDebugEnabled = log.isDebugEnabled();
 
-        final var serializationFormat = shouldUseJdkSerialization(connection);
+        final var serializationFormat = SerializationFormat.determineFormat(connection.getVersion());
         final boolean isSameNodeRequest = localNode != null && localNode.equals(connection.getNode());
 
         try (ThreadContext.StoredContext stashedContext = getThreadContext().stashContext()) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -268,14 +270,32 @@ && getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROL
...
     }
 
+    private static final String USE_JDK_SERIALIZATION = "plugins.security.use_jdk_serialization";
+    private static final Version FIRST_CUSTOM_SERIALIZATION_SUPPORTED_OS_VERSION = Version.V_2_11_0;
+    private static final Version CUSTOM_SERIALIZATION_NO_LONGER_SUPPORTED_OS_VERSION = Version.V_2_14_0;
+
+    private SerializationFormat shouldUseJdkSerialization(final Connection connection) {
+        var version = connection.getVersion();
+        if (version.after(FIRST_CUSTOM_SERIALIZATION_SUPPORTED_OS_VERSION)
+            && version.before(CUSTOM_SERIALIZATION_NO_LONGER_SUPPORTED_OS_VERSION)) {
```
peternied, line 280: @cwperks What do you think of this as the mitigation?  Being that the cost of building a custom serialization solution that also handles deduplication and compression is likely out of scope for the project in the long run I think its safe route to take.

So look as this looks good to you, then I'll start building out more test validation for these changes

cwperks: This looks ok. Let's say the de-dupe is fixed in a future minor version. Would implementing a bwc change to the new serialization format mean adding an `else if` block here?

peternied: Yup, it could.

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -249,17 +252,16 @@ final void completableRequestDecorate(
         TransportResponseHandler handler,
         DiscoveryNode localNode
     ) {
+        securityInterceptor.sendRequestDecorate(sender, connection, action, request, options, handler, localNode);
+        verifyOriginalContext(user);
+        try {
+            senderLatch.get().await(1, TimeUnit.SECONDS);
```
peternied, line 258: This waits around for either sender to complete its validation - previously those assertions were ignored

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -228,25 +230,28 @@ && getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROL
...
             }
 
-            if (useJDKSerialization) {
-                Map<String, String> jdkSerializedHeaders = new HashMap<>();
-                HeaderHelper.getAllSerializedHeaderNames()
-                    .stream()
-                    .filter(k -> headerMap.get(k) != null)
-                    .forEach(k -> jdkSerializedHeaders.put(k, Base64Helper.ensureJDKSerialized(headerMap.get(k))));
-                headerMap.putAll(jdkSerializedHeaders);
+            try {
```
peternied, line 233: Discovered a bug where if we are unable to serialize headers an exception bubbles out.  I have a feeling this is rare/impossible, but the unit test was able to trigger it.

peternied:
Note; BWC tests are failing because 2.14 needs the updated behavior - making the change directly in https://github.com/opensearch-project/security/pull/4288

peternied:
@DarshitChanpura @scrawfor99 Mind taking a look at this version of the change, nearly the same as https://github.com/opensearch-project/security/pull/4288

DarshitChanpura (approved):

scrawfor99 (approved):

Issue #4212, opened 2024-04-04
Replace Bouncy Castle Blake2bDigest in MaskedField


terryquigleysas:
**Is your feature request related to a problem?**
Replace Bouncy Castle Blake2bDigest in MaskedField
https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/MaskedField.java

**What solution would you like?**
Replace with https://mvnrepository.com/artifact/com.rfksystems/blake2b

**Do you have any additional context?**
This would help towards the following issues as Blake2b is not FIPS-compliant and is not present in the Bouncy Castle FIPS libraries:
- https://github.com/opensearch-project/security/issues/3420
- https://github.com/opensearch-project/security/issues/1500

cwperks:
[Triage] @terryquigleysas Is this issue a duplicate of this [similar issue](https://github.com/opensearch-project/security/issues/4213)? If this is a duplicate, could you please close one of the issues. Thank you!

terryquigleysas:
@cwperks They are closely related but not duplicate issues. We will be submitting standalone code for both.

terryquigleysas:
I think there is possibly a bug in the current Blake2b code. The **defaultSalt** variable is not passed to the Blake2bDigest constructor using the **salt** parameter as I would expect. It uses **personalization** instead. Is this correct?

![image](https://github.com/opensearch-project/security/assets/77437788/f08427a6-664c-447a-b75e-f085c3ad7e9b)

![image](https://github.com/opensearch-project/security/assets/77437788/399cc805-76ae-4e05-9b36-de806abcafd7)

In order to proceed which of these options is recommended?

1. Change to pass the **defaultSalt** variable to the Blake2bDigest constructor using the **salt** parameter. This will change the hash values. OR
2. Continue to pass the **defaultSalt** variable to the Blake2bDigest constructor using the **personalization** parameter, assuming there is a valid reason for this. The hash values generated will remain the same.

peternied:
@terryquigleysas Thanks for digging in on this.  I'd say we need to keep the digested values the same on existing clusters.  However, I think we should do the right thing in respect to the alternative implementation of Blake2bDigest.

terryquigleysas:
@peternied I am unsure why the digested values need to stay the same. The values are still masked, to the same length, but are now honoring the salt being passed in which appears to be the correct approach.

peternied:
[...]

PR #4271, opened 2024-04-19
Replace Bouncy Castle Blake2b

terryquigleysas:
### Description
Fix bug in existing Blake2b implementation. 
Replace Bouncy Castle Blake2bDigest in MaskedField.

Changes are backwards compatible. The only difference is that the Blake2b hash values will be different per value due to the nature of the bug mentioned above.

No documentation changes are required.

### Issues Resolved
Resolves https://github.com/opensearch-project/security/issues/4212

### Testing
Relevant tests have been updated.

### Check List

### DIFF
--- a/src/test/java/org/opensearch/security/dlic/dlsfls/CustomFieldMaskedComplexMappingTest.java
+++ b/src/test/java/org/opensearch/security/dlic/dlsfls/CustomFieldMaskedComplexMappingTest.java
@@ -167,10 +167,10 @@ public class CustomFieldMaskedComplexMappingTest extends AbstractDlsFlsTest {
                     "88783587fef740690c4fa39476fb86314d034fa3370e1a1fa186f6d9d4644a18ad85063c1e3161f8929f7ca019bb8740611eaf337709113901e7c3a6b59f4166"
                 )
         );
-        Assert.assertFalse(res.getBody().contains("e90a2fdf7b1939ec06e294321fd7d23e1a70d8fc080a3f85d0f3bf08c205b53"));
+        Assert.assertFalse(res.getBody().contains("9fe023e13d5179157b2023c21c60bc98340a12470538affd306281619d2477c3"));
         Assert.assertFalse(res.getBody().contains("*.*.*.*"));
-        Assert.assertFalse(res.getBody().contains("430a65d4b9c51de7192e048b2639db0de5c56f1901afccc2a01ef97f6a769a38"));
-        Assert.assertFalse(res.getBody().contains("7f48bb3636edf546a75968ca7cd0bdfe63e9ce7af04ef7cb642931fa15d2d7a3"));
+        Assert.assertFalse(res.getBody().contains("5115e9707d1a18772af2a1274e2d3450284754a921692d6b1aef922c33cda8fa"));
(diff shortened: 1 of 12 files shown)


cwperks:
Thank you for the contribution @terryquigleysas!

This change looks good to me. For FIPS compliance, will all references to non-FIPS BC libraries be removed when compiling/running in FIPS mode?

cwperks:
For the rfksystems library, the repo does not have a License file but the README specifies [Apache 2.0](https://github.com/rfksystems/blake2b?tab=readme-ov-file#license)

terryquigleysas:
@cwperks Do you mean that you expect to see a license file in https://github.com/rfksystems/blake2b ?

Both the README you reference and https://mvnrepository.com/artifact/com.password4j/password4j state it is licensed under Apache 2.0. Will the absence of an actual file prevent this from being accepted?

cwperks:
@terryquigleysas It's fine as long as the license is specified somewhere. I left a comment in case others were looking for the license. In my experience, open-source repos will typically have a separate LICENSE file containing the full terms of the license. 

In opensearch core there's a gradle check called `./gradlew dependencyLicenses` that requires licenses to be added to the core repo directly (Example in [`discovery-gce` plugin](https://github.com/opensearch-project/OpenSearch/tree/main/plugins/discovery-gce/licenses)) to ensure that new dependencies have acceptable licenses.

terryquigleysas:
@cwperks Thanks, sounds good. Is there anything else I need to do?

src/test/java/org/opensearch/security/dlic/dlsfls/CustomFieldMaskedComplexMappingTest.java
```diff
@@ -167,10 +167,10 @@ public void testComplexMappingSearch() throws Exception {
                     "88783587fef740690c4fa39476fb86314d034fa3370e1a1fa186f6d9d4644a18ad85063c1e3161f8929f7ca019bb8740611eaf337709113901e7c3a6b59f4166"
                 )
         );
-        Assert.assertFalse(res.getBody().contains("e90a2fdf7b1939ec06e294321fd7d23e1a70d8fc080a3f85d0f3bf08c205b53"));
+        Assert.assertFalse(res.getBody().contains("9fe023e13d5179157b2023c21c60bc98340a12470538affd306281619d2477c3"));
```
peternied, line 170: > The only difference is that the Blake2b hash values will be different [...]

This seems specifically backward **in**compatible.  I would not expect any changes to these unit tests - I don't think this is safe to backport.  Can we add an alternative provider for the hashing logic that can be 'configured on' rather than on by default?  This allows system admins to choose if/when they want to take this change

peternied (changes requested):

src/test/java/org/opensearch/security/dlic/dlsfls/CustomFieldMaskedComplexMappingTest.java
terryquigleysas: @peternied The hashing changes because the salt does not appear to be passed correctly in the existing code. Please see comment https://github.com/opensearch-project/security/issues/4212#issuecomment-2063453718

terryquigleysas:
Closing off this PR and replacing with https://github.com/opensearch-project/security/pull/4275 and bug https://github.com/opensearch-project/security/issues/4274

Issue #3903, opened 2023-12-29
[RFC] Security Performance Test Suite


nibix:
# Background

Already now, the OpenSearch project systematically conducts benchmark test runs and publishes the results: https://opensearch.org/benchmarks/ This is based on the pre-defined [benchmark-workloads](https://github.com/opensearch-project/opensearch-benchmark-workloads)  which are executed using the tool [opensearch-benchmark](https://github.com/opensearch-project/opensearch-benchmark).

Additionally, there is ongoing work to define testing processes for performance related quality assurance: https://github.com/opensearch-project/OpenSearch/issues/7499

These efforts, however, focus on the OpenSearch core; the efficiency of the code implementing data retrieval and indexing is tested. In some cases, there is a comparison between OpenSearch with security enabled vs OpenSearch without security enabled. While this can give a first idea of the overhead produced by the security plugin, the results cannot be generalized. This is because the performance overhead of the security plugin depends on a number of different factors. See below for more information on this.

Ideally, there would be an additional security-focused performance test suite. This test suite could be used for assuring that the security plugin does not suffer any performance regressions. Additionally, it could be used to determine the actual benefit of performance enhancing code changes (for example https://github.com/opensearch-project/security/issues/3870).

Goal of this RFC is to collect requirements for such a test suite, learn about ongoing efforts, and to find a way to integrate it into the software development process.

# Performance influencing factors

The performance overhead of the security plugin depends on a number of different factors:

- Number of indices on a cluster. The security plugin access controls involve resolving requested index patterns and index patterns in the role definitions to actual indices. For clusters with many indices, this can slow things down (see for example https://github.com/opensearch-project/security/issues/3870 )

- Number of roles of a user. While the example configurations usually demonstrate users with one or two roles, there are environments which act very liberal in assigning roles to users. Users with hundreds of roles do happen in reality. In order to determine the authorization status of a user, the security plugin needs to loop through every role of a user for each request.

- Size of the user object. During inter-node communication, the security plugin needs to pass on the user object to the target node; for this, the user object needs to be serialized and deserialized. Larger user objects cause higher overhead during serialization and deserialization. Additionally, larger user objects put more stress onto the network.

- Use of patterns vs constant index names in the role configuration. 

- Use of dynamic index patterns. The security plugin allows to define roles which use expressions to dynamically define the indices for which privileges are granted. Additionally, date math can be used to dynamically grant privileges. Both cases add a certain overhead to the privilege evaluation.

[...]

PR #4282, opened 2024-04-23
Performance benchmark workflow

abraakie:
### Description

This will run a benchmark workflow for each push and pull request against a base branch (main, 1.*, 2.*).
For pull requests the results of the benchmark are compared to the results of previous runs on the target branch. If the threshold (currently 150%) is exceeded, the benchmark workflow will fail.

This workflow helps identifying performance impacts of code changes in an early state.

### Issues Resolved

- Resolves https://github.com/opensearch-project/security/issues/3903

### Testing

The workflow is tested manually. The result fluctuation is reasonable.

### Check List

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,69 @@
+name: Benchmark
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - 1.*
+      - 2.*
+  pull_request:
+    branches:
+      - main
+      - 1.*
+      - 2.*
+env:
+  JAVA_VERSION: 21
+  OPENSEARCH_VERSION: 3.0.0
+  OPENSEARCH_ADMIN_PASSWORD: Hello_World123
+jobs:
+  benchmark:
+    runs-on: ubuntu-latest
+    name: Benchmark
+    steps:
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: ${{ env.JAVA_VERSION }}
+      - name: Checkout Branch
+        uses: actions/checkout@v4
+      - name: Set up gradle
+        uses: gradle/actions/setup-gradle@v3
+        with:
+          cache-disabled: false
+      - name: Assemble target plugin
+        run: ./gradlew assemble
+      - name: Prepare Security Plugin
+        run: mv ./build/distributions/opensearch-security-*.zip ./benchmark/docker/
+      - name: Build Docker Image
+        run: docker build --platform linux/amd64 --build-arg VERSION=${{ env.OPENSEARCH_VERSION }} -f benchmark/docker/benchmark.dockerfile -t opensearchproject/security-benchmark:latest benchmark/docker/
+      - name: Run Benchmark Cluster
+        uses: isbang/compose-action@v1.5.1
+        with:
+          compose-file: "./benchmark/docker-compose.yml"
+        env:
+          OPENSEARCH_INITIAL_ADMIN_PASSWORD: ${{ env.OPENSEARCH_ADMIN_PASSWORD }}
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install Benchmark
+        run: pip install opensearch-benchmark
+      - name: Execute Benchmarks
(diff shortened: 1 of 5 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Set up gradle
+        uses: gradle/actions/setup-gradle@v3
+        with:
+          cache-disabled: false
+      - name: Assemble target plugin
+        run: ./gradlew assemble
+      - name: Prepare Security Plugin
+        run: mv ./build/distributions/opensearch-security-*.zip ./benchmark/docker/
+      - name: Build Docker Image
+        run: docker build --platform linux/amd64 --build-arg VERSION=${{ env.OPENSEARCH_VERSION }} -f benchmark/docker/benchmark.dockerfile -t opensearchproject/security-benchmark:latest benchmark/docker/
```
peternied, line 39: Why containerize with docker? We already have a supported model for starting and executing tests against a cluster, see this workflow for an example:
https://github.com/opensearch-project/security/blob/main/.github/workflows/plugin_install.yml

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Run Benchmark Cluster
```
peternied, line 40: I don't think test will produce consistent metrics when run in a cluster.  If we are trying to detect 'large' issues, a cluster adds more variables.  What results have you seen in your testing - can you help build my confidence on these choices?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,69 @@
...
+        env:
+          OPENSEARCH_INITIAL_ADMIN_PASSWORD: ${{ env.OPENSEARCH_ADMIN_PASSWORD }}
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install Benchmark
+        run: pip install opensearch-benchmark
+      - name: Execute Benchmarks
+        run: opensearch-benchmark execute-test --pipeline=benchmark-only --results-format=csv --results-file=./results.csv --on-error=abort --workload=percolator --target-host=https://localhost:9200 --client-options=basic_auth_user:admin,basic_auth_password:${{ env.OPENSEARCH_ADMIN_PASSWORD }},verify_certs:false,timeout:60
```
peternied, line 53: This workflow uses `percolator` which doesn't exercise any of the security plugin's functionality, until we are doing tests with multiple users and different feature configurations this isn't any more useful than that nightly benchmarks against that are already published and managed via https://opensearch.org/benchmarks/

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Prepare Benchmark Results
+        run: python benchmark/result_rewriter.py ./results.csv ./results.json
+      - name: Download previous benchmark data
+        uses: actions/cache@v4
+        with:
+          path: ./cache
+          key: ${{ runner.os }}-benchmark
+      - name: Store Benchmark Results
```
peternied, line 61: Can you link to a couple of workflows that show no-regression vs regression so we can better understand how this looks?

benchmark/result_rewriter.py
```diff
@@ -0,0 +1,53 @@
+import json
```
peternied, line 1: Can we see about making a PR to benchmarks to support another output format natively rather than baking it into the security repo?

peternied (commented):
Thanks for creating the PR @abraakie Maybe you could help me understand more of the vision for these changes by extracting out parts of the RFC [1] that you are addressing vs those that you aren't planning vs things you are still implementing for this PR?

- [1] https://github.com/opensearch-project/security/issues/3903

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,69 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install Benchmark
+        run: pip install opensearch-benchmark
+      - name: Execute Benchmarks
+        run: opensearch-benchmark execute-test --pipeline=benchmark-only --results-format=csv --results-file=./results.csv --on-error=abort --workload=percolator --target-host=https://localhost:9200 --client-options=basic_auth_user:admin,basic_auth_password:${{ env.OPENSEARCH_ADMIN_PASSWORD }},verify_certs:false,timeout:60
+      - name: Prepare Benchmark Results
+        run: python benchmark/result_rewriter.py ./results.csv ./results.json
+      - name: Download previous benchmark data
```
cwperks, line 56: Can this eventually be extended to the last n runs? i.e. Get average or median of last 10 runs instead of only the previous run?

.github/workflows/benchmark.yml
cwperks: +1

DarshitChanpura:
@abraakie Closing this PR since there hasn't been any activity since 2+ months. Feel free to open a new one.

PR #4314, opened 2024-05-03
Flush user cache

### LINKED ISSUE #2829: [FEATURE] Endpoint to purge cache entries for specific users
As a user's backend roles may change, the user's `restRoleCache` entries may become stale. We want to invalidate stale cache items without affecting other users' cached entries. Currently there's only a single endpoint to invalidate user authentication caches `DELETE _opendistro/_security/api/cache` but this invalidates the entire cache rather than individual cache entries and so there may be many cache misses when doing that.

Additional endpoint to invalidate user authentication caches on a per user level.

Cache TTL `plugins.security.cache.ttl_minutes` was considered but if a user's backend role changes prior to the expiration of cached entry then cached entry will be stale.

dlin2028:
### Description
Currently, our system only allows for the invalidation of the entire user authentication cache, which can lead to numerous cache misses and inefficiencies. This change is required to allow for more precise cache management, specifically targeting stale cache entries at the individual user level without disrupting the cache state of other users

* Category: New feature
Currently, our system only allows for the invalidation of the entire user authentication cache. This change is required to allow for more precise cache management, specifically targeting stale cache entries at the individual user level without disrupting the cache state of other users.
Previously, invalidating a user's cache required clearing the entire cache, affecting all users. The new behavior introduces an endpoint that allows for the invalidation of cache entries on a per-user basis, thereby maintaining cache integrity for other users and reducing unnecessary cache misses.
### Issues Resolved
#2829

### Testing
Unit testing was written in FlushCacheApiTest.java

### Check List
- [X ] New functionality includes testing
- [ X] Commits are signed per the DCO using --signoff


src/test/java/org/opensearch/security/dlic/rest/api/FlushCacheApiTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+import org.opensearch.common.settings.Settings;
+import org.opensearch.common.xcontent.XContentType;
+import org.opensearch.security.test.helper.rest.RestHelper.HttpResponse;
+
+import static org.opensearch.security.OpenSearchSecurityPlugin.PLUGINS_PREFIX;
+
+public class FlushCacheApiTest extends AbstractRestApiUnitTest {
```
willyborankin, line 25: Let's move the test into `integration-tests`

DarshitChanpura:
@dlin2028 Is this being worked on actively?

dlin2028:
@DarshitChanpura No. I think someone else took this one.
https://github.com/opensearch-project/security/pull/4571

Issue #4213, opened 2024-04-04
[FEATURE] Configuration Option to Set Default Algorithm for Field Masking


terryquigleysas:
**Is your feature request related to a problem?**
The field masking algorithm defaults to Blake2b. We'd like to be able to change this via a configuration option.

**What solution would you like?**
Add configuration option, e.g. 
plugins.security.masked_fields.algorithm.default

This can be set to any of the other supported algorithms (e.g. SHA-512) https://opensearch.org/docs/latest/security/access-control/field-masking/#advanced-use-an-alternative-hash-algorithm 

**Do you have any additional context?**
As well as adding functionality for wider use it will also help for adding configurable options for FIPS compliance:
https://github.com/opensearch-project/security/issues/3420

terryquigleysas:
Please assign to me once triaged.

cwperks:
[Triage] Thank you for filing this issue @terryquigleysas. I have assigned this to you and thank you for providing a reference implementation [here](https://github.com/opensearch-project/security/compare/2.11...terryquigleysas:security:2.11).

What do you think about the setting being called `plugins.security.masked_fields.algorithm`? By having `default` on the path to the setting (`plugins.security.masked_fields.algorithm.default`) does this imply that it can be changed without a cluster reboot?

terryquigleysas:
@cwperks The algorithm can be set per field as detailed in https://opensearch.org/docs/latest/security/access-control/field-masking/#advanced-use-an-alternative-hash-algorithm

When no specific algorithm is provided for a field the default is hardcoded to choose Blake2b. We want to provide an option to choose a default of our choice when a field has no algorithm specified. Initially this would be a static setting that would require a reboot.

Issue #4274, opened 2024-04-20
[BUG] Blake2b hashing for Masked Fields does not apply salt correctly


terryquigleysas:
I think there is possibly a bug in the current Blake2b code. The **defaultSalt** variable is not passed to the Blake2bDigest constructor using the **salt** parameter as I would expect. It uses **personalization** instead. Is this correct?

![image](https://github.com/opensearch-project/security/assets/77437788/f08427a6-664c-447a-b75e-f085c3ad7e9b)

![image](https://github.com/opensearch-project/security/assets/77437788/399cc805-76ae-4e05-9b36-de806abcafd7)

In order to proceed which of these options is recommended?

1. Change to pass the **defaultSalt** variable to the Blake2bDigest constructor using the **salt** parameter. This will change the hash values. OR
2. Continue to pass the **defaultSalt** variable to the Blake2bDigest constructor using the **personalization** parameter, assuming there is a valid reason for this. The hash values generated will remain the same.

_Originally posted by @terryquigleysas in https://github.com/opensearch-project/security/issues/4212#issuecomment-2063453718_
            
Approach #1 would appear to be the correct thing to do but there are concerns that changing hashes, even to the correct values, may impact existing users.

This bug is derived from discussions on https://github.com/opensearch-project/security/pull/4271 and https://github.com/opensearch-project/security/issues/4212

scrawfor99:
[Triage] Hi @terryquigleysas thank you for filing this issue. This sounds like a worthwhile change which could help correct some unexpected behavior. We will need to handle the issues around the backwards compatibility of the code when reviewing the PR.

PR #4336, opened 2024-05-13
Configure masking algorithm default

terryquigleysas:
### Description
* New feature
* The field masking algorithm defaults to Blake2b. We'd like to be able to change this via a configuration option.

### Issues Resolved
Resolves https://github.com/opensearch-project/security/issues/4213

### Testing
New tests added. 
Tested locally against a running cluster.

### Documentation
I will be raising an issue to add documentation to https://opensearch.org/docs/latest/security/access-control/field-masking

### Check List


peternied (commented):
Thanks for this update @terryquigleysas.  Lets shift the logic around so the MaskedField uses an algorithm provider.  As masked field should not need to know which implementation is being used, but just that it is using the algorithm selected for use by the plugin.

This will pave the way for different kinds of providers to by swapped, such as a specific FIPS compliant set vs others.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -430,6 +431,19 @@ public List<Path> run() {
             }
         }
 
+        try {
+            String maskingAlgorithmDefault = settings.get(ConfigConstants.SECURITY_MASKED_FIELDS_ALGORITHM_DEFAULT);
+            if (StringUtils.isNotEmpty(maskingAlgorithmDefault)) {
+                MessageDigest.getInstance(maskingAlgorithmDefault);
+            }
+        } catch (Exception ex) {
+            throw new OpenSearchSecurityException(
```
cwperks, line 440: Can we add a test case for this scenario and ensure the node fails to boot up as expected?

What happens if a user explicitly puts `blake2b` in for this setting? Will the `MessageDigest.getInstance(...)` call fail since its not in this [list](https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#MessageDigest)?

cwperks (commented):
@terryquigleysas What do you think about passing the default in as a constructor arg to MaskedField in this [call](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/DlsFlsFilterLeafReader.java#L299) to avoid passing it into the `mask` function on every call?

terryquigleysas:
> @terryquigleysas What do you think about passing the default in as a constructor arg to MaskedField in this [call](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/DlsFlsFilterLeafReader.java#L299) to avoid passing it into the `mask` function on every call?

@cwperks Thank you for the review. I have changed the code to this effect.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
terryquigleysas: That is the purpose of FieldMaskedTest.testMaskedGetClusterDefaultDoesNotExist()

src/main/java/org/opensearch/security/configuration/DlsFlsFilterLeafReader.java
```diff
@@ -1229,11 +1236,13 @@ private static class MaskedTermsEnum extends TermsEnum {
...
         private final TermsEnum delegate;
         private final MaskedField mf;
+        private final String algorithmDefault;
 
-        public MaskedTermsEnum(TermsEnum delegate, MaskedField mf) {
+        public MaskedTermsEnum(TermsEnum delegate, MaskedField mf, String algorithmDefault) {
             super();
             this.delegate = delegate;
             this.mf = mf;
+            this.algorithmDefault = algorithmDefault;
```
cwperks, line 1245: Looks like this is no longer being used in this static inner class. Can this be removed from here?

terryquigleysas: @cwperks Good spot. Removed.

src/main/java/org/opensearch/security/dlic/rest/api/RolesApiAction.java
```diff
@@ -91,7 +91,7 @@ private ValidationResult<JsonNode> validateMaskedFields(final JsonNode content)
 
         private Pair<String, String> validateMaskedFieldSyntax(final JsonNode maskedFieldNode) {
             try {
-                new MaskedField(maskedFieldNode.asText(), SALT).isValid();
+                new MaskedField(maskedFieldNode.asText(), SALT, "").isValid();
```
peternied, line 94: The algorithm should be checked during the validity check - lets plumb this through

peternied (commented):
Thanks for the updates @terryquigleysas looks much cleaner, the refactor to a provider/factory doesn't seem a relevant with these changes.

src/main/java/org/opensearch/security/dlic/rest/api/RolesApiAction.java
terryquigleysas: Thanks @peternied  I considered this and my findings were:

- This isn't plumbed through for the existing, configured salt in this instance either. 
- If the new configured default algorithm were invalid the node would not have started in the first place.

If needed, which I don't think it is, how can the setting be easily referenced here?

src/main/java/org/opensearch/security/configuration/DlsFlsFilterLeafReader.java
```diff
@@ -130,7 +131,8 @@ class DlsFlsFilterLeafReader extends SequentialStoredFieldsLeafReader {
         this.clusterService = clusterService;
         this.auditlog = auditlog;
         this.salt = salt;
-        this.maskedFieldsMap = MaskedFieldsMap.extractMaskedFields(maskFields, maskedFields, salt);
+        this.maskingAlgorithmDefault = clusterService.getSettings().get(ConfigConstants.SECURITY_MASKED_FIELDS_ALGORITHM_DEFAULT);
```
DarshitChanpura, line 134: QQ, what happens if this setting is not explicitly set? Would it default to blake2b? If so should it not default to blake2b in case any non-supported names are passed in the setting, instead of boot-up failure?

src/main/java/org/opensearch/security/configuration/MaskedField.java
```diff
@@ -57,31 +60,31 @@ public final void isValid() throws Exception {
...
-            return blake2bHash(value);
+        if (algo != null) {
+            return customHash(value, algo);
+        } else if (regexReplacements != null) {
+            String cur = new String(value, StandardCharsets.UTF_8);
+            for (RegexReplacement rr : regexReplacements) {
+                cur = cur.replaceAll(rr.getRegex(), rr.getReplacement());
+            }
+            return cur.getBytes(StandardCharsets.UTF_8);
+        } else if (StringUtils.isNotEmpty(defaultAlgorithm)) {
```
DarshitChanpura, line 71: Does the order of execution matter here? If so why choose this particular order?

DarshitChanpura (commented):
Thanks for filing this PR @terryquigleysas. I left some clarifying comments from my end.

src/main/java/org/opensearch/security/configuration/DlsFlsFilterLeafReader.java
terryquigleysas: For backwards compatibility if not explicitly set it defaults to Blake2b as it currently does. 

In fact I had to raise https://github.com/opensearch-project/security/issues/4274 as I have been asked for it to be backwards compatible with the incorrect parameters that are currently in used in the code in order for the same hashes to be generated.

src/main/java/org/opensearch/security/configuration/MaskedField.java
terryquigleysas: Yes the order is important. The logic is the same as the existing code (albeit slightly refactored) with the addition of the new configuration option. 

Previously the masking choice as detailed in https://opensearch.org/docs/latest/security/access-control/field-masking/#advanced-use-an-alternative-hash-algorithm is basically to: 
a) use what is configured for the field(s) according to the role definition if explicitly specified 
OR 
b) default to Blake2b where nothing explicitly set

Now it is 
a) use what is configured for the field(s) according to the role definition if explicitly specified 
OR 
b) use the new configuration to specify an algorithm other than Blake2b to default to without having to do this via every role definition
OR
c) default to existing, backwards compatible Blake2b where nothing explicitly set either in the role definition or configuration

src/main/java/org/opensearch/security/dlic/rest/api/RolesApiAction.java
cwperks: @terryquigleysas `opensearchSettings` are passed as a constructor arg in this class.

cwperks: Once this is addressed, the rest looks good to me. It doesn't appear like the algo matters for the validation check, but I agree that we should pass the algo into this call.

One change we could consider making in the future is to track the supported algorithms in an Enum so that its not a generic String, but the list of supported algorithms will change depending on the version of the JDK used.

terryquigleysas: > @terryquigleysas The settings can be extracted from the `securityApiDependencies`. These dependencies are passed to all API handlers from [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/dlic/rest/api/SecurityRestApiActions.java#L69).
> 
> `securityApiDependencies.settings()` - https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/dlic/rest/api/SecurityApiDependencies.java

@cwperks Thanks. Code updated.

cwperks:
@terryquigleysas last request, can you sign all of the commits on this PR? The DCO check is failing: https://github.com/opensearch-project/security/pull/4336/checks?check_run_id=24958472355

terryquigleysas:
> @terryquigleysas last request, can you sign all of the commits on this PR? The DCO check is failing: https://github.com/opensearch-project/security/pull/4336/checks?check_run_id=24958472355

@cwperks  I have followed the process at https://github.com/opensearch-project/security/pull/4336/checks?check_run_id=24958472355 but things do not appear to have changed here. Is there another way to do this? I'm new to this specific process.

src/main/java/org/opensearch/security/configuration/DlsFlsFilterLeafReader.java
terryquigleysas: > okay. Is the future expectation that node boot-up should fail if no custom algo is passed?

Not from my point of view. This setting is optional and allows more control over the default algorithm used. It will only prevent node boot if that algorithm is not supported by the JVM.

DarshitChanpura: Should we prevent node boot-up and instead display a message stating masking defaulted to blake2b since the custom supplied algo is not recognized by JVM. I'm 50-50 on stopping node boot-up. On one hand, prevent node from coming up would allow users to detect the misconfigured value early but at the same time we are blocking node boot-up for something that can be defaulted to a value, similar to some other settings.

terryquigleysas: Hi @DarshitChanpura Yes, there are arguments either way. I leaned heavily towards stopping boot up due mainly to the following reasons:

- This will be documented under a section that is labelled as Advanced https://opensearch.org/docs/latest/security/access-control/field-masking#advanced-use-an-alternative-hash-algorithm and we could assume the user changing will be an advanced user
- The error message is very clear about why the node doesn't start and should be especially so for an advanced user
- I was concerned that logging a warning and continuing may mean this incompatibility being missed. This may be of particular concern when set for compliance purposes.

DarshitChanpura (approved):
LGTM! Thank you @terryquigleysas !

cwperks:
@willyborankin any concerns with this PR using the older test mechanism?

This PR looks good to me. I think it can be followed-up with a change to the `integrationTest` package.

willyborankin:
> @willyborankin any concerns with this PR using the older test mechanism?
> 
> This PR looks good to me. I think it can be followed-up with a change to the `integrationTest` package.

@cwperks If fine with.

cwperks (approved):

cwperks:
@terryquigleysas Did you create a documentation PR to accompany this change? Can you update the PR description to include a link to the companion PR for the [documentation-website](https://github.com/opensearch-project/documentation-website)?

terryquigleysas:
> @terryquigleysas Did you create a documentation PR to accompany this change? Can you update the PR description to include a link to the companion PR for the [documentation-website](https://github.com/opensearch-project/documentation-website)?

Yes, it can be found at https://github.com/opensearch-project/documentation-website/pull/7162

Issue #4361, opened 2024-05-23
Replace Bouncy Castle's OpenBSDBCrypt used in password hashing and verification


dancristiancecoi:
### Is your feature request related to a problem?
Currently the password hashing and verification logic is built around Bouncy Castle's OpenBSDBCrypt. This will present a problem in FIPS enabled environments as BCrypt is not a FIPS compliant algorithm & the BouncyCastle provider itself is not FIPS enabled.

### What solution would you like?

I propose we remove any references to OpenBSDBCrypt from the code & replace it with an implementation built around the [Password4J ](https://github.com/Password4j/password4j) library. 

In my opinion these are the points in favor of Password4j:

- [Open Source](https://github.com/Password4j/password4j) with an Apache License
- Actively maintained
- Supports BCrypt, Argon2, SCrypt, PBKDF2
- Straightforward API
- Backwards compatible (it can verify hashes generated by the current BouncyCastle/OpenBSDBCrypt implementation)
- Supports updating of existing hashes (to a more secure configuration or to an entirely different algorithm).
- Relatively popular

### Do you have any additional context?
This will help towards the following issues:

- https://github.com/opensearch-project/security/issues/3420
- https://github.com/opensearch-project/security/issues/4079 (support for additional BCrypt configuration & additional algorithms (PBKDF2, SCrypt, Argon2) will be done in separate PR's

dancristiancecoi:
Please assign to me once triaged.

Issue #4079, opened 2024-02-26
[FEATURE] Add support for PBKDF2 for passwords hashing


willyborankin:
**What solution would you like?**

Currently we support only `bcrypt` for passwords hashing which is resource consuming for slow CPUs.  
It would be a good idea to give customers a possibility to select different famous hashing algorithms like: `PBKDF2`.

To archive that new properties should be added in the security plugin configuration:
 - `plugins.security.password.hashing.algorithm = ...`  - default is `bcrypt`, supported values are: `pbkdf2`.
 - `plugins.security.password.hashing.<algorithm>.<param_name> = ...` - multiple parameters which are relevant to each algo 

e.g.
```yaml

plugins.security.password.hashing.algorithm='pbkdf2'
plugins.security.password.hashing.pbkdf2.iterations = 1000 
plugins.security.password.hashing.pbkdf2.hash_size_bytes = 128 
plugins.security.password.hashing.pbkdf2.salt_size_bytes = 12 

```
In case of  `PBKDF2` an additional security settings needs to be add:

peternied:
@willyborankin Looks like this would this address the perf issues around PATCH on the user API [1], do I have that right? 

- [1] https://github.com/opensearch-project/security/issues/4008

willyborankin:
@peternied, Yes, the main idea that users can try to turn hashing password algos. wdyt does it make sense?

scrawfor99:
[Triage] Hi @willyborankin, thanks for filing this issue. This change seems to have a clear direction forward and would be valuable in fixing some current performance issues.

dancristiancecoi:
Hi, can I get assigned to this please?

dancristiancecoi:
Hello!

I have some WIP changes for this feature. As I am somewhat unfamiliar with the code-base I'd like to discuss it before I proceed further in case I am taking a wrong approach. 

The solution is based around the Password4J library. The main two considerations when choosing it were: 
1) FIPS compliance
2) Built-in support of every algorithm that we want to support as part of this feature (_BCrypt, PBKDF2, SCrypt, Argon2_)

Password4j ticked both boxes while other solutions would have required pulling in multiple libraries for each algorithm & potential dependency on BouncyCastle. 

In terms of configuration, this is what the current implementation supports: 

### BCrypt:

1. Rounds
2. Minor version

### PBKDF2

1. Algorithm/Function (SHA-256 etc)
2. Iterations
3. Output Length

### Argon2

1. Memory
[...]

Issue #3420, opened 2023-09-28
[RFC] Proposal for supporting FIPS 140-2 enforced mode


terryquigleysas:
**Is your feature request related to a problem? Please describe.**
Feature Request [#1497](https://github.com/opensearch-project/security/issues/1497)

**Describe the solution you'd like**
# Problem Statement
We would like to contribute to OpenSearch to support running in FIPS-140-2 compliant mode. We propose delivering this in several phases, as discussed in the feature request above, starting with core changes and aiming towards a desired state of providing configurable options.

This RFC is to ensure our approach would be seen as a feasible and acceptable contribution.
### Phase 1: Remove hardcoded Bouncy Castle references 
**Security plugin**

Update code, retaining current functionality
  - Proposed libraries: Bouncy Castle FIPS, Password4j, rfksystems Blake2b
    - Alternatively contribute to Password4j to expose Blake2b functionality and reduce the number of libraries brought in
  - Security policy changes
    - Complicated by the plugin structure
  - NB OpenSearch must still work for rolling upgrades
 
**Performance Analyzer (potentially)**
  - This codebase is separate from OpenSearch Security and may also lead us to have to make changes to OpenSearch core
  - Who do we liaise with?

**Unknown unknowns (e.g. behavior of other plugins, scripts etc.)**
  - Emphasis on not inadvertently breaking anything

### Phase 2: Introduce FIPS-compliant alternatives as default for:
Bcrypt password hashing
  - PBKDF2

Blake2b for masking
  - e.g. SHA3

Certificate handling (potentially)

Cipher lists (potentially)

Any additional security policy changes

Add FIPS mode configuration flag
  - This may lead us to have to make changes to OpenSearch core
### Phase 3: Testing and rework
_By now we will be carrying out extensive testing and verification and expect that additional requirements may arise._

Additional work for any issues found in our testing

Extend unit tests

Extend integration tests

### Phase 4: Configurability
Additional configuration options
  - Configure additional security providers
  - Configure hashing algorithms
  - Validation

Contingency for unknown unknowns

### Phase 5: Documentation

All required configuration options and settings

JDK 11 requirement

Limitations
### Not in scope

Changing an existing cluster from non-FIPS to FIPS compliant

Dashboards, Data Prepper etc. - our focus is on server only

Any, as yet unknown, OpenSearch plugins that require extensive work for FIPS-compliance

These could be actioned by the wider community

### Help Required
We have accessed and used:
  - YouTube videos on developing and contributing
[...]

PR #4381, opened 2024-05-30
 Replace BouncyCastle's OpenBSDBCrypt use with password4j for password hashing and verification

dancristiancecoi:
### Description
This change removes the usages of Bouncy Castle's OpenBSDBCrypt from the code and replaces it with a FIPS compliant library that supports additional hashing algorithms like PBKDF2, Argon2 and SCrypt. 

Furthermore it consolidates the password hashing and verification logic in one place.

**This change will require a security review.**

### Issues Resolved

- https://github.com/opensearch-project/security/issues/4361

### Related Issues

- https://github.com/opensearch-project/security/issues/4079 (this is the first stage for this issue. I will be raising further PR's to add extra configuration for BCrypt & support for additional hashing algorithms once this gets merged)
- https://github.com/opensearch-project/security/issues/3420

#### Various authentication attempts against a local 3.x deploy

- Can create users when password is supplied
- Can create users when a hashed password is supplied
- Can authenticate an user with a correct password
- Authentication fails when the password is wrong
- hash.sh script still works

```
(base) mtldce@SAS-1B16WV3:~$ curl -k https://localhost:9200 -u admin:$PASSWORD
{
  "name" : "SAS-1B16WV3",
  "cluster_name" : "opensearch",
  "cluster_uuid" : "XE_PodgCSZCaQQjHhWzlJA",
  "version" : {
    "distribution" : "opensearch",
    "number" : "3.0.0-SNAPSHOT",
    "build_type" : "tar",
    "build_hash" : "db5240e06dc08e7d7d03432595bb4d93b0e2e32d",
    "build_date" : "2024-05-30T10:46:41.778575385Z",
    "build_snapshot" : true,
    "lucene_version" : "9.11.0",
    "minimum_wire_compatibility_version" : "2.15.0",
    "minimum_index_compatibility_version" : "2.0.0"
  },
  "tagline" : "The OpenSearch Project: https://opensearch.org/"
}

(base) mtldce@SAS-1B16WV3:~$ curl -k https://localhost:9200 -u admin:$WRONG_PASSWORD
Unauthorized

(base) mtldce@SAS-1B16WV3:~$ curl -X PUT -k -s -H "Content-Type: application/json" "https://localhost:9200/_opendistro/_security/api/internalusers/dancecoi" -u admin:$PASSWORD -d $'{"password": "strongpassword1996!","opendistro_security_roles": ["all_access"],"backend_roles": ["all_access"],"attributes": {"attribute1": "value1","attribute2": "value2"}}'
{"status":"CREATED","message":"'dancecoi' created."}


(base) mtldce@SAS-1B16WV3:~$  curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoi:strongpassword1996!
health status index                        uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   security-auditlog-2024.05.30 ihLqFsJQToS1Jih90b_2Dg   1   1          4            0     57.5kb         57.5kb
green  open   .opendistro_security         w9zyt-LAT3CcwEqsL1p2TQ   1   0         10            0     88.6kb         88.6kb

(base) mtldce@SAS-1B16WV3:~$ curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoi:wrong_password
Unauthorized(base)


(base) mtldce@SAS-1B16WV3:~$ curl -X PUT -k -s -H "Content-Type: application/json" "https://localhost:9200/_opendistro/_security/api/internalusers/dancecoihash" -u admin:$PASSWORD -d $'{"hash": "$2y$12$gdh2ecVBQmwpmcAeyReicuNtXyR6GMWSfXHxtcBBqFeFz2VQ8kDZe","opendistro_security_roles": ["all_access"],"backend_roles": ["all_access"],"attributes": {"attribute1": "value1","attribute2": "value2"}}'
{"status":"CREATED","message":"'dancecoihash' created."}


(base) mtldce@SAS-1B16WV3:~$ curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoihash:testPassword
health status index                        uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   security-auditlog-2024.05.30 ihLqFsJQToS1Jih90b_2Dg   1   1          6            0     87.5kb         87.5kb
green  open   .opendistro_security         w9zyt-LAT3CcwEqsL1p2TQ   1   0         10            1    114.6kb        114.6kb

(base) mtldce@SAS-1B16WV3:~$ curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoihash:wrongTestPassword
Unauthorized(base)

(base) mtldce@SAS-1B16WV3:~/Projects/forks/OpenSearch/build/distribution/local/opensearch-3.0.0-SNAPSHOT/plugins/opensearch-security/tools$ ./hash.sh 
WARNING: nor OPENSEARCH_JAVA_HOME nor JAVA_HOME is set, will use /usr/bin/java
[Password:] 
$2y$12$2/lx4CpmaotLUtfa.vyiCOw3.1T6pgiPaql6l49pFF8m43ZAlJcRO


curl -X PUT -k -s -H "Content-Type: application/json" "https://localhost:9200/_opendistro/_security/api/internalusers/dancecoihash2" -u admin:$PASSWORD -d $'{"hash": "$2y$12$2/lx4CpmaotLUtfa.vyiCOw3.1T6pgiPaql6l49pFF8m43ZAlJcRO","opendistro_security_roles": ["all_access"],"backend_roles": ["all_access"],"attributes": {"attribute1": "value1","attribute2": "value2"}}' "value2"}}'
{"status":"CREATED","message":"'dancecoihash2' created."}

mtldce@SAS-1B16WV3:~$ curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoihash2:geforce_rtx_3070!
health status index                        uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   security-auditlog-2024.05.30 ihLqFsJQToS1Jih90b_2Dg   1   1          8            0    117.6kb        117.6kb
green  open   .opendistro_security         w9zyt-LAT3CcwEqsL1p2TQ   1   0         10            2    127.8kb        127.8kb

(base) mtldce@SAS-1B16WV3:~$ curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoihash2:wrong_password
Unauthorized
```

##### Pre-upgrade:

- Created an user

```
curl -X PUT -k -s -H "Content-Type: application/json" "https://localhost:9200/_opendistro/_security/api/internalusers/dancecoisas" -u $ADMINUSER:$PASSWORD -d $'
{"password": "strongpassword1996!","opendistro_security_roles": ["all_access"],"backend_roles": ["all_access"],"attributes": {"attribute1": "value1","attribute2": "value2"}}'
{"status":"CREATED","message":"'dancecoisas' created."}

curl -k -s "https://localhost:9200/" -u $ADMINUSER:$PASSWORD
{
  "name" : "sas-opendistro-default-0",
  "cluster_name" : "sas-opendistro",
  "cluster_uuid" : "arjz9067TKSANhk48Xg6JQ",
  "version" : {
    "distribution" : "opensearch",
    "number" : "2.11.0",
    "build_type" : "tar",
    "build_hash" : "4dcad6dd1fd45b6bd91f041a041829c8687278fa",
    "build_date" : "2023-10-13T02:55:55.511945994Z",
    "build_snapshot" : false,
    "lucene_version" : "9.7.0",
    "minimum_wire_compatibility_version" : "7.10.0",
    "minimum_index_compatibility_version" : "7.0.0"
  },
  "tagline" : "The OpenSearch Project: https://opensearch.org/"
}

```
##### Post-upgrade:

- Passwords previously hashed with OpenBSDBCrypt still work as expected
- Persisted users can still authenticate as expected

```
curl -k -s "https://localhost:9200/" -u $ADMINUSER:$PASSWORD
{
  "name" : "sas-opendistro-default-0",
  "cluster_name" : "sas-opendistro",
  "cluster_uuid" : "arjz9067TKSANhk48Xg6JQ",
  "version" : {
    "distribution" : "opensearch",
    "number" : "2.13.0",
    "build_type" : "tar",
    "build_hash" : "7ec678d1b7c87d6e779fdef94e33623e1f1e2647",
    "build_date" : "2024-03-26T00:02:39.659767978Z",
    "build_snapshot" : false,
    "lucene_version" : "9.10.0",
    "minimum_wire_compatibility_version" : "7.10.0",
    "minimum_index_compatibility_version" : "7.0.0"
  },
  "tagline" : "The OpenSearch Project: https://opensearch.org/"
}

curl -k -s "https://localhost:9200/_cat/indices" -u dancecoisas:strongpassword1996!
Defaulted container "sas-opendistro" out of: sas-opendistro, sysctl (init), sas-certframe (init)
green  open .plugins-ml-config               A3U0wNFnS5WOrC2yUrGtKg 1 0   1 0   3.9kb   3.9kb
green  open .opensearch-observability        SPlW4oIsQ5yDO7qT5xuFrA 1 0   0 0    208b    208b
yellow open security-auditlog-2024.05.30     VLtGQpMrS1en6CyEMj-VMA 1 1   3 0    47kb    47kb
green  open .opensearch-sap-log-types-config 7rNwKfsWRImllcPVTkbgbA 1 0 452 0 101.6kb 101.6kb
green  open .opendistro_security             3q1xOXU_SZGg1ud34p_K_Q 1 0   9 0  24.4kb  24.4kb

curl -k -s "https://localhost:9200/_cat/indices" -u dancecoisas:wrongpassword

(base) curl -k -s "https://localhost:9200/_cat/indices" -u dancecoisas:wrongpassword

```

##### Extra: 
- Added unit tests
- The initial GitHub Checks are passing

### Check List


dancristiancecoi:
Integration tests were failing due to pre-set hashes in various internal_users.yml files having a different configuration to default one (BCrypt.A with 4 log rounds vs B.Crypt Y with 12 log rounds).  I've adapted the code so it verifies the passwords with a BCrypt configuration based on the hash.

src/main/java/org/opensearch/security/tools/Hasher.java
```diff
@@ -82,11 +81,7 @@ public static void main(final String[] args) {
     }
 
     public static String hash(final char[] clearTextPassword) {
-        final byte[] salt = new byte[16];
-        new SecureRandom().nextBytes(salt);
-        final String hash = OpenBSDBCrypt.generate((Objects.requireNonNull(clearTextPassword)), salt, 12);
-        Arrays.fill(salt, (byte) 0);
-        Arrays.fill(clearTextPassword, '\0');
-        return hash;
+        PasswordHasher passwordHasher = new BCryptPasswordHasher();
```
cwperks, line 84: Is it possible to this tool to produce an output based on the configured password hashing strategy?

The purpose of the [`hash.sh` tool](https://github.com/opensearch-project/security/blob/main/tools/hash.sh#L35) is to help users configure security. Using the tool allows a user to pass a plaintext password and get a bcrypt hash of the password to put in `internal_users.yml` before starting a cluster for the first time.

dancristiancecoi: Yes, I have something in the works for this but I will push this as part of a future PR where we add extra BCrypt configuration. Initially I will be adding extra command line arguments for the users to be able to configure the version of BCrypt and the log rounds. Afterwards that will be expanded to support other hashing algorithms and the corresponding parameters for each alg.

cwperks: Got it. I re-read the PR description and see that this change only changes the BCrypt implementation from BC -> Password4j. I look forward to seeing the future PRs as well!

src/main/java/org/opensearch/security/hasher/BCryptPasswordHasher.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private void cleanup(CharBuffer password) {
+        password.clear();
+        char[] passwordOverwrite = new char[password.capacity()];
+        Arrays.fill(passwordOverwrite, '\0');
+        password.put(passwordOverwrite);
+    }
+
+    private HashingFunction getDefaultBCryptFunction() {
+        return BcryptFunction.getInstance(Bcrypt.Y, 12);
```
cwperks, line 78: nit: Can this be a `private static` constant defined at the top of the class? Maybe `private static HasingFunction DEFAULT_BCRYPT_FUNCTION = BcryptFunction.getInstance(Bcrypt.Y, 12);`

src/main/java/org/opensearch/security/user/UserService.java
```diff
@@ -97,6 +98,7 @@ private static CType getUserConfigName() {
     public UserService(ClusterService clusterService, ConfigurationRepository configurationRepository, Settings settings, Client client) {
         this.clusterService = clusterService;
         this.configurationRepository = configurationRepository;
+        this.passwordHasher = new BCryptPasswordHasher();
```
cwperks, line 101: I think we may want to have a central place where this is instantiated once and passed to components that need it. Can we look into instantiating this within [OpenSearchSecurityPlugin.createComponents](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L1017-L1030)?

src/test/java/org/opensearch/security/dlic/rest/api/AccountApiActionConfigValidationsTest.java
```diff
@@ -56,13 +54,13 @@ public void updatePassword() {
...
-        assertTrue(OpenBSDBCrypt.checkPassword(u.getHash(), "cccccc".toCharArray()));
+        assertTrue(passwordHasher.check("cccccc".toCharArray(), u.getHash()));
 
         requestContent.remove("password");
-        requestContent.put("hash", Utils.hash("dddddd".toCharArray()));
+        requestContent.put("hash", passwordHasher.hash("dddddd".toCharArray()));
         result = accountApiAction.updatePassword(SecurityConfiguration.of(requestContent, "u", configuration));
         assertTrue(result.isValid());
-        assertTrue(OpenBSDBCrypt.checkPassword(u.getHash(), "dddddd".toCharArray()));
+        assertTrue(passwordHasher.check("dddddd".toCharArray(), u.getHash()));
```
willyborankin, line 63: AFAIK we started to use `hamcrest`. So lets use `assertThat(...)` from it instead

src/integrationTest/java/org/opensearch/test/framework/TestSecurityConfig.java
```diff
@@ -967,12 +967,7 @@ public void updateInternalUsersConfiguration(Client client, List<User> users) {
     }
 
     static String hash(final char[] clearTextPassword) {
-        final byte[] salt = new byte[16];
-        new SecureRandom().nextBytes(salt);
-        final String hash = OpenBSDBCrypt.generate((Objects.requireNonNull(clearTextPassword)), salt, 12);
-        Arrays.fill(salt, (byte) 0);
-        Arrays.fill(clearTextPassword, '\0');
-        return hash;
+        return passwordHasher.hash((Objects.requireNonNull(clearTextPassword)));
```
willyborankin, line 970: No need to use `Objects.requireNonNull` here AFAIU `passwordHasher.hash` already checks on the null value and throws `OpenSearchException`.

src/main/java/org/opensearch/security/hasher/BCryptPasswordHasher.java
```diff
@@ -0,0 +1,84 @@
...
+import com.password4j.BcryptFunction;
+import com.password4j.HashingFunction;
+import com.password4j.Password;
+import com.password4j.types.Bcrypt;
+
+public class BCryptPasswordHasher implements PasswordHasher {
+
+    @Override
+    public String hash(char[] password) {
+        if (password == null || password.length == 0) {
```
willyborankin, line 31: Lest use OS API for checking the `password` variable:
`org.opensearch.core.common.Strings.isNullOrEmpty`

src/main/java/org/opensearch/security/hasher/BCryptPasswordHasher.java
```diff
@@ -0,0 +1,84 @@
...
+                (PrivilegedAction<String>) () -> Password.hash(passwordBuffer).with(getDefaultBCryptFunction()).getResult()
+            );
+        } finally {
+            cleanup(passwordBuffer);
+        }
+    }
+
+    @Override
+    public boolean check(char[] password, String hash) {
+        if (password == null || password.length == 0) {
```
willyborankin, line 50: Same here: `org.opensearch.core.common.Strings.isNullOrEmpty`

src/integrationTest/java/org/opensearch/test/framework/TestSecurityConfig.java
```diff
@@ -77,8 +77,8 @@
 public class TestSecurityConfig {
 
     private static final Logger log = LogManager.getLogger(TestSecurityConfig.class);
-
-    public final static String REST_ADMIN_REST_API_ACCESS = "rest_admin__rest_api_access";
+    public static final String REST_ADMIN_REST_API_ACCESS = "rest_admin__rest_api_access";
```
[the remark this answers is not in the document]
dancristiancecoi: On this line I've only flipped the "final static" keywords around :D. But its a good point, made it private!

src/integrationTest/java/org/opensearch/test/framework/TestSecurityConfig.java
dancristiancecoi: Done!

src/main/java/org/opensearch/security/hasher/BCryptPasswordHasher.java
dancristiancecoi: That would require converting to a String first which I am trying to avoid. However will use it for checking the hash variable

src/main/java/org/opensearch/security/user/UserService.java
dancristiancecoi: That makes sense, thank you!

src/main/java/org/opensearch/security/hasher/BCryptPasswordHasher.java
willyborankin: Ahhh sorry, it uses `char[]` did not notice.

src/integrationTest/java/org/opensearch/test/framework/TestSecurityConfig.java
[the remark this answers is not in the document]
dancristiancecoi: Actually looks like it is imported in a test class outside of the package, so reverted it to public for now. We can modify it in a separate PR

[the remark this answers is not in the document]
willyborankin: OK got it. No problem.

willyborankin (approved):
LGTM

src/main/java/org/opensearch/security/hasher/PasswordHasher.java
```diff
@@ -0,0 +1,19 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.hasher;
+
+public interface PasswordHasher {
```
DarshitChanpura, line 14: is there a plan to implement multiple Password Hashers?

src/test/java/org/opensearch/security/hasher/BCryptPasswordHasherTests.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Test
+    public void shouldCleanupPasswordCharBuffer() {
+        char[] password = new char[] { 'p', 'a', 's', 's', 'w', 'o', 'r', 'd' };
+        CharBuffer passwordBuffer = CharBuffer.wrap(password);
+        passwordHasher.hash(password);
+        assertThat("\0\0\0\0\0\0\0\0", is(new String(password)));
+        assertThat("\0\0\0\0\0\0\0\0", is(passwordBuffer.toString()));
+    }
+}
```
DarshitChanpura, line 101: nice 👏

src/main/java/org/opensearch/security/hasher/PasswordHasher.java
dancristiancecoi: Yes, in the upcoming PR's!

src/main/java/org/opensearch/security/hasher/BCryptPasswordHasher.java
```diff
@@ -0,0 +1,84 @@
...
+import org.opensearch.SpecialPermission;
+
+import com.password4j.BcryptFunction;
+import com.password4j.HashingFunction;
+import com.password4j.Password;
+import com.password4j.types.Bcrypt;
+
+import static org.opensearch.core.common.Strings.isNullOrEmpty;
+
+public class BCryptPasswordHasher implements PasswordHasher {
```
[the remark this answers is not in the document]
dancristiancecoi: The HashingFunction itself is implemented as a Singleton in Password4J so I am not sure it will add much benefit to make this class singleton too. Outside of testing it only gets instantiated once. There might be other factors I am not considering though.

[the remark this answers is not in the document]
DarshitChanpura: > Outside of testing it only gets instantiated once

This might be a good reason to have it as a singleton instance.

build.gradle
```diff
@@ -582,7 +582,7 @@ dependencies {
     implementation 'org.ldaptive:ldaptive:1.2.3'
     implementation 'com.nimbusds:nimbus-jose-jwt:9.39.1'
     implementation 'com.rfksystems:blake2b:2.0.0'
-
+    implementation 'com.password4j:password4j:1.8.2'
```
shikharj05, line 585: should we remove bouncycastle dependency from here?

shikharj05 (commented):
Thanks for the PR. Quick questions -
1. How does this look like for end-users? Doesn't it break existing logins?
2. Any performance numbers we can get to understand the difference?

dancristiancecoi:
> Thanks for the PR. Quick questions -
> 
> 1. How does this look like for end-users? Doesn't it break existing logins?


1. No difference for end-users as the existing hashes still get verified!


> 2. Any performance numbers we can get to understand the difference?

I do not have any performance numbers but I can run some if required. What will be the best way to do it? Should I use the opensearch-benchmark tool locally or is there a github action we can run?

build.gradle
dancristiancecoi: Afaik BouncyCastle is still used in the Security plugin. It will be fully removed in the final stages of the FIPS work

src/main/java/org/opensearch/security/hasher/BCryptPasswordHasher.java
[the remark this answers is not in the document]
dancristiancecoi: @DarshitChanpura 

Though in the current form it is quite simple to change it to singleton, my reluctance is due to how the changes introduced in this PR will evolve once we bring in configurability for BCrypt and we start supporting other hashing algorithms. 

To elaborate: 

* BCryptPasswordHasher (and other implementations of PasswordHasher) will be modified so it takes in parameters for the hashing algorithms. Though it is possible to implement Singleton that can handle parameters, it will add extra complexity and it won't be an elegant solution.
* A factory (or something of the kind) will be implemented to handle instantiation of a specific PasswordHasher based on the user's configuration. Combining the two, while possible, is again extra added complexity.

Considering the above & that there is no harm in instantiating the PasswordHasher multiple times I personally dont think the extra complexity in code and in testing is worth it in this case.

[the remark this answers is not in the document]
DarshitChanpura: okay, in a case where there are plans to make this configurable, we can leave it as is.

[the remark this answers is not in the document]
dancristiancecoi: sounds good, thanks a lot for the review!

dancristiancecoi:
@shikharj05 I've ran the benchmarking tool with the percolator workload. It was against a one node local cluster running on my laptop so I am not sure how representative the results are :D. 

For comparison I also attached a benchmark run in similar conditions but against the latest code in main.

[testRunAgainstChanges.txt](https://github.com/user-attachments/files/15623735/testRunAgainstChanges.txt)
[testRunAgainstMain.txt](https://github.com/user-attachments/files/15623739/testRunAgainstMain.txt)

cwperks:
This change LGTM! I am also looking at the benchmarking tool to look at ways to measure the impact on performance for change this library.

There is a performance test suite that's in [Draft](https://github.com/opensearch-project/security/pull/4282) and not unavailable yet unfortunately.

dancristiancecoi:
> This change LGTM! I am also looking at the benchmarking tool to look at ways to measure the impact on performance for the change in library.
> 
> There is a performance test suite that's in [Draft](https://github.com/opensearch-project/security/pull/4282) and not unavailable yet unfortunately.

Awesome,  thanks @cwperks!

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -199,8 +201,9 @@ void updateAdminPassword() throws IOException {
      */
     private boolean isAdminPasswordSetToAdmin(String internalUsersFile) throws IOException {
         JsonNode internalUsers = YAML_MAPPER.readTree(new FileInputStream(internalUsersFile));
+        PasswordHasher passwordHasher = new BCryptPasswordHasher();
         return internalUsers.has("admin")
-            && OpenBSDBCrypt.checkPassword(internalUsers.get("admin").get("hash").asText(), DEFAULT_ADMIN_PASSWORD.toCharArray());
+            && passwordHasher.check(DEFAULT_ADMIN_PASSWORD.toCharArray(), internalUsers.get("admin").get("hash").asText());
```
derek-ho, line 206: I'm curious as to why the order is swapped around here, is that the standard/could we keep it consistent with the OpenBSDBCrypt parameter ordering?

dancristiancecoi: Don't think there is a standard as I've seen them ordered either way depending on the library. I've put the password parameter first to keep it consistent with the verification function in Password4j (and to me it seems more intuitive) however I can certainly flip them around if required!

DarshitChanpura (approved):
@dancristiancecoi Changes look good to me! I don't have any reservations against merging this as long as there is some sort of benchmarking done once the #4282 is merged. Once everything is implemented we can ship it in 2.x line.

src/test/java/org/opensearch/security/hasher/BCryptPasswordHasherTests.java
```diff
@@ -0,0 +1,101 @@
...
+        assertThat(passwordHasher.check(wrongPassword.toCharArray(), hashedPassword), is(false));
+
+    }
+
+    /**
+     * Ensures that the hashes that were previously created by OpenBSDBCrypt are still valid
+     */
+    @Test
+    public void shouldBeBackwardsCompatible() {
+        String legacyHash = "$2y$12$gdh2ecVBQmwpmcAeyReicuNtXyR6GMWSfXHxtcBBqFeFz2VQ8kDZe";
```
derek-ho, line 50: For my learning where did you get this from? Is this hard coded somewhere as the default legacy hash used by Bcrypt?

derek-ho (approved):
LGTM

src/test/java/org/opensearch/security/hasher/BCryptPasswordHasherTests.java
shikharj05: nit: we should document how this has is constructed - to keep track of this test for further references.

shikharj05 (approved):

shikharj05:
> @shikharj05 I've ran the benchmarking tool with the percolator workload. It was against a one node local cluster running on my laptop so I am not sure how representative the results are :D.
> 
> For comparison I also attached a benchmark run in similar conditions but against the latest code in main.
> 
> [testRunAgainstChanges.txt](https://github.com/user-attachments/files/15623735/testRunAgainstChanges.txt) [testRunAgainstMain.txt](https://github.com/user-attachments/files/15623739/testRunAgainstMain.txt)
> 
> **Run 2:**
> 
> [testRunAgainstMain-run2.txt](https://github.com/user-attachments/files/15647673/testRunAgainstMain-run2.txt) [testRunAgainstMyChanges-run2.txt](https://github.com/user-attachments/files/15647674/testRunAgainstMyChanges-run2.txt)
> 
> To me it doesn't look like there is a significant variation either way in these bench-marking results.

Thanks, looked at the results, Run-1 shows some drop in throughput for ``index``, however in run-2 it matches up. LGTM for now!

src/test/java/org/opensearch/security/hasher/BCryptPasswordHasherTests.java
dancristiancecoi: @derek-ho I just used OpenBSDBCrypt to hash "password" with the default parameters. 

In BCrypt every hash will be unique for the same input. However, every implementation of BCrypt should be able to verify any BCrypt hashes when specifying the same parameters. (In this case, if you look at the prefix $2y$12 it tells you that it was generated with the "2y" version of the algorithm with 12 log rounds)

dancristiancecoi: > nit: we should document how this has is constructed - to keep track of this test for further references.

I added a comment above the unit test saying "Ensures that the hashes that were previously created by OpenBSDBCrypt are still valid". Should I clarify more?

shikharj05: Thanks, looks good for now!

willyborankin (approved):

willyborankin:
@dancristiancecoi cpuld you please prepare a manual backport?

dancristiancecoi:
> @dancristiancecoi could you please prepare a manual backport?

Sure! https://github.com/opensearch-project/security/pull/4428

dancristiancecoi:
Thanks everyone for the reviews!

ruanyl:
> Integration tests were failing due to pre-set hashes in various internal_users.yml files used for integration tests having a different configuration to the default one (BCrypt.A with 4 log rounds vs BCrypt.Y with 12 log rounds). I've adapted the code so it verifies the passwords with a BCrypt configuration based on the hash.

Hi @dancristiancecoi I'm not able to start my cluster with old `internal_users.yml`, is there documentations that I can take a look on fix this?

dancristiancecoi:
Hi @ruanyl. 

Are there any other errors in the logs to pinpoint to why the Security plugin was not initialised? 

When you mention "old internal-users.yml file" is it similar to the ones used in the integration tests / demo config ? Anything that's unique about them? 

There isn't documentation for this change as it SHOULD have worked seamlessly but it's very possible we've missed something

ruanyl:
Hi @dancristiancecoi Thanks, I think I figured it out, just need to regenerate the password hash, then it works :)

Issue #4177, opened 2024-03-27
[BUG] User Injection is failing for IPV6 addresses


devardee:
**What is the bug?**
User Injection is a construct in security plugin, which will allow other plugins to enforce authorization, by setting the User Information in thread context. Security plugin will read this information and initialize a User object and apply authorization if needed. User Injected String has the following pattern :

```
"user|backendRole|remoteIpPort|customAttributes|tenant"
```
in the `remoteIpPort` part, For IPV6 address, we are observing the following exception being thrown : https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/UserInjector.java#L99

**How can one reproduce the bug?**
Added a simple UT here : https://github.com/devardee/security-1/commit/c47d9c0339899a9d544631f99484e71fa8ab59cd

**What is the expected behavior?**
Security Plugin should correctly instantiate User object, when the Injected User has IPV6 address.

**What is your host/environment?**
NA

**Do you have any screenshots?**
NA

**Do you have any additional context?**
No

peternied:
@devardee Thanks for filing this issue and creating a reproduction.  What do you think about creating a pull request to fix this?

scrawfor99:
[Triage] Thanks for filing this issue @devardee. Looks like you are going to take care of this with a PR. Thank you! Marking as triaged.

PR #4399, opened 2024-06-03
Add support for ipv6 ip address in user injection

derek-ho:
### Description
Enhances the parsing logic of user injection to support ipv6 addresses.
Bug fix
Issue: https://github.com/opensearch-project/security/issues/4177

Ipv6 IP addresses would throw a parsing error, now it doesn't

### Issues Resolved
Fix: https://github.com/opensearch-project/security/issues/4177

No

### Testing
Added unit test

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/auth/UserInjector.java
+++ b/src/main/java/org/opensearch/security/auth/UserInjector.java
@@ -93,14 +93,16 @@ public class UserInjector {
         }
 
         public void setTransportAddress(String addr) throws UnknownHostException, IllegalArgumentException {
-            // format is ip:port
-            String[] ipAndPort = addr.split(":");
-            if (ipAndPort.length != 2) {
+            int lastColonIndex = addr.lastIndexOf(':');
+            if (lastColonIndex == -1) {
                 throw new IllegalArgumentException("Remote address must have format ip:port");
             }
 
-            InetAddress iAdress = InetAddress.getByName(ipAndPort[0]);
-            int port = Integer.parseInt(ipAndPort[1]);
+            String ip = addr.substring(0, lastColonIndex);
+            String portString = addr.substring(lastColonIndex + 1);
+
+            InetAddress iAdress = InetAddress.getByName(ip);
+            int port = Integer.parseInt(portString);
 
             this.transportAddress = new TransportAddress(iAdress, port);
         }
--- a/src/main/java/org/opensearch/security/auth/UserInjector.java
+++ b/src/main/java/org/opensearch/security/auth/UserInjector.java
@@ -67,7 +67,7 @@ public class UserInjector {
     }
 
     public static class InjectedUser extends User {
-        transient TransportAddress transportAddress;
+        private transient TransportAddress transportAddress;
 
         public InjectedUser(String name) {
             super(name);
--- a/src/main/java/org/opensearch/security/auth/UserInjector.java
+++ b/src/main/java/org/opensearch/security/auth/UserInjector.java
@@ -67,7 +67,7 @@ public class UserInjector {
     }
 
     public static class InjectedUser extends User {
-        private transient TransportAddress transportAddress;
+        transient TransportAddress transportAddress;
 
         public InjectedUser(String name) {
             super(name);
--- a/src/test/java/org/opensearch/security/auth/UserInjectorTest.java
+++ b/src/test/java/org/opensearch/security/auth/UserInjectorTest.java
@@ -65,6 +65,35 @@ public class UserInjectorTest {
     }
 
     @Test
+    public void testValidInjectUserIpV6() {
+        HashSet<String> roles = new HashSet<>();
+        roles.addAll(Arrays.asList("role1", "role2"));
+        threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER, "user|role1,role2|2001:db8:3333:4444:5555:6666:7777:8888:9200");
+        User injectedUser = userInjector.getInjectedUser();
+        assertEquals(injectedUser.getName(), "user");
+        assertEquals(injectedUser.getRoles(), roles);
+    }
+
+    @Test
+    public void testInvalidInjectUserIpV6() {
+        HashSet<String> roles = new HashSet<>();
+        roles.addAll(Arrays.asList("role1", "role2"));
+        threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER, "user|role1,role2|2001:db8:3333:5555:6666:7777:8888:9200");
+        User injectedUser = userInjector.getInjectedUser();
+        assertNull(injectedUser);
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


DarshitChanpura (commented):
The change looks good. Is there an issue associate with this? If not would you mind filing one.

src/test/java/org/opensearch/security/auth/UserInjectorTest.java
```diff
@@ -64,6 +64,35 @@ public void testValidInjectUser() {
...
+        threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER, "user|role1,role2|2001:db8:3333:5555:6666:7777:8888:9200");
+        User injectedUser = userInjector.getInjectedUser();
+        assertNull(injectedUser);
+    }
+
+    @Test
+    public void testValidInjectUserBracketsIpV6() {
+        HashSet<String> roles = new HashSet<>();
+        roles.addAll(Arrays.asList("role1", "role2"));
+        threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER, "user|role1,role2|[2001:db8:3333:4444:5555:6666:7777:8888]:9200");
```
derek-ho, line 90: I couldn't find a clear answer online on how a ipv6 address might look like with a port. Some sources say something like `[ipv6 address]:port` while others just have `ipv6 address:port`. It seems like the library handles both these cases, but I added a test just to verify this.

src/main/java/org/opensearch/security/auth/UserInjector.java
```diff
@@ -93,14 +93,16 @@ public TransportAddress getTransportAddress() {
         }
 
         public void setTransportAddress(String addr) throws UnknownHostException, IllegalArgumentException {
-            // format is ip:port
-            String[] ipAndPort = addr.split(":");
-            if (ipAndPort.length != 2) {
+            int lastColonIndex = addr.lastIndexOf(':');
+            if (lastColonIndex == -1) {
```
derek-ho, line 97: Side question: are we always guaranteed to get a port with the IP address? If not then this assumption/check would probably fail. However, this is consistent with the behavior today so it is probably fine.

DarshitChanpura: Can we add a more robust check than relying on last index of colon since IPv6 addresses also contain colon in them. My suggestion here would to use regex-matching to allow for ports to be read correctly if present and not if not present.

derek-ho: My thinking behind doing it this way is that we do not need to re-invent the wheel. Since the library is already properly parsing an IP in either v4 or v6 format, all we need to do is to extract the supposed port and assume the rest of the string is the address. If it is malformed in any way the library would catch this, assuming that the port is always present (which the existing code does). Does that make sense?

cwperks: This looks fine to me and User Injection will always have port. See comment [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/UserInjector.java#L128). If the port doesn't exist, throw an IllegalArgumentException. The call to [Integer.parseInt](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/UserInjector.java#L103) will also ensure that its a valid integer value. You could get even more restrictive and check that its between 1–65535, but that is probably overkill.

src/test/java/org/opensearch/security/auth/UserInjectorTest.java
```diff
@@ -64,6 +64,44 @@ public void testValidInjectUser() {
...
+        assertNull(injectedUser);
+    }
+
+    @Test
+    public void testValidInjectUserBracketsIpV6() {
+        HashSet<String> roles = new HashSet<>();
+        roles.addAll(Arrays.asList("role1", "role2"));
+        threadContext.putTransient(
+            ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER,
+            "user|role1,role2|[2001:db8:3333:4444:5555:6666:7777:8888]:9200"
```
cwperks, line 98: Oh, nice. I should have finished reading the tests first.

src/test/java/org/opensearch/security/auth/UserInjectorTest.java
```diff
@@ -64,6 +64,44 @@ public void testValidInjectUser() {
...
+    @Test
+    public void testValidInjectUserIpV6() {
+        HashSet<String> roles = new HashSet<>();
+        roles.addAll(Arrays.asList("role1", "role2"));
+        threadContext.putTransient(
+            ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER,
+            "user|role1,role2|2001:db8:3333:4444:5555:6666:7777:8888:9200"
+        );
+        User injectedUser = userInjector.getInjectedUser();
+        assertEquals(injectedUser.getName(), "user");
```
cwperks, line 76: Can we also add assertions to ensure that it parses the expected port?

derek-ho: Done in: [7120a8d](https://github.com/opensearch-project/security/pull/4399/commits/7120a8d21d8805ec9d9240ce8d65a1436d45b85a)

src/test/java/org/opensearch/security/auth/UserInjectorTest.java
```diff
@@ -64,6 +64,47 @@ public void testValidInjectUser() {
...
+        HashSet<String> roles = new HashSet<>();
+        roles.addAll(Arrays.asList("role1", "role2"));
+        threadContext.putTransient(
+            ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER,
+            "user|role1,role2|2001:db8:3333:4444:5555:6666:7777:8888:9200"
+        );
+        UserInjector.InjectedUser injectedUser = userInjector.getInjectedUser();
+        assertEquals("user", injectedUser.getName());
+        assertEquals(9200, injectedUser.getTransportAddress().getPort());
+        assertEquals("2001:db8:3333:4444:5555:6666:7777:8888", injectedUser.getTransportAddress().getAddress());
```
shikharj05, line 78: can you add test-cases for shortened ipv6 format- https://en.wikipedia.org/wiki/IPv6_address#Representation

derek-ho: Added one here: [a680e92](https://github.com/opensearch-project/security/pull/4399/commits/a680e921e3e5f866a5f91a05883bf547309e36ee). I tried to use an address that would cover the two scenarios I can think of: a sequence of 0's shortened to `::`, as well as leading 0's cut off (`0001` shortened to `1`).

DarshitChanpura (approved):

cwperks (approved):

shikharj05 (approved):

cwperks:
DCO check is stuck. I am bypassing and manually verified that all commits are signed.

PR #4380, opened 2024-05-30
Optimized Privilege Evaluation

nibix:
### Description

This implements the optimized privilege evaluation as described in #3870. 

This introduces de-normalized data structures that are optimized for the checks that need to be done during privilege evaluation. Additionally, certain objects  (like DLS queries) are prepared ahead of time, as early as possible in order to minimize the overhead during actual privilege evaluation.

This is a big change set - in order to facilitate the review, I have split it into three major commits:

- Optimized action privilege evaluation
- Optimized DLS/FLS/FM privilege evaluation
- Removal of unused code

The code is extensively commented - I hope that will help during review.

* Category: Enhancement
* Why these changes are required? 

Performance tests indicate that the OpenSearch security layer adds a noticeable overhead to the indexing throughput of an OpenSearch cluster. The overhead may vary depending on the number of indices, the use of aliases, the number of roles and the size of the user object. The goal of these changes is to improve privilege evaluation performance and to make it less dependent on the number of indices, etc.

* What is the old behavior before changes and new behavior after changes? 

No significant behavioral changes in the "happy case", when privileges are present.

The undocumented config option `config.dynamic.multi_rolespan_enabled` is no longer evaluated. The code now behaves like it is always set to `true`  - that is the former default. See #4495 for details.

Some slight changes are present in error cases:

- More detailed error messages for missing privileges, showing a index/action matrix of missing privileges
- Errors in the role configuration might be reported (as error log messages) more early, directly after the configuration was applied
- The DLS/FLS implementation now defaults to a "deny by default" implementation. *This is not relevant for normal cases.* This will be only relevant if index requests pass through privileges evaluator even though there are no roles which grant privileges to the requested indices. *Note: This would only happen in case of a bug in the code*. In the previous versions, the DLS/FLS implementation would grant full access to the indices. Now, the DLS/FLS implementation acts as a second barrier, denying access to the indices.

### Issues Resolved

- #3870

### Testing

- Each new component is accompanied by its own unit test.
- The high-level functionality is validated by the existing integration tests
- Mixed cluster behavior is verfied by `SecurityBackwardsCompatibilityIT` (extended in #4817 )

### Check List
- [x] New functionality includes testing
- [x] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
[...]

PR #4416, opened 2024-06-08
Extracted the user attr handling methods from ConfigModelV7 into its own class

nibix:
### Description

This code change is just in preparation for the change in #4380 as requested in https://github.com/opensearch-project/security/pull/4380#discussion_r1621324272 .

The code for user attribute handling is moved from the class ConfigModelV7 into its own class, as other new code will need to use it. Additionally, it enhances the structuring of the code by moving code with a specific purpose into its own class.

Personal note: A thorough re-design of this code might be worthwhile.

* Category: Refactoring
* Why these changes are required? New code introduced in #4380 needs to use this code.

### Issues Resolved

- https://github.com/opensearch-project/security/pull/4380#discussion_r1621324272

### Testing

### Check List
- [ ] New functionality includes testing - no new functionality
- [ ] New functionality has been documented - no new functionality

### DIFF
--- a/src/main/java/org/opensearch/security/privileges/UserAttributes.java
+++ b/src/main/java/org/opensearch/security/privileges/UserAttributes.java
@@ -1,0 +1,69 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+package org.opensearch.security.privileges;
+
+import java.util.Map;
+import java.util.Set;
+
+import com.google.common.base.Joiner;
+import com.google.common.collect.Iterables;
+
+import org.opensearch.security.user.User;
+
+/**
+ * Support for interpolating user attributes used in index patterns and DLS queries.
+ *
+ * This code was moved over from ConfigModelV7.
+ */
+public class UserAttributes {
+    public static String replaceProperties(String orig, User user) {
+
+        if (user == null || orig == null) {
+            return orig;
+        }
+
+        orig = orig.replace("${user.name}", user.getName()).replace("${user_name}", user.getName());
+        orig = replaceRoles(orig, user);
+        orig = replaceSecurityRoles(orig, user);
+        for (Map.Entry<String, String> entry : user.getCustomAttributesMap().entrySet()) {
+            if (entry == null || entry.getKey() == null || entry.getValue() == null) {
+                continue;
+            }
+            orig = orig.replace("${" + entry.getKey() + "}", entry.getValue());
+            orig = orig.replace("${" + entry.getKey().replace('.', '_') + "}", entry.getValue());
+        }
+        return orig;
+    }
+
+    private static String replaceRoles(final String orig, final User user) {
+        String retVal = orig;
+        if (orig.contains("${user.roles}") || orig.contains("${user_roles}")) {
+            final String commaSeparatedRoles = toQuotedCommaSeparatedString(user.getRoles());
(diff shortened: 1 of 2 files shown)


shikharj05 (approved):
Thanks! I also like the idea of breaking https://github.com/opensearch-project/security/pull/4380 down into such PRs to make the review faster. Feel free to raise such incremental PRs :)

src/main/java/org/opensearch/security/privileges/UserAttributes.java
```diff
@@ -0,0 +1,69 @@
+/*
```
shikharj05, line 1: minor comment: As a TODO, we can now add specific UTs for this class.

src/main/java/org/opensearch/security/privileges/UserAttributes.java
```diff
@@ -0,0 +1,69 @@
...
+ * This code was moved over from ConfigModelV7.
+ */
+public class UserAttributes {
+    public static String replaceProperties(String orig, User user) {
+
+        if (user == null || orig == null) {
+            return orig;
+        }
+
+        orig = orig.replace("${user.name}", user.getName()).replace("${user_name}", user.getName());
```
shikharj05, line 33: minor: Placeholders can be moved to static constants above.

src/main/java/org/opensearch/security/privileges/UserAttributes.java
```diff
@@ -0,0 +1,69 @@
...
+            orig = orig.replace("${" + entry.getKey().replace('.', '_') + "}", entry.getValue());
+        }
+        return orig;
+    }
+
+    private static String replaceRoles(final String orig, final User user) {
+        String retVal = orig;
+        if (orig.contains("${user.roles}") || orig.contains("${user_roles}")) {
+            final String commaSeparatedRoles = toQuotedCommaSeparatedString(user.getRoles());
+            retVal = orig.replace("${user.roles}", commaSeparatedRoles).replace("${user_roles}", commaSeparatedRoles);
```
shikharj05, line 50: TODO: It would be good to check if multiple String.replace() calls can be merged into single StringBuilder.replace() and measure performance for our use-case. This should be possible once Security plugin has micro-benchmark functionality using JMH as proposed in https://github.com/opensearch-project/security/pull/4380

nibix: IMHO, it would be best to redesign the whole attributes handling to use pre-parsed templates instead of these simple string replacement operations.

This would have a couple of advantages:

- Performance
- One could detect the use of unknown attributes and log errors. At the moment, unknown attributes are just left as such in the resulting string.
- Possibility for support for multi-valued attributes. At the moment, multi-valued attributes are not really supported. If one wants to achieve that, some pre-fabricated strings in the correct syntax (JSON or index pattern) need to be defined in the auth backend.

This, however, should probably done in a separate effort.

src/main/java/org/opensearch/security/privileges/UserAttributes.java
cwperks: +1

cwperks (approved):

peternied (approved):
Since this is a refactor, let's just move in place.

@shikharj05 - want to make a follow up PR for some of those items you called out.

willyborankin (approved):

src/main/java/org/opensearch/security/privileges/UserAttributes.java
peternied: Lets not block merging for these follow ups

src/main/java/org/opensearch/security/privileges/UserAttributes.java
peternied: Lets not block merging for these follow ups

src/main/java/org/opensearch/security/privileges/UserAttributes.java
peternied: Lets not block merging for these follow ups

PR #4417, opened 2024-06-10
Pk add extensive dls fls tests

prabhask5:
### Description

### Issues Resolved

### Testing

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/DlsIntegrationTests.java
+++ b/src/integrationTest/java/org/opensearch/security/DlsIntegrationTests.java
@@ -19,6 +19,7 @@ import java.util.stream.Collectors;
 
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.cxf.common.i18n.Exception;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


prabhask5:
Ready for review!

Also related to @scrawfor99's comment on the last PR I made: 

> Hi @prabhask5, thanks for making those changes above. Looking at some of the definitions you have, it seems like the issue is because of the contradiction in some of the roles and the ordering of how the rules are applied. I can see where you tried to combine a role which can only see "titles" with a role which cannot see "titles." I think you thought that this would result in having a user which could not see any "titles" since that is more restrictive then the other role. I can definitely see where you are coming from...

> Unfortunately, things don't quite work like that. When DLS/FLS is computed, it is done so in an additive way. Here are some related discussions around DLS/FLS that may be of use for moving forward: https://github.com/opensearch-project/security/issues/2556. Basically, we end up taking a union of the roles and this can lead to weird behavior when the ordering of the rules is swapped. In this case, I suspect the read title option is being applied after the read none option meaning the final result is that the user can get the titles. I would have to step through your tests to be sure but from what I remember this is my best guess. You can try reordering the rules or redefining them to be more consistent.

Is there consensus that this is the best way to deal with overlapping roles? In these test scenarios I have overlapping roles that cause users to both only see the title field in a document and see every other field but title in a document. With the current functionality the combination of these roles leads to not a union, but just following the less restrictive role completely, leading to the output having all fields except the title field. But in the case that both of these roles are applied to the same user, wouldn't the additive property of the roles make more sense for all fields of the document to be visible here? I can work on refactoring the role filtering implementation if there is support for this idea. Also I believe that this additive implementation might fix issues like https://github.com/opensearch-project/security/issues/2556 since it would not depend on the order of overlapping roles.

prabhask5:
Also let me know how I can increase the code coverage more

src/integrationTest/java/org/opensearch/security/DlsIntegrationTests.java
```diff
@@ -605,6 +647,133 @@ public void testAggregateAndComputeStarRatings() throws IOException {
...
+            not(
+                searchHitsContainDocumentsInAnyOrder(
+                    FIRST_INDEX_SONGS_BY_ID.keySet()
+                        .stream()
+                        .filter(id -> !id.equals(FIRST_INDEX_ID_SONG_1))
+                        .map(id -> Pair.of(FIRST_INDEX_NAME, id))
+                        .collect(Collectors.toList())
+                )
+            )
+        );
```
DarshitChanpura, line 732: nit: test case on 720 and 721 can be combined to check that searchResponse only contains 1 entry and that is the song with id `FIRST_INDEX_ID_SONG_1`.

src/integrationTest/java/org/opensearch/security/FlsAndFieldMaskingTests.java
```diff
@@ -833,6 +906,789 @@ public void getFieldCapabilities() throws IOException {
...
+
+        try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(USER_NO_FIELD_TITLE_FLS)) {
+            assertProperGetResponsesForCorrespondingFLSRestrictions(restHighLevelClient, getRequest, false);
+        }
+    }
+
+    private void assertProperGetResponsesForCorrespondingFLSRestrictions(
+        RestHighLevelClient restHighLevelClient,
+        GetRequest getRequest,
+        boolean getOne
```
DarshitChanpura, line 924: nit: can be renamed to `shouldShowFieldTitle`

DarshitChanpura: also, method names can be shortened to: `assertGetForFLSRestrictions`, `assertMGetsForFLSRestrictions`, `assertSearchForFLSRestrictions`, etc.

DarshitChanpura (approved):

shikharj05 (approved):
Thank you @prabhask5 for adding these tests!

cwperks:
> Is there consensus that this is the best way to deal with overlapping roles? In these test scenarios I have overlapping roles that cause users to both only see the title field in a document and see every other field but title in a document. With the current functionality the combination of these roles leads to not a union, but just following the less restrictive role completely, leading to the output having all fields except the title field. But in the case that both of these roles are applied to the same user, wouldn't the additive property of the roles make more sense for all fields of the document to be visible here? I can work on refactoring the role filtering implementation if there is support for this idea. Also I believe that this additive implementation might fix issues like https://github.com/opensearch-project/security/issues/2556 since it would not depend on the order of overlapping roles.

Good question @prabhask5! I think the functionality is more clear in DLS, but FLS is trickier because of implementation details on the Lucene level.

For DLS it is a Union:

Imagine **UserA** is mapped to **role1** and **role2**

Also take an index called songs with these entries:

**doc1**, **doc2**, **doc3**, **doc4**, **doc5**

**role1** would allow user to see **doc1** and **doc2**

**role2** would allow user to see **doc4** and **doc5**

What should **UserA** be able to see? In the DLS implementation in the security plugin it would be **doc1**, **doc2**, **doc4** and **doc5** (all except for **doc3**) which is a union of the documents that can be seen from the user being mapped to **role1** and **role2**

This is all documented here: https://opensearch.org/docs/latest/security/access-control/document-level-security/#dls-and-multiple-roles

The limitation on FLS is mentioned here: https://opensearch.org/docs/latest/security/access-control/field-level-security/#interaction-with-multiple-roles

cwperks (approved):
Thank you for the contribution @prabhask5 !

src/integrationTest/java/org/opensearch/security/FlsAndFieldMaskingTests.java
```diff
@@ -214,6 +218,68 @@ public class FlsAndFieldMaskingTests {
...
+     */
+    static final TestSecurityConfig.User USER_ONLY_FIELD_TITLE_MASKED = new TestSecurityConfig.User("masked_user").roles(
+        ROLE_ONLY_FIELD_TITLE_MASKED
+    );
+
+    /**
+     * Example user with fls filter in which the user can only see the {@link Song#FIELD_TITLE} field and can see every field but the {@link Song#FIELD_TITLE} field- should default to showing no fields.
+     */
+    static final TestSecurityConfig.User USER_BOTH_ONLY_AND_NO_FIELD_TITLE_FLS = new TestSecurityConfig.User("inclusive_exclusive_fls_user")
+        .roles(ROLE_ONLY_FIELD_TITLE_FLS, ROLE_NO_FIELD_TITLE_FLS);
```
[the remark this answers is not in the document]
cwperks: What happens when a user defines a role that is "invalid"? Is there at least a log entry that can be looked for that mentions that the role is "invalid"? There should be some communication to the user if they try to add inclusive and exclusive rules simultaneously.

PR #4448, opened 2024-06-12
New algorithm for resolving action groups

nibix:
### Description

This code change is just in preparation for the change in #4380 as requested in https://github.com/opensearch-project/security/pull/4416#pullrequestreview-2106138499 .

This replaces the previous action group resolution algorithm done in ConfigModelV7 by a new one, which has a number of enhancements:

- The new implementation forms an independent class, enhancing testability and use in other components that will be introduced in #4380.
- The new implementation handles action group configurations containing loops gracefully. The old one would wind up StackOverflowErrors.
- The new one pre-computes the resolved action groups, thus enhancing the performance.

* Category: Enhancement
* Why these changes are required? - The changes in #4380 will no longer use ConfigModelV7 and thus need an independent action group resolution mechanism.
* What is the old behavior before changes and new behavior after changes? - No behavioral changes except for action group definitions containing groups, which will be handled gracefully now.

### Testing

- Unit tests in `org.opensearch.security.securityconf.FlattenedActionGroupsTest`

### Check List

### DIFF
--- a/src/test/java/org/opensearch/security/securityconf/FlattenedActionGroupsTest.java
+++ b/src/test/java/org/opensearch/security/securityconf/FlattenedActionGroupsTest.java
@@ -10,8 +10,10 @@
  */
 package org.opensearch.security.securityconf;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import com.google.common.collect.ImmutableMap;
@@ -26,10 +28,12 @@ import org.opensearch.security.securityconf.impl.v7.ActionGroupsV7;
 public class FlattenedActionGroupsTest {
     @Test
     public void basicTest() throws Exception {
-        TestActionGroup testActionGroups = new TestActionGroup().with("Z", "C", "A")
-            .with("A", "A1", "A2", "A3")
-            .with("B", "B1", "B2", "B3")
-            .with("C", "A", "B", "C1");
+        TestActionGroups testActionGroups = new TestActionGroups(
+            new TestActionGroup("Z").members("C", "A"), // C and A are defined below
+            new TestActionGroup("A").members("A1", "A2", "A3"), // A1-A3 are leafs
+            new TestActionGroup("B").members("B1", "B2", "B3"), // B1-B3 are leafs
+            new TestActionGroup("C").members("A", "B", "C1") // A and B are defined above, C1 is leaf
+        );
         SecurityDynamicConfiguration<ActionGroupsV7> config = SecurityDynamicConfiguration.fromMap(
             testActionGroups.map,
             CType.ACTIONGROUPS,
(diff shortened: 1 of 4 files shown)


shikharj05 (approved):
While this drops support for Configv6, I think that should be okay. LGTM 👍

willyborankin (approved):

src/test/java/org/opensearch/security/securityconf/FlattenedActionGroupsTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.Test;
+
+import org.opensearch.security.securityconf.impl.CType;
+import org.opensearch.security.securityconf.impl.SecurityDynamicConfiguration;
+import org.opensearch.security.securityconf.impl.v7.ActionGroupsV7;
+
+public class FlattenedActionGroupsTest {
+    @Test
+    public void basicTest() throws Exception {
+        TestActionGroup testActionGroups = new TestActionGroup().with("Z", "C", "A")
```
cwperks, line 29: nit: IMO this could be more readable if these are labeled as group or action. i.e. groupA instead of A or actionA1 instead of Ai.

I think it would also be nice to define a test action group like new TestActionGroup().name("groupZ").allowedActions("groupC", "groupA");

nibix: I have now made the `TestActionGroup` API more fluent. Additionally, I have added comments to the action group creation to indicate what elements get further expanded and what elements are leaves. 

However, I would like to keep the opaque naming to make clear that the resolver code does not put any semantics into the naming. In the end, it does not matter to the code what string is an action group and what is an action. It just resolves what it can resolve.

shikharj05 (approved):

cwperks (approved):
Thank you @nibix! This change looks good to me. One scenario I'd like to see covered would be a change in action group and verify its reflected in the new FlattenedActionGroups data structure. 

Would it make sense to included the flattened actions in the API response for `GET _plugins/_security/api/actiongroups/<action-group>`?

https://opensearch.org/docs/latest/security/access-control/api/#get-action-group

nibix:
@cwperks 

> Thank you @nibix! This change looks good to me. One scenario I'd like to see covered would be a change in action group and verify its reflected in the new FlattenedActionGroups data structure.

This is certainly an important thing to test. However, such configuration updates are handled by the `DynamicConfigFactory` class which is not changed by this PR.

There used to be an integration test which would test what you are asking for - that is, it changed action groups via API and tested the effectiveness of these changes via normal user operations. That was at https://github.com/opensearch-project/security/blob/069249656970e839302a0d3b8b87c82fbf3e56cc/src/test/java/org/opensearch/security/dlic/rest/api/ActionGroupsApiTest.java . There was recently a refactor of this test in #4371. I quickly skimmed the replacement code and could not find any logic which tests the effectiveness of the modified configuration, but maybe I missed something?

nibix:
@DarshitChanpura 

> Would the new runtime be more linear compared to the older implementation because we are using a combination of BFS + DFS?

The major gain comes from the fact that we calculate the resolved action group space only once instead of re-calculating it again and again for each action privilege configuration. Additionally, by using an iterative algorithm instead of a recursive algorithm it is easier for us to detect that resolution has finished and abort early. That give some more gain. However, all these gains only pertain to the config reload phase and not to the actual privilege evaluation.

Issue #3870, opened 2023-12-19
[RFC] Optimized Privilege Evaluation


nibix:
# Background

Performance tests indicate that the OpenSearch security layer adds a noticeable overhead to the indexing throughput of an OpenSearch cluster (cf https://github.com/opensearch-project/OpenSearch/issues/2461, https://github.com/opensearch-project/security/issues/3104). The overhead may vary depending on the number of indices, the use of aliases, the number of roles and the size of the user object.

As the latency is dependent on the number of indices, fresh clusters with only few indices perform well; however, users may experience with ongoing time and thus an increasing number of indices a continuous performance degradation. Especially such creeping changes are very undesirable, as they are hard to analyze and debug.

# Analysis

## High computational complexity of privilege evaluation

The privileges evaluation code uses a number of nested loops for determining if a user has privileges for an index (see for example https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java#L1103 and https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java#L488 ).

In human language, the privilege resolution algorithm is similar to this:

- Loop through all roles
  - Loop through all index patterns of role
    - Loop through all indices on cluster to find aliases and data streams
    - Loop through all aliases and data streams matching index pattern
      - Loop through all indices to find indices which are member of the alias
    - Loop through all indices to find indices which match the index pattern
    - Loop through indices requested by the operation
      - Loop through the indices produced by resolving the index patterns
        
Especially the last loop works on a cross product on two subsets of all existing indices. This is a O(n²) complexity class on the number of indices. On a cluster with many indices and in case very broad index patterns are used, this can require hundreds of thousands iterations.

A number of additional things can be observed here:

- The algorithm follows exactly the structure of the configuration files. This is not necessarily the most efficient approach.
- The algorithm newly resolves for each request the index patterns inside the role definition. This is not really necessary, as the result of these operations will remain constant until the configuration is changed or indices are created or deleted.

## Blow up of thread context data for DLS/FLS privilege evaluation

The privilege evaluation code for DLS/FLS works independently from the normal privilege evaluation code.

[...]

PR #4486, opened 2024-06-24
Remove special handling for do_not_fail_on_forbidden on cluster actions

### LINKED ISSUE #4485: do_not_fail_on_forbidden mode introduces inconsistencies for mget, msearch and similar actions
This is a facet of the work done for #3870.  A corresponding PR will immediately follow up, which is a part of the changes done in #4380

The implementation of the `do_not_fail_on_forbidden` mode in conjunction with multi-part requests like `mget`, `msearch`, `mtv` and other index-independent read operations like `scroll` seems to be inconsistent on several levels.

The `PrivilegesEvaluator` class gives certain cluster actions a special treatment for `do_not_fail_on_forbidden` mode at https://github.com/opensearch-project/security/blob/37afcaab74c8ef3133249f8ebe728ea897d0af2f/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L414 .

The affected actions are determined by `action0.startsWith("indices:data/read/")` and the conditions that determine whether an action is a cluster action: https://github.com/opensearch-project/security/blob/37afcaab74c8ef3133249f8ebe728ea897d0af2f/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L696

This results in the following actions being handled here:

- `indices:data/read/scroll`
- `indices:data/read/mget`
- `indices:data/read/msearch`
- `indices:data/read/mtv`
- `indices:data/read/search/template/render`

In any case, these permissions are checked as cluster permission with the method `securityRoles.impliesClusterPermissionPermission(action0)`.

The special treatment for `do_not_fail_on_forbidden` which is executed afterwards looks like this:

```
Set<String> reduced = securityRoles.reduce(
                            requestedResolved,
                            user,
                            new String[] { action0 },
                            resolver,
                            clusterService
                        );

                        if (reduced.isEmpty()) {
                            presponse.allowed = false;
                            return presponse;
                        }

                        if (irr.replace(request, true, reduced.toArray(new String[0]))) {
                            presponse.missingPrivileges.clear();
                            presponse.allowed = true;
                            return presponse;
                        }
```

This results in some inconsistencies described in the following.

## Privileges need to be specified both in cluster_privileges and index_privileges

As described before, the affected actions such as `indices:data/read/mget` are considered as cluster actions, so these need to be specified in the `cluster_privileges` section of a role. `securityRoles.reduce()` now however is an index-specific operation, so it checks the `index_privileges` of a role. Thus, in `do_not_fail_on_forbidden` mode, the privileges for  `indices:data/read/mget` need to be present in `index_privileges` as well. This is inconsistent to the mode where `do_not_fail_on_forbidden` is turned off, the index privileges for  `indices:data/read/mget`  are not needed in that case. 

It is also confusing and surprising to the user. The docs at https://opensearch.org/docs/latest/security/access-control/permissions/#cluster-permissions say on one hand:

> These permissions are for the cluster and can’t be applied granularly. For example, you either have permissions to take snapshots (cluster:admin/snapshot/create) or you don’t. The cluster permission, therefore, cannot grant a user privileges to take snapshots of a select set of indexes while preventing the user from taking snapshots of others.

But in the table below it mentions for `mget`  and related actions:

> Permission to run multiple GET operations in one request. This setting must be configured as both a cluster- and index-level permission

This is a contradiction to the first sentence. Additionally, it is not true when `do_not_fail_on_forbidden` is turned off.

## `_mget` gains an invalid index pattern support in `do_not_fail_on_forbidden`  mode

In OpenSearch, the `_get` API and indivual items in `_mget` API requests always refer to exactly one index, so index patterns are not supported here. If you try to use an index pattern, you will usually get an error:

```
{
    "docs": [
        {
            "_index": "test_*",
            "_id": "1",
            "error": {
                "root_cause": [
                    {
                        "type": "index_not_found_exception",
                        "reason": "no such index [test_*]",
                        "index": "test_*",
                        "resource.id": "test_*",
                        "resource.type": "index_expression",
                        "index_uuid": "_na_"
                    }
                ],
                "type": "index_not_found_exception",
                "reason": "no such index [test_*]",
                "index": "test_*",
                "resource.id": "test_*",
                "resource.type": "index_expression",
                "index_uuid": "_na_"
            }
        }
   ]
}
```

When having `do_not_fail_on_forbidden` enabled, however, index patterns are supported at least in some cases like the test at https://github.com/opensearch-project/security/blob/37afcaab74c8ef3133249f8ebe728ea897d0af2f/src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java#L281 proves.

This is caused by the `irr.replace()` call in the special handling which resolves index patterns for the `MultiGetRequest.Item` objects.

## The `indices:data/read/scroll` privilege is also required as index privilege, but any index will do

**Note:** This might sound like a security issue first, but it is actually not, as the `indices:data/read/search` privileges are still effective to limit access.

Like `indices:data/read/mget`,  the `indices:data/scroll` privileges need to be provided both as `cluster_permissions` and `index_permissions`. The special handling for  `indices:data/scroll`  also includes the attempt to gather the requested indices from the request. As the scroll request does not carry index information, the requested indices will be however assumed to be `*` (all indices). The `securityRoles.reduce()` call will then calculate the intersection between `*` and the indices with privileges and return all indices with available privileges. With these, the `irr.replace()` method is called. This however is just a no-op for scroll requests: https://github.com/opensearch-project/security/blob/37afcaab74c8ef3133249f8ebe728ea897d0af2f/src/main/java/org/opensearch/security/resolver/IndexResolverReplacer.java#L797

That means effectively, that a  `indices:data/read/scroll` index privilege is necessary, but any non-empty set of indices is sufficient to satisfy the special handling code. While this is not a security issue, as there are further safe-guards (see above), this is an inconsistent behaviour.

## Proposed solution

The special handling code for `do_not_fail_on_forbidden` mode on cluster actions is actually unnecessary and can be completely removed. 

For the multi actions like `mget`, `msearch`, `mtv`, and search template actions, the action handling will always trigger sub-actions which will be routed again through the privilege evaluation code - then, the proper access controls will be enforced.

For the scroll action, proper access control has been already enforced before when issuing the `search` request.

Thus, the special handling code actually introduces no benefits and only issues.

nibix:
The special handling for `do_not_fail_on_forbidden` on cluster actions does not provide any benefits - rather it just causes some inconsistent behavior. See #4485 for the details.

This is part of the work done for #3870

### Description

* Category: Enhancement and bug fix
* Why these changes are required? - The present behavior is inconsistent and confusing. Additionally, it does unnecessary computations on index sets, which can potentially decrease action throughput
* What is the old behavior before changes and new behavior after changes? - Both clusters with `do_not_fail_on_forbidden` set to `true`  and to `false` will need the same set of privileges for multi actions and scroll actions. The need to specify `indices:data/read/mget`, `indices:data/read/msearch`, `indices:data/read/mtv`, `indices:data/read/scroll` and `indices:data/read/search/template/render` as index permissions in the roles config will go away.

### Issues Resolved

- Resolves #4485 

### Testing

- Tests in `DoNotFailOnForbiddenTests` have been adapted.

### Check List
- [ ] New Roles/Permissions have a corresponding security dashboards plugin PR - not necessary

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
+++ b/src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java
@@ -278,7 +278,7 @@ public class DoNotFailOnForbiddenTests {
     @Test
     public void shouldMGetDocument_positive() throws IOException {
         try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(LIMITED_USER)) {
-            MultiGetRequest request = new MultiGetRequest().add(BOTH_INDEX_PATTERN, ID_1).add(BOTH_INDEX_PATTERN, ID_4);
+            MultiGetRequest request = new MultiGetRequest().add(MARVELOUS_SONGS, ID_1).add(MARVELOUS_SONGS, ID_4);
 
             MultiGetResponse response = restHighLevelClient.mget(request, DEFAULT);
 
@@ -297,11 +297,34 @@ public class DoNotFailOnForbiddenTests {
     }
 
     @Test
+    public void shouldMGetDocument_partial() throws Exception {
+        try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(LIMITED_USER)) {
+            MultiGetRequest request = new MultiGetRequest().add(MARVELOUS_SONGS, ID_1).add(HORRIBLE_SONGS, ID_4);
+
+            MultiGetResponse response = restHighLevelClient.mget(request, DEFAULT);
+
+            MultiGetItemResponse[] responses = response.getResponses();
+            assertThat(responses, arrayWithSize(2));
+            MultiGetItemResponse firstResult = responses[0];
+            MultiGetItemResponse secondResult = responses[1];
+            assertThat(firstResult.getFailure(), nullValue());
+            assertThat(
+                firstResult.getResponse(),
+                allOf(containDocument(MARVELOUS_SONGS, ID_1), documentContainField(FIELD_TITLE, TITLE_MAGNUM_OPUS))
+            );
+            assertThat(secondResult.getFailure().getMessage(), containsString("no permissions for [indices:data/read/mget[shard]]"));
+        }
+    }
+
+    @Test
     public void shouldMGetDocument_negative() throws IOException {
         try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(LIMITED_USER)) {
             MultiGetRequest request = new MultiGetRequest().add(HORRIBLE_SONGS, ID_4);
-
-            assertThatThrownBy(() -> restHighLevelClient.mget(request, DEFAULT), statusException(FORBIDDEN));
+            MultiGetResponse response = restHighLevelClient.mget(request, DEFAULT);
+            MultiGetItemResponse[] responses = response.getResponses();
+            assertThat(responses, arrayWithSize(1));
+            MultiGetItemResponse firstResult = responses[0];
+            assertThat(firstResult.getFailure().getMessage(), containsString("no permissions for [indices:data/read/mget[shard]]"));
         }
     }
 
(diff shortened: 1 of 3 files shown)


nibix:
To avoid any confusion: Yes, this is already complete. The only code-change not related to tests is the removal of the `do_not_fail_on_forbidden` special handling for cluster actions.

DarshitChanpura (approved):
LGTM!

RyanL1997 (approved):
LGTM, re-running CI, cuz that seems like fail due to flaky

willyborankin (approved):

PR #4490, opened 2024-06-26
Separated DLS/FLS privilege evaluation from action privilege evaluation

nibix:
### Description

This change is in preparation for #3870 and #4380 .

This cuts off some parts from the quite big and monolithic method `PrivilegesEvaluator.evaluate()` into separate methods and modules.

This achieves several things:

- Better separation of concerns, thus better modularization and thus better understandability, re-usability and testability
- Computations of DLS/FLS privileges are done directly in the DLS/FLS valve code. Thus, if DLS/FLS valve is disabled, no DLS/FLS privileges are computed
- The new class `PrivilegesEvaluationContext` combines commonly needed information for privilege evaluation and thus allows to shorten the parameter lists of many methods in this context. For this PR, only the `PrivilegesEvaluator.evaluate()` method itself is changed, but further adaptions will follow up when due.

* Category - Refactoring
   - For #3870, we need to re-build the privilege evaluation code both for the normal action privilege evaluation and for DLS/FLS. Even though these are quite separate and different things, these are currently intermingled in one big monolith. The changes for #3870 will be easier to understand if they are structured better.
* What is the old behavior before changes and new behavior after changes? - No changes.

### Issues Resolved

Contributes to #3870

### Testing

- Existing integration tests

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/filter/SecurityFilter.java
+++ b/src/main/java/org/opensearch/security/filter/SecurityFilter.java
@@ -82,6 +82,7 @@ import org.opensearch.security.configuration.AdminDNs;
 import org.opensearch.security.configuration.CompatConfig;
 import org.opensearch.security.configuration.DlsFlsRequestValve;
 import org.opensearch.security.http.XFFResolver;
+import org.opensearch.security.privileges.PrivilegesEvaluationContext;
 import org.opensearch.security.privileges.PrivilegesEvaluator;
 import org.opensearch.security.privileges.PrivilegesEvaluatorResponse;
 import org.opensearch.security.resolver.IndexResolverReplacer;
@@ -378,7 +379,15 @@ public class SecurityFilter implements ActionFilter {
                 log.trace("Evaluate permissions for user: {}", user.getName());
             }
 
-            final PrivilegesEvaluatorResponse pres = eval.evaluate(user, action, request, task, injectedRoles);
+            PrivilegesEvaluationContext context = null;
+            PrivilegesEvaluatorResponse pres;
+
+            try {
+                context = eval.createContext(user, action, request, task, injectedRoles);
+                pres = eval.evaluate(context);
+            } catch (PrivilegesEvaluatorResponse.NotAllowedException e) {
+                pres = e.getResponse();
+            }
 
             if (log.isDebugEnabled()) {
                 log.debug(pres.toString());
@@ -387,7 +396,7 @@ public class SecurityFilter implements ActionFilter {
             if (pres.isAllowed()) {
                 auditLog.logGrantedPrivileges(action, request, task);
                 auditLog.logIndexEvent(action, request, task);
-                if (!dlsFlsValve.invoke(action, request, listener, pres.getEvaluatedDlsFlsConfig(), pres.getResolved())) {
+                if (!dlsFlsValve.invoke(action, request, listener, context)) {
                     return;
                 }
                 final CreateIndexRequestBuilder createIndexRequestBuilder = pres.getCreateIndexRequestBuilder();
--- a/src/main/java/org/opensearch/security/filter/SecurityFilter.java
+++ b/src/main/java/org/opensearch/security/filter/SecurityFilter.java
@@ -379,15 +379,8 @@ public class SecurityFilter implements ActionFilter {
                 log.trace("Evaluate permissions for user: {}", user.getName());
             }
 
-            PrivilegesEvaluationContext context = null;
-            PrivilegesEvaluatorResponse pres;
-
-            try {
-                context = eval.createContext(user, action, request, task, injectedRoles);
-                pres = eval.evaluate(context);
-            } catch (PrivilegesEvaluatorResponse.NotAllowedException e) {
-                pres = e.getResponse();
-            }
+            PrivilegesEvaluationContext context = eval.createContext(user, action, request, task, injectedRoles);
+            PrivilegesEvaluatorResponse pres = eval.evaluate(context);
 
             if (log.isDebugEnabled()) {
                 log.debug(pres.toString());
--- a/src/main/java/org/opensearch/security/filter/SecurityFilter.java
+++ b/src/main/java/org/opensearch/security/filter/SecurityFilter.java
@@ -389,7 +389,7 @@ public class SecurityFilter implements ActionFilter {
             if (pres.isAllowed()) {
                 auditLog.logGrantedPrivileges(action, request, task);
                 auditLog.logIndexEvent(action, request, task);
-                if (!dlsFlsValve.invoke(action, request, listener, context)) {
+                if (!dlsFlsValve.invoke(context, listener)) {
                     return;
                 }
                 final CreateIndexRequestBuilder createIndexRequestBuilder = pres.getCreateIndexRequestBuilder();
--- a/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
+++ b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
@@ -1,0 +1,92 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+package org.opensearch.security.privileges;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Supplier;
+
+import com.google.common.collect.ImmutableSet;
+
+import org.opensearch.action.ActionRequest;
+import org.opensearch.cluster.ClusterState;
+import org.opensearch.cluster.metadata.IndexNameExpressionResolver;
+import org.opensearch.security.resolver.IndexResolverReplacer;
+import org.opensearch.security.support.WildcardMatcher;
+import org.opensearch.security.user.User;
+import org.opensearch.tasks.Task;
+
+/**
+ * Request-scoped context information for privilege evaluation.
+ *
+ * This class carries metadata about the request and provides caching facilities for data which might need to be
+ * evaluated several times per request.
+ *
+ * As this class is request-scoped, it is only used by a single thread. Thus, no thread synchronization mechanisms
+ * are necessary.
+ */
+public class PrivilegesEvaluationContext {
+    private final User user;
+    private final String action;
+    private final ActionRequest request;
+    private IndexResolverReplacer.Resolved resolvedRequest;
+    private final Task task;
+    private final ImmutableSet<String> mappedRoles;
+    private final IndexResolverReplacer indexResolverReplacer;
+
+    public PrivilegesEvaluationContext(
+        User user,
+        ImmutableSet<String> mappedRoles,
+        String action,
+        ActionRequest request,
+        Task task,
+        IndexResolverReplacer indexResolverReplacer
+    ) {
+        this.user = user;
+        this.mappedRoles = mappedRoles;
+        this.action = action;
+        this.request = request;
+        this.task = task;
+        this.indexResolverReplacer = indexResolverReplacer;
+    }
+
+    public User getUser() {
+        return user;
+    }
+
+    public String getAction() {
+        return action;
+    }
+
+    public ActionRequest getRequest() {
+        return request;
+    }
+
+    public IndexResolverReplacer.Resolved getResolvedRequest() {
+        IndexResolverReplacer.Resolved result = this.resolvedRequest;
+
+        if (result == null) {
+            result = indexResolverReplacer.resolveRequest(request);
+            this.resolvedRequest = result;
+        }
+
+        return result;
+    }
+
+    public Task getTask() {
+        return task;
+    }
+
+    public ImmutableSet<String> getMappedRoles() {
+        return mappedRoles;
+    }
+
+}
--- a/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
+++ b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
@@ -10,17 +10,10 @@
  */
 package org.opensearch.security.privileges;
 
-import java.util.HashMap;
-import java.util.Map;
-import java.util.function.Supplier;
-
 import com.google.common.collect.ImmutableSet;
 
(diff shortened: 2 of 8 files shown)


src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -226,18 +221,45 @@ private void setUserInfoInThreadContext(User user) {
...
+                log.info("Roles {} are not mapped to the user {}", injectedRolesValidationSet, user);
+                throw new PrivilegesEvaluatorResponse.NotAllowedException(presponse);
+            }
+            mappedRoles = ImmutableSet.copyOf(injectedRolesValidationSet);
+        }
+
+        return new PrivilegesEvaluationContext(user, mappedRoles, action0, request, task, irr);
+    }
+
+    public PrivilegesEvaluatorResponse evaluate(PrivilegesEvaluationContext context) {
```
cwperks, line 251: This signature is much cleaner now. Nice!

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -378,7 +379,15 @@ private <Request extends ActionRequest, Response extends ActionResponse> void ap
...
             }
 
-            final PrivilegesEvaluatorResponse pres = eval.evaluate(user, action, request, task, injectedRoles);
+            PrivilegesEvaluationContext context = null;
+            PrivilegesEvaluatorResponse pres;
+
+            try {
+                context = eval.createContext(user, action, request, task, injectedRoles);
+                pres = eval.evaluate(context);
+            } catch (PrivilegesEvaluatorResponse.NotAllowedException e) {
```
cwperks, line 388: I think this is going to get confusing over time if the only place NotAllowedException is thrown is within `createContext`, but not within `eval.evaluate`.

cwperks: wdyt about moving the cross-cluster specific validation to here instead of within `createContext`?

nibix: I agree - having an exception here was certainly not my first choice. However, due to the validation performed the injected roles done at https://github.com/opensearch-project/security/blob/2b5a811de599f7c7fc2ca2b9246e57fd6cfaf33b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L262 I needed a dual return type. 

The alternative would be not using an exception but a kind of `AlternativeResult<PrivilegesEvaluationContext, PrivilegesEvaluatorResponse>` type, but that would be just a different inconsistency. WDYT?

cwperks: I would be ok with `Pair<PrivilegesEvaluationContext, Optional<PrivilegesEvaluatorResponse>>`.

nibix: Using a `Pair` type would lose the semantics that the results are alternatives, though.

cwperks: > Using a Pair type would lose the semantics that the results are alternatives, though.

`AlernativeResults` sounds good.

nibix: > wdyt about moving the cross-cluster specific validation to here instead of within createContext?

You mean moving it to `eval.evaluate()`?

The problem is that validation and determination of roles is entangled.

 I'd like to keep `PrivilegesEvaluationContext` immutable - with caches as an exception. That means that the `mappedRoles` should be determined as soon as it is created.

But, maybe, I should give up the idea that the mapped roles should not change in the `PrivilegesEvaluationContext`  object. :thinking:

cwperks: > You mean moving it to eval.evaluate()? 

Potentially a method called `computeMappedRoles` that's called before createContext that can optionally return a PrivilegesEvaluatorResponse?

DarshitChanpura: > I would be ok with Pair<PrivilegesEvaluationContext, Optional<PrivilegesEvaluatorResponse>>. Curious to know what other maintainers think too.

Will we have different contexts at any given point? What are the pros and cons of having a catch block vs having a pair of AlternativeResult?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public ActionRequest getRequest() {
+        return request;
+    }
+
+    public IndexResolverReplacer.Resolved getResolvedRequest() {
+        IndexResolverReplacer.Resolved result = this.resolvedRequest;
+
+        if (result == null) {
```
DarshitChanpura, line 69: for my understanding, in what cases will the result be null

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -250,23 +272,8 @@ public PrivilegesEvaluatorResponse evaluate(
             action0 = PutMappingAction.NAME;
         }
 
-        final PrivilegesEvaluatorResponse presponse = new PrivilegesEvaluatorResponse();
+        PrivilegesEvaluatorResponse presponse = new PrivilegesEvaluatorResponse();
```
DarshitChanpura, line 275: will this be mutated at some point?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -226,18 +221,45 @@ private void setUserInfoInThreadContext(User user) {
...
+                presponse.allowed = false;
+                presponse.missingSecurityRoles.addAll(injectedRolesValidationSet);
+                presponse.reason("Injected roles validation failed");
+                log.info("Roles {} are not mapped to the user {}", injectedRolesValidationSet, user);
+                throw new PrivilegesEvaluatorResponse.NotAllowedException(presponse);
+            }
+            mappedRoles = ImmutableSet.copyOf(injectedRolesValidationSet);
+        }
+
+        return new PrivilegesEvaluationContext(user, mappedRoles, action0, request, task, irr);
```
DarshitChanpura, line 248: why pass irr instead of irr.resolveRequest (which is what we utilize at the end)?

src/main/java/org/opensearch/security/filter/SecurityFilter.java
nibix: > What are the pros and cons of having a catch block vs having a pair of AlternativeResult?

IMHO, it's mostly just a matter of taste.

Pro exception and catch:

- This is the "official" way in Java to signal exceptional returns from a method

Contra exception and catch:

- It might be confusing that not all methods throw exceptions but return `PrivilegeEvaluatorResponse` instead
- Exceptions have a slight performance penalty - even though it should be not an issue here, as we are looking at a rarely happening case

Pro AlternativeResult:

- Less confusing? (Well, I am not sure about this one)
- Bit better performance in the error case

Contra AlternativeResult:

- Requires more complicated logic to model the alternative cases.
- Correct usage cannot be verified statically during compile time. Instead, it needs to throw exceptions in case certain methods are used in the wrong state.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
nibix: This allows lazy execution. `irr.resolveRequest()` can be expensive, thus lazy execution helps us to skip it in cases where it is not needed.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
nibix: This is the initial state for the lazy execution. When it is null, the request was not resolved yet.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
nibix: Not at the moment, but it will be in #4380. But for now, I can revert that change.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
cwperks: @nibix wdyt about renaming the exception? I think it would be confusing to add an exception called NotAllowedException and many people may assume by looking at the code that a NotAllowedException is thrown when authz fails and that's not true.

DarshitChanpura: If we introduce the catch block then it will be only for this particular case, but rest of the places would still return PrivilegeEvaluatorResponse. It might be confusing to someone who reads it in future to understand why is there a catch block only at one place

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
DarshitChanpura: Is there a way to avoid lazy execution? (unless it is better to keep but I'm not sure how)

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
DarshitChanpura: I'm still confused. Are we trying to avoid another call to irr.resolveRequest if the result if already non-null?

nibix: Exactly. The call to `irr.resolveRequest()` can be amazingly expensive on clusters with many indices, so you want to avoid calling it without a good reason. That's why we have lazy execution to make sure that it is only executed when really needed. Additionally, if it is executed, we cache the reference in the `PrivilegeEvaluationContext` object to avoid having to call  `irr.resolveRequest()`  later again the resolved indices should be needed again.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
nibix: After having slept a night over it, my opinion is: Even though I'd like the `PrivilegeEvaluationContext` object as immutable as possible, I'd make the `mappedRoles` modifiable - that way we can move the validation code again to the actual `evaluate()` method and get a clean method interface instead. WDYT?

cwperks: I think it makes sense to be immutable. Did you see the comment [here](https://github.com/opensearch-project/security/pull/4490#discussion_r1661365425)? What about a new method that's called before `createContext` where the responsibility of the method is to compute the mapped roles?

scrawfor99: I prefer to not have the one off exception type here. The rest, is viable in either way but I would rather have limited exception types.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -226,18 +221,45 @@ private void setUserInfoInThreadContext(User user) {
...
-    public PrivilegesEvaluatorResponse evaluate(
-        final User user,
-        String action0,
-        final ActionRequest request,
-        Task task,
-        final Set<String> injectedRoles
-    ) {
+    public PrivilegesEvaluationContext createContext(User user, String action0, ActionRequest request, Task task, Set<String> injectedRoles)
+        throws PrivilegesEvaluatorResponse.NotAllowedException {
+        if (!isInitialized()) {
```
scrawfor99, line 226: How are we hitting a scenario where the plugin is not initialized but we are calling createContext for the evaluator? It seems this should not arise right? If we can reach this state doesn't that mean something is wrong.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
nibix: @cwperks 

> I think it makes sense to be immutable. Did you see the comment https://github.com/opensearch-project/security/pull/4490#discussion_r1661365425? What about a new method that's called before createContext where the responsibility of the method is to compute the mapped roles?

But if I understand it correctly, that method also would need to validate the roles with the             ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROLES_VALIDATION value. Thus that method also would need a dual return type: either the mapped roles or a privilege evaluation error, either in form of an exception or another form of alternative return type. That would not really win anything.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
nibix: "Not initialized" means here "the configuration roles.yml, rolesmapping.yml, and actiongroups.yml" has not yet been installed on the cluster. That's a quite common case for a freshly installed cluster.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
nibix: So, I now went ahead and moved the logic related to `OPENDISTRO_SECURITY_INJECTED_ROLES_VALIDATION` back to the evaluate() method. This allowed me to delete the exception stuff again, but at the cost of having the `mappedRoles` attribute non-final in `PrivilegesEvaluationContext`. 

Side note: All this is only necessary because of the call to `mappedRoles = ImmutableSet.copyOf(injectedRolesValidationSet)` in line 268 of the following code block:

https://github.com/opensearch-project/security/blob/20c87750524481242199ec8d63b41db56cd6e64f/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L255-L269

I have ready through the design docs which introduced the functionality and could not find a good reason why the roles need to be modified by a validation string. So, if we would determine that this is actually not necessary, we could make the attribute final again.

cwperks: Should we capture the role re-assignment on the validation set in an issue to discuss further? It looks safe to remove for me since it enforces that the roles must be a superset of the validation set.

I would try removing the line and see if there's an existing test that fails.

cwperks (approved):
The new changes look good. Let's capture the role re-assignment for the injected roles validation set in an issue to discuss further.

shikharj05 (approved):

DarshitChanpura (approved):

DarshitChanpura:
@cwperks @scrawfor99 this is blocked by each of your un-resolved conversations. Would you mind going through them and closing as needed?

scrawfor99 (approved):

PR #4524, opened 2024-07-04
Add support for PBKDF2 for password hashing & add support for configuring BCrypt and PBKDF2

dancristiancecoi:
### Description
This change adds support for PBKDF2 as an alternative, FIPS compliant, password hashing algorithm. 
Furthermore, it adds support for configuring the following parameters for these hashing algorithms:

**BCrypt:**

- Rounds (defaults to 12)
- Minor (defaults to Y)

**PBKDF2**:

- Iterations (defaults to 600000)
- Length (defaults to 256)
- Function (defaults to SHA256)

Additionally, the _hash.sh script_ has been modified to support the new algorithm and configuration.

```
usage: hash.sh [-h] [-p <password>] [-env <Environment variable name>] [-a
       <hashing algorithm>] [-r <rounds (BCrypt)>] [-min <minor (BCrypt)>]
       [-l <length (PBKDF2)>] [-f <function (PBDKF2)>] [-i <iterations
       (PBKDF2)>]
 -h,--help                               Display the help information
 -p,--password <password>                Cleartext password to hash
 -env <Environment variable name>        Environment variable name to read
                                         password from
 -a,--algorithm <hashing algorithm>      Algorithm to use for password
                                         hashing. Valid values are: BCrypt
                                         | PBKDF2. Default: BCrypt
 -r,--rounds <rounds (BCrypt)>           Number of rounds to use in
                                         logarithmic form. Valid values
                                         are: 4 to 31. Default: 12
 -min,--minor <minor (BCrypt)>           Version of BCrypt algorithm to
                                         use. Valid values are: A | B | Y.
                                         Default: Y
 -l,--length <length (PBKDF2)>           Desired length of the final
                                         derived key. Default: 256
 -f,--function <function (PBDKF2)>       Pseudo-random function applied to
                                         the password. Valid values are
                                         SHA1 | SHA224 | SHA256 | SHA384 |
                                         SHA512. Default: SHA256
 -i,--iterations <iterations (PBKDF2)>   Number of times the pseudo-random
                                         function is applied to the
                                         password. Default: 600000
```

_install_demo_configuration.sh_ was left unchanged and currently only creates demo configurations with BCrypt hashes. 

### Issues Resolved
https://github.com/opensearch-project/security/issues/4079 (partially)

### Related issues:
https://github.com/opensearch-project/security/pull/4381

### Testing
Added unit tests, integration tests and done manual tests. 

#### Deploy without any hashing configuration specified:

```
curl -k https://localhost:9200 -u admin:$PASSWORD
{
  "name" : "SAS-1B16WV3",
  "cluster_name" : "opensearch",
  "cluster_uuid" : "zeB4OjWjRACHFQikYd34Fg",
  "version" : {
    "distribution" : "opensearch",
    "number" : "3.0.0-SNAPSHOT",
    "build_type" : "tar",
    "build_hash" : "58d1164f74e921a26b7a73b6185b38cc87bbc7a9",
    "build_date" : "2024-07-04T14:13:09.807311713Z",
    "build_snapshot" : true,
    "lucene_version" : "9.12.0",
    "minimum_wire_compatibility_version" : "2.16.0",
    "minimum_index_compatibility_version" : "2.0.0"
  },
  "tagline" : "The OpenSearch Project: https://opensearch.org/"
}


curl -k https://localhost:9200 -u admin:$WRONG_PASSWORD
Unauthorized

curl -X PUT -k -s -H "Content-Type: application/json" "https://localhost:9200/_opendistro/_security/api/internalusers/dancecoi" -u admin:$PASSWORD -d $'{"password": "strongpassword1996!","opendistro_security_roles": ["all_access"],"backend_roles": ["all_access"],"attributes": {"attribute1": "value1","attribute2": "value2"}}' "value2"}}'
{"status":"CREATED","message":"'dancecoi' created."}

curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoi:strongpassword1996!9200/_cat/indices?v" -u dancecoi:strongpassword1996!
health status index                        uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   security-auditlog-2024.07.04 NeRiqXzSSIKHxWnhLBOmaw   1   1          3            0     44.4kb         44.4kb
green  open   .opendistro_security         pRLb5T_8SyObPEUWlu7wug   1   0         10            0     89.6kb         89.6kb

curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoi:wrong_password
Unauthorized


curl -X PUT -k -s -H "Content-Type: application/json" "https://localhost:9200/_opendistro/_security/api/internalusers/dancecoihash" -u admin:$PASSWORD -d $'{"hash": "$2y$12$gdh2ecVBQmwpmcAeyReicuNtXyR6GMWSfXHxtcBBqFeFz2VQ8kDZe","opendistro_security_roles": ["all_access"],"backend_roles": ["all_access"],"attributes": {"attribute1": "value1","attribute2": "value2"}}' "value2"}}'
{"status":"CREATED","message":"'dancecoihash' created."}

curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoihash:testPassword
health status index                        uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   security-auditlog-2024.07.04 NeRiqXzSSIKHxWnhLBOmaw   1   1          5            0     74.4kb         74.4kb
green  open   .opendistro_security         pRLb5T_8SyObPEUWlu7wug   1   0         10            1    115.6kb        115.6kb
```

##### opensearch.yml config:
```
.....
plugins.security.password.hashing.algorithm: pbkdf2
plugins.security.password.hashing.pbkdf2.iterations: 650000
plugins.security.password.hashing.pbkdf2.length: 512
plugins.security.password.hashing.pbkdf2.function: SHA512
.....
```

##### internal_users.yml config:
```
admin:
  hash: "$5$2576980377600512$CobxDePVXnGapo+pU3xo7hCZElhfgsnxtRh+Tox4nsgEP/3/McAWatRQz74eczIPYtbMCZbDFc4Bpd2GUZ56aw==$9aG6wcMv2ztLddg4UJN21dPtVzK2AihllX5p5w1OSY3C8je2ryFDsgXhb8Mg//YcClTTTK5Q6ZCtc1M0vQeLCA=="
  reserved: true
  backend_roles:
  - "admin"
  description: "Demo admin user"
....
```

##### Testing:

```
curl -k https://localhost:9200 -u admin:$PASSWORD
{
  "name" : "SAS-1B16WV3",
  "cluster_name" : "opensearch",
  "cluster_uuid" : "cnmnETP_R--rNPCV2csXcQ",
  "version" : {
    "distribution" : "opensearch",
    "number" : "3.0.0-SNAPSHOT",
    "build_type" : "tar",
    "build_hash" : "58d1164f74e921a26b7a73b6185b38cc87bbc7a9",
    "build_date" : "2024-07-04T14:13:09.807311713Z",
    "build_snapshot" : true,
    "lucene_version" : "9.12.0",
    "minimum_wire_compatibility_version" : "2.16.0",
    "minimum_index_compatibility_version" : "2.0.0"
  },
  "tagline" : "The OpenSearch Project: https://opensearch.org/"
}

curl -k https://localhost:9200 -u admin:$WRONG_PASSWORD
Unauthorized

./tools/hash.sh -a pbkdf2 -f SHA512 -l 512 -p testPassword
WARNING: nor OPENSEARCH_JAVA_HOME nor JAVA_HOME is set, will use /usr/bin/java
$5$2576980377600512$Z5Sk45FrRk1UR2HZWMrbBiWmd63qN37Vc2pwJQIEjs0103aG8OFCYgkZJG+FQ533XAZCw5me9ilxkjPwVtm1eQ==$3jQnsKteu/2Jq3aR/C0ME6CI58T4N/dHtGfimYWBdVFnRVSFZ91GkDC7KJgesG989zW/UpvHqLCtT+o1zkKROw==

curl -X PUT -k -s -H "Content-Type: application/json" "https://localhost:9200/_opendistro/_security/api/internalusers/dancecoihash" -u admin:$PASSWORD -d $'{"hash": "$5$2576980377600512$Z5Sk45FrRk1UR2HZWMrbBiWmd63qN37Vc2pwJQIEjs0103aG8OFCYgkZJG+FQ533XAZCw5me9ilxkjPwVtm1eQ==$3jQnsKteu/2Jq3aR/C0ME6CI58T4N/dHtGfimYWBdVFnRVSFZ91GkDC7KJgesG989zW/UpvHqLCtT+o1zkKROw==","opendistro_security_roles": ["all_access"],"backend_roles": ["all_access"],"attributes": {"attribute1": "value1","attribute2": "value2"}}'
{"status":"CREATED","message":"'dancecoihash' created."}

curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoihash:testPassword
health status index                        uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   security-auditlog-2024.07.04 pdViqtaxSx6fM4LYlxIypQ   1   1          1            0     13.9kb         13.9kb
green  open   .opendistro_security         o7CmWxd6S6GSOQw9W2enEA   1   0         10            0     90.6kb         90.6kb


curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoihash:wrongTestPassword
Unauthorized

curl -X PUT -k -s -H "Content-Type: application/json" "https://localhost:9200/_opendistro/_security/api/internalusers/dancecoi" -u admin:$PASSWORD -d $'{"password": "strongpassword1996!","opendistro_security_roles": ["all_access"],"backend_roles": ["all_access"],"attributes": {"attribute1": "value1","attribute2": "value2"}}'
{"status":"CREATED","message":"'dancecoi' created."}

curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoi:strongpassword1996!
health status index                        uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   security-auditlog-2024.07.04 pdViqtaxSx6fM4LYlxIypQ   1   1          5            0     75.3kb         75.3kb
green  open   .opendistro_security         o7CmWxd6S6GSOQw9W2enEA   1   0         10            0     93.9kb         93.9kb

curl -k -s "https://localhost:9200/_cat/indices?v" -u dancecoi:wrongpassword
Unauthorized

```

### Documentation
This requires changes to the OpenSearch documentation. 

- https://github.com/opensearch-project/documentation-website/issues/7669

### Check List


dancristiancecoi:
Tests are failing due to the switch to JDK21 and SecurityManager no longer being supported. 

I can try and add the _@SuppressWarnings("removal")_ annotation but any advice on how to proceed will be appreciated!

cwperks:
@dancristiancecoi some quick comments from the PR description. I think it would be helpful to expand on the help text for the hashing script to make it read more like a `man` page. 

Without passing any params will it default to the same BCrypt hash as it does today?

For the default values I think we should include those in the help text of the hash script as well. For any option with multiple choices (like `--minor`) I think the help text should include the list of options (if its a small amount of options) or a reference to where you can find out more about the possible options.

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -29,16 +29,15 @@
...
+import org.opensearch.security.hasher.PasswordHasherFactory;
 import org.opensearch.security.support.ConfigConstants;
 
 import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
 
 import static org.opensearch.security.DefaultObjectMapper.YAML_MAPPER;
-import static org.opensearch.security.support.ConfigConstants.SECURITY_RESTAPI_PASSWORD_MIN_LENGTH;
-import static org.opensearch.security.support.ConfigConstants.SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX;
+import static org.opensearch.security.support.ConfigConstants.*;
```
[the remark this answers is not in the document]
willyborankin: Hmmm CheckStyle should catch such things.

src/test/java/org/opensearch/security/auditlog/AbstractAuditlogiUnitTest.java
```diff
@@ -38,7 +38,7 @@ protected String getResourceFolder() {
         return "auditlog";
     }
 
-    protected final void setup(Settings settings) throws Exception {
```
cwperks, line 41: Just realized that there is a typo in the name of this class. It should be renamed to `AbstractAuditlogUnitTest`

src/test/java/org/opensearch/security/auditlog/AbstractAuditlogiUnitTest.java
```diff
@@ -56,7 +56,7 @@ protected final void setup(Settings settings) throws Exception {
         });
 
         final Settings nodeSettings = defaultNodeSettings(defaultNodeSettings.build());
-        setup(Settings.EMPTY, new DynamicSecurityConfig(), nodeSettings, init);
+        setup(Settings.EMPTY, securityConfig, nodeSettings, init);
```
cwperks, line 59: @dancristiancecoi is the CI failure related to this changes in this file?

ref: https://github.com/opensearch-project/security/actions/runs/9793063749/job/27040225350?pr=4524

```
  2> java.lang.NullPointerException: Cannot invoke "org.opensearch.security.test.helper.rest.RestHelper.executeDeleteRequest(String, org.apache.hc.core5.http.Header[])" because "this.rh" is null
        at __randomizedtesting.SeedInfo.seed([C16759D4AFC08567:D86111E1CD25905D]:0)
        at org.opensearch.security.auditlog.integration.BasicAuditlogTest.testIndexCloseDelete(BasicAuditlogTest.java:775)
```

dancristiancecoi:
> @dancristiancecoi some quick comments from the PR description. I think it would be helpful to expand on the help text for the hashing script to make it read more like a `man` page.
> 
> Without passing any params will it default to the same BCrypt hash as it does today?
> 
> For the default values I think we should include those in the help text of the hash script as well. For any option with multiple choices (like `--minor`) I think the help text should include the list of options (if its a small amount of options) or a reference to where you can find out more about the possible options.

Good points! I will document the default values in the Hasher script.

The script is backwards compatible, so if a user doesnt pass any algorithm related parameters it will default to the same BCrypt hash configuration

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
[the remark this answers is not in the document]
dancristiancecoi: Yeah, very strange that it didnt catch this specific import

cwperks:
Took a first pass and left a few comments. I'm curious to see the test coverage report when all CI checks pass. Thank you for adding extensive tests to accompany this change!

One thought crossed my mind while reviewing the code as well, I think it would be useful in a follow-up change to add an option to the [`install_demo_configuration` script](https://github.com/opensearch-project/security/blob/main/tools/install_demo_configuration.sh) to that can be used to install the demo configuration with FIPS compliance.

src/test/java/org/opensearch/security/auditlog/AbstractAuditlogiUnitTest.java
dancristiancecoi: Potentially, i will have a look! Would it be possible to re-run that test run to make sure its not an intermittent failure? (

cwperks: Sure, I just triggered a re-run. I have never seen that failure before so I think it is directly related to this PR.

dancristiancecoi: Thank you! Does look like its an issue with the PR. I will investigate

kumargu:
because PBKDF2 consumes higher CPU cycles (depending in iterations, which i see here is already very high) -- can a bad actor effect Opesearch cluster health by repeated password attempts?

dancristiancecoi:
> 1/ because PBKDF2 consumes higher CPU cycles (depending on iterations, which i see here is already very high) -- can a bad actor effect Opensearch cluster health by repeated password attempts?
> 
> 2/ Have we benchmarked how many cpu cycles are consumed and if this will work with single core cloud instances?

1) In theory that is possible. However, the high CPU cycles and slow hash computation are done on purpose to make brute-force attacks much harder so its about finding a balance between the risk of DoS attacks and brute-force attacks. OpenSearch has rate limiting so that should help with the former. 

OWASP recommends 600000 iterations or higher when using SHA256 so this is pretty much the lower bound for their PBKDF2 recommendations

2) No benchmarking has been done yet

dancristiancecoi:
> @dancristiancecoi some quick comments from the PR description. I think it would be helpful to expand on the help text for the hashing script to make it read more like a `man` page.
> 
> Without passing any params will it default to the same BCrypt hash as it does today?
> 
> For the default values I think we should include those in the help text of the hash script as well. For any option with multiple choices (like `--minor`) I think the help text should include the list of options (if its a small amount of options) or a reference to where you can find out more about the possible options.

Does something like this work? 

```
usage: hash.sh [-h] [-p <password>] [-env <Environment variable name>] [-a
       <hashing algorithm>] [-r <rounds (BCrypt)>] [-min <minor (BCrypt)>]
       [-l <length (PBKDF2)>] [-f <function (PBDKF2)>] [-i <iterations
       (PBKDF2)>]
 -h,--help                               Display the help information
 -p,--password <password>                Cleartext password to hash
 -env <Environment variable name>        Environment variable name to read
                                         password from
 -a,--algorithm <hashing algorithm>      Algorithm to use for password
                                         hashing. Valid values are: BCrypt
                                         | PBKDF2. Default: BCrypt
 -r,--rounds <rounds (BCrypt)>           Number of rounds to use in
                                         logarithmic form. Valid values
                                         are: 4 to 31. Default: 12
 -min,--minor <minor (BCrypt)>           Version of BCrypt algorithm to
                                         use. Valid values are: A | B | Y.
                                         Default: Y
 -l,--length <length (PBKDF2)>           Desired length of the final
                                         derived key. Default: 512
 -f,--function <function (PBDKF2)>       Pseudo-random function applied to
                                         the password. Valid values are
                                         SHA1 | SHA224 | SHA256 | SHA384 |
                                         SHA512. Default: SHA256
 -i,--iterations <iterations (PBKDF2)>   Number of times the pseudo-random
                                         function is applied to the
                                         password. Default: 600000
```

cwperks:
> Does something like this work @cwperks ?

Yes. Thanks for including the options and defaults in the help text.

src/integrationTest/java/org/opensearch/security/api/InternalUsersRestApiIntegrationTest.java
```diff
@@ -59,8 +61,10 @@ public class InternalUsersRestApiIntegrationTest extends AbstractConfigEntityApi
 
     private final static String SOME_ROLE = "some-role";
 
-    private final PasswordHasher passwordHasher = new BCryptPasswordHasher();
-
+    private final PasswordHasher passwordHasher = PasswordHasherFactory.createPasswordHasher(Settings
+            .builder()
+            .put(ConfigConstants.SECURITY_PASSWORD_HASHING_ALGORITHM, ConfigConstants.BCRYPT)
```
kumargu, line 66: nit: should we just say ```PASSWORD_HASHING_ALGORITHM"?

src/main/java/org/opensearch/security/hasher/AbstractPasswordHasher.java
```diff
@@ -0,0 +1,33 @@
...
+abstract class AbstractPasswordHasher implements PasswordHasher {
+
+    HashingFunction hashingFunction;
+
+    public abstract String hash(char[] password);
+
+    public abstract boolean check(char[] password, String hash);
+
+    protected void cleanup(CharBuffer password) {
+        password.clear();
```
kumargu, line 28: can we please have a ```null``` check on password check?

src/main/java/org/opensearch/security/hasher/AbstractPasswordHasher.java
```diff
@@ -0,0 +1,33 @@
...
+
+abstract class AbstractPasswordHasher implements PasswordHasher {
+
+    HashingFunction hashingFunction;
+
+    public abstract String hash(char[] password);
+
+    public abstract boolean check(char[] password, String hash);
+
+    protected void cleanup(CharBuffer password) {
```
kumargu, line 27: let's please add some java docs.

src/main/java/org/opensearch/security/hasher/PBKDF2PasswordHasher.java
```diff
@@ -0,0 +1,86 @@
...
+            cleanup(passwordBuffer);
+        }
+    }
+
+    @SuppressWarnings("removal")
+    @Override
+    public boolean check(char[] password, String hash) {
+        if (password == null || password.length == 0) {
+            throw new OpenSearchSecurityException("Password cannot be empty or null");
+        }
```
kumargu, line 65: can we enforce this check in the abstract class? this will not only reduce code duplication; but will also make sure the check is enforced across all implementation.

src/main/java/org/opensearch/security/hasher/PBKDF2PasswordHasher.java
```diff
@@ -0,0 +1,86 @@
...
+        CharBuffer passwordBuffer = CharBuffer.wrap(password);
+        try {
+            SecurityManager securityManager = System.getSecurityManager();
+            if (securityManager != null) {
+                securityManager.checkPermission(new SpecialPermission());
+            }
+            return AccessController.doPrivileged(
+                (PrivilegedAction<String>) () -> Password.hash(passwordBuffer).addRandomSalt(64).with(hashingFunction).getResult()
```
kumargu, line 53: let's add the salt's length as a static value, its too hidden out here. 

(from wikipedia)

> The US [National Institute of Standards and Technology](https://en.wikipedia.org/wiki/National_Institute_of_Standards_and_Technology) recommends a salt length of at least 128 bits.

Should we pick 128 as the length?

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
```diff
@@ -0,0 +1,88 @@
...
+        PasswordHasher passwordHasher;
+        switch (algorithm.toLowerCase()) {
+            case BCRYPT:
+                passwordHasher = getBCryptHasher(settings);
+                break;
+            case PBKDF2:
+                passwordHasher = getPBKDF2Hasher(settings);
+                break;
+            default:
+                throw new IllegalArgumentException("Password hashing algorithm not supported");
```
kumargu, line 38: can we log the "unsupported" algorithm being requested here? it will help to debug issues faster.

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private static PasswordHasher getBCryptHasher(Settings settings) {
+        int rounds = settings.getAsInt(
+            ConfigConstants.SECURITY_PASSWORD_HASHING_BCRYPT_ROUNDS,
+            ConfigConstants.SECURITY_PASSWORD_HASHING_BCRYPT_ROUNDS_DEFAULT
+        );
+        String minor = settings.get(
+            ConfigConstants.SECURITY_PASSWORD_HASHING_BCRYPT_MINOR,
+            ConfigConstants.SECURITY_PASSWORD_HASHING_BCRYPT_MINOR_DEFAULT
+        ).toUpperCase();
```
kumargu, line 51: i am not sure why we need a non-default value for this setting. the default should be always the standard, no?

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
```diff
@@ -0,0 +1,88 @@
...
+
+        if (rounds < 4 || rounds > 31) {
+            throw new IllegalArgumentException("BCrypt rounds must be between 4 and 31.");
+        }
+        if (!minor.equals("A") && !minor.equals("B") && !minor.equals("Y")) {
+            throw new IllegalArgumentException("BCrypt minor must be 'A', 'B', or 'Y'.");
+        }
```
kumargu, line 58: can we make it more readable 

```
Set<String> ALLOWED_BCRYPT_MINORS = Set.of("A", "B", "Y");

if (!ALLOWED_BCRYPT_MINORS.contains(minor)) {
    throw new IllegalArgumentException("BCrypt minor must be 'A', 'B', or 'Y'.");
}

or 

```
if (!Set.of("A", "B", "Y").contains(minor)) {
    throw new IllegalArgumentException("BCrypt minor must be 'A', 'B', or 'Y'.");
}

```

```

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -144,6 +146,23 @@ public class ConfigConstants {
...
+
+    public static final String SECURITY_PASSWORD_HASHING_BCRYPT_ROUNDS = "plugins.security.password.hashing.bcrypt.rounds";
+    public static final int SECURITY_PASSWORD_HASHING_BCRYPT_ROUNDS_DEFAULT = 12;
+    public static final String SECURITY_PASSWORD_HASHING_BCRYPT_MINOR = "plugins.security.password.hashing.bcrypt.minor";
+    public static final String SECURITY_PASSWORD_HASHING_BCRYPT_MINOR_DEFAULT = "Y";
+
+    public static final String SECURITY_PASSWORD_HASHING_ALGORITHM = "plugins.security.password.hashing.algorithm";
+    public static final String SECURITY_PASSWORD_HASHING_ALGORITHM_DEFAULT = BCRYPT;
+    public static final String SECURITY_PASSWORD_HASHING_PBKDF2_ITERATIONS = "plugins.security.password.hashing.pbkdf2.iterations";
+    public static final int SECURITY_PASSWORD_HASHING_PBKDF2_ITERATIONS_DEFAULT = 600000;
```
kumargu, line 160: nit:  ```s/60000/60_000```

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
```diff
@@ -0,0 +1,88 @@
...
+
+        if (!pbkdf2Function.matches("SHA(1|224|256|384|512)")) {
+            throw new IllegalArgumentException("PBKDF2 function must be one of SHA1, SHA224, SHA256, SHA384, or SHA512.");
+        }
+        if (iterations <= 0) {
+            throw new IllegalArgumentException("PBKDF2 iterations must be a positive integer.");
+        }
+        if (length <= 0) {
+            throw new IllegalArgumentException("PBKDF2 length must be a positive integer.");
+        }
```
kumargu, line 85: i think we should also include the invalid value in the error context.

src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
```diff
@@ -144,8 +144,11 @@ void updateAdminPassword() throws IOException {
         try {
             final PasswordValidator passwordValidator = PasswordValidator.of(
                 Settings.builder()
-                    .put(SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX, "(?=.*[A-Z])(?=.*[^a-zA-Z\\\\d])(?=.*[0-9])(?=.*[a-z]).{8,}")
-                    .put(SECURITY_RESTAPI_PASSWORD_MIN_LENGTH, DEFAULT_PASSWORD_MIN_LENGTH)
+                    .put(
+                        ConfigConstants.SECURITY_RESTAPI_PASSWORD_VALIDATION_REGEX,
+                        "(?=.*[A-Z])(?=.*[^a-zA-Z\\\\d])(?=.*[0-9])(?=.*[a-z]).{8,}"
```
kumargu, line 149: nit: let's wrap this is a constant.

src/integrationTest/java/org/opensearch/security/api/InternalUsersRestApiIntegrationTest.java
dancristiancecoi: I am happy to change it though I see that most of other constants in that file have the "SECURITY_" prefix

src/main/java/org/opensearch/security/hasher/PBKDF2PasswordHasher.java
dancristiancecoi: Agreed!

kumargu (commented):
Thanks @dancristiancecoi for the PR. left a few comments.

src/integrationTest/java/org/opensearch/security/api/InternalUsersRestApiIntegrationTest.java
kumargu: Ah ok. then i think what you have is good.

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
dancristiancecoi: The default here (Y) is actually different to what Password4j itself defaults to (B) so it is consistent with the previous OpenBSDBCrypt generated hashes.

Honestly I am split between allowing this as an option or not. On one hand its not something that most users would want to change but on the other hand it might be good to have control over this if there is a new version of BCrypt in the future

kumargu: i think if something is not needed right away or not expected to change anytime soon, we should avoiding keeping it in config.

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
```diff
@@ -0,0 +1,88 @@
...
+            ConfigConstants.SECURITY_PASSWORD_HASHING_PBKDF2_FUNCTION_DEFAULT
+        ).toUpperCase();
+
+        int iterations = settings.getAsInt(
+            ConfigConstants.SECURITY_PASSWORD_HASHING_PBKDF2_ITERATIONS,
+            ConfigConstants.SECURITY_PASSWORD_HASHING_PBKDF2_ITERATIONS_DEFAULT
+        );
+        int length = settings.getAsInt(
+            ConfigConstants.SECURITY_PASSWORD_HASHING_PBKDF2_LENGTH,
+            ConfigConstants.SECURITY_PASSWORD_HASHING_PBKDF2_LENGTH_DEFAULT
```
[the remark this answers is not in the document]
dancristiancecoi: The length is in bytes while the function is described in bits  (so technically it should be at minimum 32 byte length for SHA-256 but as far as I know thats not secure). 

Researching I am finding very different recommendations for the length based on the function, though Password4j defaults to 256 bytes for SHA-256 so I will change it to that. 

I think its valuable for users to be able to change both the pseudo-random function and the length as both will have an impact on the speed and the security of the password.

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
dancristiancecoi: Just to add more info: Y version of BCrypt is an older version than B so it still might be useful for customers to be able to change to a more up-to-date version if they wish. 

Another option is to make the default "B" and not make it configurable. That should work without issues for authentication as we validate with a BCrypt function based on the hash but my worry is that I am missing some other usages of hashing that might be affected by changing the default algorithm (like the replaceEnv functionality). 

Thoughts?

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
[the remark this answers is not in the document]
kumargu: >I think its valuable for users to be able to change both the pseudo-random function and the length as both will have an impact on the speed and the security of the password.

ok -- sounds good -- let's keep the configs.

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
kumargu: > but my worry is that I am missing some other usages of hashing that might be affected by changing the default algorithm (like the replaceEnv functionality).

ok. in that case, let's keep ```Y```  as default, don't want to add regression / backward compatibility issues from such a change. 

I think we should open a separate PR if we want to make ```B``` as the new default cc @cwperks

cwperks: Changing the default is tricky because it would be a breaking change, right?

If a user wants to keep their existing internal user list and change the password hashing strategy then that would require updating the hashes of all internal users. One way to accomplish that is to keep the old hashes when starting a cluster after changing the hashing strategy. For each user that sends a request to the cluster, on the first request they would need to authenticate with the old hash and then in the backend we can compute the new hash and store it. On subsequent requests they would authenticate with the new hash.

src/main/java/org/opensearch/security/hasher/PBKDF2PasswordHasher.java
dancristiancecoi: I've moved the null check logic into separate methods in the abstract class. Unfortunately can't enforce the class extending it to call them but at least it sorts out the code duplication

src/test/java/org/opensearch/security/hasher/PBKDF2PasswordHasherTests.java
```diff
@@ -0,0 +1,60 @@
...
+    public void setup() {
+        passwordHasher = new PBKDF2PasswordHasher(
+            ConfigConstants.SECURITY_PASSWORD_HASHING_PBKDF2_FUNCTION_DEFAULT,
+            ConfigConstants.SECURITY_PASSWORD_HASHING_PBKDF2_ITERATIONS_DEFAULT,
+            ConfigConstants.SECURITY_PASSWORD_HASHING_PBKDF2_LENGTH_DEFAULT
+        );
+    }
+
+    @Test
+    public void shouldGenerateValidHashesFromParameters() {
```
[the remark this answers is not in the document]
dancristiancecoi: This is already covered by unit tests of PasswordHasherFactory

src/test/java/org/opensearch/security/hasher/BCryptPasswordHasherTests.java
```diff
@@ -53,49 +41,23 @@ public void shouldBeBackwardsCompatible() {
...
+        assertThat(hasher.check(wrongPassword.toCharArray(), hash), is(false));
 
-    @Test
-    public void shouldCleanupPasswordCharBuffer() {
-        char[] password = new char[] { 'p', 'a', 's', 's', 'w', 'o', 'r', 'd' };
-        CharBuffer passwordBuffer = CharBuffer.wrap(password);
-        passwordHasher.hash(password);
-        assertThat("\0\0\0\0\0\0\0\0", is(new String(password)));
-        assertThat("\0\0\0\0\0\0\0\0", is(passwordBuffer.toString()));
+        hasher = new BCryptPasswordHasher("Y", 13);
```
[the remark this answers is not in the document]
dancristiancecoi: yes, inside the unit tests for PasswordHasherFactory

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
dancristiancecoi: Password4j will throw their own BadParametersException (implements IllegalArgumentException). Personally, i'd leave the validation here so we don't expose Password4j exceptions needlessly

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
dancristiancecoi: For existing hashes it should still work as when we validate we instantiate the hashing function based on the hash (the first two characters of the hash indicate if its 2y, 2b etc). But if we don't rehash every user after we change the default we will have the weird state where some stored hashes are of the 2y type and other newer hashes will be of 2b type.

cwperks (commented):
@dancristiancecoi The changes look good to me. Can you also create a PR on the [documentation-website](https://github.com/opensearch-project/documentation-website) to add documentation about the settings being introduced?

dancristiancecoi:
> @dancristiancecoi The changes look good to me. Can you also create a PR on the [documentation-website](https://github.com/opensearch-project/documentation-website) to add documentation about the settings being introduced?
 
Thank you! 

I've created it here: https://github.com/opensearch-project/documentation-website/issues/7669

kumargu:
LGTM. thanks for taking care of the comments.

kumargu (approved):

src/integrationTest/java/org/opensearch/security/api/InternalUsersRestApiIntegrationTest.java
```diff
@@ -59,8 +61,9 @@ public class InternalUsersRestApiIntegrationTest extends AbstractConfigEntityApi
 
     private final static String SOME_ROLE = "some-role";
 
-    private final PasswordHasher passwordHasher = new BCryptPasswordHasher();
-
+    private final PasswordHasher passwordHasher = PasswordHasherFactory.createPasswordHasher(
```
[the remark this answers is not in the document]
dancristiancecoi: yeah you are right, I will remove

src/integrationTest/java/org/opensearch/security/api/AbstractApiIntegrationTest.java
```diff
@@ -85,7 +87,9 @@ public abstract class AbstractApiIntegrationTest extends RandomizedTest {
 
     public static LocalCluster localCluster;
 
-    public static PasswordHasher passwordHasher = new BCryptPasswordHasher();
+    public static PasswordHasher passwordHasher = PasswordHasherFactory.createPasswordHasher(
```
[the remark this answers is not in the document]
dancristiancecoi: can be, will make it final! thanks

src/main/java/org/opensearch/security/hasher/AbstractPasswordHasher.java
```diff
@@ -0,0 +1,79 @@
...
+     * Checks if the given hash is null or empty and throws an exception if it is.
+     *
+     * @param hash the hash to check
+     * @throws OpenSearchSecurityException if the hash is null or empty
+     */
+    protected void checkHashNotNullOrEmpty(String hash) {
+        if (isNullOrEmpty(hash)) {
+            throw new OpenSearchSecurityException("Hash cannot be empty or null");
+        }
+    }
```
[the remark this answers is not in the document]
dancristiancecoi: The reason behind having them as a separate method is to avoid converting the password to String as much as I can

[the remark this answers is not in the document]
DarshitChanpura: okay, my push was to avoid repetitive code but if it adds clarity I'm okay with that.

src/main/java/org/opensearch/security/hasher/PBKDF2PasswordHasher.java
```diff
@@ -0,0 +1,82 @@
...
+
+class PBKDF2PasswordHasher extends AbstractPasswordHasher {
+
+    private static final int DEFAULT_SALT_LENGTH = 128;
+
+    @SuppressWarnings("removal")
+    PBKDF2PasswordHasher(String function, int iterations, int length) {
+        SecurityManager securityManager = System.getSecurityManager();
+        if (securityManager != null) {
+            securityManager.checkPermission(new SpecialPermission());
```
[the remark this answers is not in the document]
dancristiancecoi: Originally I had the password4j  permissions in the plugin-security.policy files but that did not seem to actually load them properly. Even the existing permissions in plugin-security.policy don't seem to be passed to the library without using doPrivileged unfortunately

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
dancristiancecoi: Agreed but I am not sure how I could approach this warning considering this configuration would be changed at deploy time (most likely). 

Is there an example of a similar warning for some other options that I can look at?

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
dancristiancecoi: Sorry, missed this comment initially! Will make the change.

src/main/java/org/opensearch/security/hasher/PBKDF2PasswordHasher.java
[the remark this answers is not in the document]
DarshitChanpura: hmm.. something to look at later, outside the scope of this PR.

src/test/java/org/opensearch/security/dlic/rest/api/InternalUsersApiActionValidationTest.java
```diff
@@ -193,7 +195,15 @@ public void validateSecurityRolesWithImmutableRolesMappingConfig() throws Except
     }
 
     private InternalUsersApiAction createInternalUsersApiAction() {
-        return new InternalUsersApiAction(clusterService, threadPool, userService, securityApiDependencies, new BCryptPasswordHasher());
+        return new InternalUsersApiAction(
+            clusterService,
+            threadPool,
+            userService,
+            securityApiDependencies,
+            PasswordHasherFactory.createPasswordHasher(
```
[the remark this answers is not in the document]
dancristiancecoi: yes, good catch!

src/main/java/org/opensearch/security/hasher/PasswordHasherFactory.java
DarshitChanpura: > Is there an example of a similar warning for some other options that I can look at?

Not that I'm aware of. This may have to be added to the setting description itself. If we are planning to include this setting in the default opensearch.yml then a comment should be added to that yml as well IMO.

dancristiancecoi:
Any suggestions on where to document these new settings? 

One option is to document them in [Configuring OpenSearch -> Security Settings > Expert-level settings](https://opensearch.org/docs/latest/install-and-configure/configuring-opensearch/security-settings/)

We could also add a new chapter under **Password settings** in [Security in OpenSearch -> Configuration -> Modifying the YAML files](https://opensearch.org/docs/latest/security/configuration/yaml/#opensearchyml)

Which option is the most suitable one if at all?

DarshitChanpura:
> Which option is the most suitable one if at all?

IMO we should add it under [Configuring OpenSearch -> Security Settings > Expert-level settings](https://opensearch.org/docs/latest/install-and-configure/configuring-opensearch/security-settings/) since it lists all available security settings and also highlights that this is an expert-level setting.

cwperks (approved):

DarshitChanpura (approved):
Thanks @dancristiancecoi for this contribution!

dancristiancecoi:
Thanks a lot for the reviews everyone!

PR #4528, opened 2024-07-05
[BUG] Blake2b hashing for Masked Fields does not apply salt correctly

prabhask5:
### Description
As per: https://github.com/opensearch-project/security/issues/4274 change the calling Blake hash function to set the salt to to the default salt rather than the personalization.

Bug fix

Refer to: https://github.com/opensearch-project/security/issues/4274

Hashes for field masking should be different, but this behavior should be the expected behavior since the salt is set to the actual default salt value.

### Issues Resolved
https://github.com/opensearch-project/security/issues/4274

No

### Testing
Fixed field masking tests accordingly to reflect new hashes.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/configuration/MaskedField.java
+++ b/src/main/java/org/opensearch/security/configuration/MaskedField.java
@@ -152,7 +152,7 @@ public class MaskedField {
     private byte[] blake2bHash(byte[] in) {
         // Salt is passed incorrectly but order of parameters is retained at present to ensure full backwards compatibility
         // Tracking with https://github.com/opensearch-project/security/issues/4274
-        final Blake2b hash = new Blake2b(null, 32, null, defaultSalt);
+        final Blake2b hash = new Blake2b(null, 32, defaultSalt, null);
         hash.update(in, 0, in.length);
         final byte[] out = new byte[hash.getDigestSize()];
         hash.digest(out, 0);


src/main/java/org/opensearch/security/configuration/MaskedField.java
```diff
@@ -152,7 +152,7 @@ private static byte[] customHash(byte[] in, final String algorithm) {
     private byte[] blake2bHash(byte[] in) {
         // Salt is passed incorrectly but order of parameters is retained at present to ensure full backwards compatibility
         // Tracking with https://github.com/opensearch-project/security/issues/4274
-        final Blake2b hash = new Blake2b(null, 32, null, defaultSalt);
+        final Blake2b hash = new Blake2b(null, 32, defaultSalt, null);
```
terryquigleysas, line 155: Hi @prabhask5 I had made the same code change in my original PR but concerns were raised that some customers may be reliant on certain hash values, even though generated by this incorrect call, so I backed the change out and raised the bug to discuss the best approach going forward.

One possibility here would be to add the fix but wrap it in an if statement that uses the newer, correct code from OpenSearch version 3.x onwards. It could then also be documented as a potential breaking change for some customers. What do you think?

Note I also had to change the corresponding tests that were reliant on this hashing in my original PR.

PR #4533, opened 2024-07-08
Update PULL_REQUEST_TEMPLATE to include an API spec change in the checklist.

dblock:
Coming from https://github.com/opensearch-project/opensearch-api-specification/issues/387, update PR template to include a checkbox that requests that all API changes be documented in the OpenAPI spec used to generate clients.


RyanL1997 (approved):

reta (approved):

DarshitChanpura:
Adding a backport label in case there are updates being made directly to 2.x

Issue #4493, opened 2024-06-26
[RFC] Retire support for V6 configuration


nibix:
# Preface

Note: I am asking for opinions and comments on the following topic because it influences the work on #3870. 

# Current situation

## Configuration file formats

At the moment, the security plugin supports two slightly different configuration file formats:

In the `*.yml` files, these can be told apart by the `_meta.config_version` property. Internally, there these can be told apart by the `V6` vs `V7` suffix of class names: `RoleV6` vs `RoleV7`. `V7` in class names corresponds to `config_version` = 2.

Example for the differences:

**New, V7:**

```
_meta:
  type: "roles"
  config_version: 2
example_role:
  cluster_permissions: 
    - CLUSTER_COMPOSITE_OPS
  index_permissions:
    - index_patterns:
        - "foo*"
      dls: 'query'
      allowed_actions:
        - CRUD
```

**Old, V6:**

```
example_role:
  cluster:
    - CLUSTER_COMPOSITE_OPS
  indices:
    'logstash-*':
      '*':
        - CRUD
        - CREATE_INDEX
    '_dls_': 'query'
```

Actually, the number 6 in the class names still stems from the Elasticsearch version 6. Thus, the `roles.yml` configuration type allows the specification of document types, even though these are not supported any more in OpenSearch 2. Consequently, the implementation actually ignores this information.

Thus, these configuration files stem from the ODFE times, when it was still available for Elasticsearch 6.

Further changes include some cleanup and restructuring of the configuration file format.

## Privilege evaluation code

Even though the central privilege evaluation code is located in the class `PrivilegeEvaluator`, the actual logic which interprets the the roles is located in the classes `ConfigModelV6` and `ConfigModelV7`:

- https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/ConfigModelV6.java
- https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java

Depending on the used configuration version, either  `ConfigModelV6` and `ConfigModelV7` is used.  Even though both classes should apply the same logic, both contain separate implementations. Both implementations are very complex: `ConfigModelV6` has 1316 lines of code and `ConfigModelV7` has 1267 lines of code.

Especially for security critical code, it seems to be an anti-pattern to have such fundamental and critical code doubled up. It increases maintenance effort and cognitive load.

[...]

PR #4546, opened 2024-07-11
Remove support for v6 configuration

nibix:
### Description

This solves #4493 and removes support for v6 configuration. On `master`, v6 configuration is not  usable any more anyway, see #4745 for the verification.

As an additional benefit, this makes it possible to use `SecurityDynamicConfiguration<>` instances in a generically type-safe manner. So far, one usually needs to use  `SecurityDynamicConfiguration<?>` instances and do unsafe casts.

If it is desired to continue supporting v6 configuration for OpenSearch 2.x, there's the PR #4753 for that.

* Category: Refactoring
* Why these changes are required? - Minimize duplicate logic for privilege evaluation, remove unnecessary code
* What is the old behavior before changes and new behavior after changes? - Any try to use v6 configuration is now fail-fast 

### Issues Resolved

- Solves #4493 

### Check List


src/test/java/org/opensearch/security/securityconf/SecurityRolesPermissionsV6Test.java
```diff
@@ -1,193 +0,0 @@
...
-import org.opensearch.security.user.User;
-
-import org.mockito.quality.Strictness;
-
-import static org.mockito.Mockito.doReturn;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
-import static org.mockito.Mockito.withSettings;
-
-public class SecurityRolesPermissionsV6Test {
```
DarshitChanpura, line 49: why delete v6 test?

DarshitChanpura (commented):
Changes look good to me @nibix. Could you please check and resolve the conflict?

src/test/java/org/opensearch/security/securityconf/SecurityRolesPermissionsV6Test.java
nibix: Now we convert "behind the scenes" any V6 config into a V7 config. Thus, the business logic which operates on the V6 config is no longer in use - that's especially the classes `ConfigModelV6` and `DynamicConfigModelV6`. I actually forgot to delete these classes in the first commit - in a second commit, I have now deleted these classes.

As the `SecurityRolesPermissionsV6Test` tests the business logic in the  `ConfigModelV6`  class, it has now nothing to test anymore.

nibix:
BTW, I felt that I should describe the idea of the change a bit more in detail. See the updated description of this PR.

cwperks:
With this change, would a user still be able to use the Migrate API to change the documents stored in the security index to the V7 format?

If a user uses securityadmin to export the security index, what format will the generated yml files be in?

nibix:
@cwperks 

> With this change, would a user still be able to use the Migrate API to change the documents stored in the security index to the V7 format?
> If a user uses securityadmin to export the security index, what format will the generated yml files be in?

In the end, it is up to us to define these behaviors. I would expect that both functionalities remain unaffected. Thus, exporting the configuratin will yield v6/v1 configuration files. Also, the migrate API and the migrate command continue to work as before.

cwperks:
> In the end, it is up to us to define these behaviors. I would expect that both functionalities remain unaffected. Thus, exporting the configuratin will yield v6/v1 configuration files. Also, the migrate API and the migrate command continue to work as before.

Sounds good. What work is remaining on this PR in order to flip it from Draft?

nibix:
@cwperks 

> What work is remaining on this PR in order to flip it from Draft?

Mainly reviewing test failures and reviewing the special cases, like migration and API - as mentioned above. 

So, I would then focus on this again?

cwperks:
If its an enabler for https://github.com/opensearch-project/security/pull/4380, then IMO I think we should move this forward

nibix:
Note: This needs #4744 to be merged beforehand to have tests passing

willyborankin (dismissed):

src/main/java/org/opensearch/security/configuration/ConfigurationMap.java
```diff
@@ -0,0 +1,93 @@
...
+
+import com.google.common.collect.ImmutableMap;
+
+import org.opensearch.security.securityconf.impl.CType;
+import org.opensearch.security.securityconf.impl.SecurityDynamicConfiguration;
+
+/**
+ * Allows type safe access of configuration instances via the configuration type
+ */
+public class ConfigurationMap {
```
cwperks, line 24: Nice! I think this class enhances the readability of the code.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -611,36 +609,37 @@ public Map<CType, SecurityDynamicConfiguration<?>> getConfigurationsFromIndex(
...
-        return retVal;
+        return result;
     }
 
-    private Map<CType, SecurityDynamicConfiguration<?>> validate(Map<CType, SecurityDynamicConfiguration<?>> conf, int expectedSize)
-        throws InvalidConfigException {
+    private ConfigurationMap validate(ConfigurationMap conf, int expectedSize) throws InvalidConfigException {
 
-        if (conf == null || conf.size() != expectedSize) {
+        if (conf == null || conf.size() < expectedSize) {
```
cwperks, line 642: Why change the operator here?

src/test/resources/legacy/securityconfig_v6/action_groups.yml
```diff
@@ -1,136 +0,0 @@
-OPENDISTRO_SECURITY_UNLIMITED:
```
cwperks, line 1: Its great to see how many files can be removed with this change. Nice!

cwperks (dismissed):
Left one minor comment. This change LGTM!

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nibix: I can rollback that change. This PR is based on the auto-conversion, where this seemed to be necessary because we sometimes created more configs than expected (the tenant config).

nibix:
@willyborankin @cwperks I did one more minor commit: https://github.com/opensearch-project/security/pull/4546/commits/5ebe5007c4a1cf64b1ed1129d0c3b1c7d9cdf11b   ... could you please re-approve?

willyborankin (approved):

cwperks (approved):

cwperks:
> @willyborankin @cwperks I did one more minor commit: https://github.com/opensearch-project/security/commit/5ebe5007c4a1cf64b1ed1129d0c3b1c7d9cdf11b ... could you please re-approve?

Done

PR #4548, opened 2024-07-11
Fixed test failures in FlsAndFieldMaskingTests

nibix:
### Description

Fixes failing tests in FlsAndFieldMaskingTests. 

The tests `flsWithIncludesRulesIncludesFieldMappersFromPlugins` and `testFlsOnAClosedAndReopenedIndex` both used the REST API to create an identical test user. However, the test which executes second would get a 200 Ok REST status instead a 201 Created REST status, as the `PUT` would operate on a resource that already exists. 

This caused the test to fail.

This moves the user to the global user definition of the test.

* Category: Test fix
* Why these changes are required? - The test failed before
* What is the old behavior before changes and new behavior after changes? - None

### Check List


willyborankin (approved):

cwperks (approved):
Good catch. Thank you @nibix !

cwperks:
@nibix Can you raise a backport PR for this change?

Issue #4515, opened 2024-07-02
[FEATURE] Remove hard coded action names, pull from core instead


derek-ho:
**Is your feature request related to a problem?**
Hard coded action names like: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/SecurityIndexAccessEvaluator.java#L113 should not exist, and instead we should pull from core. That way if there are future changes we are insulated. 
**What solution would you like?**
Pull action names from core
**What alternatives have you considered?**
None
**Do you have any additional context?**
No

scrawfor99:
[Triage] HI @derek-ho thanks for filing this issue. This seems like a smart change so going to mark as triaged.

PR #4582, opened 2024-07-19
Fixed READ_ACTIONS required by TermsAggregationEvaluator

nibix:
The READ_ACTIONS array is passed by TermsAggregationEvaluator to securityRoles.getAllPermittedIndicesForDashboards() which checks whether privileges are available for all actions specified in READ_ACTIONS. READ_ACTIONS also contained the string "indices:data/read/field_caps*", which is actually wrong, because it is not an action but looks like pattern. However, the code behind securityRoles.getAllPermittedIndicesForDashboards() will never treat these strings as patterns. The "*" is only considered a normal, bare character. Patterns (via WildcardMatcher class) will be only applied to these strings.

This had the effect that a bare privilege "indices:data/read/field_caps" was not sufficient to fulfill the requirement. It was necessary to have either "indices:data/read/field_caps*" in ones roles.yml or something broader like "indices:data/read/*". The latter is the most likely case, which is the reason why in most cases this gets unnoticed.

Observed while working on #3870 and decoupled from #4380.

### Description
* Category:  Bug fix
* Why these changes are required? - The code created an impression that patterns were supported at the particular place, while they were actually not supported.
* What is the old behavior before changes and new behavior after changes? - No significant changes. It will be sufficient to specify the privilege "indices:data/read/field_caps" instead of ""indices:data/read/field_caps*" to have field caps on indices properly working, but in practice this seems to be a rare case.

### Issues Resolved

Contributes to #3870

### Testing

Integration testing via existing tests

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/privileges/TermsAggregationEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/TermsAggregationEvaluator.java
@@ -53,9 +53,7 @@ public class TermsAggregationEvaluator {
         "indices:data/read/mget",
         "indices:data/read/get",
         "indices:data/read/search",
-        "indices:data/read/field_caps*"
-        // "indices:admin/mappings/fields/get*"
-    };
+        "indices:data/read/field_caps" };
 
     private static final QueryBuilder NONE_QUERY = new MatchNoneQueryBuilder();
 


DarshitChanpura (commented):
Thanks @nibix for addressing this! is there a way this can be tested to ensure that `*` is a bare character here and would not expand?

src/main/java/org/opensearch/security/privileges/TermsAggregationEvaluator.java
```diff
@@ -53,9 +53,7 @@ public class TermsAggregationEvaluator {
         "indices:data/read/mget",
         "indices:data/read/get",
         "indices:data/read/search",
-        "indices:data/read/field_caps*"
```
shikharj05, line 56: Thanks for identifying the issue and fixing it. However, wouldn't removing/fixing this permission in next version would break existing users using ``indices:data/read/field_caps*``, right?

nibix: Users which are use `indices:data/read/field_caps*` in their roles.yml are using a wildcard - which includes both the literal strings `indices:data/read/field_caps*` and `indices:data/read/field_caps`.

nibix:
@DarshitChanpura 

> Thanks @nibix for addressing this! is there a way this can be tested to ensure that `*` is a bare character here and would not expand?

This is quite easy to verify by a simple manual static code analysis. This is the call path:

https://github.com/opensearch-project/security/blob/26a4c0bef937a0105c9851be9ddea61e857cee7e/src/main/java/org/opensearch/security/privileges/TermsAggregationEvaluator.java#L89-L95

https://github.com/opensearch-project/security/blob/26a4c0bef937a0105c9851be9ddea61e857cee7e/src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java#L391

https://github.com/opensearch-project/security/blob/26a4c0bef937a0105c9851be9ddea61e857cee7e/src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java#L564

This shows that the `actions` parameter is never passed to a `WildcardMatcher.create()` method. However, it is only this method which gives you the possibility to interpret strings with patterns.

shikharj05 (approved):

DarshitChanpura (dismissed):

cwperks:
@nibix wdyt about removing all hardcoding and pulling the static action name from core if possible (it may not be possible in all cases)?

Related issue: https://github.com/opensearch-project/security/issues/4515

cwperks (approved):

shikharj05 (approved):

nibix:
@cwperks @DarshitChanpura is there anything still missing that blocks this?

cwperks:
> @cwperks @DarshitChanpura is there anything still missing that blocks this?

It needs one more approval. @DarshitChanpura @peternied @scrawfor99 @willyborankin @reta @RyanL1997 Can one of you take a look?

scrawfor99 (approved):

Issue #4480, opened 2024-06-21
[BUG] Exception when reloading self signed certs via REST API


reneradoi:
**What is the bug?**
As part of the TLS cert renewal process I want to use the [REST API](https://opensearch.org/docs/latest/security/configuration/tls/#reload-tls-certificates-on-the-transport-layer) for hot reload of transport and http certificates. When reloading the already used certificates, everything works fine (response code 200). But when I want to apply newly generated, self signed certificates, I always run into the same exception telling me that new certs do not have valid Issuer DN, Subject DN or SAN.

I’ve checked my configuration and the validity of the certs, everything seems to be fine here. If the application is restarted, the new certs are also picked up without any issue. Only reloading via the API produces this error.

**How can one reproduce the bug?**
- Configure and start OpenSearch with self signed certificates as in [Generate Certificates](https://opensearch.org/docs/latest/security/configuration/generate-certificates/), but with a list of multiple entries with `DNSName` in the Subject Alternative Names section
- after the application has settled, generate new self signed certificates (the same way like before, with exactly the same Issuer, Subject and Subject Alternative Names
- try to upload the new certificate via API call to `_plugins/_security/api/ssl/transport/reloadcerts`

**What is the expected behavior?**
- Validation of the new certificate should be ok because Issuer, Subject and SANs are the same
- new certificate should be accepted and API call should return `200`

**What is your host/environment?**
 - OS: Ubuntu 22.04
 - Version: Opensearch 2.14.0
 - Plugins:

**Do you have any screenshots?**
API response:
```
{"error":{"root_cause":[{"type":"i_o_exception","reason":"OpenSearchSecurityException[Error while initializing http SSL layer from PEM: java.lang.Exception: New Certs do not have valid Issuer DN, Subject DN or SAN.]; nested: Exception[New Certs do not have valid Issuer DN, Subject DN or SAN.];"}],"type":"i_o_exception","reason":"OpenSearchSecurityException[Error while initializing http SSL layer from PEM: java.lang.Exception: New Certs do not have valid Issuer DN, Subject DN or SAN.]; nested: Exception[New Certs do not have valid Issuer DN, Subject DN or SAN.];","caused_by":{"type":"security_exception","reason":"Error while initializing http SSL layer from PEM: java.lang.Exception: New Certs do not have valid Issuer DN, Subject DN or SAN.","caused_by":{"type":"exception","reason":"New Certs do not have valid Issuer DN, Subject DN or SAN."}}},"status":500}
```

[...]

PR #4624, opened 2024-08-06
Sort the DNS Names in the SANs

Mehdi-Bendriss:
### Description
* Category: Bug fix
  * During the hot reload of a certificate, comparing **equivalent** certificates with multiple DNS Names on the SAN will fail if the order of the entries doesn't match.
  * Previously, the DN list of the old vs new certificates was compared as it was read, while this PR ensures both lists are sorted prior to the comparison   

### Issues Resolved
https://github.com/opensearch-project/security/issues/4480

### Testing
manual testing - as it is a pretty straightforward change.
happy to add a unit test if needed

### Check List


willyborankin:
@Mehdi-Bendriss, thank you for your contribution. Unfortunately, it won't fix the problem. The method where you added the comparison is used only in the REST API. To address the bug, you need to sort the DN names in the method at:
https://github.com/opensearch-project/security/blob/ef19743b6aa0c56ba7d759e737785fa13381f86d/src/main/java/org/opensearch/security/ssl/DefaultSecurityKeyStore.java#L666 method.

willyborankin:
> @willyborankin thanks for the response.
> 
> I am not sure I'm following. The sorting in my PR happens in `getSubjectAlternativeNames` which is then called from within [`hasValidDNs`](https://github.com/opensearch-project/security/blob/ef19743b6aa0c56ba7d759e737785fa13381f86d/src/main/java/org/opensearch/security/ssl/DefaultSecurityKeyStore.java#L671C32-L671C58) for both the old and new certificates.
> 
> Is my understanding correct or did I miss something?

Ahh sorry didn't notice `Function<>` in the method. All good my comment is misleading.

src/main/java/org/opensearch/security/ssl/DefaultSecurityKeyStore.java
```diff
@@ -30,15 +30,7 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
```
willyborankin, line 33: Could you please, run `gradle spotlessCheck spotlessApply`?

cwperks: Looks like that didn't remove the wildcard imports. Can you adjust your IDE settings to prevent wildcard imports?

willyborankin: True `spotlessCheck spotlessApply` does not do it.

Mehdi-Bendriss: Done

willyborankin:
LGTM. One last thing you need to sign off your commits using git: 
`git commit --signoff `

Mehdi-Bendriss:
@willyborankin I signed all commits, the DCO step still fails. Any idea?

cwperks:
> @willyborankin I signed off all commits, the DCO step still fails. Any idea?

I manually set it to pass. All of the commits are signed.

cwperks (approved):

willyborankin (approved):

Issue #4613, opened 2024-07-31
[FEATURE] Support multiple keys in JWT configuration 


stephen-crawford:
**Is your feature request related to a problem?**
Currently, OpenSearch only supports a single key for JWT verification at any time. When you configure JWTs, you specify the key and that is the only key which your requests can be verified against. This leads to downtime for the cluster when you want to rotate your keys. Because updating the security configuration is not an instantaneous process, you will experience a period of time when no valid key is available for you to send JWT requests with. 

**What solution would you like?**
It would be great if multiple JWT keys could be stored at a time. This means that you could rotate a key without losing access to the cluster. Since there would be multiple keys when you need to rotate one key you can use the other key to send your requests. This avoids the period of time where there is not a valid key for your requests to be auth'd against. 

**What alternatives have you considered?**
Leaving things as is is always an option but fails to resolve the downtime problem. Figuring out a way to make the security configuration updates instantaneous would also solve this problem but that is a larger change and one that is far less likely to be completed.

cwperks:
[Triage] Thank you for this feature request @stephen-crawford. This sounds like a good feature to have. Marking as triaged.

PR #4632, opened 2024-08-09
Allow multiple signing keys to be provided

stephen-crawford:
### Description
This change is an enhancement which allows users to provide multiple signing keys when configuring JWT options. 
Previously, users could only specify a single signing key when configuring their JWT options. As a result, if a user wanted to rotate their key, they would experience some amount of downtime since the cluster would have to reinitialize the update security configuration. This change does not stop the required bouncing, but because you can now provide multiple keys, you can provide backup(s) key(s) which means even when the cluster bounces you have the ability to send JWT requests with the backup key(s). 

To specify the multiple keys, you just need to use the same method as before but separate the keys with a comma i.e. <key 1>, <key 2>, etc....

### Issues Resolved
This PR addresses: https://github.com/opensearch-project/security/issues/4613

No

### Testing
I added several new tests to the existing HTTPJWTAuthenticatorTest class.

### Check List
- [ ] ~New Roles/Permissions have a corresponding security dashboards plugin PR~

### DIFF
--- a/src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
+++ b/src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
@@ -14,6 +14,8 @@ package com.amazon.dlic.auth.http.jwt;
 import java.nio.file.Path;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
@@ -54,7 +56,7 @@ public class HTTPJwtAuthenticator implements HTTPAuthenticator {
     private static final Pattern BASIC = Pattern.compile("^\\s*Basic\\s.*", Pattern.CASE_INSENSITIVE);
     private static final String BEARER = "bearer ";
 
-    private final JwtParser jwtParser;
+    private final List<JwtParser> jwtParsers = new ArrayList<>();
     private final String jwtHeaderName;
     private final boolean isDefaultAuthHeader;
     private final String jwtUrlParameter;
@@ -68,6 +70,7 @@ public class HTTPJwtAuthenticator implements HTTPAuthenticator {
         super();
 
         String signingKey = settings.get("signing_key");
+        List<String> signingKeysSplit = new ArrayList<>();
         jwtUrlParameter = settings.get("jwt_url_parameter");
         jwtHeaderName = settings.get("jwt_header", AUTHORIZATION);
         isDefaultAuthHeader = AUTHORIZATION.equalsIgnoreCase(jwtHeaderName);
@@ -83,19 +86,31 @@ public class HTTPJwtAuthenticator implements HTTPAuthenticator {
             );
         }
 
-        final JwtParserBuilder jwtParserBuilder = KeyUtils.createJwtParserBuilderFromSigningKey(signingKey, log);
-        if (jwtParserBuilder == null) {
-            jwtParser = null;
+        boolean multipleKeys = signingKey != null && signingKey.contains(",");
+        if (multipleKeys) {
+            String[] keyArray = signingKey.replace(",\\s+", ",").split(","); // Remove spaces
+            signingKeysSplit.addAll(Arrays.asList(keyArray));
         } else {
-            if (requireIssuer != null) {
-                jwtParserBuilder.requireIssuer(requireIssuer);
-            }
+            signingKeysSplit.add(signingKey);
+        }
 
-            final SecurityManager sm = System.getSecurityManager();
-            if (sm != null) {
-                sm.checkPermission(new SpecialPermission());
+        for (String key : signingKeysSplit) {
+            JwtParser jwtParser;
+            final JwtParserBuilder jwtParserBuilder = KeyUtils.createJwtParserBuilderFromSigningKey(key, log);
+            if (jwtParserBuilder == null) {
+                jwtParser = null;
+            } else {
+                if (requireIssuer != null) {
+                    jwtParserBuilder.requireIssuer(requireIssuer);
+                }
+
+                final SecurityManager sm = System.getSecurityManager();
+                if (sm != null) {
+                    sm.checkPermission(new SpecialPermission());
+                }
+                jwtParser = AccessController.doPrivileged((PrivilegedAction<JwtParser>) jwtParserBuilder::build);
             }
-            jwtParser = AccessController.doPrivileged((PrivilegedAction<JwtParser>) jwtParserBuilder::build);
+            jwtParsers.add(jwtParser);
         }
     }
 
@@ -120,7 +135,8 @@ public class HTTPJwtAuthenticator implements HTTPAuthenticator {
     }
 
     private AuthCredentials extractCredentials0(final SecurityRequest request) {
-        if (jwtParser == null) {
+
+        if (jwtParsers.isEmpty() || jwtParsers.getFirst() == null) {
             log.error("Missing Signing Key. JWT authentication will not work");
             return null;
         }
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
```diff
@@ -68,6 +70,7 @@ public HTTPJwtAuthenticator(final Settings settings, final Path configPath) {
         super();
 
         String signingKey = settings.get("signing_key");
```
cwperks, line 72: Instead of parsing these out, can this simply be changed from `settings.get("signing_key")` to `settings.getAsList("signing_key")`

If a single value is provided, that method should give back a list of one. Alternatively, you can supply a yml list of values and parsing can be avoided.

stephen-crawford: Sure, I can swap that over if you prefer.

src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java
```diff
@@ -534,6 +535,306 @@ public void testRequiredIssuerWithIncorrectAudience() {
...
+
+        KeyPair pair2 = keyGen.generateKeyPair();
+        PrivateKey priv2 = pair2.getPrivate();
+        PublicKey pub2 = pair2.getPublic();
+
+        String jwsToken1 = Jwts.builder().setSubject("Leonard McCoy").signWith(priv1, SignatureAlgorithm.RS256).compact();
+        String jwsToken2 = Jwts.builder().setSubject("Stephen Crawford").signWith(priv2, SignatureAlgorithm.RS256).compact();
+
+        Settings settings = Settings.builder()
+            .put(
```
cwperks, line 607: Does this set need to be updated to use `putList` instead of passing a command separated string?

stephen-crawford: The tests pass so I think it works either way. I just copied the pattern from the required_audience field which already used a list.

src/test/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticatorTest.java
```diff
@@ -534,6 +535,306 @@ public void testRequiredIssuerWithIncorrectAudience() {
...
+        PrivateKey priv3 = pair3.getPrivate();
+        PublicKey pub3 = pair3.getPublic();
+
+        KeyPairGenerator keyGen4 = KeyPairGenerator.getInstance("EC");
+        keyGen4.initialize(521);
+        KeyPair pair4 = keyGen4.generateKeyPair();
+        PrivateKey priv4 = pair4.getPrivate();
+        PublicKey pub4 = pair4.getPublic();
+
+        String jwsToken1 = Jwts.builder().setSubject("Stephen Crawford").signWith(priv1, SignatureAlgorithm.RS256).compact();
```
cwperks, line 723: nit: I know it may work for the test, but usernames should not contain whitespace

stephen-crawford: All the tests do this. I think it is fine.

cwperks (approved):

src/main/java/com/amazon/dlic/auth/http/jwt/HTTPJwtAuthenticator.java
```diff
@@ -67,7 +68,8 @@ public class HTTPJwtAuthenticator implements HTTPAuthenticator {
     public HTTPJwtAuthenticator(final Settings settings, final Path configPath) {
         super();
 
-        String signingKey = settings.get("signing_key");
+        List<String> signingKeys = settings.getAsList("signing_key");
```
shikharj05, line 71: Thanks for the change @stephen-crawford ! Can we add an issue to update docs how users can rotate their keys?

Also - an example would help on how multiple keys can be specified. Looks like getAsList supports multiple different ways- https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/settings/Settings.java#L426-L448

cwperks: +1, can we use the integrationTest framework to add a test for this there as well?

stephen-crawford: Sure.

src/integrationTest/java/org/opensearch/test/framework/JwtConfigBuilder.java
```diff
@@ -33,8 +34,8 @@ public JwtConfigBuilder jwtUrlParameter(String jwtUrlParameter) {
         return this;
     }
 
-    public JwtConfigBuilder signingKey(String signingKey) {
```
cwperks, line 36: nit: Keep this method as well and have this be overloaded with 2 separate implementations. If a single signingKey is passed then you can create the list in here.

cwperks (approved):

cwperks:
Thanks for creating a documentation issue as well. Do you plan to open a PR along with the issue?

DarshitChanpura (approved):

src/main/java/org/opensearch/security/util/KeyUtils.java
```diff
@@ -54,8 +54,8 @@ public JwtParserBuilder run() {
                         PublicKey key = null;
 
                         final String minimalKeyFormat = signingKey.replace("-----BEGIN PUBLIC KEY-----\n", "")
-                            .replace("-----END PUBLIC KEY-----", "");
-
+                            .replace("-----END PUBLIC KEY-----", "")
+                            .trim();
```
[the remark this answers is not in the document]
stephen-crawford: When we add a list of the keys, we want to make sure that if the keys are provided in a weird format with preceeding whitespace etc. that the reformatted ones don't get messed up.

shikharj05 (approved):

krishna-ggk (commented):
Should we add a BWC test to ensure there is no breakage since the type of config is changing from singleton to list?

cwperks:
@krishna-ggk [Settings.getAsList](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/settings/Settings.java#L415-L448) is backwards compatible with Settings.get(). The same yaml configuration for jwt can be used with both Settings.get and Settings.getAsList for the `signing_key` portion of the config:

Example jwt config:

```
jwt_auth_domain:
    description: "Authenticate via Json Web Token"
    http_enabled: false
    transport_enabled: false
    order: 0
    http_authenticator:
      type: jwt
      challenge: false
      config:
        signing_key: "base64 encoded HMAC key or public RSA/ECDSA pem key"
        jwt_header: "Authorization"
        jwt_url_parameter: null
        roles_key: null
        subject_key: null
```

One potential issue that could arise would be trying to update the JWT authc config in a mixed cluster of 2.17 and <2.17 nodes.

If the config update hit the 2.17 node and updated the jwt config then the <2.17 nodes would not know how to parse the new config.

krishna-ggk:
Thanks @cwperks for clarifying the backward compatibility aspects.

> If the config update hit the 2.17 node and updated the jwt config then the <2.17 nodes would not know how to parse the new config.

Does the config update through API generate list or CSV? If so, it could potentially cause issue by generating incompatible version even if only single value is provided. Admittedly though this should be a remote scenario.

Potentially the least riskiest approach would be to introduce a pluralized attribute (signing_keys) which comes at expense of managing deprecation of older attribute. In interest of keeping things simple, I agree current approach is good enough.

Issue #3079, opened 2023-07-31
[FEATURE] Security plugin config entries should have a `created_at` and `updated_at` timestamp


cwperks:
**Is your feature request related to a problem?**

Config entries in the security index such as users, roles and roles_mappings have no timestamps associated with them to indicate when the entry was created or when it was most recently modified. When investigating issues, its often helpful to have this contextual information to determine when a change was made to help locate more information about the change. In many frameworks like Ruby on Rails, these timestamps come as a convention to track any objects that get persisted.

Each entry in the security index (i.e. a user or a role) should have these timestamps on the entry. This entry would help a cluster admin identify when a particular user/role/role_mapping/etc. is created or last modified.

**What solution would you like?**

Add `created_at` and `updated_at` timestamps to config entries in the security index. 

#### Example
Each entry, such as this example for user has both of these values

```yml
admin:
  hash: "$2a$12$VcCDgh2NDk07JGN0rjGbM.Ad41qVR/YFJcgHp0UGns5JDymv..TOG"
  reserved: true
  backend_roles:
  - "admin"
  description: "Demo admin user"
  created_at: 2023‐08‐30T06:05:02Z
  updated_at 2023‐09‐01T08:09:03Z
```

scrawfor99:
[Triage] Hey @cwperks, could you provide some more info as to the action items required for the change? If someone goes to complete the issue, I think we will want them to know how the created_at and updated_at fields are related to the objects you are referring to. For instance, if a user has a role, and then the role is updated, doe the user also get the new `updated_at` field?

scrawfor99:
[Triage] @cwperks just following-up.

peternied:
2 cents - I think adding a setting like this would help cluster operators troubleshoot behavior that might have been due to a config change.  Without this change that kind of diagnosis is not possible

cwperks:
@scrawfor99 I think it should be only when that particular config entry was updated. i.e. if a roles_mapping is updated to map the user `craig` to role `custom_role`, then it would be the config entry corresponding to the `roles_mapping` for `custom_role` that should update its `updated_at` timestamp.

If the user `craig` changed their password then I would expect the `updated_at` timestamp to reflect when the config entry was last updated.

peternied:
@RyanL1997 FYI - I updated the description to include an example for user simaliar to what @cwperks mentioned; consider this as you create a solution

cwperks:
[...]

PR #4638, opened 2024-08-12
Implement resource permission evaluation in security 

stephen-crawford:
### Description
This change adds resource permission evaluation into the Security plugin framework. This is accomplished by adding a new class ResourcePrivilegesEvaluator and expanding the RoleV7 definition to account for the new Resource class. 

This PR is dependent on the related change in core: https://github.com/opensearch-project/OpenSearch/pull/15230

### Issues Resolved
https://github.com/opensearch-project/security/issues/4562

### Testing

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
@@ -1,0 +1,104 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.privileges;
+
+import java.util.Set;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import org.opensearch.OpenSearchSecurityException;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.core.common.transport.TransportAddress;
+import org.opensearch.security.securityconf.ConfigModel;
+import org.opensearch.security.securityconf.SecurityRoles;
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.security.user.User;
+import org.opensearch.threadpool.ThreadPool;
+
+import org.greenrobot.eventbus.Subscribe;
+
+public class ResourceAccessEvaluator {
+    protected final Logger log = LogManager.getLogger(this.getClass());
+    private final ClusterService clusterService;
+    private ThreadContext threadContext;
+    private ConfigModel configModel;
+
+    public ResourceAccessEvaluator(final ClusterService clusterService, final ThreadPool threadPool) {
+        this.clusterService = clusterService;
+        this.threadContext = threadPool.getThreadContext();
+    }
+
+    @Subscribe
+    public void onConfigModelChanged(final ConfigModel configModel) {
+        this.configModel = configModel;
+    }
+
+    SecurityRoles getSecurityRoles(final Set<String> roles) {
+        return configModel.getSecurityRoles().filter(roles);
+    }
+
+    boolean isInitialized() {
+        return configModel != null && configModel.getSecurityRoles() != null;
+    }
+
+    public PrivilegesEvaluatorResponse evaluate(final User user, final Set<String> actions) {
+        if (!isInitialized()) {
+            throw new OpenSearchSecurityException("OpenSearch Security is not initialized.");
+        }
+
+        final PrivilegesEvaluatorResponse presponse = new PrivilegesEvaluatorResponse();
+
+        final TransportAddress caller = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
+
+        final Set<String> mappedRoles = mapRoles(user, caller);
+
+        presponse.resolvedSecurityRoles.addAll(mappedRoles);
+        final SecurityRoles securityRoles = getSecurityRoles(mappedRoles);
+
+        final boolean isDebugEnabled = log.isDebugEnabled();
+        if (isDebugEnabled) {
+            log.debug("Evaluate permissions for {} on {}", user, clusterService.localNode().getName());
+            log.debug("Action: {}", actions);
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        final PrivilegesEvaluatorResponse presponse = new PrivilegesEvaluatorResponse();
+
+        final boolean isDebugEnabled = log.isDebugEnabled();
+        if (isDebugEnabled) {
+            log.debug("Evaluate permissions for {} on {}", user, clusterService.localNode().getName());
+            log.debug("Action: {}", action);
+            log.debug("Resource: {}", request.getRequestedResources());
+            log.debug("Security roles: {}", securityRoles.toString());
```
peternied, line 59: Can you merge these into a single `log.debug` statement, otherwise they can be intertwined with other messages

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
```diff
@@ -0,0 +1,87 @@
...
+        if (resourcesRequested == null || resourcesRequested.isEmpty()) {
+            presponse.allowed = true;
+            return presponse;
+        }
+        presponse.allowed = true;
+        for (String resource : resourcesRequested) {
+            if (!securityRoles.impliesResourcePermission(resource)) {
+                presponse.missingPrivileges.add(action);
+                presponse.allowed = false;
+                log.info(
```
peternied, line 72: This message would be more useful if it included the resource(s) didn't have permission to access.

src/main/java/org/opensearch/security/securityconf/impl/v7/RoleV7.java
```diff
@@ -29,6 +29,7 @@
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Date;
```
peternied, line 32: - This is a first in this class, is this a version of https://github.com/opensearch-project/security/issues/3079 that you wanted to include for the new type?

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
```diff
@@ -0,0 +1,87 @@
...
+            throw new OpenSearchSecurityException("OpenSearch Security is not initialized.");
+        }
+
+        final PrivilegesEvaluatorResponse presponse = new PrivilegesEvaluatorResponse();
+
+        final boolean isDebugEnabled = log.isDebugEnabled();
+        if (isDebugEnabled) {
+            log.debug("Evaluate permissions for {} on {}", user, clusterService.localNode().getName());
+            log.debug("Action: {}", action);
+            log.debug("Resource: {}", request.getRequestedResources());
```
peternied, line 58: Is this a parallel change? I don't think this is wired up for core, looks like compile is failing.

peternied (commented):
Good first steps, thanks for putting this out there.

cwperks:
FYI I have been exploring the idea I posted about [here](https://github.com/opensearch-project/security/issues/4500#issuecomment-2233326609) on my fork [here](https://github.com/cwperks/security/pull/26).

Ultimately, I think the resource owner needs to be empowered to determine how they want to share the resource with others without going through the admin of the cluster. 

In order for ml-commons to adopt the new scheme that security will provide, how will the [`public`, `private` and `restricted` attributes](https://github.com/opensearch-project/documentation-website/blob/main/_ml-commons-plugin/model-access-control.md#model-groups) translate to this new scheme?

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
stephen-crawford: Yeah, I need to make a change in core first. I have not yet but it is basically going to add a "getRequestedResources" method to the ActionRequest that can be overridden. This is all coming from @DarshitChanpura's design in the linked issue.

src/main/java/org/opensearch/security/securityconf/impl/v7/RoleV7.java
stephen-crawford: To be honest, I just thought it seemed like it may be a nice thing to have since we were adding the new class. I can also remove it though and don't feel strongly about its inclusion.

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
stephen-crawford: Oh duh, good point.

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
stephen-crawford: Sure no problem. Let me fix that.

stephen-crawford:
HI @cwperks, thanks for your thoughtful comments.

I tried to address your thoughts below. 

> Ultimately, I think the resource owner needs to be empowered to determine how they want to share the resource with others without going through the admin of the cluster.

I think most patterns we follow force users to rely on the cluster operator/administrator in some ways. I understand you would like to model resources as registered fields in a per-resource indices which any user can create. This is quite different than how roles work and seems better suited outside of the roles system. Perhaps, you discussed things elsewhere, but based off https://github.com/opensearch-project/security/issues/4500#issuecomment-2233595019 it seemed like this discussion was being tabled. 

> In order for ml-commons to adopt the new scheme that security will provide, how will the [public, private and restricted attributes](https://github.com/opensearch-project/documentation-website/blob/main/_ml-commons-plugin/model-access-control.md#model-groups) translate to this new scheme?

I would expect the ML-commons plugin to override ActionRequests for their plugin and add the associated resources to the new method being added (in core): 

```
/**
     * Should be overwritten by implementing plugins with the resources each actionRequest requires.
     * Bypasses resource evaluation by default.
     */
    public List<String> getResources() {
    }
```

From there, I would expect there to be a role i.e. `ml-all-access` which included the Resource with a pattern 'ml-commons:*' or whatever the resources were called. Similarly, they would need to define roles which had the appropriate analog for `private` and `restricted` as shown in the documentation. The assignment of the appropriate roles would have to be carried out by an administrator.


_________


@DarshitChanpura @cwperks It seems like there is still decisions to be made around the design of the feature. I would describe DC's intention as more of "adding resource controls to role based access controls" while Craig seems to be more interested in adding an independent access system for resources. I suggest we reach consensus on this before moving forward.

It is probably best to carry this on over at https://github.com/opensearch-project/security/issues/4500

Issue #4614, opened 2024-08-01
[FEATURE] Create APIs to support configuration of rate limiting


derek-ho:
**Is your feature request related to a problem?**
Coming from: https://github.com/opensearch-project/security-dashboards-plugin/issues/2070, there is a request to support configuration of rate limiting/auth failure listeners via UI. In order to better support this, I am proposing that the backend security plugin expose some way to modify these settings, instead of only being able to be modified via config, securityadmin.sh, or admin cert.
**What solution would you like?**
An API that would be able to create/modify/delete these settings.
**What alternatives have you considered?**
None
**Do you have any additional context?**
None

cwperks:
[Triage] @derek-ho It sounds like a good idea to have an API for this. Can you also post some detail about the API like the route, params and responses?

derek-ho:
Hi @cwperks I was thinking of something like the following (subject to change):
```
GET /authfailurelisteners/${id} - return response like:
200
{type: ip
allowed_tries: 10
time_window_seconds: 3600
 block_expiry_seconds: 600
max_blocked_clients: 100000
max_tracked_clients: 100000}
```
```
GET /authfailurelisteners - return response like:
200
[
{...}, ....
]
```

```
DELETE /authfailurelisteners/${id} - return response like:
200
```

```
POST /authfailurelisteners
{type: ip
allowed_tries: 10
time_window_seconds: 3600
 block_expiry_seconds: 600
max_blocked_clients: 100000
max_tracked_clients: 100000}

 - return response like:
201
```

```
PUT /authfailurelisteners/${id}
{type: ip
allowed_tries: 10
time_window_seconds: 3600
 block_expiry_seconds: 600
max_blocked_clients: 100000
max_tracked_clients: 100000}

 - return response like:
201
```

PR #4641, opened 2024-08-13
Add initial changes to expose an endpoint for auth failure listener get call

derek-ho:
### Description
Feature
Creates API for GET/DELETE/CREATE auth failure listeners
Auth failure listeners could only be modified by yaml

### Issues Resolved
Fix: #4614
No

No

### Testing
Added a test
### Check List


src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,204 @@
...
+    public static final String TIME_WINDOW_SECONDS_JSON_PROPERTY = "time_window_seconds";
+    public static final String BLOCK_EXPIRY_JSON_PROPERTY = "block_expiry_seconds";
+    public static final String MAX_BLOCKED_CLIENTS_JSON_PROPERTY = "max_blocked_clients";
+    public static final String MAX_TRACKED_CLIENTS_JSON_PROPERTY = "max_tracked_clients";
+
+    private static final List<Route> ROUTES = addRoutesPrefix(
+        ImmutableList.of(
+            new Route(GET, "/authfailurelisteners"),
+            new Route(DELETE, "/authfailurelisteners/{name}"),
+            new Route(PUT, "/authfailurelisteners/{name}")
```
shikharj05, line 59: nit: can we make this more user-friendly? Also, in opensearch docs, the feature is called API rate limiting, should the names match the documentation? (while I see that the config name is ``auth_failure_listeners`` ) - https://opensearch.org/docs/latest/security/configuration/api-rate-limiting/

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,204 @@
...
+                    }
+
+                    @Override
+                    public Settings settings() {
+                        return securityApiDependencies.settings();
+                    }
+
+                    @Override
+                    public Map<String, DataType> allowedKeys() {
+                        return ImmutableMap.of("test", DataType.OBJECT);
```
shikharj05, line 116: nit: test?

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,204 @@
...
+            });
+        }).error((status, toXContent) -> response(channel, status, toXContent)))
+            .override(PUT, (channel, request, client) -> loadConfiguration(getConfigType(), false, false).valid(configuration -> {
+                ConfigV7 config = (ConfigV7) configuration.getCEntry(CType.CONFIG.toLCString());
+
+                String listenerName = request.param(NAME_JSON_PROPERTY);
+
+                ObjectNode body = (ObjectNode) DefaultObjectMapper.readTree(request.content().utf8ToString());
+                SecurityJsonNode authFailureListener = new SecurityJsonNode(body);
+                String authenticationBackend = authFailureListener.get(TYPE_JSON_PROPERTY).asString().equals("ip")
```
shikharj05, line 175: nit: extract a constant for ``ip``

src/test/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiActionTest.java
```diff
@@ -0,0 +1,113 @@
...
+import static org.hamcrest.core.IsEqual.equalTo;
+import static org.hamcrest.core.StringContains.containsString;
+
+public class AuthFailureListenersApiActionTest extends AbstractRestApiUnitTest {
+
+    private static final Header ADMIN_FULL_ACCESS_USER = encodeBasicHeader("admin_all_access", "admin_all_access");
+    private static final Header USER_NO_REST_API_ACCESS = encodeBasicHeader("admin", "admin");
+
+    @Test
+    public void testForbiddenAccess() throws Exception {
```
shikharj05, line 30: nit: add a test for delete as well

src/test/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiActionTest.java
```diff
@@ -0,0 +1,113 @@
...
+        );
+        assertThat(updateAuthFailuresResponse.getBody(), updateAuthFailuresResponse.getStatusCode(), equalTo(HttpStatus.SC_OK));
+
+        // Get after put returns the test auth failure listener
+        RestHelper.HttpResponse getAuthFailuresResponseAfterPut = rh.executeGetRequest(
+            "/_plugins/_security/api/authfailurelisteners",
+            ADMIN_FULL_ACCESS_USER
+        );
+        assertThat(getAuthFailuresResponseAfterPut.getBody(), getAuthFailuresResponseAfterPut.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertThat(getAuthFailuresResponseAfterPut.getBody(), getAuthFailuresResponseAfterPut.getBody(), containsString("test"));
```
shikharj05, line 73: nit: assert the complete config

shikharj05 (commented):
Thanks for the changes, a couple of minor comments.

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
DarshitChanpura: Is the use-case to limit API calls based on this config? if so, it should be renamed to `/api-rate-limiter`

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
DarshitChanpura: This should be replaced with actual allowed keys.

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
DarshitChanpura: Types should be an enum here.

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
cwperks: Are more than one auth_failure_listener allowed? My understanding is there can be at most 1 ip-based and 1 username-based.

If its limited to one, should we not accept a url param and have separate routes for setting up ip-based and username-based? If `{name}` is an accepted value then would the acceptable values be `ip` and `username`?

If this `{name}` is used for the key in yaml/security index for the auth_failure_listener (i.e. `ip_rate_limiting` or `internal_authentication_backend_limiting` in the demo configuration), then I don't think its necessary to provide this kind of customization. If they can only configure a single `auth_failure_listener` per type then we should keep the `ip_rate_limiting` and `internal_authentication_backend_limiting` convention

derek-ho: I took a look through the code and it seems like there can be more than one per type: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java#L454. I don't know if this makes sense though, since I don't really see a scenario that you would want multiple of each type. @cwperks should we file a bug around this?

derek-ho: As to the other points in this thread, I was thinking rate limiting might be a misleading term to brand these APIs, since they are only limiting on auth failures, not on all API requests. I opened an issue to see if there is interest in the bigger feature of rate limiting all API requests (regardless of auth failure) [here](https://github.com/opensearch-project/security/issues/4652). I can change it if the majority thinks otherwise.

src/test/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiActionTest.java
derek-ho: Done

src/test/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiActionTest.java
derek-ho: Done

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
derek-ho: Added a constant for this

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
derek-ho: Done, and added a test to assert behavior

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,204 @@
...
+                ok(channel, authFailureContent(config));
+            }).error((status, toXContent) -> response(channel, status, toXContent))
+        ).override(DELETE, (channel, request, client) -> loadConfiguration(getConfigType(), false, false).valid(configuration -> {
+            ConfigV7 config = (ConfigV7) configuration.getCEntry(CType.CONFIG.toLCString());
+
+            String listenerName = request.param("name");
+
+            // Try to remove the listener by name
+            if (config.dynamic.auth_failure_listeners.getListeners().remove(listenerName) == null) {
+                badRequest(channel, "listener not found");
```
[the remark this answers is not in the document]
derek-ho: Done, thanks!

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
cwperks: This matches the same pattern for [InternalUsers API](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java#L74-L81) and [RolesMapping API](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/dlic/rest/api/RolesMappingApiAction.java#L39-L48). I think we consider different names, but I also think the name in this PR is fine for consistency.

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,217 @@
...
+                    @Override
+                    public Settings settings() {
+                        return securityApiDependencies.settings();
+                    }
+
+                    @Override
+                    public Map<String, DataType> allowedKeys() {
+                        final ImmutableMap.Builder<String, DataType> allowedKeys = ImmutableMap.builder();
+
+                        return allowedKeys.put(TYPE_JSON_PROPERTY, DataType.STRING)
```
cwperks, line 122: FYI There was a feature in 2.16 to add `ignore_hosts` to username-based auth failure listeners. We should allow accepting that in this API:

https://github.com/opensearch-project/documentation-website/pull/7859/files

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,217 @@
...
+    public static final String AUTHENTICATION_BACKEND_JSON_PROPERTY = "authentication_backend";
+    public static final String ALLOWED_TRIES_JSON_PROPERTY = "allowed_tries";
+    public static final String TIME_WINDOW_SECONDS_JSON_PROPERTY = "time_window_seconds";
+    public static final String BLOCK_EXPIRY_JSON_PROPERTY = "block_expiry_seconds";
+    public static final String MAX_BLOCKED_CLIENTS_JSON_PROPERTY = "max_blocked_clients";
+    public static final String MAX_TRACKED_CLIENTS_JSON_PROPERTY = "max_tracked_clients";
+
+    private static final List<Route> ROUTES = addRoutesPrefix(
+        ImmutableList.of(
+            new Route(GET, "/authfailurelisteners"),
```
cwperks, line 61: Do you plan to support GET with name as well?

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,217 @@
...
+                final var config = (ConfigV7) configuration.getCEntry(CType.CONFIG.toLCString());
+                ok(channel, authFailureContent(config));
+            }).error((status, toXContent) -> response(channel, status, toXContent))
+        ).override(DELETE, (channel, request, client) -> loadConfiguration(getConfigType(), false, false).valid(configuration -> {
+            ConfigV7 config = (ConfigV7) configuration.getCEntry(CType.CONFIG.toLCString());
+
+            String listenerName = request.param("name");
+
+            // Try to remove the listener by name
+            if (config.dynamic.auth_failure_listeners.getListeners().remove(listenerName) == null) {
```
cwperks, line 171: Is there a potential for an NPE here or will `config.dynamic.auth_failure_listeners` always be non-null?

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,217 @@
...
+            });
+        }).error((status, toXContent) -> response(channel, status, toXContent)))
+            .override(PUT, (channel, request, client) -> loadConfiguration(getConfigType(), false, false).valid(configuration -> {
+                ConfigV7 config = (ConfigV7) configuration.getCEntry(CType.CONFIG.toLCString());
+
+                String listenerName = request.param(NAME_JSON_PROPERTY);
+
+                ObjectNode body = (ObjectNode) DefaultObjectMapper.readTree(request.content().utf8ToString());
+                SecurityJsonNode authFailureListener = new SecurityJsonNode(body);
+                String authenticationBackend = authFailureListener.get(TYPE_JSON_PROPERTY).asString().equals(IP_TYPE)
```
cwperks, line 188: Should we separate this logic into a separate validate method to ensure that the payload contains all required values for the respective type of auth failure listener?

i.e. `authentication_backend` is not applicable to ip-based auth failure listener so should this API send BAD_REQUEST back to the client if the payload body is invalid?

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,217 @@
...
+                    || authFailureListener.get(AUTHENTICATION_BACKEND_JSON_PROPERTY).isNull()
+                        ? null
+                        : authFailureListener.get(AUTHENTICATION_BACKEND_JSON_PROPERTY).asString();
+
+                // Try to remove the listener by name
```
cwperks, line 193: nit: update this comment

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,217 @@
...
+                config.dynamic.auth_failure_listeners.getListeners()
+                    .put(
+                        listenerName,
+                        new ConfigV7.AuthFailureListener(
+                            authFailureListener.get(TYPE_JSON_PROPERTY).asString(),
+                            Optional.ofNullable(authenticationBackend),
+                            Integer.parseInt(authFailureListener.get(ALLOWED_TRIES_JSON_PROPERTY).asString()),
```
cwperks, line 200: wdyt about adding `asInt()` to SecurityJsonNode and calling parseInt within there?

src/main/java/org/opensearch/security/securityconf/impl/v7/ConfigV7.java
```diff
@@ -248,6 +249,24 @@ public AuthFailureListener(ConfigV6.AuthFailureListener v6) {
             this.max_tracked_clients = v6.max_tracked_clients;
         }
 
+        public AuthFailureListener(
```
cwperks, line 252: I wonder if it makes sense to separate out IPAuthFailureListener and UsernameAuthFailureListener. They both could inherit from an AbstractAuthFailureListener and re-use the parent's constructor for common attributes.

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
derek-ho: I figured it is not necessary, since the overall GET should return all instances anyway, and there is no additional information that can be returned if we build a different endpoint per auth failure listener, but I can do that if/when we feel we have enough functionality to justify that addition, wdyt?

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
derek-ho: When I was reading the code I thought it would always be initialized based on this [code ref](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/impl/v7/ConfigV7.java#L137). But I am not sure if there are other ways that this would be initialized, can you let me know if this is the correct understanding?

cwperks: You're right. Based on the code-path `config.dynamic.auth_failure_listeners` can never be null`. Resolving this thread.

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
cwperks: I agree. I think list is adequate for this API. In realistic scenarios, at most 2 auth_failure_listeners would be configured. Resolving this thread.

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
derek-ho: Sure did so in: [fdaba0c](https://github.com/opensearch-project/security/pull/4641/commits/fdaba0cbf43cbb307e4ed762efd93e16ef3f5d4b)

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
derek-ho: Sure did so in: [fdaba0c](https://github.com/opensearch-project/security/pull/4641/commits/fdaba0cbf43cbb307e4ed762efd93e16ef3f5d4b)

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
derek-ho: Done: [fdaba0c](https://github.com/opensearch-project/security/pull/4641/commits/fdaba0cbf43cbb307e4ed762efd93e16ef3f5d4b)

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
derek-ho: Good idea: [fdaba0c](https://github.com/opensearch-project/security/pull/4641/commits/fdaba0cbf43cbb307e4ed762efd93e16ef3f5d4b)

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    private ValidationResult<SecurityJsonNode> validateAuthFailureListener(SecurityJsonNode authFailureListener, String name) {
+        if (name == null) {
+            return ValidationResult.error(RestStatus.BAD_REQUEST, badRequestMessage("name is required"));
+        }
+        if (authFailureListener.get(TYPE_JSON_PROPERTY).isNull()) {
+            return ValidationResult.error(RestStatus.BAD_REQUEST, badRequestMessage("type is required"));
+        }
+        if (!(authFailureListener.get(TYPE_JSON_PROPERTY).asString().equals(IP_TYPE)
```
cwperks, line 243:
```
 if (!(Set.of(IP_TYPE, USERNAME_TYPE).includes(authFailureListener.get(TYPE_JSON_PROPERTY).asString())) {
    ...
 }
```

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,265 @@
...
+        requestHandlersBuilder.override(
+            GET,
+            (channel, request, client) -> loadConfiguration(getConfigType(), false, false).valid(configuration -> {
+                final var config = (ConfigV7) configuration.getCEntry(CType.CONFIG.toLCString());
+                ok(channel, authFailureContent(config));
+            }).error((status, toXContent) -> response(channel, status, toXContent))
+        ).override(DELETE, (channel, request, client) -> loadConfiguration(getConfigType(), false, false).valid(configuration -> {
+            ConfigV7 config = (ConfigV7) configuration.getCEntry(CType.CONFIG.toLCString());
+
+            String listenerName = request.param("name");
```
cwperks, line 176: nit: Replace this with the constant, similar to below

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,265 @@
...
+            return ValidationResult.error(
+                RestStatus.BAD_REQUEST,
+                badRequestMessage("ip auth failure listeners must have 'internal' authentication backend")
+            );
+        }
+        if (authFailureListener.get(TYPE_JSON_PROPERTY).asString().equals(IP_TYPE)
+            && !authFailureListener.get(AUTHENTICATION_BACKEND_JSON_PROPERTY).isNull()) {
+            return ValidationResult.error(
+                RestStatus.BAD_REQUEST,
+                badRequestMessage("username auth failure listeners should not have an authentication backend")
```
cwperks, line 259: I think this is wrong. IP Auth failure listener should not have auth backend.

I'm actually a little surprised that authentication_backend is a required field at all. 

Relevant section of code: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java#L403-L414

src/test/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiActionTest.java
```diff
@@ -0,0 +1,153 @@
...
+        assertThat(getSettingResponse.getBody(), updateSettingResponse.getStatusCode(), equalTo(HttpStatus.SC_FORBIDDEN));
+        RestHelper.HttpResponse deleteSettingResponse = rh.executeDeleteRequest(
+            "/_plugins/_security/api/authfailurelisteners/test",
+            USER_NO_REST_API_ACCESS
+        );
+        assertThat(getSettingResponse.getBody(), updateSettingResponse.getStatusCode(), equalTo(HttpStatus.SC_FORBIDDEN));
+    }
+
+    @Test
+    public void testFullAccess() throws Exception {
```
cwperks, line 53: Can we add a test, that has a minimal payload? i.e. just include `{\"type\":\"ip\"}` and verify that it will create an auth failure listener with the default values?

Default values are here: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/impl/v7/ConfigV7.java#L230-L234

Will the integer parsing allow negative values? Can we prohibit negative ints?

derek-ho: Done in: [2a8c237](https://github.com/opensearch-project/security/pull/4641/commits/2a8c237f25239167cf854bc50143f9f7f85a4af9). I don't think we need to handle negative values/that is overkill since they could always just put negative values in the yaml right? We can bring that up as a separate issue to validate that doesn't break anything but it feels out of scope for this PR.

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
derek-ho: Yea I originally had the logic reversed, but never changed the message. I fixed it in: [2a8c237](https://github.com/opensearch-project/security/pull/4641/commits/2a8c237f25239167cf854bc50143f9f7f85a4af9)

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
derek-ho: Done in: [2a8c237](https://github.com/opensearch-project/security/pull/4641/commits/2a8c237f25239167cf854bc50143f9f7f85a4af9)

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
derek-ho: This is much cleaner! Fixed in: [2a8c237](https://github.com/opensearch-project/security/pull/4641/commits/2a8c237f25239167cf854bc50143f9f7f85a4af9)

src/main/java/org/opensearch/security/dlic/rest/api/AuthFailureListenersApiAction.java
```diff
@@ -0,0 +1,288 @@
...
+
+        int timeWindowSeconds = authFailureListener.get(TIME_WINDOW_SECONDS_JSON_PROPERTY).isNull()
+            ? TIME_WINDOW_SECONDS_DEFAULT
+            : authFailureListener.get(TIME_WINDOW_SECONDS_JSON_PROPERTY).asInt();
+
+        int blockExpirySeconds = authFailureListener.get(BLOCK_EXPIRY_JSON_PROPERTY).isNull()
+            ? BLOCK_EXPIRY_SECONDS_DEFAULT
+            : authFailureListener.get(BLOCK_EXPIRY_JSON_PROPERTY).asInt();
+
+        int maxBlockedClients = authFailureListener.get(MAX_BLOCKED_CLIENTS_JSON_PROPERTY).isNull()
```
cwperks, line 239: nit: is it possible to pass the default into the `asInt` call and move the null handling within?

i.e. reduce this call to `int maxBlockedClients = authFailureListener.get(MAX_BLOCKED_CLIENTS_JSON_PROPERTY).asInt(MAX_BLOCKED_CLIENTS_DEFAULT)`

derek-ho: Good suggestion, it cleaned up the code a good amount. Did so in this commit: [d565320](https://github.com/opensearch-project/security/pull/4641/commits/d565320611db6e052685d837957b310256aa90db)

cwperks (approved):

DarshitChanpura (approved):

Issue #2040, opened 2022-08-22
[FEATURE] Support UserInfo Endpoint for OpenID


Martin-Kemp:
**Is your feature request related to a problem?**
We're currently trying to use OpenID but can't use it for fine grained authorization because Opensearch does not support the UserInfo Endpoint.

**What solution would you like?**
Support UserInfo Endpoint as defined by OpenID: https://openid.net/specs/openid-connect-core-1_0.html#UserInfo
This is supported by Elasticsearch through xpack: https://www.elastic.co/guide/en/elasticsearch/reference/current/oidc-guide.html

**What alternatives have you considered?**
I'm currently trying to get around this by using LDAP attributes but allow-listing only some attributes does not seem to work either: https://github.com/opensearch-project/security/issues/2032

**Do you have any additional context?**

DarshitChanpura:
Thank you for filing this request.

Martin-Kemp:
Not sure if this is frowned upon, but here's how x-pack does it: https://github.com/elastic/elasticsearch/blob/53a2aa94f9f46aded071c48b696bbb514459ad36/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java#L256

Obviously I'm not suggesting a copy-paste but might be worth looking at for a reference.

hoilc:
😂 We are faced with the same issue. ID token is really not enough.

geckiss:
Hello, any update on this? We would love to see this feature implemented

peternied:
Thanks for the interest @geckiss we'd welcome a pull request to support this feature.

PR #4649, opened 2024-08-15
Add support for OpenId Connect userinfo_endpoint 

stephen-crawford:
### Description
This change reworks the OpenID Connect authentication process to allow for the configuration of the new setting `userinfo_endpoint`. It also allows the configuration of the two associated settings `client_id` and `issuer_id_url`, which are used when validating the response provided by the OIDC provider back to the client. 

The userinfo_endpoint is a configuration of OIDC originally detailed in the RFC: https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

It functions by creating an automated callback between the Client and the OIDC provider. In the RFC, these are referred to as the Relying Party (RP) and OpenID Provider (OP). In short while the existing OIDC support works by causing the OP to simply send a complete JWT to be parsed by OpenSearch, the new support works by returning that JWT back to the OP while querying the `userinfo_endpoint` configured during registration. 

In turn, the OP will return either a serialized JSON response or a new JWT. This response is then validated based on the details outlined in the RFC, before the `sub` and `roles` fields of either the JSON or JWT are formed into a new set of AuthCredentials. 

Here is an example of how this works 

OP --/app/home# Bearer <sub: stephen>--> RP
RP --GET /<userinfo_endpoint> Bearer <sub: stephen--> OP
OP --OK <sub: stephen, roles: admin, iss: ...> --> RP 

### Issues Resolved
This change should resolve: https://github.com/opensearch-project/security/issues/2040

### Testing
I have updated the existing OIDC tests and added new tests which cover the JWT and JSON flows for the userinfo behavior. 

### Check List


stephen-crawford:
Opening this up for review while I go ahead and try to test the use case. Going to use OpenSearch with the custom installation and Key\cloack as the OIDC provider.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticator.java
```diff
@@ -1,73 +0,0 @@
-/*
```
cwperks, line 1: @stephen-crawford Is the user-info endpoint only used to extract the backend roles from the external IdP or does it have another use as well? (i.e. custom attributes)

Would it make sense to have a protected method in the AbstractHttpJwtAuthenticator to `extractRoles` and have this class override the implementation?

The class renaming is making it harder to determine what the diffs are in this PR.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
```diff
@@ -0,0 +1,328 @@
...
+import static com.amazon.dlic.auth.http.jwt.keybyoidc.OpenIdConstants.SUB_CLAIM;
+
+public class HTTPOpenIdAuthenticator implements HTTPAuthenticator {
+
+    private final static Logger log = LogManager.getLogger(HTTPOpenIdAuthenticator.class);
+    private final Settings settings;
+    private final Path configPath;
+    private final int requestTimeoutMs = 10000;
+    private final SettingsBasedSSLConfigurator.SSLConfig sslConfig;
+    private final String userinfo_endpoint;
```
cwperks, line 63: nit: Keep casing consistent. wdyt of `userInfoEndpoint`?

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
```diff
@@ -0,0 +1,328 @@
...
+                if (openIdJwtAuthenticator == null) {
+                    openIdJwtAuthenticator = new HTTPJwtKeyByOpenIdConnectAuthenticator(settings, configPath);
+                }
+            }
+        }
+        return openIdJwtAuthenticator;
+    }
+
+    public String getType() {
+        return null;
```
cwperks, line 86: Type should have a value. I think this should remain `jwt-key-by-oidc` with the different being how roles are extracted.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
stephen-crawford: I took this convention because it is a read settings value. I am happy to swap it over.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticator.java
cwperks: Got it, I see an example positive response here: https://docs.aws.amazon.com/cognito/latest/developerguide/userinfo-endpoint.html#get-userinfo-response-sample

So userinfo endpoint needs to be called before extractRoles and extractSubject.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
stephen-crawford: I am unclear on the purpose of this getType outside of it being an inherited requirement. I don't think it would be `jwt-key` etc. because it is possible for the userinfo_endpoint to return a json as opposed to the a jwt. I was not sure what we wanted to put here so I left it as null. I can swap it if have an actual convention to follow, but again, I am unsure on the purpose of this method.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/OpenIdConstants.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package com.amazon.dlic.auth.http.jwt.keybyoidc;
+
+public class OpenIdConstants {
+
+    public static final String APPLICATION_JSON = "application/json";
```
cwperks, line 16: Use `import static org.apache.hc.core5.http.ContentType.APPLICATION_JSON;`

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/OpenIdConstants.java
```diff
@@ -0,0 +1,23 @@
...
+    public static final String APPLICATION_JWT = "application/jwt";
+    public static final String AUTHORIZATION_HEADER = "Authorization";
```
cwperks, line 18: Use `org.apache.http.HttpHeaders.AUTHORIZATION`

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
```diff
@@ -0,0 +1,328 @@
...
+    public HTTPOpenIdAuthenticator(Settings settings, Path configPath) throws Exception {
+        this.settings = settings;
+        this.configPath = configPath;
+        this.sslConfig = getSSLConfig(settings, configPath);
+        userinfo_endpoint = settings.get("userinfo_endpoint");
+        openIdJwtAuthenticator = new HTTPJwtKeyByOpenIdConnectAuthenticator(settings, configPath);
+    }
+
+    public HTTPJwtKeyByOpenIdConnectAuthenticator getOpenIdJwtAuthenticator() {
+        if (openIdJwtAuthenticator == null) {
```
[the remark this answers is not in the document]
stephen-crawford: I made this for the case where we want to access the wrapped jwtAuthenticator. If the overall Authenticator is not initialized, we would want to be able to access the internal authenticator for testing I believe. The synchronized is meant to prevent the creation of multiple authenticators when `getOpenIdJwtAuthenticator` is called multiple times. We don't want to create several instances of the `jwtOpenIdAuthenticator` so this makes sure we only ever make one. 

If you think there is a better way of doing this let me know.

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticatorTests.java
```diff
@@ -0,0 +1,693 @@
...
+
+        Assert.assertNotNull(creds);
+        assertThat(creds.getUsername(), is(MCCOY_SUBJECT));
+        assertThat(creds.getAttributes().get("attr.jwt.aud"), is(List.of(TestJwts.TEST_AUDIENCE).toString()));
+        assertThat(creds.getBackendRoles().size(), is(0));
+        assertThat(creds.getAttributes().size(), is(4));
+    }
+
+    @Test
+    public void userinfoEndpointReturnsJwtWithAllRequirementsTest() throws Exception {
```
stephen-crawford, line 426: This point starts the new tests.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
```diff
@@ -0,0 +1,328 @@
...
+                String contentType = httpEntity.getContentType();
+                if (!contentType.contains(APPLICATION_JSON) && !contentType.contains(APPLICATION_JWT)) {
+                    throw new AuthenticatorUnavailableException(
+                        "Error while getting " + this.userinfo_endpoint + ": Invalid content type in response"
+                    );
+                }
+
+                String userinfoContent;
+
+                try (
```
cwperks, line 180: Would it be possible to use nimbus' [`UserInfoResponse.parse(response)`](https://www.javadoc.io/doc/com.nimbusds/oauth2-oidc-sdk/5.22/com/nimbusds/openid/connect/sdk/UserInfoResponse.html)?

The KeySetRetriever use's nimbuses similarly like this: https://github.com/opensearch-project/security/blob/main/src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/KeySetRetriever.java#L92-L97

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
```diff
@@ -0,0 +1,328 @@
...
+
+            return selfRefreshingKeySet;
+        }
+
+        private JWTClaimsSet getJwtClaimsSet(SecurityRequest request) throws OpenSearchSecurityException {
+            String parsedToken = super.getJwtTokenString(request);
+            return getJwtClaimsSetFromInfoContent(parsedToken);
+        }
+
+        private JWTClaimsSet getJwtClaimsSetFromInfoContent(String userInfoContent) throws OpenSearchSecurityException {
```
stephen-crawford, line 306: This is new--how I split it out from the extract method--but the rest of the nested class was previously implemented.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticator.java
stephen-crawford: Yeah, this code is pretty complicated which is why I tried to add a lot of tests. The issue is that the existing OIDC logic was not implemented with the userinfo endpoint in mind. It is tightly coupled to the JWTAuthenticator (and the abstract version), but the userinfo can be a jwt or json which complicates things.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/OpenIdConstants.java
stephen-crawford: Goo idea

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
stephen-crawford: Hmm let me look. I just googled how to read the entity content properly but if we already have something that does that, then that would be great.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPJwtKeyByOpenIdConnectAuthenticator.java
stephen-crawford: @cwperks I added a comment on the test file and the renamed file to help make clear what is new.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
```diff
@@ -0,0 +1,328 @@
...
+                || claims.getAudience().toString().isBlank()
+                || !claims.getAudience().contains(settings.get(CLIENT_ID))) {
+                missing = missing.concat("aud");
+            }
+        }
+
+        return missing;
+    }
+
+    private final class HTTPJwtKeyByOpenIdConnectAuthenticator extends AbstractHTTPJwtAuthenticator {
```
stephen-crawford, line 260: All above is new. This internal class previously existed as the separate HttpJwtKeyByOpenIdConnectAuthenticator. I nested the class since the userinfo logic directly relies on it and the previous naming was out of convention.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
stephen-crawford: I swapped to "oidc" since saml is just "saml"

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
stephen-crawford: So, I took a look and I think trying to use the package is more hassle than its worth. For starters, we need to add a dependency on https://mvnrepository.com/artifact/com.nimbusds/oauth2-oidc-sdk. But more notably, trying to use the Parse logic will require resolving a translation between the CloseableHttpResponse we current use and the generic HttpResponse that the built in method expects. 

It seems like we are probably better off leaving things as they are for now.

src/test/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticatorTests.java
DarshitChanpura: a lot of code is repetitive in the test below. Do you think we can extract it to a single method, say assertValidCredentials and assertInvalidCredentials?

DarshitChanpura (commented):
Thank you @stephen-crawford for adding this support. Changes look close. Left a few comments. Could you look into codecov annotations about missed lines in test coverage?

stephen-crawford:
Hi @DarshitChanpura, thanks for taking a look. I will definitely address your comments before merging. Right now, I am having to backport these changes to the 2.x line to check the stability and functionality of the feature so it will be tomorrow or next week before I update--but I am going to.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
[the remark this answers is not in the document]
stephen-crawford: I am going to remove this since apparently this pattern is not recommended in modern coding practices ...

stephen-crawford:
@cwperks I swapped over to using Nimbus for sending the requests like you suggested. I ran into some problems with the verification of the fields since it seems like the Nimbus validation is based off of the server response not the actual fields equaling the settings. I also addressed @DarshitChanpura's comments while takng care of this swap and revising the tests. 

Going to go back to manually testing things now.

nibix:
@stephen-crawford Cool feature! I am wondering about one thing: Is the userinfo endpoint then queried for each incoming REST request? If so, this might add quite a bit of latency on these requests.

stephen-crawford:
Hi @nibix, to answer your question, yes. The server will need to be queried for each request. There will be some additional latency when using this feature but presumably anyone who wants to use it would be aware of this trade off. I don't think there is an alternative to this approach since I don't believe our OIDC support caches the credentials for a session. This is a possible improvement down the line but from what I can tell, we expect each request to include valid authorization info. If we were to add caching of some kind, I would expect a new setting i.e. `userinfo_ttl` which would let you configure the time length that the userinfo should remain valid for. Then a temporary credential could be provided upon the first credentials fetch and returned back to OpenSearch upon all subsequent requests inside that time period.

This would require a fair amount of additional work I suspect.

nibix:
@stephen-crawford There's already a user cache in `BackendRegistry`: 

https://github.com/opensearch-project/security/blob/359272ef4b3811ec794a1c8ac19e7dfcf42906f4/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L580-L585

Currently, the user cache is only used when there is a `AuthenticationBackend` which is non no-op or if there are `AuthorizationBackend` instances. Following the  comment in the code, it looks like that the original assumption was that classes implementing `HTTPAuthenticator` work only on the request object itself, i.e., not communicating with further backends. That would be the motivation to disable the cache in these cases - as it happens in the code. 

To a certain extend, one could argue that the `user_info` endpoint can be also considered as an authentication backend. If it would be used as such, the user cache would automatically get active - thus reducing the number of necessary requests to the `user_info` endpoint.

stephen-crawford:
Hi @nibix, my understanding from the comment in the linked code is that the cache is not active for OIDC. It reads 

```
 // noop backend configured and no authorizers 
 // that mean authc and authz was completely done via HTTP (like JWT or PKI) 
 if (authBackend.getClass() == NoOpAuthenticationBackend.class && authorizers.isEmpty()) { 
     // no cache 
     return authBackend.authenticate(ac); 
 } 
  
```

So since OIDC is based on the JWT auth I interpreted this as meaning there would not be any caching here. Perhaps, I am mistaken.

Either way, it sounds like you have reservations about making this change? Is there something you would like to see done to this PR to move forward? I still need to setup the test systems to actually confirm it works as expected but the code should mostly be complete.

cwperks:
A round trip per request is a lot of overhead and keeping the user cache in sync across nodes would also be a challenging problem.

Ideally, there is a token that has claims about the subject and roles that can be extracted after initial authentication so that only a round trip on creation of a session is necessary.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
```diff
@@ -0,0 +1,289 @@
...
+        } catch (IOException | URISyntaxException | com.nimbusds.oauth2.sdk.ParseException e) {
+            throw new AuthenticatorUnavailableException("Error while getting " + this.userInfoEndpoint + ": " + e, e);
+        }
+    }
+
+    private String validateResponseClaims(JWTClaimsSet claims, String id, boolean isSigned) {
+
+        String missing = "";
+
+        if (claims.getClaim(SUB_CLAIM) == null || claims.getClaim(SUB_CLAIM).toString().isBlank() || !claims.getClaim("sub").equals(id)) {
```
cwperks, line 203: This should also support the `subject_key` and `roles_key` settings: https://opensearch.org/docs/latest/security/authentication-backends/openid-connect/#configure-openid-connect-integration

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
```diff
@@ -0,0 +1,289 @@
...
+                throw new RuntimeException(e);
+            }
+        } catch (IOException | URISyntaxException | com.nimbusds.oauth2.sdk.ParseException e) {
+            throw new AuthenticatorUnavailableException("Error while getting " + this.userInfoEndpoint + ": " + e, e);
+        }
+    }
+
+    private String validateResponseClaims(JWTClaimsSet claims, String id, boolean isSigned) {
+
+        String missing = "";
```
cwperks, line 201: StringBuilder? Or maybe `Set<String> missingRequiredClaims`?

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
```diff
@@ -0,0 +1,289 @@
...
+     * @return AuthCredentials formed through querying the userinfo_endpoint
+     * @throws OpenSearchSecurityException On failure to extract credentials from the request
+     */
+    public AuthCredentials extractCredentials0(SecurityRequest request, ThreadContext context) throws OpenSearchSecurityException {
+
+        try {
+
+            URI userInfoEndpointURI = new URI(this.userInfoEndpoint);
+
+            String bearerHeader = request.getHeaders().get(AUTHORIZATION).getFirst();
```
cwperks, line 118: Can this be extract into a helper method where unit tests can be added?

cwperks: There's similar logic in [AbstractHTTPJwtAuthenticator](https://github.com/opensearch-project/security/blob/main/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java#L177-L180). Can we write this logic once and test it so that its re-usable?

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
```diff
@@ -0,0 +1,289 @@
...
+            if (!StringUtils.isBlank(bearerHeader)) {
+                if (bearerHeader.contains("Bearer ")) {
+                    bearerHeader = bearerHeader.substring(7);
+                }
+            }
+
+            String finalBearerHeader = bearerHeader;
+
+            AccessToken accessToken = new AccessToken(AccessTokenType.BEARER, finalBearerHeader) {
```
cwperks, line 127: FYI Nimbus has a class called [BearerAccessToken](https://www.javadoc.io/doc/com.nimbusds/oauth2-oidc-sdk/5.9/com/nimbusds/oauth2/sdk/token/BearerAccessToken.html), would that be usable here?

cwperks: Looks like this is the source code: https://bitbucket.org/connect2id/oauth-2.0-sdk-with-openid-connect-extensions/src/c7736eb688bf0a81714374822d4e7b5232222eea/src/main/java/com/nimbusds/oauth2/sdk/token/BearerAccessToken.java#lines-208:211

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
```diff
@@ -0,0 +1,289 @@
...
+
+                UserInfoResponse userInfoResponse = UserInfoResponse.parse(httpResponse);
+
+                if (!userInfoResponse.indicatesSuccess()) {
+                    throw new AuthenticatorUnavailableException(
+                        "Error while getting " + this.userInfoEndpoint + ": " + userInfoResponse.toErrorResponse()
+                    );
+                }
+
+                String contentType = String.valueOf(httpResponse.getHeaderValues("content-type"));
```
cwperks, line 155: Extract out the UserInfoSuccessResponse here and then call on UserInfoSuccessResponse.getEntityContentType to get the content type from the request.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
stephen-crawford: Sure.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
stephen-crawford: Will do

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
stephen-crawford: Oh I didn't realize that.

src/main/java/com/amazon/dlic/auth/http/jwt/keybyoidc/HTTPOpenIdAuthenticator.java
```diff
@@ -0,0 +1,289 @@
...
+        return "oidc";
+    }
+
+    private static SettingsBasedSSLConfigurator.SSLConfig getSSLConfig(Settings settings, Path configPath) throws Exception {
+        return new SettingsBasedSSLConfigurator(settings, configPath, "openid_connect_idp").buildSSLConfig();
+    }
+
+    @Override
+    public AuthCredentials extractCredentials(SecurityRequest request, ThreadContext context) throws OpenSearchSecurityException {
+        if (this.userInfoEndpoint != null && !this.userInfoEndpoint.isBlank()) {
```
cwperks, line 83: wdyt about moving the logic that calls the userinfo endpoint into a new method?

In AbstractHttpJwtAuthenticator would it make sense to add a new method called:

```
public JwtClaimsSet(String jwt) { 
   return jwt.getJWTClaimsSet(); // default implementation
}
```

That way this line can be replaced: https://github.com/opensearch-project/security/blob/main/src/main/java/com/amazon/dlic/auth/http/jwt/AbstractHTTPJwtAuthenticator.java#L131

In the OpenID authenticator can we then subclass that and either return `jwt.getJWTClaimsSet()` (if userinfo endpoint is not configured) or if userinfo is configured then follow the logic that you have below.

stephen-crawford: I am not sure what the benefit of further moving things would be. We already have separation of the userinfo vs. non-userinfo logic inside this handler. If you really prefer it that way, I can move things around, but don't see how 


```
   @Override
    public AuthCredentials extractCredentials(SecurityRequest request, ThreadContext context) throws OpenSearchSecurityException {
        return new AuthCredentials(this.getJwtClaims())
    }
     
    public JwtClaimsSet getJwtClaims() {
     if (this.userInfoEndpoint != null && !this.userInfoEndpoint.isBlank()) {
            return extractCredentials0(request, context);
        }
        return (this.openIdJwtAuthenticator.extractCredentials(request, context));
   }
```
is better than 
```
   @Override
    public AuthCredentials extractCredentials(SecurityRequest request, ThreadContext context) throws OpenSearchSecurityException {
        if (this.userInfoEndpoint != null && !this.userInfoEndpoint.isBlank()) {
            return extractCredentials0(request, context);
        }
        return (this.openIdJwtAuthenticator.extractCredentials(request, context));
    }
```


These are functionally identical in my opinion and make the code harder to follow I feel. You can decide what you prefer though.

cwperks: This class would not need to override `extractCredentials`, it would completely re-use `extractCredentials` from AbstractHTTPJwtAuthenticator, but override the implementation of `getJWTClaimsSet`.

DarshitChanpura:
@cwperks @derek-ho How should we proceed with this one?

DarshitChanpura:
Closing this PR as it is stale.

PR #4744, opened 2024-09-21
Fixed use of rolesMappingConfiguration in InternalUsersApiActionValidationTest

nibix:
### Description

While working on https://github.com/opensearch-project/security/pull/4546, I encountered weird behavior in InternalUsersApiActionValidationTest. Closer analysis revealed that the behavior came from the fact that the test erroneously parsed a roles mapping JSON config as `CType.ROLES` and not as `CType.ROLESMAPPING`.

 I have now switched the config type to `CType.ROLESMAPPING`. 

This has a couple of consequences:

- The roles mapping config type does not have a `static` property, so it cannot be marked as such.
- This causes the behavior of one test to flip: `InternalUsersApiActionValidationTest.validateSecurityRolesWithImmutableRolesMappingConfig`, last assert.

Side note: To be honest, I am a bit confused by the test `validateSecurityRolesWithImmutableRolesMappingConfig` - why should have a user no reserved roles or static roles assigned? Despite the change, the test demonstrates that reserved roles cannot be assigned.

* Category: Test fix
* Why these changes are required? - As the changes in !4546 do some more strict validations, the wrong assignment of the wrong config type leads to issues.
* What is the old behavior before changes and new behavior after changes? - as this is a test fix, no behavior changes.

### Check List

### DIFF
--- a/src/test/java/org/opensearch/security/dlic/rest/api/InternalUsersApiActionValidationTest.java
+++ b/src/test/java/org/opensearch/security/dlic/rest/api/InternalUsersApiActionValidationTest.java
@@ -72,13 +72,13 @@ public class InternalUsersApiActionValidationTest extends AbstractApiActionValid
         config.set("all_access", objectMapper.createObjectNode());
         config.set("regular_role", objectMapper.createObjectNode());
 
-        config.set("some_role_with_static_mapping", objectMapper.createObjectNode().put("static", true));
+        config.set("some_role_with_static_mapping", objectMapper.createObjectNode());
         config.set("some_role_with_reserved_mapping", objectMapper.createObjectNode().put("reserved", true));
         config.set("some_role_with_hidden_mapping", objectMapper.createObjectNode().put("hidden", true));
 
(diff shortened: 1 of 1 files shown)


cwperks:
My understanding is that reserved roles can be assigned, but that the test is pertaining to `roles_mappings`. There is one instance in the demo configuration of a `reserved` role mapping and that is [kibana_server](https://github.com/opensearch-project/security/blob/main/config/roles_mapping.yml#L47). The test is asserting that the `reserved` roles_mapping cannot be modified. 

If its not possible to create a `static` roles mapping, then I think the test scenario can be removed.

nibix:
@cwperks 

> If its not possible to create a `static` roles mapping, then I think the test scenario can be removed.

Good point, removing the part!

nibix:
@cwperks 

> My understanding is that reserved roles can be assigned, but that the test is pertaining to `roles_mappings`

Yes, it operate mostly on role mapping docs, but it is part of the internal users API. That's what confuses me a bit.

cwperks (approved):

willyborankin (approved):

PR #4746, opened 2024-09-22
[DRAFT] [Feature] Introduces resource sharing and access-control

DarshitChanpura:
companion PR: https://github.com/opensearch-project/OpenSearch/pull/16030

### Description
Introduces a new authorization mechanism to control access to resources defined by plugins. 

This PR also introduces a sample resource plugin to show the APIs in action.

* Category : New feature
  * At present, plugins have implemented in-house authorization mechanisms for a lack of centralized framework. This PR introduces a centralized way to offload resource permissions check to security plugin. Plugins will leverage the java APIs introduced in core.
 

### Issues Resolved
- Related to https://github.com/opensearch-project/security/issues/4500

### Testing
- automated and manual testing

### Check List


sample-resource-plugin/src/main/java/org/opensearch/security/sample/SampleResourcePlugin.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    @Override
+    public Collection<Class<? extends LifecycleComponent>> getGuiceServiceClasses() {
+        final List<Class<? extends LifecycleComponent>> services = new ArrayList<>(1);
+        services.add(GuiceHolder.class);
+        return services;
+    }
+
+    public static class GuiceHolder implements LifecycleComponent {
```
nibix, line 131: This `GuiceHolder` pattern from the security plugin is IMHO only a hack to get access to dependencies which are available via dependency injection, but not via the `createComponents()` API in the `Plugin` interface. The un-hackish way to give plugins access to `ResourceService` would be to add it to the param list of `createComponents()`.

DarshitChanpura:
Resource-sharing index gets created successfully upon first cluster start:
```
[2024-10-04T14:09:04,570][INFO ][o.o.p.PluginsService     ] [80a9970ace0f] PluginService:onIndexModule index:[.opensearch_resource_sharing/G4yJ1aVmRRaQStn45aWdDg]
[2024-10-04T14:09:04,604][INFO ][o.o.c.m.MetadataCreateIndexService] [80a9970ace0f] [.opensearch_resource_sharing] creating index, cause [api], templates [], shards [1]/[1]
[2024-10-04T14:09:04,607][INFO ][o.o.c.r.a.AllocationService] [80a9970ace0f] updating number_of_replicas to [0] for indices [.opensearch_resource_sharing]
[2024-10-04T14:09:04,608][WARN ][o.o.c.r.a.AllocationService] [80a9970ace0f] Falling back to single shard assignment since batch mode disable or multiple custom allocators set
[2024-10-04T14:09:04,638][INFO ][o.o.p.PluginsService     ] [80a9970ace0f] PluginService:onIndexModule index:[.opensearch_resource_sharing/G4yJ1aVmRRaQStn45aWdDg]
[2024-10-04T14:09:04,748][INFO ][o.o.c.r.a.AllocationService] [80a9970ace0f] Cluster health status changed from [YELLOW] to [GREEN] (reason: [shards started [[.opensearch_resource_sharing][0]]]).
[2024-10-04T14:09:04,768][INFO ][o.o.s.r.ResourceSharingIndexHandler] [80a9970ace0f] Resource sharing index .opensearch_resource_sharing created.
```

and subsequent starts show that the index is present:
```
[2024-10-04T14:11:08,755][INFO ][o.o.s.r.ResourceSharingIndexHandler] [80a9970ace0f] Index .opensearch_resource_sharing already exists.
```

.gitignore
```diff
@@ -43,7 +43,3 @@ out/
 build/
 gradle-build/
 .gradle/
-
-# nodejs
-node_modules/
-package-lock.json
```
derek-ho, line 49: Why is this necessary?

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -345,6 +345,7 @@ private <Request extends ActionRequest, Response extends ActionResponse> void ap
                         log.info("Transport auth in passive mode and no user found. Injecting default user");
                         user = User.DEFAULT_TRANSPORT_USER;
                         threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_USER, user);
+                        threadContext.putPersistent(ConfigConstants.OPENDISTRO_SECURITY_USER, user);
```
derek-ho, line 348: I noticed a few of these putPersistent being added. I do not know the full context, but is there any security implications with putting it in persistently? For example maybe stashing the context wouldn't work if the user is persisted into the threadcontext. Any ideas on this usage @cwperks ?

src/main/java/org/opensearch/security/resources/ResourceSharingIndexManagementRepository.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Creates the resource sharing index if it doesn't already exist.
+     * This method is called during the initialization phase of the repository.
+     * It ensures that the index is set up with the necessary mappings and settings
+     * before any operations are performed on the index.
+     */
+    public void createResourceSharingIndexIfAbsent() {
+        // TODO check if this should be wrapped in an atomic completable future
+
+        this.resourceSharingIndexHandler.createResourceSharingIndexIfAbsent(() -> null);
```
derek-ho, line 36: Do we need this () -> null

src/main/java/org/opensearch/security/resources/ResourceSharingIndexListener.java
```diff
@@ -0,0 +1,130 @@
...
+ * These indices are defined on bootstrap and configured to listen in OpenSearchSecurityPlugin.java
+ */
+public class ResourceSharingIndexListener implements IndexingOperationListener {
+
+    private final static Logger log = LogManager.getLogger(ResourceSharingIndexListener.class);
+
+    private static final ResourceSharingIndexListener INSTANCE = new ResourceSharingIndexListener();
+    private ResourceSharingIndexHandler resourceSharingIndexHandler;
+
+    private boolean initialized;
```
derek-ho, line 37: Can we not use this extra boolean and just base it on the indexlistener itself?

src/main/java/org/opensearch/security/filter/SecurityFilter.java
cwperks: I want to add the same code in https://github.com/opensearch-project/security/pull/4896. I added the notion of persistent TC headers in the core a while ago with the intent that these headers are not stashable and I'd like to move to the authenticated user as a non-stashable header.

Its added here because Security is adding an IndexOperationListener on the resourceIndex and the ThreadContext has already been stashed by the sample plugin so that it can write to the resourceIndex. By adding this call here, @DarshitChanpura is guaranteed that he can read the userinfo from the TC inside the IndexOperationListener. 

I believe the `postIndex` hook is running within the same block as the IndexRequest to the resourceIndex.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexListener.java
```diff
@@ -0,0 +1,130 @@
...
+    public void postDelete(ShardId shardId, Engine.Delete delete, Engine.DeleteResult result) {
+
+        String resourceIndex = shardId.getIndexName();
+        log.info("postDelete called on {}", resourceIndex);
+
+        String resourceId = delete.id();
+
+        boolean success = this.resourceSharingIndexHandler.deleteResourceSharingRecord(resourceId, resourceIndex);
+        if (success) {
+            log.info("Successfully deleted resource sharing entries for resource {}", resourceId);
```
derek-ho, line 123: Nit. Should this be singular since there would only be one resource sharing entry for each resource?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -2082,6 +2127,7 @@ public Collection<Class<? extends LifecycleComponent>> getGuiceServiceClasses()
 
         final List<Class<? extends LifecycleComponent>> services = new ArrayList<>(1);
         services.add(GuiceHolder.class);
+        log.info("Guice service classes loaded");
```
derek-ho, line 2130: Do we want to change this to something lower like debug?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -2109,8 +2155,12 @@ public Collection<SystemIndexDescriptor> getSystemIndexDescriptors(Settings sett
             ConfigConstants.SECURITY_CONFIG_INDEX_NAME,
             ConfigConstants.OPENDISTRO_SECURITY_DEFAULT_CONFIG_INDEX
         );
-        final SystemIndexDescriptor systemIndexDescriptor = new SystemIndexDescriptor(indexPattern, "Security index");
-        return Collections.singletonList(systemIndexDescriptor);
+        final SystemIndexDescriptor securityIndexDescriptor = new SystemIndexDescriptor(indexPattern, "Security index");
+        final SystemIndexDescriptor resourceSharingIndexDescriptor = new SystemIndexDescriptor(
+            ConfigConstants.OPENSEARCH_RESOURCE_SHARING_INDEX,
```
derek-ho, line 2160: Nit: Do we want to add a TODO to support custom resource sharing names?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -2065,6 +2099,17 @@ public void onNodeStarted(DiscoveryNode localNode) {
...
         }
+
+        // create resource sharing index if absent
+        rmr.createResourceSharingIndexIfAbsent();
+
+        for (ResourcePlugin resourcePlugin : OpenSearchSecurityPlugin.GuiceHolder.getResourceService().listResourcePlugins()) {
+            String resourceIndex = resourcePlugin.getResourceIndex();
+
+            this.indicesToListen.add(resourceIndex);
+            log.info("Preparing to listen to index: {} of plugin: {}", resourceIndex, resourcePlugin);
```
derek-ho, line 2110: Same, debug?

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -0,0 +1,343 @@
...
+     * Returns a set of accessible resources for the current user within the specified resource index.
+     *
+     * @param resourceIndex The resource index to check for accessible resources.
+     * @return A set of accessible resource IDs.
+     */
+    public <T> Set<T> getAccessibleResourcesForCurrentUser(String resourceIndex, Class<T> clazz) {
+        if (areArgumentsInvalid(resourceIndex, clazz)) {
+            return Collections.emptySet();
+        }
+        final User user = threadContext.getPersistent(ConfigConstants.OPENDISTRO_SECURITY_USER);
```
derek-ho, line 65: Why are we putting this persistently here?

derek-ho: Is it to access the system index within the handler?

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -0,0 +1,343 @@
...
+        Set<String> scopes
+    ) {
+        if (areArgumentsInvalid(resourceId, resourceIndex, revokeAccess, scopes)) {
+            return null;
+        }
+        final User user = threadContext.getPersistent(ConfigConstants.OPENDISTRO_SECURITY_USER);
+        LOGGER.info("User {} revoking access to resource {} for {} for scopes {} ", user.getName(), resourceId, revokeAccess, scopes);
+
+        // check if user is admin, if yes the user has permission
+        boolean isAdmin = adminDNs.isAdmin(user);
```
derek-ho, line 181: do we want to split this out into a utility function such as hasAdminLevelAcess? Maybe there are additional use cases that we want to support here, such as a "super" role, and not just based on the DN... wdyt?

derek-ho (commented):
Can we also add some test cases for the various handlers?

.gitignore
DarshitChanpura: it was automatically added. But we don't add node_modules and package-lock since we don't have any front-end components in this project.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
DarshitChanpura: +1 to @cwperks answer. User names are particularly required when creating or listing resources. This would ensure that a user who is current context-user is always present and thus not result in null value. At some point in future, the putTransient calls will be removed completely.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexManagementRepository.java
DarshitChanpura: it's callback functionality. Its not used currently but was added in case there were any listeners or related items are to be added.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexListener.java
DarshitChanpura: Its to ensure that ReosurceSharingIndexListener is only initialized once.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexListener.java
DarshitChanpura: addressed.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: Good catch. I will lower it to debug.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: Not right now no. Maybe in the next iteration if there is a request for it.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: This one can remain as is to allow operators to understand what resource plugin indices are being listened upon.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
DarshitChanpura: This is to get user which is then used when creating a resource or listing resources. Without this the values may be null.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -0,0 +1,343 @@
...
+            return false;
+        }
+        final User user = threadContext.getPersistent(ConfigConstants.OPENDISTRO_SECURITY_USER);
+
+        LOGGER.info("Checking if {} has {} permission to resource {}", user.getName(), scope, resourceId);
+
+        // check if user is admin, if yes the user has permission
+        if (adminDNs.isAdmin(user)) {
+            return true;
+        }
```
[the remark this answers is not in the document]
DarshitChanpura: adminDNs is super-admin.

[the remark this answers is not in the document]
DarshitChanpura: non-super-admin will not be able to access these resources.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
DarshitChanpura: This here is specific to super-admin. All "super" roles will require user to have the resource shared with them with that scope.

DarshitChanpura:
> Can we also add some test cases for the various handlers?

This is a WIP. Will flip to ready for review once test cases are added.

DarshitChanpura:
Closing in lieu of https://github.com/opensearch-project/security/pull/5016

Issue #4509, opened 2024-07-01
[BUG] Exception when hot-reloading Let's Encrypt certs issued by alternate intermediate CAs


perios101:
**What is the bug?**
As part of the process using the [REST API](https://opensearch.org/docs/latest/security/configuration/tls/#reload-tls-certificates-on-the-transport-layer) to hot reload transport and http certificates, we receive a 500 error.

The old certificate (and all previous ones to date) has been issued by Let's Encrypt's R3 CA (which seems to have been alive and well as DR CA) and the new one by R11 CA.
**Same scenario** is happening on an environment where intermediate CA changed from R11 to R10. Both are covered by parent ISRG Root X1 CA (along with the old R3).

Restarted individual pods automatically pick up new certificate as valid without issue however hot-reloading existing instances throws the same error over and over despite both CAs are Let's Encrypt issuers covered by root CA.

**How can one reproduce the bug?**
1. Configure and use OpenSearch using Let's Encrypt certificate
2. Renew certificate (using different intermediate CA to the original one
3. Hot-reload certificates using REST API

**What is the expected behavior?**
Certificate validation should succeed as both certificates were issued via valid CA - pinning to same CA as the old one is not considered good practice, especially in DR scenarios (yes, Issuers have those too).

**What is your host/environment?**
 - OS: Amazon Linux (containerised for Kubernetes)
 - Version: OpenSearch 2.14.1
 - Plugins:

**Do you have any screenshots?**
Exception:
`java.io.IOException: OpenSearchSecurityException[Error while initializing transport SSL layer from PEM: java.lang.Exception: New Certs do not have valid Issuer DN, Subject DN or SAN.]; nested: Exception[New Certs do not have valid Issuer DN, Subject DN or SAN.];
`

**Do you have any additional context?**

- From digging into the Security plugin code, it appears logic [HERE](https://github.com/opensearch-project/security/blob/2.14.0.0/src/main/java/org/opensearch/security/ssl/DefaultSecurityKeyStore.java#L655-L662) is the culprit.
- We have the same issue across three different environments (affects also OpenSearch Dashboards - using same security plugin).
- Workaround (restarting pods) exists but it's not scalable / fit as a long-term solution.

shikharj05:
Thanks for filing the issue, I see a previous issue https://github.com/opensearch-project/security/issues/2360 where this was reported and being worked on. @shubhmkr-amazon are you still working on this one?

shubhmkr-amazon:
[...]

PR #4752, opened 2024-09-24
Allow skipping hot reload dn validation

parislarkins:
### Description
This change adds two new boolean settings for the Security plugin: 
- `plugins.security.ssl.http.enforce_cert_reload_dn_verification` 
- `plugins.security.ssl.transport.enforce_cert_reload_dn_verification`

Both options will default to `True`, which will maintain the current behaviour where performing a hot reload requires the Distinguished Names (IssuerDN, SubjectDN, and SAN) in the new certificate to match the current certificate. 

When set to `False`, the Distinguished Names validation will be skipped when hot reloading the respective certificate (HTTP or Transport).

The current behaviour is preventing us from hot-reloading new Let's Encrypt certificates that were signed by a different intermediate CA than the original certificate, meaning we need to perform a rolling restart of each of our clusters in order to rotate the certificates. According to Let's Encrypt, one of their root CAs is expected to expire as soon as 2030 (https://letsencrypt.org/certificates/), so an approach that allowed different intermediate CAs but still rejected changed root CAs will still require us to perform a rolling restart for our clusters when the root CAs are inevitably rotated.

We do not believe adding the ability to disable this validation for hot-reloads creates any added security risk, as long as it is properly considered considering organisational context. There is no similar validation performed when changing certificates by restarting OpenSearch, and an actor that could trigger a hot-reload (requiring them to modify the certificate files on the OpenSearch node) would also likely be able to bypass the validation by restarting OpenSearch. 

For example, in our operational context only the super admin user can trigger a hot-reload. The certificate for the super admin user is stored only on the OpenSearch nodes, so any malicious actor who had the super admin certificate and ability to modify the OpenSearch certificate files would also have the ability to restart OpenSearch. This makes the validation performed when triggering a hot-reload irrelevant, as any malicious actor could simply restart OpenSearch instead. 

* Category: **New feature**

### Issues Resolved
- Resolves https://github.com/opensearch-project/security/issues/4509, although this approach covers a slightly larger scope by allowing skipping the validation entirely, and may not be appropriate for all deployments.

### Testing
Unit tests have been added that cover reloading certificates to a certificate signed by a different Certificate Authority (different root and signing CA), and validating that this is rejected when the new settings are set to `true`, and accepted when the settings are set to `false`.

### Check List
- [ ] New functionality has been documented - draft: https://github.com/parislarkins/documentation-website/pull/1


src/test/java/org/opensearch/security/ssl/SecuritySSLReloadCertsActionTests.java
```diff
@@ -186,6 +201,99 @@ public void testReloadHttpSSLSameCertsPass() throws Exception {
         assertReloadCertificateSuccess(rh, "http", getInitCertDetailsExpectedResponse());
     }
 
+    @Test
+    public void testReloadHttpCertDifferentTrustChain_skipDnValidationPass() throws Exception {
```
shikharj05, line 205: nit: can we add a test to verify default behavior, this is to ensure default doesn't get changed in future without explicit test modification.

shikharj05 (commented):
Thanks for the changes

src/test/java/org/opensearch/security/ssl/SecuritySSLReloadCertsActionTests.java
parislarkins: Absolutely, good suggestion

parislarkins: Added @shikharj05!

src/main/java/org/opensearch/security/ssl/OpenSearchSecuritySSLPlugin.java
```diff
@@ -638,6 +638,23 @@ public List<Setting<?>> getSettings() {
             Setting.longSetting(SSLConfigConstants.SECURITY_SSL_HTTP_CRL_VALIDATION_DATE, -1, -1, Property.NodeScope, Property.Filtered)
         );
 
+        settings.add(
+            Setting.boolSetting(
+                SSLConfigConstants.SECURITY_SSL_HTTP_ENFORCE_CERT_RELOAD_DN_VERIFICATION,
+                SSLConfigConstants.SECURITY_SSL_HTTP_ENFORCE_CERT_RELOAD_DN_VERIFICATION_DEFAULT,
```
cwperks, line 644: nit: I'd be supportive of having the default value inline here instead of creating a constant for it.

src/test/java/org/opensearch/security/ssl/SecuritySSLReloadCertsActionTests.java
```diff
@@ -294,6 +317,29 @@ public void testReloadTransportCertDifferentTrustChain_noSkipDnValidationFail()
         );
     }
 
+    @Test
+    public void testReloadTransportCertDifferentTrustChain_defaultSettingValidationFail() throws Exception {
+        updateFiles(defaultCertFilePath, pemCertFilePath);
+        updateFiles(defaultKeyFilePath, pemKeyFilePath);
+        initTestCluster(pemCertFilePath, pemKeyFilePath, pemCertFilePath, pemKeyFilePath, true, true, null);
```
cwperks, line 324: I know that `true` is the default, but to really test default behavior shouldn't this test pass `null` for both `httpEnforceReloadDnVerification` and `transportEnforceReloadDnVerification`?

cwperks (commented):
Thank you @parislarkins , this change looks reasonable to me. 

What happens if you disable dn validation and change the dn of a node, but the dn is not in the `nodes_dn` list?

src/main/java/org/opensearch/security/ssl/OpenSearchSecuritySSLPlugin.java
parislarkins: Done, thanks!

src/test/java/org/opensearch/security/ssl/SecuritySSLReloadCertsActionTests.java
parislarkins: Agreed, have updated both default test cases

parislarkins:
> Thank you @parislarkins , this change looks reasonable to me.
> 
> What happens if you disable dn validation and change the dn of a node, but the dn is not in the `nodes_dn` list?

@cwperks I have not tested this specifically, but I expect the node would no longer be able to connect to the remote cluster. As I understand it, this is the same behaviour that would happen if the cert was changed by restarting OpenSearch as well.

cwperks (dismissed):

shikharj05 (approved):

src/test/java/org/opensearch/security/ssl/SecuritySSLReloadCertsActionTests.java
```diff
@@ -211,6 +365,15 @@ private void updateFiles(String srcFile, String dstFile) {
         FileHelper.copyFileContents(FileHelper.getAbsoluteFilePathFromClassPath(srcFile).toString(), dstFile);
     }
 
+    private JsonNode getNewCertAuthorityUpdatedCertDetailsExpectedResponse(String updateChannel) {
+        String updateKey = (Objects.equals(updateChannel, "http")) ? HTTP_CERTIFICATES_LIST_KEY : TRANSPORT_CERTIFICATES_LIST_KEY;
+        String oldKey = (Objects.equals(updateChannel, "http")) ? TRANSPORT_CERTIFICATES_LIST_KEY : HTTP_CERTIFICATES_LIST_KEY;
+        final var updatedCertDetailsResponse = DefaultObjectMapper.objectMapper.createObjectNode();
+        updatedCertDetailsResponse.set(updateKey, buildCertsInfoNode(NEW_CA_NODE_CERT_DETAILS));
+        updatedCertDetailsResponse.set(oldKey, buildCertsInfoNode(NODE_CERT_DETAILS));
+        return updatedCertDetailsResponse;
```
derek-ho, line 374: I got kind of confused reading this helper function - can we refactor this for readability? I thought that this was just to help with the dn of either the http or transport certs, but why do we have keys for both http and transport (not just one or the other)?

parislarkins: Thanks @derek-ho, agreed! I have refactored this and the existing `getUpdatedCertDetailsExpectedResponse` into a simpler method. Both http and transport keys are required because the expected JSON response returned here is compared with the response from a `GET` to `_opendistro/_security/api/ssl/certs`, which includes both http and transport certs:

```json
{
    "http_certificates_list": [
        {
            "issuer_dn": "CN=Example Com Inc. Signing CA,OU=Example Com Inc. Signing CA,O=Example Com Inc.,DC=example,DC=com",
            "subject_dn": "CN=node-1.example.com,OU=SSL,O=Test,L=Test,C=DE",
            "san": "[[0, [2.5.4.3, node-1.example.com]], [2, localhost], [2, node-1.example.com], [7, 127.0.0.1], [8, 1.2.3.4.5.5]]",
            "not_before": "2023-04-14T13:22:53Z",
            "not_after": "2033-04-11T13:22:53Z"
        }
    ],
    "transport_certificates_list": [
        {
            "issuer_dn": "CN=Example Com Inc. Signing CA,OU=Example Com Inc. Signing CA,O=Example Com Inc.,DC=example,DC=com",
            "subject_dn": "CN=node-1.example.com,OU=SSL,O=Test,L=Test,C=DE",
            "san": "[[0, [2.5.4.3, node-1.example.com]], [2, localhost], [2, node-1.example.com], [7, 127.0.0.1], [8, 1.2.3.4.5.5]]",
            "not_before": "2023-04-14T13:23:00Z",
            "not_after": "2033-04-11T13:23:00Z"
        }
    ]
}
```

This way we can verify that the cert for only one channel was updated.

derek-ho (approved):
Thanks

PR #4753, opened 2024-09-24
Auto-convert V6 configuration instances into V7 configuration instances (for OpenSearch 2.x only)

nibix:
### Description

**This is secondary to #4546. Please review #4546 first.**

This relates to the proposal "Behind the scenes conversion from V6 to V7" from #4493. 
This converts the V6 config instances to V7 config instances "behind the scenes". This allows to save a lot of duplicate implementation regarding to privilege evaluation and possibly also further logic.

As an additional benefit, this makes it possible to use `SecurityDynamicConfiguration<>` instances in a generically type-safe manner. So far, one usually needs to use  `SecurityDynamicConfiguration<?>` instances and do unsafe casts.

This only touches the runtime representation of the configuration. If there's a V6 configuration in the configuration index, it remains untouched. The possibility to update configuration in V6 format will remain.

This is only for OpenSearch 2.x. For the OpenSearch main branch, #4546 removes V6 config support alltogether.

**Behind the scenes conversion**

Some more detail on what is changed by this PR:

The configuration architecture consists of these levels:

1. JSON documents in the configuration index
2. `ConfigurationRepository` which manages `SecurityDynamicConfiguration<>` instances which contain as entry instances of `org.opensearch.security.securityconf.impl.v6.RoleV6`,  `org.opensearch.security.securityconf.impl.v6.RoleMappingsV6`, `org.opensearch.security.securityconf.impl.v7.RoleV7`,  `org.opensearch.security.securityconf.impl.v7.RoleMappingsV7`, etc.
3. `DynamicConfigFactory` which consumes  `SecurityDynamicConfiguration<>` instances and creates `ConfigModel` and `DynamicConfigModel` instances. If it finds V6 classes in  `SecurityDynamicConfiguration<>`, it creates `ConfigModelV6` instances, if it finds V7 classes, it creates `ConfigModelV7` instances.
4. The `ConfigModelV6` and `ConfigModelV7` classes contain business logic which evaluates roles and permissions.
5. This is used by `PrivilegesEvaluator`.

This PR changes the architecture at level 2:

1. The JSON documents in the index do not change
2. `ConfigurationRepository` is changed in this way: If it finds V6 instances like `org.opensearch.security.securityconf.impl.v6.RoleV6`, it auto-converts these to V7 instances like `org.opensearch.security.securityconf.impl.v7.RoleV7`. The V7 classes already have functionality for that which is already used for the securityadmin migration command.
3.  This makes the distinction of V6/V7 on all layers below unnecessary.  `DynamicConfigFactory` does no longer need a distinction between two config versions.
4.  `ConfigModelV6`  becomes redundant.
5. As `PrivilegeEvaluator` consumes the super interface `ConfigModel`, it does not notice anything about this change.

* Category: Refactoring
* Why these changes are required? - Minimize duplicate logic for privilege evaluation
* What is the old behavior before changes and new behavior after changes? - none

* Category: Refactoring
* Why these changes are required? - Minimize duplicate logic for privilege evaluation
* What is the old behavior before changes and new behavior after changes? no behavior changes

### Issues Resolved

- Partially solves #4493 

### Testing

- New int test `LegacyConfigV6AutoConversionTest`

### Check List


src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -609,36 +607,47 @@ public Map<CType, SecurityDynamicConfiguration<?>> getConfigurationsFromIndex(
...
+
+        return result;
     }
 
-    private Map<CType, SecurityDynamicConfiguration<?>> validate(Map<CType, SecurityDynamicConfiguration<?>> conf, int expectedSize)
-        throws InvalidConfigException {
+    private ConfigurationMap validate(ConfigurationMap conf, int expectedSize) throws InvalidConfigException {
 
-        if (conf == null || conf.size() != expectedSize) {
+        if (conf == null || conf.size() < expectedSize) {
```
cwperks, line 650: Same comment from other PR, is the operator change necessary? Can it ever be the case that `conf.size() > expectedSize`?

src/integrationTest/java/org/opensearch/security/legacy/LegacyConfigV6AutoConversionTest.java
```diff
@@ -0,0 +1,119 @@
...
+        .rawConfigurationDocumentYaml(
+            "internalusers",
+            "admin:\n"
+                + "  readonly: true\n"
+                + "  hash: $2a$12$VcCDgh2NDk07JGN0rjGbM.Ad41qVR/YFJcgHp0UGns5JDymv..TOG\n"
+                + "  roles:\n"
+                + "  - admin\n"
+                + "  attributes:\n"
+                + "    attribute1: value1\n"
+        )//
```
cwperks, line 55: nit: remove the trailing forward slashes

src/main/java/org/opensearch/security/securityconf/impl/CType.java
```diff
@@ -116,11 +212,99 @@ public String configFileName() {
...
+                return convert(NonValidatingObjectMapper.readValue(json, javaType), ctype);
+            } else {
+                return convert(DefaultObjectMapper.readValue(json, javaType), ctype);
+            }
+        }
+
+        public SecurityDynamicConfiguration<NewType> convert(SecurityDynamicConfiguration<OldType> oldConfig, CType<NewType> ctype) {
+            SecurityDynamicConfiguration<NewType> newConfig = SecurityDynamicConfiguration.empty(ctype);
+            newConfig.setAutoConvertedFrom(oldConfig);
+            // newConfig.setSeqNoPrimaryTerm();
```
cwperks, line 300: Why is this commented out? Did you encounter any issues during testing?

src/integrationTest/java/org/opensearch/security/legacy/LegacyConfigV6AutoConversionTest.java
```diff
@@ -0,0 +1,119 @@
...
+                response.getBody()
+            );
+        }
+    }
+
+    /**
+     * This must be the last test executed, as it changes the config index
+     */
+    @Test
+    public void zzz_migrateApi() {
```
cwperks, line 104: Curious, why test the migrate API if there is auto-conversion going on behind the scenes?

If you changed the test to use securityadmin to output the security index to yaml, would the same assertions in this test apply?

cwperks (commented):
Thank you @nibix! Left a few comments. This change looks good to me. This will greatly simplify the security codebase to only have to keep track of the v7 models.

src/integrationTest/java/org/opensearch/security/legacy/LegacyConfigV6AutoConversionTest.java
nibix: done

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
cwperks: Got it, the auto-conversion occurs during Config loading so the output could be n+1 the input if tenants config is generated. Thanks!

src/integrationTest/java/org/opensearch/security/legacy/LegacyConfigV6AutoConversionTest.java
cwperks: Makes sense. Resolving this thread.

cwperks (approved):

cwperks:
@DarshitChanpura @willyborankin @derek-ho @peternied Could we get one more review on this? This change helps pave the way forward for Optimized Privileges Evaluation as it simplifies the code greatly to only depend on one version of the models.

willyborankin (approved):

Issue #4608, opened 2024-07-30
[BUG] Audit Log publish Incorrect FAILED_LOGIN event for Successful login attempt by SAML user on SAML enabled Domain


niravpi:
**What is the bug?**
For SAML enabled domain, audit log publishes ```FAILED_LOGIN``` event for successful login attempt of SAML user.

**How can one reproduce the bug?**
* Create any SAML enabled domain
*  Enable Audit log for that domain
*  Now try to lg in to dashboard via valid SAML user creds
* You can see in the audit logs event there will be ```FAILED_LOGIN```

**What is the expected behavior?**
For SAML enabled domain, there should be no ```FAILED_LOGIN``` audit log event for successful login attempt of SAML user

**What is your host/environment?**
- OS: Linux
- Plugins: Security

**Do you have any screenshots?**
![Screenshot 2024-07-30 at 1 07 04 PM](https://github.com/user-attachments/assets/bd71c21b-1082-473e-ba1a-b35fe31595bc)

**Do you have any additional context?**
NA

expani:
The code path that likely triggers https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L296-L299

cwperks:
[Triage] Thank you for filing this issue @niravpi. That auditlog entry corresponds to this [line](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L306) and it will always be hit on SAML login because the SAML login flow requires `challenge` to be set to `true` to enter the `reRequestAuthentication` flow. This line will always be hit on successful login for SAML and it is confusing since the overall login is successful. I am in favor of removing this audit log entry for the SAML login flow.

TristisOris:
got same issue with 2.19

zabbix server, https requests, login name not logged for unknown reason.
```
{
  "_index": "security-auditlog-2025.03.05",
  "_id": "TKcRZZUBx750GiUNlUtB",
  "_version": 1,
  "_score": null,
  "_source": {
    "audit_cluster_name": " ",
    "audit_node_name": " ",
    "audit_rest_request_method": "HEAD",
    "audit_category": "FAILED_LOGIN",
    "audit_request_origin": "REST",
    "audit_node_id": "XYNU66GNTJa3mpPFpOhWYw",
    "audit_request_layer": "REST",
    "audit_rest_request_path": "/",
[...]

PR #4762, opened 2024-09-30
Remove failed login attempt for saml authenticator

derek-ho:
### Description
Removes audit log entry of failed login for all SAML authentication since it will always re-request authentication

### Issues Resolved
Fix: https://github.com/opensearch-project/security/issues/4608
No
No
### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/auth/BackendRegistry.java
+++ b/src/main/java/org/opensearch/security/auth/BackendRegistry.java
@@ -75,6 +75,7 @@ import org.greenrobot.eventbus.Subscribe;
 import static org.apache.http.HttpStatus.SC_FORBIDDEN;
 import static org.apache.http.HttpStatus.SC_SERVICE_UNAVAILABLE;
 import static org.apache.http.HttpStatus.SC_UNAUTHORIZED;
+import static com.amazon.dlic.auth.http.saml.HTTPSamlAuthenticator.SAML_TYPE;
 
 public class BackendRegistry {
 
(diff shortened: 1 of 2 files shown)


willyborankin (approved):

cwperks:
@derek-ho Can you rebase with the latest from main?

derek-ho:
> @derek-ho Can you rebase with the latest from main?

Done

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -303,7 +304,10 @@ public boolean authenticate(final SecurityRequestChannel request) {
                 if (authDomain.isChallenge()) {
                     final Optional<SecurityResponse> restResponse = httpAuthenticator.reRequestAuthentication(request, null);
                     if (restResponse.isPresent()) {
-                        auditLog.logFailedLogin("<NONE>", false, null, request);
+                        // saml will always hit this to re-request authentication
```
nibix, line 307: I understand the basic problem behind this, but I am wondering whether this is the right path.

We have two challenging auth mechanisms: Basic auth and SAML. SAML will indeed always need a challenge, but also basic will cause challenges in many cases. 

What about basic auth? I have not tested it, but I THINK it will show the same behavior from Dashboards, as it will initially send an unauthenticated request to probe whether authentication is necessary.

We are here in the block guarded by the condition `if (ac == null)` in line 298. That means that no credentials have been sent. Thus, the authentication does not fail because credentials (like user name and password or SAML response) were wrong, but just because the lack of credentials.

Should this really counted then as a failed login? After all, this method won't work for brute-forcing. Maybe the call to `auditLog.logFailedLogin()` should be removed in any case?

cwperks: IMO it should be retained if the initial Basic Auth Request contains wrong credentials. If a user is making a credential-less request then I don't think its necessary. For SAML its always unnecessary.

cwperks: Marking this thread as resolved. This PR solves the linked issue. We can further enhance this for basic auth in another PR to differentiate between credential-less requests vs requests sent with wrong credentials.

PR #4776, opened 2024-10-03
Add deprecation warning for GET/POST/PUT cache

derek-ho:
### Description
Adds deprecation warning that GET/POST/PUT methods for cache endpoint is removed in version 3.0.0. 

### Issues Resolved
None
No

No

### Testing
None
### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
+++ b/src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
@@ -42,17 +42,17 @@ public class FlushCacheApiAction extends AbstractApiAction {
             new DeprecatedRoute(
                 Method.GET,
                 "/cache",
-                "GET is not supported for /cache endpoint and will be removed in the next major version."
+                "GET is not supported for /cache endpoint and will be removed in the next major version. Use DELETE instead."
             ),
             new DeprecatedRoute(
                 Method.PUT,
                 "/cache",
-                "PUT is not supported for /cache endpoint and will be removed in the next major version."
+                "PUT is not supported for /cache endpoint and will be removed in the next major version. Use DELETE instead."
             ),
             new DeprecatedRoute(
                 Method.POST,
                 "/cache",
-                "POST is not supported for /cache endpoint and will be removed in the next major version."
+                "POST is not supported for /cache endpoint and will be removed in the next major version. Use DELETE instead."
             )
         )
     );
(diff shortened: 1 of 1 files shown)


DarshitChanpura (dismissed):
LGTM!

DarshitChanpura (commented):
If you are instead trying to Deprecate the routes itself, you can do something like this:

https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/dlic/rest/api/SecuritySSLCertsApiAction.java#L60

DarshitChanpura (approved):

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
```diff
@@ -39,9 +39,21 @@ public class FlushCacheApiAction extends AbstractApiAction {
     private static final List<Route> routes = addRoutesPrefix(
         ImmutableList.of(
             new Route(Method.DELETE, "/cache"),
-            new Route(Method.GET, "/cache"),
-            new Route(Method.PUT, "/cache"),
-            new Route(Method.POST, "/cache")
+            new DeprecatedRoute(
+                Method.GET,
+                "/cache",
+                "GET is not supported for /cache endpoint and will be removed in the next major version."
```
cwperks, line 45:
```suggestion
                "GET is not supported for /cache endpoint and will be removed in the next major version. Use DELETE instead."
```

RyanL1997 (dismissed):

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
derek-ho: I don't think we can necessarily assume that they want to DELETE the cache - all we should say is we don't support GET/POST/PUT - wdyt?

cwperks: DELETE is the HTTP Verb, so its not deleting the cache, but instructing the nodes to refresh their caches from the security index.

A caller of this API using GET/POST/PUT will already get a not implemented response. I was thinking that the message could be more informative and tell the user the correct http verb to use.

willyborankin (approved):

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
cwperks: If they are using this API then we can assume the intent is to flush the cache.

derek-ho: Sure makes sense. Added that line to the end of all other routes as well in: [8c99aa0](https://github.com/opensearch-project/security/pull/4776/commits/8c99aa0c07286e28e6628e3d0e814235691a76a6)

cwperks (approved):

nibix (approved):

PR #4817, opened 2024-10-16
Fixed bulk index requests in BWC tests and hardened assertions

nibix:
### Description

So far, the BWC test `SecurityBackwardsCompatibilityIT.testDataIngestionAndSearchBackwardsCompatibility()` did not ingest any documents, as the submitted bulk requests were invalid. 

The test already asserted that `_bulk` API calls returned a 200 OK HTTP status. However, this is not sufficient, as the `_bulk` API also returns a 200 status if processing of the bulk items fails.  Such failures are only reflected item-wise in the response body. This indeed happened, as the test erroneously serialized a string containing the JSON document again as JSON, creating a JSON string containing JSON. This was rejected by OpenSearch.

This is now fixed - additional assertions have been created to verify that the items of the bulk request are actually indexed.

Additionally, the DLS rules of the user would not match any of the ingested documents. This would have caused the `_search` API always to return empty result sets - even if any documents were ingested.

The indexed test documents were adjusted to have proper `genre` attributes which can match the configured DLS rule. Additional assertions were added that verify that the DLS and FLS rules are correctly applied to the result set.

* Category: Test fix
* Why these changes are required? - the test did not cover as much functionality as it should
* What is the old behavior before changes and new behavior after changes? - no changes
 
### Testing

- Verified behavior in CI

### Check List

### DIFF
--- a/bwc-test/src/test/java/org/opensearch/security/bwc/SecurityBackwardsCompatibilityIT.java
+++ b/bwc-test/src/test/java/org/opensearch/security/bwc/SecurityBackwardsCompatibilityIT.java
@@ -27,6 +27,7 @@ import org.apache.hc.client5.http.ssl.ClientTlsStrategyBuilder;
 import org.apache.hc.client5.http.ssl.NoopHostnameVerifier;
 import org.apache.hc.core5.http.Header;
 import org.apache.hc.core5.http.HttpHost;
+import org.apache.hc.core5.http.io.entity.StringEntity;
 import org.apache.hc.core5.http.message.BasicHeader;
 import org.apache.hc.core5.http.nio.ssl.TlsStrategy;
 import org.apache.hc.core5.reactor.ssl.TlsDetails;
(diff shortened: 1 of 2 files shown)


willyborankin (approved):

bwc-test/src/test/java/org/opensearch/security/bwc/SecurityBackwardsCompatibilityIT.java
```diff
@@ -239,15 +242,21 @@ private void ingestData(String index) throws IOException {
...
+                bulkRequestBody.append(Song.randomSong().asJson() + "\n");
             }
             List<Response> responses = RestHelper.requestAgainstAllNodes(
                 testUserRestClient,
                 "POST",
                 "_bulk?refresh=wait_for",
-                RestHelper.toHttpEntity(bulkRequestBody.toString())
+                new StringEntity(bulkRequestBody.toString(), APPLICATION_NDJSON)
             );
             responses.forEach(r -> assertThat(r.getStatusLine().getStatusCode(), is(200)));
```
cwperks, line 253: I was not aware of this behavior with BulkRequests. Thank you for updating this to ensure that all items are successful 🙌

cwperks (approved):

RyanL1997 (approved):

cwperks:
@nibix Can you create a manual backport for this PR? I suspect its due to the difference between Apache HttpCliient5 on main vs Apache HttpClient4 on 2.x

nibix:
> @nibix Can you create a manual backport for this PR? I suspect its due to the difference between Apache HttpCliient5 on main vs Apache HttpClient4 on 2.x

Sorry, I missed this request. Will do asap.

PR #4818, opened 2024-10-17
Improve error message when a node with an incorrectly configured certificate attempts to connect

### LINKED ISSUE #4601: [FEATURE] Improve error message when a node with an incorrectly configured certificate attempts to connect
**Problem**

When setting up a cluster, if a node tries to join with a certificate that is not configured correctly in [plugins.security.nodes_dn](https://opensearch.org/docs/latest/security/configuration/tls/#configuring-node-certificates), OpenSearch provides an error message - "Transport client authentication no longer supported". 

```
[2024-07-26T03:02:58,137][WARN ][o.o.d.HandshakingTransportAddressConnector] [os01] handshake failed for [connectToRemoteMasterNode[172.23.0.2:9300]]
opensearch-docker-test-os01-1    | org.opensearch.transport.RemoteTransportException: [os02][172.23.0.2:9300][internal:transport/handshake]
opensearch-docker-test-os01-1    | Caused by: org.opensearch.OpenSearchException: Transport client authentication no longer supported.
```

This is a misleading error - under this circumstance, the problem has nothing to do with the transport client.

**Context of the error message**

My understanding of this error message is as follows.

OpenSearch clusters have two different ports:
- The REST API port, usually port 9200
- The internode communication port (aka 'transport' port), usually port 9300
 
In the early days of Elasticsearch, there was a popular application client, the transport client, that connected on the internode port. This old method of connecting has been deprecated since ES 7.0 in 2019. It was removed in OpenSearch 2.0 (https://github.com/opensearch-project/security/pull/1701). Modern application clients connect on the REST API port.
 
From the perspective of OpenSearch, an incoming connection on port 9300 could be coming from a node, or it could be coming from a transport client.

There is some discussion of this in issue https://github.com/opensearch-project/security/issues/4401 and the linked forum posts https://forum.opensearch.org/t/transport-client-authentication-no-longer-supported-error-when-deploying-cluster-with-security-plugin-enabled/9814/5 / https://forum.opensearch.org/t/transport-client-authentication-no-longer-supported-error-while-implementing-third-party-ca-cert-for-transport-layer/19694 . 

**Proposed solution**

I would like to make this error message clearer.

If we could distinguish between a node joining and a transport client connecting, and provide different error messages for each, that would be ideal.

If it is not easy to tell the difference from the server side, we could still add more information to the error message - something like:

"A connection was attempted on this cluster's internode communication port, but it was refused. If you are attempting to join a node to this cluster, please confirm the Distinguished Name of the new node's certificate is configured in this node's plugins.security.nodes_dn setting. If you are attempting to connect to this cluster with a transport client, please note that transport clients are no longer supported."

It would be nice to also mention the DN that was rejected - for example:

"The distinguished name of the node certificate provided was: 'CN=node.other.com,OU=SSL,O=Test,L=Test,C=DE'"

This would help the user work out what needs to be configured in plugins.security.nodes_dn .

akolarkunnu:
### Description
Updated the error message to understand what is the exact reason of error and renamed the API name to match the intention of API.

### Issues Resolved
Resolves #4601

### Testing
Ran the unit tests

### Check List
- [ ]  ~New Roles/Permissions have a corresponding security dashboards plugin PR~
- [ x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md)

### DIFF
--- a/src/main/java/org/opensearch/security/ssl/util/ExceptionUtils.java
+++ b/src/main/java/org/opensearch/security/ssl/util/ExceptionUtils.java
@@ -77,8 +77,12 @@ public class ExceptionUtils {
     }
 
     public static OpenSearchException clusterWrongNodeCertConfigException(String sslPrincipal) {
-        return new OpenSearchException("Node presenting certificate with SSL Principal {" + sslPrincipal + "} could" +
-                " not securely connect to the cluster. Please ensure the principal is correct and present in the" +
-                " nodes_dn list.");
+        return new OpenSearchException(
+            "Node presenting certificate with SSL Principal {"
+                + sslPrincipal
+                + "} could"
+                + " not securely connect to the cluster. Please ensure the principal is correct and present in the"
+                + " nodes_dn list."
+        );
     }
 }
--- a/src/main/java/org/opensearch/security/ssl/util/ExceptionUtils.java
+++ b/src/main/java/org/opensearch/security/ssl/util/ExceptionUtils.java
@@ -76,7 +76,7 @@ public class ExceptionUtils {
         return new OpenSearchException("An error occurred during the creation of Jwk: {}", cause, cause.getMessage());
     }
 
-    public static OpenSearchException createTransportClientNoLongerSupportedException() {
-        return new OpenSearchException("Transport client authentication no longer supported.");
+    public static OpenSearchException clusterWrongNodeCertConfigException() {
+        return new OpenSearchException("Node certificate configuration is wrong or certificate is invalid.");
     }
 }
--- a/src/main/java/org/opensearch/security/ssl/util/ExceptionUtils.java
+++ b/src/main/java/org/opensearch/security/ssl/util/ExceptionUtils.java
@@ -76,7 +76,9 @@ public class ExceptionUtils {
         return new OpenSearchException("An error occurred during the creation of Jwk: {}", cause, cause.getMessage());
     }
 
-    public static OpenSearchException clusterWrongNodeCertConfigException() {
-        return new OpenSearchException("Node certificate configuration is wrong or certificate is invalid.");
+    public static OpenSearchException clusterWrongNodeCertConfigException(String sslPrincipal) {
+        return new OpenSearchException("Node presenting certificate with SSL Principal {" + sslPrincipal + "} could" +
+                " not securely connect to the cluster. Please ensure the principal is correct and present in the" +
+                " nodes_dn list.");
     }
 }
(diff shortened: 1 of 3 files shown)


akolarkunnu:
This is breaking change because of renaming of API. When I backport this, I will keep the existing API as such and introduce this as a new API, so that it won't break anything.

akolarkunnu:
Can anyone please approve to start workflow(test execution) ?

src/main/java/org/opensearch/security/ssl/util/ExceptionUtils.java
```diff
@@ -76,7 +76,7 @@ public static OpenSearchException createJwkCreationException(Throwable cause) {
         return new OpenSearchException("An error occurred during the creation of Jwk: {}", cause, cause.getMessage());
     }
 
-    public static OpenSearchException createTransportClientNoLongerSupportedException() {
-        return new OpenSearchException("Transport client authentication no longer supported.");
+    public static OpenSearchException clusterWrongNodeCertConfigException() {
```
peternied, line 79: > This is breaking change because of renaming of API. When I backport this, I will keep the existing API as such and introduce this as a new API, so that it won't break anything.

Thanks for this comment, while true at the java API level, the security plugin is less strict around public API changes where the primary concern is that version upgrades are not impacted by renames/updates.

This change is safe to backport to the 2.x line.

peternied (approved):

src/main/java/org/opensearch/security/ssl/util/ExceptionUtils.java
```diff
@@ -76,7 +76,7 @@ public static OpenSearchException createJwkCreationException(Throwable cause) {
...
+        return new OpenSearchException("Node certificate configuration is wrong or certificate is invalid.");
```
cwperks, line 80: nit: wdyt about including the `principal` in the error message as well.

Something like:

`Node presenting certificate with SSL Principal {principal} could not securely connect to the cluster. Please ensure the principal is correct and present in the nodes_dn list`

`nodes_dn` list is a list of trusted distinguished names. This could either be statically defined in `opensearch.yml` or dynamic through the security index with the `nodes.yml` config file.

akolarkunnu: Yes, it's better. I updated the eception message.

nibix (approved):

willyborankin:
@akolarkunnu, could you please fix code formatting? (`gradle spotlessCheck spotlessApply`)

willyborankin (approved):

willyborankin:
@nibix could you plz aprrove it one more time and I will merge it

nibix (approved):

Issue #4495, opened 2024-06-27
[RFC] Retire support for config option `config.dynamic.multi_rolespan_enabled` in `config.yml`


nibix:
## Background

I am asking for opinions and comments on the following topic because it influences the work on #3870.

## Current situation

The OpenSearch security config file `config.yml` supports an option called `config.dynamic.multi_rolespan_enabled`. 

This config option is **undocumented**. Doing a search for it only returns hits which includes dumps of `config.yml` where the option with default value is just coincidentally contained in the config file:

- https://www.google.com/search?q=site%253Aopensearch.org+%22multi_rolespan_enabled%22

The only very brief mention of this option can be found in the docs of a related project:

- https://docs.search-guard.com/7.x-53/changelog-searchguard-7-x-35_0_0#breaking-default-changed-for-permissions-evaluation-across-different-roles-multi-rolespan

### Functionality

If this option is set to `true` (which is the default if a `config.yml` file in version 2/7 is in use), privileges for a single OpenSearch operation can be collected from several roles. If this option is set to `false`, privileges for a single operation must be located in exactly one role.

This can pertain to actions or indices.

#### Several indices

If I issue a search request on indices `a` and `b`, I need the privilege `indices:data/read/search` for both indices.

With `multi_rolespan_enabled = true`, it is possible that the user has two different roles which provide the different privileges:

```
role_1:
  index_permissions:
    index_patterns:
    - a
    allowed_actions:  
    - indices:data/read/search
```    

```
role_2:
  index_permissions:
    index_patterns:
    - b
    allowed_actions:  
    - indices:data/read/search
```

With `multi_rolespan_enabled = false`, however, these roles would be **not sufficient** for a single search operation on both indices. Instead, a user would need to have a single role with both privileges:

```
role_1:
  index_permissions:
    index_patterns:
    - a
    - b
    allowed_actions:  
    - indices:data/read/search
```    

#### Several actions

For example, in order to issue a bulk index insert request, I need two privileges:

- index privilege `indices:data/write/bulk`
- index privilege `indices:data/write/index`

With `multi_rolespan_enabled = true`, it is possible that the user has two different roles which provide the different privileges:

```
role_1:
  index_permissions:
    index_patterns:
    - foo
    allowed_actions:  
    - indices:data/write/bulk
```    

```
role_2:
  index_permissions:
    index_patterns:
    - foo
    allowed_actions:  
    - indices:data/write/index
```

[...]

PR #4898, opened 2024-11-13
[2.x Backport] Optimized Privilege Evaluation

nibix:
### Description

This implements the optimized privilege evaluation as described in #3870 and backports the changes from #4380 to the 2.x branch.

**Important: Before this can be release, the OpenSearch version in the `LEGACY_HEADERS_UNNECESSARY_AS_OF` property must be checked to be in sync with the actual release version. See review comment at https://github.com/opensearch-project/security/pull/4898#pullrequestreview-2431993806** 

This introduces de-normalized data structures that are optimized for the checks that need to be done during privilege evaluation. Additionally, certain objects  (like DLS queries) are prepared ahead of time, as early as possible in order to minimize the overhead during actual privilege evaluation.

This is a big change set - in order to facilitate the review, I have split it into three major commits:

- Optimized action privilege evaluation
- Optimized DLS/FLS/FM privilege evaluation
- Removal of unused code

The code is extensively commented - I hope that will help during review.

* Category: Enhancement

Performance tests indicate that the OpenSearch security layer adds a noticeable overhead to the indexing throughput of an OpenSearch cluster. The overhead may vary depending on the number of indices, the use of aliases, the number of roles and the size of the user object. The goal of these changes is to improve privilege evaluation performance and to make it less dependent on the number of indices, etc.

No significant behavioral changes in the "happy case", when privileges are present.

The undocumented config option `config.dynamic.multi_rolespan_enabled` is no longer evaluated. The code now behaves like it is always set to `true`  - that is the former default. See #4495 for details.

Some slight changes are present in error cases:

- More detailed error messages for missing privileges, showing a index/action matrix of missing privileges
- Errors in the role configuration might be reported (as error log messages) more early, directly after the configuration was applied
- The DLS/FLS implementation now defaults to a "deny by default" implementation. *This is not relevant for normal cases.* This will be only relevant if index requests pass through privileges evaluator even though there are no roles which grant privileges to the requested indices. *Note: This would only happen in case of a bug in the code*. In the previous versions, the DLS/FLS implementation would grant full access to the indices. Now, the DLS/FLS implementation acts as a second barrier, denying access to the indices.

### Issues Resolved

- #3870

This is a backport from #4380

### Testing

- Each new component is accompanied by its own unit test.
- The high-level functionality is validated by the existing integration tests
- Mixed cluster behavior is verfied by `SecurityBackwardsCompatibilityIT` (extended in #4817 )

### Check List


src/main/java/org/opensearch/security/privileges/dlsfls/DlsFlsLegacyHeaders.java
```diff
@@ -0,0 +1,247 @@
...
+ * <p>
+ * As soon as backward compat in mixed clusters is no longer required, this class should be removed.
+ *
+ */
+public class DlsFlsLegacyHeaders {
+    /**
+     * Defines the first OpenSearch version which does not need the legacy headers
+     * TODO this needs to be adapted
+     */
+    static final Version LEGACY_HEADERS_UNNECESSARY_AS_OF = Version.V_EMPTY;
```
nibix, line 48: Important: Before this is releases, this must be set to the OpenSearch version this functionality is released in.

src/main/java/org/opensearch/security/privileges/dlsfls/DlsFlsLegacyHeaders.java
```diff
@@ -0,0 +1,247 @@
...
+ * <p>
+ * As soon as backward compat in mixed clusters is no longer required, this class should be removed.
+ *
+ */
+public class DlsFlsLegacyHeaders {
+    /**
+     * Defines the first OpenSearch version which does not need the legacy headers
+     * TODO this needs to be adapted
+     */
+    static final Version LEGACY_HEADERS_UNNECESSARY_AS_OF = Version.V_2_19_0;
```
nibix, line 48: Important: Before this is released, this must be set to the OpenSearch version this functionality is released in.

krishna-ggk:
Thank for the significant contribution - the improvements look great.

Would you be open to incorporating a feature flag here as well? In a previous major improvement, we found that feature flagging would have allowed us to roll it out more gradually.

nibix:
@krishna-ggk 

> Would you be open to incorporating a feature flag here as well? In a previous major improvement, we found that feature flagging would have allowed us to roll it out more gradually.

Generally, of course. I just have a  few detail questions:

- Are you thinking about a flag that can be changed at runtime, without needing to restart the nodes, or a flag that needs the nodes to be restarted? 
- How shall the rollout strategy be? Ship it disabled and let users enable it optionally? Or ship it enabled and let users disable it just in case?
- Is the plan then to keep it till the last version of OpenSearch 2.x and remove the feature flag for 3.0?

JFTR: Unfortunately, adding a feature flag will require a quite a bit of effort and will significantly increase maintenance effort due to a lot of code duplication. From the top of my head, we need to do this:

- Restore the currently deleted class `SecurityRoles` from  https://github.com/opensearch-project/security/blob/2.x/src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java - this will be roughly 1000 additional LOC
- Duplicate significant parts of  `PrivilegesEvaluator` as it has been changed fundamentally. That will be roughly 500 additional LOC.
- Duplicate `PitPrivilegesEvaluator` (96 LOC), `SystemIndexAccessEvaluator` (320 LOC), `TermsAggregationEvaluator` (100 LOC)
- Duplicate significant parts of `RestLayerPrivilegesEvaluator` (100 LOC)
- Duplicate most of the DLS/FLS code (2000+ LOC)
- Revise the backwards compat concept for DLS/FLS (at the moment it assumes that the change of implementation happens with a specific OpenSearch release version)
- Duplicate all integration tests related to privileges and DLS/FLS and run it both with the feature flag enabled and disabled
- Keep in mind for future features and bug fixes related to privilege evaluation and DLS/FLS, two code parts need to be considered

krishna-ggk:
> @krishna-ggk
> 
> > Would you be open to incorporating a feature flag here as well? In a previous major improvement, we found that feature flagging would have allowed us to roll it out more gradually.
> 
> Generally, of course. I just have a few detail questions:

Thanks @nibix .

> 
> * Are you thinking about a flag that can be changed at runtime, without needing to restart the nodes, or a flag that needs the nodes to be restarted?

Ideally, a runtime flag would be preferable, but a restart-required flag could also work as a simple mitigation if issues arises.

> * How shall the rollout strategy be? Ship it disabled and let users enable it optionally? Or ship it enabled and let users disable it just in case?

Given this is a performance improvement rather than a new feature where we're gathering user feedback, I'd recommend shipping it enabled by default. That way, users can easily disable it if any issues come up, without needing to opt-in.

> * Is the plan then to keep it till the last version of OpenSearch 2.x and remove the feature flag for 3.0?

Since there are no compatibility concerns, we could maintain the feature flag for 2-4 minor versions and then remove it. I'm open to other perspectives on the right timeline as well.



> JFTR: Unfortunately, adding a feature flag will require a quite a bit of effort and will significantly increase maintenance effort due to a lot of code duplication. From the top of my head, we need to do this:
> ..
> * Keep in mind for future features and bug fixes related to privilege evaluation and DLS/FLS, two code parts need to be considered

Thanks. These are very valid points and tradeoff considerations. I understand the additional effort required to implement and maintain a feature flag along with maintenance burden due to increased code duplication. However, given the core nature of the privilege evaluation logic being changed, having that extra safeguard in the short-to-medium term could be quite valuable.

Perhaps a middle ground would be to include the feature flag for 2-3 release cycles, with the default set to "on" starting in 2.19. That way, we get the benefits of the flag without an overly long-lived maintenance burden. What are your thoughts on that approach?

krishna-ggk:
Agreed @nibix, backward compatibility is valid point and infact can remain if there is cross-cluster setup with mixed versions. 

As one thought, could we continue to populate headers irrespective for subsequent versions until we remove the old code at which point we can add back the LEGACY_HEADERS_UNNECESSARY_AS_OF for that version? I understand this may remove some performance benefit, but wondering if it is significant enough.

nibix:
@krishna-ggk 

> As one thought, could we continue to populate headers irrespective for subsequent versions until we remove the old code at which point we can add back the LEGACY_HEADERS_UNNECESSARY_AS_OF for that version? I understand this may remove some performance benefit, but wondering if it is significant enough.

Yes, that should be possible. You are right that this means that users with DLS, FLS or field masking active won't have significant performance improvements in that phase.

cwperks:
I had a PR that I previously closed for Fls/Dls/Field Masking improvements: https://github.com/opensearch-project/security/pull/4706

Would it make sense to re-open that to get the optimization put forth in that PR?

I'd also like to try to get https://github.com/opensearch-project/security/pull/4665 into the 2.x branch as well and wondering if I should re-open that PR and target it to 2.x.

nibix:
@cwperks 

> I had a PR that I previously closed for Fls/Dls/Field Masking improvements: #4706
> 
> Would it make sense to re-open that to get the optimization put forth in that PR?

Well - if the goal is managing risks caused by code changes, then I am not sure if the changes from #4706 would come with a lower risk than the risks implied by this PR.  IMHO, FLS and field masking do have a less than optimal test coverage. This makes any change in this regard risky. Maybe, another strategy of managing the risk introduced by this PR is just to invest in new tests to increase test coverage especially for FLS and field masking?

Concretely about the strategy from #4706: I am worried whether the index information obtained by the request index resolution can be sufficiently trusted in order to be usable for building the DLS/FLS/FM maps. There are are already now examples where this information cannot be used:

- Scroll requests
- PIT searches: https://opensearch.org/docs/latest/search-plugins/searching-data/point-in-time-api/
- I am not sure about CCS, but it also makes use of "unusual" request interactions: https://opensearch.org/docs/latest/search-plugins/cross-cluster-search/

Potentially, there are more similar cases, either in OpenSearch core or in plugins.

I still think the two most promising ways to tackle unknown risks are:

- Intensified testing. Analyze potential sources of issues and try to cover these by tests. Ideally, automated, otherwise manual. Of course, this can only cover "unknowns" which we can identify and thus make "knowns". "Unknowns" which remain "unknowns" are not covered here :)
- As described in https://github.com/opensearch-project/security/pull/4898#issuecomment-2502896729: Feature flag with a kind of capability information stored in the cluster state and thus made available via a `Connection` object. That would be a core change - which however also other future works would benefit from, IMHO.

Both however require a bit of work.

krishna-ggk:
Thanks folks. I'm still responding only to partial aspects mentioned as I'm currently traveling. (I'd love to collaborate more on this with you all more deeply but stuck this week due to travel)

> I still think the two most promising ways to tackle unknown risks are:
> 
> * Intensified testing. Analyze potential sources of issues and try to cover these by tests. Ideally, automated, otherwise manual. Of course, this can only cover "unknowns" which we can identify and thus make "knowns". "Unknowns" which remain "unknowns" are not covered here :)
> * As described in [[2.x Backport] Optimized Privilege Evaluation #4898 (comment)](https://github.com/opensearch-project/security/pull/4898#issuecomment-2502896729): Feature flag with a kind of capability information stored in the cluster state and thus made available via a `Connection` object. That would be a core change - which however also other future works would benefit from, IMHO.
> 
> Both however require a bit of work.

Agreed on the overall direction of addressing the risk. However I still think a feature flag to turn off is super helpful safeguard since the area we are improving is core authz capability that has security implications to users in case of any regression.

I'm wondering if we can perhaps take the approach of validating **correctness** in couple minor versions before in addition to increased coverage and then target performance gain for subsequent version where we cleanup old code? I'm open to other approaches too, but just falling back to this as a potential approach to reduce risk.

nibix:
@krishna-ggk 

> I'm wondering if we can perhaps take the approach of validating **correctness** in couple minor versions before in addition to increased coverage and then target performance gain for subsequent version where we cleanup old code?

From my point of view, that should be possible. If users want to have immediate performance gains, one could even add a second flag which turns off the headers. However users would need to take some precaution to change the flag only at the right state.

kumargu:
I am listing out the outstanding discussions and we will try to reach out to a consensus in a meeting. 

(For now I am going to also put my opinion on each of the points, I'll come back and update the final consensus of the meeting)

>   We need to further discuss if we need a flag that toggled at runtime or if toggling would need a rolling-restart of nodes.

It seems that we are in agreement that we'd need a flag to control feature enablement and disablement. That's great! 
I would vote for a flag that needs a rolling restarts of all nodes in a cluster. This will clean the old state of the node such as thread context and is more deterministic way (as compared to eventually consistency of config propagation).  

> Should we ship the feature disabled or enabled?

In agreement to Gopala, that we should enable it by default, since it involves performance improvements. Given the extent of benchmarking and testing being done here, I would expect a rare need to config switch. In case, if large number of clusters are impacted, we will rely on a patch release (open-source + managed amazon offering). 

> Is the plan then to keep it till the last version of OpenSearch 2.x and remove the feature flag for 3.0? 

Agreed with Gopala that we should keep it for a few versions, but at the same time I would want to get rid of the flag the sooner the better, (maybe in 3.1).   In general, we'll have to support backward compatibility for such large change today and in future.  I would not worry about the duplicated LOC, but would definitely want to get rid of it. Maybe we structure the code folders in a way that its becomes easier to clean the older code with minimal touching needing to the new one.

> We need to consider backwards compatibility on mixed clusters because nodes on OpenSearch versions without this change rely on the presence of thread context headers to determine whether DLS/FLS restrictions need to be applied or not:

I would like to clarify what do we mean by mixed cluster mode? 

[1] In Amazon managed Opensearch, a mixed cluster mode also refers to a migration state when clusters are upgraded from one version to another.  The upgrades could take several hours.

[2] I would like to understand how thread context is set today (sorry for ignorance). Does a peer node have any role setting thread context of a node?  I would assume the answer to be No, since thread context init seems a node property.  If my assumption is correct,  the flag on a node (picked on start/restart) decides if the node should perform DLS/FLS checks via thread context. On restarts of the flag is set to point to new node,  the thread context is never set and older ones are naturally wiped.  Naturally,  for backward compatibility, the old code and new code needs to be maintained together.

nibix:
@kumargu 

> I would like to clarify what do we mean by mixed cluster mode?

With mixed cluster mode, I mean a state of a cluster where there are nodes where the new implementation is active **and** nodes where the old implementation is active at the same time.

In the present implementation, this is just a cluster with two different versions of OpenSearch. In an implementation with a feature flag, that would be a cluster where the feature flag is in the process of being changed via a rolling restart.

I will follow up on the thread context in a second comment.

shikharj05:
>  I would like to clarify what do we mean by mixed cluster mode?
With mixed cluster mode, I mean a state of a cluster where there are nodes where the new implementation is active and nodes where the old implementation is active at the same time.

Mixed cluster also means a cluster using cross-cluster search/replication; I think using different implementations can break cross-cluster search too.

nibix:
@kumargu 

> Thanks @nibix . this is very useful context and sufficient for this discussion

You're welcome! :)

> Now, I think that we can never (or any sooner) get rid of the old code depending on the request headers, since there would be often a mixed cluster state where customers (AWS managed) would attempt to upgrade their clusters from old OS version to newer OS versions. (let's discuss more this in a meeting, but feel free to comment what you think)

Only providing backwards compat to older nodes requires very little code. In the current implementation, this is all concentrated in the class `DlsFlsLegacyHeaders`: 

https://github.com/opensearch-project/security/blob/3c8811148b62796e98c03ef3de07432c81950c95/src/main/java/org/opensearch/security/privileges/dlsfls/DlsFlsLegacyHeaders.java#L43-L247

What are the OpenSearch backwards compat guarantees? Across two major versions? Or more?

However, just to avoid confusion: In my understanding, the feature flag we are talking about would not control just the backward compat code, but the complete implementation of the whole privilege evaluation process, which includes action privileges and DLS/FLS. That will be much more code, maybe 10x as much.

kumargu:
>  In my understanding, the feature flag we are talking about would not control just the backward compat code, but the complete implementation of the whole privilege evaluation process

Yes.

kumargu:
(meeting notes 12/19/24: attendees @cwperks @nibix @kumargu)

1. Given the size of change, we will split this PR into two PRs  [1] DLS/ FLS optimizations involving changes to removal of thread context and [2] Optimizations in privilege evaluations. 
2. We will target [2] to be made available in 2.19 (the last line in 2.x version, expected to land in Feb -2025).  [1] will be made available in 3.0 (expected to land in April 2025).
3. Both optimizations will be enabled by default.
4.    [2] can be disabled via turning off a flag and performing a rolling-restarts of node. We ruled out the need of advanced runtime flag for sake of keeping things simple and deterministic. (we may introduce runtime flags for future releases, no doubt they will give more seem-less experience).  
5. There will be no feature flag for [1].   We discussed of having two feature flags but we settled on having only one the reasons being a.  the DLS/ FLS changes are less destructive and any breaking changes in this code path is expected to be caught early (in tests); since they are on the request path.  b. We didn't want to introduce too many flags which adds confusion. 
6. The feature flag will be removed in 3.x (3.1 potentially) --  this is when we remove all the old-duplicated code for PR [2].  Only for PR [1] we need to maintain backward compatibility in mixed cluster state (see previous discussion). The old-duplicated code for this change can be removed only in 4.0. (to get to 4.0 customers have to be at 3.x and hence no mixed cluster state during version upgrades).
7. There were few (3?) security issues reported around DLS/FLS working which shows the lack of test coverage. To begin with, we will fix all known issues in this area and then start adding more coverage. We will need help from (@cwperks) to collect those issues and (@kumargu) will help in prioritization.

(let me know if I have missed anything, also for those who missed the meeting, please leave your comments if you are not aligned with the decisions)

krishna-ggk:
Thank you @nibix  @cwperks @kumargu for working through the concerns - the outcome summarized [here](https://github.com/opensearch-project/security/pull/4898#issuecomment-2556854032) makes sense.

nibix:
@krishna-ggk @kumargu @cwperks 

Here is the new PR as discussed: #4998 It introduces old and new implementation alongside of each other. They are switchable by the setting `plugins.security.privileges_evaluation.use_legacy_impl`. The PR only contains action privileges and not DLS/FLS.

I still need to review test coverage to see if I need to add some more tests to cover the legacy implementation.

nibix:
Closing this as this is superseded by https://github.com/opensearch-project/security/pull/4998

kumargu:
> However, if we also need to merge similar changes to the main branch, quite big additional efforts will be required in order to roll back all the code at one point in time

@cwperks can we hold back until 3.0/3.1 until merging this is mainline. In 3.0/3.1 we'll get rid of all the duplicated code and merge to mainline would be cleaner.

cwperks:
> > However, if we also need to merge similar changes to the main branch, quite big additional efforts will be required in order to roll back all the code at one point in time
> 
> @cwperks can we hold back until 3.0/3.1 until merging this is mainline. In 3.0/3.1 we'll get rid of all the duplicated code and merge to mainline would be cleaner.

What do you mean? Are you suggesting to revert https://github.com/opensearch-project/security/pull/4380 and then rebase https://github.com/opensearch-project/security/pull/4998 to main and backport to 2.x?

kumargu:
> Are you suggesting to revert https://github.com/opensearch-project/security/pull/4380 and then rebase https://github.com/opensearch-project/security/pull/4998 to main and backport to 2.x?

Yes.

PR #4921, opened 2024-11-19
Scaffolding for POST/DELETE/GET api tokens calls

derek-ho:
### Description
This PR puts the high level scaffolding in place to handle GET/POST/DELETE api tokens api calls. 
Notably:
- function to create an index to hold api tokens if it doesn't exist
- function to create a document representing api tokens
- function to delete a document corresponding to a given name
- function to get all documents representing api tokens

This PR is not a finalized/api token structure may be subject to change up until merge into main branch (this is designated for feature branch), and does not perform any authc/authz checks using these newly introduced API tokens. 

### Issues Resolved

### Testing

### Check List


cwperks:
A couple of high-level comments:

1. This PR currently looks like scaffolding to build on, what is the goal with this PR?

2. Can you add tests for different scenarios that this PR handles?

3. IMO Tokens should have permissions/scopes directly associated with them - not roles. We could consider allowing the API to request issuing a token to accept a role and copy the permissions from the role.

derek-ho:
> A couple of high-level comments:
> 
> 1. This PR currently looks like scaffolding to build on, what is the goal with this PR?

Updated the description. The goal of this PR is to mainly have a working POC/high level scaffolding in place to interact with the underlying opensearch cluster, such as creating/getting/deleting documents in an opensearch index representing API tokens. This PR does not finalize/use those documents to perform any authc/z checks yet. 

> 2. Can you add tests for different scenarios that this PR handles?

Yes, I am working on that now, and I will add those before this PR is merged. 

> 3. IMO Tokens should have permissions/scopes directly associated with them - not roles. We could consider allowing the API to request issuing a token to accept a role and copy the permissions from the role.

Yup - I am pushing a commit to address this. My high level approach will be to ask for user input for cluster permissions and index permissions (the same input when creating a role in the first place) and then performing authz checks on the ephemeral role, which will be created from these cluster/index permissions but not persisted to the index. Let me know if this high level approach makes sense to you CC @DarshitChanpura

src/main/java/org/opensearch/security/action/apitokens/ApiToken.java
```diff
@@ -0,0 +1,214 @@
...
+            throw new IllegalArgumentException("creation_time is required");
+        }
+
+        return new ApiToken(description, jti, clusterPermissions, indexPermissions, creationTime);
+    }
+
+    private static RoleV7.Index parseIndexPermission(XContentParser parser) throws IOException {
+        List<String> indexPatterns = new ArrayList<>();
+        List<String> allowedActions = new ArrayList<>();
+        String dls = "";
```
cwperks, line 116: Supporting FLS/DLS/Field Masking will introduce complications on token issuance. As part of a check to see that a user can issue a token it needs to be 2 criteria:

1. The user has permissions to issue a token
2. The permissions associated with the token are a subset of the permissions that the user has. With Fls/Dls/Field Masking its not clear how to achieve #2.

I think we should make tokens simple from the start and only allow index patterns and allowed_actions for index permissions.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenIndexHandler.java
```diff
@@ -0,0 +1,147 @@
...
+    private Client client;
+    private ClusterService clusterService;
+    private static final Logger LOGGER = LogManager.getLogger(ApiTokenIndexHandler.class);
+
+    public ApiTokenIndexHandler(Client client, ClusterService clusterService) {
+        this.client = client;
+        this.clusterService = clusterService;
+    }
+
+    public String indexToken(ApiToken token) {
```
cwperks, line 57: nit: name this `indexTokenMetadata` or `indexTokenPayload`. The token is not persisted so the name of this method could be misleading.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenIndexHandler.java
```diff
@@ -0,0 +1,147 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+    }
+
+    public void deleteToken(String name) throws ApiTokenException {
+        try (final ThreadContext.StoredContext ctx = client.threadPool().getThreadContext().stashContext()) {
+            DeleteByQueryRequest request = new DeleteByQueryRequest(ConfigConstants.OPENSEARCH_API_TOKENS_INDEX).setQuery(
+                QueryBuilders.matchQuery("description", name)
```
cwperks, line 84: Why "description" here?

src/main/java/org/opensearch/security/action/apitokens/ApiTokenIndexHandler.java
```diff
@@ -0,0 +1,147 @@
...
+            long deletedDocs = response.getDeleted();
+
+            if (deletedDocs == 0) {
+                throw new ApiTokenException("No token found with name " + name);
+            }
+            LOGGER.info("Deleted " + deletedDocs + " documents");
+        }
+    }
+
+    public Map<String, ApiToken> getApiTokens() {
```
cwperks, line 98: Should we consider building filtering into this. i.e. passing `name` as a param to filter the results?

cwperks (commented):
This is a good base to continue working off of! Good start to this much requested feature :). It would be great to see an end-to-end integ test where a user can request a token is created and then use it to make requests against the cluster.

src/main/java/org/opensearch/security/action/apitokens/ApiToken.java
derek-ho: This is a good callout. I made a sub - IndexPermission class within ApiToken which represents the barebones index_pattern and allowed_actions only, will dis-regard the dls, fls, and field masking for now.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenIndexHandler.java
derek-ho: Renamed to name

src/main/java/org/opensearch/security/action/apitokens/ApiTokenIndexHandler.java
derek-ho: I just wanted to limit the code for now. In the future this may make sense, but I also don't expect users to create too many tokens to begin with, so not sure if this will make the cut or not, so for now I left it out.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenIndexHandler.java
derek-ho: Good callout. I also renamed the get method to be more clear in: [dad767b](https://github.com/opensearch-project/security/pull/4921/commits/dad767b9559dd2e9f4d059059a68c642a8aae935)

src/main/java/org/opensearch/security/action/apitokens/ApiTokenIndexHandler.java
```diff
@@ -0,0 +1,149 @@
...
+            client.index(request, irListener);
+            return token.getName();
+
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+    }
+
+    public void deleteToken(String name) throws ApiTokenException {
```
cwperks, line 83: Any audit logging present on internal user endpoints should be applicable here as well. Same audit logging for create user or delete user should have an equivalent entry here.

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -124,6 +126,13 @@ protected AbstractAuditLog(
             ConfigConstants.SECURITY_CONFIG_INDEX_NAME,
             ConfigConstants.OPENDISTRO_SECURITY_DEFAULT_CONFIG_INDEX
         );
+        // TODO: support custom api tokens index?
+        this.securityOrApiTokensIndexMatcher = WildcardMatcher.from(
```
DarshitChanpura, line 130: you can rename this to : `securityIndicesMatcher`

src/main/java/org/opensearch/security/action/apitokens/ApiToken.java
```diff
@@ -0,0 +1,222 @@
...
+        String name = null;
+        String jti = null;
+        List<String> clusterPermissions = new ArrayList<>();
+        List<IndexPermission> indexPermissions = new ArrayList<>();
+        Instant creationTime = null;
+
+        XContentParser.Token token;
+        String currentFieldName = null;
+
+        while ((token = parser.nextToken()) != XContentParser.Token.END_OBJECT) {
```
DarshitChanpura, line 99: mind adding a comment above the while loop about what a pre-parsed doc would look like?

src/main/java/org/opensearch/security/compliance/ComplianceConfig.java
```diff
@@ -508,7 +508,8 @@ public boolean writeHistoryEnabledForIndex(String index) {
             return false;
         }
         // if security index (internal index) check if internal config logging is enabled
-        if (securityIndex.equals(index)) {
+        // TODO: Add support for custom api token index?
+        if (securityIndex.equals(index) || securityIndex.equals(ConfigConstants.OPENSEARCH_API_TOKENS_INDEX)) {
```
DarshitChanpura, line 512: this should be replaced by a matcher as done in AbstractAuditLog class.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenRepository.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public String createApiToken(String name, List<String> clusterPermissions, List<ApiToken.IndexPermission> indexPermissions) {
+        apiTokenIndexHandler.createApiTokenIndexIfAbsent();
+        // TODO: Implement logic of creating JTI to match against during authc/z
+        // TODO: Add validation on whether user is creating a token with a subset of their permissions
+        return apiTokenIndexHandler.indexTokenMetadata(new ApiToken(name, "test-token", clusterPermissions, indexPermissions));
+    }
+
+    public void deleteApiToken(String name) throws ApiTokenException {
+        apiTokenIndexHandler.createApiTokenIndexIfAbsent();
```
DarshitChanpura, line 35: why do we delete token if Index doesn't exist?

src/main/java/org/opensearch/security/action/apitokens/ApiTokenRepository.java
```diff
@@ -0,0 +1,44 @@
...
+
+public class ApiTokenRepository {
+    private final ApiTokenIndexHandler apiTokenIndexHandler;
+
+    public ApiTokenRepository(Client client, ClusterService clusterService) {
+        apiTokenIndexHandler = new ApiTokenIndexHandler(client, clusterService);
+    }
+
+    public String createApiToken(String name, List<String> clusterPermissions, List<ApiToken.IndexPermission> indexPermissions) {
+        apiTokenIndexHandler.createApiTokenIndexIfAbsent();
```
DarshitChanpura, line 28: i think we should create this index on bootstrap. Can be done in later PRs.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
```diff
@@ -0,0 +1,287 @@
...
+import static org.opensearch.security.action.apitokens.ApiToken.INDEX_PERMISSIONS_FIELD;
+import static org.opensearch.security.action.apitokens.ApiToken.NAME_FIELD;
+import static org.opensearch.security.dlic.rest.support.Utils.addRoutesPrefix;
+
+public class ApiTokenAction extends BaseRestHandler {
+    private final ApiTokenRepository apiTokenRepository;
+
+    private static final List<RestHandler.Route> ROUTES = addRoutesPrefix(
+        ImmutableList.of(
+            new RestHandler.Route(POST, "/apitokens"),
```
DarshitChanpura, line 49: if we are handling only one token at a time, IMO we should change the endpoint to `apitoken`. Thoughts?

derek-ho: I think the standard across the repo is to have it be a plural instead of a singular, even when handling PUT/POST. I think it is just to maintain the same route across all routes of the same action. To follow the standards already set i think plural makes sense here.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenRepository.java
derek-ho: Sure I was thinking about that - I will add a TODO here to circle back to make a decision.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenRepository.java
derek-ho: I have logic handling the delete of the api token if it doesn't exist. My thinking around creating the index here is that if users are trying to call on the delete of this endpoint in the first place they may be interested in this feature, so we create the index for them (even if the token doesn't exist). Again this can be changed in future/if index is created on bootstrap, but we can re visit this when we decide on the correct approach for index creation.

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
derek-ho: Good suggestion, made the change.

src/main/java/org/opensearch/security/compliance/ComplianceConfig.java
derek-ho: Sure done in: [58cf8ca](https://github.com/opensearch-project/security/pull/4921/commits/58cf8cac4696ee0e0c8ea6eb58eea10f1fe7a460)

src/main/java/org/opensearch/security/action/apitokens/ApiToken.java
derek-ho: Good idea added in: [58cf8ca](https://github.com/opensearch-project/security/pull/4921/commits/58cf8cac4696ee0e0c8ea6eb58eea10f1fe7a460)

src/main/java/org/opensearch/security/action/apitokens/ApiTokenIndexHandler.java
derek-ho: I added a hack to do so by preserving the user in the indexhandler. This seems to be taken care of by the abstractAPIaction, so I think we will need to consolidate/abstract this away to a common place. We can also handle the authorization of this API in that abstraction, since that is something else that needs to happen. For now confirmed that the hack is working and an audit log entry for api tokens actions is created.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenRepository.java
```diff
@@ -0,0 +1,44 @@
...
+        // TODO: Implement logic of creating JTI to match against during authc/z
+        // TODO: Add validation on whether user is creating a token with a subset of their permissions
+        return apiTokenIndexHandler.indexTokenPayload(new ApiToken(name, "test-token", clusterPermissions, indexPermissions));
```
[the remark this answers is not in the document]
derek-ho: Updated the name.

src/main/java/org/opensearch/security/action/apitokens/ApiToken.java
```diff
@@ -0,0 +1,239 @@
...
+                                indexPermissions.add(parseIndexPermission(parser));
+                            }
+                        }
+                        break;
+                }
+            }
+        }
+
+        if (name == null) {
+            throw new IllegalArgumentException(NAME_FIELD + " is required");
```
DarshitChanpura, line 150: can this be empty? if not you can leverage isNullOrEmpty supplied by core

DarshitChanpura: Another option would be to create a method for this validation checks:
```java
public void validateFields(String... fields) throws IllegalArgumentException {
   for (String field : fields) { 
   
   // switch block to validate against the field variables declared at the start of the class
   }
}
```

DarshitChanpura: You can also remove it there are validations in place to prevent null or Empty values from being fed into the index document.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
```diff
@@ -0,0 +1,287 @@
...
+        )
+    );
+
+    public ApiTokenAction(ClusterService clusterService, ThreadPool threadPool, Client client) {
+        this.apiTokenRepository = new ApiTokenRepository(client, clusterService);
+    }
+
+    @Override
+    public String getName() {
+        return "Actions to get and create API tokens.";
```
DarshitChanpura, line 61: change this to a name with snake case: `manage_api_tokens`

src/main/java/org/opensearch/security/action/apitokens/ApiTokenIndexHandler.java
```diff
@@ -0,0 +1,169 @@
...
+        this.client = client;
+        this.clusterService = clusterService;
+    }
+
+    public String indexTokenMetadata(ApiToken token) {
+        // TODO: move this out of index handler class, potentially create a layer in between baseresthandler and abstractapiaction which can
+        // abstract this complexity away
+        final var originalUserAndRemoteAddress = Utils.userAndRemoteAddressFrom(client.threadPool().getThreadContext());
+        try (final ThreadContext.StoredContext ctx = client.threadPool().getThreadContext().stashContext()) {
+            client.threadPool()
```
DarshitChanpura, line 65: this should be retrieved as a Persistent Header. But this should be done in BackendRegistry and can be taken as a separate PR.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenRepository.java
DarshitChanpura: We should not be creating an index. Instead we can modify the method definition to throw an exception when Index is not found.

DarshitChanpura: Same for below.

DarshitChanpura (commented):
Final few comments. Looks ready to be merged otherwsie. Thanks @derek-ho !

src/main/java/org/opensearch/security/action/apitokens/ApiTokenIndexHandler.java
derek-ho: Yes will follow up on this when seeing what the right approach to consolidating the logic is.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
derek-ho: Done

DarshitChanpura (approved):

PR #4967, opened 2024-12-13
Adds JTI and expiration field support for API Tokens

derek-ho:
### Description
This change adds support for expiration field (to be used to determine validity of api tokens), as well as support for generating the JTI that represents the token. To generate the JTI the existing securityTokenManager's jwtvendor logic was re-used, and new Api tokens specific dynamic config model settings were added. 

Enhancement
Support vending the token itself and expiration
Return dummy jti, no support for expiration

### Issues Resolved

### Testing

### Check List


src/main/java/org/opensearch/security/securityconf/impl/v7/ConfigV7.java
```diff
@@ -495,4 +498,52 @@ public String toString() {
...
+            return encryptionKey;
+        }
+
+        public void setEncryptionKey(String encryptionKey) {
+            this.encryptionKey = encryptionKey;
+        }
+
+        @Override
+        public String toString() {
+            return "ApiTokens [ enabled=" + enabled + ", signing_key=" + signingKey + ", encryption_key=" + encryptionKey + "]";
```
DarshitChanpura, line 544: nit: "ApitTokenSettings ..."

src/test/java/org/opensearch/security/authtoken/jwt/JwtVendorTest.java
```diff
@@ -270,4 +272,103 @@ public void testCreateJwtLogsCorrectly() throws Exception {
...
+        final String expectedClusterPermissions = "cluster:admin/*";
+        final String expectedIndexPermisions = indexPermission.toString();
+
+        LongSupplier currentTime = () -> (long) 100;
+        String claimsEncryptionKey = "1234567890123456";
+        Settings settings = Settings.builder()
+            .put("signing_key", signingKeyB64Encoded)
+            .put("encryption_key", claimsEncryptionKey)
+            // CS-SUPPRESS-SINGLE: RegexpSingleline get Extensions Settings
+            .put(ConfigConstants.EXTENSIONS_BWC_PLUGIN_MODE, true)
```
DarshitChanpura, line 293: why this setting? also we should add a test for unsupported algorithm exception?

src/main/java/org/opensearch/security/action/apitokens/ApiTokenRepository.java
```diff
@@ -17,19 +17,36 @@
...
+    private final SecurityTokenManager securityTokenManager;
 
-    public ApiTokenRepository(Client client, ClusterService clusterService) {
+    public ApiTokenRepository(Client client, ClusterService clusterService, SecurityTokenManager tokenManager) {
         apiTokenIndexHandler = new ApiTokenIndexHandler(client, clusterService);
+        securityTokenManager = tokenManager;
     }
 
-    public String createApiToken(String name, List<String> clusterPermissions, List<ApiToken.IndexPermission> indexPermissions) {
+    public ApiTokenRepository(ApiTokenIndexHandler apiTokenIndexHandler1, SecurityTokenManager tokenManager) {
```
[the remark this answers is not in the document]
derek-ho: This is a constructor that I created just for testing to inject a mocked apiTokenIndexHandler. Otherwise I had to do some hacky things with setting it via reflection. I'm not sure about the standards around this - is it bad to create a constructor that is only used in testing?

src/main/java/org/opensearch/security/action/apitokens/ApiToken.java
```diff
@@ -28,37 +36,50 @@ public class ApiToken implements ToXContent {
...
         this.indexPermissions = indexPermissions;
+        this.expiration = expiration;
+    }
 
+    public ApiToken(String name, List<String> clusterPermissions, List<IndexPermission> indexPermissions) {
+        this.creationTime = Instant.now();
+        this.name = name;
+        this.clusterPermissions = clusterPermissions;
+        this.indexPermissions = indexPermissions;
+        this.expiration = Long.MAX_VALUE;
```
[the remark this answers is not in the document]
derek-ho: Changed to default to 30 day expiry from now in: [98301f8](https://github.com/opensearch-project/security/pull/4967/commits/98301f80c8e597502a8ce3113005705f665c055b)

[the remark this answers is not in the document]
derek-ho: Good call out, default infinite expiry seems less secure!

src/main/java/org/opensearch/security/action/apitokens/ApiToken.java
```diff
@@ -28,37 +36,50 @@ public class ApiToken implements ToXContent {
...
-        this.jti = jti;
+        this.name = name;
         this.clusterPermissions = clusterPermissions;
         this.indexPermissions = indexPermissions;
         this.creationTime = creationTime;
+        this.expiration = expiration;
 
     }
 
+    public void setJti(String jti) {
```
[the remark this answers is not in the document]
derek-ho: Changed all fields to be final, and changed the call signature so that we get the JTI from the securityTokenManager first in order to create the immutable ApiToken.

src/main/java/org/opensearch/security/action/apitokens/ApiToken.java
```diff
@@ -84,6 +105,37 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
...
+                actionsArray.add(action);
+            }
+
+            json.add(INDEX_PATTERN_FIELD, patternsArray);
+            json.add(ALLOWED_ACTIONS_FIELD, actionsArray);
+
+            return json.toString();
+        }
+
+        public static IndexPermission fromString(String str) {
```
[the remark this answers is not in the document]
derek-ho: Added a TODO for that

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
```diff
@@ -182,6 +188,13 @@ void validateRequestParameters(Map<String, Object> requestBody) {
             throw new IllegalArgumentException("Missing required parameter: " + NAME_FIELD);
         }
 
+        if (requestBody.containsKey(EXPIRATION_FIELD)) {
+            Object expiration = requestBody.get(EXPIRATION_FIELD);
+            if (!(expiration instanceof Long)) {
+                throw new IllegalArgumentException(EXPIRATION_FIELD + " must be an long");
```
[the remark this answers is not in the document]
derek-ho: Corrected the grammar.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -148,4 +152,68 @@ public ExpiringBearerAuthToken createJwt(
...
+        claimsBuilder.subject(subject);
+        claimsBuilder.audience(audience);
+        claimsBuilder.notBeforeTime(now);
+
+        final Date expiryTime = new Date(expiration);
+        claimsBuilder.expirationTime(expiryTime);
+
+        if (clusterPermissions != null) {
+            final String listOfClusterPermissions = String.join(",", clusterPermissions);
+            claimsBuilder.claim("cp", encryptionDecryptionUtil.encrypt(listOfClusterPermissions));
```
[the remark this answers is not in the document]
derek-ho: Yup, that is cleaner and I changed it to use that.

src/main/java/org/opensearch/security/securityconf/impl/v7/ConfigV7.java
derek-ho: Fixed.

src/test/java/org/opensearch/security/authtoken/jwt/JwtVendorTest.java
derek-ho: Removed this setting, it was unnecessary. The unsupported algorithm I got was from when my key was not long enough, when it is long enough it does not throw that exception. In any case I also added a test for it in the JWT vendor (which is where it originates from) in: [b84d2cf](https://github.com/opensearch-project/security/pull/4967/commits/b84d2cfb34369a11717d102fb04f27977b8d3510)

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -148,4 +152,68 @@ public ExpiringBearerAuthToken createJwt(
...
+        if (indexPermissions != null) {
+            List<String> permissionStrings = new ArrayList<>();
+            for (ApiToken.IndexPermission permission : indexPermissions) {
+                permissionStrings.add(permission.toString());
+            }
+            final String listOfIndexPermissions = String.join(",", permissionStrings);
+            claimsBuilder.claim("ip", encryptionDecryptionUtil.encrypt(listOfIndexPermissions));
+        }
+
+        final JWSHeader header = new JWSHeader.Builder(JWSAlgorithm.parse(signingKey.getAlgorithm().getName())).build();
```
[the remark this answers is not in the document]
derek-ho: No it is based on if the key is not long enough, which I added a test to verify in: [b84d2cf](https://github.com/opensearch-project/security/pull/4967/commits/b84d2cfb34369a11717d102fb04f27977b8d3510)

src/main/java/org/opensearch/security/action/apitokens/ApiToken.java
```diff
@@ -109,6 +152,7 @@ public static ApiToken fromXContent(XContentParser parser) throws IOException {
         List<String> clusterPermissions = new ArrayList<>();
         List<IndexPermission> indexPermissions = new ArrayList<>();
         Instant creationTime = null;
+        long expiration = Long.MAX_VALUE;
```
DarshitChanpura, line 155: should this be Long.MAX_VALUE?

src/main/java/org/opensearch/security/action/apitokens/ApiTokenRepository.java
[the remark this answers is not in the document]
DarshitChanpura: Yes, general practice is to not create public constructors. Ideally you should create a static method, say `forTest()` that returns a test constructor.

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
```diff
@@ -131,6 +141,33 @@ public ExpiringBearerAuthToken issueOnBehalfOfToken(final Subject subject, final
...
     }
 
+    public ExpiringBearerAuthToken issueApiToken(
+        final String name,
+        final Long expiration,
+        final List<String> clusterPermissions,
+        final List<ApiToken.IndexPermission> indexPermissions
+    ) {
+        final User user = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+        if (user == null) {
```
DarshitChanpura, line 151: in what cases will this be nul?? super-admin or plugins?

src/main/java/org/opensearch/security/action/apitokens/ApiToken.java
```diff
@@ -84,6 +95,38 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
             builder.endObject();
             return builder;
         }
+
+        // TODO: look into integrating with default object mapper
+        @Override
```
cwperks, line 100: This is what `toXContent` is meant for. For objects that implement ToXContent you can jsonify them like this:

I believe you can do something like this to get json for debugging:

```
token.toXContent(XContentFactory.jsonBuilder(), ToXContent.EMPTY_PARAMS).toString();
```

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -148,4 +152,68 @@ public ExpiringBearerAuthToken createJwt(
...
+        claimsBuilder.subject(subject);
+        claimsBuilder.audience(audience);
+        claimsBuilder.notBeforeTime(now);
+
+        final Date expiryTime = new Date(expiration);
+        claimsBuilder.expirationTime(expiryTime);
+
+        if (clusterPermissions != null) {
+            final String listOfClusterPermissions = String.join(",", clusterPermissions);
+            claimsBuilder.claim("cp", encryptString(listOfClusterPermissions));
```
cwperks, line 180: Why the use of encryption?

Encryption was used with claims like `backend_roles` and `roles` in obo because we made the choice that roles information was sensitive and should be opaque to extensions. One of the goals was not to propagate some of the existing problems with plugins to the extensions ecosystem. For example, we did not want to propagate roles injection to extensions. If roles injection was supported, then an extension would have been able to send roles (maybe either by http header or transport) to the opensearch cluster to tell it how to authorize. Given that extensions were 3pp and designed to be zero-trust that was not a practice that could be extended to extensions.

Back then we discussed tokens as the strategic way for async tasks for extensions...and we can do the same for plugins as well to eventually deprecate roles injection.

For the extensions use-case, the thinking was that a user can obtain a token (if they have the correct permissions) and then upload it to the extension (via the extension's frontend ui). This was the extension can make calls with the token on the original user's behalf, without the original user needing an active session.

cwperks: Ok I see, in the event a token is leaked the entity obtaining the token should not be able to determine its scope.

I'm not sure if these values need to be included in the payload. We are storing the metadata (including permissions) in a system index and then mirror that system index in a cache for performance. When a client makes a request using a token, we can authenticate it and then look up its permissions in the cache using the jti.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenRepository.java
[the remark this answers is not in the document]
derek-ho: Fixed in: [08e0890](https://github.com/opensearch-project/security/pull/4967/commits/08e08900b914f7df626692108b8099a745343bee) according to suggestions.

src/main/java/org/opensearch/security/action/apitokens/ApiToken.java
derek-ho: It shouldn't matter here, since when the api token is created and expiration is not specified it defaults to 30 days from creation time. But I changed it to 0 to feel safer in: [08e0890](https://github.com/opensearch-project/security/pull/4967/commits/08e08900b914f7df626692108b8099a745343bee)

src/main/java/org/opensearch/security/action/apitokens/ApiToken.java
derek-ho: Nice! That's what I was looking for. I also implemented fromXcontent for the reverse (which I will need to use after decrypting the indexpermission from the token) in: [08e0890](https://github.com/opensearch-project/security/pull/4967/commits/08e08900b914f7df626692108b8099a745343bee)

src/main/java/org/opensearch/security/identity/SecurityTokenManager.java
derek-ho: It should only be null if the threadcontext is stashed. Since I am always calling it with the user in the threadcontext, I removed the null check

derek-ho: In: [09018dc](https://github.com/opensearch-project/security/pull/4967/commits/09018dc5dce3143acb896088887550956874175c)

DarshitChanpura (approved):

Issue #4760, opened 2024-09-30
[BUG/performance] Serialization/Deserialization check for safe classes adds performance overhead


krishna-ggk:
**What is the bug?**
While debugging latency contributors in an OpenSearch 2.13 cluster using JDK serialization, we noticed [SafeSerializationUtils.isSafeClass](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/SafeSerializationUtils.java#L70) adds ~20% overhead to latency although it isn't as visible in flamegraphs (0.02%) (This was validated by short circuiting isSafeClass to return).

<img width="541" alt="image" src="https://github.com/user-attachments/assets/2a35519c-f86e-4bc8-8c55-921ae5140b70">

![image](https://github.com/user-attachments/assets/f3a6b412-62f2-4f14-babd-c00e5dee7223)

**What is the expected behavior?**
It seems like a very low hanging fruit to gain significant performance benefit.

PR #4973, opened 2024-12-16
Refactor SafeSerializationUtils for better performance

shikharj05:
### Description
* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation) - Enhancement
* Why these changes are required? Resolves https://github.com/opensearch-project/security/issues/4760
Adds a cache to avoid repeated checks for same class. Removes usage of stream() in favor of a for loop

### Issues Resolved
https://github.com/opensearch-project/security/issues/4760

### Testing
UTs

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/support/SafeSerializationUtils.java
+++ b/src/main/java/org/opensearch/security/support/SafeSerializationUtils.java
@@ -17,12 +17,11 @@ import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 
 import org.opensearch.security.auth.UserInjector;
@@ -57,7 +56,7 @@ public final class SafeSerializationUtils {
         LdapAttribute.class
     );
 
-    private static final List<Class<?>> SAFE_ASSIGNABLE_FROM_CLASSES = ImmutableList.of(
+    private static final Set<Class<?>> SAFE_ASSIGNABLE_FROM_CLASSES = ImmutableSet.of(
         InetAddress.class,
         Number.class,
         Collection.class,
@@ -66,12 +65,23 @@ public final class SafeSerializationUtils {
     );
 
     private static final Set<String> SAFE_CLASS_NAMES = Collections.singleton("org.ldaptive.LdapAttribute$LdapAttributeValues");
+    private static final Map<Class<?>, Boolean> safeClassCache = new ConcurrentHashMap<>();
 
     static boolean isSafeClass(Class<?> cls) {
-        return cls.isArray()
-            || SAFE_CLASSES.contains(cls)
-            || SAFE_CLASS_NAMES.contains(cls.getName())
-            || SAFE_ASSIGNABLE_FROM_CLASSES.stream().anyMatch(c -> c.isAssignableFrom(cls));
+        return safeClassCache.computeIfAbsent(cls, SafeSerializationUtils::computeIsSafeClass);
+    }
+
+    private static boolean computeIsSafeClass(Class<?> cls) {
+        return cls.isArray() || SAFE_CLASSES.contains(cls) || SAFE_CLASS_NAMES.contains(cls.getName()) || isAssignableFromSafeClass(cls);
+    }
+
+    private static boolean isAssignableFromSafeClass(Class<?> cls) {
+        for (Class<?> safeClass : SAFE_ASSIGNABLE_FROM_CLASSES) {
+            if (safeClass.isAssignableFrom(cls)) {
+                return true;
+            }
+        }
+        return false;
     }
 
     static void prohibitUnsafeClasses(Class<?> clazz) throws IOException {
@@ -79,5 +89,4 @@ public final class SafeSerializationUtils {
             throw new IOException("Unauthorized serialization attempt " + clazz.getName());
         }
     }
-
 }
--- a/src/test/java/org/opensearch/security/support/SafeSerializationUtilsTest.java
+++ b/src/test/java/org/opensearch/security/support/SafeSerializationUtilsTest.java
@@ -1,0 +1,88 @@
+package org.opensearch.security.support;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.util.*;
+import java.util.regex.Pattern;
+
+import org.junit.Test;
+
+import org.opensearch.security.auth.UserInjector;
+import org.opensearch.security.user.User;
+
+import com.amazon.dlic.auth.ldap.LdapUser;
+import org.ldaptive.AbstractLdapBean;
+import org.ldaptive.LdapAttribute;
+import org.ldaptive.LdapEntry;
+import org.ldaptive.SearchEntry;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class SafeSerializationUtilsTest {
+
+    @Test
+    public void testSafeClasses() {
+        assertTrue(SafeSerializationUtils.isSafeClass(String.class));
+        assertTrue(SafeSerializationUtils.isSafeClass(InetSocketAddress.class));
+        assertTrue(SafeSerializationUtils.isSafeClass(Pattern.class));
+        assertTrue(SafeSerializationUtils.isSafeClass(User.class));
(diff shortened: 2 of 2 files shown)


src/test/java/org/opensearch/security/support/SafeSerializationUtilsTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class SafeSerializationUtilsTest {
+
+    @Test
+    public void testSafeClasses() {
```
cwperks, line 39: Would it be possible to add a test to ensure that there is a cache hit the second time `SafeSerializationUtils.isSafeClass` is called for the same class?

shikharj05: Thanks, I have added a test to verify cache size; I think that should help.

src/main/java/org/opensearch/security/support/SafeSerializationUtils.java
```diff
@@ -57,7 +56,7 @@ public final class SafeSerializationUtils {
         LdapAttribute.class
     );
 
-    private static final List<Class<?>> SAFE_ASSIGNABLE_FROM_CLASSES = ImmutableList.of(
+    private static final Set<Class<?>> SAFE_ASSIGNABLE_FROM_CLASSES = ImmutableSet.of(
```
kumjiten, line 59: nit: Since we're iterating over it, List or Set will not make any difference.

src/main/java/org/opensearch/security/support/SafeSerializationUtils.java
```diff
@@ -66,18 +65,28 @@ public final class SafeSerializationUtils {
...
 
     private static final Set<String> SAFE_CLASS_NAMES = Collections.singleton("org.ldaptive.LdapAttribute$LdapAttributeValues");
+    static final Map<Class<?>, Boolean> safeClassCache = new ConcurrentHashMap<>();
 
     static boolean isSafeClass(Class<?> cls) {
-        return cls.isArray()
-            || SAFE_CLASSES.contains(cls)
-            || SAFE_CLASS_NAMES.contains(cls.getName())
-            || SAFE_ASSIGNABLE_FROM_CLASSES.stream().anyMatch(c -> c.isAssignableFrom(cls));
+        return safeClassCache.computeIfAbsent(cls, SafeSerializationUtils::computeIsSafeClass);
```
kumjiten, line 71: query: What is the expected number of classes we're considering? As Map will keep on growing for unsafe classes, it might impact as large objects will not fit in cache lines.

cwperks: From my own understand, there should only ever be safe classes as the security plugin controls what is serialized in the [SecurityInterceptor](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java#L133-L360). If an unsafe class is seen, there is an error thrown on [deserialization](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/Base64JDKHelper.java#L153). (Need to verify there's an existing test for this case.)

This map is expected to be quite small. For SAFE_CLASSES and SAFE_CLASS_NAMES, we know how many entries there could be in total. For SAFE_ASSIGNABLE_FROM_CLASS_NAMES, there could be a variable number of entries depending on the concrete classes, but there are only a limited amount of subclasses for each class in that list.

derek-ho (approved):
LGTM

src/main/java/org/opensearch/security/support/SafeSerializationUtils.java
kumjiten: Thanks, @cwperks, for clarifying. If there are only safe classes, now I'm wondering why this check was added initially?

cwperks: Not sure. It comes from the initial ODFE commit: https://github.com/opendistro-for-elasticsearch/security/blob/436fdd97dcbf0f6be9fc9808bfb0df24e31dad67/src/main/java/com/amazon/opendistroforelasticsearch/security/support/Base64Helper.java

I see that the interceptor also exists in the initial commit as well.

kumjiten: +1, which comes from [search-guards](https://github.com/floragunncom/search-guard/commit/2d867255b9664e186f7546d587725b8dc6995bc1#diff-265b33cfc4754f739a44739c48b4be039a2b8ca3ad7da3b26bcf18d6b4592f95)

shikharj05: From the look of it, not all classes are known in advance, right? Classes inherited from ``SAFE_ASSIGNABLE_FROM_CLASSES`` could be any number of classes and hence the check. For example, if any unknown class were allowed to be deserialized, it could potentially lead to various issues like cluster instability, exploits, etc. I will check the classes/number of classes on a running test cluster and revert back here

nibix: Just adding some background:

All the trouble is caused by the LDAP auth module as it uses an custom sub-class of `User`, namely [LdapUser](https://github.com/opensearch-project/security/blob/main/src/main/java/com/amazon/dlic/auth/ldap/LdapUser.java) .

If that class would not exist, serialization of the User object would be constrained to a very limited set of types.

The `LdapUser` subclass only exists to allow interaction between the `LDAPAuthenticationBackend` and the `LDAPAuthorizationBackend`. Otherwise, it is not necessary. 

It's still on my todo list to write an issue about optimization potentials around the `User` object which also includes abolishing the `LdapUser` subclass and making the class `User` final and immutable. There will be actually quite a few advantages coming from that. I hope that I'll get the time to take care of this soon.

nibix (approved):

kumjiten (approved):

src/main/java/org/opensearch/security/support/SafeSerializationUtils.java
shikharj05: Thanks @nibix  for the context.

I ran a OSB test locally with a 2-node cluster, the map size did not grow beyond 9 entries.

> opensearch-benchmark execute-test --pipeline=benchmark-only --workload=http_logs --target-host=https://localhost:9200 --client-options=basic_auth_user:admin,basic_auth_password:<password>,verify_certs:false

```
|                                                     Store size |                                                        |     17.9018 |      GB |
|                                                  Translog size |                                                        | 2.71481e-06 |      GB |
|                                         Heap used for segments |                                                        |           0 |      MB |
|                                       Heap used for doc values |                                                        |           0 |      MB |
|                                            Heap used for terms |                                                        |           0 |      MB |
|                                            Heap used for norms |                                                        |           0 |      MB |
|                                           Heap used for points |                                                        |           0 |      MB |
|                                    Heap used for stored fields |                                                        |           0 |      MB |
|                                                  Segment count |                                                        |          66 |         |
|                                                 Min Throughput |                                           index-append |     17861.7 |  docs/s |
|                                                Mean Throughput |                                           index-append |     62341.8 |  docs/s |
|                                              Median Throughput |                                           index-append |     94204.8 |  docs/s |
|                                                 Max Throughput |                                           index-append |     98542.1 |  docs/s |
```


Node 1-
```
[2024-12-18T11:50:54,759][WARN ][o.o.s.s.SafeSerializationUtils] [smoketestnode] Size of safeClassCache map is 9
[2024-12-18T11:50:54,759][WARN ][o.o.s.s.SafeSerializationUtils] [smoketestnode] Size of safeClassCache map is 9
[2024-12-18T11:50:54,759][WARN ][o.o.s.s.SafeSerializationUtils] [smoketestnode] Size of safeClassCache map is 9
[2024-12-18T11:50:55,766][WARN ][o.o.s.s.SafeSerializationUtils] [smoketestnode] Size of safeClassCache map is 9
[2024-12-18T11:50:55,766][WARN ][o.o.s.s.SafeSerializationUtils] [smoketestnode] Size of safeClassCache map is 9
```

Node 2-
```
[2024-12-18T14:50:53,834][WARN ][o.o.s.s.SafeSerializationUtils] [smoketestnode2] Size of safeClassCache map is 9
[2024-12-18T14:50:53,834][WARN ][o.o.s.s.SafeSerializationUtils] [smoketestnode2] Size of safeClassCache map is 9
[2024-12-18T14:50:53,834][WARN ][o.o.s.s.SafeSerializationUtils] [smoketestnode2] Size of safeClassCache map is 9
[2024-12-18T14:50:53,834][WARN ][o.o.s.s.SafeSerializationUtils] [smoketestnode2] Size of safeClassCache map is 9
[2024-12-18T14:50:53,835][WARN ][o.o.s.s.SafeSerializationUtils] [smoketestnode2] Size of safeClassCache map is 9
```

kumjiten: Thanks, @shikharj05, for running OSB and verifying.

PR #4975, opened 2024-12-17
correct authentication_backend type internal

yangsongbai:


cwperks (changes requested):
FYI both `intern` and `internal` are accepted for this value: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java#L259-L260

There are many more [places in the code](https://github.com/search?q=repo%3Aopensearch-project%2Fsecurity+%5C%22intern%5C%22&type=code) that refer to the value `intern`. IMO they should either all be updated or the use of `intern` should remain. 

For fresh values, this update would get sourced into the security index, but for clusters with an existing `.opendistro_security` index in order to see the change applied a cluster admin would either need to use securityadmin or the securityconfig api (if enabled)

I like the intent of the change because `internal` is more readable then the shorthand `intern`, but we would need to ensure there aren't any backward compatibility issues with this config change.

PR #4992, opened 2024-12-24
Api token authc/z implementation with Cache

derek-ho:
### Description
This PR implements authc/z for api tokens. For authc it is based on token validity for as well as presence in a cache, which listens to index and delete operations on the api token index. For Authz, it onboards onto action privileges class, where instead of returning not authorized for cluster and index privileges, requests go through a final api token check, where if the user is an api token, it will evaluate whether the permissions in the cache are sufficient to execute the request.

### Issues Resolved

### Testing

### Check List


src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -584,22 +584,24 @@ public void logDocumentWritten(ShardId shardId, GetResult originalResult, Index
...
-                            originalResult.internalSourceRef(),
-                            XContentType.JSON
-                        )
-                    ) {
-                        Object base64 = parser.map().values().iterator().next();
-                        if (base64 instanceof String) {
-                            originalSource = (new String(BaseEncoding.base64().decode((String) base64), StandardCharsets.UTF_8));
-                        } else {
-                            originalSource = XContentHelper.convertToJson(originalResult.internalSourceRef(), false, XContentType.JSON);
+                    if (originalSource == null) {
```
derek-ho, line 587: Changes in this file are to correct a mis-merge that happened in prior PRs to this feature branch.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
```diff
@@ -45,7 +46,7 @@ public class PrivilegesEvaluationContext {
     private final IndexResolverReplacer indexResolverReplacer;
     private final IndexNameExpressionResolver indexNameExpressionResolver;
     private final Supplier<ClusterState> clusterStateSupplier;
-
+    private final ApiTokenIndexListenerCache apiTokenIndexListenerCache = ApiTokenIndexListenerCache.getInstance();
```
cwperks, line 49: Why add this to PrivilegesEvaluatorContext?

I think this should be handled similarly to the security configCache from ConfigurationRepository: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java

cwperks: Note that security uses a publish-subscribe model to [notify consumers](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java#L569-L579) about changes to the security config. See [BackendRegistry.onDynamicConfigModelChanged](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L175-L193) for an example of a subscriber.

src/test/java/org/opensearch/security/action/apitokens/ApiTokenAuthenticatorTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+        AuthCredentials credentials = authenticator.extractCredentials(request, threadcontext);
+
+        assertNull("Should return null when JTI is not in allowlist cache", credentials);
+    }
+
+    @Test
+    public void testExtractCredentialsPassWhenJtiInCache() {
+        String testJti =
+            "eyJhbGciOiJIUzUxMiJ9.eyJzdWIiOiJ0ZXN0LXRva2VuIiwiYXVkIjoidGVzdC10b2tlbiIsIm5iZiI6MTczNTMyNjM0NywiaXAiOiJnaGlWTXVnWlBtcHZJMjZIT2hmUTRnaEJ1Qkh0Y2x6c1REYVlxQjVBRklyTkE4SzJCVTdxc2toMVBCOEMzQWpTdVBaREM0THVSM2pjZkdpLzlkU2ZicDBuQTNGMkhtSi9jaDA3cDY2ZWJ6OD0iLCJpc3MiOiJvcGVuc2VhcmNoLWNsdXN0ZXIiLCJleHAiOjIyMTg3NjU2NDMzMCwiaWF0IjoxNzM1MzI2MzQ3LCJjcCI6IjlPS3N4aFRaZ1pmOVN1akprRzFXVWIzVC9RVTZ4YmZTc2hrbllkVUhraHM9In0.kqMSnn5YwhLmeiI_8iIBQ5uhmI52n2MNniAa52Zpfs3TiE_PXKiNbDNs08hNqzGYW772gT7lfvp6kZnFxQ4v2Q";
```
cwperks, line 82: Use `Jwts.builder()` to construct jwts for testing: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L175-L193

src/main/java/org/opensearch/security/action/apitokens/ApiTokenIndexListenerCache.java
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * This method is called after an index operation is performed.
+     * It adds the JTI of the indexed document to the cache and maps the document ID to the JTI (for deletion handling).
+     * @param shardId The shard ID of the index where the operation was performed.
+     * @param index The index where the operation was performed.
+     * @param result The result of the index operation.
+     */
+    @Override
+    public void postIndex(ShardId shardId, Engine.Index index, Engine.IndexResult result) {
```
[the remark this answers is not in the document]
derek-ho: Cleaned it up to use the transport action to refresh all nodes here: [bc8aacf](https://github.com/opensearch-project/security/pull/4992/commits/bc8aacf42eb1ee1d85b8bfa46c798430d24f7291)

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
derek-ho: The reason why I am adding the instance to the context is for evaluation within the authz flow. I am not sure if the publish/subscribe model would work well for api tokens, which seems to be different than the other roles within the cluster. I feel like it is wrong to hook it up such that the actionprivileges need to be re-computed for all roles in the cluster on creation of a new api token (which does not affect the existing roles of the cluster). That is why I went forward with the instance approach that can be modified under the hood without re-computation of the other roles datastructures. Is there any concern with this instance approach other than not following the patterns existing within the security repo?

src/integrationTest/java/org/opensearch/security/privileges/ActionPrivilegesTest.java
```diff
@@ -258,6 +262,69 @@ public void hasAny_wildcard() throws Exception {
                 isForbidden(missingPrivileges("cluster:whatever"))
             );
         }
+
+        @Test
+        public void apiToken_explicit_failsWithWildcard() throws Exception {
+            SecurityDynamicConfiguration<RoleV7> roles = SecurityDynamicConfiguration.fromYaml("test_role:\n" + //
```
cwperks, line 268: If the roles in this are unused, can we use `SecurityDynamicConfiguration.empty(CType.ROLES);` instead?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
```diff
@@ -172,4 +173,8 @@ public String toString() {
             + mappedRoles
             + '}';
     }
+
+    public ApiTokenIndexListenerCache getApiTokenIndexListenerCache() {
```
cwperks, line 177: IMO Cache is an implementation detail. The class should be about accessing metadata about the tokens like the permissions associated with a token.

cwperks: See [ConfigurationRepository](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java) which has methods for accessing the security configuration (from the security index or from the cache). I think repository is a good name here to let the caller of the class know that its used to access info about the security configuration.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
derek-ho: Changed it such that I extract the permissions from the repository upon creation of the context.

src/integrationTest/java/org/opensearch/security/privileges/ActionPrivilegesTest.java
derek-ho: Applied the suggested change to all instances where I did not need to use the roles.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
derek-ho: Removed the cache listener and moved it into the repository to handle all data operations in one place.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
```diff
@@ -172,4 +180,41 @@ public String toString() {
...
+        this.action = action;
+        this.request = request;
+        this.clusterStateSupplier = clusterStateSupplier;
+        this.indexResolverReplacer = indexResolverReplacer;
+        this.indexNameExpressionResolver = indexNameExpressionResolver;
+        this.task = task;
+        this.permissionsForApiToken = permissions;
+    }
+
+    private Permissions extractApiTokenPermissionsForUser(ApiTokenRepository apiTokenRepository) {
```
cwperks, line 211: Can this call be moved to within the PrivilegesEvaluator and then pass the resulting permissions into this context?

Since there is overlap here between roles/token permissions I'm wondering if this can be structured in a way to make it obvious that privileges are being evaluated on a user that is mapped to roles or on a token that has a set of permissions associated with it.

src/main/java/org/opensearch/security/privileges/ActionPrivileges.java
```diff
@@ -407,7 +414,56 @@ PrivilegesEvaluatorResponse providesPrivilege(PrivilegesEvaluationContext contex
...
+
+            // Check for exact match
+            if (!Collections.disjoint(resolvedClusterPermissions, actions)) {
+                return PrivilegesEvaluatorResponse.ok();
+            }
+
+            // Check for pattern matches (like "cluster:*")
+            for (String permission : resolvedClusterPermissions) {
+                // skip pure *, which was evaluated above
+                if (permission != "*") {
```
cwperks, line 448: `!=` operator shouldn't be used on strings.

Is this check necessary? When `permission` is `*` it creates a Matcher that always returns true without having to do any regex or glob matching. I think this case should be covered by L454-459

src/main/java/org/opensearch/security/authtoken/jwt/ApiTokenJwtVendor.java
```diff
@@ -0,0 +1,83 @@
...
+        claimsBuilder.issueTime(now);
+        claimsBuilder.subject(subject);
+        claimsBuilder.audience(audience);
+        claimsBuilder.notBeforeTime(now);
+        final Date expiryTime = new Date(expiration);
+        claimsBuilder.expirationTime(expiryTime);
+
+        final JWSHeader header = new JWSHeader.Builder(JWSAlgorithm.parse(signingKey.getAlgorithm().getName())).build();
+
+        final SignedJWT signedJwt = AccessController.doPrivileged(
```
cwperks, line 71: Why id doPrivileged needed here?

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -11,63 +11,43 @@
...
-
-        if (isKeyNull(settings, "encryption_key")) {
-            throw new IllegalArgumentException("encryption_key cannot be null");
-        } else {
-            this.encryptionDecryptionUtil = new EncryptionDecryptionUtil(settings.get("encryption_key"));
-        }
-        this.timeProvider = timeProvider.orElse(System::currentTimeMillis);
+public abstract class JwtVendor {
+    public ExpiringBearerAuthToken createJwt(
+        final String issuer,
```
cwperks, line 35: I think we should work towards a single `JwtVendor` that can be initialized with Settings and separate claims builders for payloads for different types of tokens that this repo may issue.

i.e. `OBOClaims.Builder().issuer(<issuer>)` or `ApiTokenClaims.Builder()...`

src/main/java/org/opensearch/security/privileges/ActionPrivileges.java
derek-ho: This is to cover the case where when we want to check for explicit permission matching, if it was rejected in the first block, we don't want it to pass the wildcard matcher below.

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
derek-ho: Agree. The reason why I had to split it up here was the Obo jwt vendor requires an encryption key, whereas my api token jwt vendor does not use the encryption key at all... what do you think is a good approach to centralize them when one has a required setting (errors out without it), whereas one does not use it?

src/main/java/org/opensearch/security/action/apitokens/ApiTokenRepository.java
```diff
@@ -13,19 +13,48 @@
...
 public class ApiTokenRepository {
     private final ApiTokenIndexHandler apiTokenIndexHandler;
     private final SecurityTokenManager securityTokenManager;
+    private static final Logger log = LogManager.getLogger(ApiTokenRepository.class);
 
+    private final Map<String, Permissions> jtis = new ConcurrentHashMap<>();
+
+    void reloadApiTokensFromIndex() {
+        jtis.clear();
+        Map<String, ApiToken> tokensFromIndex = apiTokenIndexHandler.getTokenMetadatas();
```
cwperks, line 38: I think there's a risk of a race-condition here where `jtis` will be empty for a brief period. 

With the security index, the `configCache` is keyed by `CType` and it does a blanket replacement after reloading: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java#L561

If the same instance of `jtis` is kept, does it make sense to first replace any of the existing keys and remove and keys that don't exist in `tokensFromIndex`, but exist in the `jtis`?

src/main/java/org/opensearch/security/authtoken/jwt/ApiTokenJwtVendor.java
derek-ho: Without this block creating it results in the following error:

```
 18:01:52.700 [opensearch[opensearch-node1][transport_worker][T#1]] ERROR org.opensearch.security.identity.SecurityTokenManager - Error creating Api Token for admin
opensearch-node1-1       | java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "accessDeclaredMembers")
opensearch-node1-1       | 	at java.base/java.security.AccessControlContext.checkPermission(AccessControlContext.java:488) ~[?:?]
opensearch-node1-1       | 	at java.base/java.security.AccessController.checkPermission(AccessController.java:1085) ~[?:?]
opensearch-node1-1       | 	at java.base/java.lang.SecurityManager.checkPermission(SecurityManager.java:411) ~[?:?]
opensearch-node1-1       | 	at java.base/java.lang.Class.checkMemberAccess(Class.java:3299) ~[?:?]
opensearch-node1-1       | 	at java.base/java.lang.Class.getDeclaredConstructor(Class.java:3000) ~[?:?]

```

src/main/java/org/opensearch/security/privileges/ActionPrivileges.java
derek-ho: Switched to .equals() in: [685dcfc](https://github.com/opensearch-project/security/pull/4992/commits/685dcfc8b7752fc4b118b1b85e3b9e6f827e7c0d), thanks!

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
derek-ho: Moved in: [685dcfc](https://github.com/opensearch-project/security/pull/4992/commits/685dcfc8b7752fc4b118b1b85e3b9e6f827e7c0d)

src/main/java/org/opensearch/security/action/apitokens/ApiTokenRepository.java
derek-ho: I got what you are saying. I switched it such that first I get all tokens from the index, then boot out any ones that don't exist anymore, then add any new entries in: [685dcfc](https://github.com/opensearch-project/security/pull/4992/commits/685dcfc8b7752fc4b118b1b85e3b9e6f827e7c0d). I think this would still result in a short period of time in which revoked api tokens could still be used, but I think it is short enough, and this addresses the original concern. Thoughts @cwperks ?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
```diff
@@ -45,7 +46,7 @@ public class PrivilegesEvaluationContext {
     private final IndexResolverReplacer indexResolverReplacer;
     private final IndexNameExpressionResolver indexNameExpressionResolver;
     private final Supplier<ClusterState> clusterStateSupplier;
-
+    private final Permissions permissionsForApiToken;
```
cwperks, line 49: nit: just call this `permissions`. API Tokens may directly map to permissions, but this could be used generically for any such similar use case.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -304,7 +309,17 @@ public PrivilegesEvaluationContext createContext(
         TransportAddress caller = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
         ImmutableSet<String> mappedRoles = ImmutableSet.copyOf((injectedRoles == null) ? mapRoles(user, caller) : injectedRoles);
 
-        return new PrivilegesEvaluationContext(user, mappedRoles, action0, request, task, irr, resolver, clusterStateSupplier);
+        return new PrivilegesEvaluationContext(
```
cwperks, line 312: I think this should be split into 2 separate classes: `RoleBasedPrivilegesEvaluatorContext` vs `PermissionBasedPrivilegesEvaluatorContext` instead of trying to overload this class to support 2 separate use-cases.

src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java
```diff
@@ -377,6 +386,24 @@ private void buildAAA() {
...
         }
 
+        /*
+         * If the Api token authentication is configured:
+         * Add the ApiToken authbackend in to the auth domains
+         * Challenge: false - no need to iterate through the auth domains again when ApiToken authentication failed
+         * order: -2 - prioritize the Api token authentication when it gets enabled
+         */
+        Settings apiTokenSettings = getDynamicApiTokenSettings();
+        log.info("APITOKENSETTINGS" + apiTokenSettings.toString());
```
cwperks, line 396: nit: remove this line

src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java
```diff
@@ -241,6 +247,9 @@ public Settings getDynamicApiTokenSettings() {
             .build();
     }
 
+    @Inject
```
cwperks, line 250: Remove these lines

src/main/java/org/opensearch/security/authtoken/jwt/JwtVendor.java
```diff
@@ -103,7 +94,7 @@ static Tuple<JWK, JWSSigner> createJwkFromSettings(final Settings settings) {
         }
     }
 
-    public ExpiringBearerAuthToken createJwt(
```
cwperks, line 106: Ideally, I'd like to keep this generic `public ExpiringBearerAuthToken createJwt(JwtClaimsSet claims);` and rely on the caller to use the correct instance of a JWTVender.

With the changes in this PR, its designed in a way that the same JWTVendor can issue OBO Tokens and API Tokens.

Should we raise a PR against main to make this more generic for the existing OBO use-case?

derek-ho: This suggestion makes sense. Let's not split it up/PR against main. Let's just keep changes in this feature branch: it will get picked up with the merge to main later.

src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java
derek-ho: Done.

src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java
derek-ho: Done.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
derek-ho: Done - this is cleaner IMO

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluationContext.java
derek-ho: Done - good idea to keep it general to support more use cases in the future.

src/main/java/org/opensearch/security/authtoken/jwt/claims/JwtClaimsBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+
+public class JwtClaimsBuilder {
+    private final JWTClaimsSet.Builder builder;
+    private EncryptionDecryptionUtil encryptionDecryptionUtil;
+
+    public JwtClaimsBuilder() {
+        this.builder = new JWTClaimsSet.Builder();
+    }
+
+    public JwtClaimsBuilder encryptionKey(String encryptionKey) {
```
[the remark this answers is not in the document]
derek-ho: Done.

src/main/java/org/opensearch/security/privileges/ActionPrivileges.java
```diff
@@ -449,7 +456,58 @@ PrivilegesEvaluatorResponse providesPrivilege(PrivilegesEvaluationContext contex
                 }
             }
 
-            return PrivilegesEvaluatorResponse.insufficient(action);
```
cwperks, line 452: wdyt about creating 2 `providesPrivilege` methods. One for roles-based and one for permissions-based? The last arg of this method takes a set of roles so it already looks like its specifically designed for roll-based permission evaluation. Maybe there can be a branch in https://github.com/opensearch-project/security/blob/a6859519806bd9f22d6e653ff02aa3a5fc86e3de/src/main/java/org/opensearch/security/privileges/ActionPrivileges.java#L147-L149 depending on the type of PrivilegeEvaluationContext?

derek-ho: Done in: [77ad6c3](https://github.com/opensearch-project/security/pull/4992/commits/77ad6c3722479750b86afd2984460e063defaf3b)

src/main/java/org/opensearch/security/privileges/ActionPrivileges.java
```diff
@@ -145,11 +145,29 @@ public ActionPrivileges(
...
+            return cluster.apiTokenProvidesClusterPrivilege((PermissionBasedPrivilegesEvaluationContext) context, Set.of(action), false);
+        } else {
+            // Not supported
+            return PrivilegesEvaluatorResponse.insufficient(action);
+        }
     }
 
     public PrivilegesEvaluatorResponse hasAnyClusterPrivilege(PrivilegesEvaluationContext context, Set<String> actions) {
-        return cluster.providesAnyPrivilege(context, actions, context.getMappedRoles());
+        if (context instanceof RoleBasedPrivilegesEvaluationContext) {
```
cwperks, line 159: This looks like a good contender for polymorphism.

src/main/java/org/opensearch/security/privileges/ActionPrivileges.java
```diff
@@ -901,11 +1022,69 @@ PrivilegesEvaluatorResponse providesExplicitPrivilege(
...
+                        }
+                    }
+
+                    if (indexHasAllPermissions) {
+                        break; // Found a permission that covers all actions for this index
+                    }
+                }
+
+                if (!indexHasAllPermissions) {
+                    return PrivilegesEvaluatorResponse.insufficient("Insufficient permissions for the index" + concreteIndex);
```
nibix, line 1082: This error message does not indicate which permission is missing. It also does not mention any other potential indices which also might have missing permissions. Thus, it is not really user friendly.

I would recommend to use the `CheckTable` class as it is also used by the other index privilege evaluation methods.

nibix: Additionally, the API contract for `hasIndexPrivilege()` which calls this method defines that situations where there are privileges available for sub-sets of indices are reported with a `isPartiallyOk()` response, combined with the set of available indices:

https://github.com/opensearch-project/security/blob/472f7c7979502606ccb0707908ce5f689dde6164/src/main/java/org/opensearch/security/privileges/ActionPrivileges.java#L165-L179

This is necessary for do_not_fail_on_forbidden functionality. Without that behaviour, that functionality won't work correctly for API tokens.

src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java
```diff
@@ -127,6 +128,7 @@ public final static <T> SecurityDynamicConfiguration<T> addStatics(SecurityDynam
     private final Path configPath;
     private final InternalAuthenticationBackend iab;
     private final ClusterInfoHolder cih;
+    private final ApiTokenRepository ar;
```
nibix, line 131: Nit: Despite the abbreviations that are used in the existing code, I would recommend to use non-abbreviated attribute names, as it strongly improves code readability.

src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java
```diff
@@ -377,6 +382,23 @@ private void buildAAA() {
...
+         * Challenge: false - no need to iterate through the auth domains again when ApiToken authentication failed
+         * order: -2 - prioritize the Api token authentication when it gets enabled
+         */
+        Settings apiTokenSettings = getDynamicApiTokenSettings();
+        if (!isKeyNull(apiTokenSettings, "signing_key")) {
+            final AuthDomain _ad = new AuthDomain(
+                new NoOpAuthenticationBackend(Settings.EMPTY, null),
+                new ApiTokenAuthenticator(getDynamicApiTokenSettings(), this.cih.getClusterName(), ar),
+                false,
+                -2
```
nibix, line 397: If a user has configured an auth domain with the priority -2, this will be removed in favor of the `ApiTokenAuthenticator`. This might be unexpected behavior.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenRepository.java
```diff
@@ -13,18 +13,59 @@
...
 
 public class ApiTokenRepository {
     private final ApiTokenIndexHandler apiTokenIndexHandler;
     private final SecurityTokenManager securityTokenManager;
+    private static final Logger log = LogManager.getLogger(ApiTokenRepository.class);
+
+    private final Map<String, Permissions> jtis = new ConcurrentHashMap<>();
+
+    void reloadApiTokensFromIndex() {
+        Map<String, ApiToken> tokensFromIndex = apiTokenIndexHandler.getTokenMetadatas();
```
nibix, line 39: ApiTokenIndexHandler seems to do a `actionGet`, i.e., a sync  operation. Whenever we're on transport operations, I'd strongly recommend to prefer async operations using action listeners, as otherwise there's a slight deadlock risk due to thread pool exhaustion.

src/main/java/org/opensearch/security/privileges/ActionPrivileges.java
```diff
@@ -901,11 +1022,69 @@ PrivilegesEvaluatorResponse providesExplicitPrivilege(
...
+
+            for (String concreteIndex : resolvedIndices.getAllIndices()) {
+                boolean indexHasAllPermissions = false;
+
+                // Check each index permission
+                for (ApiToken.IndexPermission indexPermission : indexPermissions) {
+                    // First check if this permission applies to this index
+                    boolean indexMatched = false;
+                    for (String pattern : indexPermission.getIndexPatterns()) {
+                        if (WildcardMatcher.from(pattern).test(concreteIndex)) {
```
nibix, line 1047: OpenSearch has the property that privileges on aliases also grant privileges on indices contained inside the alias. Similar properties hold for data streams and backing indices. I think this support is missing here.

src/main/java/org/opensearch/security/privileges/PermissionBasedPrivilegesEvaluationContext.java
```diff
@@ -0,0 +1,71 @@
...
+            + '}';
+    }
+
+    public Permissions getPermissions() {
+        return permissions;
+    }
+
+    @Override
+    public ImmutableSet<String> getMappedRoles() {
+        return ImmutableSet.of();
```
nibix, line 66: There are quite a few dependencies on mapped roles.

- There's `ProtectedIndexEvaluator` which uses the setting `plugins.security.protected_indices.roles` to define roles which can access protected indices
- `PrivilegeEvaluator` uses mapped roles to realize the Dashboards multi tenancy implementation
- User attributes exposes the mapped roles via `${user.securityRoles}`, thus they can be used in DLS statements

Is it anticipated that this functionality won't be available with API tokens?

src/integrationTest/java/org/opensearch/security/privileges/ActionPrivilegesTest.java
```diff
@@ -314,9 +383,20 @@ public void positive_full() throws Exception {
                 assertThat(result, isAllowed());
             }
 
+            @Test
+            public void apiTokens_positive_full() throws Exception {
```
nibix, line 387: Getting index privileges right is a quite challenging thing, as many non-obvious cases (aliases, data streams, do_not_fail_on_forbidden) must be considered. 

Therefore, I would recommend to add another dimension to the test parameters defined by the `IndexPrivileges` test. This dimension would produce a `PermissionBasedPrivilegesEvaluationContext` instead of a `RoleBasedPrivilegesEvaluationContext` using a central utility method. Then, you could use the existing test methods and be sure to have all cases covered.

derek-ho:
Created a tracking issue for unresolved comments, I will create a PR for those shortly: https://github.com/opensearch-project/security/issues/5088. Merging this as functionality is there and github is slightly lagging due to large number of files changed, making it hard to review. Merging into the feature branch.

PR #4998, opened 2024-12-30
[2.x Backport] Optimized Privilege Evaluation: Action privileges ONLY, with feature flag

nibix:
### Description

This implements the optimized privilege evaluation as described in https://github.com/opensearch-project/security/issues/3870 and partially backports the changes from https://github.com/opensearch-project/security/pull/4380 to the 2.x branch.

This supersedes the PR #4898 . The changes are an outcome from the requests that came up in the review of #4898. 

Particularly, this PR now contains both the new privilege evaluation implementation and the old privilege evaluation implementation. The new implementation is active by default; the old implementation can be restored by setting the opensearch.yml setting `plugins.security.privileges_evaluation.use_legacy_impl` to `true`. 

Additionally, this contains ONLY the changes related to action privileges. The DLS/FLS implementation is still the old one.

* Category: Enhancement

Performance tests indicate that the OpenSearch security layer adds a noticeable overhead to the indexing throughput of an OpenSearch cluster. The overhead may vary depending on the number of indices, the use of aliases, the number of roles and the size of the user object. The goal of these changes is to improve privilege evaluation performance and to make it less dependent on the number of indices, etc.

The additional preservation of the old implementation has been added on request in order to have an option to roll back to the old implementation in case issues are found in the new implementation. Unfortunately, this makes it necessary to double up quite a lot of code.

No significant behavioral changes in the "happy case", when privileges are present.

The undocumented config option config.dynamic.multi_rolespan_enabled is no longer evaluated. The code now behaves like it is always set to true - that is the former default. See https://github.com/opensearch-project/security/issues/4495 for details.

Some slight changes are present in error cases:
- More detailed error messages for missing privileges, showing a index/action matrix of missing privileges
- Errors in the role configuration might be reported (as error log messages) more early, directly after the configuration was applied

### Issues Resolved

Partially #3870

This is a partial backport from https://github.com/opensearch-project/security/pull/4380

### Testing

- Each new component is accompanied by its own unit test.
- The high-level functionality is validated by the existing integration tests

### Check List


nibix:
Note: This still needs some more integration tests to provide test coverage for the legacy implementation. I will add this ASAP.

kumargu:
Thanks @nibix !

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1120,18 +1121,43 @@ public Collection<Object> createComponents(
...
-            cr,
-            resolver,
-            auditLog,
-            settings,
-            privilegesInterceptor,
-            cih,
-            irr,
-            namedXContentRegistry.get()
-        );
+        if (PrivilegesEvaluator.USE_LEGACY_PRIVILEGE_EVALUATOR.get(settings)) {
```
cwperks, line 1124: Nice! Thank you for introducing this setting for rollout. Basically anything referencing this setting will be yanked out to complete https://github.com/opensearch-project/security/issues/5013 when its time to remove this setting to get the full benefits of Optimized Privilege Evaluation?

cwperks:
Primary review performed in https://github.com/opensearch-project/security/pull/4380

The changes in this PR look reasonable to me. Thank you for making sure that any tests that can are run on both code paths. I don't have any major concerns. @kumargu can you take a look as well?

kumargu:
sure. I will take a look at it.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
nibix: Yes. Bascially, the code to be yanked out exactly corresponds to these commits:

- https://github.com/opensearch-project/security/pull/4998/commits/cc1d22ad795ff6c9597f95c6f9d219edfe0b1823
- https://github.com/opensearch-project/security/pull/4998/commits/310b77ddcfe871706771418b6c9f985515ced149
- https://github.com/opensearch-project/security/pull/4998/commits/ec08e4d9fbc9ea782349d19cb136bba351ebe4b1

cwperks (approved):

DarshitChanpura:
Thank you @nibix @cwperks @kumargu  for bringing this home!

Issue #1504, opened 2021-11-23
API-Key generation API is missing


SebastianGode:
**Is your feature request related to a problem? Please describe.**
Instead of using normal user+password authentication I would like to have an never expiring API key as Elasticsearch also offers.
https://www.elastic.co/guide/en/elasticsearch/reference/7.16/security-api-create-api-key.html
Currently I have not found any way to generate such a key as the security plugin does not have such an option yet:
https://opensearch.org/docs/latest/security-plugin/access-control/api/

**Describe the solution you'd like**
It would be great to have an Elasticsearch compatible API for generating API-keys

jcannell:
Was just looking for this feature today...would love to have this.

maddisondavid:
+1 I was also searching around thinking I'd missed something in the documentation!

feli0821:
+1 also looking for this feature

jansyren:
+1 yes this would be a great feature, lots of applications writing to elastic has this support so adding it to Opensearch would be a good improvement.

djfranzelttec:
Would be an excellent feature to have.

emollusion:
+1 on this one, good to have when implementing "machine to machine" where cert auth is not possible.

bidetzz:
+1 this feature would be a great addition

nundys:
+1

derek-ho:
Requirements

Taking inspiration from the heavily thumbs upped github issue requesting API keys, we can see that API tokens would be useful in several situations, including user and service to service authentication/authorization. Once available in the OpenSearch security ecosystem, API tokens can also be used as the foundational building blocks for other features such as enforcing rate or resource usage limits to asynchronous/long running jobs. 

Based on the above the main user stories we are trying to solve would be the following:

* As a user, if I have permissions to do so, I should be able to create an API token (with a subset of my permissions?) for a specific amount of time (or unlimited) that can be used for authc/authz.
    * These API tokens should be able to perform actions that fall under the given permissions at creation time for the specified lifetime of the api token, and there should be some mechanism to track these requests (for future enhancements).
* As a user, if I have permissions to do so, I should be able to list the currently existing and revoked API tokens. 
    * It would be a nice to have to be able to list the last used date/time of tokens
* As a user, if I have permissions to do so, I should be able to delete/revoke a valid API token. After doing so, that token should immediately not be able to be used for authc/authz. 



Design 

Approach 1:  Index a Document for each Token, AllowList Cache for Usage (Preferred)

[...]

PR #5012, opened 2025-01-07
Subset of permissions check on creation

derek-ho:
### Description
This PR adds permission checks around API token creation, and cleans up misc. items from the feature branch, including authorizing API token endpoint properly and stashing the context in order to perform the API token refresh action without explicit permission to do so. This PR is dependent on #4992 to be merged first, and this PR only contains the following changes:

https://github.com/opensearch-project/security/pull/5012/files/aa506e78eb5699d2580e28d4bb0f0060971449e1..e13c055e063ae3e5ee40386a53155e5c65893b58

### Issues Resolved

### Testing

### Check List


src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
```diff
@@ -295,4 +354,99 @@ private void sendErrorResponse(RestChannel channel, RestStatus status, String er
         }
     }
 
+    /**
+     *  Validates that the user has the required permissions to create an API token (must be a subset of their own permissions)
+     * */
+    @SuppressWarnings("unchecked")
+    void validateUserPermissions(List<String> clusterPermissions, List<ApiToken.IndexPermission> indexPermissions) {
```
nibix, line 361: So, the token they are requesting is validated to be a subset of the permissions available to them in roles.yml, right?

What is supposed to happen if the permissions in roles.yml are later changed, especially if they are reduced?

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
```diff
@@ -295,4 +354,99 @@ private void sendErrorResponse(RestChannel channel, RestStatus status, String er
...
+        final User user = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+        final TransportAddress caller = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
+        final Set<String> roles = privilegesEvaluator.mapRoles(user, caller);
+
+        // Early return conditions
+        if (roles.isEmpty()) {
+            throw new OpenSearchException("User does not have any roles");
+        } else if (roles.contains("all_access")) {
```
nibix, line 369: Is this assumption safe?

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
```diff
@@ -295,4 +354,99 @@ private void sendErrorResponse(RestChannel channel, RestStatus status, String er
...
+                    RoleV7 role = (RoleV7) rolesConfiguration.getCEntry(roleName);
+                    if (role == null || role.getIndex_permissions() == null) continue;
+
+                    // Check each index permission block in the role
+                    for (RoleV7.Index indexPermission : role.getIndex_permissions()) {
+                        List<String> rolePatterns = indexPermission.getIndex_patterns();
+                        List<String> roleIndexPerms = indexPermission.getAllowed_actions();
+
+                        // Check if this role's pattern covers the requested pattern
+                        if (WildcardMatcher.from(rolePatterns).test(requestedPattern)) {
```
nibix, line 422: Assume that I have a role with the `index_pattern` `/index_[0-9]+/`. How can a `requestedPattern` look like that replicates this pattern?

At the moment, I have doubts that simple pattern matching can be used to support the full pattern semantics.

nibix: This also does not cover alias semantics.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
```diff
@@ -295,4 +354,99 @@ private void sendErrorResponse(RestChannel channel, RestStatus status, String er
...
+            }
+        }
+    }
+
+    private SecurityDynamicConfiguration<?> load(final CType<?> config) {
+        SecurityDynamicConfiguration<?> loaded = configurationRepository.getConfiguration(config);
+        return DynamicConfigFactory.addStatics(loaded);
+    }
+
+    protected void authorizeSecurityAccess(RestRequest request) throws IOException {
```
nibix, line 445: What do you think about having a dedicated transport action backing the REST action? Then, one would get access control out of the box without needing explicit checks within the action code.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
cwperks: I think the same problem exists with roles as well. If you stored roles information and a user's roles mappings changed then the tokens would need to be invalidated.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
cwperks: Agree with this though it depends on the action.

If the security admin is going to a page in the security dashboards plugin to manage the existing tokens (listing them and performing actions such as revoke) then these should follow the authorization model of other security APIs.

If regular users are allowed to request a token (with a subset of their own permissions) then that could be authorized like other cluster actions in OpenSearch.

derek-ho: I think (for now) my operating assumption would be that this is an admin-level action to issue api tokens to connect with other services programmatically. Since API tokens (currently) do not follow the new optimized privileges evaluation, and thus have higher performance penalty than other means of authc/z, and thus we should restrict this as much as possible. However, if there is community feedback/requests for this feature, then we can consider it at a later time. Since we want to authorize these APIs like the others, I added this logic.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
derek-ho: This is a good point. I added a todo on this issue: https://github.com/opensearch-project/security/issues/5088#issuecomment-2637627943

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
derek-ho: Probably not and it's unnecessary so I removed it in: [bf31791](https://github.com/opensearch-project/security/pull/5012/commits/bf317912bcc7ca22df56c2db2d787aedb454a0b3)

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
derek-ho: Added alias support in this commit: [bf31791](https://github.com/opensearch-project/security/pull/5012/commits/bf317912bcc7ca22df56c2db2d787aedb454a0b3). I also shimmed in a way to not pass in a privilegesEvaluationContext into the indexpattern class, as long as you pass in the used components out of the context. What do y'all think of this change?

cwperks: > This also does not cover alias semantics.

How would this not cover aliases? i.e. if a user has a role mapping with read access to `logs` and requests a token with read access to `logs` (assuming `logs` is an alias) won't this work?

Agree on regexp support here, @derek-ho can we throw an IllegalArgumentException if the requested permission is for a regexp?

derek-ho: I think it doesn't cover alias support for underlying indices, i.e. Users with access to `logs` alias would not be able to request api tokens for a subset of any of the underlying indices, even when they have permission to do so.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
```diff
@@ -418,8 +426,15 @@ void validateUserPermissions(List<String> clusterPermissions, List<ApiToken.Inde
...
                         List<String> roleIndexPerms = indexPermission.getAllowed_actions();
 
+                        IndexPattern indexPattern = IndexPattern.from(rolePatterns);
+
                         // Check if this role's pattern covers the requested pattern
-                        if (WildcardMatcher.from(rolePatterns).test(requestedPattern)) {
+                        if (indexPattern.matches(
+                            requestedPattern,
+                            indexNameExpressionResolver,
+                            (String template) -> WildcardMatcher.NONE,
```
nibix, line 435: I am a bit confused about returning `WildcardMatcher.NONE` here. Why do we do this?

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
nibix: I kind of feel that the handling of later privilege changes is under-specified here, as it is actually a quite challenging topic.

Changes of privileges can be caused by different effects:

- Changes in roles.yml
- Changes in roles_mapping.yml
- Change of a user's IP address (roles_mapping.yml can match IP addresses to privileges)
- Changes to the user-to-role-mapping in external authentication systems such as LDAP, OIDC or SAML


Especially the last item makes it impossible to get automatic notifications about the privileges of a user being changed. Thus, IMHO, there needs to be a conceptual review on what can be done and what can't be done and how a compensation for that could look like.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
nibix: To be honest, it is a bit difficult for me to review this without a proper scope definition. During review, I learned from you about a couple of assumptions and assumed limitations (such as not using the optimized privilege evaluation code paths, not supporting regexes, etc).  IMHO, there should be some definitions in the issue #1504 what's the initial goal to be achieved and what not. That will be then an information to guide the code review process. Otherwise, this review process will get quite inefficient. 

If the assumption is indeed that issuing tokens is an admin-level action, I do wonder why we need the matching to existing privileges at all. Won't an admin-level action just be similar to granting new privileges?

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
nibix: Regarding the regexp issue:

I think, one of the fundamental issues here is the following:

Normal users generally do not have access to the `roles.yml` configuration. Thus, they do not know how they gain access to the indices - they just know that they have access to the indices they can observe.

Thus, they need to resort to trial-and-error process to see whether they can gain access to an index or whether the roles.yml definition uses a construct that is not supported.

Obviously, this is sub-optimal from a UX point of view.

Ways around that issue - just from the top of my head:

- Just support API  token creation for admin users only. These have full access anyway, and all the checks for intersection become unnecessary  (as described in https://github.com/opensearch-project/security/pull/5012#discussion_r1943377145). 
- Just support API token creation for concrete index names, not for patterns. Then you do not have to match regexes against regexes.
- Move the process of determining the intersection of requested privileges and available privileges from the token creation to the privilege evaluation phase. Thus, one would change the check *"does roles.yml provide privileges"* into *"does roles.yml provide privileges AND does the api token request the privileges"*. This would also solve - to a certain extent - the issue of changes to roles.yml.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
cwperks: If its rolled out to admin users initially at first than I agree, but if the longer term plan is for other users to be able to request tokens then the logic to determine whether the requested permissions are a subset of the user's permissions would be needed. I was thinking to keep the logic simple and only consider: 1) `cluster_permissions` and 2) `index_permissions`. For index permissions, would it be sufficient to limit the checks to concrete indices and glob-like patterns and prohibit `regexes`?

In practice, I have not come across many roles definitions where cluster administrators use regex when define the index patterns in a role. I'm sure its used, but not prevalent from my experience.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
cwperks: > Obviously, this is sub-optimal from a UX point of view.

Can't this be handled another way though? i.e. prevent using tokens entirely if any role uses regex in `index_permissions` or skip the subset check for any role using `regexes`? 

One other way this could be handled is creating an API that a user can call that helps them comprehend what actions and index patterns they are able to select. This API could also be used to power UX for the feature.

> Just support API token creation for concrete index names, not for patterns. Then you do not have to match regexes against regexes.

What about glob-like patterns?

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
derek-ho: Since I changed the index pattern class to take in a matcher generator, here I need to pass in a function to generate a wildcardmatcher (as I don't have a PrivilegesEvaluationContext to pass in). In the Index pattern class, the wildcardmatcher is used to evaluate pattern templates/user attributes, which we don't plan to support for API tokens, and thus I passed in a WildCardMatcher that doesn't match anything to not support that use case.

src/main/java/org/opensearch/security/action/apitokens/ApiTokenAction.java
nibix: > What about glob-like patterns?

It is quite a brain-twister to think about matching patterns against patterns :sweat_smile: .. but I tought this through again, and there might be another issue. Consider: 

Consider:

- I have in roles.yml privileges for the pattern `index_1?`. This gives me privileges on `index_10`, `index_11`, etc., but not on `index_1`. 

- I now try to request an API token for the index pattern `index_1*`. The pattern `index_1?` will positively match `index_1*`. The `*` character will be bound to the `?`. However, the pattern `index_1*` is broaded than `index_1?`. It will also match `index_1`, which was not matched by `index_1?`. Thus, we would have a privilege escalation vulnerability.

Of course, this might be quite a theoretical issue where it real world applicability is questionable. Still, IMHO, security software should be also safe of such scenarios. 

Thus, I would strongly recommend against matching any patterns against patterns. It is just too fragile. 

I would still recommend the approach sketched out above:

> Move the process of determining the intersection of requested privileges and available privileges from the token creation to the privilege evaluation phase. Thus, one would change the check "does roles.yml provide privileges" into "does roles.yml provide privileges AND does the api token request the privileges". This would also solve - to a certain extent - the issue of changes to roles.yml.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -2143,7 +2157,11 @@ public Collection<SystemIndexDescriptor> getSystemIndexDescriptors(Settings sett
             ConfigConstants.OPENDISTRO_SECURITY_DEFAULT_CONFIG_INDEX
         );
         final SystemIndexDescriptor systemIndexDescriptor = new SystemIndexDescriptor(indexPattern, "Security index");
-        return Collections.singletonList(systemIndexDescriptor);
+        final SystemIndexDescriptor apiTokenSystemIndexDescriptor = new SystemIndexDescriptor(
+            ConfigConstants.OPENSEARCH_API_TOKENS_INDEX,
+            "Security API token index"
+        );
+        return List.of(systemIndexDescriptor, apiTokenSystemIndexDescriptor);
```
cwperks, line 2164: nit: we can start creating these inline.

cwperks (commented):
Overall looks good to me. Left a comment about try to unify endpoint authorization of these new endpoints and existing endpoints that subclass AbstractApiAction.

src/main/java/org/opensearch/security/privileges/IndexPattern.java
```diff
@@ -61,29 +61,33 @@ private IndexPattern(WildcardMatcher staticPattern, ImmutableList<String> patter
         this.hashCode = staticPattern.hashCode() + patternTemplates.hashCode() + dateMathExpressions.hashCode();
     }
 
-    public boolean matches(String index, PrivilegesEvaluationContext context, Map<String, IndexAbstraction> indexMetadata)
-        throws PrivilegesEvaluationException {
+    @FunctionalInterface
+    public interface MatcherGenerator {
```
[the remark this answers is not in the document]
derek-ho: Removed

Issue #5014, opened 2025-01-09
[Question] Should resource-sharing information be stored on a dedicated index or on the resource index inside the documents?


DarshitChanpura:
__Please read original proposal: #4500  in-case you need context on the feature itself.__

## Proposal 
With the recent changes in design approach of the Resource Sharing and Permissions feature, we've reached at a fork in the path where we must choose from 2 paths forward for storing resource-sharing data. Below, I've listed these two approaches in more detail:

## Approach 1 : Dedicated index

This approach emphasizes storing resource sharing information on a dedicated index `.opensearch-resource-sharing`. A document will look something like:
```
{
    "_index": ".opensearch-resource-sharing",
    "_id": "ki95HkHK8H5Gm9O5T23i",
    "_score": 1.0,
    "_source": {
        "resource_id": "ej10HpQB9Y5Gm9L4T17p",
        "source_idx": ".sample-resource-plugin"
        "created_by": {
          "user": "darshit"
        },
        "share_with": {
          "restricted": {
             "users": ["craig", "derek"],
             "roles": [],
             "backend_roles": []
          }
        }
    }
}
```
This approach present a more simplistic path towards achieving resource-level access-control with some performance overhead.

1. Pros:
   1. This approach maintains separation of concerns .
   2. With this approach no additional protections need to be put in place to ensure that the resource-sharing related data is not unintentionally exposed to the user.
 2. Cons:
    1. This approach requires a separate index call to this document when applying DLS which would add a performance overhead (significance yet to be determined)

## Approach 2: Same Index

This approach emphasizes that the resource-sharing information will be stored in the resource document to keep it closer to the resource intended to have access-control. This would look something like:

```
{
    "_index": ".sample-resource-plugin",
    "_id": "ej10HpQB9Y5Gm9L4T17p",
    "_score": 1.0,
    "_source": {
        "name": "sample_test",
        "description": "Sample test resource for ResourcePlugin",
        "attributes": {
            "a": "b"
        },
        "created_by": {
          "user": "darshit"
        },
        "share_with": {
          "restricted": {
             "users": ["craig", "derek"],
             "roles": [],
             "backend_roles": []
          }
        }
    }
}
```
Although this approach looks similar to the current model where plugins store the user information in their own index, the biggest difference is that now security-plugin will be aware of this data and has control over it.

[...]

PR #5016, opened 2025-01-10
[Feature] Introduces Centralized Resource Access Control and Sharing

DarshitChanpura:
### Description
Introduces a new authorization mechanism to control access to resources defined by plugins. 
There are 4 major components to this PR:
1. Introduces an SPI that defines the contracts for a resource sharing plugin as well as Resource Access Control plugin.
2. Introduces a client that can be used by plugins to implement resource-access-control. 
3. Introduces 4 new APIs that expose resource-sharing and access verification capabilities.
4. Adds  a sample resource plugin with tests to demonstrate the usage of this new feature.
5. Adds a feature flag to toggle the feature. Feature is _**enabled**_ by default.

* Category : New feature
  * At present, plugins have implemented in-house authorization mechanisms for a lack of centralized framework. This PR introduces a centralized way to offload resource permissions check to security plugin. 
 

### Issues Resolved
- Related to https://github.com/opensearch-project/security/issues/4500

### Testing
- automated and manual testing

### Check List


sample-resource-plugin/src/main/java/org/opensearch/sample/SampleResourcePlugin.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    @Override
+    public Collection<Class<? extends LifecycleComponent>> getGuiceServiceClasses() {
+        final List<Class<? extends LifecycleComponent>> services = new ArrayList<>(1);
+        services.add(GuiceHolder.class);
+        return services;
+    }
+
+    public static class GuiceHolder implements LifecycleComponent {
```
nibix, line 135: IMHO, the GuiceHolder pattern should be just considered as a hack. The static reference will make trouble in cases where there are several nodes in one JVM (for example in testing environments).

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
```diff
@@ -383,6 +384,7 @@ public class SearchOperationTest {
             new AuditConfiguration(true).compliance(new AuditCompliance().enabled(true))
                 .filters(new AuditFilters().enabledRest(true).enabledTransport(true))
         )
+        .nodeSettings(Map.of(OPENSEARCH_RESOURCE_SHARING_ENABLED, false))
```
nibix, line 387: Why is this necessary here?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -712,9 +775,19 @@ public void onIndexModule(IndexModule indexModule) {
...
+                    resourceAccessHandler
                 )
             );
+
+            // Listening on POST and DELETE operations in resource indices
+            ResourceSharingIndexListener resourceSharingIndexListener = ResourceSharingIndexListener.getInstance();
+            resourceSharingIndexListener.initialize(threadPool, localClient, auditLog);
+            if (RESOURCE_INDICES.contains(indexModule.getIndex().getName())) {
+                indexModule.addIndexOperationListener(resourceSharingIndexListener);
+                log.warn("Security plugin started listening to operations on resource-index {}", indexModule.getIndex().getName());
```
nibix, line 788: Why is this a warning? Warning log messages have the potential to raise concerns amongst admins which monitor the logs.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -224,6 +224,7 @@ public boolean authenticate(final SecurityRequestChannel request) {
         if (adminDns.isAdminDN(sslPrincipal)) {
             // PKI authenticated REST call
             threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_USER, new User(sslPrincipal));
+            threadContext.putPersistent(ConfigConstants.OPENDISTRO_SECURITY_USER, new User(sslPrincipal));
```
nibix, line 227: How is this change related to resource sharing?

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -0,0 +1,1209 @@
...
+     * @param boolQuery Query to execute with the request
+     */
+    private void executeSearchRequest(Set<String> resourceIds, Scroll scroll, SearchRequest searchRequest, BoolQueryBuilder boolQuery) {
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder().query(boolQuery)
+            .size(1000)
+            .fetchSource(new String[] { "resource_id" }, null);
+
+        searchRequest.source(searchSourceBuilder);
+
+        SearchResponse searchResponse = client.search(searchRequest).actionGet();
```
nibix, line 666: Sync operations in low level OpenSearch code are bound to make issues. As soon this is executed on a transport thread, this will fail.

src/main/java/org/opensearch/security/configuration/SecurityFlsDlsIndexSearcherWrapper.java
```diff
@@ -116,6 +127,37 @@ protected DirectoryReader dlsFlsWrap(final DirectoryReader reader, boolean isAdm
             log.trace("dlsFlsWrap(); index: {}; privilegeEvaluationContext: {}", index.getName(), privilegesEvaluationContext);
         }
 
+        String indexName = shardId != null ? shardId.getIndexName() : null;
+        Set<String> resourceIds;
+        if (this.isResourceSharingEnabled
+            && !Strings.isNullOrEmpty(indexName)
+            && OpenSearchSecurityPlugin.getResourceIndices().contains(indexName)) {
+            resourceIds = this.resourceAccessHandler.getAccessibleResourceIdsForCurrentUser(indexName);
```
nibix, line 135: Under the hood, this makes a sync index operation. 

We are here in an extremely low level operation. I am pretty sure that this will not work, even if it would be an async operation (which it cannot be due to the API constraints).

sample-resource-plugin/src/main/java/org/opensearch/sample/SampleResourcePlugin.java
DarshitChanpura: this is no longer needed. I will be removing it in next commit.

spi/src/main/java/org/opensearch/security/spi/resources/Resource.java
```diff
@@ -0,0 +1,33 @@
...
+import java.io.IOException;
+
+import org.opensearch.core.common.io.stream.NamedWriteable;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.xcontent.ToXContentFragment;
+
+/**
+ * Marker interface for all resources
+ */
+public abstract class Resource implements NamedWriteable, ToXContentFragment {
```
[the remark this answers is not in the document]
DarshitChanpura: The idea is that a Resource must only extend this class. The reason why it is abstract is that I want to enforce a constructor of type Resource(StreamInput in) be created.

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
DarshitChanpura: The feature is enabled by default. Which means that the resource sharing index will be created which caused `shouldGetFieldCapabilitiesForAllIndexes_positive` to fail on `containsExactlyIndices` assertion. The other option to not adding this flag is to add that assertion. However, for the sake of this test, since it is not resource-sharing related, I added a  flag to disable it.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: i followed pattern from other reference in the same class but that no longer seem to exist. I will change it to info.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
DarshitChanpura: In long term we do intend to change this to a persistent header and remove transient completely.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: Could you explain a bit more in detail?

src/main/java/org/opensearch/security/configuration/SecurityFlsDlsIndexSearcherWrapper.java
DarshitChanpura: Could you explain a bit more in detail?

src/integrationTest/java/org/opensearch/security/SearchOperationTest.java
nibix: I would guess that this is only because the resource sharing index is not marked as hidden. Marking it as hidden should fix this issue without changing the observable behavior.

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -371,6 +378,11 @@ public enum RolesMappingResolution {
     // Variable for initial admin password support
     public static final String OPENSEARCH_INITIAL_ADMIN_PASSWORD = "OPENSEARCH_INITIAL_ADMIN_PASSWORD";
 
+    // Resource sharing index
+    public static final String OPENSEARCH_RESOURCE_SHARING_INDEX = ".opensearch_resource_sharing";
```
nibix, line 382: As this is not a configuration key, I do not think that it should be in `ConfigConstants`.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
nibix: The OpenSearch request handling model is a completely async design, based on action listeners. All requests are processed by thread pools which usually have a **fixed** size. Any sync operation will cause one of these threads to wait, i.e., be unavailable while another thread will be potentially active. If there's a cyclic dependency between the thread pools (and in some cases there is), this can lead to thread pool exhaustion, i.e. a operation waiting for getting a thread it is holding locked at the very moment.

Thus, all core OpenSearch operations do not use actionGet, but action listeners.

See here for an example of such an issue: https://github.com/opensearch-project/OpenSearch/pull/8377

nibix: This is even asserted in the OpenSearch core code:

https://github.com/opensearch-project/OpenSearch/blob/f9c239d340423099699be52aa1594ef37e35005f/server/src/main/java/org/opensearch/common/util/concurrent/BaseFuture.java#L109

DarshitChanpura:
Integ tests blocked by: https://github.com/opensearch-project/security/pull/5040

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: updated the calls to be async.

src/main/java/org/opensearch/security/support/ConfigConstants.java
DarshitChanpura: done.

DarshitChanpura:
> I would guess that this is only because the resource sharing index is not marked as hidden. Marking it as hidden should fix this issue without changing the observable behavior.

@nibix should the index be marked as hidden? On other note, this PR has been updated with integration test for sample plugin and these tests run with CI as well.

DarshitChanpura:
In case anyone needs help getting started with review, here is 1 order that you can follow.
1. Start with the SPI folder.
2. Review OpenSearchSecurityPlugin to understand how the SPI is utilized to setup resource index listeners.
3. Next, review `resources` package, particularly ResourceAccessHandler and ResourceSharingIndexHandler.
4. Next review DLS/FLS related changes.
5. Once you review the main package, you can then review Sample-resource plugin to understand how a plugin might utilize it and checkout the integration tests to understand two paths, i.e. feature flag enabled and disabled.

spi/src/main/java/org/opensearch/security/spi/resources/Resource.java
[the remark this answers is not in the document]
cwperks: wdyt about the convention of calling these `SharableResource` and update all instances of `Resource` -> `SharableResource`. Resource is pretty generic and I think `SharableResource` gets to the core of what this is.

spi/src/main/java/org/opensearch/security/spi/resources/ResourceSharingExtension.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.opensearch.security.spi.resources;
+
+/**
+ * This interface should be implemented by all the plugins that define one or more resources.
+ *
+ * @opensearch.experimental
+ */
+public interface ResourceSharingExtension {
```
cwperks, line 16: nit: Similar to previous comment, wdyt about naming this `SharableResourceExtension`?

src/main/java/org/opensearch/security/DefaultObjectMapper.java
```diff
@@ -287,12 +287,26 @@ public static TypeFactory getTypeFactory() {
         return objectMapper.getTypeFactory();
     }
 
+    @SuppressWarnings("removal")
```
cwperks, line 290: Is the change in this class used anywhere?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -755,7 +796,10 @@ public Weight doCache(Weight weight, QueryCachingPolicy policy) {
 
                 @Override
                 public void onPreQueryPhase(SearchContext context) {
-                    dlsFlsValve.handleSearchContext(context, threadPool, namedXContentRegistry.get());
+                    CompletableFuture.runAsync(
```
cwperks, line 799: Is this change necessary?

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -0,0 +1,635 @@
...
+        }
+
+        // StepListener for the user’s "own" resources
+        StepListener<Set<String>> ownResourcesListener = new StepListener<>();
+
+        // StepListener for resources shared with the user’s name
+        StepListener<Set<String>> userNameResourcesListener = new StepListener<>();
+
+        // StepListener for resources shared with the user’s roles
+        StepListener<Set<String>> rolesResourcesListener = new StepListener<>();
```
cwperks, line 119: TIL about the StepListener. Nice!

I see that its marked as [`@opensearch.internal`](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/StepListener.java#L67), should we open up a PR in core to mark this public and solicit feedback?

cwperks: Looks like its used in quite a few places: https://github.com/search?q=org%3Aopensearch-project%20StepListener&type=code

I opened up a [PR in core](https://github.com/opensearch-project/OpenSearch/pull/17076) to mark this as `@opensearch.api`

nibix:
@DarshitChanpura 

> should the index be marked as hidden?

yes, any system index which is not directly relevant to the end user should be not visible by default to the user. Thus, it should be marked as hidden.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
nibix: Please see the comment in https://github.com/opensearch-project/security/pull/5016#discussion_r1916634398 ... i fear, the issue described there still holds.

While the method suggests that the call is "async", the `join()` call again blocks an OpenSearch thread which brings the risk of thread pool exhaustion and thus deadlocks. It just circumvents the assertion which is done in the `actionGet()` method.

I am sorry to say, but I am pretty confident that it is not safely possible to make search requests within low level operations like `onPreQueryPhase`  or the lucene filter leaf readers. Even if it would be possible, the performance would be miserable, as this would cause the query to be fired thousands of times for a single user search.

I fear that a completely different approach is necessary:

- Filter level DLS was designed to allow DLS for term lookup queries which seems a thing which comes closest to the current approach.
- Alternatively, the whole resource sharing information must be kept in the Java heap.

Still, I have severe doubts whether DLS is really the correct was to tackle this issue.

Reasons:

- DLS does only cover read operations. It does not cover write operations and cannot be quickly extended to cover it.
- I have doubts whether the equation "one resource is exactly one document" is really true. See also the question here: https://github.com/opensearch-project/security/issues/5014#issuecomment-2586831810

src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
```diff
@@ -374,57 +385,93 @@ public void handleSearchContext(SearchContext searchContext, ThreadPool threadPo
...
                 return;
             }
 
             DlsFlsProcessedConfig config = this.dlsFlsProcessedConfig.get();
 
-            DlsRestriction dlsRestriction = config.getDocumentPrivileges().getRestriction(privilegesEvaluationContext, index);
+            if (this.isResourceSharingEnabled && OpenSearchSecurityPlugin.getResourceIndices().contains(index)) {
+                CountDownLatch latch = new CountDownLatch(1);
+
+                threadPool.generic()
```
nibix, line 398: See comment at https://github.com/opensearch-project/security/pull/5016/files#r1924173306

src/main/java/org/opensearch/security/configuration/SecurityFlsDlsIndexSearcherWrapper.java
```diff
@@ -103,36 +113,125 @@ public SecurityFlsDlsIndexSearcherWrapper(
...
+            latch.countDown();
+        }, ex -> {
+            log.error("Failed to fetch resource IDs for index '{}': {}", indexName, ex.getMessage(), ex);
+            exceptionRef.set(ex);
+            latch.countDown();
+        }));
+
+        // 5. Block until the async call completes
         try {
+            latch.await();
```
nibix, line 164: Blocking such threads is dangerous. See comment at https://github.com/opensearch-project/security/pull/5016/files#r1924173306

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
cwperks: > DLS does only cover read operations. It does not cover write operations and cannot be quickly extended to cover it. I guess, resource sharing, however, also needs support for write operations?

I think with the approach @DarshitChanpura took in this PR, there is a separate index that security owns and keeps track of the `resource_user` (the owner of the resource) along with sharing info. Plugins reserve the right to write to their system indices, but if the index contains resources (as put forward in this feature) then behind-the-scenes security will filter the search request to only return resources pertinent to the authenticated user.

DarshitChanpura: The idea behind this approach was to prevent plugin dev errors. It was discussed as part of the PR in core. ([comment](https://github.com/opensearch-project/OpenSearch/pull/16030#issuecomment-2569237398)). By implementing a DLS approach we do not require plugin devs to make any calls manually.

@reta @nibix @cwperks I think we need to align on the path forward. DLS or not? I've already introduced a sharing api as part of this PR.

src/main/java/org/opensearch/security/resources/RecipientType.java
```diff
@@ -0,0 +1,24 @@
...
+package org.opensearch.security.resources;
+
+/**
+ * This class determines a type of recipient a resource can be shared with.
+ * An example type would be a user or a role.
+ * This class is used to determine the type of recipient a resource can be shared with.
+ *
+ * @opensearch.experimental
+ */
+public record RecipientType(String type) {
```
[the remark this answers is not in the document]
DarshitChanpura: To keep the type and the actual value separate. This can be moved to the SPI to make it easier.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -0,0 +1,635 @@
...
+     *
+     */
+    public Future<Void> getAccessibleResourceIdsForCurrentUser(String resourceIndex, ActionListener<Set<String>> listener) {
+        final UserSubjectImpl userSubject = (UserSubjectImpl) threadContext.getPersistent(
+            ConfigConstants.OPENDISTRO_SECURITY_AUTHENTICATED_USER
+        );
+        final User user = (userSubject == null) ? null : userSubject.getUser();
+
+        // If no user is authenticated, return an empty set
+        if (user == null) {
```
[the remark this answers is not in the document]
DarshitChanpura: requireNonNull may fail with NPE . In that case it might better to just remove the check.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
DarshitChanpura: I saw the PR. Looks like we decided to not mark it as public.

DarshitChanpura:
> @DarshitChanpura
> 
> > should the index be marked as hidden?
> 
> yes, any system index which is not directly relevant to the end user should be not visible by default to the user. Thus, it should be marked as hidden.

In this case the index is not technically directly relevant however when user shares the info or wants to check who has access to this resource. So I'm not sure if hidden still applies here.

src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
DarshitChanpura: https://github.com/opensearch-project/security/pull/5016#discussion_r1927298282

DarshitChanpura: i think we should align on this asap.

src/main/java/org/opensearch/security/configuration/SecurityFlsDlsIndexSearcherWrapper.java
DarshitChanpura: https://github.com/opensearch-project/security/pull/5016#discussion_r1927298282.

src/main/java/org/opensearch/security/DefaultObjectMapper.java
DarshitChanpura: its causing warnings without this suppress.

spi/src/main/java/org/opensearch/security/spi/resources/ResourceSharingExtension.java
DarshitChanpura: IMO ResourceSharingExtension is more clear.

DarshitChanpura:
I've opened a discussion issue to finalize the approach. #5062

spi/src/main/java/org/opensearch/security/spi/resources/Resource.java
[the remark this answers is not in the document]
DarshitChanpura: Resource could be anything and not necessarily shareable.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: Closing as the discussion is moved to #5062

DarshitChanpura:
@nibix @reta @willyborankin @cwperks Mind giving this another review? I've stripped off changes related to DLS from this PR.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/AbstractSampleResourcePluginTests.java
```diff
@@ -0,0 +1,71 @@
...
+import static org.opensearch.sample.utils.Constants.RESOURCE_INDEX_NAME;
+import static org.opensearch.sample.utils.Constants.SAMPLE_RESOURCE_PLUGIN_PREFIX;
+import static org.opensearch.security.dlic.rest.support.Utils.PLUGIN_RESOURCE_ROUTE_PREFIX;
+
+/**
+ * These tests run with security enabled
+ */
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class AbstractSampleResourcePluginTests {
```
nibix, line 18: Are these tests just for a demonstration on how to write tests for resource plugins or are they actually contributing to testing the functionality in the security project?

spi/src/main/java/org/opensearch/security/spi/resources/Resource.java
```diff
@@ -0,0 +1,33 @@
...
+package org.opensearch.security.spi.resources;
+
+import java.io.IOException;
+
+import org.opensearch.core.common.io.stream.NamedWriteable;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.xcontent.ToXContentFragment;
+
+/**
+ * Marker interface for all resources
```
nibix, line 18: Nit: Actually, this is not an interface but an abstract class. Personally, I'd still vote for making this an interface :)

nibix: I have seen the argument that the constructor of type Resource(StreamInput in) shall be enforced. However, i guess, Resource instance will usually be used as part of ActionRequests, right? In order to use them there, the presence of such a constructor is already implicitly necessary.

spi/src/main/java/org/opensearch/security/spi/resources/Resource.java
nibix, line 20: Why is this a `NamedWriteable` and not just a `Writeable`?

src/main/java/org/opensearch/security/DefaultObjectMapper.java
```diff
@@ -287,12 +287,26 @@ public static TypeFactory getTypeFactory() {
...
     public static Set<String> getFields(Class<?> cls) {
```
nibix, line 291: I have not seen any use of `getFields()` in the code of this PR. Wondering if the change below is necessary then.

nibix:
Following up on this, sorry I forgot about it:

> should the index be marked as hidden? 

IMHO, system indices introduced by plugins should be always hidden and protected and be made only available via REST APIs.

Main reason: OpenSearch does not have a name-spacing concept. Nothing prevents a user of OpenSearch to already have an index named `resource` or use an index pattern named `resource*` or even `*`.

A sudden appearance of such an index could break their applications.

The only way to separate such index from user index is to prefix them with `.` and mark them as hidden.

.github/workflows/ci.yml
```diff
@@ -146,12 +158,18 @@ jobs:
     - name: Checkout security
       uses: actions/checkout@v4
 
-    - name: Build and Test
+    - name: Publish SPI to Local Maven
```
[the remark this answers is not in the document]
DarshitChanpura: This was needed to compile the sample plugin and run the sample resource-plugin tests. Yes, there will be documentation to publish SPI to local maven.

build.gradle
```diff
@@ -403,6 +403,7 @@ opensearchplugin {
     name 'opensearch-security'
     description 'Provide access control related features for OpenSearch'
     classname 'org.opensearch.security.OpenSearchSecurityPlugin'
+    extendedPlugins = ['lang-painless']
```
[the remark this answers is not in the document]
DarshitChanpura: to run updateByQuery calls. Here is one usage: https://github.com/opensearch-project/security/pull/5016/files#diff-da0512f9e69c7b7f3522cb8e0baa56500969d7457ebbdc1603b57436b825963eR830

src/main/java/org/opensearch/security/DefaultObjectMapper.java
DarshitChanpura: I remember encountering compilation issues which is why I had added. I have removed it for now.

spi/src/main/java/org/opensearch/security/spi/resources/Resource.java
DarshitChanpura: I anticipate the need for using unique names for resource identification down the line.

spi/src/main/java/org/opensearch/security/spi/resources/Resource.java
DarshitChanpura: Yes, this was originally meant to be an interface. However, with non-api approach I needed a way to parse the [resource document](https://github.com/DarshitChanpura/security/blob/0e7f96ba73424788b8c84d44230ddf2a89eb7628/src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java#L1361) when sending the response back to the plugin. Which is when I had changed it to enforce StreamInput constructor. But I have since changed it to implement a [ResourceParser](https://github.com/DarshitChanpura/security/blob/8bf699365c887bb2a4db0eaf521f3c1a78c36d99/spi/src/main/java/org/opensearch/security/spi/resources/ResourceParser.java#L15) that must be registered by the resource plugin. 

And so, I've changed Resource back to being an interface and remove the need for StreamInput constructor.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/AbstractSampleResourcePluginTests.java
DarshitChanpura: They are meant to be run with CI to ensure that the resource-sharing functionality works as intended.
https://github.com/DarshitChanpura/security/blob/6d5c80fdf847985061a60ed984413f3b9fa791a9/.github/workflows/ci.yml#L127

DarshitChanpura:
> IMHO, system indices introduced by plugins should be always hidden and protected and be made only available via REST APIs.
> 
> Main reason: OpenSearch does not have a name-spacing concept. Nothing prevents a user of OpenSearch to already have an index named `resource` or use an index pattern named `resource*` or even `*`.
> 
> A sudden appearance of such an index could break their applications.
> 
> The only way to separate such index from user index is to prefix them with `.` and mark them as hidden.

The index was originally [named](https://github.com/DarshitChanpura/security/blob/47cb89dba6a84cfb7e07ae13a06cc242ebd3cc9f/src/main/java/org/opensearch/security/resources/ResourceSharingConstants.java#L15) to start with ".". I've since updated its settings to mark the index as [hidden](https://github.com/DarshitChanpura/security/blob/0e7f96ba73424788b8c84d44230ddf2a89eb7628/src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java#L100).

.github/workflows/ci.yml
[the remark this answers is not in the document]
nibix: Is this the case because the SPI artifact was not published yet?

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/AbstractSampleResourcePluginTests.java
nibix: I am not sure if it makes sense to have tests for functionality of the `security` plugin in the `sample-resource-plugin` module. IMHO, as the functionality is in `src/main`, the tests should be also in `src/test`. At least that would be intuitive for developers working on the code.

src/main/java/org/opensearch/security/DefaultObjectMapper.java
nibix: I was referring to the changes inside the method, especially the `AccessController.doPrivileged()` ... this still seems to be changed.

spi/src/main/java/org/opensearch/security/spi/resources/ResourceProvider.java
```diff
@@ -0,0 +1,33 @@
...
+public class ResourceProvider {
+    private final String resourceType;
+    private final String resourceIndexName;
+    private final ResourceParser resourceParser;
+
+    public ResourceParser getResourceParser() {
+        return resourceParser;
+    }
+
+    public String getResourceIndexName() {
```
nibix, line 20: If I see it correctly, `getResourceIndexName()` is not used anywhere. What would it be used for?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -274,6 +288,10 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
     private volatile OpensearchDynamicSetting<Boolean> transportPassiveAuthSetting;
     private volatile PasswordHasher passwordHasher;
     private volatile DlsFlsBaseContext dlsFlsBaseContext;
+    private ResourceSharingIndexManagementRepository rmr;
+    private ResourceAccessHandler resourceAccessHandler;
+    private static final Map<String, ResourceProvider> RESOURCE_PROVIDERS = new HashMap<>();
```
nibix, line 293: I usually advise against using `static` attributes in OpenSearch classes. So far, it is possible to instantiate several OpenSearch nodes and even clusters in parallel in one JVM. That's cool, as this allows to run several OpenSearch clusters in a JVM for testing. This is possible because OpenSearch managed mostly to avoid static fields and singletons. 

Having this static can lead to funny concurrency issues and side-effects.

src/test/java/org/opensearch/security/IndexIntegrationTests.java
```diff
@@ -846,19 +846,19 @@ public void testIndexResolveMinus() throws Exception {
         resc = rh.executeGetRequest("/*,-foo*/_search", encodeBasicHeader("foo_all", "nagilum"));
         assertThat(resc.getStatusCode(), is(HttpStatus.SC_FORBIDDEN));
 
-        resc = rh.executeGetRequest("/*,-*security/_search", encodeBasicHeader("foo_all", "nagilum"));
+        resc = rh.executeGetRequest("/*,-*security,-*resource*/_search", encodeBasicHeader("foo_all", "nagilum"));
```
nibix, line 849: If the resource index is now hidden, can we roll back these changes?

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/CreateResourceTransportAction.java
```diff
@@ -0,0 +1,79 @@
...
+    private void createResource(CreateResourceRequest request, ActionListener<CreateResourceResponse> listener) {
+        Resource sample = request.getResource();
+        try (XContentBuilder builder = jsonBuilder()) {
+            IndexRequest ir = nodeClient.prepareIndex(RESOURCE_INDEX_NAME)
+                .setWaitForActiveShards(1)
+                .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE)
+                .setSource(sample.toXContent(builder, ToXContent.EMPTY_PARAMS))
+                .request();
+
+            log.info("Index Request: {}", ir.toString());
```
nibix, line 69: Log messages with level `info` are visible by default. Thus, these messages should have value for an average OpenSearch admin. This might be rather a `debug` level?

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/DeleteResourceTransportAction.java
```diff
@@ -0,0 +1,76 @@
...
+            }));
+        }
+    }
+
+    private void deleteResource(DeleteResourceRequest request, ActionListener<DeleteResponse> listener) {
+        DeleteRequest deleteRequest = new DeleteRequest(RESOURCE_INDEX_NAME, request.getResourceId()).setRefreshPolicy(
+            WriteRequest.RefreshPolicy.IMMEDIATE
+        );
+
+        nodeClient.delete(deleteRequest, listener);
```
nibix, line 73: I'm a bit lost here: How is here made sure that I cannot delete resources that I do not own (or, resp., which are not shared with me).

.github/workflows/ci.yml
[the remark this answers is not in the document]
DarshitChanpura: I see that this is duplicated. It is not required for Resource Focused tests. I have removed it.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/AbstractSampleResourcePluginTests.java
nibix: Moving the tests from `sample-resource-plugin/src/integrationTest` to `src/integrationTest`. Reason: The code being tested is also located in `src/main`. Thus, the average developers looking for tests would look for them also in `src/integrationTest` or `src/test`. 

On the other hand, a sample is just a sample and should not contribute to the verification of the main components.

DarshitChanpura: This has been updated. Code being tested is now under sample-resource-plugin/src/main.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: I have moved it to a new class called ResourcePluginInfo.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/DeleteResourceTransportAction.java
DarshitChanpura: I have added a new client and modified this transport action to use that client to call verify access before deleting it.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/CreateResourceTransportAction.java
DarshitChanpura: I will got through an update the new added log statements soon.

spi/src/main/java/org/opensearch/security/spi/resources/exceptions/ResourceSharingException.java
```diff
@@ -0,0 +1,56 @@
...
+    public ResourceSharingException(StreamInput in) throws IOException {
+        super(in);
+    }
+
+    @Override
+    public RestStatus status() {
+        String message = getMessage();
+        if (message.contains("not authorized")) {
+            return RestStatus.FORBIDDEN;
+        } else if (message.contains("no authenticated")) {
```
derek-ho, line 46: not authenticated

client/src/main/java/org/opensearch/security/client/resources/ResourceSharingNodeClient.java
```diff
@@ -0,0 +1,117 @@
...
+        if (!resourceSharingEnabled) {
+            log.warn("Resource Access Control feature is disabled. Resource is not shareable.");
+            listener.onFailure(
+                new OpenSearchException(
+                    "Resource Access Control feature is disabled. Resource is not shareable.",
+                    RestStatus.NOT_IMPLEMENTED
+                )
+            );
+            return;
+        }
```
derek-ho, line 75: We can extract out this as a logical block that is run before each function

DarshitChanpura:
I will be adding concrete java doc and update any stale docs soon.

common/src/main/java/org/opensearch/security/common/resources/ResourceAccessHandler.java
```diff
@@ -0,0 +1,588 @@
...
+                    return;
+                }
+
+                boolean isAdmin = adminDNs.isAdmin(user);
+                boolean isOwner = isOwnerOfResource(document, user.getName());
+
+                if (!isAdmin && !isOwner) {
+                    LOGGER.info("User {} does not have access to delete the record {}", user.getName(), resourceId);
+                    listener.onResponse(false);
+                } else {
```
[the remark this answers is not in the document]
DarshitChanpura: block is removed since preIndex is no longer implemented

client/src/main/java/org/opensearch/security/client/resources/ResourceSharingNodeClient.java
DarshitChanpura: addressed

spi/src/main/java/org/opensearch/security/spi/resources/exceptions/ResourceSharingException.java
DarshitChanpura: not a type. it is intentional. i've updated the condition with startsWith to make it more readable.

src/test/java/org/opensearch/security/IndexIntegrationTests.java
DarshitChanpura: I tried but the tests fail without this change even though the index is marked as hidden which suggests that /* is resolving to hidden indices as well.
```json
{
  ".opensearch_resource_sharing" : {
    "settings" : {
      "index" : {
        "replication" : {
          "type" : "DOCUMENT"
        },
        "hidden" : "true",
        "number_of_shards" : "1",
        "auto_expand_replicas" : "0-all",
        "provided_name" : ".opensearch_resource_sharing",
        "creation_date" : "1741386115210",
        "number_of_replicas" : "0",
        "uuid" : "iDibox3zS8CNC0MhqrkTYQ",
        "version" : {
          "created" : "137217827"
        }
      }
    }
  }
}
```
As you can see the index is marked as hidden.

DarshitChanpura:
I have pushed the final changes with updated documentation.

I've also documented the final changes here: https://github.com/opensearch-project/security/issues/4500#issuecomment-2715832633

A detailed README is added for plugin devs here: https://github.com/opensearch-project/security/blob/4ba46a13342db50251d2cc0a5d250d6949ff68f5/RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md

src/test/java/org/opensearch/security/IndexIntegrationTests.java
DarshitChanpura: Closing this comment. As it will get lost in the long github history. Please open a new comment to continue the discussion.

DarshitChanpura:
<details>
<summary>Sample-resource-plugin integrations tests pass locally:</summary>

```
➜  security git:(resource-sharing-spi) ./gradlew clean :opensearch-sample-resource-plugin:integrationTest
=======================================
OpenSearch Build Hamster says Hello!
  Gradle Version        : 8.13
  OS Info               : Mac OS X 15.3.1 (aarch64)
  JDK Version           : 21 (Eclipse Temurin JDK)
  JAVA_HOME             : /.sdkman/candidates/java/21.0.4-tem
  Random Testing Seed   : 20F70E3D35573259
  In FIPS 140 mode      : false
=======================================

> Task :opensearch-security-common:compileJava
Note: ~/security/common/src/main/java/org/opensearch/security/common/support/Utils.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.

> Task :compileJava
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.

> Task :compileIntegrationTestJava
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

[Incubating] Problems report is available at: file:///~/security/build/reports/problems/problems-report.html

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/8.13/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD SUCCESSFUL in 42s
19 actionable tasks: 19 executed
➜  security git:(resource-sharing-spi) 

```
</details>

client/README.md
```diff
@@ -0,0 +1,233 @@
+Here's a **refined and corrected** version of your `README.md` file with improved clarity, grammar, and formatting:
+
+---
+
+# **Resource Sharing Client**
+
+This package provides a **ResourceSharing client** that resource plugins can use to **implement access control** by communicating with the **OpenSearch Security Plugin**.
```
nibix, line 7: How do plugins consume this? Is this published as a separate maven package?

This README should probably have instructions on how to consume it.

common/src/main/java/org/opensearch/security/common/package-info.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/**
+ * This package defines common classes required to implement resource access control in OpenSearch.
```
nibix, line 10: It is unclear to me why these classes need to be moved into an own module. I see that the `client` module depends on it via an `implementation` dependency, right? So, will each plugin then carry a JAR with these implementations?

As some of them feel rather security-core-ish, like `AdminDNs`, I am not sure if it is good to move these out of the security core.

common/src/main/java/org/opensearch/security/common/resources/ResourceAccessHandler.java
```diff
@@ -0,0 +1,549 @@
...
+     */
+    public void getAccessibleResourceIdsForCurrentUser(String resourceIndex, ActionListener<Set<String>> listener) {
+        final UserSubjectImpl userSubject = (UserSubjectImpl) threadContext.getPersistent(
+            ConfigConstants.OPENDISTRO_SECURITY_AUTHENTICATED_USER
+        );
+        final User user = (userSubject == null) ? null : userSubject.getUser();
+
+        // If no user is authenticated, return an empty set
+        if (user == null) {
+            LOGGER.warn("Unable to fetch user details. User is null.");
```
nibix, line 91: A resource sharing plugin indirectly cosumes the code via the `client` dependency, right?

Will this be also the case when security is disabled?

If so, what's the strategy for the case that security is disabled? Here it looks like that resources are not available at all when security is disabled (i.e., the user object is null).

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -224,7 +225,7 @@ public boolean authenticate(final SecurityRequestChannel request) {
         if (adminDns.isAdminDN(sslPrincipal)) {
             // PKI authenticated REST call
             User superuser = new User(sslPrincipal);
-            UserSubject subject = new UserSubjectImpl(threadPool, superuser);
+            UserSubject subject = new UserSubjectImpl(threadPool, new org.opensearch.security.common.user.User(sslPrincipal));
```
nibix, line 228: Why is this change necessary?

src/test/java/org/opensearch/security/IndexIntegrationTests.java
```diff
@@ -846,19 +846,16 @@ public void testIndexResolveMinus() throws Exception {
...
```
nibix, line 849: Following up on this: https://github.com/opensearch-project/security/pull/5016/files#r1992120543

I think this indicates a deeper problem that needs to be investigated.

nibix (commented):
This is a huge change. Meanwhile, it's architectural decisions like the `common` module are not entirely clear to me, which makes reviewing quite difficult.

Thus, I could only skim the changes, and added a few comments.

For a thorough review, I do not have the time ATM, though :-(

src/test/java/org/opensearch/security/IndexIntegrationTests.java
DarshitChanpura: I believe this might require updating the test framework. But I believe that could be done in a separate PR.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
DarshitChanpura: At present, the persistent header is only utilized in ResourceAccessHandler. And since we have an implementation of User in common package, we utilize that package's User here to avoid mismatch issues during runtime. This will be cleaned up once User is completely moved to common package.

common/src/main/java/org/opensearch/security/common/resources/ResourceAccessHandler.java
DarshitChanpura: When security is disabled, the call will not reach here. Instead it would fail [here in the client] (https://github.com/opensearch-project/security/blob/e6d43b606608617667e5823aaf7ba3a022f56cec/client/src/main/java/org/opensearch/security/client/resources/ResourceSharingNodeClient.java#L68) with action not found

nibix: Does that mean that resource plugins cannot be used when security is disabled?

DarshitChanpura:
As reviewers are leaning towards smaller PR, I'm closing this monolith, in lieu of individual PRs: 
- #5185 - introduces `opensearch-resource-sharing-spi`
- #5186 - introduces `opensearch-security-common` and `opensearch-security-client`
- #5187 - introduces `sample-resource-plugin`

nibix:
@DarshitChanpura 

> I understand this is a huge change, but I've tried to split it into multiple components, and highlighted the same in PR description.

Apologies, I did not want to sound harsh. Just wanted to point out that I cannot give a thorough and complete review at the moment due to the size and time constraints.

Issue #4739, opened 2024-09-13
[BUG] Opensearch does not respect "discovery.type=single-node" setting during index creation


jazzl0ver:
### Describe the bug

Opensearch was launched using the following `docker-compose.yml` (derived from https://opensearch.org/docs/latest/install-and-configure/install-opensearch/docker/#sample-docker-composeyml) on Debian 12 privileged LXC container:
```
version: '3'
services:
  opensearch: # This is also the hostname of the container within the Docker network (i.e. https://opensearch-node1/)
    image: opensearchproject/opensearch:latest # Specifying the latest available image - modify if you want a specific version
    container_name: opensearch
    environment:
      - discovery.type=single-node
      - bootstrap.memory_lock=true # Disable JVM heap memory swapping
      - plugins.security.system_indices.enabled=false
      - "OPENSEARCH_JAVA_OPTS=-Xms4096m -Xmx4096m" # Set min and max JVM heap sizes to at least 50% of system RAM
      - OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_INITIAL_ADMIN_PASSWORD}    # Sets the demo admin user password when using demo configuration, required for OpenSearch 2.12 and later
    ulimits:
      memlock:
        soft: -1 # Set memlock to unlimited (no soft or hard limit)
        hard: -1
      nofile:
        soft: 65536 # Maximum number of open files for the opensearch user - set to at least 65536
        hard: 65536
    volumes:
      - /root/data:/usr/share/opensearch/data # Creates volume called opensearch-data1 and mounts it to the container
    ports:
      - 9200:9200 # REST API
      - 9600:9600 # Performance Analyzer
    networks:
      - opensearch-net # All of the containers will join the same Docker bridge network
  opensearch-dashboards:
    image: opensearchproject/opensearch-dashboards:latest # Make sure the version of opensearch-dashboards matches the version of opensearch installed on other nodes
    container_name: opensearch-dashboards
    ports:
      - 5601:5601 # Map host port 5601 to container port 5601
    expose:
      - "5601" # Expose port 5601 for web access to OpenSearch Dashboards
    environment:
      OPENSEARCH_HOSTS: '["https://opensearch:9200"]' # Define the OpenSearch nodes that OpenSearch Dashboards will query
    networks:
      - opensearch-net

networks:
  opensearch-net:
```
In my understanding in a single node installation the replica count should be set to 0 for all indices by default, but this doesn't happen:
[...]

PR #5030, opened 2025-01-15
Improve Health Status of security-auditlog Index in Single Node Clusters

Ganesh-RB:
#### Bug Fix: Improve Health Status of security-auditlog Index in Single Node Clusters

This pull request addresses an issue where the `security-auditlog` index was showing a yellow health status in single node clusters. The problem was caused by the use of `indexrequest` default settings, which were not optimal for single node environments.

#### Changes:

* Modified the replica settings for the `security-auditlog` index to:
  ```json
  {
    "index.number_of_shards": 1,
    "index.auto_expand_replicas": "0-all"
  }


### Issues Resolved
* https://github.com/opensearch-project/security/issues/4739 

### Testing
* Created a single node cluster and verified that the security-auditlog index health changed from yellow to green.
* Confirmed that the new settings do not negatively impact multi-node clusters.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/auditlog/sink/AbstractInternalOpenSearchSink.java
+++ b/src/main/java/org/opensearch/security/auditlog/sink/AbstractInternalOpenSearchSink.java
@@ -12,8 +12,12 @@
 package org.opensearch.security.auditlog.sink;
 
 import java.io.IOException;
+import java.util.Map;
 
+import com.google.common.collect.ImmutableMap;
+import org.opensearch.ResourceAlreadyExistsException;
 import org.opensearch.action.DocWriteRequest;
+import org.opensearch.action.admin.indices.create.CreateIndexRequest;
 import org.opensearch.action.index.IndexRequestBuilder;
 import org.opensearch.action.support.WriteRequest.RefreshPolicy;
 import org.opensearch.client.Client;
@@ -51,6 +55,29 @@ public abstract class AbstractInternalOpenSearchSink extends AuditLogSink {
 
     }
 
+    private boolean indexExists(String  indexName) {
+        return clientProvider
+            .admin()
+            .indices()
+            .prepareExists(indexName)
+            .execute()
+            .actionGet()
+            .isExists();
+    }
+
+    private boolean createIndexIfAbsent(String indexName){
+        try {
+            final Map<String, Object> indexSettings = ImmutableMap.of("index.number_of_shards", 1, "index.auto_expand_replicas", "0-all");
+            final CreateIndexRequest createIndexRequest = new CreateIndexRequest(indexName).settings(indexSettings);
+            final boolean ok = clientProvider.admin().indices().create(createIndexRequest).actionGet().isAcknowledged();
+            log.info("Index {} created?: {}", indexName, ok);
+            return ok;
+        } catch (ResourceAlreadyExistsException resourceAlreadyExistsException) {
+            log.info("Index {} already exists", indexName);
+            return false;
+        }
+    }
+
     public boolean doStore(final AuditMessage msg, String indexName) {
 
         if (Boolean.parseBoolean(
@@ -60,6 +87,10 @@ public abstract class AbstractInternalOpenSearchSink extends AuditLogSink {
                 log.trace("audit log of audit log will not be executed");
             }
             return true;
+        }
+
+        if(!indexExists(indexName)){
+            createIndexIfAbsent(indexName);
         }
 
         try (StoredContext ctx = threadPool.getThreadContext().stashContext()) {


src/main/java/org/opensearch/security/auditlog/sink/AbstractInternalOpenSearchSink.java
```diff
@@ -51,6 +55,29 @@ public void close() throws IOException {
...
+            .indices()
+            .prepareExists(indexName)
+            .execute()
+            .actionGet()
+            .isExists();
+    }
+
+    private boolean createIndexIfAbsent(String indexName){
+        try {
+            final Map<String, Object> indexSettings = ImmutableMap.of("index.number_of_shards", 1, "index.auto_expand_replicas", "0-all");
```
shikharj05, line 70: Hi @Ganesh-RB , thanks for the work! 
nit: any specific reason we are setting the index replicas to auto-expand? this will create a replica on each node, I am not sure if it's required here. Can you check how settings for other indices like ``.kibana`` or any other non-critical index created by the system?

src/main/java/org/opensearch/security/auditlog/sink/AbstractInternalOpenSearchSink.java
```diff
@@ -51,6 +55,29 @@ public void close() throws IOException {
...
+        return clientProvider
+            .admin()
+            .indices()
+            .prepareExists(indexName)
+            .execute()
+            .actionGet()
+            .isExists();
+    }
+
+    private boolean createIndexIfAbsent(String indexName){
```
shikharj05, line 68: nit: this could be moved to common/utils class.

src/main/java/org/opensearch/security/auditlog/sink/AbstractInternalOpenSearchSink.java
kumargu: linking the actual issue would be useful https://github.com/opensearch-project/security/issues/4739

src/main/java/org/opensearch/security/auditlog/sink/AbstractInternalOpenSearchSink.java
```diff
@@ -51,6 +55,29 @@ public void close() throws IOException {
 
     }
 
+    private boolean indexExists(String  indexName) {
+        return clientProvider
```
cwperks, line 59: Will this do an ExistsRequest on every doStore operation? What would be the impact to performance?

It may be possible to do a lookup in the cluster metadata similar to https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java#L603-L611 to avoid an async request here.

Ganesh-RB: > Will this do an ExistsRequest on every doStore operation?

yes, with this approach. I am checking if the index exists on every new log message (every `dostore` function call).

> What would be the impact to performance?

Yes, It will have effects on performance. 
- In terms of network call: Call to index API will become twice. 
- In terms of CPU usage: Relatively small increase. Checking if index exists is lite operation as compared to adding documents.`indexrequest` also checks if index exists or not if its not it creates the index(may be with some optimized way) and then add the document.

>   It may be possible to do a lookup in the cluster metadata

Thank you for your input. I am not sure using it If we can change the setting as soon as the index is created? 

From my understanding, we can detect the yellow index by checking the metadata periodically and updating the setting. As the default naming convention for audit logs follows the pattern `"'security-auditlog-'YYYY.MM.DD" `. This structure results in the creation of a new index daily, triggered by the first log write event of each day. Implementing periodic checks can result in index being in yellow state for some time interval.

Ganesh-RB: May be If there is a way to trigger the index creation with required setting after the indexName changes, we can resolve the issue without impacting the performance.

One way is to keep a variable with previous indexName(or previous date) and create index only if it is changed. It will add some complexity in code.

src/main/java/org/opensearch/security/auditlog/sink/AbstractInternalOpenSearchSink.java
Ganesh-RB: Thank you @shikharj05 will update the settings to  
```.json
{
  "index.number_of_shards": 1,
  "index.auto_expand_replicas": "0-1"
}
```
So that it can have maximum of 1 replica.

src/main/java/org/opensearch/security/auditlog/sink/AbstractInternalOpenSearchSink.java
```diff
@@ -62,6 +88,8 @@ public boolean doStore(final AuditMessage msg, String indexName) {
             return true;
         }
 
+        createIndexIfAbsent(indexName);
+
         try (StoredContext ctx = threadPool.getThreadContext().stashContext()) {
```
shikharj05, line 93: I think we will need to move this within the try/catch post stashing context - 
1. Moving to try/catch will help handle exceptions during createIndex
2. You might need to stashContext of the thread to allow index to be created for requests that do not have access to create index. (needs testing)

src/main/java/org/opensearch/security/auditlog/sink/AbstractInternalOpenSearchSink.java
```diff
@@ -30,6 +36,8 @@ public abstract class AbstractInternalOpenSearchSink extends AuditLogSink {
     protected final Client clientProvider;
     private final ThreadPool threadPool;
     private final DocWriteRequest.OpType storeOpType;
+    private String lastUsedIndexName;
+    final Map<String, Object> indexSettings = ImmutableMap.of("index.number_of_shards", 1, "index.auto_expand_replicas", "0-1");
```
shikharj05, line 40: nit: can be static

src/main/java/org/opensearch/security/auditlog/sink/AbstractInternalOpenSearchSink.java
```diff
@@ -64,6 +88,12 @@ public boolean doStore(final AuditMessage msg, String indexName) {
 
         try (StoredContext ctx = threadPool.getThreadContext().stashContext()) {
             try {
+                boolean ok = createIndexIfAbsent(indexName);
+                if (!ok) {
+                    log.error("Server not acknowledge for creation of index {}", indexName);
```
shikharj05, line 93: nit: error message could be something like
```
 Failed to create index [index_name]
```

shikharj05 (approved):

src/main/java/org/opensearch/security/auditlog/sink/AbstractInternalOpenSearchSink.java
```diff
@@ -44,13 +52,29 @@ public AbstractInternalOpenSearchSink(
...
+        if (Objects.equals(indexName, lastUsedIndexName)) return true;
+
+        try {
+            final CreateIndexRequest createIndexRequest = new CreateIndexRequest(indexName).settings(indexSettings);
+            final boolean ok = clientProvider.admin().indices().create(createIndexRequest).actionGet().isAcknowledged();
+            lastUsedIndexName = indexName;
+            log.info("Index {} created?: {}", indexName, ok);
+            return ok;
+        } catch (ResourceAlreadyExistsException resourceAlreadyExistsException) {
+            log.info("Index {} already exists", indexName);
```
cwperks, line 73: Shouldn't `lastUsedIndexName` be set here?

What happens in the case of node reboot and the audit log index exists? `lastUsedIndexName` would be null when the node starts and shouldn't this `catch` clause be used to populate it if it is null and the exist already exists?

cwperks: FYI I experimented around with passing the `clusterService` to the internal opensearch sinks which can be used to check if an index exists without having to issue an Exists request. Take a look at the change in https://github.com/cwperks/security/pull/44. That change also shows an example of how to use the integration test framework in this repo to write an integration test for this case.

cwperks (approved):

RyanL1997 (approved):

Issue #4094, opened 2024-03-04
[BUG] `audit_request_body` missing for REST since 2.11


jackevans43:
**What is the bug?**
OpenSearch audit logs included the REST request payload in the audit logs in `audit_request_body` in 2.10 but not 2.11 (or 2.12)

**How can one reproduce the bug?**
1. Use docker-compose to bring up OpenSearch stack
2. In Dashboards, create index pattern for `security-auditlog*`
3. In Dashboard, configure audit logs to have no REST disabled categories
4. `curl -vk https://127.0.0.1:9200/twitter -u admin:admin -XPUT -d '{"settings":{"index":{"number_of_shards":3}}}' -H 'content-type: application/json`
5. `curl -vk https://127.0.0.1:9200/_search -u admin:admin -XPUT -d '{"query":{"match_all": {}}}' -H 'content-type: application/json'`

**What is the expected behavior?**
In OpenSearch Dashboards 2.10 discover view for the `security-auditlog*` index pattern, can see `audit_request_body` field set for both queries. Since 2.11 it's missing for the search (but still present for the index creation - `audit_request_layer: TRANSPORT`, but not REST).

**What is your host/environment?**
 - OS: Fedora
 - Version 39
 - Plugins

scrawfor99:
[Triage] Hi @jackevans43, thanks for filing this issue. This is the expected behavior for failed log in requests but not for other requests. Can you confirm that this is the behavior for requests other than failed login attempts?


Thanks

west117:
Hi, just to add to this - we're seeing the same behaviour as described by @jackevans43. Can confirm it is for requests other than login requests, e.g. a normal search will not yield an audited http body.

jackevans43:
@scrawfor99 Sorry for the slow reply - yes this behaviour is for successful search requests etc. Have you managed to reproduce the issue?

peternied:
If you are looking to get started on this issue, you might want to look at the code modified in this pull request [1], it should give you a good place to start.

- [1] https://github.com/opensearch-project/security/pull/691/files

cwperks:
Marking as closed as https://github.com/opensearch-project/security/pull/5071 is merged. The fix for this will be released in OpenSearch 3.0.0 and 2.19.2.

PR #5071, opened 2025-01-30
Log request body to audit logs

olkkoti:
### Description

Currently we do not get request body to audit log, as requests are already authenticated in header verifier level, which does not have access to request body.

SecurityRestFilter has the whole RestRequest object, but it does not authenticate it again. Also, it does not authorize the request if the route is not a named one. The idea here is to do logGrantedPrivileges call even if we do not authorize. This way we will get the request body to audit log.

### Issues Resolved

[[BUG] audit_request_body missing for REST since 2.11
](https://github.com/opensearch-project/security/issues/4094)

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/InternalAuditLogTest.java
+++ b/src/integrationTest/java/org/opensearch/security/InternalAuditLogTest.java
@@ -9,7 +9,7 @@
 */
 package org.opensearch.security;
 
-import java.io.IOException;
+import java.time.Duration;
 
 import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
 import org.apache.logging.log4j.LogManager;
@@ -48,13 +48,24 @@ public class InternalAuditLogTest {
         .build();
 
     @Test
-    public void testAuditLogShouldBeGreenInSingleNodeCluster() throws IOException {
+    public void testAuditLogShouldBeGreenInSingleNodeCluster() throws InterruptedException {
         try (TestRestClient client = cluster.getRestClient(USER_ADMIN)) {
             client.get(""); // demo request for insuring audit-log index is created beforehand
-            TestRestClient.HttpResponse indicesResponse = client.get("_cat/indices");
-
-            assertThat(indicesResponse.getBody(), containsString("security-auditlog"));
-            assertThat(indicesResponse.getBody(), containsString("green"));
+            int retriesLeft = 5;
+            while (retriesLeft > 0) {
+                retriesLeft = retriesLeft - 1;
+                try {
+                    TestRestClient.HttpResponse indicesResponse = client.get("_cat/indices");
+                    assertThat(indicesResponse.getBody(), containsString("security-auditlog"));
+                    assertThat(indicesResponse.getBody(), containsString("green"));
+                    break;
+                } catch (AssertionError e) {
+                    if (retriesLeft == 0) {
+                        throw e;
+                    }
+                    Thread.sleep(Duration.ofSeconds(1));
+                }
+            }
         }
     }
 }
--- a/src/main/java/org/opensearch/security/auth/BackendRegistry.java
+++ b/src/main/java/org/opensearch/security/auth/BackendRegistry.java
@@ -228,7 +228,6 @@ public class BackendRegistry {
             UserSubject subject = new UserSubjectImpl(threadPool, superuser);
             threadContext.putPersistent(ConfigConstants.OPENDISTRO_SECURITY_AUTHENTICATED_USER, subject);
             threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_USER, superuser);
-            auditLog.logSucceededLogin(sslPrincipal, true, null, request);
             return true;
         }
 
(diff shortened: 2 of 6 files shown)


cwperks (commented):
wdyt about adding `auditLog.logSucceededLogin(user.getName(), request);` before each call to `delegate.handleRequest(request, channel, client);`?

We could also remove these calls from the BackendRegistry and rely on them to be called within here. 

To enable audit logs for this category, you would need to add then `AUTHENTICATED` category as one of the enabled categories.

olkkoti:
How about like this?

olkkoti:
Hmm. Apparently at least in test cases user is allowed to be null here. I guess then we just dont do audit log?

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -393,9 +392,12 @@ public boolean authenticate(final SecurityRequestChannel request) {
             final User impersonatedUser = impersonate(request, authenticatedUser);
             final User effectiveUser = impersonatedUser == null ? authenticatedUser : impersonatedUser;
             threadPool.getThreadContext().putTransient(ConfigConstants.OPENDISTRO_SECURITY_USER, effectiveUser);
+            if (effectiveUser != authenticatedUser) {
+                threadPool.getThreadContext()
+                    .putTransient(ConfigConstants.OPENDISTRO_SECURITY_IMPERSONATION_INITIATING_USER, authenticatedUser.getName());
+            }
             UserSubject subject = new UserSubjectImpl(threadPool, effectiveUser);
             threadPool.getThreadContext().putPersistent(ConfigConstants.OPENDISTRO_SECURITY_AUTHENTICATED_USER, subject);
```
cwperks, line 400: Thank you @olkkoti . Apologies for not getting back sooner.

Curious, why not use the value from this persistent header?

In the case where impersonation is being used, looks like the current behavior is to audit log the authenticated user (not the user being impersonated - makes sense). In that case, maybe we can use the new header you are introducing in here and use it if the header is populated.

cwperks: Just triggered the PR checks to run. ~30 min for some of them to complete.

olkkoti: Ok, another problem. To preserve the original functionality, we should do logSucceededLogin right after we have fetched security user, and only if it is set. Now at least the tests that I am able to run succesfully locally pass.

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -120,6 +120,9 @@ public class ConfigConstants {
 
     public static final String OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT = OPENDISTRO_SECURITY_CONFIG_PREFIX + "user_info";
 
+    public static final String OPENDISTRO_SECURITY_IMPERSONATION_INITIATING_USER = OPENDISTRO_SECURITY_CONFIG_PREFIX
+        + "impersonation_initiating_user";
```
cwperks, line 124: wdyt about shortening this to `initiating_user`? This value will be set regardless of impersonation is being used or not and would be used for holding the identity of the original caller.

olkkoti: Yes, I renamed it.

Also, one new integration test failed on windows build. It also failed on my local machine. It looks like there is a race condition in the test case, so I added a few retries.

cwperks: Just triggered the tests to run again

cwperks:
@olkkoti I sync'ed this branch w/ main to re-run the CI checks bc there was a PR that went into main yesterday to resolve compilation errors after a breaking change was merged into the core repo.

src/integrationTest/java/org/opensearch/security/InternalAuditLogTest.java
```diff
@@ -48,13 +48,24 @@ public class InternalAuditLogTest {
...
     @Test
-    public void testAuditLogShouldBeGreenInSingleNodeCluster() throws IOException {
+    public void testAuditLogShouldBeGreenInSingleNodeCluster() throws InterruptedException {
         try (TestRestClient client = cluster.getRestClient(USER_ADMIN)) {
             client.get(""); // demo request for insuring audit-log index is created beforehand
-            TestRestClient.HttpResponse indicesResponse = client.get("_cat/indices");
-
-            assertThat(indicesResponse.getBody(), containsString("security-auditlog"));
-            assertThat(indicesResponse.getBody(), containsString("green"));
+            int retriesLeft = 5;
```
cwperks, line 54: Are you seeing flakiness in this test? I'm ok w/ putting the retry logic in, but the tests in the `integrationTest` sourceSet are already run with retries. This test is new and ideally should not have introduced flakiness. Is there a race condition between when the audit log index is created and when its available in the output of `_cat/indices`?

cwperks:
This PR LGTM, waiting for the CI checks to finish running.

cwperks (dismissed):

src/integrationTest/java/org/opensearch/security/InternalAuditLogTest.java
olkkoti: Yes. It failed on my CI run, and when I tested on my own computer it always failed without small sleep just before client.get("_cat/indices"). To be honest, I did not examine it any further.

cwperks: We'll address the flakiness in another PR to wait for the cluster to be green before calling `_cat/indices`. Thanks for putting the basic retry logic in here.

cwperks:
We should add a test for this as well to make sure the body is being audit logged as expected and will fail if this change is reverted.

olkkoti:
Ok, I added one test case for audit log request body.

cwperks:
> Hopefully fixed checkstyle scan. :)

Awesome, the bwc checks will be fixed in https://github.com/opensearch-project/security/pull/5137 which should be merging shortly

cwperks:
@olkkoti You can use `./gradlew spotlessApply` to automatically apply the spotless rules like unused imports.

olkkoti:
Yeah, I fixed checkstyle but broke spotless... Both should be ok now.

cwperks (approved):

olkkoti:
Is there something I still should do to get this merged? As far as I can tell, other things from checklist are not applicable to this PR?

willyborankin (approved):

willyborankin:
@cwperks, do we need to backport this?

cwperks:
@willyborankin yes, I think so. This is a bug fix which can be shipped in 2.19.2. 2.x is now in maintenance mode with no plan for 2.20 so backports will be limited to security fixes and bug fixes while its in maintenance mode. New features should only go in main atm.

willyborankin:
@olkkoti, could you please prepare a manual backport?

olkkoti:
> @olkkoti, could you please prepare a manual backport?

https://github.com/opensearch-project/security/pull/5199

Issue #4441, opened 2024-06-11
Once in a while receiving Unexpected exception java.util.ConcurrentModificationException when using REST API


javad87:
I'm using OS v 2.12.0, and when ingesting data with kafka-connector after some days, get this exception. I explained issue with details in follwoing link: (https://forum.opensearch.org/t/once-in-a-while-receiving-security-exception-form-opensearch-using-aiven-kafka-connector/19652) and you can refer to stack trace of exception:

[2024-06-11T03:14:28,034][ERROR][o.o.s.f.SecurityFilter   ] [os01] Unexpected exception java.util.ConcurrentModificationException
java.util.ConcurrentModificationException: null
	at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1605) ~[?:?]
	at java.base/java.util.HashMap$KeyIterator.next(HashMap.java:1628) ~[?:?]
	at java.base/java.util.Collections$UnmodifiableCollection$1.next(Collections.java:1078) ~[?:?]
	at java.base/java.util.AbstractCollection.addAll(AbstractCollection.java:337) ~[?:?]
	at java.base/java.util.HashSet.<init>(HashSet.java:121) ~[?:?]
	at org.opensearch.security.securityconf.ConfigModelV7$RoleMappingHolder.map(ConfigModelV7.java:1304) ~[opensearch-security-2.12.0.0.jar:2.12.0.0]
	at org.opensearch.security.securityconf.ConfigModelV7.mapSecurityRoles(ConfigModelV7.java:1368) ~[opensearch-security-2.12.0.0.jar:2.12.0.0]
	at org.opensearch.security.privileges.PrivilegesEvaluator.mapRoles(PrivilegesEvaluator.java:584) ~[opensearch-security-2.12.0.0.jar:2.12.0.0]
	at org.opensearch.security.privileges.PrivilegesEvaluator.evaluate(PrivilegesEvaluator.java:254) ~[opensearch-security-2.12.0.0.jar:2.12.0.0]
	at org.opensearch.security.filter.SecurityFilter.apply0(SecurityFilter.java:381) [opensearch-security-2.12.0.0.jar:2.12.0.0]
	at org.opensearch.security.filter.SecurityFilter.apply(SecurityFilter.java:165) [opensearch-security-2.12.0.0.jar:2.12.0.0]
[...]

Issue #4642, opened 2024-08-14
[BUG] java.util.ConcurrentModificationException in org.opensearch.security.securityconf.ConfigModelV7$RoleMappingHolder.map


nibix:
**What is the bug?**

In load tests performed on the most recent snapshot of the OpenSearch main branch, we sporadically encountered this exception:

```
2024-08-14T09:10:30,251][ERROR][o.o.s.f.SecurityFilter   ] [os-cluster-data-0] Unexpected exception java.util.ConcurrentModificationException
java.util.ConcurrentModificationException: null
	at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1605) ~[?:?]
	at java.base/java.util.HashMap$KeyIterator.next(HashMap.java:1628) ~[?:?]
	at java.base/java.util.Collections$UnmodifiableCollection$1.next(Collections.java:1078) ~[?:?]
	at java.base/java.util.AbstractCollection.addAll(AbstractCollection.java:337) ~[?:?]
	at java.base/java.util.HashSet.<init>(HashSet.java:121) ~[?:?]
	at org.opensearch.security.securityconf.ConfigModelV7$RoleMappingHolder.map(ConfigModelV7.java:1201) ~[opensearch-security-3.0.0.0.jar:3.0.0.0]
	at org.opensearch.security.securityconf.ConfigModelV7.mapSecurityRoles(ConfigModelV7.java:1265) ~[opensearch-security-3.0.0.0.jar:3.0.0.0]
	at org.opensearch.security.privileges.PrivilegesEvaluator.mapRoles(PrivilegesEvaluator.java:563) ~[opensearch-security-3.0.0.0.jar:3.0.0.0]
	at org.opensearch.security.privileges.PrivilegesEvaluator.createContext(PrivilegesEvaluator.java:236) ~[opensearch-security-3.0.0.0.jar:3.0.0.0]
	at org.opensearch.security.filter.SecurityFilter.apply0(SecurityFilter.java:382) [opensearch-security-3.0.0.0.jar:3.0.0.0]
```

Line 1201 in `ConfigModelV7` is this:

https://github.com/opensearch-project/security/blob/a6fb2d4880c55cddfbe5b88a3ef36c014a886ee4/src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java#L1201

So, it seems that the set returned by `user.getSecurityRoles()` was concurrently modified while the call  `new HashSet<>()` was iterating through it to create a copy.

[...]

PR #5084, opened 2025-02-04
[2.x] Avoid ConcurrentModificationException for User class fields

shikharj05:
### Description
* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation) Bug fix
Fix https://github.com/opensearch-project/security/issues/4441
Previously, since the fields were ```Collections.synchronizedSet()``` /``Collections.synchronizedMap()`` - explicit synchronization is required (see https://github.com/opensearch-project/security/issues/4642#issue-2465496301) 

By switching to ``CopyOnWriteArraySet`` and ``ConcurrentHashMap``, we get lock-free reads, safe iteration without explicit synchronization. 

However, CopyOnWriteArraySet can lead to memory overhead - should be minimal as updates/writes are expected to be lower compared to reads.

### Issues Resolved
https://github.com/opensearch-project/security/issues/4441

### Testing

tested locally.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/user/User.java
+++ b/src/main/java/org/opensearch/security/user/User.java
@@ -38,8 +38,6 @@ import java.util.Set;
 
 import com.google.common.collect.Lists;
 
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.CopyOnWriteArraySet;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.core.common.io.stream.Writeable;
@@ -74,10 +72,10 @@ public class User implements Serializable, Writeable, CustomAttributesAware {
     /**
      * roles == backend_roles
      */
-    private final Set<String> roles = new CopyOnWriteArraySet<>();
-    private final Set<String> securityRoles = new CopyOnWriteArraySet<>();
+    private final Set<String> roles = Collections.synchronizedSet(new HashSet<String>());
+    private final Set<String> securityRoles = Collections.synchronizedSet(new HashSet<String>());
     private String requestedTenant;
-    private final Map<String, String> attributes = new ConcurrentHashMap<>();
+    private Map<String, String> attributes = Collections.synchronizedMap(new HashMap<>());
     private boolean isInjected = false;
 
     public User(final StreamInput in) throws IOException {
@@ -88,10 +86,7 @@ public class User implements Serializable, Writeable, CustomAttributesAware {
         if (requestedTenant.isEmpty()) {
             requestedTenant = null;
         }
-        Map<String, String> readAttributes = in.readMap(StreamInput::readString, StreamInput::readString);
-        if (readAttributes != null) {
-            attributes.putAll(readAttributes);
-        }
+        attributes = Collections.synchronizedMap(in.readMap(StreamInput::readString, StreamInput::readString));
         securityRoles.addAll(in.readList(StreamInput::readString));
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/security/user/User.java
```diff
@@ -72,10 +74,10 @@ public class User implements Serializable, Writeable, CustomAttributesAware {
     /**
      * roles == backend_roles
      */
-    private final Set<String> roles = Collections.synchronizedSet(new HashSet<String>());
-    private final Set<String> securityRoles = Collections.synchronizedSet(new HashSet<String>());
+    private final Set<String> roles = new CopyOnWriteArraySet<>();
```
nibix, line 77: Does such a change affect the serialization output of the user object?

nibix: What is the computational complexity of operations of the `CopyOnWriteArraySet`? Looking at its code, it seems that most operations that have a *O(1)* complexity in  `HashSet` have a *O(n)*  complexity in  `CopyOnWriteArraySet`.

shikharj05: :) was waiting for BWC tests- https://github.com/opensearch-project/security/actions/runs/13130361763/job/36634124101?pr=5084#step:4:2765

```
»  Caused by: java.io.IOException: Unauthorized deserialization attempt ; java.util.concurrent.ConcurrentHashMap$Segment
```
 

CopyOnWriteArraySet is assumed safe I think because it is assignable from Collection - https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CopyOnWriteArraySet.html

For ConcurrentHashMap, we need to add to https://github.com/opensearch-project/security/blob/2.x/src/main/java/org/opensearch/security/support/SafeSerializationUtils.java

@nibix - do you think it's safe to add ConcurrentHashMap in https://github.com/opensearch-project/security/blob/2.x/src/main/java/org/opensearch/security/support/SafeSerializationUtils.java#L59 ? I don't think it should cause ser/de-serialization issues during upgrades, thoughts?

nibix: If you add these classes to `SageSerializationUtils` in (say) 2.19, any node that is still on 2.18 probably will still reject it. This will happen during rolling upgrades of a cluster.

nibix: If I see it correctly, the check is also applied when deserializing user objects. This means that any node which does not have that change will reject these classes. Thus, this will break on mixed clusters (during rolling upgrades).

nibix: But we do not have a guarantee that this remains the only point where the data structure is used.

Here's another case:

https://github.com/opensearch-project/security/blob/3c62a83df24da9fd3f4f3b0601565b62336afb49/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L352

As this code is touched for each request, this will have a noticeable CPU usage impact on users with many roles.

shikharj05: > If I see it correctly, the check is also applied when deserializing user objects. This means that any node which does not have that change will reject these classes. Thus, this will break on mixed clusters (during rolling upgrades).


makes sense. For User ``attributes``, we cannot change it to a ConcurrentHashMap. For CopyOnWriteArraySet, if there are other places (still checking) where we use ``CopyOnWriteArraySet`` directly,  O(n) will be degrading for lookups. Another option as you suggested it to add the roles to the new HashSet in a synchronized block - however this can lead to a performance penalty.

nibix: IMHO, as the sets of the `User` object are already synchronized anyway, I would consider that performance penalty as acceptable as a quick fix. Yes, it limits concurrency, but that single copy operation should be extremely fast, so in practice I do not expect a significant change.

Still, I also have the strong opinion that nothing on the user object should be synchronized or even subject to concurrent modification. It's on my todo list for this year to refactor the whole user object to be immutable and thus thread safe out of the box. That would solve any concurrency headaches on the user object.

cwperks (approved):

kumjiten (approved):

PR #5099, opened 2025-02-07
Add support for CIDR ranges in ignore_hosts setting.

shikharj05:
### Description

This change adds support to specify CIDR ranges in ignore_hosts settings. 

* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation) Enhancement
While ignore_hosts currently supports specifying IP addresses and hostnames, it would be good to support adding CIDR ranges as well. For example, see comment here- https://github.com/opensearch-project/security/issues/4927#issuecomment-2502563665
Adding support for CIDR ranges in ``ignore_hosts``

### Issues Resolved
https://github.com/opensearch-project/security/issues/4927

### Testing

### Check List
- [TODO] New functionality has been documented

Example config-

```
auth_failure_listeners:
      ip_rate_limiting:
        type: ip
        allowed_tries: 1
        time_window_seconds: 5
        block_expiry_seconds: 20
        max_blocked_clients: 100000
        max_tracked_clients: 100000
        ignore_hosts:
          - 127.0.0.0/16
```


cwperks:
Thank you for this PR @shikharj05. We will also want to update the [documentation](https://opensearch.org/docs/latest/security/configuration/api-rate-limiting/) accordingly.

src/main/java/org/opensearch/security/auth/limiting/AbstractRateLimiter.java
```diff
@@ -35,6 +39,8 @@ public abstract class AbstractRateLimiter<ClientIdType> implements AuthFailureLi
     protected final RateTracker<ClientIdType> rateTracker;
     protected final List<String> ignoreHosts;
     private WildcardMatcher ignoreHostMatcher;
+    // Cache for subnet matchers
+    private static final Map<String, SubnetUtils.SubnetInfo> SUBNET_CACHE = new ConcurrentHashMap<>();
```
cwperks, line 43: Given that the `auth_failure_listeners` are part of the dynamic config, we will need to ensure this cache is updated on a config change to the `ignore_hosts` setting for completeness. I think it would be ok to address in a follow-up, but we should capture it in an issue if its not already being handled.

See the `@Subscribe` section of [BackendRegistry](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L176-L194) for an example of how to listen to changes in the dynamic configuration.

src/main/java/org/opensearch/security/auth/AuthFailureListener.java
```diff
@@ -18,12 +18,19 @@
...
 import java.net.InetAddress;
+import java.util.List;
+
+import org.apache.commons.net.util.SubnetUtils.SubnetInfo;
 
 import org.opensearch.security.support.WildcardMatcher;
 import org.opensearch.security.user.AuthCredentials;
 
 public interface AuthFailureListener {
+    List<String> getIgnoreHosts();
```
cwperks, line 29: There may be some overlap between this and the WildcardMatcher down below. Do you think we should remove the one below on favor of a single `List<String> getIgnoreHosts();` which covers both static IPs and CIDR blocks?

src/main/java/org/opensearch/security/support/SecurityUtils.java
```diff
@@ -140,4 +149,35 @@ private static String resolveEnvVar(String envVarName, String mode, boolean bc,
...
+
+    public static boolean matchesCidrPatterns(ClientBlockRegistry<InetAddress> clientBlockRegistry, InetAddress address) {
+        if (clientBlockRegistry == null || address == null) {
+            return false;
+        }
+        AuthFailureListener authFailureListener = (AuthFailureListener) clientBlockRegistry;
+        String hostAddress = address.getHostAddress();
+        return authFailureListener.getIgnoreHosts().parallelStream().filter(pattern -> pattern.indexOf('/') != -1).anyMatch(pattern -> {
+            try {
+                SubnetInfo subnetInfo = cidrCache.computeIfAbsent(pattern, authFailureListener::getSubnetForCidr);
```
cwperks, line 175: Ignore the comment above. Looks like the dynamic config is already being handled 🙌

shikharj05:
> Thank you for this PR @shikharj05. We will also want to update the [documentation](https://opensearch.org/docs/latest/security/configuration/api-rate-limiting/) accordingly.

Ack, will create a separate PR for docs.

src/main/java/org/opensearch/security/auth/limiting/AbstractRateLimiter.java
```diff
@@ -66,6 +77,16 @@ public WildcardMatcher getIgnoreHostsMatcher() {
         return hostMatcher;
     }
 
+    @Override
+    public SubnetUtils.SubnetInfo getSubnetForCidr(String cidr) {
+        return SUBNET_CACHE.computeIfAbsent(cidr, pattern -> {
```
nibix, line 82: Why do we need a cache here? Can't the `SubnetUtils` instances be created upon initialization of an instance of this class (or a subclass)?

src/main/java/org/opensearch/security/support/SecurityUtils.java
```diff
@@ -46,6 +54,7 @@ public final class SecurityUtils {
     static final Pattern ENVBC_PATTERN = Pattern.compile("\\$\\{envbc" + ENV_PATTERN_SUFFIX);
     static final Pattern ENVBASE64_PATTERN = Pattern.compile("\\$\\{envbase64" + ENV_PATTERN_SUFFIX);
     public static Locale EN_Locale = forEN();
+    private static final Map<String, SubnetInfo> cidrCache = new ConcurrentHashMap<>();
```
nibix, line 57: Why do we need a cache on top on the cache in `AbstractRateLimiter`?

src/main/java/org/opensearch/security/support/SecurityUtils.java
```diff
@@ -140,4 +149,35 @@ private static String resolveEnvVar(String envVarName, String mode, boolean bc,
...
+        return valuesToCheck.stream().anyMatch(hostMatcher);
+    }
+
+    public static boolean matchesCidrPatterns(ClientBlockRegistry<InetAddress> clientBlockRegistry, InetAddress address) {
+        if (clientBlockRegistry == null || address == null) {
+            return false;
+        }
+        AuthFailureListener authFailureListener = (AuthFailureListener) clientBlockRegistry;
+        String hostAddress = address.getHostAddress();
+        return authFailureListener.getIgnoreHosts().parallelStream().filter(pattern -> pattern.indexOf('/') != -1).anyMatch(pattern -> {
```
nibix, line 173: Is using a parallelStream really justified here? AFAIK, it should be rather used in situations where the operation per-item operation clearly outweighs the cost of the thread management. I have doubts whether CIDR matching is really so heavy. 

See: https://www.baeldung.com/java-when-to-use-parallel-stream#1-the-overhead

src/main/java/org/opensearch/security/auth/limiting/AbstractRateLimiter.java
shikharj05: Yes, while the current implementation handles additions, removals are not handled. i.e. since there's no TTL - entries are persisted for lifetime of the node. Let me check if I can switch this with a TTL based cache or Subscribe to changes. (could also be a follow-up PR)

src/main/java/org/opensearch/security/auth/AuthFailureListener.java
shikharj05: We need both - since WildcardMatcher supports IPs, hostnames and wildcards in IPs & hostnames as well. I can move the IP matching to one method and keep hostname match in another.

src/main/java/org/opensearch/security/support/SecurityUtils.java
shikharj05: My bad, I had initially kept the cache in ``AbstractRateLimiter`` and then moved it to ``SecurityUtils``, removing the one from AbstractRateLimiter

src/main/java/org/opensearch/security/auth/limiting/AbstractRateLimiter.java
shikharj05: The cache is meant to ensure dynamic changes are also captured.

src/main/java/org/opensearch/security/support/SecurityUtils.java
shikharj05: Thanks, moved to ``streams``

src/main/java/org/opensearch/security/auth/limiting/AbstractRateLimiter.java
nibix: By dynamic changes, do you mean changes dynamic changes to the configuration (like `config.yml`, etc)?

The `AbstractRateLimiter` instances are stored in the list `ipClientBlockRegistry` in `BackendRegistry.java`. 

This list is updated upon changes of the dynamic configuration:

https://github.com/opensearch-project/security/blob/6efe00b6c1b978730dd9a0d6ef6e9d17ab2daec3/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L178-L195

Thus, I'd believe it is possible to pre-compute the `SubnetUtils` instances during creation of `AbstractRateLimiter` instances and thus being able to avoid the use of any caches at all.

src/main/java/org/opensearch/security/support/SecurityUtils.java
```diff
@@ -140,4 +149,52 @@ private static String resolveEnvVar(String envVarName, String mode, boolean bc,
...
+
+        AuthFailureListener authFailureListener = (AuthFailureListener) clientBlockRegistry;
+        final String hostAddress = address.getHostAddress();
+
+        return authFailureListener.getIgnoreHosts().stream().anyMatch(pattern -> {
+            // Handle CIDR patterns
+            if (pattern.indexOf('/') != -1) {
+                try {
+                    SubnetInfo subnetInfo = cidrCache.computeIfAbsent(pattern, authFailureListener::getSubnetForCidr);
+                    return subnetInfo.isInRange(hostAddress);
```
nibix, line 189: If I see it correctly, `SubnetInfo` only supports IPv4. If `hostAddress` is in IPv6 format, it will throw an `IllegalArgumentException`:

https://github.com/apache/commons-net/blob/aeaa37e9753ce9920c91d22493e074781ef69ff3/src/main/java/org/apache/commons/net/util/SubnetUtils.java#L395C19-L395C43

I guess, we also need to support IPv6?

src/main/java/org/opensearch/security/support/SecurityUtils.java
```diff
@@ -140,4 +149,52 @@ private static String resolveEnvVar(String envVarName, String mode, boolean bc,
...
+        if (address == null || clientBlockRegistry == null) {
+            return false;
+        }
+
+        AuthFailureListener authFailureListener = (AuthFailureListener) clientBlockRegistry;
+        final String hostAddress = address.getHostAddress();
+
+        return authFailureListener.getIgnoreHosts().stream().anyMatch(pattern -> {
+            // Handle CIDR patterns
+            if (pattern.indexOf('/') != -1) {
```
nibix, line 186: I do not think that using `/` is sufficient for identifying CIDRs. The `ignore_hosts` property supports the whole wildcard matcher feature set which also supports specifying regexes using the `/regex/` syntax. This would also hit on these regexes.

shikharj05: good catch. one of the options is to implement regex check for ipv4/ipv6 addresses to identify CIDRs specifically - will check other options and update the PR

src/main/java/org/opensearch/security/support/SecurityUtils.java
shikharj05: sure, will add support for ipv6.

src/main/java/org/opensearch/security/auth/limiting/AbstractRateLimiter.java
shikharj05: By dynamic changes, I also mean changes made using Security APIs- https://opensearch.org/docs/latest/security/access-control/api/#patch-configuration

We could still pre-compute during initialization and later use a subscriber to update instances of SubnetUtils, thoughts?

nibix: > By dynamic changes, I also mean changes made using Security APIs

I would believe that these also cause an invocation of onDynamicConfigModelChanged(), no?

src/main/java/org/opensearch/security/support/SecurityUtils.java
cwperks: You can also keep it simple and get the last index of `/` as long as `/` is not the final char in the string

src/main/java/org/opensearch/security/auth/limiting/AbstractRateLimiter.java
shikharj05: yes - made the changes to pre-compute.

src/main/java/org/opensearch/security/support/SecurityUtils.java
```diff
@@ -140,4 +150,53 @@ private static String resolveEnvVar(String envVarName, String mode, boolean bc,
...
+        if (address == null || clientBlockRegistry == null) {
+            return false;
+        }
+
+        AuthFailureListener authFailureListener = (AuthFailureListener) clientBlockRegistry;
+        final String hostAddress = address.getHostAddress();
+
+        for (String pattern : authFailureListener.getIgnoreHosts()) {
+            // Handle CIDR patterns - SubnetUtilsMatcherMap holds a SubnetInfo object for valid IPv4 patterns
+            if (authFailureListener.getSubnetUtilsMatcherMap().containsKey(pattern)) {
```
nibix, line 187: How about directly iterating through the getSubnetUtilsMatcherMap().values()? Actually, then you'd not need a `Map`, but just a `Collection`, which would simplify the implementation.

src/main/java/org/opensearch/security/support/SecurityUtils.java
```diff
@@ -140,4 +150,53 @@ private static String resolveEnvVar(String envVarName, String mode, boolean bc,
...
+                SubnetUtils.SubnetInfo subnetInfo = authFailureListener.getSubnetUtilsMatcherMap().get(pattern);
+                if (subnetInfo != null && subnetInfo.isInRange(hostAddress)) {
+                    return true;
+                }
+            } else {
+                // Handle both direct IPs and wildcard IP patterns
+                if (authFailureListener.getIgnoreHostsMatcher().test(hostAddress)) {
```
nibix, line 194: This part of the loop  does not use the looping variable `pattern`. Can't we move it out of the loop then?

src/main/java/org/opensearch/security/support/SecurityUtils.java
```diff
@@ -140,4 +150,53 @@ private static String resolveEnvVar(String envVarName, String mode, boolean bc,
...
+        return false;
+    }
+
+    // This is used to match IP addresses - for e.g. 192.168.0.1 to 192.168.0.0/16
+    public static boolean matchesIpAndCidrPatterns(ClientBlockRegistry<InetAddress> clientBlockRegistry, InetAddress address) {
+        if (address == null || clientBlockRegistry == null) {
+            return false;
+        }
+
+        AuthFailureListener authFailureListener = (AuthFailureListener) clientBlockRegistry;
```
nibix, line 182: Nit: As this method only works for `AuthFailureListener` instances, it should be probably an instance method of `AuthFailureListener`. 

Alternatively: Have you considered to create a new class that encapsulates a combination of `WildcardMatcher` and CIDR matching? 

This combination of matching is something what is not only useful for the `ignore_hosts` setting, but for quite a few functionalities. Having a reusable component for that would be nice.

src/main/java/org/opensearch/security/support/SecurityUtils.java
nibix: Just as a recommendation: I have used in other projects this library:  

https://seancfoley.github.io/IPAddress/ 

It supports both IPv4 and IPv6, is super flexible and has a clean API.

shikharj05: @nibix yes, had found this library - wasn't sure if we should use it. If it seems okay, I will try this one out.

src/main/java/org/opensearch/security/support/SecurityUtils.java
shikharj05: makes sense. will refactor the changes

willyborankin (approved):

src/main/java/org/opensearch/security/support/HostAndCidrMatcher.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import inet.ipaddr.IPAddressString;
+
+/**
+ * A utility class that performs matching of IP addresses against hostname patterns and CIDR ranges.
+ * This matcher supports both wildcard hostname patterns (e.g., *.example.com) and CIDR notation (e.g., 192.168.1.0/24).
+ */
+public class HostAndCidrMatcher {
```
nibix, line 27: Cool! :+1:

src/main/java/org/opensearch/security/support/HostAndCidrMatcher.java
```diff
@@ -0,0 +1,114 @@
...
+    private static final String IP_HOSTNAME = "ip-hostname";
```
nibix, line 28: Nit: Would it be possible to use an enum for `IP_HOSTNAME` and `IO_HOSTNAME_LOOKUP`?

src/main/java/org/opensearch/security/support/HostAndCidrMatcher.java
```diff
@@ -0,0 +1,114 @@
...
+            return false;
+        }
+
+        try {
+            IPAddressString addressString = new IPAddressString(address.getHostAddress());
+            return cidrMatchers.stream().anyMatch(cidrAddress -> cidrAddress.contains(addressString));
+        } catch (Exception e) {
+            log.warn("Failed to process IP address {}: {}", address, e.getMessage());
+            return false;
+        }
```
nibix, line 70: I am wondering about the cases which would trigger an exception here. Are these known?

If this is just covers unknown error conditions, I would recommend not to catch exceptions here and returning `false` instead. The reason is that there might be a remote chance that there is an issue hidden in the code that can be exploited by attackers which can then be used to bypass filters.

src/test/java/org/opensearch/security/support/HostAndCidrMatcherTest.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    @Test
+    public void shouldHandleInvalidCidrNotation() throws Exception {
+        matcher = new HostAndCidrMatcher(Arrays.asList("invalid/cidr/notation"));
+        InetAddress address = InetAddress.getByName(PRIVATE_CLASS_C_IP);
+        assertThat(matcher.matchesCidr(address), is(false));
+    }
+
+    @Test(expected = Exception.class)
```
nibix, line 180: What line is actually throwing the exception in the method below? My guess would be `InetAddress.getByName()`. If that's the case, then this test can be omitted, as it does not really verify behavior of the matcher itself.

nibix (commented):
Look very good! I have left a couple of further minor comments.

src/test/java/org/opensearch/security/support/HostAndCidrMatcherTest.java
```diff
@@ -0,0 +1,225 @@
...
+    public void shouldHandleMalformedIpAddresses() throws Exception {
+        matcher = new HostAndCidrMatcher(Arrays.asList(PRIVATE_CLASS_C_CIDR));
+        InetAddress address = InetAddress.getByName("invalid.ip.address");
+        matcher.matchesCidr(address);
```
nibix, line 184: What line is actually throwing the exception in the method below? My guess would be InetAddress.getByName(). If that's the case, then this test can be omitted, as it does not really verify behavior of the matcher itself.

src/main/java/org/opensearch/security/support/HostAndCidrMatcher.java
shikharj05: I don't think the exceptions are known, we can thrown an exception here to avoid hidden failures.

src/main/java/org/opensearch/security/support/HostAndCidrMatcher.java
shikharj05: ack, done.

src/main/java/org/opensearch/security/support/HostAndCidrMatcher.java
```diff
@@ -66,7 +63,7 @@ public boolean matchesCidr(InetAddress address) {
             return cidrMatchers.stream().anyMatch(cidrAddress -> cidrAddress.contains(addressString));
         } catch (Exception e) {
             log.warn("Failed to process IP address {}: {}", address, e.getMessage());
-            return false;
+            throw new RuntimeException("Invalid Address format used");
```
nibix, line 66: I'd recommend to throw the original exception e as cause. Alternatively, one could also remove the try block alltogether.

shikharj05: ack, removed the block for simplicity.

nibix (approved):

Issue #5098, opened 2025-02-07
Deprecate opendistro in API requests and responses


shikharj05:
Is your feature request related to a problem?
https://github.com/opensearch-project/security/issues/5092
 The security repo has lingering usages of the legacy terms like ``opendistro_security`` is API requests and responses, these should be replaced with ``opensearch_security``,  however legacy API contracts are not explicitly marked as deprecated. This task is to track marking opendistro API request/responses as deprecated.

What solution would you like?
Mark opendistro API requests/responses as deprecated, removal in a next major version. Replace with opensearch - for e.g. opendistro_security_roles in Create User - https://opensearch.org/docs/latest/security/access-control/api/#create-user

Issue #5092, opened 2025-02-05
Deprecate legacy language from the security plugin repo in 3.0.0 for removal in 4.0.0


cwperks:
**Is your feature request related to a problem?**

The security repo has lingering usages of the terms `opendistro` and `kibana`. This issue will be a tracking issue about removing all of the legacy usages and replacing them with `opensearch` and `opensearch_dashboards` respectively.

Some example usages are:

- The security index (`.opendistro_security`)
- static/reserved roles like `kibana_user` and `kibanaserver`
- The `.kibana` indices in OpenSearch dashboards
- `config.dynamic.kibana` portion of the securityconfig which includes `kibana_url`

shikharj05:
Created Subtasks
*  #5097
*  #5098
*  #4117

PR #5113, opened 2025-02-18
Deprecate opendistro_security_roles and add opensearch_security_roles

shikharj05:
Deprecate opendistro_security_roles and add opensearch_security_roles for internal-user APIs.

### Description
* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation) Maintenance
https://github.com/opensearch-project/security/issues/5092
```
curl -XPUT -k -u admin:<password> "https://localhost:9200/_plugins/_security/api/internalusers/testuser" -H 'Content-Type: application/json' -d'
{
  "password": "kirkpass!!",
  "opensearch_security_roles": ["manage_snapshots"],
  "backend_roles": ["role 1", "role 2"],
  "attributes": {
    "attribute1": "value1",
    "attribute2": "value2"
  }
}
'

curl -k -u admin:<password> 'https://localhost:9200/_plugins/_security/api/user/testuser?pretty'                                                
{
  "testuser" : {
    "hash" : "",
    "reserved" : false,
    "hidden" : false,
    "backend_roles" : [
      "role 1",
      "role 2"
    ],
    "attributes" : {
      "attribute1" : "value1",
      "attribute2" : "value2"
    },
    "opendistro_security_roles" : [
      "manage_snapshots"
    ],
    "opensearch_security_roles" : [
      "manage_snapshots"
    ],
    "static" : false
  }
}



curl -k -u admin:<password> 'https://localhost:9200/_plugins/_security/api/internalusers/testuser?pretty' -X PATCH -d '[{"op": "replace", "path": "/opensearch_security_roles", "value": ["notebooks_full_access","all_access"]}]' -H 'content-type: application/json'

{
  "status" : "OK",
  "message" : "'testuser' updated."
}


curl -k -u admin:<password> 'https://localhost:9200/_plugins/_security/api/user/testuser?pretty'
{
  "testuser" : {
    "hash" : "",
    "reserved" : false,
    "hidden" : false,
    "backend_roles" : [
      "role 1",
      "role 2"
    ],
    "attributes" : {
      "attribute1" : "value1",
      "attribute2" : "value2"
    },
    "opendistro_security_roles" : [
      "notebooks_full_access",
      "all_access"
    ],
    "opensearch_security_roles" : [
      "notebooks_full_access",
      "all_access"
    ],
    "static" : false
  }
}

curl -k -u admin:<password> 'https://localhost:9200/_plugins/_security/api/internalusers/testuser?pretty' -X PATCH -d '[{"op": "replace", "path": "/opendistro_security_roles", "value": ["notebooks_full_access"]}]' -H 'content-type: application/json'     

{
  "status" : "OK",
  "message" : "'testuser' updated."
}


curl -k -u admin:<password> 'https://localhost:9200/_opendistro/_security/api/user/testuser?pretty'                                             
{
  "testuser" : {
    "hash" : "",
    "reserved" : false,
    "hidden" : false,
    "backend_roles" : [
      "role 1",
      "role 2"
    ],
    "attributes" : {
      "attribute1" : "value1",
      "attribute2" : "value2"
    },
    "opendistro_security_roles" : [
      "manage_snapshots"
    ],
    "opensearch_security_roles" : [
      "manage_snapshots"
    ],
    "static" : false
  }
}


```

### Issues Resolved
https://github.com/opensearch-project/security/issues/5098

### Testing
Manual testing results above.

Logs
```
[2025-02-18T15:03:11,188][WARN ][org.opensearch.security.dlic.rest.api.InternalUsersApiAction] The field 'opendistro_security_roles' is deprecated and will be removed in a future release. Please use 'opensearch_security_roles' instead.
```

### Check List
- [TODO] New functionality has been documented

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/api/InternalUsersRestApiIntegrationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/api/InternalUsersRestApiIntegrationTest.java
@@ -43,6 +43,8 @@ import static org.hamcrest.Matchers.containsString;
 import static org.opensearch.security.api.PatchPayloadHelper.addOp;
 import static org.opensearch.security.api.PatchPayloadHelper.patch;
 import static org.opensearch.security.api.PatchPayloadHelper.replaceOp;
+import static org.opensearch.security.dlic.rest.api.InternalUsersApiAction.OPENDISTRO_SECURITY_ROLES;
+import static org.opensearch.security.dlic.rest.api.InternalUsersApiAction.OPENSEARCH_SECURITY_ROLES;
 import static org.opensearch.security.dlic.rest.api.InternalUsersApiAction.RESTRICTED_FROM_USERNAME;
 
 public class InternalUsersRestApiIntegrationTest extends AbstractConfigEntityApiIntegrationTest {
@@ -149,11 +151,15 @@ public class InternalUsersRestApiIntegrationTest extends AbstractConfigEntityApi
                 builder.field("attributes", attributes);
             }
             if (securityRoles != null) {
-                builder.field("opendistro_security_roles");
+                builder.field(getRoleField());
                 securityRoles.toXContent(builder, params);
             }
             return builder.endObject();
         };
+    }
+
+    private static String getRoleField() {
+        return randomFrom(List.of(OPENDISTRO_SECURITY_ROLES, OPENSEARCH_SECURITY_ROLES));
     }
 
     static ToXContentObject defaultServiceUser() {
--- a/src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
+++ b/src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
@@ -18,6 +18,8 @@ import java.util.Map;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
@@ -51,7 +53,16 @@ import static org.opensearch.security.dlic.rest.support.Utils.addRoutesPrefix;
 
 public class InternalUsersApiAction extends AbstractApiAction {
 
+    private static final Logger log = LogManager.getLogger(InternalUsersApiAction.class);
     private final PasswordHasher passwordHasher;
+
+    /**
+     * @deprecated Use opensearch_security_roles instead. This field will be removed in a future release.
+     */
+    @Deprecated
+    public static final String OPENDISTRO_SECURITY_ROLES = "opendistro_security_roles";
+
+    public static final String OPENSEARCH_SECURITY_ROLES = "opensearch_security_roles";
 
     @Override
     protected void consumeParameters(final RestRequest request) {
@@ -183,8 +194,7 @@ public class InternalUsersApiAction extends AbstractApiAction {
             rolesConfiguration -> loadConfiguration(CType.ROLESMAPPING, false, false).map(roleMappingsConfiguration -> {
                 final var contentAsNode = (ObjectNode) securityConfiguration.requestContent();
                 final var securityJsonNode = new SecurityJsonNode(contentAsNode);
-                var securityRoles = securityJsonNode.get("opendistro_security_roles").asList();
-                securityRoles = securityRoles == null ? List.of() : securityRoles;
+                var securityRoles = getRoles(securityJsonNode);
                 final var rolesValid = endpointValidator.validateRoles(securityRoles, rolesConfiguration);
                 if (!rolesValid.isValid()) {
                     return ValidationResult.error(rolesValid.status(), rolesValid.errorMessage());
@@ -200,6 +210,22 @@ public class InternalUsersApiAction extends AbstractApiAction {
                 return ValidationResult.success(securityConfiguration);
             })
         );
+    }
+
+    private List<String> getRoles(SecurityJsonNode content) {
+        List<String> roles = content.get(OPENSEARCH_SECURITY_ROLES).asList();
+        // Fall back to deprecated field if new field is not present
+        if (roles == null) {
+            roles = content.get(OPENDISTRO_SECURITY_ROLES).asList();
+            if (roles != null) {
+                log.warn(
+                    "The field '{}' is deprecated and will be removed in a future release. Please use '{}' instead.",
+                    OPENDISTRO_SECURITY_ROLES,
+                    OPENSEARCH_SECURITY_ROLES
+                );
+            }
+        }
+        return roles != null ? roles : List.of();
     }
 
     ValidationResult<SecurityConfiguration> createOrUpdateAccount(
@@ -309,7 +335,8 @@ public class InternalUsersApiAction extends AbstractApiAction {
                         return allowedKeys.put("backend_roles", DataType.ARRAY)
                             .put("attributes", DataType.OBJECT)
                             .put("description", DataType.STRING)
-                            .put("opendistro_security_roles", DataType.ARRAY)
+                            .put(OPENDISTRO_SECURITY_ROLES, DataType.ARRAY)
+                            .put(OPENSEARCH_SECURITY_ROLES, DataType.ARRAY)
                             .put("hash", DataType.STRING)
                             .put("password", DataType.STRING)
                             .build();
--- a/src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java
+++ b/src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java
@@ -359,12 +359,19 @@ public class DynamicConfigFactory implements Initializable, ConfigurationChangeL
 
             // Security roles should only contain roles that exist in the roles dynamic config.
             // We should filter out any roles that have hidden rolesmapping.
-            return tmp == null
-                ? ImmutableList.of()
-                : tmp.getOpendistro_security_roles()
+            if (tmp == null) {
+                return ImmutableList.of();
+            } else if (!tmp.getOpensearch_security_roles().isEmpty()) {
+                return tmp.getOpensearch_security_roles()
                     .stream()
                     .filter(role -> !isRolesMappingHidden(role) && rolesV7SecurityDynamicConfiguration.exists(role))
                     .collect(ImmutableList.toImmutableList());
+            } else {
+                return tmp.getOpendistro_security_roles()
+                    .stream()
+                    .filter(role -> !isRolesMappingHidden(role) && rolesV7SecurityDynamicConfiguration.exists(role))
+                    .collect(ImmutableList.toImmutableList());
+            }
         }
 
         // Remove any hidden rolesmapping from the security roles
--- a/src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
+++ b/src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
@@ -51,6 +51,7 @@ public class InternalUserV7 implements Hideable, Hashed, StaticDefinable {
     private Map<String, String> attributes = Collections.emptyMap();
     private String description;
     private List<String> opendistro_security_roles = Collections.emptyList();
+    private List<String> opensearch_security_roles = Collections.emptyList();
 
     private InternalUserV7(String hash, boolean reserved, boolean hidden, List<String> backend_roles, Map<String, String> attributes) {
         super();
@@ -117,6 +118,16 @@ public class InternalUserV7 implements Hideable, Hashed, StaticDefinable {
 
     public void setOpendistro_security_roles(List<String> opendistro_security_roles) {
         this.opendistro_security_roles = opendistro_security_roles;
+        this.opensearch_security_roles = opendistro_security_roles;
+    }
+
+    public List<String> getOpensearch_security_roles() {
+        return opensearch_security_roles;
+    }
+
+    public void setOpensearch_security_roles(List<String> opensearch_security_roles) {
+        this.opendistro_security_roles = opensearch_security_roles;
+        this.opensearch_security_roles = opensearch_security_roles;
     }
 
     public Map<String, String> getAttributes() {
--- a/config/internal_users.yml
+++ b/config/internal_users.yml
@@ -20,7 +20,7 @@ admin:
 anomalyadmin:
   hash: "$2y$12$TRwAAJgnNo67w3rVUz4FIeLx9Dy/llB79zf9I15CKJ9vkM4ZzAd3."
   reserved: false
-  opendistro_security_roles:
+  opensearch_security_roles:
   - "anomaly_full_access"
   description: "Demo anomaly admin user, using internal role"
 


src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
```diff
@@ -51,6 +51,7 @@ public class InternalUserV7 implements Hideable, Hashed, StaticDefinable {
     private Map<String, String> attributes = Collections.emptyMap();
     private String description;
     private List<String> opendistro_security_roles = Collections.emptyList();
+    private List<String> opensearch_security_roles = Collections.emptyList();
```
nibix, line 54: One high-level remark: 

IMHO, config property names should be as self-descriptive as possible. However, neither `opendistro_security_roles` nor `opensearch_security_roles` are really self-descriptive.

The function of this property is as follows: Roles which are specified via `role` property first go through role-mapping so that they are mapped onto the effective roles of the user at runtime. Roles specified via `opendistro_security_roles`  do not go through role mapping, but get effective immediately.

Can we find a property name that is (kind of) self-descriptive?

shikharj05: what do you think about ``opensearch_security_inline_roles``/``opensearch_security_static_roles``/``opensearch_security_direct_roles`` or any other name :) ?

nibix: I do not think, the `opensearch_security` prefix is necessary, as the other attribute is also just called `roles`.

So, that would be `inline_roles` or `static_roles` or `direct_roles`. I would not recommend `static_roles`, as this could be confused with the static role configuration in https://github.com/opensearch-project/security/blob/main/src/main/resources/static_config/static_roles.yml .

I could also imagine `directly_effective_roles`, that would be most telling.

@cwperks @DarshitChanpura @derek-ho wdyt?

cwperks: Personally, I think we should consider removing this attribute from the InternalUser object and only do roles mappings with RolesMappings.

What is the advantage of having multiple ways to map roles to usernames?

shikharj05: > What is the advantage of having multiple ways to map roles to usernames?

I would consider these roles as default permissions that a user has when you create such an user. Also, it's simpler to setup for small number of users. 

Using role mappings is better for more centralized and flexible management, better for managing similar users, etc. 

I think it should be okay to keep this feature for it's simplicity (single API setup/configuration) - also, maybe it will prove helpful with service/system accounts feature in the plugin - i.e. using inline_roles or such directly_effective_roles for accounts.

cwperks: > @cwperks should we create an issue to track the long-term re-evaluation of role mappings.

Sounds like we are going to keep supporting inline roles based on this conversation. I think we can add an issue for migration assistance here. In past versions, there was a Migrate API to help migrating the security config so should we start implementing one to account for these upcoming changes as well?

src/integrationTest/java/org/opensearch/security/api/InternalUsersRestApiIntegrationTest.java
```diff
@@ -149,13 +151,17 @@ static ToXContentObject internalUser(
...
-                builder.field("opendistro_security_roles");
+                builder.field(getRoleField());
                 securityRoles.toXContent(builder, params);
             }
             return builder.endObject();
         };
     }
 
+    private static String getRoleField() {
+        return randomFrom(List.of(OPENDISTRO_SECURITY_ROLES, DIRECT_SECURITY_ROLES));
```
nibix, line 162: I see that there are also other tests which use `randomFrom`. However, I am kind of skeptical of using random sources in tests, especially if the random space is only two elements. This has high potential to lead to test flakyness.

Shoudlnt we just test both cases?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -228,6 +241,37 @@ ValidationResult<SecurityConfiguration> validateSecurityRoles(final SecurityConf
...
+
+        // Add OpenSearch roles if present
+        if (openSearchRoles != null) {
+            mergedRoles.addAll(openSearchRoles);
+        }
+
+        // Add OpenDistro roles if present
+        if (openDistroRoles != null) {
+            mergedRoles.addAll(openDistroRoles);
+            log.warn(
```
nibix, line 265: Indirectly related: What's the migration path towards having only the new attribute?

This will only log warnings when the REST API is used. No warnings will be logged for usage of the securityadmin tool or just the existing data in the config index.

Will there be an additional change in the future to address existing data?

src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java
```diff
@@ -359,12 +359,19 @@ public List<String> getSecurityRoles(String user) {
 
             // Security roles should only contain roles that exist in the roles dynamic config.
             // We should filter out any roles that have hidden rolesmapping.
-            return tmp == null
-                ? ImmutableList.of()
-                : tmp.getOpendistro_security_roles()
+            if (tmp == null) {
+                return ImmutableList.of();
+            } else if (!tmp.getDirect_security_roles().isEmpty()) {
```
nibix, line 364: This will silently ignore  the old attribute - even if still exists. This can lead to confusion.

src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
```diff
@@ -117,6 +118,16 @@ public List<String> getOpendistro_security_roles() {
...
     public void setOpendistro_security_roles(List<String> opendistro_security_roles) {
         this.opendistro_security_roles = opendistro_security_roles;
+        this.direct_security_roles = opendistro_security_roles;
+    }
+
+    public List<String> getDirect_security_roles() {
+        return direct_security_roles;
+    }
+
+    public void setDirect_security_roles(List<String> direct_security_roles) {
```
nibix, line 128: How does the serialized JSON produced from this bean look like? I reckon, it will contain both attributes, even if one is unused, correct?

src/integrationTest/java/org/opensearch/security/api/InternalUsersRestApiIntegrationTest.java
shikharj05: I am okay with testing both cases for this change.

src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java
shikharj05: will add a log here

cwperks (approved):

src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
nibix: I am generally wondering how clients consuming that API should handle this.

This actually already applies to the security Dashboards plugin, which uses this API.

Ideally, an API could just switch the the new attribute and use that. However, in this case, all API clients would also need to have a kind of detection mechanism to see which attribute is the one that is actually in use.

This is especially critical for the PATCH API, which such clients cannot use anymore without making a GET in advance to find out what attribute would be in use.

Generally, that seems to be not a good design of an API.

shikharj05: > I am generally wondering how clients consuming that API should handle this. This actually already applies to the security Dashboards plugin, which uses this API.

Since both the fields are maintained and updated on changes, clients using existing field will not break. As the existing field is being deprecated, guidance to use new field will be documented. As both the fields are available, I don't think a detection mechanism is required. I agree it would have been cleaner to switch to the new attribute, however I am not sure if we can do it without breaking existing clients. 


See example below-
```
curl -k -u admin:<password> 'https://localhost:9200/_plugins/_security/api/internalusers/testuser?pretty' -X PATCH -d '[{"op": "replace", "path": "/opendistro_security_roles", "value": ["notebooks_full_access"]}]' -H 'content-type: application/json'     

{
  "status" : "OK",
  "message" : "'testuser' updated."
}


curl -k -u admin:<password> 'https://localhost:9200/_opendistro/_security/api/user/testuser?pretty'                                             
{
  "testuser" : {
    "hash" : "",
    "reserved" : false,
    "hidden" : false,
    "backend_roles" : [
      "role 1",
      "role 2"
    ],
    "attributes" : {
      "attribute1" : "value1",
      "attribute2" : "value2"
    },
    "opendistro_security_roles" : [
      "manage_snapshots"
    ],
    "opensearch_security_roles" : [
      "manage_snapshots"
    ],
    "static" : false
  }
}
```

src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
```diff
@@ -116,7 +121,18 @@ public List<String> getOpendistro_security_roles() {
     }
 
     public void setOpendistro_security_roles(List<String> opendistro_security_roles) {
+        log.warn("Deprecated configuration opendistro_security_roles set. Kindly use direct_security_roles instead.");
         this.opendistro_security_roles = opendistro_security_roles;
+        this.direct_security_roles = opendistro_security_roles;
+    }
```
nibix, line 127: Sorry for the delay, I was a bit busy with other stuff. I have though about this log message again.

This method will be called by Jackson when it de-serializes the internal user JSON. It will be called for each user object.

As  in the security config index, all user records are stored as one big JSON in a single document, Jackson needs to de-serialize all  user objects, even if it only needs one. This will have the effect that whenever the user config is being loaded from the index, you'll get this log message repeated for each user stored in that index. On clusters with many users, these will be quite a lot repeated log messages at once.

This would happen in these cases:

- On each node: On cluster startup, when the index is initially read
- On each node: Whenever the configuration is updated
- On a single node: Whenever internalusers REST API is used (be it `GET` or `PUT` or `PATCH`), the index is freshly read initially, so these logs would also appear for all users, even if the REST API only addresses a single user.

I think such repeated log messages won't be helpful to the user. I would guess that we need a different solution to notify the user about the deprecated attribute.

shikharj05: No worries. Thanks for the detailed explanation. My understanding was that the log lines would be added only during initial read and during updates (limited to a specific user);  And that once the configuration is in memory, it won't be fetched from the index unless a flush was called.

I will remove the log line to avoid excessive logging. 

> I would guess that we need a different solution to notify the user about the deprecated attribute.

Do you think documentation should be enough? something like- https://opensearch.org/docs/latest/breaking-changes/#deprecate-non-inclusive-terms

shikharj05: > Which one is shown is determined by the last update: If the client used the new attribute name in the update, the new one is also used when retrieving the document. If the client used the old attribute name in the update, the old one is also used when retrieving the document.


nit: This might get problematic with number of clients? For e.g. client A using old attribute v/s client B using the new attribute. 

I am more aligned with option 1 - i.e. track with a major update if required. Alternatively, we could introduce new APIs to deprecate old functionality. i.e. new APIs return new attribute names, existing ones use old one. (would have been cleaner if this was done when APIs moved to _plugins/_security from _opendistro API paths)

nibix: > This might get problematic with number of clients? For e.g. client A using old attribute v/s client B using the new attribute.

Yes, good point.

shikharj05: @cwperks @nibix - how do you think we should proceed here?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
DarshitChanpura: Can we do a clean break with a deprecation message that clearly states that this will be remove in 3.0 GA?

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
```diff
@@ -335,7 +379,8 @@ public Map<String, RequestContentValidator.DataType> allowedKeys() {
                         return allowedKeys.put("backend_roles", DataType.ARRAY)
                             .put("attributes", DataType.OBJECT)
                             .put("description", DataType.STRING)
-                            .put("opendistro_security_roles", DataType.ARRAY)
+                            .put(OPENDISTRO_SECURITY_ROLES, DataType.ARRAY)
```
DarshitChanpura, line 382: Do we want either to be set or both should be allowed for now?

What is the plan to remove this deprecated key from allowed_keys?

src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java
```diff
@@ -357,14 +359,35 @@ public String getHash(String user) {
...
-                : tmp.getOpendistro_security_roles()
+            if (tmp.getDirect_security_roles() != null && !tmp.getDirect_security_roles().isEmpty()) {
+                return tmp.getDirect_security_roles()
                     .stream()
                     .filter(role -> !isRolesMappingHidden(role) && rolesV7SecurityDynamicConfiguration.exists(role))
                     .collect(ImmutableList.toImmutableList());
+            }
+
+            if (tmp.getOpendistro_security_roles() != null && !tmp.getOpendistro_security_roles().isEmpty()) {
+                return tmp.getOpendistro_security_roles()
```
DarshitChanpura, line 385: nit: line 368 logger can be moved here.

src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
DarshitChanpura: Can we show both for now with deprecation documented on the website? We do set the direct_security_roles atttribute even when opendistro_security_roles is set, which should not affect any roles being missed in the new key when populating the deprecated key. Apologies if I didn't understand the problem correctly.

src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
DarshitChanpura: are `opensearch_security_roles` and `direct_security_roles` same? Asking, as i see `direct_security_roles` being used in config/internal_users.yml file in this PR.

src/main/java/org/opensearch/security/dlic/rest/api/InternalUsersApiAction.java
shikharj05: The plan is to mark these deprecated in 3.0, removal in 4.0. Hence, we need to keep both in 3.0 or decide on whether we just document in 3.0 and remove cleanly in 4.0

src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
shikharj05: The initial name was ``opensearch_security_roles``, changed to ``direct_security_roles`` as per discussion on the PR. The test results are from before the rename :)

src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
nibix: The following is really just IMHO, so please be invited to have different opinions: ;-)

1. The change of the attribute names provides too little value to justify the investment of significant effort. Meanwhile, like stated above, the security config APIs have many shortcomings. Thus, one option would be to postpone the renaming until the security config APIs get a new version. I could offer to write an RFC on new security config APIs.
2. Alternatively, of course, we can have the changes as in this PR. The log message in its present form won't be possible, though.  We can also document this as a deprecation in the user docs. Still, I am then unsure of the path forward. Is there a possibility to drop the old attribute name at some point? And what would be the strategy for that?  I tried to find a document on OpenSearch deprecation strategies to get some guidance, but I could not find one.

shikharj05: > Still, I am then unsure of the path forward. Is there a possibility to drop the old attribute name at some point? And what would be the strategy for that? I tried to find a document on OpenSearch deprecation strategies to get some guidance, but I could not find one.

I am not sure if there's opensearch specific guidance. @cwperks do you know? 
The general guidance is to follow Semver. Per semver, if an API contract is to be changed - it needs a new major version. I couldn't find anything explicit about whether it's absolutely required to allow for migration path (deprecation cycle) - just that a new major version is a MUST with changes in API terms. I am okay to keep the behavior as is and update documentation to say ``opendistro_security_roles`` will be replaced by ``direct_security_roles`` in 4.0.

cwperks: I don't think I've seen a versioned API in opensearch and a quick search of the opensearch-api-specification repo doesn't reveal any either. This is the published APIs: https://opensearch-project.github.io/opensearch-api-specification/

When it comes to renaming an API parameter the steps would be:

1. Introduce new param name and support old param name, but mark it deprecated
2. Remove it in the next major version

To follow that we would need to mark it deprecated in 3.0 and removed in 4.0 to follow that convention. That being said, there have been efforts historically to remove legacy references like `opendistro` and `kibana` so there is a general understand that they are being phased out but we should do targeted deprecations where we can. Particular around things like security role names or the security index name.

cwperks: Off the top of my head, there is some awkwardness in the core repo around index templates.

There are 2 index template APIs:

1. `PUT _template/<template name>` (v1)
2. `PUT _index_template/<template name>` (v2)

From the code it looks like `_index_template` is a replacement for `_template` but they renamed it instead of versioning and the old APIs have yet to be removed.

src/main/java/org/opensearch/security/securityconf/DynamicConfigFactory.java
shikharj05: we need to log early so that we always log irrespective of ``direct_security_roles`` or ``opendistro_security_roles`` being used.

src/main/java/org/opensearch/security/securityconf/impl/v7/InternalUserV7.java
shikharj05: 3.0 is around the corner, @cwperks / @nibix - do you think we should just document this as an upcoming change in 4.0 and change it in 4.0 directly? If not, does this PR look okay?

DarshitChanpura:
@shikharj05 Would you please resolve merge conflicts? We can also pick this back up once 4.0 is closer?

DarshitChanpura:
@shikharj05 closing this PR as last update was 3 months ago. Feel free to open a new one.

Issue #2756, opened 2023-05-10
java.lang.IllegalStateException: Username cannot have '|' in the security plugin.


ran2806:
Hi All, 

We are using cognito with custom OIDC to authenticate users on Amazon Opensearch cluster. The OIDC provider is sending sub field in following pattern. sub = "idp-provider|email@test.com|tenant-account-id|user-id". I believe this sub field is used as a username in authentication. So when the  user comes authenticated opensearch is throwing error as 

java.lang.IllegalStateException: Username cannot have '|' in the security plugin.

Is there a way to use a different property for the username? 

Thanks.

cwperks:
[Triage] @cwperks Can you add some more details around the reason for this character being forbidden in a username? This delimeter is being used to serialize/deserialize a user on the thread context.

cwperks:
@RyanL1997 Are the forbidden characters in usernames displayed anywhere on the documentation website?

dtaivpp:
Adding myself to the thread for visibility :D

ran2806:
@cwperks currently the OIDC is configured and used by lot of services, So I can not change the subject pattern which is used as username. Also congnito does not allow to change the mapping of OIDC sub to some other value. I have a support ticket with AWS team for the same to use the different field for username.

mmehrten:
I've run into the same issue setting up Cognito as the OIDC provider for OpenSearch, where Cognito is backed by an Okta IDP. In this case Okta doesn't let you change the sub response, and Cognito won't let you use anything other than sub for the username attribute.

michal-plebanski-ideals:
I also faced this problem, makes it impossible to use Opensearch in AWS with Cognito and Federated IdP via OIDC to Okta / Auth0.

peternied:
I think this can be fixed by changing the logic to 'escape' uses of the pipe character.  We'd welcome a pull request to add this functionality.

cwperks:
> I think this can be fixed by changing the logic to 'escape' uses of the pipe character. We'd welcome a pull request to add this functionality.

The escaping can be done when the username is serialized [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L278)

shikharj05:
[...]

PR #5175, opened 2025-03-17
Escape pipe character for injected users

shikharj05:
### Description

Escape pipe while setting UserInfo in ThreadContext

Bug fix
Allow OIDC use-cases where username contain pipes.
usernames, roles & tenants will escape pipe character if present.

### Issues Resolved
https://github.com/opensearch-project/security/issues/2756

### Testing

### Check List
- [] New functionality includes testing

### DIFF
--- a/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
@@ -111,6 +111,7 @@ import org.greenrobot.eventbus.Subscribe;
 
 import static org.opensearch.security.OpenSearchSecurityPlugin.traceAction;
 import static org.opensearch.security.support.ConfigConstants.OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT;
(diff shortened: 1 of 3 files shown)


shikharj05:
> java.lang.IllegalStateException: Username cannot have '|' in the security plugin.

This error is specifically from AOS. The actual issue reported mentions usage of usage of Cognito with Amazon OpenSearch Service (AOS).

cwperks:
The user's reporting the issue have only mentioned a problem with usernames in certain SSO providers. Do you think we should limit this change to username and add logic to forbid it in other fields like roles and backend roles or should we do a blanket escape for all fields?

shikharj05:
> Thank you for the PR @shikharj05! The user's reporting the issue have only mentioned a problem with usernames in certain SSO providers. Do you think we should limit this change to username and add logic to forbid it in other fields like roles and backend roles or should we do a blanket escape for all fields?

IMO, it's better to escape all fields like roles, tenants, etc to be consistent and prevent future issues in the same area.

derek-ho (approved):
Thank you @shikharj05

PR #5187, opened 2025-03-17
[Resource Access Control] [Part3] Introduces a sample plugin to demonstrate usage of Resource Access Control feature

DarshitChanpura:
**#5185 and #5186 must be merged before merging this.**

#5016 is being broken down into smaller pieces. This is part 3.

### Description
Introduces a sample resource plugin that demonstrate usage of `opensearch-security-client` introduced in #5186 for resource access control.

### Issues Resolved
Related to #4500

### Testing
- automated tests

### Check List
~- [ ] New Roles/Permissions have a corresponding security dashboards plugin PR


DarshitChanpura:
CI should pass once #5185 and #5186 are merged.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/SampleResourcePluginTests.java
```diff
@@ -0,0 +1,194 @@
...
+            resourceId = response.getTextFromJsonBody("/message").split(":")[1].trim();
+            Thread.sleep(1000);
+        }
+
+        // Create an entry in resource-sharing index
+        try (TestRestClient client = cluster.getRestClient(cluster.getAdminCertificate())) {
+            // Since test framework doesn't yet allow loading ex tensions we need to create a resource sharing entry manually
+            String json = String.format(
+                "{"
+                    + "  \"source_idx\": \""
```
cwperks, line 86: nit: Now the main targets JDK 21 we can use text blocks to make multi-line json blocks like this.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/nonsystemindex/ResourceNonSystemIndexSIDisabledTests.java
```diff
@@ -0,0 +1,42 @@
...
+import org.opensearch.test.framework.cluster.LocalCluster;
+
+import static org.opensearch.security.spi.resources.FeatureConfigConstants.OPENSEARCH_RESOURCE_SHARING_ENABLED;
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+import static org.opensearch.test.framework.TestSecurityConfig.User.USER_ADMIN;
+
+/**
+ * These tests run with resource sharing enabled but the plugin does not declare a system index and system index protection is disabled
+ */
+public class ResourceNonSystemIndexSIDisabledTests extends AbstractResourcePluginNonSystemIndexTests {
```
cwperks, line 27: Why make the distinction? I think its reasonable to assume that resource index will always be a system index and will not be accessed directly by regular users. If a plugin declares a resource index without also specifying it as a system index then consider that invalid. If there was a way to hook into the bootup flow then it would be good to have a log message accordingly.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/AbstractSampleResourcePluginTests.java
```diff
@@ -0,0 +1,132 @@
...
+ */
+
+package org.opensearch.sample;
+
+import org.opensearch.security.spi.resources.ResourceAccessScope;
+import org.opensearch.test.framework.TestSecurityConfig;
+
+import static org.opensearch.sample.utils.Constants.RESOURCE_INDEX_NAME;
+import static org.opensearch.sample.utils.Constants.SAMPLE_RESOURCE_PLUGIN_PREFIX;
+import static org.opensearch.security.dlic.rest.support.Utils.PLUGIN_RESOURCE_ROUTE_PREFIX;
```
nibix: It might be more a personal taste and nitpicking but I'd recommend to avoid using constants from the code base to be tested in the code base that executes the tests.

The reason:

Suppose I accidentially add a random a character into the value of `PLUGIN_RESOURCE_ROUTE_PREFIX` .. this would break all plugins and clients which use this path. However, the test would happily pick up the changes and continue being green.

Using values/constants that are separate from the main code base, however, would allow us to notice such accidential changes, as the tests would fail then.

sample-resource-plugin/src/main/java/org/opensearch/sample/SampleResourceScope.java
```diff
@@ -0,0 +1,40 @@
...
+package org.opensearch.sample;
+
+import org.opensearch.security.spi.resources.ResourceAccessScope;
+
+/**
+ * This class implements two scopes  for the sample plugin.
+ * The first scope is SAMPLE_FULL_ACCESS, which allows full access to the sample plugin.
+ * The second scope is PUBLIC, which allows public access to the sample plugin.
+ */
+public enum SampleResourceScope implements ResourceAccessScope<SampleResourceScope> {
```
nibix: This is obsolete, right?

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/rest/create/CreateResourceRestAction.java
```diff
@@ -0,0 +1,96 @@
...
+            case PUT:
+                return createResource(source, client);
+            case POST:
+                return updateResource(source, request.param("resource_id"), client);
+            default:
+                throw new IllegalArgumentException("Illegal method: " + request.method());
+        }
+    }
+
+    @SuppressWarnings("unchecked")
```
nibix, line 63: Using `@SuppressWarnings("unchecked")` on untrusted input (the request body source) bears the risk of getting `ClassCastExceptions` deep inside the application code. While this is usually not security critical, it might locating the root cause of issues diificult - both for admins and for end users.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/rest/revoke/RevokeResourceAccessRestAction.java
```diff
@@ -0,0 +1,68 @@
...
+    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) {
+        String resourceId = request.param("resource_id");
+        if (Strings.isNullOrEmpty(resourceId)) {
+            throw new IllegalArgumentException("resource_id parameter is required");
+        }
+        Map<String, Object> source;
+        try (XContentParser parser = request.contentParser()) {
+            source = parser.map();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
nibix, line 59: `prepareRequest()` already throws `IOException`. Thus, it is not necessary to rethrow it as `RuntimException`.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/nonsystemindex/ResourceNonSystemIndexSIDisabledTests.java
DarshitChanpura: removed these tests to avoid confusion.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/rest/create/CreateResourceRestAction.java
DarshitChanpura: I agree. However this is is a sample plugin and hence should be fine IMO.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/rest/revoke/RevokeResourceAccessRestAction.java
DarshitChanpura: ack. modified the signature of prepareRequest() above.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/AbstractSampleResourcePluginFeatureEnabledTests.java
```diff
@@ -0,0 +1,476 @@
...
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.nullValue;
+import static org.opensearch.sample.utils.Constants.RESOURCE_INDEX_NAME;
+import static org.opensearch.security.resources.ResourceSharingConstants.OPENSEARCH_RESOURCE_SHARING_INDEX;
+import static org.opensearch.test.framework.TestSecurityConfig.User.USER_ADMIN;
+
+/**
+ * This abstract class defines common tests between different feature flag scenarios
+ */
+public abstract class AbstractSampleResourcePluginFeatureEnabledTests extends AbstractSampleResourcePluginTests {
```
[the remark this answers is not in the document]
DarshitChanpura: The tests are common for feature enabled scenarios. This class is inherited by 3 tests which implement 3 scenarios: User with full access, user with limited access and direct access when system index feature is disabled. The super-class test suite was added only to avoid duplicated tests-code.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/AbstractSampleResourcePluginTests.java
DarshitChanpura: This is no longer used.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/AbstractSampleResourcePluginFeatureEnabledTests.java
```diff
@@ -0,0 +1,476 @@
...
+            // Also update the in-memory map and get
+            ResourcePluginInfo.getInstance().getResourceIndicesMutable().add(RESOURCE_INDEX_NAME);
+            ResourceProvider provider = new ResourceProvider(
+                SampleResource.class.getCanonicalName(),
+                RESOURCE_INDEX_NAME,
+                new SampleResourceParser()
+            );
+            ResourcePluginInfo.getInstance().getResourceProvidersMutable().put(RESOURCE_INDEX_NAME, provider);
+
+            Thread.sleep(1000);
```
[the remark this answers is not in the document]
DarshitChanpura: replace with awaitility.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/rest/create/CreateResourceRestAction.java
nibix: Well, the primary purpse of sample plugins is to give other developers an idea/blueprint on how to design their own code. Thus, there is a big chance that developers will follow any pattern displayed in that code base, including bad patterns.

And, experience shows that developers have the tendency to follow patterns they are finding in existing code bases, no matter how bad these actually might be.

IMHO, to cope for this, we should add in such cases at least a comment like "this is just done for simplicity, do not do that for real plugins".

DarshitChanpura: We do use this pattern at a decent number of places in security plugin. Here is one such [example](https://github.com/opensearch-project/security/blob/3c635c9f64ec58be206d9c43a865bd8cfa423fa1/src/main/java/org/opensearch/security/privileges/dlsfls/IndexToRuleMap.java#L57). I will add a comment here stating suppressing should be avoided in real-world, however there is no mechanism to stop plugin's from implementing it.

.github/workflows/ci.yml
```diff
@@ -304,6 +304,93 @@ jobs:
...
+        run: |
+          ./gradlew clean \
+            :opensearch-security-spi:publishToMavenLocal \
+            -Dbuild.snapshot=false
+
+      - name: Run SampleResourcePlugin Integration Tests
+        uses: gradle/gradle-build-action@v3
+        with:
+          arguments: |
+            :opensearch-sample-resource-plugin:integrationTest -Dbuild.snapshot=false
```
cwperks, line 346: Why `-Dbuild.snapshot=false` everywhere?

sample-resource-plugin/README.md
```diff
@@ -0,0 +1,187 @@
...
+    - A user **with `sample-resource-plugin` permissions** but **without a shared resource**:
+        - **Cannot access the resource**, since resource-level access control applies.
+
+
+## API Endpoints
+
+The plugin exposes the following six API endpoints:
+
+### 1. Create Resource
+- **Endpoint:** `POST /_plugins/sample_resource_sharing/create`
```
cwperks, line 74: nit: can we introduce or user a convention?

`/_plugins/_{name_of_plugin}`/api/{resource_type}/{action}`

Iin this case the action would be create.

DarshitChanpura: this is sample plugin's readme that pertains to actions specific to the sample plugin. This README aims to provide information about setup and testing the sample plugin. IMO it should be fine as is.

DarshitChanpura: I've made it verbose intentionally.

DarshitChanpura:
Merged this feature-branch PR as i didn't see amy blocking reviews. The PR against main is already open and is the last step to getting this feature merged.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/rest/create/CreateResourceRestAction.java
nibix: There are different "risk classes" of using `@SuppressWarnings("unchecked")`. For the example you have linked, it is possible to argue within a minute that this is actually a safe cast and no late `ClassCastExceptions` can be caused by it.

PR #5212, opened 2025-03-25
Immutable user object

### LINKED ISSUE #5168: [Performance] Make User object immutable
# Background

This is a follow-up to the optimization efforts started in #3870 

In #3870, #4380 we have made already good progress in shaving off the performance overhead which is added by the security plugin to the overall performance shown by an OpenSearch cluster. In the course of this, we performed further performance tests which indicate that an OpenSearch cluster with the security plugin still shows a throughput with is about 10% lower than an OpenSearch cluster without the security plugin: https://github.com/opensearch-project/security/pull/4380#issuecomment-2324239103

# Analysis

It is already well known that the serialization and de-serialization of the user object adds a significant performance penalty: #2780, #4760 . There were a couple of efforts in this area already: One approach was to change the serialization protocol: #2802 . However, unexpected side-effects were encountered, which caused the change to be rolled back: #3771, #3776, #4670 .

In the course of this issue, we want to approach this again. However, this time not by optimizing the serialization protocol, but just by reducing the amount of serializations.

## Life cycle of the `org.opensearch.security.user.User` object

The major class of which objects are often serialized and deserialized in an OpenSearch cluster is `org.opensearch.security.user.User`. So, let's have a look at its life cycle:

- Instances of `org.opensearch.security.user.User` are usually created when a REST requested is authenticated by the security plugin. This happens in the class `BackendRegistry`, which either directly instantiates `User` objects or delegates it to classes implementing the `AuthenticationBackend` interface. Additionally, `BackendRegistry` maintains a cache of user objects in the case the same user sends several requests in sequence.

- In a second step,  `BackendRegistry` uses classes implementing `AuthorizationBackend` to add roles to the `User` object.

- The class `PrivilegesEvaluator` performs role mapping and adds a number of effective roles to the `User` object to make these available to dynamic attributes used in the role configuration.

- If the current node needs to delegate work to other nodes, the `User` object is serialized, put into a transport header and then de-serialized again on the receiving node. In case the current node needs to send messages to *n* other nodes, the user object will be serialized and de-serialized *n* times.

- On the receiving node, the de-serialized `User` object will be used. Actual authentication via `BackendRegistry` and `AuthenticationBackend` does not happen. That means, the `User` object stays constant.

## An immutable `User` object

The life cycle shows that the `User` object mostly stays constant after the authentication phase was finished. An outlier is the modification performed in `PrivilegesEvaluator`. The purpose of this, attribute interpolation, can be also achieved by other context variables. We already did changes in  #4380 with the goal to make this modification unnecessary; specifically, this is the introduction of the `PrivilegeEvaluationContext` class, which also holds the effective roles.

If we can achieve a `User` object which is guaranteed to stay constant after initialization, i.e., which is immutable, we gain capabilities which are very helpful for optimizing serialization performance:

- An immutable object means that the binary data created by serialization does not change as well. That means, the serialized data can be computed *once* per user object and then re-used again and again. The existing authentication cache even allows us to re-use the serialized data across different requests by the same user.

- Additionally, a cache similar to the cache maintained by `BackendRegistry` can be utilized to also cut down the number of de-serialization operations to a single one per user per node. Such a cache would map the serialized binary data to the actual user object.

- Immutable objects are inherently thread-safe. That makes any synchronization or locking mechanisms on the user object unnecessary.

## Necessary changes

Obviously, in order to make the `User` object immutable, all code which modifies the `User` object needs to be changed.

Besides from the creation pefrormed in  `BackendRegistry` and the  `AuthenticationBackend` implementations, this pertains especially to two further places.

### `AuthorizationBackend`

The `AuthorizationBackend` interface looks currently like this:

https://github.com/opensearch-project/security/blob/487fb05dfb6a6e1f7a2d06bf9ad5edee827a70af/src/main/java/org/opensearch/security/auth/AuthorizationBackend.java#L49-L70

Implementors of the `AuthorizationBackend` interface are supposed to call `addRoles()` on the supplied `User` object in order to provide the user roles they have found.

This will not be possible any more with immutable user objects. Thus, the  `AuthorizationBackend` interface must be changed. There are two options to handle this:

1. Pass a builder class for user objects to the  `AuthorizationBackend` interface: 

```
    void fillRoles(User.Builder userBuilder, AuthCredentials credentials) throws OpenSearchSecurityException;
```

2. Keep the `User` parameter, but let the `fillRoles()` create a new, modified `User` object and return it from the method:

```
    User fillRoles(User user, AuthCredentials credentials) throws OpenSearchSecurityException;
```

### `PrivilegesEvaluator`

As described above, the adding of effective roles needs to be removed from `PrivilegesEvaluator`:

https://github.com/opensearch-project/security/blob/487fb05dfb6a6e1f7a2d06bf9ad5edee827a70af/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L351-L352

The same information is already made available via `PrivilegeEvaluationContext`; it is just necessary to change all usage of this information to the new class.

# Additional Benefits

It is generally agreed that the use of immutable objects contributes to high-quality, robust software, especially in multi-threaded environments. The security plugin was struggling with concurrent modification issues on the `User` object in the past: #4642, #2263 .  Thus, the change will contribute to the general robustness of the security plugin.

nibix:
### Description

Fixes #5168 

This PR makes the User object immutable. This brings a number of significant advantages:

- An immutable object means that the binary data created by serialization does not change as well. That means, the serialized data can be computed once per user object and then re-used again and again. The existing authentication cache even allows us to re-use the serialized data across different requests by the same user.

- Additionally, a cache similar to the cache maintained by BackendRegistry can be utilized to also cut down the number of de-serialization operations to a single one per user per node. Such a cache would map the serialized binary data to the actual user object.

- Immutable objects are inherently thread-safe. That makes any synchronization or locking mechanisms on the user object unnecessary.

* Category: 
  * Enhancement
  * The serialization of the user object takes significant time, reducing the throughput. The changes help to minimize serialization operations. Additionally, they make the code more mature. 
  * No behavioral changes
### Issues Resolved

Fixes #5168 

### Testing

- Unit testing for parsing of serialized user object (from OpenSearch 2.19)
- Integration testing

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/auth/internal/InternalAuthenticationBackend.java
+++ b/src/main/java/org/opensearch/security/auth/internal/InternalAuthenticationBackend.java
@@ -30,7 +30,6 @@ import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
-import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 
@@ -69,13 +68,11 @@ public class InternalAuthenticationBackend implements AuthenticationBackend, Imp
         if (exists) {
             // FIX https://github.com/opendistro-for-elasticsearch/security/pull/23
             // Credits to @turettn
-            final Map<String, String> customAttributes = internalUsersModel.getAttributes(user.getName());
+            ImmutableMap<String, String> customAttributes = internalUsersModel.getAttributes(user.getName());
             ImmutableMap.Builder<String, String> attributeMap = ImmutableMap.builder();
 
-            if (customAttributes != null) {
-                for (Entry<String, String> attributeEntry : customAttributes.entrySet()) {
-                    attributeMap.put("attr.internal." + attributeEntry.getKey(), attributeEntry.getValue());
-                }
+            for (Entry<String, String> attributeEntry : customAttributes.entrySet()) {
+                attributeMap.put("attr.internal." + attributeEntry.getKey(), attributeEntry.getValue());
             }
 
             return Optional.of(
--- a/src/main/java/org/opensearch/security/user/User.java
+++ b/src/main/java/org/opensearch/security/user/User.java
@@ -33,6 +33,7 @@ import java.io.ObjectOutputStream;
 import java.io.ObjectStreamField;
 import java.io.Serial;
 import java.io.Serializable;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -57,14 +58,14 @@ import org.opensearch.security.support.Base64Helper;
  */
 public class User implements Serializable, CustomAttributesAware {
 
-    public static final User ANONYMOUS = new User("opendistro_security_anonymous").withRole("opendistro_security_anonymous_backendrole");
+    public static final User ANONYMOUS = new User("opendistro_security_anonymous").withRoles("opendistro_security_anonymous_backendrole");
 
     // This is a default user that is injected into a transport request when a user info is not present and passive_intertransport_auth is
     // enabled.
     // This is to be used in scenarios where some of the nodes do not have security enabled, and therefore do not pass any user information
     // in threadcontext, yet we need the communication to not break between the nodes.
     // Attach the required permissions to either the user or the backend role.
-    public static final User DEFAULT_TRANSPORT_USER = new User("opendistro_security_default_transport_user").withRole(
+    public static final User DEFAULT_TRANSPORT_USER = new User("opendistro_security_default_transport_user").withRoles(
         "opendistro_security_default_transport_backendrole"
     );
 
@@ -156,17 +157,10 @@ public class User implements Serializable, CustomAttributesAware {
     }
 
     /**
-     * Returns a new User object that additionally contains the provided backend role.
+     * Returns a new User object that additionally contains the provided backend roles.
      */
-    public User withRole(String role) {
-        return new User(
-            this.name,
-            new ImmutableSet.Builder<String>().addAll(this.roles).add(role).build(),
-            this.securityRoles,
-            this.requestedTenant,
-            this.attributes,
-            this.isInjected
-        );
+    public User withRoles(String... roles) {
+        return withRoles(Arrays.asList(roles));
     }
 
     /**
--- a/src/integrationTest/java/org/opensearch/security/user/UserFactoryTest.java
+++ b/src/integrationTest/java/org/opensearch/security/user/UserFactoryTest.java
@@ -1,0 +1,65 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.user;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+import com.google.common.collect.ImmutableSet;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import org.opensearch.OpenSearchException;
+import org.opensearch.common.settings.Settings;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+@RunWith(Parameterized.class)
+public class UserFactoryTest {
+    UserFactory subject;
+
+    @Test
+    public void parse_successful() {
+        User source = new User("test_user").withRoles(ImmutableSet.of("a", "b"));
+        User target = subject.fromSerializedBase64(source.toSerializedBase64());
+
+        assertEquals(source, target);
+    }
+
+    @Test
+    public void parse_invalid() {
+        try {
+            User target = subject.fromSerializedBase64("invaliddata123");
+            fail("Should have failed; got " + target);
+        } catch (Exception e) {
+            assertTrue(
+                "Got invalid stream header " + e,
+                e instanceof OpenSearchException && e.getMessage().contains("invalid stream header")
+            );
+        }
+    }
+
+    public UserFactoryTest(UserFactory subject, String name) {
+        this.subject = subject;
+    }
+
+    @Parameterized.Parameters(name = "{1}")
+    public static Collection<Object[]> params() {
+        return Arrays.asList(
+            new Object[] { new UserFactory.Simple(), "Simple" },
+            new Object[] { new UserFactory.Caching(Settings.EMPTY), "Caching" }
+        );
+    }
+}
--- a/src/integrationTest/java/org/opensearch/security/user/UserTest.java
+++ b/src/integrationTest/java/org/opensearch/security/user/UserTest.java
@@ -11,13 +11,16 @@
 package org.opensearch.security.user;
 
 import java.util.Arrays;
+import java.util.Map;
 
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import org.junit.Test;
 
 import org.opensearch.security.support.Base64Helper;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertSame;
 
 public class UserTest {
     @Test
@@ -75,5 +78,61 @@ public class UserTest {
                 .withAttributes(ImmutableMap.of("a", "v_a", "b", "v_b")),
             user
         );
+    }
+
+    @Test
+    public void withRoles() {
+        User original = new User("test_user").withRoles("a");
+        User modified = original.withRoles("b");
+
+        assertEquals(ImmutableSet.of("a"), original.getRoles());
+        assertEquals(ImmutableSet.of("a", "b"), modified.getRoles());
+    }
+
+    @Test
+    public void withRoles_unmodified() {
+        User original = new User("test_user").withRoles("a");
+        User unmodified = original.withRoles(ImmutableSet.of());
+
+        assertSame(original, unmodified);
+    }
+
+    @Test
+    public void withAttributes() {
+        User original = new User("test_user").withAttributes(Map.of("a", "1"));
+        User modified = original.withAttributes(Map.of("b", "2"));
+
+        assertEquals(ImmutableMap.of("a", "1"), original.getCustomAttributesMap());
+        assertEquals(ImmutableMap.of("a", "1", "b", "2"), modified.getCustomAttributesMap());
+    }
+
+    @Test
+    public void withAttributes_unmodified() {
+        User original = new User("test_user").withAttributes(Map.of("a", "1"));
+        User unmodified = original.withAttributes(Map.of());
+
+        assertSame(original, unmodified);
+    }
+
+    @Test
+    public void withRequestedTenant() {
+        User original = new User("test_user").withRequestedTenant("a");
+        User modified = original.withRequestedTenant("b");
+
+        assertEquals("a", original.getRequestedTenant());
+        assertEquals("b", modified.getRequestedTenant());
+    }
+
+    @Test
+    public void withRequestedTenant_unmodified() {
+        User original = new User("test_user").withRequestedTenant("a");
+        User unmodified = original.withRequestedTenant("a");
+
+        assertSame(original, unmodified);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void illegalName() {
+        new User("");
     }
 }
--- a/src/main/java/org/opensearch/security/auth/ldap2/LDAPAuthorizationBackend2.java
+++ b/src/main/java/org/opensearch/security/auth/ldap2/LDAPAuthorizationBackend2.java
@@ -186,9 +186,7 @@ public class LDAPAuthorizationBackend2 implements AuthorizationBackend, Destroya
         }
 
         if (skipUsersMatcher.test(authenticatedUser)) {
-            if (isDebugEnabled) {
-                log.debug("Skipped search roles of user {}/{}", authenticatedUser, originalUserName);
-            }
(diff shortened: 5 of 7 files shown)


src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -642,16 +650,18 @@ private User impersonate(final SecurityRequest request, final User originalUser)
             final boolean isDebugEnabled = log.isDebugEnabled();
             // loop over all http/rest auth domains
             for (final AuthDomain authDomain : restAuthDomains) {
-                final AuthenticationBackend authenticationBackend = authDomain.getBackend();
+                if (!(authDomain.getBackend() instanceof ImpersonationBackend impersonationBackend)) {
```
cwperks, line 653: Is this a typo here? Shorten to `if (!(authDomain.getBackend() instanceof ImpersonationBackend)) {`

cwperks: TIL of Pattern Matching introduced in JDK 17: https://docs.oracle.com/en/java/javase/17/language/pattern-matching-instanceof.html

src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java
```diff
@@ -293,10 +293,7 @@ private Set<String> map(final User user, final TransportAddress caller) {
                 return Collections.emptySet();
             }
 
-            final Set<String> securityRoles = new HashSet<>();
-            synchronized (user.getSecurityRoles()) {
-                securityRoles.addAll(user.getSecurityRoles());
-            }
+            final Set<String> securityRoles = new HashSet<>(user.getSecurityRoles());
```
nibix, line 296: The synchronization can be removed again because `user.getSecurityRoles()`  cannot be any longer concurrently modified as observed in https://github.com/opensearch-project/security/issues/4441

src/main/java/org/opensearch/security/auth/UserInjector.java
```diff
@@ -66,54 +62,7 @@ public UserInjector(Settings settings, ThreadPool threadPool, AuditLog auditLog,
 
     }
 
-    public static class InjectedUser extends User {
```
nibix, line 69: FYI: The class `InjectedUser` is no longer necessary, as it was only used to transport the value  `transportAddress` back to its direct caller, which would set it into the thread context. Afterwares, the additional fields in `InjectedUser` would not be used any more. Additionally, the `InjectdUser` was serialized to a normal `User` object.

Thus, we have replaced `Injected` user by the `Result` class below which is evaluated by the direct caller and then discarded.

src/main/java/org/opensearch/security/user/User.java
```diff
@@ -305,4 +337,27 @@ public boolean isServiceAccount() {
...
+
+        return result;
+    }
+
+    /**
+     * Creates a backwards compatible object that can be used for serialization
+     */
+    Object writeReplace() {
+        return new org.opensearch.security.user.serialized.User(this);
+    }
```
nibix, line 356: FYI: This does not work correctly yet. The serialized output will refer to the class `org.opensearch.security.user.serialized.User` which is not known in previous OpenSearch versions. Thus, mixed clusters fail. 

I am not completely sure how to fix this. One option would be to move the "actual" User class to a different package and just keep the dummy for serialization here. But I do not like this approach too much.

src/main/java/org/opensearch/security/user/User.java
```diff
@@ -305,4 +342,47 @@ public boolean isServiceAccount() {
...
+            throws IOException {
+        ObjectOutputStream.PutField fields = out.putFields();
+        fields.put("name", name);
+        fields.put("roles", Collections.synchronizedSet(new HashSet<>(this.roles)));
+        fields.put("securityRoles", Collections.synchronizedSet(new HashSet<>(this.securityRoles)));
+        fields.put("requestedTenant", requestedTenant);
+        fields.put("attributes", Collections.synchronizedMap(new HashMap<>(this.attributes)));
+        fields.put("isInjected", this.isInjected);
+
+        out.writeFields();
```
nibix, line 386: This is not particularly nice, but it is necessary to keep backwards compatibility with nodes on older OpenSearch versions. To a certain extent, I am inclined again to use a different serialization format on clusters with only new nodes in order to get rid of these `synchronizedSet()` calls.

nibix:
@cwperks @kkhatua @DarshitChanpura 

I'd like to hear an opinion from you regarding the following:

In the immutable user object implementation, we need to use the following quite inefficient code to create serialized forms of the user object that are backwards compatible. That's especially the wrapping of collections inside `Collections.synchronizedSet(new HashSet<>(...))`. The new code does not need this any more, but nodes on old versions need this.

https://github.com/opensearch-project/security/blob/7a49f7c39fd4ecef635b9af05f9c0721d8a854d3/src/main/java/org/opensearch/security/user/User.java#L361-L384

Due to this, I am leaning towards again using a custom serialization protocol for the immutable user object. Thus, in perspective, we could retire the wrapping inside `Collections.synchronizedSet(new HashSet<>(...))`. I am aware of the previous issues with custom serialization formats. However, in my perception, the issues all came from the huge data structures created by DLS/FLS. The user object, on the other hand, has a quite bounded size. Thus, I believe that this would work well and bring us benefits.

cwperks:
> However, in my perception, the issues all came from the huge data structures created by DLS/FLS.

yes, that was the case and led to large duplication when using custom serialization for that data structure: https://github.com/opensearch-project/OpenSearch/issues/11430

> Due to this, I am leaning towards again using a custom serialization protocol for the immutable user object.

+1 to this if you are seeing noticeable improvements with the new format. 

I

nibix:
@cwperks 

> +1 to this if you are seeing noticeable improvements with the new format.

I believe the performance improvements will be noticeable in graphs, but not as pronounced as the improvements we see already now.

I rather look at this from a code quality perspective: Without changing the protocol at one point, we have to carry the baggage of doing this ...

```
fields.put("roles", Collections.synchronizedSet(new HashSet<>(this.roles))); 
fields.put("securityRoles", Collections.synchronizedSet(new HashSet<>(this.securityRoles))); 
fields.put("attributes", Collections.synchronizedMap(new HashMap<>(this.attributes)));
```

... without a real reason to use `Collections.synchronizedSet()` and `Collections.synchronizedMap()` other than backwards compatibility. Such "magic code" (in the terms of "do not touch the following code, even though it fulfils no real function") is IMHO something one should avoid.

src/main/java/org/opensearch/security/auth/UserInjector.java
```diff
@@ -123,59 +87,74 @@ public InjectedUser getInjectedUser() {
...
 
         if (log.isTraceEnabled()) {
             log.trace("Injected user object:{} ", injectedUser.toString());
         }
 
-        return injectedUser;
+        return new Result(injectedUser, transportAddress);
+    }
+
+    public TransportAddress parseTransportAddress(String addr) throws UnknownHostException, IllegalArgumentException {
```
nibix, line 141: FYI: This method is not new, it was just moved from the now deleted inner class `InjectedUser`

src/main/java/org/opensearch/security/user/User.java
```diff
@@ -278,4 +320,78 @@ public boolean isServiceAccount() {
...
+    private void writeObject(ObjectOutputStream out) throws IOException {
+        ObjectOutputStream.PutField fields = out.putFields();
+        fields.put("name", name);
+        fields.put("roles", Collections.synchronizedSet(new HashSet<>(this.roles)));
+        fields.put("securityRoles", Collections.synchronizedSet(new HashSet<>(this.securityRoles)));
+        fields.put("requestedTenant", requestedTenant);
+        fields.put("attributes", Collections.synchronizedMap(new HashMap<>(this.attributes)));
+        fields.put("isInjected", this.isInjected);
+
+        out.writeFields();
```
nibix, line 395: This is not particularly nice, but it is necessary to keep backwards compatibility with nodes on older OpenSearch versions. To a certain extent, I am inclined again to use a different serialization format on clusters with only new nodes in order to get rid of these synchronizedSet() calls.

src/integrationTest/java/org/opensearch/security/privileges/ActionPrivilegesTest.java
```diff
@@ -1071,8 +1071,7 @@ static SecurityDynamicConfiguration<RoleV7> createRoles(int numberOfRoles, int n
     }
 
     static PrivilegesEvaluationContext ctx(String... roles) {
-        User user = new User("test_user");
-        user.addAttributes(ImmutableMap.of("attrs.dept_no", "a11"));
+        User user = new User("test_user").withAttributes(ImmutableMap.of("attrs.dept_no", "a11"));
```
cwperks, line 1074: Should we consider using the Builder pattern which is pretty prevalently used across OpenSearch?

src/main/java/org/opensearch/security/auth/UserInjector.java
```diff
@@ -188,15 +167,15 @@ boolean injectUser(SecurityRequestChannel request) {
...
+        auditLog.logSucceededLogin(injectedUser.getUser().getName(), true, null, request);
 
         return true;
     }
 
-    protected Map<String, String> mapFromArray(String... keyValues) {
+    protected ImmutableMap<String, String> mapFromArray(String... keyValues) {
         if (keyValues == null) {
-            return Map.of();
+            return ImmutableMap.of();
```
cwperks, line 178: FYI Since the min java compatibility version is now JDK21, we can take advantage of immutable classes from within the JDK: https://docs.oracle.com/en/java/javase/21/core/creating-immutable-lists-sets-and-maps.html

`Map.of()` now creates an immutable map and gives an UnsupportedOperationException if you try to put a new entry into the map.

cwperks: The only downside I see of using JDK APIs is that there is no way for the consumer to know if a map is immutable or not until actually trying to make a modification at which point they would get a RuntimeException.

src/main/java/org/opensearch/security/auth/internal/NoOpAuthorizationBackend.java
```diff
@@ -45,8 +45,7 @@ public String getType() {
     }
 
     @Override
-    public void fillRoles(final User user, final AuthCredentials authCreds) {
-        // no-op
+    public User addRoles(User user, AuthenticationContext context) {
```
cwperks, line 48: much better 🙌

src/main/java/org/opensearch/security/auth/UserInjector.java
nibix: These are also my thoughts, using the explicit ImmutableSet class, I document that something's is immutable and I can also trust that something's immutable if it is passed as argument. 

That saves quite a bit of copying/wrapping because one is afraid that the object might be mutated somewhere else.

cwperks (approved):

src/main/java/org/opensearch/security/auth/internal/InternalAuthenticationBackend.java
```diff
@@ -133,21 +131,14 @@ public User authenticate(final AuthCredentials credentials) {
...
-
-                final User user = new User(credentials.getUsername(), roles, credentials);
-
-                final List<String> securityRoles = internalUsersModel.getSecurityRoles(credentials.getUsername());
-                if (securityRoles != null) {
-                    user.addSecurityRoles(securityRoles);
-                }
-                return user;
+                ImmutableSet<String> backendRoles = internalUsersModel.getBackendRoles(credentials.getUsername());
+                ImmutableSet<String> securityRoles = internalUsersModel.getSecurityRoles(credentials.getUsername());
```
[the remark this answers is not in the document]
nibix: In this PR, I just have been sticking to the terminology that was present in the respective code parts before. I just corrected the typo `backenRoles`.

Still, indeed, all the different roles in the code base are quite confusing.

As far as I know, it is like this:

- `User.roles` refers to backend roles (coming from external sources); in the remaining code base, these are often referred to as [backendRoles](https://github.com/search?q=repo%3Aopensearch-project%2Fsecurity+backendroles&type=code&p=1) (or [backenRoles](https://github.com/search?q=repo%3Aopensearch-project%2Fsecurity+backenroles&type=code) ) 
- `User.securityRoles` refers to roles that should not go through role mapping. 

In non-user contexts we have:

- `mappedRoles`, which is the union of  `securityRoles`,  the result of role mapping and injected roles
- The `roles.yml` configuration, where the role names correspond to the `mappedRoles`, but usually not to `backendRoles`. 

For my taste, in our code base, roles assume just too many different roles. Thus, I would vote for a very clear differentiation each for each use. Thus, either strive for clear qualification - like `backendRoles` vs `mappedRoles`. Or even rethink the scope of roles. Actually, IMHO, all of the role mapping process pertaining to backend roles could be already done during authentication time - which could severely reduce the scope leak of backend roles. 

I might file an RFC for this.

kumargu:
I'll start taking a look at this tomorrow.

src/main/java/org/opensearch/security/auth/AuthenticationContext.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * This class allows HTTPAuthenticators and authentication backends to provide context data to authorization backends.
+ * This is especially useful for siblings of authentication backends and authorization backends (like LDAP authc and
+ * LDAP authz) to pass data which is specific to the auth type (like the LDAP user entry).
+ * <p>
+ * This allows to abolish specialized sub-classes of the User object (like LdapUser).
+ */
+public class AuthenticationContext {
+    private final AuthCredentials credentials;
+    private final Map<Class<?>, Object> contextData = new HashMap<>();
```
kumargu, line 29: queries to ```contextData```  are always from a single thread?  if not do we need a ConcurrentHashMap?

nibix: The AuthenticationContext class is scoped just to a single thread; that's the thread that processes the current request. As soon as we might spawn into other threads, instances of this class are already out of scope.

kumargu (approved):

willyborankin:
Thank you, @nibix, for the excellent work. If all comments have been addressed, I'll go ahead and merge it

Issue #4054, opened 2024-02-16
[BUG] Annoying warning message "No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'" still not fixed


ComBin:
**What is the bug?**
While basic+saml security configuration every user login generate a lot of WARN message in logs:
```
[2024-02-16T11:24:43,253][WARN ][o.o.s.h.HTTPBasicAuthenticator] [elk-h01.test.logs.infra.rbc.ru] No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'
[2024-02-16T11:24:43,261][WARN ][o.o.s.h.HTTPBasicAuthenticator] [elk-h01.test.logs.infra.rbc.ru] No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'
[2024-02-16T11:24:43,368][WARN ][o.o.s.h.HTTPBasicAuthenticator] [elk-h01.test.logs.infra.rbc.ru] No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'
[2024-02-16T11:24:43,374][WARN ][o.o.s.h.HTTPBasicAuthenticator] [elk-h01.test.logs.infra.rbc.ru] No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'
```
This complicates debugging, takes up extra disk space, etc.

**How can one reproduce the bug?**
1. Take some adfs server
2. Create security config like this:
```
---
_meta:
  type: config
  config_version: 2
config:
  dynamic:
    kibana:
      server_username: opensearch_dashboards
      index: ".kibana"
    authc:
      basic_internal_auth_domain:
        description: Authenticate via HTTP Basic against internal users database
        http_enabled: true
        transport_enabled: true
        order: 1
        http_authenticator:
          type: basic
          challenge: false
        authentication_backend:
          type: intern
      saml_auth_domain:
        http_enabled: true
        transport_enabled: false
        order: 2
        http_authenticator:
          type: saml
          challenge: true
          config:
            idp:
              metadata_url: <metadata_url>
              entity_id: <entity_id>
            sp:
              entity_id: <entity_id>
            kibana_url: <kibana_url>
            roles_key: roles
            exchange_key: <some_key>
        authentication_backend:
          type: noop
```
3. Login adfs user
4. See log file in OpenSearch node

**What is the expected behavior?**
This log message see only on TRACE/DEBUG level.

**What is your host/environment?**
[...]

PR #3364, opened 2023-09-14
Change log message from warning to trace on WWW-Authenticate challenge

bagipriyank:
### Description
Change warning message No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic' to trace

### Issues Resolved
- Resolves #3273

### Check List
- [ ] New functionality includes testing
- [ ] New functionality has been documented
- [X] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #3273, opened 2023-08-31
[BUG] Lots of waring message `No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'` in log


Hailong-am:
**What is the bug?**

When i enable both SAML and basic auth and login with SAML, there have lots of basic auth warning message 
`[2023-08-31T13:35:04,053][WARN ][o.o.s.h.HTTPBasicAuthenticator] [bcd07463160c] No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'` printed in OpenSearch log file 

**How can one reproduce the bug?**

**What is the expected behavior?**
Don't print warning message or should this be debug level

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

![image](https://github.com/opensearch-project/security/assets/113890546/2e8e6588-9f2c-41b6-b567-aaad845d6c05)

enable both SAML and basic 
![image](https://github.com/opensearch-project/security/assets/113890546/e3eb7410-50fc-4b7e-bc2d-ab371b89c32f)

**Do you have any additional context?**
Add any other context about the problem.

scrawfor99:
[Triage] Thank you for filing this issue. This message is a warning message that is expected when you run with the SAML endpoint configuration you are using. This is a response from not having a Basic auth header (you have none because you are using SAML). You can disregard this message. 

We can follow-up this issue by making a change so that this log is TRACE instead of WARNING level. Then you should not see this output.

taltsafrirpx:
any update with a solution?

Hailong-am:
> any update with a solution?

@taltsafrirpx the log level have changed to trace, so you should not see those warning by default.

jkrishnarao2003:
We are using OpenSearch 2.10.0. We are seeing this log message hugely.  Which version this was fixed?

Hailong-am:
https://github.com/opensearch-project/security/commit/d7d8472eeab0d0f497d89267bd1e7ec360c6f18c
should be 2.11.0
![image](https://github.com/opensearch-project/security/assets/113890546/4f9ba3a2-a526-4844-8122-0f2265545d66)

SeemsLikeIt:
For information just migrated my cluster to 2.11.1 and got log as warning with SAMl configuration.
[2024-01-31T20:12:52,211][WARN ][o.o.s.h.HTTPBasicAuthenticator] [os1] No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'

[...]

PR #5221, opened 2025-03-27
Change HTTPBasicAuthenticator log message from warning to trace for 'No basic auth' error

StewartWBrown:
### Description

When basic auth and SAML authentications are both enabled, the logs are filled with messages stating: 
`"level": "WARN", "component": "o.o.s.h.HTTPBasicAuthenticator", "message": "No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'"`

This was thought to be addressed by https://github.com/opensearch-project/security/pull/3364, however the logging message remained in another area which this PR has tuned to trace level instead. 

### Issues Resolved

- https://github.com/opensearch-project/security/issues/4054
- https://github.com/opensearch-project/security/issues/3273

### Testing

Bulk Integration testing workflow with this change seems to look fine!

### Check List


src/main/java/org/opensearch/security/support/HTTPHelper.java
```diff
@@ -40,9 +40,11 @@ public class HTTPHelper {
 
     public static AuthCredentials extractCredentials(String authorizationHeader, Logger log) {
 
+        final boolean isTraceEnabled = log.isTraceEnabled();
+
         if (authorizationHeader != null) {
-            if (!authorizationHeader.trim().toLowerCase().startsWith("basic ")) {
-                log.warn("No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'");
+            if (!authorizationHeader.trim().toLowerCase().startsWith("basic ") && isTraceEnabled) {
+                log.trace("No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'");
```
shikharj05, line 47: Thanks for the PR. While I understand noise created when multiple auth types are created, moving this line to trace logging might be an issue for clusters with only basic auth enabled. I don't think there's another log line available. 

For https://github.com/opensearch-project/security/issues/4054, would it be possible to change the ``order``  , i.e. mark SAML as order 1 and basic auth as order 2? Do you still see same amount of logs?

RyanL1997: +1 to @shikharj05

terryquigleysas: I also agree that changing to trace isn't the answer but the behavior should change for valid scenarios requiring multiple authenticators using fallthrough.

The fix required could be one of the following:

1. Pass the _isChallenge_ flag from the _authDomain_ to the _extractCredentials_ calls. Only _log.warn_ when _isChallenge_ is _true_
2. Alternatively, and probably a more complete fix, would be be to pass the _firstChallengingHttpAuthenticator_ to the _extractCredentials_ calls. Only _log.warn_ when the _firstChallengingHttpAuthenticator_ is of type _HTTPBasicAuthenticator_

The starting point for these changes would be the _httpAuthenticator.extractCredentials_ call at https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L282

StewartWBrown: Thanks all! Initially raised this PR as i felt that it was odd that basic auth was the only authentication type logging a warn message on checking that the header was in the expected format- leaving this message-spam only occurring when basic auth was earlier in the order.

But i understand this is could be a helpful message, so banishing it away to Trace or Debug may not be useful (especially for a more common auth type like Basic). I agree with Terry that passing isChallenge may be the way forward in this case. Especially after reading the docs, which at the moment seems inconsistent with this logging message: https://opensearch.org/docs/latest/security/authentication-backends/basic-authc/#the-challenge-setting

Will look into this a little more later this week with a bit more of a complete fix!

StewartWBrown:
Have added in the isChallenge flag to extractCredentials as suggested! Log the error message as a warn if 'challenge' is enabled, and at trace level if 'trace' is enabled. Otherwise do not print the error which will hopefully reduce the log spam being seen when security configs are set up in this specific ordering. 

Sorry for delay, other things popped up over past few weeks! Still need to test this works as expected, and will have time to do so at the start of next week.

cwperks:
@StewartWBrown Sorry for the delay on review. I was out at OpenSearchCon EU last week and back in the office this week.

cwperks:
@StewartWBrown I think the fix for this can be even simpler. wdyt about only logging the `No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'` message if the `Authorization` header is missing?

It doesn't make sense to log that out when the Authorization header is present, but of a different type of token that is not a basic auth header.

StewartWBrown:
@cwperks No worries at all - hope OpenSearchCon EU went well! Looking forward to catch up on some sessions if they become available online :) 

Although this would stop the log message if using JWT authentication for example (where the header is present, but with a different type of token). Wouldn't this issue still persist if using SAML which I believe doesn't have an Authorization header? Not too familiar with some of the other Auth methods OpenSearch can use so not sure if there's others that fall in the same boat too.

Do appreciate alternative ideas though, as i don't want an overkill solution to stop the noise of one log message! From my understanding of the codebase though i think this extra flag could be the most straightforward method.

nibix:
I'd just give my five cents here, but then I'll immediately disappear again, because I have two days off ;-)

It can be helpful to just consult the guidelines for using log4j log levels at https://commons.apache.org/proper/commons-logging/guide.html#Message_Priorities.2FLevels

> WARN: Use of deprecated APIs, poor use of API, 'almost' errors, other runtime situations that are undesirable or unexpected, but not necessarily "wrong". Expect these to be immediately visible on a status console.

IMHO, these criteria do not match the absence of an "Authorization" header.  This is not an undesirable or unexpected condition, it is part of a normal protocol flow.

Let's check the guidelines for INFO:

> INFO: Interesting runtime events (startup/shutdown). Expect these to be immediately visible on a console, so be conservative and keep to a minimum. 

Also for this, I'd say that the criteria is not met. It will be impossible to keep them at a minimum if these are caused by normal protocol flows.

Thus, my opinion would be to just generally lower the log level to DEBUG.

StewartWBrown:
For further support of changing the message logging level, I can see similar checks to this one for different Auth types happening at either Debug or Trace: 

[HTTPProxyAuthenticator](https://github.com/opensearch-project/security/blob/da21abf4e5bce0de99d66ded52fba04239c96c8e/src/main/java/org/opensearch/security/http/HTTPProxyAuthenticator.java#L86),  [OnBehalfOf](https://github.com/opensearch-project/security/blob/da21abf4e5bce0de99d66ded52fba04239c96c8e/src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java#L169), [HTTPClientCertAuthenticator](https://github.com/opensearch-project/security/blob/da21abf4e5bce0de99d66ded52fba04239c96c8e/src/main/java/org/opensearch/security/http/HTTPClientCertAuthenticator.java#L95),[ JWT Authentication](https://github.com/opensearch-project/security/blob/6e78dd9d1a1e5e05d50b626d796bd3011ac5c530/src/main/java/org/opensearch/security/auth/http/jwt/HTTPJwtAuthenticator.java#L150).

terryquigleysas:
@cwperks Thank you for revisiting this. We found that the change submitted here was the least intrusive that would actually fix the bug correctly.

The basic auth check should only log that it is returning a 401 when set to challenge so we passed that flag through. Note we also have a finalizing basic authenticator that is set to challenge and will log and behave appropriately when authentication details are missing.

The first basic auth check cannot challenge as we have others that can follow, e.g. JWT , that use the header. For other auth types it seems better to check for the challenge flag too rather than relying on specific text included in the header - this is also to ensure it is future-proof.

cwperks:
@terryquigleysas I think there's an opportunity to make the logic much simpler and achieve the same outcome without having to keep track of whether the authenticated is a challenging authenticator.

I think if these 2 conditions are met:

- Request lacks `Authorization` HTTP Header
- The route is not `/_opendistro/_security/saml/acs` <- This is the ACS SAML endpoint where the request body contains the XML SAML assertion that we change to a JWT for the remainder of the session and will not have the `Authorization` http header.

Then we should log out that warning, otherwise skip logging out the warning. The current logic is definitely wrong and should not log a warning when an `Authorization` header is present..of any kind.

terryquigleysas:
> @terryquigleysas I think there's an opportunity to make the logic much simpler and achieve the same outcome without having to keep track of whether the authenticator is a challenging authenticator.
> 
> If these 2 conditions are met:
> 
> * Request lacks `Authorization` HTTP Header
> * The route is not `/_opendistro/_security/saml/acs` <- This is the ACS SAML endpoint where the request body contains the XML SAML assertion that we change to a JWT for the remainder of the session and will not have the `Authorization` http header.
> 
> Then we should log out that warning, otherwise skip it. The current logic is definitely wrong and should not log a warning when an `Authorization` header is present..of any kind.

Thanks @cwperks , we will take a look again with this suggestion.

src/main/java/org/opensearch/security/support/HTTPHelper.java
```diff
@@ -38,11 +38,17 @@
...
 
-    public static AuthCredentials extractCredentials(String authorizationHeader, Logger log) {
+    public static AuthCredentials extractCredentials(String authorizationHeader, Logger log, boolean isChallenge) {
+
+        boolean isTraceEnabled = log.isTraceEnabled();
 
         if (authorizationHeader != null) {
             if (!authorizationHeader.trim().toLowerCase().startsWith("basic ")) {
-                log.warn("No 'Basic Authorization' header, send 401 and 'WWW-Authenticate Basic'");
+                if (isChallenge) {
```
DarshitChanpura, line 47: Apologies for joining the party late, i'm slightly confused on what exactly are we trying to solve here? Is it that we log at warn only when Basic Auth fails or is  it that we will log at warn when no other domains to be evaluated against are present? 
I'm okay with current implementation if the idea is to simply reduce warning messages for auth types other than Basic.

cwperks: The goal is to only log this in valid scenarios. The following criteria needs to be met:

1) Request has no Authorization header
2) There is a challenging basic auth authenticator
3) Request is not for the SAML ACS endpoint which does not accept an Authorization header (its the endpoint that accepts the XML assertion) and is used in the SAML flow.

Currently, this is getting logged if a request has an Authorization header of a different kind (like a bearer token) and its annoying because its being logged excessively and its not accurate.

DarshitChanpura: If the idea is to limit logging of this message only when basic auth is the challenging authenticator and creds fail, I'm okay with this flag.

src/main/java/org/opensearch/security/support/HTTPHelper.java
cwperks: @shikharj05 This log line is available elsewhere in the BackendRegistry here: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L352-L354

tbh, if the log level is a concern I would suggest to remove this line and make the one in BackendRegistry warn level to solve this issue.

cwperks: Something like this: https://github.com/cwperks/security/pull/57

terryquigleysas: > Something like this: [cwperks#57](https://github.com/cwperks/security/pull/57)

@cwperks Yes, removing the problematic log line, as you have here, should also resolve it.

terryquigleysas:
@StewartWBrown This can be closed off now https://github.com/opensearch-project/security/pull/5377 has been merged.
Thanks for the PR.

Issue #5219, opened 2025-03-26
[RFC] Fine grained settings permissions


sam-herman:
## Overview
This RFC proposes adding a new Settings Permission Valve to OpenSearch Security that provides fine-grained control over cluster and index settings modifications. The valve will allow administrators to define which users and roles can modify specific settings using wildcard patterns.

## Problem Statement
Currently, OpenSearch Security lacks granular control over settings modifications. Users with cluster or index management `UPDATE` permissions can modify any settings within their scope, which may lead to security and operational risks on shared clusters.

## Proposed Solution
Implement a new SettingsPermissionValve that:
1. Validates settings modifications against role-based permissions
2. Supports wildcard patterns for settings names
3. Handles both cluster and index settings
4. Integrates with existing security model

## Benefits
1. Enhanced security through granular settings control
2. Prevent accidental misconfiguration
3. Support for role-based access control
4. Audit trail for settings modifications

## Architecture

The suggested implementation will be similar to document level security.

### Example Usage

Role Configuration with index and cluster level fine grained settings permissions:

```yaml
cluster_settings_role:
  cluster_permissions:
    - "cluster:admin/settings/*"
  allowed_cluster_settings:
    - "cluster.routing.*"
    - "cluster.metadata.*"

index_settings_role:
  index_permissions:
    - index_patterns: ["*"]
      allowed_actions: ["indices:admin/settings/*"]
      allowed_settings: ["index.number_of_replicas", "index.refresh_interval"]
```

Overall role configuration:
```yaml
# Example role configuration
roles:
  monitoring_role:
    cluster_permissions:
      - "cluster:admin/settings/*"
    allowed_cluster_settings:
      - "cluster.routing.allocation.*"
      - "cluster.routing.rebalance.*"
    
  index_manager_role:
    index_permissions:
      - index_patterns: ["*"]
        allowed_actions: ["indices:admin/settings/*"]
        allowed_settings:
          - "index.number_of_replicas"
          - "index.refresh_interval"
          - "index.blocks.*"
```

## Alternatives Considered
NA

sam-herman:
If this suggestion looks good, feel free to assign to me, I have a PR ready.

shikharj05:
Hi, Thanks @sam-herman for the proposal! Sounds like a good addition. Couple of questions or points that should be added in the proposal - 
* How would backward compatibility be handled? i.e. for an existing role with ``allowed_settings`` as empty, would this allow all access or now start to restrict access?
[...]

PR #5222, opened 2025-03-27
5219/fine grained settings permissions

sam-herman:
### Description
Introduces fine grained permissions for settings modifications on index and cluster. For more details see https://github.com/opensearch-project/security/issues/5219

### Issues Resolved
Resolves https://github.com/opensearch-project/security/issues/5219

### Testing
`./gradlew clean build`

### Check List
- [ x] New functionality has been documented
- [ x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md)


src/main/java/org/opensearch/security/configuration/SettingsPermissionValve.java
```diff
@@ -0,0 +1,26 @@
...
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.configuration;
+
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.security.privileges.PrivilegesEvaluationContext;
+
+public interface SettingsPermissionValve {
```
nibix, line 17: The presence of the `DlsFlsRequestValve` has mostly historical reasons, as it was once an optional module. 

I do not believe we need to make the settings access control optional (do we?). If we do not have the need to make it optional, I would recommend to make it non-pluggable to reduce complexity.

src/main/java/org/opensearch/security/configuration/SettingsPermissionValveImpl.java
```diff
@@ -0,0 +1,206 @@
...
+        this.clusterService = clusterService;
+        this.adminDNs = adminDNs;
+        this.auditLog = auditLog;
+
+        // Add listener for configuration updates
+        clusterService.addListener(event -> {
+            SecurityDynamicConfiguration<RoleV7> config = rolesConfiguration.get();
+            if (config != null) {
+                // Handle any cluster state related updates if needed
+                // TODO: is this needed here? It's used in DlsFlsValveImpl but seems to be working well without it here on cluster updates
```
nibix, line 60: This is because the DLS/FLS feature uses denormalized data structures which are updated whenever a new index is added to the cluster. This is for performance reason. If we do not use denormalized data structures here, we do not need listening to the cluster state.

src/main/java/org/opensearch/security/configuration/SettingsPermissionValveImpl.java
```diff
@@ -0,0 +1,206 @@
...
+    public boolean invoke(PrivilegesEvaluationContext context, ActionListener<?> listener) {
+        final ActionRequest request = context.getRequest();
+        
+        // Skip validation for admin users
+        if (adminDNs.isAdmin(context.getUser())) {
+            return true;
+        }
+
+        try {
+            if (request instanceof ClusterUpdateSettingsRequest) {
```
nibix, line 75: I would recommend to add an interface to core which marks requests which operate on settings. That could be similar to the `IndicesRequest` interface:

https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/IndicesRequest.java

Otherwise, there is an elevated risk that developers of new actions are not aware of this handling whenever actions are introduced which handle settings.

src/main/java/org/opensearch/security/configuration/SettingsPermissionValveImpl.java
```diff
@@ -0,0 +1,206 @@
...
+            listener.onFailure(
+                new OpenSearchSecurityException("User not authorized to use these settings during index creation/update: " + requestSettings.keySet(), RestStatus.FORBIDDEN)
+            );
+            return false;
+        }
+
+        return true;
+    }
+
+    private Set<String> getAllowedSettingsFromRoles(PrivilegesEvaluationContext context) {
```
nibix, line 149: We do not support pattern matching on names of settings this way. Do we want to support pattern matching?

src/main/java/org/opensearch/security/configuration/SettingsPermissionValveImpl.java
```diff
@@ -0,0 +1,206 @@
...
+        Set<String> allowedSettings = new HashSet<>();
+        SecurityDynamicConfiguration<RoleV7> roles = rolesConfiguration.get();
+
+        if (roles != null && roles.getCEntries() != null) {
+            for (String role : context.getUser().getRoles()) {
```
nibix, line 154: Role handling in the security plugin can be confusing as there are many different roles :)

Here, one should use `privilegesEvaluationContext.getMappedRoles()` instead. These will be the roles that went through the role mapping and role injection process. `user.getRoles()` would only return the backend roles of a user.

src/main/java/org/opensearch/security/configuration/SettingsPermissionValveImpl.java
```diff
@@ -0,0 +1,206 @@
...
+                    List<RoleV7.Index> indexPermissions = roleConfig.getIndex_permissions();
+                    if (indexPermissions != null) {
+                        for (RoleV7.Index indexPermission : indexPermissions) {
+                            Set<String> indexSettings = indexPermission.getAllowed_settings();
+                            if (indexSettings != null) {
+                                allowedSettings.addAll(indexSettings);
+                            }
+                        }
+                    }
+                }
```
nibix, line 173: I am not sure if I understand this correctly: We are merging here the settings from index permissions and cluster permissions. Don't we need to differentiate?

src/test/java/org/opensearch/security/configuration/SettingsPermissionValveTests.java
```diff
@@ -0,0 +1,362 @@
...
+import org.opensearch.common.settings.Settings;
+import org.opensearch.security.test.SingleClusterTest;
+import org.opensearch.security.test.helper.rest.RestHelper;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class SettingsPermissionValveTests extends SingleClusterTest {
```
nibix, line 28: I would recommend to use the int test style like it is used here:

https://github.com/opensearch-project/security/blob/main/src/integrationTest/java/org/opensearch/security/DoNotFailOnForbiddenTests.java

That way, you can create all configuration using a declarative style. You won't need to create a new directory containing test resources then (experience has shown that these test resource directories are quite hard to manage).

src/main/java/org/opensearch/security/configuration/SettingsPermissionValveImpl.java
```diff
@@ -0,0 +1,206 @@
...
+
+        // For backwards compatibility we will allow all settings if no allowed settings are defined
+        if (allowedSettings.isEmpty()) {
+            return true;
+        }
+        log.debug("Allowed settings: {} for user: {}", allowedSettings, context.getUser().getName());
+
+        // Validate persistent settings
+        if (!validateSettingsMap(request.persistentSettings(), allowedSettings)) {
+            auditLog.logMissingPrivileges(context.getAction(), request, context.getTask());
```
nibix, line 106: I guess we need to introduce a new audit logging category here. This way, it would look in the audit logs like the user is missing the action privilege indicated by `context.getAction()`.

src/main/java/org/opensearch/security/configuration/SettingsPermissionValveImpl.java
```diff
@@ -0,0 +1,206 @@
...
+        ClusterUpdateSettingsRequest request,
+        ActionListener<?> listener
+    ) {
+        // Get allowed settings patterns from user's roles
+        Set<String> allowedSettings = getAllowedSettingsFromRoles(context);
+
+        // For backwards compatibility we will allow all settings if no allowed settings are defined
+        if (allowedSettings.isEmpty()) {
+            return true;
+        }
```
nibix, line 101: Do we need a way to indicate that a user is not allowed to modify settings at all?

nibix (commented):
Thank you for this PR! 

I know the discussion on the concept is still ongoing. Yet, I think, it cannot hurt to give this a first pass.

The security plugin has a couple of complexities and oddities; I  have added a couple of comments to clarify these things.

DarshitChanpura:
@sam-herman is this PR still being worked upon?

PR #5240, opened 2025-04-02
[Resource Access Control] [Part2] Updates SPI to introduce a ResourceSharingClient interface and implements it within security plugin

### LINKED ISSUE #4500: [Proposal] Resource Permissions
### Status Quo

Currently, there is no mechanism provided by the security plugin for plugins to utilize to provide fine-grained access control to own resources created by each plugin. Some plugins have addressed this by implementing their own custom authorization mechanism, however it is not sustainable for each plugin to handle resource-level FGAC using their own custom implementation. For example, in [anomaly detection](https://opensearch.org/docs/latest/observing-your-data/ad/security/), granting permissions to delete detectors requires the user to be granted `cluster:admin/opendistro/ad/detector/delete`, which allows the user to delete all detectors in the cluster. Anomaly Detection has implemented their own authz mechanism. To avoid this the Security plugin should be the only place where authorization happens. To address this, we are introducing a new concept of Resource Permissions. This feature adds an additional step to the authorization evaluation, allowing plugins to define resource-specific permissions.

### **Feature Implementation**

Once plugins define the resource permissions associated with resources, cluster admins can assign the relevant resource permissions to roles, which will then be evaluated within the Security plugin. Resource permissions can be defined in two ways:

1. Starts with `resource:<plugin-identifier>/<permission>`  **(proposed)**
2. Starts with `<plugin-identifier>:<permission>`

The main difference is the presence of the standard prefix `resource:`, **which simplifies validation [here](https://github.com/opensearch-project/OpenSearch/blob/2069bd805804dd93bd69695a4c6521cc6f2b9bb6/server/src/main/java/org/opensearch/transport/TransportService.java#L1097)**. The structure would look like:

```
# Allows users to use all Anomaly Detection functionality on selected resources, i.e., detectors
anomaly_full_access:
  reserved: true
  cluster_permissions:
    - 'cluster_monitor'
    - 'cluster:admin/opendistro/ad/*'
  index_permissions:
    - index_patterns:
        - '*'
      allowed_actions:
        - 'indices_monitor'
        - 'indices:admin/aliases/get'
        - 'indices:admin/mappings/get'
  resource_permissions:
    - identifier_pattern:
        - 'detector1'
        - 'detector2'
      allowed_actions:
        - 'resource:ad/detectors/read'
```

**This allows admins to add a layer of control, restricting access to selected detectors instead of all detectors. Plugins like [alerting](https://opensearch.org/docs/latest/observing-your-data/alerting/security/), [ml-commons](https://quip-amazon.com/CBKNACJHvnRY/ML-Commons-Security-model-access-control), and flow-framework can similarly benefit by defining permissions at the resource level.**

### **Evaluation Flow Diagram**

* ResourceAccessEvaluator.java
```mermaid
flowchart TD
    A[Resource Access Evaluator.evaluate]
    A --> B{Does request have resources defined?}
    B -->|No| C[Return response as incomplete]
    B -->|Yes| D{Do roles have resource_permissions defined?}
    D -->|No| E[Return response as denied and mark complete]
    D -->|Yes| F[Evaluate resource_permissions for all roles]
    F --> G{Do resource_permissions match with supplied resource names?}
    G -->|Yes| H[Return response as allowed and mark complete]
    G -->|No| E
    H --> J[Return presponse]
    E --> J
    C --> J
```

* PrivilegesEvaluator.java’s evaluate() method with resource permissions check:
```mermaid
flowchart TD
    A[PrivilegesEvaluator.evaluate]
    A --> B[Resolve roles]
    B --> C{Is bulk request for default tenant?}
    C -->|Yes| D[Evaluate and return presponse]
    C -->|No| E[Resolve all indices in the request]
    E --> F[Evaluate snapshot restore request]
    F -->|Complete| G[Return response]
    F -->|Incomplete| H[Evaluate security index access request]
    H -->|Complete| I[Return response]
    H -->|Incomplete| J[Evaluate protected index access request]
    J -->|Complete| K[Return response]
    J -->|Incomplete| L[Evaluate Resource Access request]
    L -->|Complete| M[Return response]
    L -->|Incomplete| N[Evaluate PIT request]
    N -->|Complete| O[Return response]
    N -->|Incomplete| P{Is cluster permission request?}
    P -->|Yes| Q[Evaluate cluster-level permissions and return response]
    P -->|No| R[Evaluater terms aggregation request]
    R -->|Complete| S[Return response]
    R -->|Incomplete| T[Perform DNFOF evaluation if enabled]
    T --> U[Return result]
    style L fill:green,stroke:#333,stroke-width:4px
```

### **Design Questions**

* Are we allowing a mix of roles with and without resource permissions?
    * If yes, what is the default behavior when none of the roles resolved for a user contain a resource permission?
    * Should the user be allowed permission to that resource?
        * If yes, there won’t be backward compatibility issues, but it is not ideal as the feature can be circumvented by not adding any resource permissions.
        * **(Proposed)** If no, there might be backward compatibility issues, especially in mixed cluster requests, which need thorough testing.
* Should we allow `*`  (match_all) patterns for accessing resources? OR should they be individually listed under the role?
    * *Note: `detector_c*`  can still be allowed but the question here is, should  `*`  be prohibited as a resource_name pattern.*
* Should `ressource_permissions` be hidden behind feature flag?
    * No, the primary purpose of introducing this feature is to move the authorization logic, implemented by individual plugins with resources, to Security plugin.

### File Changes

1. [ActionRequest.java](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/ActionRequest.java)
    1. This base class is extended by every plugin’s implementation of transport request calls. Hence, the proposal is to add a standard class property `List<String> resources` . This should then be implemented by plugins to populate this property when handling an actionRequest after which [SecurityFilter.java](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/filter/SecurityFilter.java) will intercept this call to perform privilege evaluation. *(It is important to note this change when creating a documentation to onboard plugin developers to this change)*
        1. *Note: this is done for standardization purposes, however this can be skipped and left on the plugin developers to add this property to each request class. (Not recommended)*
2. [TransportService.java](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/transport/TransportService.java)
    1. Add `resource:`  as a valid transport action prefix [here](https://github.com/opensearch-project/OpenSearch/blob/2069bd805804dd93bd69695a4c6521cc6f2b9bb6/server/src/main/java/org/opensearch/transport/TransportService.java#L1097).
3. [RoleV7.java](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/impl/v7/RoleV7.java)
    1. This class will now define a new type of class: `Resource` (Use the class [Index](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/impl/v7/RoleV7.java#L102) as reference)
    2. Add this newly defined class as a property to the parent
4. ResourceAccessEvaluator.java *(new file)*
    1. A new class that defines the evaluator method to verify the resource permissions.
    2. This method should assume the resource names are passed in the request (the string list defined above).
    3. **If no resources are present in the list, it should assume that the plugin has not opted into this feature and so should bypass this authorization check to return a success response.**
5. [PrivilegeEvaluator.java](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java)
    1. Modify [evaluate()](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L229) method to add a check that calls the evaluator defined above.

### **Test Plan**

1. Add local tests in the security plugin.
2. Add mixed cluster tests. (To evaluate requests coming in during rolling-upgrade scenario)
3. Add tests in anomaly detection or other plugins that could leverage this feature.

### **Will the feature require a security review?**

This feature will require a security review to confirm that it correctly evaluates the scope of the requested resource.

### Documentation

The documentation should be added for this feature detailing how users can utilize this feature.

--

## _Next Steps_

- [ ] Open a META issue that tracks this feature
    - [ ] Add a task to update TransportService, in core repo, to accept a new action type prefix
    - [ ] Open a task to add this feature in Security Plugin (preferred if done via feature branch)
    - [ ] Open a follow-up issue to add tests for this feature
    - [ ] Open a PoC PR in anomaly detector plugin to demonstrate how this feature allows admins to allocate permissions to individual resources.

DarshitChanpura:
Continuation of #5194 with a no-client approach.

#5016 is being broken down into smaller pieces. This is part 2.

### Description
Modifies SPI to add a client to invoke resource-access-control APIs.

There are 4 java APIs as well as 4 REST APIs introduced as part of this PR. Plugins will leverage the client to call the java APIs to implement resource access control.

Refer to the [RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md](https://github.com/DarshitChanpura/security/blob/resource-sharing-no-client/RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md file to understand in-depth implementation of this feature.

### Issues Resolved
- Resolves #4500

### Testing
- automated tests + manual tests

### Check List


src/main/java/org/opensearch/security/resources/ResourceIndexListener.java
```diff
@@ -0,0 +1,107 @@
...
+            log.debug("Skipping resource sharing entry creation as this was an update operation for resource {}", resourceId);
+            return;
+        }
+
+        final UserSubjectImpl userSubject = (UserSubjectImpl) threadPool.getThreadContext()
+            .getPersistent(ConfigConstants.OPENDISTRO_SECURITY_AUTHENTICATED_USER);
+        final User user = userSubject.getUser();
+
+        try {
+            Objects.requireNonNull(user);
```
DarshitChanpura, line 72: Continuing: https://github.com/opensearch-project/security/pull/5194#discussion_r2020484813 @nibix

src/main/java/org/opensearch/security/resources/ResourceAccessControlClient.java
```diff
@@ -0,0 +1,240 @@
...
+     */
+    private String getFeatureDisabledReason() {
+        boolean sharingEnabled = settings.getAsBoolean(
+            ConfigConstants.OPENSEARCH_RESOURCE_SHARING_ENABLED,
+            ConfigConstants.OPENSEARCH_RESOURCE_SHARING_ENABLED_DEFAULT
+        );
+
+        if (!sharingEnabled) return "ShareableResource Access Control feature is disabled.";
+        Version nodeVersion = clusterService.localNode().getVersion();
+        if (nodeVersion.before(RESOURCE_SHARING_MIN_SUPPORTED_VERSION)) {
```
DarshitChanpura, line 189: Continuing discussion:https://github.com/opensearch-project/security/pull/5194#discussion_r2025237637 @nibix

src/main/java/org/opensearch/security/resources/ResourceAccessControlTransportClient.java
```diff
@@ -0,0 +1,257 @@
...
+import static org.opensearch.security.resources.ResourceSharingConstants.RESOURCE_SHARING_MIN_SUPPORTED_VERSION;
+
+/**
+ * Access control transport client responsible for handling resource sharing operations such as verifying,
+ * sharing, revoking, and listing access to shareable resources.
+ * NOTE: The calls are gated behind `cluster:admin/security/resource_access/*` action permissions
+ *
+ * @opensearch.experimental
+ */
+public final class ResourceAccessControlTransportClient implements ResourceSharingClient {
```
DarshitChanpura, line 49: This class is not used and is added for future purposes in-case we decide to gate the transport calls for resource access-control.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1125,6 +1183,33 @@ public Collection<Object> createComponents(
...
+        // Assign resource sharing client to each extension
+        // Using the non-gated client (i.e. no additional permissions required)
+        ResourceSharingClient resourceAccessControlClient = new ResourceAccessControlClient(
+            resourceAccessHandler,
+            settings,
+            clusterService
+        );
+        resourcePluginInfo.getResourceSharingExtensions().forEach(extension -> {
+            extension.assignResourceSharingClient(resourceAccessControlClient);
+        });
```
DarshitChanpura, line 1210: Continuing discussion: https://github.com/opensearch-project/security/pull/5194#discussion_r2025308079 @nibix @cwperks

src/test/java/org/opensearch/security/IndexIntegrationTests.java
```diff
@@ -846,19 +846,16 @@ public void testIndexResolveMinus() throws Exception {
...
 
-        resc = rh.executeGetRequest("/*,-*security/_search", encodeBasicHeader("foo_all", "nagilum"));
+        resc = rh.executeGetRequest("/*,-*security,-foo*,-*resource*/_search", encodeBasicHeader("foo_all", "nagilum"));
         assertThat(resc.getStatusCode(), is(HttpStatus.SC_OK));
 
-        resc = rh.executeGetRequest("/*,-*security,-foo*/_search", encodeBasicHeader("foo_all", "nagilum"));
-        assertThat(resc.getStatusCode(), is(HttpStatus.SC_OK));
-
-        resc = rh.executeGetRequest("/_all,-*security/_search", encodeBasicHeader("foo_all", "nagilum"));
+        resc = rh.executeGetRequest("/_all,-*security,-*resource*/_search", encodeBasicHeader("foo_all", "nagilum"));
```
derek-ho, line 855: nit: here why are you removing one of the cases/should the only modifications be adding -*resource*?

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -382,6 +385,10 @@ public enum RolesMappingResolution {
     // Variable for initial admin password support
     public static final String OPENSEARCH_INITIAL_ADMIN_PASSWORD = "OPENSEARCH_INITIAL_ADMIN_PASSWORD";
 
+    // Resource sharing feature-flag
+    public static final String OPENSEARCH_RESOURCE_SHARING_ENABLED = SECURITY_SETTINGS_PREFIX + "resource_sharing.enabled";
+    public static final boolean OPENSEARCH_RESOURCE_SHARING_ENABLED_DEFAULT = true;
```
derek-ho, line 390: I see this is by design but just want to discuss with other reviewers - is there any concern with having this feature be default enabled? I do not see too much concern with it since it would be opt in only via implementing the SPI.

RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md
```diff
@@ -0,0 +1,691 @@
...
+| Field          | Type    | Description |
+|---------------|---------|-------------|
+| `sharing_info` | Object  | Contains information about the updated resource sharing state. |
+| `source_idx`   | String  | The OpenSearch index containing the resource. |
+| `resource_id`  | String  | The unique identifier of the resource. |
+| `created_by`   | Object  | Information about the user who created the sharing entry. |
+| `share_with`   | Object  | Defines users, roles, and backend roles that still have access to the resource. |
+
+
+### **4. List Accessible Resources**
```
cwperks, line 630: I think this is better left for each resource plugin to define the endpoints for their plugin. What's the advantage of having a REST API in the security plugin for this?

RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md
```diff
@@ -0,0 +1,691 @@
...
+---
+
+## **9. REST APIs Introduced by the Security Plugin**
+
+In addition to client methods, the **Security Plugin** introduces new **REST APIs** for managing resource access when the feature is enabled. These APIs allow users to **verify, grant, revoke, and list access** to resources.
+
+### **1. Verify Access**
+- **Endpoint:**
+  ```
+  POST /_plugins/_security/resources/verify_access
```
cwperks, line 474: In what instances would plugins call this REST API vs using the service class that has a similar API?

One API I could see useful for powering frontend UX for sharing would be an API to get the sharing info of a current resource.

i.e. in Quip if I own a document, I can click on the share button and it will show me who the document has been shared with an at what access levels.

src/main/java/org/opensearch/security/support/ConfigConstants.java
cwperks: experimental feature flags should be false while experimental. We can certainly change default values in the future in major releases.

spi/src/main/java/org/opensearch/security/spi/resources/client/NoopResourceSharingClient.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * Throws 501 exception as a no-op implementation.
+     *
+     * @param resourceIndex The index to search for accessible resources.
+     * @param listener      Callback receiving a set of {@link ShareableResource} instances.
+     */
+    @Override
+    public <T extends ShareableResource> void listAllAccessibleResources(String resourceIndex, ActionListener<Set<T>> listener) {
+        handleSecurityDisabled("Unable to list all accessible resources.", listener);
+    }
```
derek-ho, line 96: Do we want to just return all instead of not returning anything? Similar to verifyResourceAccess method?

src/test/java/org/opensearch/security/IndexIntegrationTests.java
DarshitChanpura: its a duplicate.

src/main/java/org/opensearch/security/support/ConfigConstants.java
DarshitChanpura: This feature flag is for initial versions. It is intended to be permanently enabled.

On another hand, we can simply say that if security is enabled then we assume that the feature is enabled. For now, i have kept these as separate.

RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md
DarshitChanpura: This simply allows listing all resources you have access to. Like say you open a photo app. You should be able to see all photos shared with you regardless of your permission level. However you should not be able to see the photos that have not been shared with you at all.

RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md
```diff
@@ -0,0 +1,691 @@
...
+}
+```
+
+#### **Response Fields:**
+| Field            | Type    | Description |
+|-----------------|---------|-------------|
+| `has_permission` | Boolean | `true` if the user has access, `false` otherwise. |
+
+
+### **2. Grant Access**
```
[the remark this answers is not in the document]
DarshitChanpura: They accept a different structure in arguments. If you were to remove user from multiple groups you will have to pass multiple PATCH operations for same user on different action-groups.

spi/src/main/java/org/opensearch/security/spi/resources/client/NoopResourceSharingClient.java
DarshitChanpura: I feel like this should be handled by plugin, but I'm open to suggestions. Remember, if we decide to implement it here, we will need a parser which can only be passed when registering itself as resource-sharing plugin. We store these inside ResourcePluginInfo in sec plugin, but in No-op case we would need that information to be passed inside this API.

RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md
cwperks: But doesn't the photo APP define the API it needs? What's security's role. My understanding is that the ResourceSharingClient provided to the plugin can be used by the plugin for filtering resources visible to the user.

DarshitChanpura: It does but how does it fetch all accessible photos. We don't have a filtering mechanism (aside from DLS, but we benched that approach) for plugins to call a list api and they be able to list all photos. I've implemented this usage in SamplResourcePlugin.

src/main/java/org/opensearch/security/resources/ResourceIndexListener.java
nibix: I have written a small test (see below) to test whether `postIndex()` is only called once per doc or once per shard.

It's basically this:

```
    @Test
    public void index() {
        try (Client client = cluster.getInternalNodeClient()) {
            client.admin().indices().create(new CreateIndexRequest("test").settings(Settings.builder().put("index.number_of_shards", 3).put("index.number_of_replicas", 3).build())).actionGet();
            client.index(new IndexRequest("test").id("test_doc").source(Map.of("test", "data"))).actionGet();
        }
    }

    public static class TestPlugin extends Plugin  {
        @Override
        public void onIndexModule(IndexModule indexModule) {
            if (indexModule.getIndex().getName().equals("test")) {
                indexModule.addIndexOperationListener(new IndexingOperationListener() {

                    @Override
                    public void postIndex(ShardId shardId, Engine.Index index, Engine.IndexResult result) {
                        System.out.println("postIndex " + shardId + " id: " + index.id() + " origin: " + index.origin() + " created: " + result.isCreated());
                    }
                });
            }
        }
    }
```

The output is:

```
postIndex [test][0] id: test_doc origin: PRIMARY created: true
postIndex [test][0] id: test_doc origin: REPLICA created: true
```

So, it is called one additional time for a replica shard (even though I am confused why it is only called once, given we requested 3 replicas. but anyway).

So, in order to make sure that the ResourceSharing document is only created once, I'd recommend to add the check `if (index.origin() == Origin.PRIMARY)`

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
nibix: My comment from https://github.com/opensearch-project/security/pull/5194#discussion_r2033139112

After a while of back and forth thinking, I must say that - even though this solution feels a bit fragile - this is probably the best solution we can get. 

So, in order to get this thing moving, we can go with this solution for now.

Still, I would recommend two things:

- File an RFC in core discussing this dependency situation and asking for opinions how this is achieved best and what could be done in core to improve the situation. 
- I think we still need to work a bit on the "disabled security" or "diabled resource sharing access control" case; in these cases, the `ResourceSharingClient` reference would be just `null`, right? So, a `null`  `ResourceSharingClient`  would mean: skip access controls. This is a bit fragile, as a  `ResourceSharingClient`   could be also `null` for other reasons, for example due to exceptions in the code path leading to the initialization. I would recommend to check whether we can find solutions/patterns which can increase robustness in this regard a bit.

src/main/java/org/opensearch/security/resources/ResourceAccessControlTransportClient.java
nibix: IMHO, we should not add unused code to the code base. I'd recommend to keep it in a branch and add it when it is actually needed.

src/main/java/org/opensearch/security/support/ConfigConstants.java
nibix: I am also a bit worried regarding the "experimental" status. I tried to find some guidelines regarding experimental features.

There's this page in the docs: https://opensearch.org/docs/latest/install-and-configure/configuring-opensearch/experimental/ ... however it is not too helpful.

There is another issue where a discussion takes place whether experimental features should be enabled or disabled: https://github.com/opensearch-project/OpenSearch/issues/13411 .. the consensus there is that experimental features should be not enabled by default.

So, I guess we have two choices here:

- Disable the feature by default
- Do not mark the feature as experimental

RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md
nibix: > It does but how does it fetch all accessible photos. 

Cant we expose a method in the client that yields the IDs of all accessible photos?

In any case, a photo app also needs to provide an endpoint to list all resources, if it wants to remain usable in case the security plugin is disabled (and therefore the `GET /_plugins/_security/resources/list/{resource_index}` endpoint does not exist).

Any external client will likely use the photo app specific endpoint, as it also would want to support the "security disabled" case.

RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md
DarshitChanpura: okay, I will remove the code.

RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md
DarshitChanpura: > Cant we expose a method in the client that yields the IDs of all accessible photos?

We do that and a step further by listing the photos itself. If the consensus is to only return IDs I will go ahead and make that change.

src/main/java/org/opensearch/security/support/ConfigConstants.java
DarshitChanpura: For this iteration, I will update it to be disabled by default.

DarshitChanpura: just  a note. If this feature is enabled it will be enabled for all plugins.

src/main/java/org/opensearch/security/resources/ResourceAccessControlTransportClient.java
DarshitChanpura: removed.

RESOURCE_ACCESS_CONTROL_FOR_PLUGINS.md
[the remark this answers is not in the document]
DarshitChanpura: Resolving as APIs have been removed in this iteration.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: > File an RFC in core discussing this dependency situation and asking for opinions how this is achieved best and what could be done in core to improve the situation.

Opened this PR: https://github.com/opensearch-project/OpenSearch/pull/17893

> I think we still need to work a bit on the "disabled security" or "diabled resource sharing access control" case; in these cases, the ResourceSharingClient reference would be just null, right? So, a null ResourceSharingClient would mean: skip access controls. This is a bit fragile, as a ResourceSharingClient could be also null for other reasons, for example due to exceptions in the code path leading to the initialization. I would recommend to check whether we can find solutions/patterns which can increase robustness in this regard a bit.

Explained in SPI readme, on how to handle the null reference to ResourceSharingClient. [link](https://github.com/opensearch-project/security/pull/5240/files#diff-c2a8ed85c34a39abf525178db791719c19d70d19d6319c0e5421af4da5e83f45R136)

src/main/java/org/opensearch/security/resources/ResourceAccessControlClient.java
DarshitChanpura: You are correct.  I'll remove this version control check.

spi/README.md
```diff
@@ -16,24 +16,24 @@ public class SampleResourcePlugin extends Plugin implements SystemIndexPlugin, R
...
 
     @Override
-    public String getResourceType() {
-        return SampleResource.class.getCanonicalName();
+    public Set<ResourceProvider> getResourceProviders() {
+      return Set.of(
+              new ResourceProvider(
+                      SampleResource.class.getCanonicalName(), // class-name of the resource
+                      RESOURCE_INDEX_NAME,                     // the index that stores resource, **must only store the type of resource defined in the line above**
+                      new SampleResourceParser()               // parser to parse the resource
```
cwperks, line 29: Are you still able to retain the ResourceProvider but also have a `compileOnly` dependency on the SPI in the extending plugin?

I think the SPI needs to be compileOnly for plugin that optionally extend security.

spi/src/main/java/org/opensearch/security/spi/resources/client/NoopResourceSharingClient.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * Throws 501 exception as a no-op implementation.
+     *
+     * @param resourceId    The ID of the resource to share.
+     * @param resourceIndex The index containing the resource.
+     * @param recipients    The recipients of the resource, including users, roles, and backend roles.
+     * @param listener      Callback receiving the updated {@link ResourceSharing} document.
+     */
+    @Override
+    public void shareResource(
```
cwperks, line 60: Can we shorten these method names to names like `share` or `revoke`. This is already a ResourceSharingClient so having resource in the method names is overly verbose.

When is `verifyResourceAccess` used?

spi/README.md
```diff
@@ -112,6 +111,263 @@ Ensure that your **plugin declaration class** implements `ResourceSharingExtensi
...
+protected void doExecute(Task task, DeleteResourceRequest request, ActionListener<DeleteResourceResponse> listener) {
+  String resourceId = request.getResourceId();
+  if (resourceId == null || resourceId.isEmpty()) {
+    listener.onFailure(new IllegalArgumentException("Resource ID cannot be null or empty"));
+    return;
+  }
+
+  // Check permission to resource
+  ResourceSharingClient resourceSharingClient = ResourceSharingClientAccessor.getResourceSharingClient();
+  resourceSharingClient.verifyResourceAccess(resourceId, RESOURCE_INDEX_NAME, ActionListener.wrap(isAuthorized -> {
```
cwperks, line 165: Do this still make sense to have absent of the `scopes` change that you had earlier?

Authorization will be done as it is currently in the SecurityFilter and based on `cluster_permissions`. In what scenarios would a plugin dev use this method?

src/main/java/org/opensearch/security/resources/rest/revoke/RevokeResourceAccessRequest.java
```diff
@@ -0,0 +1,181 @@
...
+import org.opensearch.action.ActionRequest;
+import org.opensearch.action.ActionRequestValidationException;
+import org.opensearch.common.xcontent.LoggingDeprecationHandler;
+import org.opensearch.common.xcontent.XContentFactory;
+import org.opensearch.common.xcontent.XContentType;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.security.spi.resources.ResourceAccessActionGroups;
```
cwperks, line 27: I think we should minimize what's in the SPI vs what's in the security repo. SPI should be the minimum set of files (mostly or exclusively interfaces necessary for the plugins that optionally extend this SPI). 

SharedWithActionGroup is quite a verbose name, can we encapsulate all sharing info into a single data structure. Maybe `SharingInfo` that contains all information with regards to sharing.

spi/README.md
DarshitChanpura: without `implementation` declaration on SPI, there would be RunTimeException (NoClassDefFoundError) and prevent cluster from bootstrapping.

For now these need to be `implementation`, unless we get rid of ShareableResource and ShareableResourceParser interfaces, which are needed for listing all resources API, and without these contracts, resources could be of any type.

spi/src/main/java/org/opensearch/security/spi/resources/client/NoopResourceSharingClient.java
DarshitChanpura: > Can we shorten these method names to names like share or revoke. This is already a ResourceSharingClient so having resource in the method names is overly verbose.

Sure.

> When is verifyResourceAccess used?

Here is a usage in [ADPlugin](https://github.com/opensearch-project/anomaly-detection/pull/1400/files#diff-cf370d40fdd2abc404283fa1e37768646f30ee1b8bfbb0b5c5302fc8a2a080fcR721). Another one in [ml-plugin](https://github.com/opensearch-project/ml-commons/pull/3715/files#diff-7e30388762cbeaaf7af2391b1f6a623882537b3363ba3d3801f62223b98097aaR173).

spi/README.md
DarshitChanpura: I'm confused. Are you stating the individual resource access will be evaluated at cluster level?

src/main/java/org/opensearch/security/resources/rest/revoke/RevokeResourceAccessRequest.java
DarshitChanpura:
```
{
   "source_idx": ".plugins-ml-model-group",
   "resource_id": "model-group-123",
   "created_by": {
      "user": "darshit"
   },
   "share_with": {
      "action-group1": {
         "users": ["user1", "user2"],
         "roles": ["viewer_role"],
      },
      "action-group2": {
         "users": ["admin_user"],
         "roles": ["editor_role"],
      }
   }
}
```

This data-structure has been split into:
```java
 * The index where the resource is defined
     */
    private String sourceIdx;

    /**
     * The unique identifier of the resource
     */
    private String resourceId;

    /**
     * Information about who created the resource
     */
    private CreatedBy createdBy;

    /**
     * Information about with whom the resource is shared with
     */
    private ShareWith shareWith;
```
here. https://github.com/opensearch-project/security/blob/feature/resource-permissions/spi/src/main/java/org/opensearch/security/spi/resources/sharing/ResourceSharing.java

I have stripped SPI of everything that is not needed. Moving the classes inside `sharing` package in SPI to security will result in cyclic dependency between SPI and security as share and revoke methods require reference to ResourceSharing class and list method references ShareableResourceClass.

spi/README.md
DarshitChanpura: I have changed it to be compileOnly and converted the listResources API to getResourceIds.

DarshitChanpura:
@nibix @cwperks @derek-ho The SPI is now required to be declared as compileOnly instead of implementation. This was done by removing the requirement for ResourceParser and converting listAccessibleResources API from returning actual resources to just resourceIds.

spi/README.md
cwperks: Got it, this fn is used to determine if the resource has been shared w/ the current user.

.github/workflows/ci.yml
```diff
@@ -243,12 +208,18 @@ jobs:
       - name: Checkout security
         uses: actions/checkout@v4
 
+      - name: Publish components to Maven Local
```
cwperks, line 211: Are these steps required?

spi/src/main/java/org/opensearch/security/spi/resources/ResourceSharingExtension.java
```diff
@@ -16,20 +20,17 @@
...
+    Set<ResourceProvider> getResourceProviders();
 
     /**
-     * The index where resource is stored
-     * @return the name of the parent index where resource is stored
+     * Assigns the ResourceSharingClient to the resource plugin. Plugins can then utilize this to call the methods for access control.
+     * @param client the ResourceSharingClient instance
      */
-    String getResourceIndex();
+    void assignResourceSharingClient(ResourceSharingClient client);
```
cwperks, line 34: For plugins that implement this extension, can't we rely on the the ResourceSharingClient never being provided as having either 1) security uninstalled, 2) security disabled or 3) this feature disabled?

I see that there are checks added in the AD PR to see if security is disabled, but I don't think the checks should be present in those repos.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -0,0 +1,1338 @@
...
+     * @param listener    The listener to be notified when the operation completes.
+     *                    The listener receives a set of resource IDs as a result.
+     * @apiNote This method:
+     * <ul>
+     *   <li>Uses source filtering for optimal performance</li>
+     *   <li>Performs exact matching on the source_idx field</li>
+     *   <li>Returns an empty get instead of throwing exceptions</li>
+     * </ul>
+     */
+    public void fetchAllDocuments(String pluginIndex, ActionListener<Set<String>> listener) {
```
cwperks, line 237: I think we should handle this case w/o getting all resource IDs. It looks like this is used in cases where the logged in user has all resources visible to them? (i.e. superuser or cluster admin)

If that's the case then can we communicate with the resource plugin to perform an search unrestricted.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -0,0 +1,506 @@
...
+        LOGGER.debug("Listing accessible resources within the resource index {} for user: {}", resourceIndex, user.getName());
+
+        // 2. If the user is admin, simply fetch all resources
+        if (adminDNs.isAdmin(user)) {
+            loadAllResources(resourceIndex, ActionListener.wrap(listener::onResponse, listener::onFailure));
+            return;
+        }
+
+        // StepListener for the user’s "own" resources
+        StepListener<Set<String>> ownResourcesListener = new StepListener<>();
```
cwperks, line 92: Is it possible to do this in a single query instead of 4 separate queries?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -2203,10 +2281,28 @@ private void tryAddSecurityProvider() {
...
+            for (ResourceSharingExtension extension : loader.loadExtensions(ResourceSharingExtension.class)) {
+                for (ResourceProvider provider : extension.getResourceProviders()) {
+                    resourceProviders.put(provider.resourceIndexName(), provider);
+                    resourceIndices.add(provider.resourceIndexName());
+                }
+                resourceSharingExtensions.add(extension);
+            }
+            resourcePluginInfo.setResourceIndices(resourceIndices);
+            resourcePluginInfo.setResourceProviders(resourceProviders);
+            resourcePluginInfo.setResourceSharingExtensions(resourceSharingExtensions);
```
cwperks, line 2304: wdyt about only having this setter and then computing everything else within the ResourcePluginInfo class?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1125,6 +1151,29 @@ public Collection<Object> createComponents(
...
+        final var resourceSharingIndex = ResourceSharingConstants.OPENSEARCH_RESOURCE_SHARING_INDEX;
+        ResourceSharingIndexHandler rsIndexHandler = new ResourceSharingIndexHandler(resourceSharingIndex, localClient, threadPool);
+
+        ResourceAccessHandler resourceAccessHandler = new ResourceAccessHandler(threadPool, rsIndexHandler, adminDns, resourcePluginInfo);
+        // Resource Sharing index is enabled by default
+        boolean isResourceSharingEnabled = settings.getAsBoolean(
+            FeatureConfigConstants.OPENSEARCH_RESOURCE_SHARING_ENABLED,
+            FeatureConfigConstants.OPENSEARCH_RESOURCE_SHARING_ENABLED_DEFAULT
+        );
+        resourceSharingIndexManagementRepository = ResourceSharingIndexManagementRepository.create(
```
cwperks, line 1163: Is this wrapper class needed? Looks like you can call `this.resourceSharingIndexHandler.createResourceSharingIndexIfAbsent` directly down below. What is the purpose of the callable as an arg to that method?

cwperks (commented):
Thank you @DarshitChanpura, I think most large structural items have been addressed. I think we should keep references to the feature flag only in this repo and not expose those outside the repo.

I am having trouble reviewing the queries in the ResourceSharingIndexHandler where its computing the resources visible to the current user, but the code and the intent is easy to follow.

I think this PR is a big step forward towards sensible authz for resources created by plugins where this lays the groundwork towards removing the copies of the users stored across various plugin system indices. Even with the placeholder for action groups at this point, these set of PRs can be built upon towards the desired end state of resource authorization where the owner of the resource gets to specify the access level when sharing. 

While this change is large, it is done in a way not to cause disruption to the certain system and is enabled by a new feature flag. This is particularly good for anyone implementing a plugin from scratch to have a mechanism they can leverage from the security plugin and also allows for an opt-in model where existing plugins can opt into this mechanism from the security plugin to avoid a lot of existing code duplication around the plugin ecosystem.

I'm very eager to see the next steps after this initial groundwork to truly move towards the new model of resource authorization. This is a benefit for 1) OpenSearch users that will be empowered to control how their resources are shared, 2) plugin developers which can now delegate resource authz to the security plugin and 3) cluster administrators will not be bottlenecked figuring out how to do resource sharing using the existing filter_by_backend_role mechanism.

This PR is against a feature branch and a separate PR will be done to the main branch.

spi/src/main/java/org/opensearch/security/spi/resources/ResourceSharingExtension.java
DarshitChanpura: > I see that there are checks added in the AD PR to see if security is disabled, but I don't think the checks should be present in those repos.

The checks are meant to be for the feature to be disabled. If you do see a security enabled/disabled check then it is a miss on my end. Would you mind pointing me to the exact code change?

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: we are only returning the ids. This method fetches resource-ids from the individual entries in resource-sharing-index and not the pluginIndex.

i;ve renamed this to make it clearer.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -0,0 +1,1338 @@
...
+     * </pre>
+     *
+     * @param pluginIndex   The source index to match against the source_idx field
+     * @param entities      Set of values to match in the specified Recipient field
+     * @param recipient     The type of recipient {@link Recipient}
+     * @param actionGroup   The action group to match against the action-group field
+     * @param listener      The listener to be notified when the operation completes.
+     *                      The listener receives a set of resource IDs as a result.
+     */
+    public void fetchDocumentsForAGivenActionGroup(
```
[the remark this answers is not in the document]
DarshitChanpura: removed.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
DarshitChanpura: done.-

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: We have an opportunity to preload the information that will not change, and will allows us to have constant time look-ups instead of computing these from resourceSharingExtensions. I think we should keep it as is.

cwperks:
My highest-level comments have been addressed. Have you been able to organize it in such a way where only this repo refers to the `plugins.security.resource_sharing.enabled` feature flag? 

Marking PR approved to merge into the feature branch, another review will be performed before merging to main.

cwperks (approved):

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: I've followed suit with existing pattern: https://github.com/opensearch-project/security/blob/7603dc28a233a7df72f86537fb11d674f92f5155/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java#L320

This will allows us to separate any index management tasks to its own class. Callable was a placeholder. I have removed extra abstraction since the only method offer by the repository was to create the index. I've replaced it with direct call

DarshitChanpura:
Merging this. All future comments can be addressed in PR towards main.

PR #5281, opened 2025-04-18
[Resource Permissions] Introduces Centralized Resource Access Control Framework

DarshitChanpura:
### Description
Introduces Centralized Resource Access Control framework by declaring a new SPI for plugins to extend ResourceSharingExtension and use ResourceSharingClient to utilize the resource access control APIs. Design outlined in the proposal below.

### Issues Resolved
- Related to #4500 

### Testing
- unit and integration tests, manual testing

### Check List


src/main/java/org/opensearch/security/resources/ResourceAccessControlClient.java
```diff
@@ -0,0 +1,189 @@
...
+import org.opensearch.security.spi.resources.sharing.ShareWith;
+import org.opensearch.security.spi.resources.sharing.SharedWithActionGroup;
+
+/**
+ * Access control client responsible for handling resource sharing operations such as verifying,
+ * sharing, revoking, and listing access to shareable resources.
+ *
+ * @opensearch.experimental
+ */
+public final class ResourceAccessControlClient implements ResourceSharingClient {
```
nibix, line 33: We have significant test coverage holes in this code. I guess this is because the test coverage is supposed to come from the sample plugin - which however is not included in the coverage configuration. Is this correct? Can we change the coverage configuration to include this?

DarshitChanpura: Not sure what exactly is causing coverage drop but it also seems to be coming from files unrelated to this code. I will take a deeper look.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/client/ResourceSharingClientAccessor.java
```diff
@@ -0,0 +1,35 @@
...
+package org.opensearch.sample.resource.client;
+
+import org.opensearch.security.spi.resources.client.NoopResourceSharingClient;
+import org.opensearch.security.spi.resources.client.ResourceSharingClient;
+
+/**
+ * Accessor for resource sharing client supplied by the SPI.
+ */
+public class ResourceSharingClientAccessor {
+    private static ResourceSharingClient CLIENT;
```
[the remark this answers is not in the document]
DarshitChanpura: The getters and setters are static here as the idea is to [load it once](https://github.com/DarshitChanpura/security/blob/3594c5a4213389f0a37dc6cc6a6f72d6f5e14fa7/sample-resource-plugin/src/main/java/org/opensearch/sample/SampleResourceExtension.java#L34) and then use it throughout. No reassignments are intended here. This pattern is used in [other places as well](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java#L88).  

Since loadExtension is called before createComponents() we cannot bind the instance of Accessor for DI later.  Since in our case, this client could either be null (later assigned to Noop) or ResourceAccessControlClient, I couldn't find a way to signal the main plugin class if the loaded extension class had received a null reference (in case, where security or feature is disabled) or ResourceAccessControlClient reference. if you are already aware of a pattern to solve this, please let me know.

src/main/java/org/opensearch/security/resources/ResourceAccessControlClient.java
DarshitChanpura: looks like the coverage file did not process: https://app.codecov.io/gh/opensearch-project/security/commit/adf96f1d377ea673fbe5a31644fa731ffff85623

sample-plugin is already added as part of report-coverage task: https://github.com/DarshitChanpura/security/blob/3594c5a4213389f0a37dc6cc6a6f72d6f5e14fa7/.github/workflows/ci.yml#L118

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/client/ResourceSharingClientAccessor.java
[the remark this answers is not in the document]
DarshitChanpura: > In any case where several nodes run in one JVM, reassingments will happen, though. This pattern might work for many cases, yet there are situations where this will just break with weird behavior. Of course, one can try to establish the rule "only one node per JVM", but I guess this will take a bit of preparation and work.

I understand, however I didn't find a way to handle `compileOnly` SPI dependency and also keep this non-static.

src/main/java/org/opensearch/security/resources/ResourceAccessControlClient.java
DarshitChanpura: This commit https://github.com/opensearch-project/security/pull/5281/commits/6173cc77e1ea4d368f196d4ccffb5ae3fec9da5b should add coverage for files covered by sample-plugin.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/client/ResourceSharingClientAccessor.java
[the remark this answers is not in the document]
cwperks: I think one way we can avoid static on the getters and setters is to create an instances of the accessor in createComponents and return it in the list of components. At transport it would then be injectable.

wdyt about the call to `getResourceSharingClient` returning `null` in the case where its unassigned? (unassigned could mean security is not installed, disabled or the feature itself is disabled) and then the resource plugin should handle that case accordingly. I'm not convinced atm for the need of the NoopResourceSharingClient class.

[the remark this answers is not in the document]
nibix: Oh, i see. So far, I did not realize that the interface is not present if the security plugin is not present. 

To be honest, I am now also running out of ideas in this regard.  

Still, then I am struggling to understand why other classes in the sample plugin do not suffer from this issue. I will open another comment from this.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/GetResourceTransportAction.java
```diff
@@ -0,0 +1,173 @@
...
+    ) {
+        super(GetResourceAction.NAME, transportService, actionFilters, GetResourceRequest::new);
+        this.transportService = transportService;
+        this.nodeClient = nodeClient;
+        this.settings = settings;
+    }
+
+    @Override
+    protected void doExecute(Task task, GetResourceRequest request, ActionListener<GetResourceResponse> listener) {
+        ResourceSharingClient client = ResourceSharingClientAccessor.getResourceSharingClient();
```
nibix, line 80: Following up to the conversation here: https://github.com/opensearch-project/security/pull/5281#discussion_r2054496615

I just now completely realized that the "optionality" of the security plugin can mean that the classes provided by the security plugin are actually not present in the classpath. 

However, then I do not understand how the code above can work without the security plugin present. If the security plugin is not present, the class `org.opensearch.security.spi.resources.client.ResourceSharingClient` will not be in the class path. I would expect, then class loading of `GetResourceTransportAction` to fail with a `NoClassDefFoundError`. Do I understand something wrong?

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/client/ResourceSharingClientAccessor.java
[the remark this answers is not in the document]
nibix: Follow up at https://github.com/opensearch-project/security/pull/5281#discussion_r2054746763

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/GetResourceTransportAction.java
cwperks: Only if the code is trying to access the client at runtime when the class is not present on the runtime classpath. Its up to the resource plugins to ensure to only access it if provided.

The class can be referenced here because its available at compile time.

nibix: But in the code above, it is accessed whenever the get resource action is executed. Even if the security plugin is not present.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/client/ResourceSharingClientAccessor.java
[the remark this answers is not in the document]
DarshitChanpura: > wdyt about the call to getResourceSharingClient returning null in the case where its unassigned?

Maybe i'm not understanding this, but if we avoid static then we have to make it injectable. How do I make an instance of ResourceSharingClientAccessor available to SampleResourceSharingExtension when extension is loaded. Also, this makes testing significantly hard with current framework as we have no mechanism of loading the extensions.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/GetResourceTransportAction.java
cwperks: @DarshitChanpura for the sample plugin tests can we have 2 tracks?

1. Security installed
2. Security not installed

DarshitChanpura: NoopResourceSharingClient is assigned using the client accessor if security is disabled. SamplePlugin demonstrate a combination usage of the feature flag and using client directly. However, since Noop returns 501 on share and revoke, we use feature flag instead of check whether client should be called.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/client/ResourceSharingClientAccessor.java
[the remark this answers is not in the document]
nibix: TBH, I do not think that the security int test framework is really suitable for testing this. The security int test framework was designed for the special needs when testing the security plugin.

However, as it runs in a single JVM and in a single class loader, it cannot replicate complex plugin loading scenarios. Take just for example the case when the security plugin is not present: I guess we would need to create a separate gradle configuration to run these tests with a classpath where the security plugin is actually not present.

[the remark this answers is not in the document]
cwperks: singleton pattern which is similar to what you already have. Is there a difference between have the CLIENT as a non-static field on a singleton static instance of `ResourceSharingClientAccessor` vs what's in the file now where the CLIENT is a static field of `ResourceSharingClientAccessor`?

One important thing to note is that singletons are only single instance per classLoader and each plugin is loaded with a different classLoader.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/GetResourceTransportAction.java
DarshitChanpura: I've removed NoopClient.

src/main/java/org/opensearch/security/resources/ResourceAccessControlClient.java
DarshitChanpura: Coverage has now gone up significantly for patch. Feel free to close this thread if you do not have any more questions.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/client/ResourceSharingClientAccessor.java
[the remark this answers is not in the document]
DarshitChanpura: Created singleton pattern. Closing this thread.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/GetResourceTransportAction.java
DarshitChanpura: > @DarshitChanpura for the sample plugin tests can we have 2 tracks?
Security installed
Security not installed

They are already present. `SampleResourcePluginSecurityDisabledTests` cover the disabled scenario and rest cover enabled scenarios along with feature enabled/disabled.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/SampleResourcePluginLimitedPermissionsTests.java
```diff
@@ -0,0 +1,456 @@
...
+
+            TestRestClient.HttpResponse response = client.putJson(SAMPLE_RESOURCE_CREATE_ENDPOINT, sampleResource);
+            response.assertStatusCode(HttpStatus.SC_OK);
+
+            resourceId = response.getTextFromJsonBody("/message").split(":")[1].trim();
+        }
+
+        // Create an entry in resource-sharing index
+        try (TestRestClient client = cluster.getRestClient(cluster.getAdminCertificate())) {
+            // Since test framework doesn't yet allow loading ex tensions we need to create a resource sharing entry manually
```
cwperks, line 112: Did we consider finished https://github.com/opensearch-project/OpenSearch/pull/16908 to get this officially supported?

If its officially supported will that obviate the need for the block below:

```
// Also update the in-memory map and get
resourcePluginInfo.getResourceIndicesMutable().add(RESOURCE_INDEX_NAME);
ResourceProvider provider = new ResourceProvider(SampleResource.class.getCanonicalName(), RESOURCE_INDEX_NAME);
resourcePluginInfo.getResourceProvidersMutable().put(RESOURCE_INDEX_NAME, provider);

ResourceSharingClientAccessor.getInstance().setResourceSharingClient(createResourceAccessControlClient(cluster));
```

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/SampleResourcePluginTestHelper.java
```diff
@@ -0,0 +1,76 @@
...
+                "users": ["%s"]
+              }
+            }
+            """.formatted(user);
+    }
+
+    protected static String revokeAccessPayload(String user) {
+        return """
+            {
+              "entities_to_revoke": {
```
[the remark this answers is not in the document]
DarshitChanpura: This is a design that I chose for sample plugin's Revoke API. It should not matter either or. But this gives more structure to my design.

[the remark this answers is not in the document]
cwperks: I see, so if security is not installed or disable then this API becomes a noop?

I think it makes sense to have a generic share api from the security plugin. Something like `_plugins/_security/api/resource/{resource_type}/{resource_id}/share` to handle all cases generally across any resource plugin.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/GetResourceTransportAction.java
nibix: There's a difference between security disabled and security absent. Did we test security absent - and that's in a real cluster, not with int tests?

DarshitChanpura: Yes, I can show a demo over slack if you'd like.

nibix: Thanks 👍

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -0,0 +1,401 @@
...
+                // Since no sharing entry exists, requests is allowed to implement a non-breaking behaviour
+                listener.onResponse(true);
+                return;
+            }
+
+            // All public entities are designated with "*"
+            userRoles.add("*");
+            userBackendRoles.add("*");
+            if (isOwnerOfResource(document, user.getName())
+                || isSharedWithEveryone(document)
```
cwperks, line 166: Why not have this as methods within `ResourceSharing`? i.e. `document.isSharedWithEveryone()`.

Similarly wdyt about other methods like `document.isSharedWithUser()`, `document.isSharedWithRole()`, etc. You can then abstract the handling for the `*` case where the document is public. 

I don't think you need separate handling for `*` because `isSharedWithEveryone` should already return true by the time you evaluate the next conditions in this clause.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -0,0 +1,401 @@
...
+        return document.getShareWith()
+            .getSharedWithActionGroups()
+            .stream()
+            .filter(sharedWithActionGroup -> actionGroups.contains(sharedWithActionGroup.getActionGroup()))
+            .findFirst()
+            .map(sharedWithActionGroup -> {
+                SharedWithActionGroup.ActionGroupRecipients aGs = sharedWithActionGroup.getSharedWithPerActionGroup();
+                Map<Recipient, Set<String>> recipients = aGs.getRecipients();
+
+                return switch (recipient) {
```
cwperks, line 379: Why a `switch - case` here instead of directly calling `recipients.getOrDefault(recipient, Set.of()).contains(entity);`

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -0,0 +1,401 @@
...
+                    case Recipient.USERS, Recipient.ROLES, Recipient.BACKEND_ROLES -> recipients.getOrDefault(recipient, Set.of())
+                        .contains(entity);
+                };
+            })
+            .orElse(false); // Return false if no matching action-group is found
+    }
+
+    private void validateArguments(Object... args) {
```
cwperks, line 387: I think we should use annotations like `@NonNull` instead of having this generic method. When annotating the args in the method signature, it would have the advantage of showing up in the IDE as the arg must be NonNull instead of finding out at runtime.

src/main/java/org/opensearch/security/resources/ResourcePluginInfo.java
```diff
@@ -0,0 +1,73 @@
...
+        resourceProviderMap.clear();
+        resourceProviderMap.putAll(providerMap);
+    }
+
+    public void setResourceIndices(Set<String> indices) {
+        resourceIndices.clear();
+        resourceIndices.addAll(indices);
+    }
+
+    public void setResourceSharingExtensions(Set<ResourceSharingExtension> extensions) {
```
cwperks, line 46: Can we have only this setter and compute the others from this?

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -0,0 +1,401 @@
...
+        if (adminDNs.isAdmin(user)) {
+            loadAllResources(resourceIndex, ActionListener.wrap(listener::onResponse, listener::onFailure));
+            return;
+        }
+
+        // 1) collect all entities we’ll match against share_with arrays
+        // for users:
+        Set<String> users = new HashSet<>();
+        users.add(user.getName());
+        users.add("*"); // for matching against publicly shared resource
```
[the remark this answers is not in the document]
DarshitChanpura: `*` simply annotates comparison against a publicly shared resource, as we defined in the contract of the definition of public resource. It is not used for wild-card comparison using expansion.

src/main/java/org/opensearch/security/resources/ResourcePluginInfo.java
```diff
@@ -0,0 +1,73 @@
...
+        resourceSharingExtensions.clear();
```
cwperks, line 47: Why clear()? Can this be designed in a way that it is set once.

spi/src/main/java/org/opensearch/security/spi/resources/FeatureConfigConstants.java
```diff
@@ -0,0 +1,15 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.security.spi.resources;
+
+public class FeatureConfigConstants {
+    // Resource sharing feature-flag
+    public static final String OPENSEARCH_RESOURCE_SHARING_ENABLED = "plugins.security.resource_sharing.enabled";
```
cwperks, line 13: I think we should include `experimental` in the setting.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/GetResourceTransportAction.java
```diff
@@ -0,0 +1,173 @@
...
+        if (client == null) {
+            fetchResourcesByIds(null, listener);
+            return;
+        }
+
+        client.getAccessibleResourceIds(RESOURCE_INDEX_NAME, ActionListener.wrap(ids -> {
+            if (ids.isEmpty()) {
+                listener.onResponse(new GetResourceResponse(Collections.emptySet()));
+            } else {
+                fetchResourcesByIds(ids, listener);
```
cwperks, line 98: For real cases in sharable resource plugins, I think we should have another way of signaling to the plugin that the current user has access to everything instead of returning all ids. For users that can only see a subset then we should return what they can see. (The alternative to that would be the DLS approach, but that has been ruled out to make this forward looking for the remote metadata sdk).

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1123,6 +1150,20 @@ public Collection<Object> createComponents(
             namedXContentRegistry.get()
         );
 
+        final var resourceSharingIndex = ResourceSharingConstants.OPENSEARCH_RESOURCE_SHARING_INDEX;
```
cwperks, line 1153: Can we also only instantiate these and return them from createComponents if the feature is enabled?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -2141,8 +2209,12 @@ public Collection<SystemIndexDescriptor> getSystemIndexDescriptors(Settings sett
             ConfigConstants.SECURITY_CONFIG_INDEX_NAME,
             ConfigConstants.OPENDISTRO_SECURITY_DEFAULT_CONFIG_INDEX
         );
-        final SystemIndexDescriptor systemIndexDescriptor = new SystemIndexDescriptor(indexPattern, "Security index");
-        return Collections.singletonList(systemIndexDescriptor);
+        final SystemIndexDescriptor securityIndexDescriptor = new SystemIndexDescriptor(indexPattern, "Security index");
```
cwperks, line 2212: Same here, let's add the descriptor if the feature flag is enabled. As long as everything in here is behind the feature flag then we have the right controls.

src/main/java/org/opensearch/security/resources/ResourcePluginInfo.java
DarshitChanpura: I added it to avoid doing computations twice. I've refactored the code to compute on the fly in onIndexModule()

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
DarshitChanpura: ah good point. done.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
DarshitChanpura: replaced.

src/main/java/org/opensearch/security/resources/ResourcePluginInfo.java
DarshitChanpura: it is used only in 1 place. I have removed this redundant clear.

spi/src/main/java/org/opensearch/security/spi/resources/FeatureConfigConstants.java
DarshitChanpura: done.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/GetResourceTransportAction.java
DarshitChanpura: There is no way of knowing that user has access to everything unless they are super-admins or are mapped to all-access. Even today, the permissions are computed from mapped roles and evaluated against requested actions.

In this case sample plugin, security plugin will return all IDs in current index that are accessible. This can be expanded in future to evaluate associated permissions before fetching individual ids, however, that make its pretty open and invites room for mis-configuration. IMO, only admin certificates should have access to all resources.
To your suggestion, yes we can introduce a method like `hasAllAccess` that plugin's can leverage , in later iteration.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: This is only used in 1 place, which is already controlled by feature-flag: https://github.com/DarshitChanpura/security/blob/adf96f1d377ea673fbe5a31644fa731ffff85623/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L2169

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: done.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
DarshitChanpura: Are you suggesting we change the design of the document?

This computation is fairly quick since it will only be evaluating for 1 resource at any time. We should not trade this to introduces new pre-computations.

OR are you suggesting we simply move this computation to ResourceSharing.java?

DarshitChanpura: I've moved them to ResourceSharing.java

cwperks (dismissed):
Thank you for the persistence @DarshitChanpura. Approving this PR with some of the comments remaining above, particularly around setting up extendedPlugins within the integrationTest framework.

I think this change is sufficiently isolated behind the experimental feature flag where the new classes initialized in OpenSearchSecurityPlugin are not used/instantiated unless the feature flag is enabled.

I really wanted to laud the introduction of this new extensibility model which I can see extended further in the future for other use-cases for plugins integrating with security and really paves the path forward for getting rid of the awkward existing plugin use-cases reading the user from the threadcontext and storing a copy in their own system indices either for this sharing use-case or for the job scheduler use case where they inject the roles back in at job runtime.

I also wanted to leave this quote: https://github.com/opensearch-project/OpenSearch/pull/4459#issuecomment-1275269995

cwperks (dismissed):

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -0,0 +1,1175 @@
...
+                listener.onFailure(
+                    new OpenSearchStatusException(
+                        "User " + requestUserName + " is not authorized to share resource " + resourceId,
+                        RestStatus.FORBIDDEN
+                    )
+                );
+            }
+
+            Script updateScript = new Script(ScriptType.INLINE, "painless", """
+                if (ctx._source.share_with == null) {
```
cwperks, line 544: The painless script here is hard the grok. Mentally, I think it makes sense to get the doc corresponding to the sharing info of the resource being acted upon, update it in memory and save it back.

I think one challenge is the centralized approach of single .resource-sharing index with sharing info of all sharable resources in the cluster.

One thing that could help is instead of having a single index, to have indexes matching 1-to-1 with the resource index.

So for example if we had a Quip Plugin with Sharable documents then there could be

.quip_documents

and the sharing info backed index managed by the security plugin

.quip_documents_sharing_info

For each index we could keep the same id for documents in both indices so that we can make GET requests instead of searches.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -0,0 +1,1175 @@
...
+     * @param isAdmin         Boolean indicating whether the user is an admin or not
+     * @param listener        Listener to be notified when the operation completes
+     * @throws IllegalArgumentException if resourceId, sourceIdx is null/empty, or if revokeAccess is null/empty
+     * @throws RuntimeException         if the update operation fails or encounters an error
+     * @apiNote This method modifies the existing document. If no modifications are needed (i.e., specified
+     * entities don't exist in the current configuration), the original document is returned unchanged.
+     * @see Recipient
+     * @see ResourceSharing
+     */
+    public void revokeAccess(
```
cwperks, line 645: I don't think it makes sense to have this in the java API. From a plugin dev point of view I want to know which resources the user has access to before issuing a query for them (akin to DLS, but shifting the responsibility to the plugin dev)

Meanwhile, security can expose 2 REST APIs;

- Get sharing info of the resource (can be queried by resource_type and resource_id)
- Update sharing info - again can be called with the resource_type and resource_id

When this feature is more mature there would additionally be re-usable components that call the security API behind the scenes if security is installed and enabled.

cwperks (approved):
Marking as approved again with additional comments.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: The painless script is intended to move the two calls (fetch and update) from security to the indexing eco-system. 

Individual indexes, while separating concerns, may make the management little cumbersome. And our current setup allows us to move into that direction if need be.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: Since no plugins actually call revoke atm, this method was added as a working example that is available for plugins to call.

In future, we can open this by replacing share and revoke from client to REST APIs, however that will move away of current default behaviour of sharing with backend-roles automatically and will be considered breaking change for new resources being created. Before introducing this a migration path will have to be in place to allow existing resources and their sharing to be migrated to central index.

derek-ho (commented):
Thank you, Darshit Chanpura

derek-ho (approved):
Thanks

PR #5307, opened 2025-04-29
Bug fix: Fixing stale cache issue post snapshot restore

nagarajg17:
### Description
Bug fix: Fixing stale cache issue post snapshot restore

Currently when a snapshot is restored on a domain including security index, the security index is updated but the in memory cache is not updated or reloaded with data from security index. Due to this AuthNZ for API requests and dashboard access is incorrect. It can either provide access when it shouldn't or deny access when it shouldn't

This change address this issue by adding a listener to index shard update events which reloads the cache with data from security index. This ensures cache is always updated when security index is updated

### Issues Resolved
[[Bug] Stale cache post snapshot restore](https://github.com/opensearch-project/security/issues/5308)

### Testing
- Tested with a local snapshot repository. Took a snapshot of current security index data. Made new changes to index by creating a new role with certain index permissions and mapping the role to existing user. Restored the snapshot and verified
    - API call to access index data which the new role had permissions was not accessible (before it was accessible because cache wasn't updated)
    - Verified cache and security index holds same data

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
+++ b/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
@@ -75,9 +75,13 @@ import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.common.util.concurrent.ThreadContext.StoredContext;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.Strings;
+import org.opensearch.core.index.Index;
+import org.opensearch.core.index.shard.ShardId;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.MediaTypeRegistry;
 import org.opensearch.env.Environment;
+import org.opensearch.index.shard.IndexEventListener;
+import org.opensearch.index.shard.IndexShard;
 import org.opensearch.security.auditlog.AuditLog;
 import org.opensearch.security.auditlog.config.AuditConfig;
 import org.opensearch.security.securityconf.DynamicConfigFactory;
@@ -93,8 +97,9 @@ import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.client.Client;
 
 import static org.opensearch.security.support.ConfigConstants.SECURITY_ALLOW_DEFAULT_INIT_USE_CLUSTER_STATE;
+import static org.opensearch.security.support.SnapshotRestoreHelper.isSecurityIndexRestoredFromSnapshot;
 
-public class ConfigurationRepository implements ClusterStateListener {
+public class ConfigurationRepository implements ClusterStateListener, IndexEventListener {
     private static final Logger LOGGER = LogManager.getLogger(ConfigurationRepository.class);
 
     private final String securityIndex;
@@ -666,6 +671,25 @@ public class ConfigurationRepository implements ClusterStateListener {
                     return null;
                 }
             });
+        }
+    }
+
+    @Override
+    public void afterIndexShardStarted(IndexShard indexShard) {
+        final ShardId shardId = indexShard.shardId();
+        final Index index = shardId.getIndex();
+
+        // Check if this is a security index shard
+        if (securityIndex.equals(index.getName())) {
+            // Only trigger on primary shard to avoid multiple reloads
+            if (indexShard.routingEntry() != null && indexShard.routingEntry().primary()) {
+                threadPool.generic().execute(() -> {
+                    if (isSecurityIndexRestoredFromSnapshot(clusterService, index, securityIndex)) {
+                        LOGGER.info("Security index primary shard {} started - config reloading for snapshot restore", shardId);
+                        reloadConfiguration(CType.values());
+                    }
+                });
+            }
         }
     }
 }
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -870,6 +870,8 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
                     }
                 }
             }.toListener());
+
+            indexModule.addIndexEventListener(cr);
         }
     }
 
--- a/src/main/java/org/opensearch/security/support/SnapshotRestoreHelper.java
+++ b/src/main/java/org/opensearch/security/support/SnapshotRestoreHelper.java
@@ -37,7 +37,11 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.SpecialPermission;
 import org.opensearch.action.admin.cluster.snapshots.restore.RestoreSnapshotRequest;
 import org.opensearch.action.support.PlainActionFuture;
+import org.opensearch.cluster.ClusterState;
+import org.opensearch.cluster.RestoreInProgress;
+import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.util.IndexUtils;
+import org.opensearch.core.index.Index;
 import org.opensearch.repositories.RepositoriesService;
 import org.opensearch.repositories.Repository;
 import org.opensearch.security.OpenSearchSecurityPlugin;
@@ -87,6 +91,24 @@ public class SnapshotRestoreHelper {
             setCurrentThreadName(threadName);
         }
         return snapshotInfo;
+    }
+
+    public static boolean isSecurityIndexRestoredFromSnapshot(ClusterService clusterService, Index index, String securityIndex) {
+        try {
+            ClusterState clusterState = clusterService.state();
+            RestoreInProgress restoreInProgress = clusterState.custom(RestoreInProgress.TYPE);
+
+            if (restoreInProgress != null) {
+                for (RestoreInProgress.Entry entry : restoreInProgress) {
+                    if (entry.indices().contains(securityIndex)) {
+                        return true;
+                    }
+                }
+            }
+        } catch (Exception e) {
+            log.warn("Could not determine if index {} was restored from snapshot, assuming new index", index.getName(), e);
+        }
+        return false;
     }
 
     @SuppressWarnings("removal")
--- a/src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
+++ b/src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
@@ -15,7 +15,9 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.time.Instant;
+import java.util.Collections;
 import java.util.Set;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeoutException;
 
 import com.fasterxml.jackson.databind.InjectableValues;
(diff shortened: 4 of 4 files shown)


shikharj05:
Thanks @nagarajg17 for the PR! can you take a look at failing tests?

cwperks:
The unit tests are failing because they perform the security index initialization using the code from [here](https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/test/AbstractSecurityUnitTest.java#L262-L288). To solve the issue, 1 solution may be to differentiate between when the security index is creating using a CreateIndexRequest vs when its created from a RestoreSnapshotRequest. I was able to write a quite [demo](https://github.com/cwperks/security/pull/50) to differentiate between these, but the code looks a bit strange because of needing to rewrite the name of the thread. I borrowed the idea from existing code in the security plugin [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/SnapshotRestoreHelper.java#L64-L90).

nibix:
I am not sure if the described behavior should be considered as a bug. Also all other operations which write to the security index do not automatically trigger a reload of the security configuration. In all cases, a subsequent call to https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/rest/SecurityConfigUpdateAction.java is performed to update the effective configuration.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -668,4 +674,39 @@ public Void run() {
...
+                        reloadConfiguration(CType.values());
+                    }
+                } finally {
+                    setCurrentThreadName(threadName);
+                }
+            }
+        }
+    }
+
+    private boolean isSecurityIndexRestoredFromSnapshot(Index index) {
```
cwperks, line 702: Can we move this into SnapshotRestoreHelper and keep `setCurrentThreadName` private?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -668,4 +674,39 @@ public Void run() {
...
+        try {
+            return Optional.ofNullable(clusterService.state().<RestoreInProgress>custom(RestoreInProgress.TYPE))
```
cwperks, line 704: I think this is fine, but I think it will always return `true` after the first instance of restoring from a snapshot unless the cluster state removes this from history at some point in time. What happens if 1) delete security index, 2) restore security index from a snapshot, 3) delete security index, 4) create empty security index from create index request

nagarajg17: It removes post successful restore process
https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/snapshots/RestoreService.java#L152-L154

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -668,4 +674,39 @@ public Void run() {
...
+        final Index index = shardId.getIndex();
+
+        // Check if this is a security index shard
+        if (securityIndex.equals(index.getName())) {
+            // Only trigger on primary shard to avoid multiple reloads
+            if (indexShard.routingEntry() != null && indexShard.routingEntry().primary()) {
+                final String threadName = Thread.currentThread().getName();
+
+                try {
+                    setCurrentThreadName("[" + ThreadPool.Names.GENERIC + "]");
```
nibix, line 690: This is a bit unconventional. Why is this necessary?

nagarajg17: Without this all the UTs started failing with error https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/service/ClusterApplierService.java#L443

Its due to difference in the way security index is created in test vs actual flow. @cwperks has explained [here](https://github.com/opensearch-project/security/pull/5307#issuecomment-2848102261). For this same reason in SnapshotHelper also they have done similar way https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/SnapshotRestoreHelper.java#L74-L87

nibix: What's the stack trace of the error?

Usually, these checks are there for a reason. OpenSearch uses quite a few tightly bounded thread pools for its async processing architecture.

The cluster state applier thread pool has a max size of 1:

https://github.com/opensearch-project/OpenSearch/blob/bf8cd257fd544e3bb33c062547be25bbe8b1045a/server/src/main/java/org/opensearch/cluster/service/ClusterApplierService.java#L178-L185

Any blocking operation on these threads bears the high risk of causing deadlocks, as the thread pool gets exhausted (as its only thread is blocked and cannot be returned into the pool). As soon as another operation triggered by this tries to modify the cluster state again, that operation will wait forever for the cluster state applier thread to become available again.

That's why the OpenSearch core code contains assertions like this.

Possible solutions might be:

- Refactor the whole code to be async 
- Offload blocking operations to a separate thread

cwperks: The error is thrown from [here](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/service/ClusterApplierService.java#L433-L448)

nagarajg17:
> BTW, I asked here a question to clarify whether we actually face a bug:

https://github.com/opensearch-project/security/pull/5307#issuecomment-2848112597

@nibix Reload of configurations is done [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java#L128) for any updates on index

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nagarajg17: > BTW, I asked here a question to clarify whether we actually face a bug:
https://github.com/opensearch-project/security/pull/5307#issuecomment-2848112597

@nibix Reload of configurations is done [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java#L128) for any updates on index

nagarajg17: Snapshot restore flow is different from flow of index update. What do you suggest?

nagarajg17: @nibix and @cwperks What approach would we want to take on fixing this error in tests `java.lang.AssertionError: should not be called by a cluster state applier. reason [the applied cluster state is not yet available]`. Could we align and close this discussion?

cwperks: > Could you elaborate what's the essential difference between snapshot restore and normal index update that configuration should be automatically reloaded on snapshot restore, but not on index update?

Restore from a snapshot is a more common operational action then direct index update, but you are right that the same applies for both cases.

nagarajg17: For index update `TransportConfigUpdateAction` is taking care of reloading configuration. Do we want to use this only for Snapshot restore? Since snapshot restore creates new index, `TransportConfigUpdateAction` flow won't be invoked for this

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
cwperks: Thank you for confirming.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nibix: > For index update TransportConfigUpdateAction is taking care of reloading configuration. 

I am sorry, but I do think this is not accurate :-)  TransportConfigUpdateAction is never automatically called if you consider basic index operations.

If you do an index update (via `PUT /.../_index/...` or via `PUT /_bulk` or via one of its transport counterparts), only the index will be updated. However, the security plugin  does *not* reload the security configuration. It only does if after the update an additional call to the `TransportConfigUpdateAction` is performed).

You can see that in the code of the `securityadmin` tool:

https://github.com/opensearch-project/security/blob/6e78dd9d1a1e5e05d50b626d796bd3011ac5c530/src/main/java/org/opensearch/security/tools/SecurityAdmin.java#L1362-L1392

In lines 1364 to 1379, the index is modified. Only afterwards, in line 1386 and 1387, there's a separate call to `PUT /_plugins/_security/configupdate`. 

Of course, this is a pattern that could be also applied to snapshot restore operations: Just do the snapshot restore and call `PUT /_plugins/_security/configupdate` afterwards.

My general opinion on this is:

- We should strive for consistent implementations. Consistent code will be easier to understand for developers and thus more efficient to maintain. Of course, inconsistencies are sometimes inevitable. But we should have good reasons to do so and name these (possibly even in the code as comment).
- Following the KISS principle, simple solutions should be preferred over complex or challenging solutions. At the moment, the proposed solution does complex (and possibly time consuming) operations in the very low level cluster state applier thread. This involves accessing the index itself, that's a task the cluster state applier thread was originally not created for (as it by default refuses these operations, compare the calls to `setCurrentThreadName()`). There might be ways to solve this, but research/tests are necessary to find these. This is also an indicator for complexity. On the other hand, there's the very simple solution, which is extremely easy to understand and very robust: Just call the additional `TransportConfigUpdateAction` after any manipulation of the index.  That's a solution that would be available already today, without any code change.

*Note: I won't veto against a solution with built-in automatic refresh of the configuration. However, I strongly believe that we should be able to give good reasons for introducing an inconsistency in the behavior between snapshot restore and index update. We should know how to react if afterwards a user files a bug report "If I do a `PUT /.opendistro_security/_doc/...`, security configuration is not automatically refreshed. However, it is automatically refreshed for snapshot restore. Why isnt it happening for the normal `PUT`?"*

*Note 2: I will be out of office for the next two days.*

cwperks: FWIW I have only ever operationally done a direct index update to the security index once to resolve an issue with a corrupt index. In that case, there was a separate bug which led to invalid values in the index.

Restore from a snapshot is a more common operational action that could be automated and I think its ok to focus on this scenario.

nagarajg17: Made changes for performing `ReloadConfigurations` in different thread

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -668,4 +673,23 @@ public Void run() {
...
+    public void afterIndexShardStarted(IndexShard indexShard) {
+        final ShardId shardId = indexShard.shardId();
+        final Index index = shardId.getIndex();
+
+        // Check if this is a security index shard
+        if (securityIndex.equals(index.getName())) {
+            // Only trigger on primary shard to avoid multiple reloads
+            if (indexShard.routingEntry() != null && indexShard.routingEntry().primary()) {
+                threadPool.generic().execute(() -> {
+                    if (isSecurityIndexRestoredFromSnapshot(clusterService, index, securityIndex)) {
```
shikharj05, line 687: nit: would this work if restore is already complete? for e.g. very small security index (few KBs)?

shikharj05 (commented):
Thanks @nagarajg17 ! one quick question on the approach.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nibix: @cwperks 

> FWIW I have only ever operationally done a direct index update to the security index once to resolve an issue with a corrupt index. 

Good point! Convinced :)

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nagarajg17: Yes, security index i tested had just 1 custom role and 2 users and rest default ones

cwperks (approved):

DarshitChanpura (approved):
Changes look good to me! It looks like we didn't have tests that checked validity of cache post restore before this change. And with this change those tests are no longer needed.

PR #5323, opened 2025-05-10
Allow for secure LDAP password settings

chriswhite199:
### Description
* Category: Enhancement
* Why these changes are required: To allow for secure storage of LDAP password settings
* What is the old behavior before changes and new behavior after changes
    * Before - settings for ldap `password` and `pem_password` were only configurable in an insecure manner (plaintext in yml or via cli / env variables)
    * After - Settings can now be acquired from the keystore (or still provided in an insecure manner, for which a warning will be shown in the logs, prompting the administrator to migrate to the secure versions)

If this PR is approved, i plan to submit similar PRs for the other secure settings using the `SecurableLegacySetting` class:
* OIDC: `pemkey_password, client_secret, passphrase` (https://docs.opensearch.org/docs/latest/security/authentication-backends/openid-connect/)
* JWT: `signing_key` (https://docs.opensearch.org/docs/latest/security/authentication-backends/jwt/)
* SAML: `signature_private_key_password, exchange_key, pemkey_password` (https://docs.opensearch.org/docs/latest/security/authentication-backends/saml/)

### Issues Resolved
* Related to (extension and refactor of) #2296 (for which I'm the original author)

Is this a backport: no

Do these changes introduce new permission(s) to be displayed in the static dropdown on the front-end: no

### Testing
* Unit tests updated for LDAP password (with insecure legacy setting and secure setting)
* There is currently no existing test for pemkey_password

### Check List
    - To be completed in another PR on opensearch-project/documentation-website repo once this PR is approved
- [ ] New Roles/Permissions have a corresponding security dashboards plugin PR - **N/A**
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md) - **N/A**

### DIFF
--- a/src/main/java/org/opensearch/security/auth/ldap2/LDAPConnectionFactoryFactory.java
+++ b/src/main/java/org/opensearch/security/auth/ldap2/LDAPConnectionFactoryFactory.java
@@ -184,7 +184,7 @@ public class LDAPConnectionFactoryFactory {
         BindConnectionInitializer result = new BindConnectionInitializer();
 
         String bindDn = settings.get(ConfigConstants.LDAP_BIND_DN, null);
-        String password = settings.get(ConfigConstants.LDAP_PASSWORD, null);
+        String password = ConfigConstants.LDAP_PASSWORD.getSetting(settings);
 
         if (password != null && password.length() == 0) {
             password = null;
--- a/src/main/java/org/opensearch/security/setting/SecurableLegacySetting.java
+++ b/src/main/java/org/opensearch/security/setting/SecurableLegacySetting.java
@@ -1,0 +1,99 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.setting;
+
+import java.util.Optional;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import org.opensearch.common.settings.SecureSetting;
+import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.core.common.settings.SecureString;
+
+/**
+ * Wrapper for legacy settings that support a secure variant located in the Keystore.
+ * <p>
+ * Secure name is the insecure name with "_secure" appended to it.
+ */
+public class SecurableLegacySetting {
+    private static final Logger LOG = LogManager.getLogger(SecurableLegacySetting.class);
+
+    public static final String SECURE_SUFFIX = "_secure";
+
+    public final String insecurePropertyName;
+
+    public final String propertyName;
+
+    public final String defaultValue;
+
+    public SecurableLegacySetting(String insecurePropertyName) {
+        this(insecurePropertyName, null);
+    }
+
+    public SecurableLegacySetting(String insecurePropertyName, String defaultValue) {
+        this(insecurePropertyName, String.format("%s%s", insecurePropertyName, SECURE_SUFFIX), defaultValue);
+    }
+
+    public SecurableLegacySetting(String insecurePropertyName, String propertyName, String defaultValue) {
+        super();
+        this.insecurePropertyName = insecurePropertyName;
+        this.propertyName = propertyName;
+        this.defaultValue = defaultValue;
+    }
+
+    public Setting<SecureString> asSetting() {
+        final Setting<SecureString> fallback = new InsecureFallbackStringSetting(this.insecurePropertyName, this.propertyName);
+        return SecureSetting.secureString(this.propertyName, fallback);
+    }
+
+    public Setting<SecureString> asInsecureSetting() {
+        return new InsecureFallbackStringSetting(this.insecurePropertyName, this.propertyName);
+    }
+
+    public String getSetting(Settings settings) {
+        return this.getSetting(settings, this.defaultValue);
+    }
+
+    public String getSetting(Settings settings, String defaultValue) {
+        return Optional.of(this.asSetting().get(settings)).filter(ss -> ss.length() > 0).map(SecureString::toString).orElse(defaultValue);
+    }
+
+    /**
+     * Alternative to InsecureStringSetting, which doesn't raise an exception if allow_insecure_settings is false, but
+     * instead log.WARNs the violation. This is to appease a potential cyclic dependency between commons-utils
+     */
+    private static class InsecureFallbackStringSetting extends Setting<SecureString> {
+        private final String name;
+        private final String secureName;
+
+        private InsecureFallbackStringSetting(String name, String secureName) {
+            super(name, "", s -> new SecureString(s.toCharArray()), Property.Deprecated, Property.Filtered, Property.NodeScope);
+            this.name = name;
+            this.secureName = secureName;
+        }
+
+        public SecureString get(Settings settings) {
+            if (this.exists(settings)) {
+                LOG.warn(
+                    "Setting [{}] has a secure counterpart [{}] which should be used instead. Allowing use of {} for legacy setups",
+                    this.name,
+                    this.secureName,
+                    this.name
+                );
+            }
+
+            return super.get(settings);
+        }
+    }
+}
--- a/src/main/java/org/opensearch/security/ssl/SecureSSLSettings.java
+++ b/src/main/java/org/opensearch/security/ssl/SecureSSLSettings.java
@@ -16,17 +16,11 @@ package org.opensearch.security.ssl;
 
 import java.util.Arrays;
 import java.util.List;
-import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-
-import org.opensearch.common.settings.SecureSetting;
 import org.opensearch.common.settings.Setting;
-import org.opensearch.common.settings.Settings;
-import org.opensearch.core.common.settings.SecureString;
+import org.opensearch.security.setting.SecurableLegacySetting;
 
 import static org.opensearch.security.ssl.util.SSLConfigConstants.DEFAULT_STORE_PASSWORD;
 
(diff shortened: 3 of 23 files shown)


cwperks:
Thank you for this PR @chriswhite199 ! Its been on my agenda to support the keystore here for some time now :)

src/test/java/org/opensearch/security/ssl/SecureSSLSettingsTest.java
```diff
@@ -7,46 +7,11 @@
...
 import org.junit.Test;
 
-import org.opensearch.common.settings.MockSecureSettings;
-import org.opensearch.common.settings.Settings;
-
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.is;
-import static org.opensearch.security.ssl.SecureSSLSettings.SSLSetting.SECURITY_SSL_HTTP_PEMKEY_PASSWORD;
-
 public class SecureSSLSettingsTest {
```
DarshitChanpura, line 10: Should this class be renamed. It seems to only have 1 test now. It doesn't make sense to have this test class file.

src/test/java/org/opensearch/security/setting/SecurableLegacySettingTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Test
+    public void testGetInsecureSetting() {
+        final var settings = Settings.builder().put(settingName, "test-password").build();
+        final var password = secureSetting.getSetting(settings);
+        assertThat(password, is("test-password"));
+    }
+
+    @Test
+    public void testShouldFavorSecureOverInsecureSetting() {
```
DarshitChanpura, line 50: 👏

src/main/java/org/opensearch/security/ssl/SecureSSLSettings.java
```diff
@@ -16,115 +16,77 @@
...
 
     public static List<Setting<?>> getSecureSettings() {
-        return Arrays.stream(SSLSetting.values())
+        return Arrays.stream(SecureSSLSettings.class.getDeclaredFields())
+            .filter(field -> SecurableLegacySetting.class.isAssignableFrom(field.getType()))
+            .map(field -> {
+                try {
+                    return (SecurableLegacySetting) field.get(null);
+                } catch (IllegalAccessException e) {
+                    throw new RuntimeException("Unable to access field: " + field.getName(), e);
```
DarshitChanpura, line 86: This should be [SettingsException](https://github.com/opensearch-project/OpenSearch/blob/78eea275a3e1812a008e815d846ca871d7f09d20/server/src/main/java/org/opensearch/common/settings/SettingsException.java#L48)

src/main/java/org/opensearch/security/setting/SecurableLegacySetting.java
```diff
@@ -0,0 +1,99 @@
...
+        return this.getSetting(settings, this.defaultValue);
+    }
+
+    public String getSetting(Settings settings, String defaultValue) {
+        return Optional.of(this.asSetting().get(settings)).filter(ss -> ss.length() > 0).map(SecureString::toString).orElse(defaultValue);
+    }
+
+    /**
+     * Alternative to InsecureStringSetting, which doesn't raise an exception if allow_insecure_settings is false, but
+     * instead log.WARNs the violation. This is to appease a potential cyclic dependency between commons-utils
```
DarshitChanpura, line 74: Interesting. For my curiosity, how did you discover this?

cwperks:
@chriswhite199 have you tested this outside of the unit tests? I think there may be larger structural changes required for settings in security yaml files to be able to fetch values from the keystore. For settings that reside in `opensearch.yml` this approach works, but I don't think that the security plugin currently has the ability to read from the keystore (unless I'm mistaken).

chriswhite199:
> @chriswhite199 have you tested this outside of the unit tests? I think there may be larger structural changes required for settings in security yaml files to be able to fetch values from the keystore. For settings that reside in `opensearch.yml` this approach works, but I don't think that the security plugin currently has the ability to read from the keystore (unless I'm mistaken).

I have not yet (for authn / authz plugins, I know it works for tls certs for https / transport as this was tested in the original PR for that scope). I didn't think it would be any different for authNZ plugins but I will test and get back to you. 

On a side not the `DEVELOPING_WITH_DOCKER.md` instructions didn't work for me under 3.0, something about a missing plugin.properties file (which could be related to the base image being `opensearchstaging/opensearch`, maybe that image doesn't have the security plugin unpacked correctly)

src/main/java/org/opensearch/security/setting/SecurableLegacySetting.java
chriswhite199: I can't remember now, but setting `allow_insecure_settings` and defining the property as secure would raise an exception - hence the original implementation crafted two properties - `password` and `password_secure` to not break existing installations.

If you look at the original review for https://github.com/opensearch-project/security/pull/2296, @peternied and I were discussing the integration test issues related to this - hopefully that comment can help you here

src/main/java/org/opensearch/security/auth/ldap2/LDAPConnectionFactoryFactory.java
```diff
@@ -184,7 +184,7 @@ private ConnectionInitializer getConnectionInitializer() {
         BindConnectionInitializer result = new BindConnectionInitializer();
 
         String bindDn = settings.get(ConfigConstants.LDAP_BIND_DN, null);
-        String password = settings.get(ConfigConstants.LDAP_PASSWORD, null);
+        String password = ConfigConstants.LDAP_PASSWORD.getSetting(settings);
```
nibix, line 187: What would be the complete name of the secure setting for LDAP passwords?

In the security plugin, you can configure more than one LDAP authentication domain. This can then also have different passwords. If I get this right, we won't be able to configure more than one LDAP domain with different passwords, correct?

src/test/java/org/opensearch/security/auth/ldap/LdapBackendTest.java
```diff
@@ -116,7 +117,27 @@ public void testLdapAuthenticationBindDn() throws Exception {
...
+    public void testLdapAuthenticationBindDnWithSecurePassword() throws Exception {
+        final var mockSecureSettings = new MockSecureSettings();
+        mockSecureSettings.setString(ConfigConstants.LDAP_PASSWORD.propertyName, "spocksecret");
+
+        final Settings settings = Settings.builder()
+            .putList(ConfigConstants.LDAP_HOSTS, "localhost:" + ldapPort)
+            .put(ConfigConstants.LDAP_AUTHC_USERSEARCH, "(uid={0})")
+            .put(ConfigConstants.LDAP_AUTHC_USERBASE, "ou=people,o=TEST")
+            .put(ConfigConstants.LDAP_BIND_DN, "cn=Captain Spock,ou=people,o=TEST")
+            .setSecureSettings(mockSecureSettings)
```
nibix, line 140: I think we need one integration test for secure settings. You can find integration tests for LDAP here:

https://github.com/opensearch-project/security/blob/main/src/integrationTest/java/org/opensearch/security/http/LdapAuthenticationTest.java

src/main/java/org/opensearch/security/setting/SecurableLegacySetting.java
```diff
@@ -0,0 +1,99 @@
...
+import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.core.common.settings.SecureString;
+
+/**
+ * Wrapper for legacy settings that support a secure variant located in the Keystore.
+ * <p>
+ * Secure name is the insecure name with "_secure" appended to it.
+ */
+public class SecurableLegacySetting {
```
nibix, line 29: I am a bit confused about the word "legacy" here. Could you please elaborate what "legacy settings" are?

src/main/java/org/opensearch/security/ssl/SecureSSLSettings.java
nibix, line 86: It took me a while to understand this until I figured out that you are using reflection to retrieve static members (as a replacement for enum.values()). 

As this is solution is maybe a bit non-obvious, I would recommend to add a comment somewhere to explain the approach.

Another question: With the SecurityManager, this would needed to be wrapped in a doPrivileged() block in order to be able to do reflection. With the new security manager replacement, did this need go away?

nibix (commented):
Thank you for this. Secure settings for the security plugin are certainly an important topic.

Still, it feels to me that we have kind of a concept mismatch between the normal *security settings* (stored in an index) and the *secure settings* (stored in the opensearch-keystore file on each node).

- It is unclear how the config names in *security settings* config.yml map to to the setting in the keystore, especially if we have lists of auth domains.
- The processes of config hot reloading (without starting the node) seem to unaligned for security settings and secure settings (I think it is possible, but it will involve quite a few steps). 

See below for more elaboration on that.

src/main/java/org/opensearch/security/auth/ldap2/LDAPConnectionFactoryFactory.java
chriswhite199: The AuthenticationBackend settings are created from a subset of the properties here:

* https://github.com/opensearch-project/security/blob/3.0/src/main/java/org/opensearch/security/securityconf/DynamicConfigModelV7.java#L267:L276

So it should allow for each individual backend to have it's own secure value. But examining the above only the settings from the index are used - there are no secure settings subset created and used when configuring the settings object for the backend.

Also a  misunderstanding on my part renders this whole PR void, as the settings are centralized in an index, not pulled from a yaml on each node. We could allow for pulling the secure settings from the keystore by amending the above to subset and inject secure settings when creating the Settings for each backend, but then you'd almost certainly run into issues where the admin would have to remember to update the keystore on every node when secure values are set, and when changed (also leading to confusion if those keystores are not in sync - where auth make work against one node, but fail against another).

I'm leaning towards withdrawing this PR in light of the above, unless anyone has any ideas on a way forward?

chriswhite199: Maybe this is the direction the PR should take (in relation to #4004):
* Amend the securityadmin script to allow for the passwords / secret settings to be sourced from the opensearch keystore (or config.yaml for legacy reasons)
* securityadmin still stores the secrets in the .opendistro-security index

My original intent of this PR is to avoid having passwords in cleartext config files on the filesystem (currently an admin has to remember to remove the config.yaml or redact the passwords after syncing).

PR #4571, opened 2024-07-16
[FEATURE] Endpoint to purge cache entries for specific users

prabhask5:
### Description

Finishing previous PR here: https://github.com/opensearch-project/security/pull/4314

Currently, our system only allows for the invalidation of the entire user authentication cache, which can lead to numerous cache misses and inefficiencies. This change is required to allow for more precise cache management, specifically targeting stale cache entries at the individual user level without disrupting the cache state of other users

Category: New feature
Why these changes are required?
Currently, our system only allows for the invalidation of the entire user authentication cache. This change is required to allow for more precise cache management, specifically targeting stale cache entries at the individual user level without disrupting the cache state of other users.
What is the old behavior before changes and new behavior after changes?
Previously, invalidating a user's cache required clearing the entire cache, affecting all users. The new behavior introduces an endpoint that allows for the invalidation of cache entries on a per-user basis, thereby maintaining cache integrity for other users and reducing unnecessary cache misses.

### Issues Resolved
https://github.com/opensearch-project/security/issues/2829

Is this a backport? If so, please add backport PR # and/or commits #
no

Do these changes introduce new permission(s) to be displayed in the static dropdown on the front-end? If so, please open a draft PR in the security dashboards plugin and link the draft PR here

### Testing
Unit testing was written in FlushCacheApiTest.java

### Check List
- [x] New functionality includes testing
- [ ] New functionality has been documented
- [ ] New Roles/Permissions have a corresponding security dashboards plugin PR
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md)
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/security/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #5325, opened 2025-05-12
[FEATURE] Endpoint to purge cache entries for specific users

Rishav9852Kumar:
### Description
Currently, our system only allows for the invalidation of the entire user authentication cache, which can lead to numerous cache misses and inefficiencies. This change is required to allow for more precise cache management, specifically targeting stale cache entries at the individual user level without disrupting the cache state of other users

Adding my changes on top of prev contributors [4571](https://github.com/opensearch-project/security/pull/4571) [changes in progress]

Category: New feature
Why these changes are required?
Currently, our system only allows for the invalidation of the entire user authentication cache. This change is required to allow for more precise cache management, specifically targeting stale cache entries at the individual user level without disrupting the cache state of other users.
What is the old behavior before changes and new behavior after changes?
Previously, invalidating a user's cache required clearing the entire cache, affecting all users. The new behavior introduces an endpoint that allows for the invalidation of cache entries on a per-user basis, thereby maintaining cache integrity for other users and reducing unnecessary cache misses.
### Issues Resolved

[2829](https://github.com/opensearch-project/security/issues/2829)

### Testing
Unit testing was written in FlushCacheApiTest.java

### Check List


shikharj05:
Thanks @Rishav9852Kumar! can you add some testing details?

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -171,6 +172,31 @@ public void invalidateCache() {
         restRoleCache.invalidateAll();
     }
 
+    public void invalidateUserCache(String[] usernames) {
+        if (usernames == null || usernames.length == 0) {
+            log.debug("No usernames given, not invalidating user cache.");
```
shikharj05, line 177: nit: this should be a ``warning`` message?

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -171,6 +172,31 @@ public void invalidateCache() {
...
+            return;
+        }
+
+        List<String> usernamesAsList = Arrays.asList(usernames);
```
shikharj05, line 181: nit: can we use a Set?

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
```diff
@@ -105,6 +111,8 @@ protected CType<?> getConfigType() {
 
     @Override
     protected void consumeParameters(final RestRequest request) {
-        // not needed
+        if (request.path().contains("/user/")) {
+            request.param("username");
```
shikharj05, line 115: nit: is the condition required? can this be something like- https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/dlic/rest/api/CertificatesApiAction.java#L71-L75

src/integrationTest/java/org/opensearch/security/http/LdapAuthenticationCacheTest.java
```diff
@@ -0,0 +1,187 @@
+/*
```
shikharj05, line 1: these look unrelated? should this be a separate PR?

src/main/java/org/opensearch/security/action/configupdate/ConfigUpdateRequest.java
```diff
@@ -51,10 +54,17 @@ public ConfigUpdateRequest(String[] configTypes) {
...
+        this();
+        setConfigTypes(new String[] { configType });
+        setEntityNames(entityNames);
+    }
+
     @Override
     public void writeTo(final StreamOutput out) throws IOException {
         super.writeTo(out);
         out.writeStringArray(configTypes);
+        out.writeOptionalStringArray(entityNames);
```
nibix, line 67: This will break in mixed clusters, i.e. in clusters having a version of OpenSearch with this change on some nodes and a version of OpenSearch on other nodes.

If such a change is necessary, the OpenSearch version of the remote node needs to be checked via the `StreamOutput` object. The format of the written information then has to be adapted according to the version of the remote node.

src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java
```diff
@@ -125,8 +126,14 @@ protected ConfigUpdateResponse newResponse(
...
     @Override
     protected ConfigUpdateNodeResponse nodeOperation(final NodeConfigUpdateRequest request) {
-        boolean didReload = configurationRepository.reloadConfiguration(CType.fromStringValues((request.request.getConfigTypes())));
-        if (didReload) {
+        if (request.request.getConfigTypes() != null
+            && request.request.getEntityNames() != null
+            && request.request.getConfigTypes().length == 1
+            && Arrays.asList(request.request.getConfigTypes()).contains(CType.INTERNALUSERS.toLCString())
+            && request.request.getEntityNames().length > 0) {
+            backendRegistry.get().invalidateUserCache(request.request.getEntityNames());
```
nagarajg17, line 134: This will not reload/refresh the cache in ConfigurationRepository

src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java
```diff
@@ -125,8 +126,14 @@ protected ConfigUpdateResponse newResponse(
 
     @Override
     protected ConfigUpdateNodeResponse nodeOperation(final NodeConfigUpdateRequest request) {
-        boolean didReload = configurationRepository.reloadConfiguration(CType.fromStringValues((request.request.getConfigTypes())));
-        if (didReload) {
+        if (request.request.getConfigTypes() != null
+            && request.request.getEntityNames() != null
+            && request.request.getConfigTypes().length == 1
+            && Arrays.asList(request.request.getConfigTypes()).contains(CType.INTERNALUSERS.toLCString())
```
nagarajg17, line 132: Why cache misses and inefficiencies problem only happens for InternalUsers entity update and not roles and roles mapping? We seem to invalidate all entries in all the caches in backendRegistry for other entity updates

src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java
nagarajg17, line 134: Cache in ConfigurationRepository isn't updated for InternalUsers entity update. Have we verified if behaviour same as before?

src/main/java/org/opensearch/security/auth/BackendRegistry.java
nagarajg17: +1 HashSet

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
```diff
@@ -65,37 +65,43 @@ public List<DeprecatedRoute> deprecatedRoutes() {
...
-
-                        @Override
-                        public void onFailure(final Exception e) {
-                            LOGGER.error("Cannot flush cache due to", e);
-                            internalServerError(channel, "Cannot flush cache due to " + e.getMessage() + ".");
-                        }
-
+        requestHandlersBuilder.allMethodsNotImplemented().override(Method.DELETE, (channel, request, client) -> {
+            final ConfigUpdateRequest configUpdateRequest;
+            if (request.path().contains("/user/")) {
```
nagarajg17, line 70: There seems to be patch API where more than 1 user can be updatec `new Route(Method.PATCH, "/internalusers")`. How will this work then?

DarshitChanpura: To follow convention, this should use legacy as well as current paths: `/user` and `/internalusers`. What are the expected paths here? Also, this should be documented.

@nagarajg17 PATCH api should not be on concern as it doesn't aim to remove user entries correct?

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
DarshitChanpura: it should not be required as the param is only consumed if it exists, directly via request.param("username")

src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java
nibix: Chiming in here:

I would be also interested in the actual problems which we want to solve here.

I have looked into the original issue #2829. This is a bit abstract, it does not give a real reason (i.e., an operational issue) we want to solve; it just gives the abstract goal to make cache invalidation more refined. 

However, it mentions explicitly the `restRoleCache` (from BackendRegistry). This cache is however *only* in effectively used if authorization backends are configured. The only authorization backend the security plugin defines is the LDAP authorization backend. 

The original need for the issue might come from that direction. Indeed, LDAP is the biggest performance concern in the auth/auth phase. But the issue does not really mention that or how it wants it to be solved. If this is really about purging users authenticated/authorized by LDAP, I guess the only need is to have a REST API  which allows you to selectively purge a user. Then, we do not need to change the behavior when reloading internal users.

src/integrationTest/java/org/opensearch/security/api/DefaultApiAvailabilityIntegrationTest.java
```diff
@@ -16,7 +16,6 @@
 import org.opensearch.test.framework.cluster.TestRestClient;
 
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.core.Is.is;
```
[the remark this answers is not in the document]
Rishav9852Kumar: imports were removed as part of moving all flushCache test to a new test file `FlushCacheApiIntegrationTest.java` which was not picked up by the git rebase, will discard this Pr to create new fresh PR + will mention  previous commits original contributors code ref in new PR desc finsted of picking their commits.

src/integrationTest/java/org/opensearch/security/http/LdapAuthenticationCacheTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+            response.assertStatusCode(200);
+
+            assertThat(response.getTextArrayFromJsonBody("/backend_roles"), contains("admin"));
+            assertThat(response.getTextArrayFromJsonBody("/backend_roles"), not(contains("enterprise")));
+        }
+
+        embeddedLDAPServer.loadLdifData(LDIF_DATA_UPDATED_BACKEND_ROLES);
+
+        try (TestRestClient client = cluster.getRestClient(ADMIN_USER)) {
```
cwperks, line 164: @nibix I think this test illustrates the issue well. Without these 3 lines the test fails because it would do a cache lookup for the user and have a stale view of the LDAP user. By having an endpoint to remove a specific user from the cache it will force LDAP to do another roundtrip to get the latest user info.

I thought this was a good issue to both solve an existing problem, while also touching a lot of areas of plugin development like writing a REST API, writing an integrationTest ans setting up authentication with an external provider.

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
DarshitChanpura: the if block can be removed

src/integrationTest/java/org/opensearch/security/http/LdapAuthenticationCacheTest.java
cwperks: These are directly related to the real problem that this PR addresses which is most relevant when using LDAP auth. See comment https://github.com/opensearch-project/security/pull/5325/files#r2086565354

src/main/java/org/opensearch/security/action/configupdate/ConfigUpdateRequest.java
cwperks: Since this issue is in particular solving an issue for LDAP Users, I think we should just clear the caches in BackendRegistry for the username for this PR. 

That being said, I think this raised an interesting point. If we add an attribute to an internal user does it make sense to re-read all `internal_users` from the security index or only the user that was updated?

IMO that is a separate optimization that would warrant a separate discussion.

nibix: > Since this issue is in particular solving an issue for LDAP Users, I think we should just clear the caches in BackendRegistry for the username for this PR.

With the background of https://github.com/opensearch-project/security/pull/5325#discussion_r2088012433 , I would also recommend to focus this PR on just introducing the extended delete cache REST API. 

Extensions of the config updates are a separate concern; things might be more easy if we work iteratively and do this in a later step.

> That being said, I think this raised an interesting point. If we add an attribute to an internal user does it make sense to re-read all internal_users from the security index or only the user that was updated?
> IMO that is a separate optimization that would warrant a separate discussion.

Agreed. (Just some background in advance: Actually, that won't be a quick change as the configuration index stores entire config types as documents; so, you actually cannot selectively read single entries of a config type from the index.)

Rishav9852Kumar:
https://github.com/opensearch-project/security/pull/5337

PR #5337, opened 2025-05-14
Selective User Cache Invalidation Enhancement

Rishav9852Kumar:
### Description
Currently, our system only allows for the invalidation of the entire user authentication cache, which can lead to numerous cache misses and inefficiencies. When dealing with LDAP user updates, this global cache invalidation affects all users unnecessarily.

## Solution
This PR introduces a new REST endpoint that enables selective cache invalidation at the individual user level. This allows for more precise cache management, particularly useful when handling LDAP user updates where only specific user entries become stale.

## Key Changes
- Introduces new REST endpoint `/api/security/authcache/users/{username}` for selective cache invalidation
- Adds integration tests to verify the behavior with LDAP authentication
- Implements proper authentication and security controls for the new endpoint

## Benefits
- Prevents unnecessary cache misses for unaffected users
- Improves system efficiency by maintaining cache integrity for other users
- Provides better handling of LDAP user updates without impacting the entire cache

## Developer Note
This issue was aimed as an excellent introduction to the plugin development workflow, touching all crucial aspects including REST API development, integration testing, unit testing, and authentication handling with external providers - making it an ideal "good first issue" for new contributors.

## Acknowledgments
This work builds upon the contributions of:
- @prabhask5 (PR #4571)
- @dlin2028 (PR #4314)
- @cwperks ( https://github.com/opensearch-project/security/pull/4571/commits/7cbd113d68213513fcb42b1210281a1e25d0a3fb)

## Related Work
- Building upon changes from #4571
- Extends functionality from #5325

### Issues Resolved
https://github.com/opensearch-project/security/issues/2829

#### IntegrationTest

[LdapAuthenticationCacheTest.java](https://github.com/opensearch-project/security/pull/5337/files#diff-380f96fc3a7fb060714375db09406f96ce2b810f481b3d03167a1fd1b15f213b)

[FlushCacheApiIntegrationTest.java](https://github.com/opensearch-project/security/pull/5337/files#diff-389437bde272a68d68240c1d2d3072fc647a5ece639e02963d504d1b4575232c)

#### Unit tesing 
[FlushCacheApiTest.java](https://github.com/opensearch-project/security/pull/5337/files#diff-97415602a617a109ffbd3ff7839053f779b77c7b21f8c6f2b2de675bde00e6ce)

### Check List
- [ X] Commits are signed per the DCO using --signoff


src/main/java/org/opensearch/security/action/configupdate/ConfigUpdateRequest.java
```diff
@@ -51,10 +53,17 @@ public ConfigUpdateRequest(String[] configTypes) {
...
+        this();
+        setConfigTypes(new String[] { configType });
+        setEntityNames(entityNames);
+    }
+
     @Override
     public void writeTo(final StreamOutput out) throws IOException {
         super.writeTo(out);
         out.writeStringArray(configTypes);
+        out.writeOptionalStringArray(entityNames);
```
nibix, line 66: This will fail on mixed clusters,  i.e. clusters that have nodes with this change and without this change. Such extensions are only possible by inspecting the version of the peer which is available from the StreamOutput class

src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java
```diff
@@ -125,8 +126,14 @@ protected ConfigUpdateResponse newResponse(
...
-        if (didReload) {
+        if (request.request.getConfigTypes() != null
+            && request.request.getEntityNames() != null
+            && request.request.getConfigTypes().length == 1
+            && Arrays.asList(request.request.getConfigTypes()).contains(CType.INTERNALUSERS.toLCString())
+            && request.request.getEntityNames().length > 0) {
+            backendRegistry.get().invalidateUserCache(request.request.getEntityNames());
+        } else {
+            configurationRepository.reloadConfiguration(CType.fromStringValues((request.request.getConfigTypes())));
             backendRegistry.get().invalidateCache();
```
cwperks, line 137: We should retain the `didReload` check here and only invalidate the cache if the configuration was reloaded.

src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java
```diff
@@ -125,8 +126,14 @@ protected ConfigUpdateResponse newResponse(
 
     @Override
     protected ConfigUpdateNodeResponse nodeOperation(final NodeConfigUpdateRequest request) {
-        boolean didReload = configurationRepository.reloadConfiguration(CType.fromStringValues((request.request.getConfigTypes())));
-        if (didReload) {
+        if (request.request.getConfigTypes() != null
```
cwperks, line 129: Do you think its possible to remove the static inner `NodeConfigUpdateRequest`? It looks to be a simple wrapper around `ConfigUpdateRequest` that doesn't do much.

Seeing `request.request.` multiple times looks odd.

src/main/java/org/opensearch/security/action/configupdate/ConfigUpdateRequest.java
cwperks: ^ yes, let's make sure this is serialized accordingly. I'll give a sample PR from core in a sec.

cwperks: Like this: https://github.com/opensearch-project/OpenSearch/pull/17312/files#diff-392c93c4b7e6a7e1d232c0dc0437a1b368e8c0be1698662586912c4b5477cf34R119-R138

src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java
```diff
@@ -125,8 +126,14 @@ protected ConfigUpdateResponse newResponse(
...
+            && request.request.getEntityNames() != null
+            && request.request.getConfigTypes().length == 1
+            && Arrays.asList(request.request.getConfigTypes()).contains(CType.INTERNALUSERS.toLCString())
+            && request.request.getEntityNames().length > 0) {
+            backendRegistry.get().invalidateUserCache(request.request.getEntityNames());
```
nagarajg17, line 134: There is another cache in ConfigurationRepository which can become stale

src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java
nagarajg17: +1 This can performance issues

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -185,6 +186,31 @@ public void invalidateCache() {
         restRoleCache.invalidateAll();
     }
 
+    public void invalidateUserCache(String[] usernames) {
+        if (usernames == null || usernames.length == 0) {
+            log.warn("No usernames given, not invalidating user cache.");
+            return;
+        }
+
+        List<String> usernamesAsList = Arrays.asList(usernames);
```
nagarajg17, line 195: Can you check previous PR comments as well? I guess there was a comment to use set for this

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
```diff
@@ -38,10 +38,15 @@ public class FlushCacheApiAction extends AbstractApiAction {
...
-    private static final List<Route> routes = addRoutesPrefix(ImmutableList.of(new Route(Method.DELETE, "/cache")));
+    private static final List<Route> routes = addRoutesPrefix(
+        ImmutableList.of(new Route(Method.DELETE, "/cache"), new Route(Method.DELETE, "/cache/user/{username}"))
+    );
 
     private static final List<DeprecatedRoute> deprecatedRoutes = addLegacyRoutesPrefix(
-        ImmutableList.of(new DeprecatedRoute(Method.DELETE, "/cache", OPENDISTRO_API_DEPRECATION_MESSAGE))
+        ImmutableList.of(
+            new DeprecatedRoute(Method.DELETE, "/cache", OPENDISTRO_API_DEPRECATION_MESSAGE),
+            new DeprecatedRoute(Method.DELETE, "/cache/user/{username}", OPENDISTRO_API_DEPRECATION_MESSAGE)
```
nagarajg17, line 48: Can you explain/clarify, I'm not clear on this
- Why this affects only for LDAP
- Backend registry caches are invalidated for all other entity updates as well. Why is it not a problem then?

src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java
Rishav9852Kumar: I think `NodeConfigUpdateRequest ` is just  adding a wrapper similar to other files , eg : https://github.com/opensearch-project/security/blob/c31bd3220c5caf2bd0070e392493a5769232047c/src/main/java/org/opensearch/security/dlic/rest/api/ssl/TransportCertificatesInfoNodesAction.java#L77


> Seeing request.request. multiple times looks odd.

i can follow how `request.request.` is removed using a var similar to https://github.com/opensearch-project/security/blob/c31bd3220c5caf2bd0070e392493a5769232047c/src/main/java/org/opensearch/security/dlic/rest/api/ssl/TransportCertificatesInfoNodesAction.java#L88,

src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java
Rishav9852Kumar: will add `didReload` check , as a condition for  full cache invalidation.

```

    @Override
    protected ConfigUpdateNodeResponse nodeOperation(final NodeConfigUpdateRequest request) {
        final var configupdateRequest = request.request;
        if (configupdateRequest.getConfigTypes() != null
            && configupdateRequest.getEntityNames() != null
            && configupdateRequest.getConfigTypes().length == 1
            && Arrays.asList(configupdateRequest.getConfigTypes()).contains(CType.INTERNALUSERS.toLCString())
            && configupdateRequest.getEntityNames().length > 0) {
            backendRegistry.get().invalidateUserCache(configupdateRequest.getEntityNames());
        } else {
            boolean didReload = configurationRepository.reloadConfiguration(CType.fromStringValues((configupdateRequest.getConfigTypes())));
            if (didReload) {
                backendRegistry.get().invalidateCache();
            }
        }
        return new ConfigUpdateNodeResponse(clusterService.localNode(), configupdateRequest.getConfigTypes(), null);
    }
```

src/main/java/org/opensearch/security/action/configupdate/ConfigUpdateRequest.java
Rishav9852Kumar: updated the optional string read and write inside a conditional varsion check for `V_3_0_0 `

```
        if (in.getVersion().onOrAfter(Version.V_3_0_0)) {
            this.entityNames = in.readOptionalStringArray();
        }
    
        if (out.getVersion().onOrAfter(Version.V_3_0_0)) {
            out.writeOptionalStringArray(entityNames);
        }
```

src/main/java/org/opensearch/security/action/configupdate/ConfigUpdateRequest.java
```diff
@@ -51,10 +56,19 @@ public ConfigUpdateRequest(String[] configTypes) {
...
+        this();
+        setConfigTypes(new String[] { configType });
+        setEntityNames(entityNames);
+    }
+
     @Override
     public void writeTo(final StreamOutput out) throws IOException {
         super.writeTo(out);
         out.writeStringArray(configTypes);
+        if (out.getVersion().onOrAfter(Version.V_3_0_0)) {
```
cwperks, line 69: This should be 3.1.0 since 3.0.0 is already released without this change.

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
```diff
@@ -65,37 +70,49 @@ public List<DeprecatedRoute> deprecatedRoutes() {
...
+            final String username = request.path().contains("/user/") ? request.param("username") : null;
+            LOGGER.debug("cache flush requested for {}", username);
+            if (request.path().contains("/user/")) {
+                // Extract the username from the request
+                if (username == null || username.isEmpty()) {
+                    internalServerError(channel, "No username provided for cache invalidation.");
+                    return;
+                }
+                // Validate and handle user-specific cache invalidation
+                configUpdateRequest = new ConfigUpdateRequest(CType.INTERNALUSERS.toLCString(), new String[] { username });
```
nibix, line 84: I am a bit confused here: We are specifying the config type `INTERNALUSERS`. However, the flush cash API is supposed to work for users from any auth type, incl. users authenticated by LDAP (LDAP is actually the core of the issue according to the description) or other means, not only `INTERNALUSERS`. 

Could you explain the reason why `INTERNALUSERS` is used here?

src/main/java/org/opensearch/security/action/configupdate/ConfigUpdateRequest.java
```diff
@@ -36,10 +37,14 @@
 public class ConfigUpdateRequest extends BaseNodesRequest<ConfigUpdateRequest> {
 
     private String[] configTypes;
+    private String[] entityNames;
 
     public ConfigUpdateRequest(StreamInput in) throws IOException {
         super(in);
         this.configTypes = in.readStringArray();
+        if (in.getVersion().onOrAfter(Version.V_3_0_0)) {
```
shikharj05, line 45: nit: this should be 3.1?

src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java
```diff
@@ -125,11 +126,20 @@ protected ConfigUpdateResponse newResponse(
...
     @Override
     protected ConfigUpdateNodeResponse nodeOperation(final NodeConfigUpdateRequest request) {
-        boolean didReload = configurationRepository.reloadConfiguration(CType.fromStringValues((request.request.getConfigTypes())));
-        if (didReload) {
-            backendRegistry.get().invalidateCache();
+        final var configupdateRequest = request.request;
+        if (configupdateRequest.getConfigTypes() != null
+            && configupdateRequest.getEntityNames() != null
+            && configupdateRequest.getConfigTypes().length == 1
+            && Arrays.asList(configupdateRequest.getConfigTypes()).contains(CType.INTERNALUSERS.toLCString())
```
shikharj05, line 133: nit: the allowed ctypes for selective validation can be a Set? i.e. it would easier to extend for other types in future like Roles

src/main/java/org/opensearch/security/action/configupdate/ConfigUpdateRequest.java
Rishav9852Kumar: ack

src/main/java/org/opensearch/security/auth/BackendRegistry.java
Rishav9852Kumar: modified the Array with Hashset as gives better performance for `.contains()`
```
Set<String> usernamesAsSet = new HashSet<>(Arrays.asList(usernames));

        // Invalidate entries in the userCache by iterating over the keys and matching the username.
        userCache.asMap()
            .keySet()
            .stream()
            .filter(authCreds -> usernamesAsSet.contains(authCreds.getUsername()))
            .forEach(userCache::invalidate);
```

src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java
Rishav9852Kumar: updated code with 

```
    private static final Set<CType<?>> SELECTIVE_VALIDATION_TYPES = 
    Set.of(CType.INTERNALUSERS);
```
to extend for other types in future

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
Rishav9852Kumar: The use of INTERNALUSERS here is serving as a marker/identifier for user-specific cache invalidation operations, rather than limiting the cache invalidation to only internal users. 

When the ConfigUpdateRequest is created with INTERNALUSERS type and a specific username, the TransportConfigUpdateAction's nodeOperation method checks if it's a selective validation request (via canHandleSelectively method) and then calls backendRegistry.invalidateUserCache() which invalidates the cache for the specified user regardless of their authentication type (internal, LDAP, etc.).

nibix: Using the type INTERNALUSERS for other user types like LDAP sounds misleading and could be confusing for future developers. 

In an API, if I specify INTERNALUSERS, I would expect only INTERNALUSERS to be affected. 

To be honest, it feels like this API is not really suitable for the task. The main purpose of the API is reloading the internal configuration of the security plugin. However, LDAP users do not belong to the *internal* config of the security plugin. It is only a secondary effect of this API - the clearing of the user cache - that shall be leveraged here. 

Why don't we just create a new API dedicated to clearing the user cache? This would be much clearer and easier.

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
Rishav9852Kumar: The issue isn't exclusive to LDAP, but it's more noticeable with LDAP users because of how frequently their attributes might change

[LDIF_DATA_UPDATED_BACKEND_ROLES](https://github.com/opensearch-project/security/blob/e232066790b0ed264765c77840bf492b8ca6b2d6/src/integrationTest/java/org/opensearch/security/http/DirectoryInformationTrees.java#L125)
[loadLdifData](https://github.com/opensearch-project/security/blob/e232066790b0ed264765c77840bf492b8ca6b2d6/src/integrationTest/java/org/opensearch/security/http/LdapAuthenticationCacheTest.java#L162)

nibix: IMHO, the situation is slightly different;

LDAP is the *only* auth type which requires a *network roundtrip* to an external system to verify the credentials of a user. 

That's why LDAP authentication requests will be always one order of magnitude slower. Additionally, LDAP/active directory servers often cannot scale as good as OpenSearch. Especially in high load situations on OpenSearch, sending a auth request to LDAP for each request to OpenSearch would just DDOS the LDAP server.

Thus, the main reason the cache exists is LDAP.

If LDAP is not in use, one could argue that the cache could be disabled alltogether. For authentication with a noop backend (thats primarily JWT, SAML, OIDC, client cert and possibly proxy), the cache is actually already now disabled by special logic in the BackendRegistry code: 

https://github.com/opensearch-project/security/blob/344673a455de956f6a8f3217e61d0636b46a3527/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L599-L604

The cache might still help for the internal user backend, because the auth process is possibly not as efficient as it could be; but if the code would be optimized there, the cache might be also unnecessary for the internal user backend case.

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
nibix: > If that's the case, wdyt about creating a separate internal transport action for this use-case but re-use this Rest action?

Yes, that's what I meant. At the moment, we have two REST endpoints which use the same internal transport action. The main goal of the transport action is however only updating configuration; that is reflected by the name  `TransportConfigUpdateAction`.   The `DELETE /cache` API actually just uses a side effect of that action, which is clearing the cache.

The new code path actually does not perform the primary function any more; no configuration reloading takes place any more (compare the block from line 133 to 134):

https://github.com/opensearch-project/security/blob/e232066790b0ed264765c77840bf492b8ca6b2d6/src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java#L133-L140

src/main/java/org/opensearch/security/dlic/rest/api/FlushCacheApiAction.java
shikharj05: nit: we should ideally avoid adding new APIs to deprecated route. 
```
new DeprecatedRoute(Method.DELETE, "/cache/user/{username}", OPENDISTRO_API_DEPRECATION_MESSAGE)
```

Rishav9852Kumar: Overrided method to only run test with the new api 

```
@Override
    protected String apiPathPrefix() {
    protected String apiPathPrefix() {
        return PLUGINS_PREFIX;
    }
```

PR #5370, opened 2025-06-01
Easing debugging for security not initialized error

nagarajg17:
### Description

If security index is not created during node bootup and if any API call is made we receive generic error `OpenSearch Security is not initialized`. Security index may not be created for various number of reasons but commonly seen due to master not initialized. To ease the debugging and fixing the issue, adding more details to error statement and possible cause will help. This PR starts with one of the possible cause, it can be extended in future appropriately

Future adding few debug lines to ease debugging during issues

Maintenance

### Issues Resolved
Easing debugging and fixing error 

### Testing
To minic the scenario added this setting in opensearch.yml `node.roles: [data, ingest]` to have no master node. Then made API calls to reproduce the error

```
curl -X GET "https://localhost:9200/my-index-000001/_search?pretty"  -H 'Content-Type: application/json' -u '<>:<>' --insecure

OpenSearch Security is not initialized. Cluster manager not present
```

**Earlier**
```
curl -X GET "https://localhost:9200/my-index-000001/_search?pretty"  -H 'Content-Type: application/json' -u '<>:<>' --insecure

OpenSearch Security is not initialized
```

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/configuration/ClusterInfoHolder.java
+++ b/src/main/java/org/opensearch/security/configuration/ClusterInfoHolder.java
@@ -36,6 +36,7 @@ import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.cluster.node.DiscoveryNodes;
 
 public class ClusterInfoHolder implements ClusterStateListener {
+    public static final String CLUSTER_MANAGER_NOT_PRESENT = "Cluster manager not present";
 
     protected final Logger log = LogManager.getLogger(this.getClass());
     private volatile DiscoveryNodes nodes = null;
@@ -92,5 +93,12 @@ public class ClusterInfoHolder implements ClusterStateListener {
 
     public String getClusterName() {
         return this.clusterName;
+    }
+
+    public Boolean hasClusterManager() {
+        if (nodes != null) {
+            return nodes.getClusterManagerNode() != null;
+        }
+        return false;
     }
 }
--- a/src/test/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluatorTest.java
+++ b/src/test/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluatorTest.java
@@ -115,7 +115,6 @@ public class RestLayerPrivilegesEvaluatorTest {
         assertThat(exception.getMessage(), equalTo("OpenSearch Security is not initialized."));
 
         when(clusterInfoHolder.hasClusterManager()).thenReturn(false);
-        when(ClusterInfoHolder.CLUSTER_MANAGER_NOT_PRESENT).thenReturn("Cluster manager not present");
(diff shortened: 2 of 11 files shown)


src/test/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluatorTest.java
```diff
@@ -104,11 +107,20 @@ public void testEvaluate_Initialized_Success() throws Exception {
...
         );
         assertThat(exception.getMessage(), equalTo("OpenSearch Security is not initialized."));
+
+        when(clusterInfoHolder.hasClusterManager()).thenReturn(false);
+        when(clusterInfoHolder.getClusterManagerNotPresentStatus()).thenReturn("Cluster manager not present");
+        exception = assertThrows(
+                OpenSearchSecurityException.class,
+                () -> restPrivilegesEvaluator.evaluate(TEST_USER, "route_name", null)
+        );
+        assertThat(exception.getMessage(), equalTo("OpenSearch Security is not initialized.Cluster manager not present"));
```
cwperks: nit: Let's add a space before the second sentence here.

src/main/java/org/opensearch/security/configuration/ClusterInfoHolder.java
```diff
@@ -93,4 +94,15 @@ public Boolean hasNode(DiscoveryNode node) {
...
+
+    public Boolean hasClusterManager() {
+        if (nodes != null) {
+            return nodes.getClusterManagerNode() != null;
+        }
+        return false;
+    }
+
+    public String getClusterManagerNotPresentStatus() {
+        return CLUSTER_MANAGER_NOT_PRESENT;
```
cwperks: Why have this as a function? wdyt about making the constant `public static final String` and then reference the constant where needed? We don't need to pass `ClusterInfoHolder` instance around to reference this constant.

nagarajg17: Don't have any strong opinion on making the constant public scope, either is fine but we would still need `ClusterInfoHolder` instance for `hasClusterManager()` atleast.

cwperks: > we would still need ClusterInfoHolder instance for hasClusterManager() atleast.

For that definitely, but to refer to the constant we shouldn't need an instance of this class.

cwperks (approved):

DarshitChanpura (approved):

DarshitChanpura:
> @nagarajg17 can we also add a CHANGELOG entry for this? I think its a significant enough change to have an entry in the CHANGELOG.

+1

DarshitChanpura:
@nagarajg17 Would you mind checking the test failures?

cwperks (approved):

nagarajg17:
I see PR is merged. I didn't get a chance to check test failures yesterday, will check it and create another PR for CHANGELOG

PR #5374, opened 2025-06-03
Extracted ActionPrivileges interface and created RoleBasedActionPrivileges implementation and SubjectBasedActionPrivileges implementation

nibix:
### Description

This is a refactoring of the `ActionPrivileges` class: Its interface is decoupled from the implementation. `ActionPrivileges` is not the interface, `RoleBasedActionPrivileges` contains the actual implementation. 

Additionally, a new class `SubjectBasedActionPrivileges` was introduced which provides an alternative mode of privilege evaluation;  `SubjectBasedActionPrivileges` does not consider mapped roles. A single instance of  `SubjectBasedActionPrivileges` is meant for a single subject. This is useful for providing privileges for plugins (see #5341 ) and will be also useful for API tokens.

* Category:  Refactoring
* Why these changes are required?: For  #5341, we needed a new way to define privileges for plugins. As plugins do not have roles, a decoupling of the `RoleBasedActionPrivileges` seemed sensible. This principle can be also used for API tokens.
* What is the old behavior before changes and new behavior after changes?: No behavioral changes

### Testing

- New unit test `SubjectBasedActionPrivilegesTest`

### Check List


src/integrationTest/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivilegesTest.java
```diff
@@ -380,23 +387,6 @@ public void negative_wrongAction() throws Exception {
                 }
             }
 
-            @Test
-            public void positive_hasExplicit_full() {
```
nibix, line 384: FYI: This was moved to the "Misc" section

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -153,11 +155,18 @@ public class PrivilegesEvaluator {
     private final TermsAggregationEvaluator termsAggregationEvaluator;
     private final PitPrivilegesEvaluator pitPrivilegesEvaluator;
     private DynamicConfigModel dcm;
-    private final NamedXContentRegistry namedXContentRegistry;
     private final Settings settings;
-    private final Map<String, Set<String>> pluginToClusterActions;
-    private final AtomicReference<ActionPrivileges> actionPrivileges = new AtomicReference<>();
+    private final AtomicReference<RoleBasedActionPrivileges> actionPrivileges = new AtomicReference<>();
     private final AtomicReference<TenantPrivileges> tenantPrivileges = new AtomicReference<>();
+    private final Map<String, SubjectBasedActionPrivileges> pluginIdToActionPrivileges = new HashMap<>();
```
cwperks, line 161: Should we consider renaming this to `subjectToActionPrivileges`? It would be consistent with the naming `SubjectBasedActionPrivileges`.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -311,9 +310,32 @@ public PrivilegesEvaluationContext createContext(
...
         TransportAddress caller = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
-        ImmutableSet<String> mappedRoles = ImmutableSet.copyOf((injectedRoles == null) ? mapRoles(user, caller) : injectedRoles);
 
-        return new PrivilegesEvaluationContext(user, mappedRoles, action0, request, task, irr, resolver, clusterStateSupplier);
+        ActionPrivileges actionPrivileges;
+        ImmutableSet<String> mappedRoles;
+
+        if (user.isPluginUser()) {
+            mappedRoles = ImmutableSet.of();
+            actionPrivileges = this.pluginIdToActionPrivileges.get(user.getName());
```
cwperks, line 323: `this.pluginIdToActionPrivileges.getOrDefault(user.getName(), ActionPrivileges.EMPTY);`

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
nibix: We can rename it. I have chosen the present name because it illustrate that only subjects from plugins will be stored here. If at one point we use this for API tokens, we will likely need to use an additional data structure which is prepared for handling the dynamic adding/removing of API tokens (i.e., concurrency control, bounded size, ...)

But I don't have a strong opinion on the name.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
nibix: I really liked that suggestion ... when trying it out, I noticed that it does not work because the map is of type `Map<String, SubjectBasedActionPrivileges>` ... so the second parameter of `getOrDefault()` must be also of type `SubjectBasedActionPrivileges`.

I could solve it by chaning the type of `pluginIdToActionPrivileges` or by weird casting expressions, but both seem to have some bigger downsides compared to the additional if block.

nibix: An additional thought: In a second step, one could think about also migrating the privileges for the plugin's system index into this. Thus, one could concentrate all privilege handling into one artifact, removing it from `SystemIndexAccessEvaluator`.

cwperks (approved):

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
```diff
@@ -0,0 +1,889 @@
...
+     * This setting controls the allowed heap size of the precomputed index privileges (in the inner class StatefulIndexPrivileges).
+     * If the size of the indices exceed the amount of bytes configured here, it will be truncated. Privileges evaluation will
+     * continue to work correctly, but it will be slower.
+     * <p>
+     * This settings defaults to 10 MB. This is a generous limit. Experiments have shown that an example setup with
+     * 10,000 indices and 1,000 roles requires about 1 MB of heap. 100,000 indices and 100 roles require about 9 MB of heap.
+     * (Of course, these numbers can vary widely based on the actual role configuration).
+     */
+    public static Setting<ByteSizeValue> PRECOMPUTED_PRIVILEGES_MAX_HEAP_SIZE = Setting.memorySizeSetting(
+        "plugins.security.privileges_evaluation.precomputed_privileges.max_heap_size",
```
DarshitChanpura, line 74: nit: this should be stored in ConfigConstants?

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
```diff
@@ -0,0 +1,889 @@
...
+    public RoleBasedActionPrivileges(SecurityDynamicConfiguration<RoleV7> roles, FlattenedActionGroups actionGroups, Settings settings) {
+        super(new ClusterPrivileges(roles, actionGroups), new IndexPrivileges(roles, actionGroups));
+        this.roles = roles;
+        this.actionGroups = actionGroups;
+        this.statefulIndexMaxHeapSize = PRECOMPUTED_PRIVILEGES_MAX_HEAP_SIZE.get(settings);
+    }
+
+    /**
+     * Updates the stateful index configuration with the given indices. Should be normally only called by
+     * updateStatefulIndexPrivilegesAsync(). Package visible for testing.
```
DarshitChanpura: We should mark this as @VisibleForTesting for documentation purposes?

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
```diff
@@ -0,0 +1,889 @@
...
+    public void updateStatefulIndexPrivileges(Map<String, IndexAbstraction> indices, long metadataVersion) {
+        StatefulIndexPrivileges statefulIndex = this.statefulIndex.get();
+
+        indices = StatefulIndexPrivileges.relevantOnly(indices);
+
+        if (statefulIndex == null || !statefulIndex.indices.equals(indices)) {
+            long start = System.currentTimeMillis();
+            this.statefulIndex.set(new StatefulIndexPrivileges(roles, actionGroups, indices, metadataVersion, statefulIndexMaxHeapSize));
+            long duration = System.currentTimeMillis() - start;
+            log.debug("Updating StatefulIndexPrivileges took {} ms", duration);
```
DarshitChanpura, line 112: what is the metric used for?

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
```diff
@@ -0,0 +1,889 @@
...
+    }
+
+    int getEstimatedStatefulIndexByteSize() {
+        StatefulIndexPrivileges statefulIndex = this.statefulIndex.get();
+
+        if (statefulIndex != null) {
+            return statefulIndex.estimatedByteSize;
+        } else {
+            return 0;
+        }
```
DarshitChanpura, line 131:
```suggestion
        return statefulIndex == null ? 0 : statefulIndex.estimatedByteSize;
```

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
```diff
@@ -0,0 +1,889 @@
...
+                        // Otherwise, we match the pattern against the provided well-known cluster actions and add
+                        // these to the "actionToRoles" map. Additionally, for the case that the well-known cluster
+                        // actions are not complete, we also collect the matcher to be used as a last resort later.
+
+                        if (WildcardMatcher.isExact(permission)) {
+                            actionToRoles.computeIfAbsent(permission, k -> roleSetBuilder.createSubSetBuilder()).add(roleName);
+                        } else if (permission.equals("*")) {
+                            // Special case: Roles with a wildcard "*" giving privileges for all actions. We will not resolve
+                            // this stuff, but just note separately that this role just gets all the cluster privileges.
+                            rolesWithWildcardPermissions.add(roleName);
```
DarshitChanpura, line 220: this will be only for strings that '*', not for strings that contains * with other chars. Those will be evaluated in the else block below, correct?

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
```diff
@@ -0,0 +1,889 @@
...
+                    for (RoleV7.Index indexPermissions : role.getIndex_permissions()) {
+                        ImmutableSet<String> permissions = actionGroups.resolve(indexPermissions.getAllowed_actions());
+
+                        for (String permission : permissions) {
+                            // If we have a permission which does not use any pattern, we just simply add it to the
+                            // "rolesToActionToIndexPattern" map.
+                            // Otherwise, we match the pattern against the provided well-known index actions and add
+                            // these to the "rolesToActionToIndexPattern" map. Additionally, for the case that the
+                            // well-known index actions are not complete, we also collect the actionMatcher to be used
+                            // as a last resort later.
```
DarshitChanpura, line 360: qq, why do we not treat "*" as a special case here just like in the cluster privileges above?

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
```diff
@@ -0,0 +1,889 @@
...
+                    }
+                }
+            }
+
+            if (!doFilter) {
+                return indices;
+            }
+
+            // Second pass: Only if we actually need filtering, we will do it
+            ImmutableMap.Builder<String, IndexAbstraction> builder = ImmutableMap.builder();
```
DarshitChanpura, line 864: why not do it in the first pass. It seems like first pass will only break if doFilter will be set to true, and if doFilter is set to true then we need second pass any way. Otherwise first pass will run through all indices. So if we combine both passes we will only visit each index once.

src/main/java/org/opensearch/security/privileges/actionlevel/SubjectBasedActionPrivileges.java
```diff
@@ -0,0 +1,402 @@
...
+import org.opensearch.security.support.WildcardMatcher;
+
+import com.selectivem.collections.CheckTable;
+
+import static org.opensearch.security.privileges.actionlevel.WellKnownActions.allWellKnownIndexActions;
+
+/**
+ * An ActionPrivileges implementation that is valid only for a single entity.
+ * This means that individual instances of this class must be created for individual entities. The mapped roles
+ * from the context are not regarded by this method.
```
DarshitChanpura: nit: by this class.

src/main/java/org/opensearch/security/privileges/actionlevel/SubjectBasedActionPrivileges.java
```diff
@@ -0,0 +1,402 @@
...
+ * <p>
+ * The method PrivilegesEvaluator.createContext() is responsible for making sure that the correct class is used.
+ * <p>
+ * This class is useful for plugin users and API tokens.
+ */
+public class SubjectBasedActionPrivileges extends RuntimeOptimizedActionPrivileges implements ActionPrivileges {
```
DarshitChanpura: we can skip explicit declaration of  `implements ActionPrivileges` here.

src/main/java/org/opensearch/security/privileges/actionlevel/SubjectBasedActionPrivileges.java
```diff
@@ -0,0 +1,402 @@
...
+     */
+    public SubjectBasedActionPrivileges(RoleV7 role, FlattenedActionGroups actionGroups) {
+        super(new ClusterPrivileges(actionGroups.resolve(role.getCluster_permissions())), new IndexPrivileges(role, actionGroups));
+    }
+
+    /**
+     * At the moment, this class does not provide StatefulIndexPrivileges.
+     * Thus, always the slightly slower index matching code path will be used. For plugins, however,
+     * that should be okay, as they likely request specific indices without patterns.
+     */
```
DarshitChanpura, line 70: It should be easier to compute stateful privileges for plugins and API tokens right? Especially for plugins since each plugin will have a limited subset of explicitly specified index names.

src/main/java/org/opensearch/security/privileges/actionlevel/SubjectBasedActionPrivileges.java
```diff
@@ -0,0 +1,402 @@
...
+         * This does not include privileges obtained via "*" action patterns. This is both meant as a
+         * optimization and to support explicit privileges.
+         */
+        private final ImmutableSet<String> grantedActions;
+
+        /**
+         * This contains all role names that provide wildcard (*) privileges for cluster actions.
+         * This avoids a blow-up of the actionToRoles object by such roles.
+         */
+        private final boolean providesWildcardPrivilege;
```
DarshitChanpura, line 100: the comment above suggest that this is a collection rather than a boolean. Can you update the comment?

DarshitChanpura: also why not use collection of roles here? Is it because we don't consider roles at all? How will this work for API tokens?

src/integrationTest/java/org/opensearch/security/privileges/actionlevel/SubjectBasedActionPrivilegesTest.java
```diff
@@ -0,0 +1,717 @@
...
+                result.getEvaluationExceptionInfo().startsWith("Exceptions encountered during privilege evaluation:")
+            );
+        }
+
+    }
+
+    static RoleV7 config(String config) {
+        return RoleV7.fromYamlStringUnchecked(config);
+    }
+}
```
DarshitChanpura, line 717: we should test this class with API tokens before that feature goes GA.

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
nibix: Maybe I am weird, but I do believe that `ConfigConstants` is an anti pattern :) It is 300 lines of diverse strings, some are config option, but some are also something completely different. Even worse, config option metadata like type, scope and partially default values are again *not* defined in `ConfigConstants` - even though they are both config related and constant. Instead, you'll have to look into `OpenSearchSecurityPlugin` for this.

You'll never look at `ConfigConstants` with the goal of "let's see what configuration options we have". It is just too unstructured.

Furthermore, there is no need to reuse this constant ... it is only used in `RoleBasedActionPrivileges`. Additionally, it needs to be referenced in `OpenSearchSecurityPlugin` to be registered. But that's it. 

That's why I believe in that such constants should be always placed as closely as possible to the place where they are used. I can look at the class `RoleBasedActionPrivileges` and I will see after less than one minute the config options that I can use configure instances of this.

nibix: I took this opportunity to write an RFC about this: https://github.com/opensearch-project/security/issues/5378

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
nibix: It's on debug level, so it is only used in debugging situations. Rebuilding the object can take - in code terms - a slightly longer time; on big clusters, it might be a second or so. If there is the suspicion that the rebuilding process requires too many resources, this gives you the easy oppurnity to check this without greater effort.

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
nibix: Hmm ...  IMHO, both versions are reasonable and acceptable code. What's the motivation to change it?

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
nibix: Exactly, the wildcard is the `*` string that matches just everything.

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
nibix: We do:

https://github.com/opensearch-project/security/blob/b4b8f3c1caac63e610931728a1b4776ae8b91572/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java#L383-L387

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
nibix: We avoid creating copies if unnecessary. These cluster metadata objects can be potentially quite big.

src/main/java/org/opensearch/security/privileges/actionlevel/SubjectBasedActionPrivileges.java
nibix: We have to discuss/research whether we want to do this. 

You have already stated that plugins usually give explicit index names, without patterns. In such a case, the stateful privilege data structure does not provide any benfit anyway, as the static object will already contain basically the same data structure.

For API tokens: For the case that you should have *many* API tokens (like hundrets), one has to check whether it is economical to update API tokens for each cluster change. My guess would that it will get unfeasible at one point. We should revisit this when it comes to taking care of API tokens.

src/main/java/org/opensearch/security/privileges/actionlevel/SubjectBasedActionPrivileges.java
nibix: Yes, the `SubjectBasedActionPrivileges` has no notion of roles. One instance corresponds to exactly subject.

For API tokens, the idea would be that one instance corresponds to exactly one token.

src/integrationTest/java/org/opensearch/security/privileges/actionlevel/SubjectBasedActionPrivilegesTest.java
nibix: With "that feature", do you mean API tokens? Or merging this PR?

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
nibix: The comment was actually not true. It was used for non testing purposes. I changed the comment.

src/main/java/org/opensearch/security/privileges/actionlevel/SubjectBasedActionPrivileges.java
nibix: >  Can you update the comment?

Done. Actually, I forgot to update quite a few comments. This is fixed now.

cwperks (dismissed):

willyborankin (dismissed):

src/integrationTest/java/org/opensearch/security/privileges/actionlevel/SubjectBasedActionPrivilegesTest.java
DarshitChanpura: I meant API tokens.

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
DarshitChanpura: I mean wildcard by itself. Instead of .contains("\*"), match with .equals("\*")

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
DarshitChanpura: ~Are you worried about creating objects for the entries only to not use them later as doFilter is set to false. Won't GC take care of that?~

I'm fine with current approach as it seems like GC might not be much help if the builder object is too large and we already paid memory and CPU overhead.

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
DarshitChanpura: clean and compact :)

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
DarshitChanpura: I see. How often do we expect this to be logged?

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
DarshitChanpura: My understanding is that any String that may be used across classes and has a chance that it might be modified in future should be place in a separate constants file. Now with ConfigConstants, it seems like the idea is to store only those constants that are placed in opensearch.yml. Since this particular string qualifies that I suggest we move it to ConfigConstants.

nibix: > My understanding is that any String that may be used across classes and has a chance that it might be modified in future should be place in a separate constants file.

The OpenSearch code base has no explicit coding style guide lines. However, I do not know any common style guide that gives this recommendation. 

IMHO, the general reason for using string constants is actually not changeability; all common editors have quite good search/replace functionalities. Rather, it is about avoiding typos. If you use literal strings, there is nothing that keeps you from using "opensorch.x.y" ... except possibly failing tests. Rather, if you use string constants, the compiler will complain about a typo in a constant. 

Still, this is not limited to string contants. Using the pattern in the code above, you can also get ahold of the string using `PRECOMPUTED_PRIVILEGES_MAX_HEAP_SIZE.getKey()`. Also, the property of "avoiding typos" is not dependent on a single big class containing all the constants. 

Actually, there are many guides that discourage collection of constants in big classes; for example https://www.baeldung.com/java-constants-good-practices#2-a-large-global-constants-class explicitly names "A Large Global Constants Class" as an anti-pattern.

On a high level, I try always to think about "what helps me and other codes" rather than about "how has it been done". I for myself, can say that I do not find the style employed in `ConfigConstants` helpful. It is difficult to navigate and I have to look at several files in order to learn about the details of a setting. On the other hand, keeping this stuff as closely as possible to the code that actually uses it, helps me a lot. If I one a class and see right away, what settings are applicable to that class. That's something, I'd consider very helpful.

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
nibix: As it is logged on debug level, it will be never logged in default configurations. One needs to selectively activate the level on the logger to have it looged. Then, it will logged whenever an index/alias/datastream is created or deleted.

Note: Most of the code of this PR is not new, it is just being moved from one file to another file.

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
nibix: The first one is also clean and compact, IMHO :)

src/integrationTest/java/org/opensearch/security/privileges/actionlevel/SubjectBasedActionPrivilegesTest.java
nibix: Sure!

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
DarshitChanpura: okay.. not a blocker.  I will resolve this conversation.

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
DarshitChanpura: I understand, I probly missed the context when the file was added.

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
DarshitChanpura: I agree with it becoming unstructured, but at the very least these should be declared as class-level variables. This solves both issues. This particular instance seems to only be used once and  hence it is not a blocker, I will resolve the conversation.

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
nibix: Could you elaborate on this? Not sure if I get this right ... `getIndex_patterns()` returns a `List<String>`.

src/main/java/org/opensearch/security/privileges/actionlevel/SubjectBasedActionPrivileges.java
cwperks: FYI for the security plugin to get access to the audit log index would require a pattern like `security-auditlog*` since its rolled over daily. It may not generally be the case to give access to a _specific_ index

nibix:
@willyborankin @cwperks I needed to resolve a merge conflict.  Can you approve again, please? :)

willyborankin (approved):

src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
DarshitChanpura: I see, I got confused with the name `actionToRolesWithWildcardIndexPrivileges` thinking that this will contains roles with index-patterns like `abc*` and not just "\*". But it looks like that list will hold all roles that contains explicit "*" in their index-pattern.

PR #5375, opened 2025-06-04
TLS support for auxiliary transports

finnegancarroll:
### Description

In core auxiliary transports may implement [secure variants ](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/plugins/NetworkPlugin.java#L172-L182). Secure aux transports expect to consume a [SecureAuxTransportSettingsProvider](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/plugins/SecureAuxTransportSettingsProvider.java) as provided by the security plugin. `SecureAuxTransportSettingsProvider` delivers an `SSLContext` to the transport, enabling TLS on client/server connections.

Currently the is no way to configure TLS for an auxiliary transport in the security plugin and the provided `SecureAuxTransportSettingsProvider` is always empty. This PR introduces settings which allow users to configure each aux transport to provide a valid `SecureAuxTransportSettingsProvider`.

Configuration settings for each aux transport exist under distinct namespaces, where the identifier for the transport is the same as the key for enabling it under setting `aux.transport.types`.
```
plugins.security.ssl.aux.arrow.<setting option...>
plugins.security.ssl.aux.grpc.<setting option...>
plugins.security.ssl.aux.other.<setting option...>
```

Note that these changes provide <ins>**no authorization component**</ins> and only work to deliver an SSL configuration. Users can configure `clientauth_mode` to allow/disallow certificates access to a particular transport, but no permissions are enforced and no user is associated with a successfully authenticated client.

### Testing

These changes add mock aux transports to unit tests for `SslContextHandler`.
However security plugin is agnostic as to the existence of other plugins available to OpenSearch.
Since auxiliary transports are implemented as plugins validating these changes was done manually by installing security plugin and the secure gRPC transport on a single node cluster.

Env variables for convenience:
```
export PLAT=darwin
export SNAP_VERSION=3.2.0
export REPOS_DIR=<path to clone OS and security feature branches>

export OS_REPO=${REPOS_DIR}OpenSearch
export SEC_REPO=${REPOS_DIR}security
export SEC_TAR=${SEC_REPO}/build/distributions/opensearch-security-${SNAP_VERSION}.0-SNAPSHOT.zip
export GRPC_TAR=${OS_REPO}/plugins/transport-grpc/build/distributions/transport-grpc-${SNAP_VERSION}-SNAPSHOT.zip
export OS_INSTALL=${OS_REPO}/distribution/archives/${PLAT}-tar/build/install/opensearch-${SNAP_VERSION}-SNAPSHOT
```

Build OpenSearch/security plugin/gRPC plugin:
```
cd ${OS_REPO} && ./gradlew :distribution:archives:${PLAT}-tar:assemble 
cd ${OS_REPO} && ./gradlew plugins:transport-grpc:assemble
cd ${OS_REPO} && ./gradlew publishToMavenLocal
cd ${SEC_REPO} && ./gradlew :assemble
```

Install both plugins:
```
${OS_INSTALL}/bin/opensearch-plugin install file://${GRPC_TAR}
${OS_INSTALL}/bin/opensearch-plugin install file://${SEC_TAR}
```

Create demo certs for security plugin - Respond yes yes no (keeping cluster mode disabled):
```
export OPENSEARCH_INITIAL_ADMIN_PASSWORD=PrTestPass369char! && \
chmod +x "${OS_INSTALL}/plugins/opensearch-security/tools/install_demo_configuration.sh" && \
"${OS_INSTALL}/plugins/opensearch-security/tools/install_demo_configuration.sh"
```

Enable the secure gRPC auxiliary transport:
```
echo "aux.transport.types: experimental-secure-transport-grpc" >> ${OS_INSTALL}/config/opensearch.yml
echo "aux.transport.experimental-secure-transport-grpc.port: '9400-9500'" >> ${OS_INSTALL}/config/opensearch.yml
```

Configure security settings with demo certs (.pem format):
Note: This example configures `clientauth_mode: REQUIRE`.
```
echo "plugins.security.ssl.aux.experimental-secure-transport-grpc.enabled: true" >> ${OS_INSTALL}/config/opensearch.yml
echo "plugins.security.ssl.aux.experimental-secure-transport-grpc.pemcert_filepath: esnode.pem" >> ${OS_INSTALL}/config/opensearch.yml
echo "plugins.security.ssl.aux.experimental-secure-transport-grpc.pemkey_filepath: esnode-key.pem" >> ${OS_INSTALL}/config/opensearch.yml
echo "plugins.security.ssl.aux.experimental-secure-transport-grpc.clientauth_mode: REQUIRE" >> ${OS_INSTALL}/config/opensearch.yml
echo "plugins.security.ssl.aux.experimental-secure-transport-grpc.pemtrustedcas_filepath: root-ca.pem" >> ${OS_INSTALL}/config/opensearch.yml
```

Start OpenSearch:
```
${OS_INSTALL}/bin/opensearch
```

Test plaintext connection:
```
grpcurl -plaintext localhost:9400 list
```
Test insecure TLS with no client certificate provided:
```
grpcurl -insecure localhost:9400 list
```
Test TLS with client certificate provided:
Note: `-insecure` to skip host name verification.
```
grpcurl -insecure -cert ${OS_INSTALL}/config/esnode.pem -key ${OS_INSTALL}/config/esnode-key.pem localhost:9400 list
```

### Issues Resolved
[#17795](https://github.com/opensearch-project/OpenSearch/issues/17795)

### Testing
Unit tests and integration tests for mock auxiliary transport CertTypes.

### Check List
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md)


sample-resource-plugin/build.gradle
```diff
@@ -56,6 +56,13 @@ configurations.all {
         force 'org.hamcrest:hamcrest:2.2'
         force 'org.mockito:mockito-core:5.18.0'
         force 'org.slf4j:slf4j-api:1.7.36'
+
+        // Duplicate dependencies with conflicting versions from org.opensearch.plugin:transport-grpc
+        // Force versions present in security plugin
+        force "com.google.errorprone:error_prone_annotations:2.36.0"
+        force "com.google.protobuf:protobuf-java:3.25.5"
```
cwperks: This version is present in the version catalog: https://github.com/opensearch-project/OpenSearch/blob/main/gradle/libs.versions.toml#L24

Can we reference that version using the `"com.google.protobuf:protobuf-java:${versions.protobuf}"` syntax

src/integrationTest/java/org/opensearch/security/grpc/GrpcHelpers.java
```diff
@@ -0,0 +1,158 @@
...
+        true,
+        SECURITY_SSL_AUX_PEMKEY_FILEPATH.getConcreteSettingForNamespace(GRPC_SECURE_TRANSPORT_SETTING_KEY).getKey(),
+        TEST_CERTIFICATES.getNodeKey(0, null).getAbsolutePath(),
+        SECURITY_SSL_AUX_PEMCERT_FILEPATH.getConcreteSettingForNamespace(GRPC_SECURE_TRANSPORT_SETTING_KEY).getKey(),
+        TEST_CERTIFICATES.getNodeCertificate(0).getAbsolutePath(),
+        SECURITY_SSL_AUX_PEMTRUSTEDCAS_FILEPATH.getConcreteSettingForNamespace(GRPC_SECURE_TRANSPORT_SETTING_KEY).getKey(),
+        TEST_CERTIFICATES.getRootCertificate().getAbsolutePath()
+    );
+
+    public static LocalCluster.Builder baseGrpcCluster() {
```
cwperks: I know there might be quite a bit of repetition, but I would prefer not to have this method and instead copy this config where needed so that we can do other tests in the future like multiple nodes, with authc, etc. This is a very base level of config with single node and SSL only mode.

src/integrationTest/java/org/opensearch/security/grpc/GrpcHelpers.java
```diff
@@ -0,0 +1,158 @@
...
+        ClientAuth.OPTIONAL.name()
+    );
+    protected static final Map<String, Object> CLIENT_AUTH_REQUIRE = Map.of(
+        SECURITY_SSL_AUX_CLIENTAUTH_MODE.getConcreteSettingForNamespace(GRPC_SECURE_TRANSPORT_SETTING_KEY).getKey(),
+        ClientAuth.REQUIRE.name()
+    );
+
+    private static final PortsRange PORTS_RANGE = new PortsRange("9400-9500");
+    private static final String HOST_ADDR = "localhost";
+    private static final Map<String, Object> SECURE_GRPC_TRANSPORT_SETTINGS = Map.of(
```
cwperks: We need to figure out how to extend this in the future to support testing with multiple nodes.

src/integrationTest/java/org/opensearch/security/grpc/GrpcHelpers.java
```diff
@@ -0,0 +1,158 @@
...
+            .build();
+        return Grpc.newChannelBuilderForAddress(HOST_ADDR, PORTS_RANGE.ports()[0], credentials).build();
+    }
+
+    /*
+    TLS with client certificate not trusted by server.
+    */
+    public static ManagedChannel secureUntrustedChannel() throws IOException {
+        ChannelCredentials credentials = TlsChannelCredentials.newBuilder()
+            .keyManager(UN_TRUSTED_TEST_CERTIFICATES.getNodeCertificate(0), UN_TRUSTED_TEST_CERTIFICATES.getNodeKey(0, null))
```
cwperks, line 130: What is the keyManager for here and in the method above? I thought the difference between secureChannel and insecureChannel would be in the trust manager. In insecure (ssl verification disabled) it would trust all, but in trusted then it needs to know the root certs to trust.

src/integrationTest/java/org/opensearch/security/grpc/GrpcHelpers.java
```diff
@@ -0,0 +1,158 @@
...
+            .trustManager(InsecureTrustManagerFactory.INSTANCE.getTrustManagers())
+            .build();
+        return Grpc.newChannelBuilderForAddress(HOST_ADDR, PORTS_RANGE.ports()[0], credentials).build();
+    }
+
+    public static BulkResponse doBulk(ManagedChannel channel, String index, long numDocs) {
+        BulkRequest.Builder requestBuilder = BulkRequest.newBuilder().setRefresh(Refresh.REFRESH_TRUE);
+        for (int i = 0; i < numDocs; i++) {
+            String docBody = "{\"field\": \"doc " + i + " body\"}";
```
cwperks: nit: prefer to use TextBlocks for JSON payloads now that java min compatibility is 21

src/integrationTest/java/org/opensearch/security/grpc/GrpcHelpers.java
finnegancarroll: This trust store is from client perspective and informs which server certificates the client trusts. This is up to the users client configuration and purely client functionality.

The client key manager here contains the relevant certificates which are presented to the server during handshake and compared against the server's own truststore. The server will allow or reject the connection based on these.

src/integrationTest/java/org/opensearch/security/grpc/GrpcHelpers.java
finnegancarroll: Renamed this for clarity. I figure these tests do not require multi node clusters but it should be easy to implement this in the future as `LocalCluster` supports and uses this functionality in other places.

sample-resource-plugin/build.gradle
```diff
@@ -56,6 +56,13 @@ configurations.all {
         force 'org.hamcrest:hamcrest:2.2'
         force 'org.mockito:mockito-core:5.18.0'
         force 'org.slf4j:slf4j-api:1.7.36'
+
+        // Duplicate dependencies with conflicting versions from org.opensearch.plugin:transport-grpc
+        // Force versions present in security plugin
+        force "com.google.errorprone:error_prone_annotations:2.36.0"
+        force "com.google.protobuf:protobuf-java:${versions.protobuf}"
+        force "com.google.guava:guava:33.4.8-jre"
```
cwperks: We already have a variable called `guava_version`, can we use that value here?

cwperks: ok, I see this is in the sample resource plugin. 

This is fine as is, but I'd like to define a common version in the root project and use it in sub projects.

src/main/java/org/opensearch/security/ssl/config/CertType.java
```diff
@@ -11,32 +11,142 @@
...
+        return this.id().equals(certType.id());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(this.id());
+    }
+
+    /*
+    Write only set for tracking certificate types discovered and registered on a node.
```
cwperks: nit: fix the javadoc here

src/main/java/org/opensearch/security/ssl/config/CertType.java
```diff
@@ -11,32 +11,142 @@
...
+    Not all ssl context configurations are known at compile time, so we track newly discovered CertTypes here.
+    */
+    public static class NodeCertTypeRegistry implements Iterable<CertType> {
+        private final Set<CertType> RegisteredCertType = new HashSet<>();
```
cwperks: nit: use lowerCamelCase for this var

cwperks (dismissed):
The changes in this PR look ok to me. I understand that gRPC tests will evolve as more functionality is added so we can add additional tests like multi-node tests in subsequent PRs. Its nice to see the test infrastructure put into place for testing out gRPC in a secure channel.

sample-resource-plugin/build.gradle
finnegancarroll: Woops, yes moving to `guava_version` here.

cwperks:
@finnegancarroll as more gRPC tests are added, it would be necessary to use the PortAllocator (see [example](https://github.com/opensearch-project/security/blob/main/src/integrationTest/java/org/opensearch/test/framework/cluster/LocalOpenSearchCluster.java#L162-L167)) for assigning ports to in-memory nodes to avoid collision. Port collision is a cause for flakiness in this repo.

cwperks:
We could also consider a separate test group for gRPC tests to separate them from the rest

cwperks (approved):

finnegancarroll:
Hi @reta, would you mind taking a look at these changes when you have a chance?

build.gradle
```diff
@@ -530,6 +533,10 @@ allprojects {
         integrationTestImplementation("org.opensearch.plugin:reindex-client:${opensearch_version}"){
             exclude(group: 'org.slf4j', module: 'slf4j-api')
         }
+        integrationTestImplementation "org.opensearch:protobufs:0.3.0"
```
reta, line 536: @finnegancarroll @cwperks I don't think this is a sustainable approach for `security` plugin: the plugin only provides relevant security settings and that is about it. gRPC transport plugin could (and should) have the tests assuming the security settings are provided (in general, by any plugin out there).

Would we also add tests for Apache Arrow Flight? Or any other auxiliary plugin that is outside of the OpenSearch organization? I don't think we would. In any case, @cwperks leaving the decision to you.

src/main/java/org/opensearch/security/ssl/config/CertType.java
```diff
@@ -11,32 +11,139 @@
...
+        public Iterator<CertType> iterator() {
+            return Collections.unmodifiableSet(registeredCertType).iterator();
+        }
     }
 
+    /*
+    Mandatory transports.
+    */
+    public static CertType HTTP = new CertType(SSL_HTTP_PREFIX);
+    public static CertType TRANSPORT = new CertType(SSL_TRANSPORT_PREFIX);
```
reta, line 123: Why we have 3 settings for `TRANSPORT` ? What is difference between TRANSPORT , TRANSPORT_CLIENT and TRANSPORT_SERVER?

src/main/java/org/opensearch/security/dlic/rest/api/ssl/CertificatesInfo.java
```diff
@@ -32,20 +31,20 @@ public CertificatesInfo(final Map<CertType, List<CertificateInfo>> certificates)
     }
 
     public CertificatesInfo(final StreamInput in) throws IOException {
-        certificates = in.readMap(keyIn -> keyIn.readEnum(CertType.class), listIn -> listIn.readList(CertificateInfo::new));
+        certificates = in.readMap(CertType::new, listIn -> listIn.readList(CertificateInfo::new));
```
reta: In the mixed clusters, this is going to fail since the serde between old CertType (enum) and new CertType (class) is incompatible

build.gradle
cwperks: But what about when auth is introduced for grpc transport? Any auth related tests would be appropriate for the security plugin.

nibix: How is the dependency structure going to look like? If we introduce auth of grpc, does the security plugin need to depend on the grpc plugin?

src/integrationTest/java/org/opensearch/security/grpc/GrpcClientAuthNoneTests.java
```diff
@@ -0,0 +1,97 @@
...
+        SearchResponse searchResp = GrpcHelpers.doMatchAll(channel, testIndex, 10);
+        assertThat("Search response should not be null", searchResp != null);
+        assertThat(
+            "Search response should indicate success",
+            searchResp.getResponseCase().getNumber() == SearchResponse.ResponseCase.RESPONSE_BODY.getNumber()
+        );
+        assertThat(
+            "Search response has correct hits count",
+            searchResp.getResponseBody().getHits().getTotal().getTotalHits().getValue() == testDocs
+        );
```
nibix: Nit: IMHO, `assertThat()` only makes sense when a `org.hamcrest.Matcher` is passed. If just boolean expressions are used, one should prefer the more expressive `assertEquals()` (or similar) from JUnit.

src/integrationTest/java/org/opensearch/security/grpc/GrpcHelpers.java
```diff
@@ -0,0 +1,169 @@
...
+import static org.opensearch.security.ssl.util.SSLConfigConstants.SECURITY_SSL_AUX_PEMTRUSTEDCAS_FILEPATH;
+import static org.opensearch.transport.AuxTransport.AUX_TRANSPORT_TYPES_KEY;
+import static com.carrotsearch.randomizedtesting.RandomizedTest.randomFrom;
+import static io.grpc.internal.GrpcUtil.NOOP_PROXY_DETECTOR;
+
+public class GrpcHelpers {
+    protected static final TestCertificates TEST_CERTIFICATES = new TestCertificates();
+    protected static final TestCertificates UN_TRUSTED_TEST_CERTIFICATES = new TestCertificates();
+    protected static final Map<String, Object> CLIENT_AUTH_NONE = Map.of(
+        SECURITY_SSL_AUX_CLIENTAUTH_MODE.getConcreteSettingForNamespace(GRPC_SECURE_TRANSPORT_SETTING_KEY).getKey(),
```
nibix: Nit: In test code, I would already recomment NOT to use constants for user facing values (such as settings names) from the tested code base.

The reason: If the value contained in the constant is accidentially changed, the tests continue to be green. However, for all actual users, the functionality will break, as the previously assumed value won't work any more.

src/test/java/org/opensearch/security/ssl/SslSettingsManagerTest.java
```diff
@@ -52,22 +58,121 @@
...
+            .put(MOCK_AUX_CERT_TYPE_FOO_PEMCERT_FILEPATH, path(MOCK_AUX_CERT_TYPE_FOO_PEMCERT_NAME))
+            .put(MOCK_AUX_CERT_TYPE_FOO_PEMKEY_FILEPATH, path(MOCK_AUX_CERT_TYPE_FOO_PEMKEY_NAME));
+    }
+
+    private void withAuxBarSslSettings(final Settings.Builder settingsBuilder) {
+        settingsBuilder.put(ENABLE_BAR_SETTINGS_BUILDER)
+            .put(MOCK_AUX_CERT_TYPE_BAR_PEMTRUSTEDCAS_FILEPATH, path(MOCK_AUX_CERT_TYPE_BAR_PEMTRUSTEDCAS_NAME))
+            .put(MOCK_AUX_CERT_TYPE_BAR_PEMCERT_FILEPATH, path(MOCK_AUX_CERT_TYPE_BAR_PEMCERT_NAME))
+            .put(MOCK_AUX_CERT_TYPE_BAR_PEMKEY_FILEPATH, path(MOCK_AUX_CERT_TYPE_BAR_PEMKEY_NAME));
+    }
```
nibix: Nit: This code looks fragile. It will break when at one point someone forgets to update a property in the settingsBuilder. Then, an old value from an undefined previous test will be reused.

Can't we use a structure where a fresh settings builder is created for each test?

nibix (commented):
Thank you for this, cool achievement!

I mostly have nit comments. The comment from @reta on the serialization of Enum values is a blocker, though: https://github.com/opensearch-project/security/pull/5375#pullrequestreview-3031286191

src/integrationTest/java/org/opensearch/security/grpc/GrpcClientAuthRequireTests.java
```diff
@@ -0,0 +1,75 @@
...
+import static org.opensearch.security.grpc.GrpcHelpers.CLIENT_AUTH_REQUIRE;
+import static org.opensearch.security.grpc.GrpcHelpers.SINGLE_NODE_SECURE_GRPC_TRANSPORT_SETTINGS;
+import static org.opensearch.security.grpc.GrpcHelpers.TEST_CERTIFICATES;
+import static org.opensearch.security.grpc.GrpcHelpers.getSecureGrpcEndpoint;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class GrpcClientAuthRequireTests {
```
nibix, line 36: I have one question: We are testing here client auth - yet, after the client auth on the TLS layer, do I understand it correctly that no further internal authc (or authz) is performed?

Especially regarding authz, what will be the behavior of OpenSearch when a GRPC client successfully authenticated by client auth ?

build.gradle
cwperks: I imagine it would be somewhat similar to the default netty transport where it overrides the header verifier: https://github.com/opensearch-project/security/blob/1f6fc7826b35fe610d57ebedb2ee8444be72c55e/src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java#L206-L211

There would need to be an extension point on the [serverBuilder](https://github.com/opensearch-project/OpenSearch/blob/main/plugins/transport-grpc/src/main/java/org/opensearch/plugin/transport/grpc/Netty4GrpcServerTransport.java#L358-L369) that security could hook into to supply a [ServerInterceptor](https://github.com/grpc/grpc-java/blob/master/api/src/main/java/io/grpc/ServerInterceptor.java#L38-L59).

reta: >But what about when auth is introduced for grpc transport? Any auth related tests would be appropriate for the security plugin.

This is very different subject and I suspect not easy to solve - the aux transports are oblivious to core and have no injection points / hooks for auth or alike (it is just a port or range of ports).

cwperks: good point wrt auth for auxiliary transports. Ideally, any abstraction we introduce would also be applicable to default netty as well.

First thoughts I have are to add a method to [TransportAdapterProvider](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/transport/TransportAdapterProvider.java#L16-L39) that could be as simple as `boolean verifyHeaders() throws OpenSearchSecurityException` and then utilize that in the core accordingly.

src/integrationTest/java/org/opensearch/security/grpc/GrpcClientAuthRequireTests.java
finnegancarroll: > after the client auth on the TLS layer, do I understand it correctly that no further internal authc (or authz) is performed?

Yes this is correct. Here we only address EIT with authz as a follow up. Users connecting with TLS will not be associated with security plugin roles (the APIs and settings to do so do not exist). gRPC clients execute commands agnostic to index/document/operation level permissions.

src/main/java/org/opensearch/security/ssl/config/CertType.java
finnegancarroll: Users are given the option of configuring TLS settings separately for when the node-to-node transport operates as a client vs as a server.

https://docs.opensearch.org/docs/latest/security/configuration/tls/#separate-client-and-server-certificates-for-transport-layer-tls

build.gradle
finnegancarroll: > But what about when auth is introduced for grpc transport? Any auth related tests would be appropriate for the security plugin.

I think for these EIT changes [similar tests](https://github.com/opensearch-project/OpenSearch/blob/main/plugins/transport-grpc/src/internalClusterTest/java/org/opensearch/plugin/transport/grpc/ssl/SecureNetty4GrpcServerTransportIT.java) are present in core for this feature set and are definitely a bit redundant with the tests included in this PR. However pretty much all of authz functionality will be located within security plugin, with any given transport bearing only the responsibility of injecting a handler/interceptor/ect into its execution pipeline. It makes sense to me that integration tests for this feature would need to be in security plugin for these future changes.

>How is the dependency structure going to look like? If we introduce auth of grpc, does the security plugin need to depend on the grpc plugin?

Yes but only as a test dependency. I think as support in security plugin is expanded for non-REST client/server transports, gRPC is a good candidate to test these new features. It is both a native plugin to the core repo and mirrors the feature set of the http/REST transport, so it's well positioned to take advantage of the existing test suites present in security plugin and validate security resources provided to auxiliary transports work end-to-end.

src/main/java/org/opensearch/security/ssl/config/CertType.java
reta: So TRANSPORT (server) and TRANSPORT_CLIENT  (client) should be enough?

src/integrationTest/java/org/opensearch/security/grpc/GrpcHelpers.java
finnegancarroll: Makes sense. Replacing these settings in test suites with literals.

src/test/java/org/opensearch/security/ssl/SslSettingsManagerTest.java
finnegancarroll: Updated these to provide a new builder instance.

src/main/java/org/opensearch/security/dlic/rest/api/ssl/CertificatesInfo.java
finnegancarroll: Added version guards here and verified the fix by adding `/certificates` case to the BWC suite.

build.gradle
finnegancarroll: Resolving this as there doesn't seem to be clear alternative testing plan for authn/authz aux transport functionality. Additionally, while the gRPC transport is currently limited in API support, the long term goal for this transport is to mirror the REST API and provide a comprehensive alternative. I think mimicking the existing REST tests as they exist in security plugin is in line with this plan.

finnegancarroll: Moving discussion to a meta issue: https://github.com/opensearch-project/security/issues/5524
Can these changes take gRPC plugin as a test only dependency for now while I follow up on this AI?

cwperks:
@DarshitChanpura @nibix wdyt about having aux transport tests in this repo and testing for EIT? I would be happy to accept the maintenance in this case and if there are other aux transports in the future (f.e. Apache arrow as an alternative to default netty) then we can find another appropriate place for the tests at a later time.

The changes in this PR should get merged to supply the settings, so the current point of contention is where the appropriate place for the tests is.

build.gradle
cwperks: Marking as resolved and we will continue the discussion on https://github.com/opensearch-project/security/issues/5524

nibix:
@cwperks 

> wdyt about having aux transport tests in this repo and testing for EIT? I would be happy to accept the maintenance in this case and if there are other aux transports in the future (f.e. Apache arrow as an alternative to default netty) then we can find another appropriate place for the tests at a later time.

Well, as a temporary solution, it's okay. Still, I would also think that it won't be sustainable in the long run, as each further transport has these effects:

- increase of dependencies, and thus increased chance of issues with dependencies (as we see right now)
- increase of test turnaround times
- it's unclear what shall be done when a test fails

IMHO, the infrastructure that is provided by the security plugin should be tested by a minimal test implementation that is co-located with the tests.

Wouldn't it be possible to have the integration tests with security on the GRPC plugin side?

cwperks:
> Wouldn't it be possible to have the integration tests with security on the GRPC plugin side?

gRPC plugin is in core and currently none of the core plugins test with security...not saying its not possible but typically only plugins with their own repo test with security.

I see 2 solutions:

1. Introduce testing w/ security in core plugins
2. Move gRPC out of core to a separate repo

nibix:
Ah, I did not know that gRPC is part of the core project. That changes the situation a bit, I thought it was a separate plugin.

cwperks:
> I am wondering: what are the criteria to decide what goes to the core repo and what goes to its own repo?

That's an art, not a science ;). In all seriousness, I agree that we should have criteria somewhere to help people distinguish when to do either.

My general guidance is this: if you want a separate set of maintainers go with a separate repo. If the core maintainers agree to do the maintenance then it can go in core. So really it boils down to who does the maintenance..I think.

cwperks:
IMO Workload Management is another area that was introduced in the core that might be better as a separate plugin repo as they are also talking about doing testing with security.

For example, they are assigning requests to workload groups and some of the matching rules involve user attributes like username and backend roles.

CC: @kkhatua @kaushalmahi12 @jainankitk you may be interested in this conversation

cwperks:
> Are there BWC implications of doing this since it would no longer be available natively?

No issues with bwc. Only functional changes to plugins have bwc considerations..not where the plugin actually resides. If a plugin is the same code, then it would produce the same binary whether located in its own repo or in the core.

kaushalmahi12:
@cwperks I believe even if we move authN/authZ into or out of the core, it will still execute before the actual RestHandlers hence I assume rest layer will continue to access the same authN/authZ info via ThreadContext/SomeOtherMechanism in future.

Is my understanding incorrect on this ?

PR #5376, opened 2025-06-05
Integrate search-relevance functionalities with security plugin

fen-qin:
### Description
Adds search-relevance related roles to the plugin: 
 - read only
 - full access.

Adds search-relevance system indices to demo scripts:
- `.plugins-search-relevance-experiment`
- `.plugins-search-relevance-judgment-cache`

### Issues Resolved
This is a new feature for 3.1 release, no backfort

### Testing
N/A

### Check List

### DIFF
--- a/config/roles.yml
+++ b/config/roles.yml
@@ -471,3 +471,24 @@ ltr_full_access:
     reserved: true
     cluster_permissions:
       - cluster:admin/ltr/*
+
+# Allow users to use all Search Relevance functionalities
+search_relevance_full_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/opensearch/search_relevance/*'
+  index_permissions:
+    - index_patterns:
+        - '*'
+      allowed_actions:
+        - 'indices:admin/mappings/get'
+        - 'indices:data/read/search*'
+
+# Allow users to read Search Relevance resources
+search_relevance_read_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/opensearch/search_relevance/experiment/get'
+    - 'cluster:admin/opensearch/search_relevance/judgment/get'
+    - 'cluster:admin/opensearch/search_relevance/queryset/get'
+    - 'cluster:admin/opensearch/search_relevance/search_configuration/get'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Github workflow for changelog verification ([#5318](https://github.com/opensearch-project/security/pull/5318))
 - Register cluster settings listener for `plugins.security.cache.ttl_minutes` ([#5324](https://github.com/opensearch-project/security/pull/5324))
 - Add flush cache endpoint for individual user ([#5337](https://github.com/opensearch-project/security/pull/5337))
+- Integrate search-relevance functionalities with security plugin ([#5376](https://github.com/opensearch-project/security/pull/5376))
 
 ### Changed
 - Use extendedPlugins in integrationTest framework for sample resource plugin testing ([#5322](https://github.com/opensearch-project/security/pull/5322))
--- a/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
+++ b/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java
@@ -74,7 +74,9 @@ public class SecuritySettingsConfigurer {
         ".geospatial-ip2geo-data*",
         ".plugins-flow-framework-config",
         ".plugins-flow-framework-templates",
-        ".plugins-flow-framework-state"
+        ".plugins-flow-framework-state",
+        ".plugins-search-relevance-experiment",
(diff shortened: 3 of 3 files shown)


heemin32 (approved):

config/roles.yml
```diff
@@ -471,3 +471,27 @@ ltr_full_access:
...
       - cluster:admin/ltr/*
+
+# Allow users to use all Search Relevance functionalities
+search_relevance_full_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/search_relevance/query_sets/*'
+    - 'cluster:admin/search_relevance/search_configurations/*'
+    - 'cluster:admin/search_relevance/judgments/*'
+    - 'cluster:admin/search_relevance/experiments/*'
```
martin-gaievski: some indices/apis are missing here: evaluation_result and experiment_variant, is this intentional?

config/roles.yml
```diff
@@ -471,3 +471,27 @@ ltr_full_access:
...
+        - 'indices:data/read/search*'
+
+# Allow users to read Search Relevance resources
+search_relevance_read_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/search_relevance/query_sets/get'
+    - 'cluster:admin/search_relevance/search_configurations/get'
+    - 'cluster:admin/search_relevance/judgments/get'
+    - 'cluster:admin/search_relevance/experiments/get'
```
martin-gaievski: same question as in previous comment - with read-only access they still need to see evaluation results, and maybe experiment_variant (I'm not 100% sure because we don't have UI for it yet)

config/roles.yml
fen-qin: oh, I see. the cluster admin is for index based instead of api based. will update the commit.

fen-qin: I don't think we need to add standard indices here.

config/roles.yml
fen-qin: probably if it's not a system indices, then it should be accessible to all.

config/roles.yml
fen-qin: let me correct my words.

- `cluster:admin` is for transport API based and it should be defined in transport actions
- https://github.com/fen-qin/flow-framework/blob/main/src/main/java/org/opensearch/flowframework/common/CommonValue.java#L64
- so, since evalaution_results and experiment_variant have no API wrapper, so we don't have to add them to cluster admin

config/roles.yml
martin-gaievski: but why other indices are there? I think we do have only 2 system indices -  `.plugins-search-relevance-experiment` and `.plugins-search-relevance-judgment-cache`, the rest is regular/general indices. How `evaluation_result` and `experiment_variant` are different in this sense?

fen-qin: I tried to update this PR
- cluster:admin is for transport actions. In search-relevance, we defined transport actions with this prefix
```
public static final String TRANSPORT_ACTION_NAME_PREFIX = "cluster:admin/opensearch/search_relevance/";
```
- system indices was added to SecuritySettingConfigurer.

cwperks: FYI The list of system indices in the security plugin is no longer the recommended way to declare system indices.

Prefer to use [SystemIndexPlugin.getSystemIndexDescriptors()](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/plugins/SystemIndexPlugin.java#L55-L57) extension point. Security uses the registry from core now for system index protection.

fen-qin: yes. we've registered the system indicies here - https://github.com/opensearch-project/search-relevance/blob/main/src/main/java/org/opensearch/searchrelevance/plugin/SearchRelevancePlugin.java#L120
thanks, Craig!

fen-qin:
> Let's also add a CHANGELOG entry

Hello, Craig
I've addressed all comments. put the PR from draft to review. Thank you!

cwperks (approved):

config/roles.yml
```diff
@@ -471,3 +471,24 @@ ltr_full_access:
...
+        - 'indices:admin/mappings/get'
+        - 'indices:data/read/search*'
+
+# Allow users to read Search Relevance resources
+search_relevance_read_access:
+  reserved: true
+  cluster_permissions:
+    - 'cluster:admin/opensearch/search_relevance/experiment/get',
+    - 'cluster:admin/opensearch/search_relevance/judgment/get',
+    - 'cluster:admin/opensearch/search_relevance/queryset/get',
```
cwperks: @fen-qin remove the commas at the end of these lines

cwperks: no need for commas in this file, its yaml

DarshitChanpura (commented):
only one comment, commas are not needed at the end of actions in the yaml file.

config/roles.yml
fen-qin: raised another revision, resolve the issue. Thanks, Craig

fen-qin:
> only one comment, commas are not needed at the end of actions in the yaml file.

Thanks, Darshit. resolve this comment

DarshitChanpura (dismissed):

RyanL1997 (approved):

PR #5389, opened 2025-06-11
[Resource Sharing] Adds REST API to migrate existing sharing info to security plugin

DarshitChanpura:
### Description
This change will allow cluster-admins to migrate resource-sharing info from resource-documents stored in plugin-index to the centrally managed backing indices in security plugin.
* Category : Enhancement
    * To enable usage of resource-sharing feature on existing clusters
   *  Without this API, resource-sharing feature will only be available on fresh clusters. 

### Issues Resolved
- #5391 

### Testing
- automated tests

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,20 +20,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 * Fix compilation issue after change to Subject interface in core and bump to 3.2.0 ([#5423](https://github.com/opensearch-project/security/pull/5423))
 * Provide SecureHttpTransportParameters to complement SecureTransportParameters counterpart ([#5432](https://github.com/opensearch-project/security/pull/5432))
+* Use isClusterPerm instead of requestedResolved.isLocalAll() to determine if action is a cluster action ([#5445](https://github.com/opensearch-project/security/pull/5445))
+
 
 ### Refactoring
 
 * Refactor JWT Vender to take a claims builder and rename oboEnabled to enabled ([#5436](https://github.com/opensearch-project/security/pull/5436))
 
 ### Maintenance
-
 - Bump `org.eclipse.platform:org.eclipse.core.runtime` from 3.33.0 to 3.33.100 ([#5400](https://github.com/opensearch-project/security/pull/5400))
 - Bump `org.eclipse.platform:org.eclipse.equinox.common` from 3.20.0 to 3.20.100 ([#5402](https://github.com/opensearch-project/security/pull/5402))
 - Bump `spring_version` from 6.2.7 to 6.2.8 ([#5403](https://github.com/opensearch-project/security/pull/5403))
 - Bump `stefanzweifel/git-auto-commit-action` from 5 to 6 ([#5401](https://github.com/opensearch-project/security/pull/5401))
-- Bump `com.github.spotbugs` from 5.2.5 to 6.3.0 and checkstyle to 10.25.0 ([#5409](https://github.com/opensearch-project/security/pull/5409))
+- Bump `com.github.spotbugs` from 5.2.5 to 6.2.1 ([#5409](https://github.com/opensearch-project/security/pull/5409), [#5450](https://github.com/opensearch-project/security/pull/5450))
 - Bump `org.codehaus.plexus:plexus-utils` from 3.3.0 to 3.6.0 ([#5429](https://github.com/opensearch-project/security/pull/5429))
 - Bump `net.bytebuddy:byte-buddy` from 1.17.5 to 1.17.6 ([#5427](https://github.com/opensearch-project/security/pull/5427))
+- Bump `io.dropwizard.metrics:metrics-core` from 4.2.32 to 4.2.33 ([#5428](https://github.com/opensearch-project/security/pull/5428))
+- Bump `org.junit.jupiter:junit-jupiter-api` from 5.13.1 to 5.13.2 ([#5446](https://github.com/opensearch-project/security/pull/5446))
 - Bump `io.dropwizard.metrics:metrics-core` from 4.2.32 to 4.2.33 ([#5428](https://github.com/opensearch-project/security/pull/5428))
 
 ### Documentation
--- a/build.gradle
+++ b/build.gradle
@@ -71,7 +71,7 @@ plugins {
     id 'com.netflix.nebula.ospackage' version "11.11.2"
     id "org.gradle.test-retry" version "1.6.2"
     id 'eclipse'
-    id "com.github.spotbugs" version "6.2.0"
+    id "com.github.spotbugs" version "6.2.1"
     id "com.google.osdetector" version "1.7.3"
 }
 
@@ -756,8 +756,8 @@ dependencies {
     testImplementation 'commons-validator:commons-validator:1.9.0'
     testImplementation "org.springframework.kafka:spring-kafka-test:4.0.0-M2"
     testImplementation "org.springframework:spring-beans:${spring_version}"
-    testImplementation 'org.junit.jupiter:junit-jupiter:5.13.1'
-    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.13.1'
+    testImplementation 'org.junit.jupiter:junit-jupiter:5.13.2'
+    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.13.2'
     testImplementation('org.awaitility:awaitility:4.3.0') {
         exclude(group: 'org.hamcrest', module: 'hamcrest')
     }
--- a/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/secure/SecurePluginTests.java
+++ b/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/secure/SecurePluginTests.java
@@ -79,6 +79,21 @@ public class SecurePluginTests {
     }
 
     @Test
+    public void testRunClusterStateWithPluginSubject() {
+        try (TestRestClient client = cluster.getRestClient(USER_ADMIN)) {
+            HttpResponse response = client.postJson(SAMPLE_RESOURCE_PLUGIN_PREFIX + "/run_action", """
+                {
+                    "action": "cluster:monitor/state",
+                    "index": ".opendistro_security"
+                }
+                """);
+
+            assertThat(response.getStatusCode(), equalTo(RestStatus.OK.getStatus()));
+            assertThat(response.getBody(), containsString("clusterState"));
+        }
+    }
+
+    @Test
     public void testRunCreateIndexWithPluginSubject() {
         try (TestRestClient client = cluster.getRestClient(USER_ADMIN)) {
             HttpResponse response = client.postJson(SAMPLE_RESOURCE_PLUGIN_PREFIX + "/run_action", """
--- a/sample-resource-plugin/src/main/java/org/opensearch/sample/secure/actions/transport/SecurePluginTransportAction.java
+++ b/sample-resource-plugin/src/main/java/org/opensearch/sample/secure/actions/transport/SecurePluginTransportAction.java
@@ -13,6 +13,8 @@ import org.apache.logging.log4j.Logger;
 
 import org.opensearch.action.admin.cluster.health.ClusterHealthAction;
 import org.opensearch.action.admin.cluster.health.ClusterHealthRequest;
+import org.opensearch.action.admin.cluster.state.ClusterStateAction;
+import org.opensearch.action.admin.cluster.state.ClusterStateRequest;
 import org.opensearch.action.admin.indices.create.CreateIndexAction;
 import org.opensearch.action.admin.indices.create.CreateIndexRequest;
 import org.opensearch.action.support.ActionFilters;
@@ -60,6 +62,17 @@ public class SecurePluginTransportAction extends HandledTransportAction<SecurePl
                             String.valueOf(clusterHealthResponse.toXContent(XContentFactory.jsonBuilder(), ToXContent.EMPTY_PARAMS))
                         )
                     ),
+                    listener::onFailure
+                )
+            );
+            return;
+        } else if (ClusterStateAction.NAME.equals(action)) {
+            String index = request.getIndex();
+            pluginClient.execute(
+                ClusterStateAction.INSTANCE,
+                new ClusterStateRequest().indices(index),
+                ActionListener.wrap(
+                    clusterStateResponse -> listener.onResponse(new SecurePluginResponse(clusterStateResponse.toString())),
                     listener::onFailure
                 )
             );
--- a/sample-resource-plugin/src/main/resources/plugin-additional-permissions.yml
+++ b/sample-resource-plugin/src/main/resources/plugin-additional-permissions.yml
@@ -1,5 +1,6 @@
 cluster_permissions:
   - "cluster:monitor/health"
+  - "cluster:monitor/state"
 index_permissions:
   - index_patterns:
       - "test-index*"
--- a/src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
@@ -49,6 +49,8 @@ import org.opensearch.security.support.WildcardMatcher;
 import org.opensearch.security.user.User;
 import org.opensearch.tasks.Task;
 
+import static org.opensearch.security.privileges.PrivilegesEvaluator.isClusterPerm;
+
 /**
  * This class performs authorization on requests targeting system indices
  * NOTE:
@@ -302,9 +304,8 @@ public class SystemIndexAccessEvaluator {
                 }
         }
 
-        // cluster actions will return true for requestedResolved.isLocalAll()
         // the following section should only be run for index actions
-        if (this.isSystemIndexEnabled && user.isPluginUser() && !requestedResolved.isLocalAll()) {
+        if (this.isSystemIndexEnabled && user.isPluginUser() && !isClusterPerm(action)) {
             Set<String> matchingPluginIndices = SystemIndexRegistry.matchesPluginSystemIndexPattern(
                 user.getName().replace("plugin:", ""),
                 requestedResolved.getAllIndices()


src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -0,0 +1,309 @@
...
+import static org.opensearch.security.dlic.rest.api.Responses.response;
+import static org.opensearch.security.dlic.rest.support.Utils.addRoutesPrefix;
+import static org.opensearch.security.spi.resources.ResourceAccessLevels.PLACE_HOLDER;
+
+/**
+ * This REST api allows REST admins to migrate resources sharing information from plugin indices to the resource-sharing indices managed by security plugin.
+ * @apiNote Only super-admins or REST admins can invoke this api.
+ * We skip migration of records with no valid creator(user).
+ * Defines:
+ *  - POST `/resources/migrate`
```
cwperks: Can you update this documentation to include the route prefix? i.e. `_plugins/_security/api/resources/migrate`

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -0,0 +1,309 @@
...
+        throws IOException {
+        JsonNode body = Utils.toJsonNode(request.content().utf8ToString());
+
+        String sourceIndex = body.get("source_index").asText();
+        String userNamePath = body.get("user_name_path").asText();
+        String backendRolesPath = body.get("backend_roles_path").asText();
+        String defaultAccessLevel = body.has("default_access_level") ? body.get("default_access_level").asText() : PLACE_HOLDER;
+
+        SearchRequest sr = new SearchRequest(sourceIndex);
+        sr.source(new SearchSourceBuilder().query(QueryBuilders.matchAllQuery()).size(10_000)); // TODO verify
```
cwperks: why not use scroll to ensure all records are traversed?

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -0,0 +1,309 @@
...
+    protected CType<?> getConfigType() {
+        return null;
+    }
+
+    private void migrateApiRequestHandlers(RequestHandler.RequestHandlersBuilder b) {
+        b.allMethodsNotImplemented().override(POST, this::handleMigrate);
+    }
+
+    private void handleMigrate(RestChannel channel, RestRequest request, Client client) throws IOException {
+        endpointValidator.createRequestContentValidator()
```
cwperks, line 119: Can we extract the params from the request here and pass them around when needed? Passing a Triple into `createNewSharingRecords` is hard to read...I would prefer that we define what each parameter in the triple is through the method signature.

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -0,0 +1,309 @@
...
+                    @Override
+                    public Set<String> mandatoryKeys() {
+                        return ImmutableSet.of("source_index", "user_name_path", "backend_roles_path");
+                    }
+
+                    @Override
+                    public Map<String, RequestContentValidator.DataType> allowedKeys() {
+                        return ImmutableMap.<String, RequestContentValidator.DataType>builder()
+                            .put("source_index", RequestContentValidator.DataType.STRING) // name of the resource plugin index
+                            .put("user_name_path", RequestContentValidator.DataType.STRING) // path to resource creator's name
```
cwperks: nit: prefer to use `username` as one word. i.e. `username_path`

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -0,0 +1,309 @@
...
+                            .put("default_access_level", RequestContentValidator.DataType.STRING) // default access level for the new
+                                                                                                  // structure
+                            .build();
+                    }
+                });
+            }
+        };
+    }
+
+    static class SourceDoc {
```
cwperks, line 329: nit: `SourceDoc` is not very descriptive for this class. maybe `LegacySharingInfo` or `LegacyExtractedSharingInfo`?

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
DarshitChanpura: I created a triple to avoid replication of source-index and default-access-level (via SourceDoc) as those values will stay same for all docs.

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
DarshitChanpura: SourceDoc should be fine. i named it as SourceDoc since the context is already set by outer class which is to migrate sharing info and sourceDoc represents values from source to be migrated. But if you feel strongly I'll rename it.

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
DarshitChanpura: done

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
DarshitChanpura: comment on line 132 shows exactly what is contained in the triple.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/SampleResourcePluginMigrationApiTests.java
```diff
@@ -0,0 +1,234 @@
...
+                RESOURCE_SHARING_MIGRATION_ENDPOINT,
+                migrationPayload_valid_withSpecifiedAccessLevel()
+            );
+            migrateResponse.assertStatusCode(HttpStatus.SC_OK);
+            assertThat(migrateResponse.bodyAsMap().get("message"), equalTo("Migration complete. migrated 1; skippedNoUser 1; failed 0"));
+
+            TestRestClient.HttpResponse sharingResponse = client.get(RESOURCE_SHARING_INDEX + "/_search");
+            sharingResponse.assertStatusCode(HttpStatus.SC_OK);
+            assertThat(sharingResponse.bodyAsJsonNode().get("hits").get("hits").size(), equalTo(1)); // 1 of 2 entries was skipped
+            assertThat(sharingResponse.getBody().contains("read_only"), is(true)); // with custom access-level
```
cwperks: Can we strengthen this assertion to also assert the expected username and any other existing sharing info was copied correctly?

sample-resource-plugin/src/main/java/org/opensearch/sample/SampleResource.java
```diff
@@ -35,6 +36,7 @@ public class SampleResource implements NamedWriteable, ToXContentObject {
     private String name;
     private String description;
     private Map<String, String> attributes;
+    private User user;
```
cwperks, line 40: I know this is added to test migrate API out specifically, but can we leave a comment the any sharable resources should not have a separate field for User in case plugin devs try to look at the code from this sample plugin?

If a plugin dev needs a way to get info about the current owner of a resource then we should create a separate mechanism to obtain that info.

cwperks (dismissed):
Left a few minor comments, but otherwise LGTM

sample-resource-plugin/src/main/java/org/opensearch/sample/SampleResource.java
DarshitChanpura: done.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/SampleResourcePluginMigrationApiTests.java
DarshitChanpura: done.

cwperks (dismissed):

DarshitChanpura:
@derek-ho mind reviewing this one?

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -0,0 +1,334 @@
...
+     */
+    private ValidationResult<String> createNewSharingRecords(Triple<String, String, List<SourceDoc>> sourceInfo) throws IOException {
+        AtomicInteger migratedCount = new AtomicInteger();
+        AtomicInteger skippedNoUser = new AtomicInteger();
+        AtomicInteger failureCount = new AtomicInteger();
+
+        List<SourceDoc> docs = sourceInfo.getRight();
+        int total = docs.size();
+
+        CountDownLatch latch = new CountDownLatch(total);
```
willyborankin: Why is `CountDownLatch` used in case?

DarshitChanpura: For tracking resource migration stats, which will be returned in response.

cwperks (dismissed):

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/SampleResourcePluginMigrationApiTests.java
```diff
@@ -0,0 +1,271 @@
...
+import static org.opensearch.sample.utils.Constants.RESOURCE_INDEX_NAME;
+import static org.opensearch.security.resources.ResourceSharingIndexHandler.getSharingIndex;
+import static org.opensearch.security.spi.resources.FeatureConfigConstants.OPENSEARCH_RESOURCE_SHARING_ENABLED;
+import static org.opensearch.security.support.ConfigConstants.SECURITY_SYSTEM_INDICES_ENABLED_KEY;
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+import static org.opensearch.test.framework.TestSecurityConfig.User.USER_ADMIN;
+
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class SampleResourcePluginMigrationApiTests {
```
nibix, line 55: For me, it feels always a bit confusing to find tests for the plugin inside the sample plugin module. Does this test really depend on the sample plugin?

src/integrationTest/java/org/opensearch/test/framework/TestSecurityConfig.java
```diff
@@ -449,9 +449,9 @@ public int hashCode() {
 
     public static final class User implements UserCredentialsHolder, ToXContentObject {
 
-        public final static TestSecurityConfig.User USER_ADMIN = new TestSecurityConfig.User("admin").roles(
+        public final static TestSecurityConfig.User USER_ADMIN = new User("admin").roles(
             new Role("allaccess").indexPermissions("*").on("*").clusterPermissions("*")
-        );
+        ).backendRoles("admin");
```
nibix: This is a global constant that can be referenced by all tests that provides a reusable admin user. As the user has a predefined role which gives full privileges on all indices and cluster permissions, the additional backend role should be not necessary. Also, adding a backend role here makes the constant actually less reusable because backend roles are always subject to role mapping. Even though it is named "admin", you'll never know wether the actually exists a role to which it is mapped or not.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -165,22 +172,23 @@ public ResourceSharing indexResourceSharing(String resourceId, String resourceIn
...
-                (failResponse) -> {
-                    LOGGER.error(failResponse.getMessage());
+            ActionListener<IndexResponse> irListener = ActionListener.wrap(idxResponse -> {
+                LOGGER.info("Successfully created {} entry for resource {} in index {}.", resourceSharingIndex, resourceId, resourceIndex);
+                listener.onResponse(entry);
+            }, (e) -> {
+                if (ExceptionsHelper.unwrapCause(e) instanceof VersionConflictEngineException) {
+                    // already exists → treat as a skip if you like
+                    LOGGER.warn("Entry for [{}] already exists in [{}], skipping", resourceId, resourceSharingIndex);
+                    listener.onResponse(entry);
```
nibix: Is a duplicate doc a case which we are expecting in regular operations?

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -0,0 +1,334 @@
...
+ *      {
+ *          source_index: "abc",                                    // name of plugin index
+ *          username_path: "/path/to/username/node",               // path to user-name in resource document in the plugin index
+ *          backend_roles_path: "/path/to/user_backend-roles/node"  // path to backend-roles in resource document in the plugin index
+ *          default_access_level: "<some-default-access-level>"     // default value that should replace the otherwise ResourceAccessLevels.PLACE_HOLDER assigned to the new ResourceSharing object
+ *      }
+ *   - Response:
+ *      200 OK Migration Complete. migrate X, skippedNoUser Y, failed Z    // migrate -> successful migration count, skippedNoUser -> records with no creator info, failed -> records that failed to migrate
+ */
+public class MigrateResourceSharingInfoApiAction extends AbstractApiAction {
```
nibix, line 85: I am wonderign whether it really makes sense to extend `AbstractApiAction` here. Originally this super class was meant for modifcation of config records, as the `getConfigType()` method and a few other methods demonstrate.

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
nibix: But we are blocking on it in line 262, right? So, this is for synchronizing on completion?

I'd recommend to rename this to `pendingResources` or something similar in this case.

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -0,0 +1,334 @@
...
+
+        CountDownLatch latch = new CountDownLatch(total);
+
+        for (SourceDoc doc : docs) {
+            // 1) get resource ID
+            String resourceId = doc.resourceId;
+            if (resourceId == null) {
+                failureCount.getAndIncrement();
+                latch.countDown();
+                continue;
```
nibix: When does this happen?

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -0,0 +1,334 @@
...
+            }
+
+            // 2) skip if no username node
+            String username = doc.username;
+            if (username == null || username.isEmpty()) {
+                LOGGER.warn("Record without associated user, skipping entirely: {}", doc.resourceId);
```
nibix: I am wondering whether these warnings should be not only logged but collected and returned in the REST response to this action.

Let's think about the workflow: An admin user is using this action; in the results, they will see a numeric value that indices that not all resources were successfully migrated. What shall they do then?

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
DarshitChanpura: Correct. I'll rename it to migrationStats.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/SampleResourcePluginMigrationApiTests.java
DarshitChanpura: Sample plugin has been setup to test real world plugin use-cases. In this case, I can add migrate Api tests separately, but since sample plugin is available I can simply add the tests here since the only applicable use-case for the migrate API is for plugins, like sample-plugin, moving to this new authorization setup.

src/integrationTest/java/org/opensearch/test/framework/TestSecurityConfig.java
DarshitChanpura: This change is required to test case where resources are default shared with user's backend-roles. AD is one example that does this. But I have modified the test to create a local user, instead of using USER_ADMIN.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: It is added for robustness.

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
DarshitChanpura: It was extended for re-usability, i.e. endpoint validation + route-handlers setup + admin-api privilege restriction , instead of having to rewrite it. Also, the AbstractApiAction class doesn't mention about usage reserved for configs.

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
DarshitChanpura: It is added for robustness. We do not expect this to happen.

src/main/java/org/opensearch/security/resources/migrate/MigrateResourceSharingInfoApiAction.java
DarshitChanpura: They can then look-up from logs. But I have modified response to send a list of skipped resource Ids and change message level to debug.

DarshitChanpura:
> Regarding the JSON path approach, I am wondering how this action shall be used. Will end users call it - and they can source the request parameters from plugin docs?

> Or will this API be called from another piece of code, like a Dashboard plugin, a migration script, etc?

Great questions. At present a rest-admin will be to invoke the API to migrate resource-sharing info for each index. We do not intend to introduce a migration script as the action simply moves the sharing info from plugin indices to security managed indices. Old info will still be present until eventually all plugins migrate to this new model, and then we can write a script to scrub existing user info from plugin indices.

Issue #5367, opened 2025-05-27
[FEATURE] Index pattern resolution improvements


nibix:
# Overview

Goal of this issue is to implement broad improvements regarding the index pattern resolution in the security plugin.

Index pattern resolution is a vital component of the security plugin; it extracts from action requests metadata about the indices, aliases and data streams a request is going to touch. All other components of the security plugin rely on the index pattern resolution to reliably and efficiently provide this information.

However, as described in the following sections, the index pattern resolution currently operates in ways that are less than optimal. 

# Current situation

## Performance

CPU profiling on the security plugin code shows that the core index pattern resolution code is one of the most CPU intense tasks of the request processing performed by the security plugin. 

This is the code in question:

https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/resolver/IndexResolverReplacer.java

An analysis of the code shows that many things are done unnecessarily:

- The alias resolution is expensive because it is done by filtering the whole index set (which can be quite big); yet, the result of the alias resolution is used only in a single component which is often bypassed: the `PrivilegeInterceptorImpl` for implementing Dashboards multi tenancy.

- Data streams are broken apart into individual indices even if the privilege is granted on the data stream itself; as data streams can have many backing indices, this can be also expensive.

- Generally, lots of copying and filtering of potentially large data structures is going on

## Structure

The class IndexResolverReplacer must have white-box knowledge of the requests it wants to process. This can be seen here:

https://github.com/opensearch-project/security/blob/19a41b8f01d00adbeec51b35788d38e18a2f2adc/src/main/java/org/opensearch/security/resolver/IndexResolverReplacer.java#L657-L836

This is a fragile way to implement things and requires a big maintenance effort. Newly actions added to the core must/shoule be reviewed whether they need special treatment here.

## `do_not_fail_on_forbidden`

The class IndexResolverReplacer additionally implements the index replace algorithm required by the do_not_fail_on_forbidden mode. As the class does not implement it in a universal way, there are many issues with that mode as described in #3905. 

# New approach

Goal of this issue is to completely replace the `IndexResolverReplacer` implementation which:

- Leverages core interfaces to efficiently get the "true" resolved indices from a request
- Avoids unnecessary index resolutions
- Can be leveraged to implement the refined index authorization as described in #3905 

# Strategy

- [ ] Conceive core interfaces to be implemented by action requests to provide true index information
- [ ] Create prototype implementation (might also leverage caching of resolution result in request object)
[...]

PR #5399, opened 2025-06-16
Introduced explicit index resolution API

nibix:
### Description

**This is a branch merge request to illustrate the changes needed for the improved index resolution #5367. It is not yet finished yet**

* Category
  * Enhancement
  * The current index resolution mechanism employed by the security plugin (with the class `IndexResolverReplacer` in the center) has several downsides: It is slow, it is fragile, it is messy. The goal of the changes in this and the related PR is to fix this.

### Issues Resolved

- #5367

### Check List


nibix:
# Preliminary benchmark results

We now have some preliminary benchmark results. 

## Disclaimer

The usual benchmarking disclaimer: Of course, these figures can only represent very specific scenarios. Other scenarios can look quite different, as there are so many variables which can be different. Yet, especially the "slow parts" of the benchmark results can give one an impression where real performance issues are.

## Test setup

- We used a OpenSearch main snapshot for all tests
- All plugins were removed, except the security plugin
- Audit logging was disabled
- Authentication: Only the internal backend auth was active. All other authentication types were disabled.

### K8s resources per node

- Java Heap: 32 GB
- RAM: 64 GB
- CPU: 8 cores
- Kubernetes version: v1.32.2

## Tested dimensions

### Users

We compared four different users:

- A simple user with normal privileges
- User authenticated by admin certificate

### Indices on cluster

We executed the benchmarks with various numbers of indices on the cluster:

- 30 indices
- 100 indices
- 300 indices
- 1000 indices
- 3000 indices

### Operations

We tested four different operations:

- Bulk indexing with 10 items per request
- Bulk indexing with 100 items per request
- Searching on single index
- Searching on 20 indices (using index pattern)

## Test results

A commented summary of the results follows in the upcoming sections.

The raw test results can be also reviewed at  https://docs.google.com/spreadsheets/d/17Vyolu2UZ0ZjDNySyb0uyPeQIPMw-LrHQ2bvd_dNqP0/edit?usp=sharing

### Indexing throughput 

#### Bulk size 10

Bulk indexing size 10 results

In this and the following charts, the dashed lines represent OpenSearch with the standard security plugin. The solid lines represent OpenSearch with the security plugin with the optimized index resolution code.

The blue line represents requests authenticated by the super admin certificate, which by-passes most of the security plugin. Thus, the blue line forms a kind of "hull curve", it can be seen as a rough theoretical maximum from a security plugin point of view.

We see a consistent improvement of the indexing throughput (in docs/s) by 5 to 12%.  Interestingly, a correlation to the number of indices on the cluster cannot be really observed (except for maybe the 3000 index case). This is different from my expectations. I have checked this with micro benchmarks; indeed, we get a runtime improvement in the security filter, which is however less than 0.5 ms and thus not really visible in these benchmarks.

We still have quite a gap to the admin certificate performance; this indicates that we could do more research on the performance impact of the security plugin.

#### Bulk size 100

Bulk indexing size 100 results

Bulk requests with more items can be processed more efficiently. Thus, as we are looking at a docs/s throughput, the improvements are a bit smaller in this case. The improvement range is between 2% and 8%.

### Search throughput

#### Single index

Search on single index results

In this benchmark, we compare the operation throughput for search requests performed on a single index.

Plain and simple: For this case, we do not see an improvement. This is probably because search operations only rely on a single operation going through the security filter. Further messages are exchanged around the cluster in order to communicate with the shards; however, these are not ActionRequests. Thus, these are not affected by the security plugin. In addition, an index expression with a single index does not require any matching against the existing indices. This makes the whole process quite efficient up-front.

#### 20 indices

Search on 20 indices results

This benchmark tests searches operations that search through 20 indices at once (specified by an index pattern).

Here we do see an improvement of the throughput by about 2 to 4%.

@cwperks @kkhatua

nibix:
superseded by https://github.com/opensearch-project/security/pull/5827

Issue #4915, opened 2024-11-18
[BUG] Current version of OpenSAML is incompatible with the Bouncy Castle FIPS library


dancristiancecoi:
**What is the bug?**
Current version of OpenSAML (4.3.2) does not work in FIPS mode as it has a hard dependency (`org.bouncycastle.jce.ECNamedCurveTable`) on the non-FIPS distribution of Bouncy Castle

**How can one reproduce the bug?**
1. Replace BC with BC-FIPS 
2. Run OpenSAML related tests
3. You will see a error due to missing the `ECNamedCurveTable` dependency

**Potential solutions**
1. Downgrade OpenSAML to 3.x
2. Downgrade OpenSAML to 4.0 (Tests seem to pass under that version, however some untested functionality might still be broken)
3. Replace OpenSAML with Keycloak as [it supports running in FIPS 140-2 compliant mode 
](https://www.keycloak.org/server/fips)

**Relevant links**
https://github.com/opensearch-project/security/issues/3420
https://github.com/elastic/elasticsearch/issues/71983
https://shibboleth.atlassian.net/wiki/spaces/DEV/pages/1159627167/FIPS
https://shibboleth.net/pipermail/dev/2023-August/011111.html
https://www.keycloak.org/server/fips

cwperks:
[Triage] Thank you for filing an issue for this @dancristiancecoi. Should I assign this issue to you?

dancristiancecoi:
@cwperks I can give it a try, thanks!

Would also be great to get some feedback from the community on what potential solution would be the best way forward.

shshankvikrram:
@dancristiancecoi 
Removing service  'META-INF/services/org.opensaml.security.crypto.ec.NamedCurve' from opensaml will ensure EC classes are not loaded eagerly and you can remove bc non fips dependencies. 

Elastic search have implemented this already, please refer to this : https://github.com/elastic/elasticsearch/pull/98199/files

dancristiancecoi:
Excellent find @shshankvikrram, thanks!

Am I correct saying that this fixes the error on start-up for Elasticsearch but the functionality that relies on EC will still be broken?

In OpenSearch's case the error does not happen on start-up but when the OpenSAML code gets called. If we used this solution and we do not rely on any EC related functionality then we might be okay. 

Are we confident that we don't rely on any of that functionality and that the test coverage in that area is sufficient to prove this?

shshankvikrram:
[...]

PR #5404, opened 2025-06-16
Moved OpenSAML jars to a Shadow Jar configuration to facilitate its use in FIPS enabled environments

dancristiancecoi:
### Description
Moved OpenSAML jars to a Shadow Jar configuration to facilitate its use in FIPS enabled environments. 

OpenSAML past version 4.0 can't run in FIPS environments as it relies on classes that are only present in the standard distribution of BouncyCastle

Changes: 
- NamedCurve class was excluded from the shadow jar so OpenSAML can run in environments that do not have the standard BouncyCastle jars
- Also excluded various downstream dependencies from the shadow jar to avoid jarHell failures

### Issues Resolved
https://github.com/opensearch-project/security/issues/4915

### Relevant Issues
https://github.com/opensearch-project/security/issues/3420

### Testing
All existing OpenSAML unit tests are passing. 

Also manually tested its behaviour under a FIPS environment by replacing the standard BC jars with BC-FIPS jars and BouncyCastleProvider to BouncyCastleFIPSProvider in the code 

### Check List


src/main/java/org/opensearch/security/auth/http/saml/HTTPSamlAuthenticator.java
```diff
@@ -316,6 +317,10 @@ static void ensureOpenSamlInitialization() {
     private static void initializeOpenSAMLConfiguration() throws InitializationException {
         log.info("Initializing OpenSAML using the Java Services API");
 
+        System.setProperty(
```
willyborankin, line 320: Is it possible to avoid using system configuration via the `-D` option?

dancristiancecoi: I think its possible by setting  -Dopensaml.config.ecdh.defaultKDF=PBKDF2 but I have not had success with it (and the can't find this option anywhere in the docs)

I've only tested it by adding it to the test run configuration so this might be a reason why it's not working properly.

willyborankin: OK, then I don't see any problem.

willyborankin (approved):
LGTM

libs/opensaml/build.gradle
```diff
@@ -0,0 +1,63 @@
...
+        exclude(group: 'org.apache.velocity', module: 'velocity')
+    }
+    implementation "org.opensaml:opensaml-messaging-api:${open_saml_version}"
+    implementation "org.opensaml:opensaml-storage-api:${open_saml_version}"
+
+    runtimeOnly "org.opensaml:opensaml-profile-api:${open_saml_version}"
+    runtimeOnly "org.opensaml:opensaml-soap-api:${open_saml_version}"
+    runtimeOnly "org.opensaml:opensaml-soap-impl:${open_saml_version}"
+
+    runtimeOnly "org.apache.httpcomponents.client5:httpclient5:5.4.4"
```
cwperks: Let's see if there's a way to refer to these versions from core's version catalog: https://github.com/opensearch-project/OpenSearch/blob/main/gradle/libs.versions.toml#L45-L46

src/main/java/org/opensearch/security/auth/http/saml/HTTPSamlAuthenticator.java
cwperks: Is this only for FIPS-Approved mode? Should this throw an error if this system prop is previously set?

cwperks:
Thank you @dancristiancecoi . The changes mostly LGTM, but would like to see if its possible to refer to the gradle catalog for those httpclient5 deps in the new library.

libs/opensaml/build.gradle
willyborankin: @cwperks Good point. I didn’t notice it during the first round of review.

cwperks: iirc you can have a dep on build-tools like this: https://github.com/opensearch-project/security/blob/main/build.gradle#L57-L59

and then use the `api "org.module.example:example-impl:${versions.example}"` syntax

src/main/java/org/opensearch/security/auth/http/saml/HTTPSamlAuthenticator.java
dancristiancecoi: Honestly I was not entirely sure on this one. I thought it will be easier to just set it to PBKDF2 going forward to avoid another configuration an admin needs to make when switching to FIPS mode.

However it is important we test this on a setup that has OpenSearch Dashboards and preexisting users to see if there are any issues with backwards compatibility. If that is the case then I can make it a configurable options & document it

libs/opensaml/build.gradle
dancristiancecoi: Good catch! Made these changes

src/main/java/org/opensearch/security/auth/http/saml/HTTPSamlAuthenticator.java
nibix: Just chiming in here briefly to add one detail:

The OpenSearch security SAML implementation is special, as it actually is based on two different SAML libraries:

- `com.onelogin.saml2`
- OpenSAML

AFAIK, OpenSAML is *only*  used to resolve and to retrieve SAML metadata.xml files from external endpoints. The actual construction and verification of SAML requests and responses is done by the onelogin library in https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/http/saml/AuthTokenProcessorHandler.java

That means that is likely that any cryptographic code in OpenSAML is actually not in use in OpenSearch (even if OpenSAML requires it to be initialized).

Maybe that's without consequences to this PR, I just wanted to share this detail.

cwperks (approved):

src/main/java/org/opensearch/security/auth/http/saml/HTTPSamlAuthenticator.java
dancristiancecoi: Thanks a lot for clarifying @nibix! Does this mean that we should also check that the crypto used in onelogin is FIPS compliant?

nibix: If the goal of this PR is FIPS compliance, then possibly yes. For me the goal of this PR wasn't exactly clear, first my impression was more something like "make it work in a FIPS constrained environment".

PR #5408, opened 2025-06-17
[Resource Sharing] Adds a Resource Access Evaluator for standalone Resource access authorization

DarshitChanpura:
### Description
This PR adds a new privilege evaluator for evaluating access to a resource. #5281 introduced a way for plugin offload sharing and access evaluation to security plugin but that was done by requiring plugins to call verifyAccess method on their end. This leaves room for error. This new evaluator will filter all resource access requests through SecurityFilter class without requiring plugins to explicitly call verifyAccess method. It also adds support for access-levels instead of just the default one declared in the previous PR.

### Testing
- Automated Tests

### Check List

### DIFF
--- a/sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/rest/revoke/RevokeResourceAccessRequest.java
+++ b/sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/rest/revoke/RevokeResourceAccessRequest.java
@@ -14,14 +14,17 @@ import java.util.Set;
 
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.ActionRequestValidationException;
+import org.opensearch.action.DocRequest;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.security.spi.resources.sharing.Recipient;
 
+import static org.opensearch.sample.utils.Constants.RESOURCE_INDEX_NAME;
+
 /**
  * Request object for revoking access to a sample resource
  */
-public class RevokeResourceAccessRequest extends ActionRequest {
+public class RevokeResourceAccessRequest extends ActionRequest implements DocRequest {
 
     String resourceId;
     Map<Recipient, Set<String>> entitiesToRevoke;
@@ -57,5 +60,15 @@ public class RevokeResourceAccessRequest extends ActionRequest {
 
     public Map<Recipient, Set<String>> getEntitiesToRevoke() {
         return entitiesToRevoke;
+    }
+
+    @Override
+    public String index() {
+        return RESOURCE_INDEX_NAME;
+    }
+
+    @Override
+    public String id() {
+        return resourceId;
     }
 }
--- a/sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/rest/revoke/RevokeResourceAccessRequest.java
+++ b/sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/rest/revoke/RevokeResourceAccessRequest.java
@@ -9,15 +9,13 @@
 package org.opensearch.sample.resource.actions.rest.revoke;
 
 import java.io.IOException;
-import java.util.Map;
-import java.util.Set;
 
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.ActionRequestValidationException;
 import org.opensearch.action.DocRequest;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
-import org.opensearch.security.spi.resources.sharing.Recipient;
+import org.opensearch.security.spi.resources.sharing.ShareWith;
 
 import static org.opensearch.sample.utils.Constants.RESOURCE_INDEX_NAME;
 
@@ -27,26 +25,22 @@ import static org.opensearch.sample.utils.Constants.RESOURCE_INDEX_NAME;
 public class RevokeResourceAccessRequest extends ActionRequest implements DocRequest {
 
     String resourceId;
-    Map<Recipient, Set<String>> entitiesToRevoke;
+    ShareWith revokeAccess;
 
-    public RevokeResourceAccessRequest(String resourceId, Map<Recipient, Set<String>> entitiesToRevoke) {
+    public RevokeResourceAccessRequest(String resourceId, ShareWith entitiesToRevoke) {
         this.resourceId = resourceId;
-        this.entitiesToRevoke = entitiesToRevoke;
+        this.revokeAccess = entitiesToRevoke;
     }
 
     public RevokeResourceAccessRequest(StreamInput in) throws IOException {
         resourceId = in.readString();
-        entitiesToRevoke = in.readMap(key -> key.readEnum(Recipient.class), input -> input.readSet(StreamInput::readString));
+        revokeAccess = new ShareWith(in);
     }
 
     @Override
     public void writeTo(final StreamOutput out) throws IOException {
         out.writeString(resourceId);
-        out.writeMap(
-            entitiesToRevoke,
-            StreamOutput::writeEnum,
-            (streamOutput, strings) -> streamOutput.writeCollection(strings, StreamOutput::writeString)
-        );
+        revokeAccess.writeTo(out);
     }
 
     @Override
@@ -58,8 +52,8 @@ public class RevokeResourceAccessRequest extends ActionRequest implements DocReq
         return resourceId;
     }
 
-    public Map<Recipient, Set<String>> getEntitiesToRevoke() {
-        return entitiesToRevoke;
+    public ShareWith getEntitiesToRevoke() {
+        return revokeAccess;
     }
 
     @Override
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -1145,6 +1145,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
 
         final CompatConfig compatConfig = new CompatConfig(environment, transportPassiveAuthSetting);
 
+        rsIndexHandler = new ResourceSharingIndexHandler(localClient, threadPool);
         evaluator = new PrivilegesEvaluator(
             clusterService,
             clusterService::state,
@@ -1255,7 +1256,6 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
             FeatureConfigConstants.OPENSEARCH_RESOURCE_SHARING_ENABLED,
             FeatureConfigConstants.OPENSEARCH_RESOURCE_SHARING_ENABLED_DEFAULT
         )) {
-            rsIndexHandler = new ResourceSharingIndexHandler(localClient, threadPool);
 
             ResourceAccessHandler resourceAccessHandler = new ResourceAccessHandler(threadPool, rsIndexHandler, adminDns);
 
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -1156,7 +1156,9 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
             settings,
             privilegesInterceptor,
             cih,
-            irr
+            irr,
+            resourcePluginInfo.getResourceIndices(),
+            rsIndexHandler
         );
 
         dlsFlsBaseContext = new DlsFlsBaseContext(evaluator, threadPool.getThreadContext(), adminDns);
--- a/src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
@@ -1,0 +1,127 @@
+package org.opensearch.security.privileges;
+
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+
+import org.opensearch.action.ActionRequest;
+import org.opensearch.action.DocRequest;
+import org.opensearch.action.DocWriteRequest;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.security.auth.UserSubjectImpl;
+import org.opensearch.security.privileges.actionlevel.RoleBasedActionPrivileges;
+import org.opensearch.security.resources.ResourceSharingIndexHandler;
+import org.opensearch.security.spi.resources.sharing.Recipient;
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.security.user.User;
+import org.opensearch.threadpool.ThreadPool;
+
+public class ResourceAccessEvaluator {
+    private final Set<String> resourceIndices;
+    private final ThreadContext threadContext;
+    private final ResourceSharingIndexHandler resourceSharingIndexHandler;
+
+    public ResourceAccessEvaluator(
+        Set<String> resourceIndices,
+        ThreadPool threadPool,
+        ResourceSharingIndexHandler resourceSharingIndexHandler
+    ) {
+        this.resourceIndices = resourceIndices;
+        this.threadContext = threadPool.getThreadContext();
+        this.resourceSharingIndexHandler = resourceSharingIndexHandler;
+    }
+
+    public PrivilegesEvaluatorResponse evaluate(
+        final ActionRequest request,
+        final String action,
+        final PrivilegesEvaluationContext context,
+        final PrivilegesEvaluatorResponse presponse
+    ) throws IOException {
+
+        if (!(request instanceof DocRequest)) {
+            return presponse;
+        }
+
+        final UserSubjectImpl userSubject = (UserSubjectImpl) this.threadContext.getPersistent(
+            ConfigConstants.OPENDISTRO_SECURITY_AUTHENTICATED_USER
+        );
+        final User user = (userSubject == null) ? null : userSubject.getUser();
+
+        if (user == null) {
+            presponse.allowed = false;
+            return presponse.markComplete();
+        }
+
+        // If user was super-admin, the request would have already been granted. So no need to check whether user is admin
+
+        if (request instanceof DocWriteRequest<?>) {
+            // check write permissions <- maybe punt it to the regular evaluator since it requires write permissions to the index
+        }
+
+        DocRequest req = (DocRequest) request;
+
+        // if requested index is not a resource sharing index, move on to the next evaluator
+        if (!resourceIndices.contains(req.index())) {
+            return presponse;
+        }
+
+        // Fetch the ResourceSharing document
+
+        CountDownLatch latch = new CountDownLatch(1);
+
+        Set<String> userRoles = new HashSet<>(user.getSecurityRoles());
+        Set<String> userBackendRoles = new HashSet<>(user.getRoles());
+
+        this.resourceSharingIndexHandler.fetchSharingInfo(req.index(), req.id(), ActionListener.wrap(document -> {
+            if (document == null) {
+                presponse.allowed = false; // TODO should we move this to next evaluator if no document is present, probs not since this
+                                           // index is protected
+                latch.countDown();
+                return;
+            }
+
+            // If document is public, action is allowed
+            // If user is the owner, action is allowed
+            if (document.isSharedWithEveryone() || document.isCreatedBy(user.getName())) {
+                presponse.allowed = true;
+                latch.countDown();
+                return;
+            }
+
+            Set<String> accessLevels = new HashSet<>();
+            accessLevels.addAll(document.fetchAccessLevels(Recipient.USERS, Set.of(user.getName())));
+            accessLevels.addAll(document.fetchAccessLevels(Recipient.ROLES, userRoles));
+            accessLevels.addAll(document.fetchAccessLevels(Recipient.BACKEND_ROLES, userBackendRoles));
+
+            if (accessLevels.isEmpty()) {
+                presponse.allowed = false;
+                latch.countDown();
+                return;
+            }
+
+            // Expand access-levels and check if any match the action supplied
+            if (context.getActionPrivileges() instanceof RoleBasedActionPrivileges roleBasedActionPrivileges) {
+                Set<String> actions = roleBasedActionPrivileges.flattenedActionGroups().resolve(accessLevels);
+                presponse.allowed = actions.contains(action);
+                latch.countDown();
+            } else {
+                // we don't yet support Plugins to access resources
+                presponse.allowed = false;
+                latch.countDown();
+            }
+
+        }, e -> {
+            presponse.allowed = false;
+            latch.countDown();
+        }));
+        try {
+            latch.await();
+        } catch (InterruptedException ie) {
+            Thread.currentThread().interrupt();
+        }
+
+        return presponse.markComplete();
+    }
+}
--- a/src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
@@ -4,10 +4,10 @@ import java.io.IOException;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.DocRequest;
-import org.opensearch.action.DocWriteRequest;
 import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.security.auth.UserSubjectImpl;
@@ -15,6 +15,7 @@ import org.opensearch.security.privileges.actionlevel.RoleBasedActionPrivileges;
 import org.opensearch.security.resources.ResourceSharingIndexHandler;
 import org.opensearch.security.spi.resources.sharing.Recipient;
 import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.security.support.WildcardMatcher;
 import org.opensearch.security.user.User;
 import org.opensearch.threadpool.ThreadPool;
 
@@ -40,7 +41,10 @@ public class ResourceAccessEvaluator {
         final PrivilegesEvaluatorResponse presponse
     ) throws IOException {
 
-        if (!(request instanceof DocRequest)) {
+        // TODO: Check whether resource access should be disabled system index protection is off
+
+        // TODO need to check whether "cluster:" perms should be handled heeyah
+        if (!(request instanceof DocRequest req)) {
             return presponse;
         }
 
@@ -54,19 +58,15 @@ public class ResourceAccessEvaluator {
             return presponse.markComplete();
         }
 
-        DocRequest req = (DocRequest) request;
-
         // If user was super-admin, the request would have already been granted. So no need to check whether user is admin
 
         // Creation Request
         // TODO Check if following is the correct way to identify the create request
-        if (request instanceof DocWriteRequest<?> && req.id() == null) {
+        if (req.id() == null) {
             // check write permissions
-            // TODO verif that this can be punted to the regular evaluator since it requires write permissions to the index
+            // TODO verify that this can be punted to the regular evaluator since it requires write permissions to the index
             return presponse;
         }
-
-
 
         // if requested index is not a resource sharing index, move on to the next evaluator
         if (!resourceIndices.contains(req.index())) {
@@ -80,10 +80,11 @@ public class ResourceAccessEvaluator {
         Set<String> userRoles = new HashSet<>(user.getSecurityRoles());
         Set<String> userBackendRoles = new HashSet<>(user.getRoles());
 
+        AtomicBoolean shouldMarkAsComplete = new AtomicBoolean(false);
         this.resourceSharingIndexHandler.fetchSharingInfo(req.index(), req.id(), ActionListener.wrap(document -> {
             if (document == null) {
-                presponse.allowed = false; // TODO should we move this to next evaluator if no document is present, probs not since this
-                                           // index is protected
+                // TODO check whether we should mark response as not allowed. At present, it just returns incomplete response and hence is
+                // delegated to next evaluator
                 latch.countDown();
                 return;
             }
@@ -92,6 +93,7 @@ public class ResourceAccessEvaluator {
             // If user is the owner, action is allowed
             if (document.isSharedWithEveryone() || document.isCreatedBy(user.getName())) {
                 presponse.allowed = true;
+                shouldMarkAsComplete.set(true);
                 latch.countDown();
                 return;
             }
@@ -103,6 +105,9 @@ public class ResourceAccessEvaluator {
 
             if (accessLevels.isEmpty()) {
                 presponse.allowed = false;
+                // TODO check why following addition doesn't reflect in the final response message and find an alternative
+                presponse.getMissingPrivileges().add(action);
+                shouldMarkAsComplete.set(true);
                 latch.countDown();
                 return;
             }
@@ -110,16 +115,25 @@ public class ResourceAccessEvaluator {
             // Expand access-levels and check if any match the action supplied
             if (context.getActionPrivileges() instanceof RoleBasedActionPrivileges roleBasedActionPrivileges) {
                 Set<String> actions = roleBasedActionPrivileges.flattenedActionGroups().resolve(accessLevels);
-                presponse.allowed = actions.contains(action);
+                // a matcher to test against all patterns in `actions`
+                WildcardMatcher matcher = WildcardMatcher.from(actions, true);
+                if (matcher.test(action)) {
+                    presponse.allowed = true;
+                } else {
+                    // TODO check why following addition doesn't reflect in the final response message and find an alternative
+                    presponse.getMissingPrivileges().add(action);
+                }
                 latch.countDown();
             } else {
                 // we don't yet support Plugins to access resources
                 presponse.allowed = false;
                 latch.countDown();
             }
+            shouldMarkAsComplete.set(true);
 
         }, e -> {
             presponse.allowed = false;
+            shouldMarkAsComplete.set(true);
             latch.countDown();
         }));
         try {
@@ -128,6 +142,6 @@ public class ResourceAccessEvaluator {
             Thread.currentThread().interrupt();
         }
 
-        return presponse.markComplete();
+        return shouldMarkAsComplete.get() ? presponse.markComplete() : presponse;
     }
 }
--- a/src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
@@ -54,13 +54,19 @@ public class ResourceAccessEvaluator {
             return presponse.markComplete();
         }
 
+        DocRequest req = (DocRequest) request;
+
         // If user was super-admin, the request would have already been granted. So no need to check whether user is admin
 
-        if (request instanceof DocWriteRequest<?>) {
-            // check write permissions <- maybe punt it to the regular evaluator since it requires write permissions to the index
+        // Creation Request
+        // TODO Check if following is the correct way to identify the create request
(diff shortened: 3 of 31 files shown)


DarshitChanpura:
I will be flipping this out of draft once #5389 is merged. But feel free to review as this is ready.

CHANGELOG.md
```diff
@@ -15,27 +15,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 * Create a mechanism for plugins to explicitly declare actions they need to perform with their assigned PluginSubject ([#5341](https://github.com/opensearch-project/security/pull/5341))
 * Moves OpenSAML jars to a Shadow Jar configuration to facilitate its use in FIPS enabled environments ([#5400](https://github.com/opensearch-project/security/pull/5404))
+* [Resource Sharing] Adds a Resource Access Evaluator for standalone Resource access authorization ([#5408](https://github.com/opensearch-project/security/pull/5408))
+* Moves OpenSAML jars to a Shadow Jar configuration to facilitate its use in FIPS enabled environments ([#5400](https://github.com/opensearch-project/security/pull/5404))
```
cwperks: Looks like CHANGELOG got messed up when syncing. Can you fix that?

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
```diff
@@ -0,0 +1,212 @@
...
+        final PrivilegesEvaluatorResponse presponse
+    ) {
+
+        // If feature is disabled we skip evaluation through this evaluator
+        if (!isResourceSharingFeatureEnabled) {
+            return presponse;
+        }
+
+        // Skip evaluation if request is not a DocRequest type
+        if (!(request instanceof DocRequest req)) {
```
cwperks: Should we also add logic to skip DocWriteRequest?

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
```diff
@@ -0,0 +1,212 @@
...
+            log.debug("Request index {} is not a protected resource index", req.index());
+            return presponse;
+        }
+
+        final UserSubjectImpl userSubject = (UserSubjectImpl) this.threadContext.getPersistent(
+            ConfigConstants.OPENDISTRO_SECURITY_AUTHENTICATED_USER
+        );
+        final User user = (userSubject == null) ? null : userSubject.getUser();
+
+        if (user == null) {
```
cwperks: I don't think this condition can ever be true. We already have short-circuiting logic here: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/filter/SecurityFilter.java#L327-L364

There is some [logic](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/filter/SecurityFilter.java#L342-L345) in that block that I am not super familiar with.

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
```diff
@@ -0,0 +1,212 @@
...
+        CountDownLatch latch = new CountDownLatch(1);
+
+        Set<String> userRoles = new HashSet<>(user.getSecurityRoles());
+        Set<String> userBackendRoles = new HashSet<>(user.getRoles());
+
+        AtomicBoolean shouldMarkAsComplete = new AtomicBoolean(false);
+
+        // Fetch the ResourceSharing document and evaluate access
+        this.resourceSharingIndexHandler.fetchSharingInfo(req.index(), req.id(), ActionListener.wrap(document -> {
+            if (document == null) {
```
cwperks: error condition? If resource sharing is enabled and we don't even know the resource owner than that is a problem.

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
```diff
@@ -0,0 +1,212 @@
...
+                // TODO check whether we should mark response as not allowed. At present, it just returns incomplete response and hence is
+                // delegated to next evaluator
+                log.debug("No resource sharing record found for resource {} and index {}, skipping evaluation.", req.id(), req.index());
+                latch.countDown();
+                return;
+            }
+
+            // If the document is public, action is allowed
+            // If user is the owner, action is allowed
+            if (document.isSharedWithEveryone() || document.isCreatedBy(user.getName())) {
```
cwperks: Does `isSharedWithEveryone` means that everyone has full access or that its visible to everyone?

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
```diff
@@ -0,0 +1,212 @@
...
+            if (accessLevels.isEmpty()) {
+                presponse.allowed = false;
+                log.debug("Resource {} is not shared with user {}", req.id(), user.getName());
+                shouldMarkAsComplete.set(true);
+                latch.countDown();
+                return;
+            }
+
+            // Expand access-levels and check if any actions match the action supplied
+            if (context.getActionPrivileges() instanceof RoleBasedActionPrivileges roleBasedActionPrivileges) {
```
cwperks: Can we put this condition above and skip this evaluator if its not an instance of RoleBasedActionPrivileges?

nibix: Is this only about accessing FlattenedActionGroups? 

If so, how about keeping a reference to it in `PrivilegesEvaluator` and pass a supplier to this class?

The reference could be updated here:

https://github.com/opensearch-project/security/blob/68c0d94ecb856f52b5e77f3a0b898ee88eed19f8/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L240-L266

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
```diff
@@ -0,0 +1,212 @@
...
+                Set<String> actions = roleBasedActionPrivileges.flattenedActionGroups().resolve(accessLevels);
+                // a matcher to test against all patterns in `actions`
+                WildcardMatcher matcher = WildcardMatcher.from(actions, true);
+                if (matcher.test(action)) {
+                    presponse.allowed = true;
+                    log.debug("Resource {} is shared with user {}, granting access.", req.id(), user.getName());
+                } else {
+                    // TODO check why following addition doesn't reflect in the final response message and find an alternative
+                    presponse.getMissingPrivileges().add(action);
```
nibix: I'd believe that's because the collections managed by PrivilegesEvaluatorResponse are immutable. It's a bit surprising that no exception is thrown.

Generally, I would recommend to avoid the pattern of modifying `PrivilegesEvaluatorResponse` inline. An alternative would be:

```
return PrivilegesEvaluatorResponse.insufficient(action)
```

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
```diff
@@ -0,0 +1,212 @@
...
+            shouldMarkAsComplete.set(true);
+
+        }, e -> {
+            presponse.allowed = false;
+            log.debug("Something went wrong while evaluating resource {}. Marking request as unauthorized.", req.id());
+            shouldMarkAsComplete.set(true);
+            latch.countDown();
+        }));
+        try {
+            latch.await();
```
nibix: Unfortunately, blocking on index operations here bears the risk of spurious deadlocks due to thread pool exhaustion ...

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
DarshitChanpura: DocWriteRequest extends DocRequest class, so it will be skipped if I understand correctly

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
DarshitChanpura: I added it as I'm also not super familiar with that logic, but it seems like evaluator is not called if user doesn't exist.

DarshitChanpura: I'll remove it in next commit.

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
DarshitChanpura: We will know the resource owner if a record is created, this block handles the situation where for some reason a resource-sharing record doesn't yet exist. One possible situation is when this resource evaluator is triggered on an updated cluster before resource-sharing records are migrated.

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
DarshitChanpura: Good catch. I thought i had pushed an update to this condition.

isSharedWithEveryone means that this record is shared with everyone at that particular level.

This flag doesn't by default grant full-access

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
DarshitChanpura: The only way we make this non-blocking is to keep the resource-sharing information in memory similar to security plugin's core authz model. Let me know if you are aware of any other way to approach this.

We can approach this in one of two ways:
1. Create a separate threadpool for resource evaluation (similar to a discussion around actionPrivileges threadPool for OPE).
2. Maintain a cache of topK  frequently accessed resources per resource-index. K = 1000?

Both of these are optimization issues and can be tackled in a separate PR.

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
DarshitChanpura: since its inside the actionlistener wrapper i couldn't use insufficient() method as i cannot directly return response.

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
DarshitChanpura: I have moved the condition further up. IMO keeping it here should be fine to avoid unnecessary complexity.

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
nibix: But that's a risk of dead locks; i am not sure that this can be classified as optimization issue. One solution would be to make PrivilegeEvaluator async capable. But that's some more refactoring :-/

cwperks: Is there a way to specify a threadpool executor for a RestAction?

cwperks: oh I see, it would be a netty thread for HTTP Request handling, but at the transport level we can specify executor for a TransportAction. For example: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/explain/TransportExplainAction.java#L203-L209

nibix: I believe if we only get requests via REST, this is not an issue, as we do not have a cyclic dependency. However, if we also get resource requests from other transport actions (possibly from plugins), we run the risk of exhausting the thread pool.

cwperks: right, do you think its possible to only call ResourceAccessEvaluator when applicable and return from the apply fn? Within the ResourceAccessEvaluation we could do an async call to get sharing info, calculate whether the caller is permitted and then accordingly process through the action filter chain (if authorized), else send back a forbidden message on the RestChannel. 

pseudo-code

```
if (isRequestForSharableResource) { 
     resourceAccessEvaluator.evaluate(...) // async call to get sharing info from within here
     return; // exit from the ActionFilter, but don't call chain.proceed
}
```

Within `resourceAccessEvaluator.evaluate()` it would then asynchronously get the sharing info, determine whether the caller has access and proceed accordingly.

nibix: Yes, I would believe that deferring `ActionFilterChain.proceed()` to an `ActionListener.onResponse()` would be the right way to do this.

DarshitChanpura: I have moved it to be async in latest commit.

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
DarshitChanpura: With move to async evaluator, I was able to call .insufficient() method.

DarshitChanpura:
@nibix @cwperks All outstanding concerns have been addressed. mind re-reviewing the PR?

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -384,98 +398,114 @@ private <Request extends ActionRequest, Response extends ActionResponse> void ap
...
                 }
                 log.debug(err);
-
                 listener.onFailure(new OpenSearchSecurityException(err, RestStatus.FORBIDDEN));
-            }
+            };
+            // NOTE: Since resource-access evaluation requires fetching documents from index, we make the call async otherwise it would
+            // require blocking transport threads leading to thread exhaustion and request timeouts
+            // We perform the rest of the evaluation as normal if the request is not for resource-access or if the feature is disabled
+            resourceAccessEvaluator.evaluateAsync(request, action, context, ActionListener.wrap(response -> {
```
cwperks: I'd like to see if we can move this down into PrivilegesEvaluator along with the rest of the authorizers like system index and snapshot restore.

I know that one of the problems is `PrivilegesEvaluatorResponse pres = eval.evaluate(context);` expects a return value of `PrivilegesEvaluatorResponse` and not designed to be async.

I'm wondering if it makes sense to change this to either return an `Optional` or just handle the case where it can return `null` and then simply `return`. i.e. if the return value indicates that privilege evaluation is not done, then it means its being done async so stop here and the decision to proceed will be handled in the async call where it either proceeds through the chain or returns early with forbidden error if the user does not have access.

DarshitChanpura: There is a markPending() method in PrivilegeEvalResponse class. I tried using it but was not successful. I can give it another try. With rest of PrivilegeEval being synchronous there are only 2 paths forward, as i see:
1. Create a delegating handler that handles rest of the evaluation when feature is disabled or request is not of Resource type
2. Utilize markPending() to indicate that access is currently being evaluated and will be processed by the evaluator. However, due to the nature of PrivilegeEval being synchronous i'm not entirely sure if that will be helpful.

cwperks: `markPending()` looks perfectly suited for this use-case :). I see its not utilized atm as well.

DarshitChanpura: I tried with markPending(), the end result will still create a blocking operation since in SecurityFilter we will have to wait for response to be complete before proceeding. I saw the same timeouts in test runs for sample plugin. I think the only path forward is to utilize the current async approach. 

To make the code cleaner, i've moved the logic just before the privilege eval call in Security Filter which checks whether resource evaluator should be used before proceeding to actually use it.

DarshitChanpura:
@nibix @shikharj05 Mind reviewing this PR. Targeting 3.2 release for this PR.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -384,6 +399,45 @@ private <Request extends ActionRequest, Response extends ActionResponse> void ap
...
+                log.debug(err);
+                listener.onFailure(new OpenSearchSecurityException(err, RestStatus.FORBIDDEN));
+            };
+
+            // NOTE: Since resource-access evaluation requires fetching documents from index, we make the call async otherwise it would
+            // require blocking transport threads leading to thread exhaustion and request timeouts
+            // We perform the rest of the evaluation as normal if the request is not for resource-access or if the feature is disabled
+            if (resourceAccessEvaluator.shouldEvaluate(request)) {
+                resourceAccessEvaluator.evaluateAsync(request, action, context, ActionListener.wrap(response -> {
+                    if (handlePermissionCheckRequest(listener, response, action)) {
```
cwperks, line 425: its not clear from looking at this that the response will be sent from within this method if it is a dry run.

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
```diff
@@ -0,0 +1,123 @@
...
+        // if it reached this evaluator, it is safe to assume that the request if of DocRequest type
+        DocRequest req = (DocRequest) request;
+
+        resourceAccessHandler.hasPermission(req.id(), req.index(), action, context, ActionListener.wrap(hasAccess -> {
+            if (hasAccess) {
+                pResponse.allowed = true;
+                pResponseListener.onResponse(pResponse.markComplete());
+                return;
+            }
+            ;
```
cwperks: nit: remove dangling semi-colon

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
```diff
@@ -0,0 +1,123 @@
...
+            pResponseListener.onResponse(PrivilegesEvaluatorResponse.insufficient(action).markComplete());
+        }, e -> { pResponseListener.onResponse(pResponse.markComplete()); }));
```
cwperks, line 95: Let's log out the error message if this condition is ever met.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
DarshitChanpura: Checkout this test class that I specifically added to test dry-run scenario:
https://github.com/opensearch-project/security/pull/5408/files#diff-070be969106813dafad9545cb4dc2c41f8ec5800d3e52841712b002a0b3bc381

src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java
DarshitChanpura: it is logged in hasPermissions method in ResourceAccessHandler on line 208:
https://github.com/opensearch-project/security/pull/5408/files#diff-2e067d91792a15778a6e386b4dc89f3015138348e704d933d1b9ff98965215cc

src/main/java/org/opensearch/security/filter/SecurityFilter.java
DarshitChanpura: I think the method name `handleIfDryRunRequest` or `checkAndHandleDryRunRequest` might be more suitable.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -365,126 +353,68 @@ public void fetchSharingInfo(String resourceIndex, String resourceId, ActionList
...
-        LOGGER.debug(
-            "Fetching document from {}, matching source_idx: {}, resource_id: {}",
-            resourceSharingIndex,
-            resourceIndex,
-            resourceId
-        );
-
+        LOGGER.debug("Fetching document from {}, matching resource_id: {}", resourceSharingIndex, resourceId);
+
+        String failureResponse = "Failed to parse document matching resource_id: "
```
cwperks: extract this to a function and only create this when needed.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -493,64 +423,50 @@ public void onFailure(Exception e) {
...
                 sharingInfo.share(accessLevel, target);
             }
 
             String resourceSharingIndex = getSharingIndex(resourceIndex);
             try (ThreadContext.StoredContext ctx = threadPool.getThreadContext().stashContext()) {
-                assert sharingInfo != null;
                 IndexRequest ir = client.prepareIndex(resourceSharingIndex)
                     .setId(sharingInfo.getResourceId())
-                    .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE)
+                    .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE) // TODO check this policy
```
cwperks: nit: remove this TODO, any index into a system index should have refresh immediate

cwperks: ...that actually should be enforced by the core for system indices as a special attribute of a system index. i.e. index, update and delete operations to a sys index are transactional in nature and should be reflected immediately post-op.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: i don't think that is needed. We print same message in both catch blocks. I've changed the response to be unique to each catch block.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -147,44 +156,57 @@ public void hasPermission(
...
         }
 
+        PrivilegesEvaluationContext effectiveContext = context != null ? context : privilegesEvaluator.createContext(user, action);
+
         Set<String> userRoles = new HashSet<>(user.getSecurityRoles());
         Set<String> userBackendRoles = new HashSet<>(user.getRoles());
 
-        this.resourceSharingIndexHandler.fetchSharingInfo(resourceIndex, resourceId, ActionListener.wrap(document -> {
+        // At present, plugins and tokens are not supported for access to resources
+        if (!(effectiveContext.getActionPrivileges() instanceof RoleBasedActionPrivileges roleBasedActionPrivileges)) {
```
cwperks, line 165: Looks like this is added to get the latest FlattenedActionGroups? We may need to create a mechanism where we can access the latest FlattenedActionGroup after any config change.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
cwperks: goal is to avoid creating a String unless its needed so better to create JIT.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
DarshitChanpura: This is how we fetch action groups today in other privilege evaluators as well. Here is one such example:
https://github.com/opensearch-project/security/blob/d738b3c6bc22c1e1a6c0bcd2f94711ba98a725fe/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L279

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: I agree. I think i added this todo while I was debugging Windows failures where I tried WAIT_UNTIL. But eventually those failures turned out to be sync vs async nature issue of ResourceAccessEvaluator.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -384,6 +398,45 @@ private <Request extends ActionRequest, Response extends ActionResponse> void ap
...
+                    }
+                    if (response.isAllowed()) {
+                        auditLog.logGrantedPrivileges(action, request, task);
+                        auditLog.logIndexEvent(action, request, task);
+                        chain.proceed(task, action, request, listener);
+                    } else {
+                        handleUnauthorized.accept(response);
+                    }
+                }, listener::onFailure));
+                return;
```
cwperks, line 438: I want to make sure my understanding is correct and this is the last comment that I have.

It looks like returning early here without proceeding will also cause this [method](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/TransportAction.java#L218) to return and free up the thread back to the threadpool. When the ActionListener here gets a response its picked up by a different thread that then chooses to proceed through the chain or respond back with a failure message or unauthorized. This looks fine to me as long as there's no blocking action.

DarshitChanpura: Return statement here ensures that if it resource permission request we want to unblock the threadpool, then perform the async eval operation, then return 403 or call chain.proceed depending on the response. Without the return statement rest of the code will continue execution. Other option to avoid return; here is to put rest of the code in an else block, but `return;` is much cleaner.

cwperks (approved):

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/SecurityDisabledTests.java
```diff
@@ -84,20 +110,26 @@ public void testSamplePluginAPIs() {
...
 
         }
     }
 
-    private void assertNotImplementedResponse(TestRestClient.HttpResponse response) {
-        response.assertStatusCode(HttpStatus.SC_BAD_REQUEST);
-        assertThat(response.getTextFromJsonBody("/error"), containsString("no handler found for uri"));
+    private void assertNotImplementedResponse(TestRestClient.HttpResponse response, String msg) {
+        response.assertStatusCode(HttpStatus.SC_NOT_IMPLEMENTED);
+        assertThat(response.getTextFromJsonBody("/error/reason"), containsString(msg));
```
nibix: Nit: We have the class `RestMatchers` which provides a couple of hamcrest matchers which do basically the same as this method. We do not have a matcher for "not implemented", but it should be a matter of 3 minutes to add one.

https://github.com/opensearch-project/security/blob/9b578353fbbcc58d850d86a7e7f726d6439d6762/src/integrationTest/java/org/opensearch/test/framework/matcher/RestMatchers.java#L50-L96

It's not necessary to change now, but in the long run it would make sense to use uniform assertion methods.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/rest/revoke/RevokeResourceAccessRequest.java
```diff
@@ -9,41 +9,38 @@
...
 
 /**
  * Request object for revoking access to a sample resource
  */
-public class RevokeResourceAccessRequest extends ActionRequest {
+public class RevokeResourceAccessRequest extends ActionRequest implements DocRequest {
 
     String resourceId;
-    Map<Recipient, Set<String>> entitiesToRevoke;
+    ShareWith revokeAccess;
```
nibix, line 28: Nit: Is it necessary to expose this package private?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1177,7 +1182,40 @@ public Collection<Object> createComponents(
...
+            // Assign resource sharing client to each extension
+            // Using the non-gated client (i.e. no additional permissions required)
+            ResourceSharingClient resourceAccessControlClient = new ResourceAccessControlClient(resourceAccessHandler);
+            resourcePluginInfo.getResourceSharingExtensions().forEach(extension -> {
+                extension.assignResourceSharingClient(resourceAccessControlClient);
+            });
+            components.add(resourcePluginInfo);
+            components.add(resourceAccessControlClient);
+            components.add(resourceAccessHandler);
+            components.addAll(resourceSharingExtensions);
```
nibix, line 1202: What's the reason we add the `resourceSharingExtensions` to the components available for dependency injection? These are objects that come from other plugins, right?

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -384,6 +398,45 @@ private <Request extends ActionRequest, Response extends ActionResponse> void ap
...
+                }
+                log.debug(err);
+                listener.onFailure(new OpenSearchSecurityException(err, RestStatus.FORBIDDEN));
+            };
+
+            // NOTE: Since resource-access evaluation requires fetching documents from index, we make the call async otherwise it would
+            // require blocking transport threads leading to thread exhaustion and request timeouts
+            // We perform the rest of the evaluation as normal if the request is not for resource-access or if the feature is disabled
+            if (resourceAccessEvaluator.shouldEvaluate(request)) {
+                resourceAccessEvaluator.evaluateAsync(request, action, context, ActionListener.wrap(response -> {
```
nibix, line 424: Looks cool! 👍

src/main/java/org/opensearch/security/filter/SecurityFilter.java
nibix: I also think return is okay. Maybe it helps to add a comment that indicates that the control flow continues in the ActionListener above.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
nibix: ActionPrivileges is kind of unrelated to resource sharing. It just happens to hold a reference to the FlattenedActionGroups object for its own internal purposes. 

The lifecycle of the object is currently managed by PrivilegesEvaluator: 

https://github.com/opensearch-project/security/blob/90adaefefac7a51811d941c56c733c0b628ae0b9/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L239-L265

How about either add a line there which forwards the instance to here; or just keep a reference to FlattenedActionGroups in PrivilegesEvaluator? We could also add the reference directly to the PrivilegeEvaluationContext.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -312,6 +319,14 @@ && getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_ORIGIN_HEADE
...
             }
 
+            // put user as userSubject, we'll recreate it in messageReceivedDecorate
+            String authSubjectHeader = getThreadContext().getHeader(ConfigConstants.OPENDISTRO_SECURITY_AUTHENTICATED_USER_HEADER);
+            if (authSubjectHeader == null && authSubject != null) {
+                getThreadContext().putHeader(
+                    ConfigConstants.OPENDISTRO_SECURITY_AUTHENTICATED_USER_HEADER,
+                    authSubject.getUser().toSerializedBase64()
+                );
+            }
```
nibix: If I understood the other code correctly, we'll always have an authSubject, even if it is identical to the `User`. Is that correct?

Some environments can have surprisingly large user objects. Sending these twice along with each request could significantly increase network latency.

nibix (commented):
The async handling for resource privilege evaluation looks good! :+1:

I am a bit surprised by the additional ThreadContext header introduced in this PR. It feels a bit unrelated. This will be enabled independent of the feature flag, correct? As this will significantly increase the request size for each request, this is a bit critical, IMHO.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: these will be use by the plugins in their transport actions to retrieve resource sharing client. that way they don't have to rely on checking for feature flag

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/SecurityDisabledTests.java
DarshitChanpura: done. https://github.com/opensearch-project/security/pull/5408/commits/f294e1f0a3a35114c9b35ef440c77d37e47f2940

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/rest/revoke/RevokeResourceAccessRequest.java
DarshitChanpura: yes, as the plugins sending share and revoke request need a standardized way to form the share-with object. Allows to re-use same signature for share() and revoke() client methods.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
DarshitChanpura: I see the concern here. I've added a block that serializes auth header user only if it is not same as request user header, therefore serializing user object only once if they are same.

https://github.com/opensearch-project/security/pull/5408/commits/46e9e2bb23cc4824d06c0e034121ca0bad27791f

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
```diff
@@ -113,16 +113,28 @@ protected void messageReceivedDecorate(
...
+            // when auth subject user is same request user.
+            if (Boolean.parseBoolean(shouldUseUserHeader) && userHdr != null) {
+                User user = this.userFactory.fromSerializedBase64(userHdr);
+
+                getThreadContext().putPersistent(
+                    ConfigConstants.OPENDISTRO_SECURITY_AUTHENTICATED_USER,
+                    new UserSubjectImpl(getThreadPool(), user)
+                );
+            } else if (authUsrHdr != null) {
+                User user = this.userFactory.fromSerializedBase64(authUsrHdr);
```
nibix: The original user handling is done in these lines:

https://github.com/opensearch-project/security/blob/46e9e2bb23cc4824d06c0e034121ca0bad27791f/src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java#L197-L214

Can't we combine these into one? With the current code, we parse the user header two times

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
nibix: Do we have an example for that?

I think this behavior should be pointed out in a comment, as this is an approach one might not understand immediately.

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
DarshitChanpura: let me try it

DarshitChanpura: done [00dd8b4](https://github.com/opensearch-project/security/pull/5408/commits/00dd8b4f9464acfb0c7611b4e62287a9ac208798)

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
cwperks: @nibix this is a pattern I used recently in job scheduler to enforce that Job Scheduler was the plugin to instantiate a LockService and other plugins can obtain it through dependency injection if needed (although it is separately available from the JobExecutionContext)

1. JS PR: https://github.com/opensearch-project/job-scheduler/pull/670
2. Example of a plugin consuming the instance: https://github.com/opensearch-project/security-analytics/pull/1555

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
nibix: I am not sure if i get something wrong, but there's another parsing in line 217

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: here are few examples from:
- AD Plugin: https://github.com/opensearch-project/anomaly-detection/pull/1533/files#diff-48dd2ca0f511c609289566156af2a361f8d8c738fc0c1c0113b0483a8f36dec9R41

- ML Plugin: https://github.com/opensearch-project/ml-commons/pull/3715/files#diff-c341f5a67c1d221e5593dd55fb689e9d0c9a60d33e0fa26090803cfa90818fa7R117

- Sample plugin: https://github.com/opensearch-project/security/pull/5408/files#diff-e9dda849ebac77d4f55c0d8ccaf90db98c9a46386bdb23c027d87ea1c73d416dR61
-

DarshitChanpura: I've added a code comment

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
DarshitChanpura: done [5d41f95](https://github.com/opensearch-project/security/pull/5408/commits/5d41f95a2ba8969afe90ccb4a0bce52d16672108)

nibix (approved):
Thank you!

Approving. Still, I think we need to put a bit more thought into the serialization handling of the UserSubject; IMHO, it has quite a bit potential for inconsistencies at the moment:

https://github.com/opensearch-project/security/blob/5d41f95a2ba8969afe90ccb4a0bce52d16672108/src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java#L179-L239

But as this is not actually the central point of this PR, this can and should be handled separately.

PR #5425, opened 2025-06-23
Fix 5 second delay while closing RestHighLevelClient in integration tests

nibix:
### Description

While running integration tests manually, I have noticed that single test methods often required an unexpectedly long time to execute, often around 5 seconds, sometimes 10 seconds or 15 seconds. This is significantly too long, these methods should execute within a few milliseconds.

Testing has shown that the `close()` method of `RestHighLevelClient` incurs this delay, indirectly, while shutting down the `PoolingAsyncClientConnectionManager`.

I have tried a couple of approaches to fix this, but none proved really good/feasible in the end. Thus, here's just a quick fix which executes the `close()` method asynchronously. That fixes the delay.

* Category: Test fix
* Why these changes are required? Tests require an unproportionally long time to execute
* What is the old behavior before changes and new behavior after changes? No change

### Testing

- Manually verified the lower execution time

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/test/framework/cluster/OpenSearchClientProvider.java
+++ b/src/integrationTest/java/org/opensearch/test/framework/cluster/OpenSearchClientProvider.java
@@ -28,6 +28,7 @@
 
 package org.opensearch.test.framework.cluster;
 
+import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.URI;
@@ -38,6 +39,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.ForkJoinPool;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.net.ssl.KeyManager;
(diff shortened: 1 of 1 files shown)


shikharj05:
Great @nibix !  looks like a nice reduction in overall test runs as well! thanks :)

shikharj05:
nit: looks like a change is required here- https://github.com/opensearch-project/security/blob/main/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/SampleResourcePluginFeatureDisabledTests.java#L54-L60

nibix:
@shikharj05 

It seems there was a dependency issue in the resource sharing sample. That should be fixed now.

sample-resource-plugin/build.gradle
```diff
@@ -66,6 +66,7 @@ configurations.all {
         force 'commons-codec:commons-codec:1.18.0'
         force 'org.hamcrest:hamcrest:2.2'
         force 'org.mockito:mockito-core:5.18.0'
+        force 'org.slf4j:slf4j-api:1.7.36'
```
nagarajg17, line 69: why this dependency needed for this change?

nibix: Otherwise, the sample plugin won't compile; the same force is also in the build.gradle of the main security plugin.

But, indeed the slf4j dependencies are odd. In build.gradle, there are two different versions used at different places: 1.7.36 but also 2.0.12.

DarshitChanpura (approved):

PR #5357, opened 2025-05-22
Versioned Security Configuration Management

divyanshi-0402:
### Description

Implementing a comprehensive versioning system for the OpenSearch security index. This system will enable rollback and roll-forward capabilities, provide a history of changes, and allow administrators to easily view and select specific configurations from the past.

**New System Index**: Introduce a dedicated system index named `.opendistro_security_config_versions` to store versioned security configurations.

**Versioning Mechanism**:

- Generate a new version (e.g., v1, v2, v3) for each change in the security index.
- Store the complete security configuration for each version in the new system index.
- Have retention policy to purge old versions

**Updated Security Config Update Flow**: Modify the existing security configuration update process to

- Create a new version entry in the `.opendistro_security_config_versions` index
- Store the full configuration snapshot for each version

Index status with 10 security config versions : 

`green open .opendistro_security_config_versions 4Hvxs5KOST6gN2YtFNHlZg 1 0  1 0  60.1kb  60.1kb
`

Index Structure : 

```
{
  "_index": ".opendistro_security_config_versions",
  "_id": "opendistro_security_config_versions",
  "_version": 1,
  "_seq_no": 0,
  "_primary_term": 1,
  "found": true,
  "_source": {
    "versions": [
      {
        "version_id": "v1",
        "timestamp": "2025-05-22T08:46:11.887620466Z",
        "modified_by": "system",
        "security_configs": {
          "rolesmapping": {
            "lastUpdated": "2025-05-22T08:46:11.887620466Z",
            "configData": {
              "all_access": {
                "hosts": [],
                "users": [],
                "reserved": false,
                "hidden": false,
                "backend_roles": ["admin"],
                "and_backend_roles": [],
                "description": "Maps admin to all_access"
              },
              "kibana_server": {
                "hosts": [],
                "users": ["kibanaserver"],
                "reserved": true,
                "hidden": false,
                "backend_roles": [],
                "and_backend_roles": []
              }
              ... 
            }
          },
          ...
        }
      },
      ...
    ]
  }
}

```

### Issues Resolved

Related to [#5093](https://github.com/opensearch-project/security/issues/5093)

### Testing

**Manual Testing :** 

- No version update in case of no change in security configurations

```
 Index request for .opendistro_security_config_versions
 Index .opendistro_security_config_versions already exists
 Version index already exists, skipping initialization.
 Node started, try to initialize it. Wait for at least yellow cluster state....
 No changes detected in security configuration.
 No changes detected in security configuration. Skipping version update.

```

[...]

Issue #5093, opened 2025-02-05
Create a mechanism for tracking patches to the security index to eventually enable rollback/rollforward of adjacent changes


cwperks:
**Is your feature request related to a problem?**

Currently, its possible to use the security audit log to track changes to the security index when listening for `COMPLIANCE_INTERNAL_CONFIG_WRITE` events. Its also possible to combine this with `config.compliance.enabled: true`, `config.compliance.write_log_diffs: true` and `config.compliance.write_metadata_only: false` to only capture the diffs for security index requests.

The problem with this is that its not intuitive to configure and leaves cluster operators filtering the audit log for these types of events to figure out security config changes.

I'm opening this issue to spark a discussion about separating these config changes out to a separate place with the eventual goal of supporting rollback/rollforward in case a cluster operator wants to revert to the last previously known good cluster state.

nagarajg17:
## Proposal :  Versioned Security Configuration Management

To address the current limitations in security configuration management, proposal is implementing a comprehensive versioning system for the OpenSearch security index. This system will enable rollback and roll-forward capabilities, provide a history of changes, and allow administrators to easily view and select specific configurations from the past.

### Key Components:

**New System Index**: Introduce a dedicated system index named `.opendistro_security_config_versions` to store versioned security configurations.

**Versioning Mechanism**: 
- Generate a new version (e.g., v1, v2, v3) for each change in the security index. 
- Store the complete security configuration for each version in the new system index.

**Updated Security Config Update Flow**: Modify the existing security configuration update process to
- Create a new version entry in the `.opendistro_security_config_versions index`
- Store the full configuration snapshot for each version

**Rollback/Roll-forward Functionality**: 
- Implement a mechanism to revert to or advance to a specific version. 
- Include compatibility checks to ensure the selected version is compatible with the current system state.

**Change History Viewer**: Develop a API to allow administrators or security managers to
- View the complete history of security configuration changes.
- Compare differences between versions.
- Select a specific version for rollback or roll-forward


### Alternative Approach
Instead of storing complete security configurations, only store the differences (diffs) between versions in the `.opendistro_security_config_versions` index. While this approach is more space-efficient, it is creating increased complexity in reconstructing complete configurations for rollbacks and potential for errors in the diff chain, which could ultimately result in compromising the integrity of the security settings

### API Design

#### View Security Configuration Versions

* Endpoint:  
    * To view all the versions : GET _opendistro/_security/api/config/version/_all
[...]

PR #5431, opened 2025-06-23
Added View API and Rollback API for Versioned Security Configurations

divyanshi-0402:
### Description

Implementing View API and Rollback API for Versioned Security Configurations,  build upon the changes from the PR : [Versioned Security Configuration Management #5357](https://github.com/opensearch-project/security/pull/5357).

### **View Security Configuration Versions**
Purpose: View the complete history of security configuration changes, or any specific version of the security configurations.

 Endpoint:
- To view all the versions : GET /_plugins/_security/api/version
- To view a specific version : GET /_plugins/_security/api/version/{versionId}

Access Control: Admin/Security Manager permissions required

Response Format for specific version:

```
curl -XGET "https://localhost:9200/_plugins/_security/api/version/v2?pretty" -u 'admin:Divvv@admin123' --insecure

{
  "versions": [
    {
      "version_id": "v2",
      "timestamp": "2025-05-23T06:56:20.081933886Z",
      "modified_by": "admin",
      "security_configs": {
        "rolesmapping": {
          "lastUpdated": "2025-05-23T06:56:20.081933886Z",
          "configData": {
             "_meta" : {
              "type" : "rolesmapping",
              "config_version" : 2
            },
            "all_access": {
              "backend_roles": ["admin"],
              "users": [],
              "hosts": [],
              "reserved": false,
              "hidden": false,
              "description": "Maps admin to all_access",
              "and_backend_roles": []
            },
            ...
          }
        },
        "roles": {
          "lastUpdated": "2025-05-23T06:56:20.081933886Z",
          "configData": {
            "_meta" : {
              "type" : "roles",
              "config_version" : 2
            },
            "alerting_full_access": {
              "cluster_permissions": [
                "cluster:admin/opendistro/alerting/*",
                "cluster:admin/opensearch/alerting/*",
                ...
              ],
              "index_permissions": [
                {
                  "index_patterns": ["*"],
                  "allowed_actions": ["indices:admin/mappings/get", ...]
                }
              ],
              "reserved": true,
              "hidden": false,
              "static": false
            },
            ...
          }
        },
        "internalusers": {
          "lastUpdated": "2025-05-23T06:56:20.081933886Z",
          "configData": {
            "_meta" : {
              "type" : "internalusers",
              "config_version" : 2
            },
            "admin": {
              "backend_roles": ["admin"],
              "reserved": true,
              "description": "Demo admin user",
              ...
            },
            "testuser": {
              "backend_roles": [],
              "reserved": false,
              ...
            },
            ...
          }
        },
        ...
      }
    }
  ]
}

```

Response Format for all versions :

```
curl -XGET "https://localhost:9200/_plugins/_security/api/version?pretty" -u 'admin:Divvv@admin123' --insecure

{
  "versions": [
    {
      "version_id": "v1",
      "timestamp": "2025-05-22T08:46:11.887620466Z",
      "modified_by": "system",
      "security_configs": {
        "rolesmapping": {
          "lastUpdated": "2025-05-22T08:46:11.887620466Z",
          "configData": {
            "_meta" : {
              "type" : "rolesmapping",
              "config_version" : 2
            },
            "all_access": {
              "backend_roles": ["admin"],
              "users": [],
              "hosts": [],
              "reserved": false,
              "hidden": false,
              "description": "Maps admin to all_access",
              "and_backend_roles": []
            },
            ...
          }
        },
        "internalusers": {
          "lastUpdated": "2025-05-22T08:46:11.887620466Z",
          "configData": {
             "_meta" : {
              "type" : "internalusers",
              "config_version" : 2
            },
            "admin": {
              "backend_roles": ["admin"],
              "reserved": true,
              "hidden": false,
              "description": "Demo admin user",
              ...
            },
            ...
          }
        },
        ...
      }
    },
    {
      "version_id": "v2",
      "timestamp": "2025-05-23T06:56:20.081933886Z",
      "modified_by": "admin",
      "security_configs": {
        "rolesmapping": {
          "lastUpdated": "2025-05-23T06:56:20.081933886Z",
          "configData": {
             "_meta" : {
              "type" : "rolesmapping",
              "config_version" : 2
            },
            "all_access": {
              "backend_roles": ["admin"],
              "users": [],
              "hosts": [],
              "reserved": false,
              "hidden": false,
              "description": "Maps admin to all_access",
              "and_backend_roles": []
            },
            ...
          }
        },
        "internalusers": {
          "lastUpdated": "2025-05-23T06:56:20.081933886Z",
          "configData": {
            "_meta" : {
              "type" : "internalusers",
              "config_version" : 2
            },
            "admin": {
              "backend_roles": ["admin"],
              "reserved": true,
              "hidden": false,
              "description": "Demo admin user",
              ...
            },
            "testuser": {
              "backend_roles": [],
              "reserved": false,
              "hidden": false,
              ...
            },
            ...
          }
        },
        ...
      }
    },
    ...
  ]
}

```

Response format if the version doesn't exist : 

```
curl -XGET "https://localhost:9200/_plugins/_security/api/version/v3?pretty" -u 'admin:Divvv@admin123' --insecure
{
  "status" : "NOT_FOUND",
  "message" : "Version v3 not found"
}

```

Response if try to access with non admin role : 

```
curl -XGET "https://localhost:9200/_plugins/_security/api/version?pretty" -u 'test_user:TestAPI@123' --insecure                       
{
  "status" : "FORBIDDEN",
  "message" : "No permission to access REST API: User test_user with Security roles [read_only_role, own_index] does not have any role privileged for admin access. No client TLS certificate found in request"
}
```

### **Rollback to desired Security Configuration Versions**
Purpose : Allows rollback to immediate or any desired version of security 

Endpoint:
- To rollback to immediate previous version : POST /_plugins/_security/api/rollback
- To rollback to a specific version : POST /_plugins/_security/api/rollback/version/{versionID}

Access Control: Admin/Security Manager permissions required

Response format to rollback to immediate previous version : 

```
[.opendistro_security_config_versions] index has 8 versions 

curl -XPOST "https://localhost:9200/_plugins/_security/api/rollback" -u 'admin:Divvv@admin123' --insecure           
{
  "status" : "OK",
  "message" : "config rolled back to version v7"
}


 Detected changes in security configuration: [{"op":"add","path":"/internalusers/testuser","value":{"attributes":{},"backend_roles":[],"hash":"$2y$12$3q6lxV8DBxvifcRJ4tMuUeclwyJzawJTzhY.JcKEpL1TdWy.JeUPu","hidden":false,"opendistro_security_roles":[],"reserved":false,"static":false}}]
 Successfully saved version v9 to .opendistro_security_config_versions
```

Response format to rollback to a specific version : 

```
curl -XPOST "https://localhost:9200/_plugins/_security/api/rollback/version/v3" -u 'admin:Divvv@admin123' --insecure
{
  "status" : "OK",
  "message" : "config rolled back to version v3"
}% 


Detected changes in security configuration: [{"op":"remove","path":"/internalusers/testuser"}]
Successfully saved version v10 to .opendistro_security_config_versions
```

Response format if the version doesn't exist : 

```
curl -XPOST "https://localhost:9200/_plugins/_security/api/rollback/version/v13" -u 'admin:Divvv@admin123' --insecure
{
  "status" : "NOT_FOUND",
  "message" : "Version v13 not found"
}
```

Response if try to access with non admin role : 

```
curl -XPOST "https://localhost:9200/_plugins/_security/api/rollback?pretty" -u 'test_user:TestAPI@123' --insecure 
{
  "status" : "FORBIDDEN",
  "message" : "No permission to access REST API: User test_user with Security roles [read_only_role, own_index] does not have any role privileged for admin access. No client TLS certificate found in request"
}
```

### Test details for multi-node testing: 

Created two nodes : `opensearch-node1` and `opensearch-node1`

node 1 :  
```
./bin/opensearch -E cluster.name=my-cluster -E node.name=node1 -E path.data=./data1 -E http.port=9200 -E transport.port=9300

[node1] Successfully saved version v1 to .opensearch_security_config_versions

[INFO ][o.o.s.c.ConfigurationRepository] [node1] Node 'node1' initialized
```

node 2 : 

```
./bin/opensearch -E cluster.name=my-cluster -E node.name=node2 -E path.data=./data2 -E http.port=9201 -E transport.port=9301

[INFO ][o.o.s.c.ConfigurationRepository] [node2] Node 'node2' initialized
```

Rollback initiated from node-1: 

```
curl -XPOST "https://172.19.119.134:9200/_plugins/_security/api/rollback" -u 'admin:admin' --insecure
{
  "status" : "OK",
  "message" : "config rolled back to version v1"
}%                   

[node1] Detected changes in security configuration: [{"op":"remove","path":"/internalusers/testuser"}]
[node1] Successfully saved version v3 to .opensearch_security_config_versions
```

Rollback initiated from node-2 : 

```
curl -XPOST "https://172.19.119.134:9201/_plugins/_security/api/rollback/version/v1" -u 'admin:admin' --insecure                                       
{
  "status" : "OK",
  "message" : "config rolled back to version v1"
}

[node1] Detected changes in security configuration: [{"op":"remove","path":"/internalusers/testuser"}]
[node1] Successfully saved version v5 to .opensearch_security_config_versions
```

### Issues Resolved
Related to [#5093](https://github.com/opensearch-project/security/issues/5093)

### Testing
Unit Tests and Integration Tests included

### Check List

### DIFF
--- a/.github/workflows/dependabot_pr.yml
+++ b/.github/workflows/dependabot_pr.yml
@@ -60,6 +60,7 @@ jobs:
         uses: dangoslen/dependabot-changelog-helper@v4
         with:
           version: 'Unreleased 3.x'
+          sectionHeader: "Maintenance"
 
       - name: Commit the changes
         uses: stefanzweifel/git-auto-commit-action@v6
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,15 +5,29 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 3.x]
 ### Added
-- Introduced new experimental versioned security configuration management feature ([#5357] (https://github.com/opensearch-project/security/pull/5357))
-- Create a mechanism for plugins to explicitly declare actions they need to perform with their assigned PluginSubject ([#5341](https://github.com/opensearch-project/security/pull/5341))
-- Introduced View API and Rollback API for experimental versioned security configurations ([#5431](https://github.com/opensearch-project/security/pull/5431))
+- Provide SecureHttpTransportParameters to complement SecureTransportParameters counterpart ([#5432](https://github.com/opensearch-project/security/pull/5432))
 
-### Changed
-- Moved OpenSAML jars to a Shadow Jar configuration to facilitate its use in FIPS enabled environments ([#5400](https://github.com/opensearch-project/security/pull/5404))
-- Fix compilation issue after change to Subject interface in core and bump to 3.2.0 ([#5423](https://github.com/opensearch-project/security/pull/5423))
+### Features
 
-### Dependencies
+* Introduced new experimental versioned security configuration management feature ([#5357] (https://github.com/opensearch-project/security/pull/5357))
+* Introduced View API and Rollback API for experimental versioned security configurations ([#5431](https://github.com/opensearch-project/security/pull/5431))
+
+### Enhancements
+
+* Create a mechanism for plugins to explicitly declare actions they need to perform with their assigned PluginSubject ([#5341](https://github.com/opensearch-project/security/pull/5341))
+* Moved OpenSAML jars to a Shadow Jar configuration to facilitate its use in FIPS enabled environments ([#5400](https://github.com/opensearch-project/security/pull/5404))
+* Replaced the standard distribution of BouncyCastle with BC-FIPS ([#5439](https://github.com/opensearch-project/security/pull/5439))
+
+### Bug Fixes
+
+* Fix compilation issue after change to Subject interface in core and bump to 3.2.0 ([#5423](https://github.com/opensearch-project/security/pull/5423))
+
+### Refactoring
+
+
+
+### Maintenance
+
 - Bump `org.eclipse.platform:org.eclipse.core.runtime` from 3.33.0 to 3.33.100 ([#5400](https://github.com/opensearch-project/security/pull/5400))
 - Bump `org.eclipse.platform:org.eclipse.equinox.common` from 3.20.0 to 3.20.100 ([#5402](https://github.com/opensearch-project/security/pull/5402))
 - Bump `spring_version` from 6.2.7 to 6.2.8 ([#5403](https://github.com/opensearch-project/security/pull/5403))
@@ -23,16 +37,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Bump `net.bytebuddy:byte-buddy` from 1.17.5 to 1.17.6 ([#5427](https://github.com/opensearch-project/security/pull/5427))
 - Bump `io.dropwizard.metrics:metrics-core` from 4.2.32 to 4.2.33 ([#5428](https://github.com/opensearch-project/security/pull/5428))
 
-
-### Deprecated
-
-
-### Removed
+### Documentation
 
 
-### Fixed
-
-
-### Security
-
-[Unreleased 3.x]: https://github.com/opensearch-project/security/compare/3.0...main
+[Unreleased 3.x]: https://github.com/opensearch-project/security/compare/3.1...main
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,9 @@ buildscript {
         guava_version = '33.4.8-jre'
         jaxb_version = '2.3.9'
         spring_version = '6.2.8'
-        bouncycastle_version = '1.81'
+        bc_fips_version = '2.0.0'
+        bcpkix_fips_version = '2.0.7'
+        bcutil_fips_version = '2.0.3'
 
         if (buildVersionQualifier) {
             opensearch_build += "-${buildVersionQualifier}"
@@ -536,8 +538,9 @@ allprojects {
         integrationTestImplementation "org.apache.logging.log4j:log4j-core:${versions.log4j}"
         integrationTestImplementation "org.apache.logging.log4j:log4j-jul:${versions.log4j}"
         integrationTestImplementation 'org.hamcrest:hamcrest:2.2'
-        integrationTestImplementation "org.bouncycastle:bcpkix-jdk18on:${bouncycastle_version}"
-        integrationTestImplementation "org.bouncycastle:bcutil-jdk18on:${bouncycastle_version}"
+        integrationTestImplementation "org.bouncycastle:bc-fips:${bc_fips_version}"
+        integrationTestImplementation "org.bouncycastle:bcpkix-fips:${bcpkix_fips_version}"
+        integrationTestImplementation "org.bouncycastle:bcutil-fips:${bcutil_fips_version}"
         integrationTestImplementation('org.awaitility:awaitility:4.3.0') {
             exclude(group: 'org.hamcrest', module: 'hamcrest')
         }
@@ -641,12 +644,6 @@ tasks.integrationTest.finalizedBy(jacocoTestReport) // report is always generate
 //run the integrationTest task before the check task
 check.dependsOn integrationTest
 
-configurations.all {
-    exclude group:"org.bouncycastle", module: "bc-fips"
-    exclude group:"org.bouncycastle", module:"bctls-fips"
-    exclude group:"org.bouncycastle", module:"bcutil-fips"
-    exclude group:"org.bouncycastle", module:"bcpkix-fips"
-}
 
 dependencies {
     implementation project(path: ":${rootProject.name}-spi", configuration: 'shadow')
@@ -659,7 +656,9 @@ dependencies {
     implementation "com.google.guava:guava:${guava_version}"
     implementation 'org.greenrobot:eventbus-java:3.3.1'
     implementation 'commons-cli:commons-cli:1.9.0'
-    implementation "org.bouncycastle:bcprov-jdk18on:${bouncycastle_version}"
+    implementation "org.bouncycastle:bc-fips:${bc_fips_version}"
+    implementation "org.bouncycastle:bcpkix-fips:${bcpkix_fips_version}"
+    implementation "org.bouncycastle:bcutil-fips:${bcutil_fips_version}"
     implementation 'org.ldaptive:ldaptive:1.2.3'
     implementation 'com.nimbusds:nimbus-jose-jwt:9.48'
     implementation 'com.rfksystems:blake2b:2.0.0'
@@ -720,7 +719,6 @@ dependencies {
     runtimeOnly 'org.apache.santuario:xmlsec:2.3.5'
     runtimeOnly "com.github.luben:zstd-jni:${versions.zstd}"
     runtimeOnly 'org.checkerframework:checker-qual:3.49.4'
-    runtimeOnly "org.bouncycastle:bcpkix-jdk18on:${bouncycastle_version}"
     runtimeOnly 'org.scala-lang.modules:scala-java8-compat_3:1.0.2'
 
 
@@ -763,9 +761,9 @@ dependencies {
     testImplementation('org.awaitility:awaitility:4.3.0') {
         exclude(group: 'org.hamcrest', module: 'hamcrest')
     }
-    testImplementation "org.bouncycastle:bcpkix-jdk18on:${bouncycastle_version}"
-    testImplementation "org.bouncycastle:bcutil-jdk18on:${bouncycastle_version}"
-
+    testImplementation "org.bouncycastle:bc-fips:${bc_fips_version}"
+    testImplementation "org.bouncycastle:bcpkix-fips:${bcpkix_fips_version}"
+    testImplementation "org.bouncycastle:bcutil-fips:${bcutil_fips_version}"
     // JUnit build requirement
     testCompileOnly 'org.apiguardian:apiguardian-api:1.1.2'
     // Kafka test execution
--- a/src/integrationTest/java/org/opensearch/test/framework/certificate/CertificatesIssuerFactory.java
+++ b/src/integrationTest/java/org/opensearch/test/framework/certificate/CertificatesIssuerFactory.java
@@ -12,7 +12,7 @@ package org.opensearch.test.framework.certificate;
 import java.security.Provider;
 import java.util.Optional;
 
-import org.bouncycastle.jce.provider.BouncyCastleProvider;
+import org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider;
 
 import static org.opensearch.test.framework.certificate.AlgorithmKit.ecdsaSha256withEcdsa;
 import static org.opensearch.test.framework.certificate.AlgorithmKit.rsaSha256withRsa;
@@ -30,7 +30,7 @@ class CertificatesIssuerFactory {
 
     }
 
-    private static final Provider DEFAULT_SECURITY_PROVIDER = new BouncyCastleProvider();
+    private static final Provider DEFAULT_SECURITY_PROVIDER = new BouncyCastleFipsProvider();;
 
     /**
     * @see {@link #rsaBaseCertificateIssuer(Provider)}
--- a/src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java
+++ b/src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java
@@ -25,6 +25,7 @@ import org.opensearch.http.HttpServerTransport;
 import org.opensearch.http.netty4.ssl.SecureNetty4HttpServerTransport;
 import org.opensearch.plugins.SecureAuxTransportSettingsProvider;
 import org.opensearch.plugins.SecureHttpTransportSettingsProvider;
+import org.opensearch.plugins.SecureHttpTransportSettingsProvider.SecureHttpTransportParameters;
 import org.opensearch.plugins.SecureSettingsFactory;
 import org.opensearch.plugins.SecureTransportSettingsProvider;
 import org.opensearch.plugins.TransportExceptionHandler;
@@ -83,6 +84,63 @@ public class OpenSearchSecureSettingsFactory implements SecureSettingsFactory {
 
                     @Override
                     public Optional<String> sslProvider() {
+                        return sslSettingsManager.sslConfiguration(CertType.TRANSPORT)
+                            .map(config -> config.sslParameters().provider().name());
+                    }
+
+                    @Override
+                    public Optional<String> clientAuth() {
+                        return sslSettingsManager.sslConfiguration(CertType.TRANSPORT)
+                            .map(config -> config.sslParameters().clientAuth().name());
+                    }
+
+                    @Override
+                    public Collection<String> protocols() {
+                        return sslSettingsManager.sslConfiguration(CertType.TRANSPORT)
+                            .map(config -> config.sslParameters().allowedProtocols())
+                            .orElse(Collections.emptyList());
+                    }
+
+                    @Override
+                    public Collection<String> cipherSuites() {
+                        return sslSettingsManager.sslConfiguration(CertType.TRANSPORT)
+                            .map(config -> config.sslParameters().allowedCiphers())
+                            .orElse(Collections.emptyList());
+                    }
+
+                    @Override
+                    public Optional<KeyManagerFactory> keyManagerFactory() {
+                        return sslSettingsManager.sslConfiguration(CertType.TRANSPORT).map(SslConfiguration::keyStoreFactory);
+                    }
+
+                    @Override
+                    public Optional<TrustManagerFactory> trustManagerFactory() {
+                        return sslSettingsManager.sslConfiguration(CertType.TRANSPORT).map(SslConfiguration::trustStoreFactory);
+                    }
+
+                });
+            }
+
+            @Override
+            public Optional<SSLEngine> buildSecureServerTransportEngine(Settings settings, Transport transport) throws SSLException {
+                return sslSettingsManager.sslContextHandler(CertType.TRANSPORT).map(SslContextHandler::createSSLEngine);
+            }
+
+            @Override
+            public Optional<SSLEngine> buildSecureClientTransportEngine(Settings settings, String hostname, int port) throws SSLException {
+                return sslSettingsManager.sslContextHandler(CertType.TRANSPORT_CLIENT).map(c -> c.createSSLEngine(hostname, port));
+            }
+        });
+    }
+
+    @Override
+    public Optional<SecureHttpTransportSettingsProvider> getSecureHttpTransportSettingsProvider(Settings settings) {
+        return Optional.of(new SecureHttpTransportSettingsProvider() {
+            @Override
+            public Optional<SecureHttpTransportParameters> parameters(Settings settings) {
+                return Optional.of(new SecureHttpTransportParameters() {
+                    @Override
+                    public Optional<String> sslProvider() {
                         return sslSettingsManager.sslConfiguration(CertType.HTTP).map(config -> config.sslParameters().provider().name());
                     }
 
@@ -118,21 +176,6 @@ public class OpenSearchSecureSettingsFactory implements SecureSettingsFactory {
                 });
             }
 
-            @Override
-            public Optional<SSLEngine> buildSecureServerTransportEngine(Settings settings, Transport transport) throws SSLException {
-                return sslSettingsManager.sslContextHandler(CertType.TRANSPORT).map(SslContextHandler::createSSLEngine);
-            }
-
-            @Override
-            public Optional<SSLEngine> buildSecureClientTransportEngine(Settings settings, String hostname, int port) throws SSLException {
-                return sslSettingsManager.sslContextHandler(CertType.TRANSPORT_CLIENT).map(c -> c.createSSLEngine(hostname, port));
-            }
-        });
-    }
-
-    @Override
-    public Optional<SecureHttpTransportSettingsProvider> getSecureHttpTransportSettingsProvider(Settings settings) {
-        return Optional.of(new SecureHttpTransportSettingsProvider() {
             @Override
             public Collection<TransportAdapterProvider<HttpServerTransport>> getHttpTransportAdapterProviders(Settings settings) {
                 return List.of(new TransportAdapterProvider<HttpServerTransport>() {
--- a/src/test/java/org/opensearch/security/ssl/CertificatesRule.java
+++ b/src/test/java/org/opensearch/security/ssl/CertificatesRule.java
@@ -45,7 +45,7 @@ import org.bouncycastle.cert.X509CertificateHolder;
 import org.bouncycastle.cert.X509v3CertificateBuilder;
 import org.bouncycastle.cert.jcajce.JcaX509CertificateConverter;
 import org.bouncycastle.cert.jcajce.JcaX509ExtensionUtils;
-import org.bouncycastle.jce.provider.BouncyCastleProvider;
+import org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider;
 import org.bouncycastle.operator.OperatorCreationException;
 import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder;
 
@@ -53,7 +53,7 @@ import org.opensearch.common.collect.Tuple;
 
 public class CertificatesRule extends ExternalResource {
 
-    private final static BouncyCastleProvider BOUNCY_CASTLE_PROVIDER = new BouncyCastleProvider();
+    private final static BouncyCastleFipsProvider BOUNCY_CASTLE_FIPS_PROVIDER = new BouncyCastleFipsProvider();
 
     private final TemporaryFolder temporaryFolder = new TemporaryFolder();
 
@@ -128,7 +128,7 @@ public class CertificatesRule extends ExternalResource {
     }
 
     public KeyPair generateKeyPair() throws NoSuchAlgorithmException {
-        KeyPairGenerator generator = KeyPairGenerator.getInstance("RSA", BOUNCY_CASTLE_PROVIDER);
+        KeyPairGenerator generator = KeyPairGenerator.getInstance("RSA", BOUNCY_CASTLE_FIPS_PROVIDER);
         generator.initialize(4096);
         return generator.generateKeyPair();
     }
@@ -180,7 +180,7 @@ public class CertificatesRule extends ExternalResource {
             endDate
         ).addExtension(Extension.basicConstraints, true, new BasicConstraints(true))
             .addExtension(Extension.keyUsage, true, new KeyUsage(KeyUsage.digitalSignature | KeyUsage.keyCertSign | KeyUsage.cRLSign))
-            .build(new JcaContentSignerBuilder("SHA256withRSA").setProvider(BOUNCY_CASTLE_PROVIDER).build(parentKeyPair.getPrivate()));
+            .build(new JcaContentSignerBuilder("SHA256withRSA").setProvider(BOUNCY_CASTLE_FIPS_PROVIDER).build(parentKeyPair.getPrivate()));
         // CS-ENFORCE-SINGLE
     }
 
@@ -303,7 +303,7 @@ public class CertificatesRule extends ExternalResource {
             )
             .addExtension(Extension.extendedKeyUsage, true, new ExtendedKeyUsage(KeyPurposeId.id_kp_clientAuth))
             .addExtension(Extension.subjectAlternativeName, false, new DERSequence(sans.toArray(sans.toArray(new ASN1Encodable[0]))))
-            .build(new JcaContentSignerBuilder("SHA256withRSA").setProvider(BOUNCY_CASTLE_PROVIDER).build(parentKeyPair.getPrivate()));
+            .build(new JcaContentSignerBuilder("SHA256withRSA").setProvider(BOUNCY_CASTLE_FIPS_PROVIDER).build(parentKeyPair.getPrivate()));
         // CS-ENFORCE-SINGLE
         return Tuple.tuple(keyPair.getPrivate(), certificate);
     }


src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
```diff
@@ -0,0 +1,197 @@
...
+            }
+        } catch (Exception e) {
+            return ValidationResult.error(RestStatus.INTERNAL_SERVER_ERROR, payload(RestStatus.INTERNAL_SERVER_ERROR, e.getMessage()));
+        }
+    }
+
+    private ValidationResult<SecurityConfiguration> viewAllVersions() throws IOException {
+        SecurityConfigVersionDocument doc = versionsLoader.loadFullDocument();
+
+        return ValidationResult.error(OK, (builder, params) -> {
```
nagarajg17, line 99: Why error? Should it be `ValidationResult.success`?

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
```diff
@@ -0,0 +1,197 @@
...
+        builder.startArray("versions");
+        for (SecurityConfigVersionDocument.Version<?> ver : versions) {
+            builder.startObject();
+            builder.field("version_id", ver.getVersion_id());
+            builder.field("timestamp", ver.getTimestamp());
+            builder.field("modified_by", ver.getModified_by());
+            builder.field("security_configs");
+            Map<String, Object> plainConfigs = new LinkedHashMap<>();
+            for (Map.Entry<String, SecurityConfigVersionDocument.HistoricSecurityConfig<?>> entry : ver.getSecurity_configs().entrySet()) {
+                Map<String, Object> scMap = new LinkedHashMap<>();
```
nagarajg17: nit: Update variable name

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    private ValidationResult<SecurityConfiguration> rollbackToSpecificVersion(String versionId) throws IOException {
+        SecurityConfigVersionDocument doc = versionsLoader.loadFullDocument();
+
+        var maybeVer = doc.getVersions().stream().filter(v -> versionId.equals(v.getVersion_id())).findFirst();
+
+        if (maybeVer.isEmpty()) {
+            return ValidationResult.error(NOT_FOUND, payload(NOT_FOUND, "Version " + versionId + " not found"));
+        }
```
nagarajg17, line 163: Do we need this here? We seem to be doing in rollbackCommon() as well

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+                if (cType == null) {
+                    throw new IOException("Rollback aborted: Unknown config type '" + cTypeName + "' found in version");
+                }
+
+                SecurityDynamicConfiguration<?> currentConfig = currentConfigs.get(cType);
+                if (currentConfig == null) {
+                    throw new IOException("Rollback aborted: Could not fetch current config for cType '" + cTypeName + "'");
+                }
+
+                SecurityDynamicConfiguration<?> sdc = SecurityDynamicConfiguration.empty(cType);
```
nagarajg17: nit: rename or expand it

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+
+                if (sc == null || sc.getConfigData() == null) {
+                    log.warn("Skipping cType '{}' due to null configData", cTypeName);
+                    continue;
+                }
+
+                CType<?> cType = CType.fromString(cTypeName);
+                if (cType == null) {
+                    throw new IOException("Rollback aborted: Unknown config type '" + cTypeName + "' found in version");
+                }
```
nagarajg17: nit: This check needs to be done before checking data for that cType?

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+
+                SecurityDynamicConfiguration<?> currentConfig = currentConfigs.get(cType);
+                if (currentConfig == null) {
+                    throw new IOException("Rollback aborted: Could not fetch current config for cType '" + cTypeName + "'");
+                }
```
nagarajg17: Not sure if we need to have data for all cTypes always, check on this

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
nagarajg17: Also exception type `IOException` doesn't seem appropriate. `IllegalStateException`?
Check in other places as well

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+
+                SecurityDynamicConfiguration<?> sdc = SecurityDynamicConfiguration.empty(cType);
+                sdc.setSeqNo(currentConfig.getSeqNo());
+                sdc.setPrimaryTerm(currentConfig.getPrimaryTerm());
+
+                for (Map.Entry<String, ?> configEntry : sc.getConfigData().entrySet()) {
+                    if ("_meta".equals(configEntry.getKey())) {
```
nagarajg17, line 224: Why do we skip `_meta`? Good to add comments for these kinds for readers of code to understand

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+                            latch.countDown();
+                        }
+                    }
+                );
+                log.info("Rollback: wrote config data for cType={}", entry.getKey().toLCString());
+            }
+
+            if (!latch.await(20, java.util.concurrent.TimeUnit.SECONDS)) {
+                throw new IOException("Timeout while writing rolled-back configs");
+            }
```
nagarajg17: - Can we move this to another function?
- Timeout value can be constant

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+        }
+    }
+
+    private void revertRollbackOnFailure(Map<CType<?>, SecurityDynamicConfiguration<?>> backups, Exception originalException)
+        throws IOException {
+        log.error("Rollback failed mid-way. Reverting previous updates...", originalException);
+
+        for (Map.Entry<CType<?>, SecurityDynamicConfiguration<?>> entry : backups.entrySet()) {
+            try {
+                AbstractApiAction.saveAndUpdateConfigs(securityApiDependencies, client, entry.getKey(), entry.getValue()).actionGet();
```
nagarajg17: - High: This won't update cache
- Can we test for this as well

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+    private XContentBuilder buildRollbackResponseJson(String versionId) throws IOException {
+        XContentBuilder builder = XContentFactory.jsonBuilder().prettyPrint();
+        builder.startObject();
+        builder.field("status", "OK");
+        builder.field("message", "config rolled back to version " + versionId);
+        builder.endObject();
+        return builder;
+    }
+
+    private boolean isConfigEqual(SecurityDynamicConfiguration<?> currentConfig, SecurityDynamicConfiguration<?> targetConfig) {
```
nagarajg17: nit: This can be util function I think. Its reusable by any other code

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+                return builder;
+            });
+
+        } catch (Exception e) {
+            log.error("Rollback to version {} failed", versionId, e);
+            return ValidationResult.error(INTERNAL_SERVER_ERROR, payload(INTERNAL_SERVER_ERROR, "Rollback failed: " + e.getMessage()));
+        }
+    }
+
+    private void rollbackConfigsToSecurityIndex(SecurityConfigVersionDocument.Version<?> versionData) throws IOException {
```
nagarajg17, line 187: Variable names are not clear. Reading is bit tough here. Need to go up and down to know what variable is. 
Rename such that names give meaning of that it is intended for. For ex: currentConfig used here is a good naming, easy to understand

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+    private ValidationResult<SecurityConfiguration> rollbackToSpecificVersion(String versionId) throws IOException {
+        SecurityConfigVersionDocument doc = versionsLoader.loadFullDocument();
+
+        var maybeVer = doc.getVersions().stream().filter(v -> versionId.equals(v.getVersion_id())).findFirst();
+
+        if (maybeVer.isEmpty()) {
+            return ValidationResult.error(NOT_FOUND, payload(NOT_FOUND, "Version " + versionId + " not found"));
+        }
+
+        SecurityConfigVersionsLoader.sortVersionsById(doc.getVersions());
```
nagarajg17: Do we need sorting here?

src/main/java/org/opensearch/security/dlic/rest/api/Endpoint.java
```diff
@@ -29,5 +29,7 @@ public enum Endpoint {
     VALIDATE,
     ALLOWLIST,
     NODESDN,
-    SSL;
+    SSL,
+    VIEW_VERSION,
+    ROLLBACK_VERSION;
```
DarshitChanpura, line 34: These two should be grouped as version? A singular permission that allows rest-admin privileges to the view and rollback. Similar to other endpoints already present here.

src/integrationTest/java/org/opensearch/security/api/RollbackVersionApiIntegrationTest.java
```diff
@@ -0,0 +1,165 @@
...
+import static org.opensearch.security.OpenSearchSecurityPlugin.PLUGINS_PREFIX;
+import static org.opensearch.security.support.ConfigConstants.EXPERIMENTAL_SECURITY_CONFIGURATIONS_VERSIONS_ENABLED;
+
+public class RollbackVersionApiIntegrationTest extends AbstractApiIntegrationTest {
+
+    private String endpointPrefix() {
+        return PLUGINS_PREFIX + "/api";
+    }
+
+    private String RollbackBase() {
```
cwperks: If these are all constants can we declare them as constant fields at the top of this file?

src/main/java/org/opensearch/security/dlic/rest/api/RestApiAdminPrivilegesEvaluator.java
```diff
@@ -71,6 +71,8 @@ default String build() {
         .put(Endpoint.ROLESMAPPING, action -> buildEndpointPermission(Endpoint.ROLESMAPPING))
         .put(Endpoint.TENANTS, action -> buildEndpointPermission(Endpoint.TENANTS))
         .put(Endpoint.SSL, action -> buildEndpointActionPermission(Endpoint.SSL, action))
+        .put(Endpoint.VIEW_VERSION, action -> buildEndpointPermission(Endpoint.VIEW_VERSION))
+        .put(Endpoint.ROLLBACK_VERSION, action -> buildEndpointPermission(Endpoint.ROLLBACK_VERSION))
```
DarshitChanpura, line 75: similar to the comment above, this should be clubbed as one permission.

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
```diff
@@ -0,0 +1,197 @@
...
+ * REST endpoint:
+ *   GET /_plugins/_security/api/version
+ *   GET /_plugins/_security/api/version/{versionId}
+ */
+public class ViewVersionApiAction extends AbstractApiAction {
+
+    private static final Logger LOGGER = LogManager.getLogger(ViewVersionApiAction.class);
+
+    private static final List<Route> routes = addRoutesPrefix(
+        ImmutableList.of(new Route(Method.GET, "/version"), new Route(Method.GET, "/version/{versionID}"))
```
DarshitChanpura: get alll versions endpoint should be changed to `/versions`

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
DarshitChanpura: +1. Should be ValidationResult.success here.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+ * REST endpoint:
+ *   POST /_plugins/_security/api/rollback
+ *   POST /_plugins/_security/api/rollback/version/{versionID}
+ */
+public class RollbackVersionApiAction extends AbstractApiAction {
+
+    private static final Logger log = LogManager.getLogger(RollbackVersionApiAction.class);
+
+    private static final List<Route> routes = addRoutesPrefix(
+        ImmutableList.of(new Route(POST, "/rollback"), new Route(POST, "/rollback/version/{versionID}"))
```
DarshitChanpura: should be renamed to `/version/rollback` and `/version/rollback/{versionId}`

cwperks: By passing `versionID` in does this effectively support moving forward as well?

cwperks: Does it make sense to introduce a `dryRun` url param that would return the patch applied as its output without actually applying the patch?

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+
+        this.versionsLoader = versionsLoader;
+        this.configRepository = configRepository;
+        this.client = client;
+
+        this.requestHandlersBuilder.add(RestRequest.Method.GET, methodNotImplementedHandler)
+            .add(RestRequest.Method.PATCH, methodNotImplementedHandler)
+            .add(RestRequest.Method.PUT, methodNotImplementedHandler)
+            .add(RestRequest.Method.DELETE, methodNotImplementedHandler)
+            .add(POST, (channel, request, unusedclient) -> {
```
DarshitChanpura: use `requestHandlersBuilder.allMethodsNotImplemented().override(` here

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+                ValidationResult<SecurityConfiguration> result = handlePostRequest(request);
+                result.valid(securityConfiguration -> {
+                    try {
+                        XContentBuilder builder = XContentFactory.jsonBuilder().prettyPrint();
+                        builder.startObject();
+                        builder.field("status", "OK");
+                        builder.field("message", "config rolled back to version " + request.param("versionID", "previous"));
+                        builder.endObject();
+                        channel.sendResponse(new BytesRestResponse(OK, builder));
```
DarshitChanpura: Use methods from [Responses.java](https://github.com/opensearch-project/security/blob/af358d0e41b9eb28b9cadc6eb7679e83bb083b84/src/main/java/org/opensearch/security/dlic/rest/api/Responses.java) class here to send response. That way we can avoid duplicating this logic. This applies to both APIs.

example: 
```suggestion
                        ok(channel, "config rolled back to version " + request.param("versionID", "previous"));
```

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+                XContentBuilder inner = buildRollbackResponseJson(versionId);
+                builder.copyCurrentStructure(JsonXContent.jsonXContent.createParser(null, null, BytesReference.bytes(inner).streamInput()));
+                return builder;
+            });
+
+        } catch (Exception e) {
+            log.error("Rollback to version {} failed", versionId, e);
+            return ValidationResult.error(INTERNAL_SERVER_ERROR, payload(INTERNAL_SERVER_ERROR, "Rollback failed: " + e.getMessage()));
+        }
+    }
```
DarshitChanpura:
```suggestion
private ValidationResult<SecurityConfiguration> handlePostRequest(RestRequest request) throws IOException {
	        String versionParam = request.param("versionID");
	        try {
	            rollBack(versionParam);
	        } catch (Exception e) {
	            log.error("Rollback request failed", e);
	            return ValidationResult.error(INTERNAL_SERVER_ERROR, payload(INTERNAL_SERVER_ERROR, e.getMessage()));
	        }
	    }
	
	    private ValidationResult<SecurityConfiguration> rollBack(String rollBackVersionId) throws IOException {
	        SecurityConfigVersionDocument doc = versionsLoader.loadFullDocument();
	
	        if (doc.getVersions().size() < 2) {
	            return ValidationResult.error(NOT_FOUND, payload(NOT_FOUND, "No previous version available to rollback"));
	        }
	        SecurityConfigVersionsLoader.sortVersionsById(doc.getVersions());
	
	        var docVersions = doc.getVersions();
	
	        if (rollBackVersionId != null) {
	            rollBackVersionId = doc.getVersions().stream().filter(v -> rollBackVersionId.equals(v.getVersion_id())).findFirst();
	            if (rollBackVersionId.isEmpty()) {
	                return ValidationResult.error(NOT_FOUND, payload(NOT_FOUND, "Version " + rollBackVersionId + " not found"));
	            }
	        } else {
	            rollBackVersionId = docVersions.get(docVersions.size() - 2).getVersion_id();
	        }
	       
	        try {
	            rollbackConfigsToSecurityIndex(rollBackVersionI);
	
	            ThreadContext threadContext = threadPool.getThreadContext();
	
	            EventBus.getDefault()
	                .register(
	                    new SecurityConfigVersionHandler(configRepository, clusterService.getSettings(), threadContext, threadPool, client)
	                );
	
	            EventBus.getDefault().post(new SecurityConfigChangeEvent());
	
	            return ValidationResult.error(OK, (builder, params) -> {
	                XContentBuilder inner = buildRollbackResponseJson(versionId);
	                builder.copyCurrentStructure(JsonXContent.jsonXContent.createParser(null, null, BytesReference.bytes(inner).streamInput()));
	                return builder;
	            });
	
	        } catch (Exception e) {
	            log.error("Rollback to version {} failed", versionId, e);
	            return ValidationResult.error(INTERNAL_SERVER_ERROR, payload(INTERNAL_SERVER_ERROR, "Rollback failed: " + e.getMessage()));
	        }
	    }

```

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+
+    private void rollbackConfigsToSecurityIndex(SecurityConfigVersionDocument.Version<?> versionData) throws IOException {
+        Map<String, SecurityConfigVersionDocument.HistoricSecurityConfig<?>> securityConfigs = versionData.getSecurity_configs();
+        if (securityConfigs == null || securityConfigs.isEmpty()) {
+            throw new IOException("No security configs to rollback in version " + versionData.getVersion_id());
```
DarshitChanpura: why are we throwing IOException here? This should be NotFoundException and response should be 400 Bad Request.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+        }
+
+        Map<CType<?>, SecurityDynamicConfiguration<?>> configsToApply = new java.util.HashMap<>();
```
DarshitChanpura: no inline imports.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
DarshitChanpura: this should never be null since we are responsible for storing the information. A Ctype will be unknown only when security plugin moves away from current CTypes. Also this should throw NPE since we use cTypes as keys. Another option would be NoSuchFieldException.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
DarshitChanpura: This may be null when no rollback version exist, i.e. either previous version was deleted or it never existed. In such cases, we should return 400 Bad Request along with probly throwing an IllegalArgumentException.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
DarshitChanpura: since everything is in context of security you can name the variables as `updateConfig` or `rollBackConfig` since this indicates a new entry to be created from existing entry.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
DarshitChanpura: +1.

What is the expected behavior when config version changes? i.e. roles change from v7 to v8.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
DarshitChanpura: Magic variables should be declared as class level constants.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,368 @@
...
+        Map<CType<?>, SecurityDynamicConfiguration<?>> backups = new java.util.HashMap<>();
```
DarshitChanpura: Not a blocker but you can declare something like:

// map of CTypes to a Pair<updatedConfig, oldConfig>
```java
Map<CType<?>, Pair<SecurityDynamicConfiguration<?>, SecurityDynamicConfiguration<?>>> configUpdateMap = new HashMap<>();
```

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
DarshitChanpura: Have we tested the sync vs async behavior here?

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
```diff
@@ -0,0 +1,197 @@
...
+    ) {
+        super(Endpoint.VIEW_VERSION, clusterService, threadPool, securityApiDependencies);
+        this.versionsLoader = versionsLoader;
+
+        this.requestHandlersBuilder.add(RestRequest.Method.POST, RequestHandler.methodNotImplementedHandler)
+            .add(RestRequest.Method.PATCH, RequestHandler.methodNotImplementedHandler)
+            .add(RestRequest.Method.PUT, RequestHandler.methodNotImplementedHandler)
+            .add(RestRequest.Method.DELETE, RequestHandler.methodNotImplementedHandler)
+
+            .onGetRequest((restRequest) -> {
```
DarshitChanpura: similar to comment in rollback action class, this can be simplified to:
```java
 requestHandlersBuilder.allMethodsNotImplemented().onGetRequest(
```

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    private ValidationResult<SecurityConfiguration> viewSpecificVersion(String versionId) throws IOException {
+        SecurityConfigVersionDocument doc = versionsLoader.loadFullDocument();
+
+        var maybeVer = doc.getVersions().stream().filter(v -> versionId.equals(v.getVersion_id())).findFirst();
+
+        if (maybeVer.isEmpty()) {
+            return ValidationResult.error(NOT_FOUND, payload(NOT_FOUND, "Version " + versionId + " not found"));
+        }
```
DarshitChanpura, line 113: this method is pretty much identical with the one above, see if you can club it together to re-use code.

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
```diff
@@ -0,0 +1,197 @@
...
+
+        return ValidationResult.error(OK, (builder, params) -> {
+            XContentBuilder inner = buildVersionsJsonBuilder(List.of(maybeVer.get()));
+            builder.copyCurrentStructure(JsonXContent.jsonXContent.createParser(null, null, BytesReference.bytes(inner).streamInput()));
```
DarshitChanpura, line 117: is see this pattern being repeated, 2 things:
1. Can you use fromXContent and toXContent instead of buildVersionsJsonBuilder and copyCurrentStructure.
2. See if you can over-load implementation of **payload** method in Responses.java that accepts JsonContent.

src/test/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiActionValidationTest.java
```diff
@@ -0,0 +1,78 @@
...
+            versionsLoader,
+            configurationRepository,
+            client
+        );
+    }
+
+    @Test
+    public void testOnConfigDelete_isForbidden() throws IOException {
+        var result = rollbackVersionApiAction.createEndpointValidator().onConfigDelete(SecurityConfiguration.of(null, configuration));
+        assertThat(result.status(), is(org.opensearch.core.rest.RestStatus.FORBIDDEN));
```
DarshitChanpura: no inline imports. I see this pattern at other places as well that should be addressed.

src/test/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiTest.java
```diff
@@ -0,0 +1,174 @@
...
+            "/.opensearch_security_config_versions/_doc/opensearch_security_config_versions",
+            docPayload
+        );
+
+        rh.executePostRequest("/.opensearch_security_config_versions/_refresh", "");
+
+        assertThat("Failed to insert config versions doc", response.getStatusCode(), is(oneOf(200, 201)));
+    }
+
+    protected String getEndpointPrefix() {
```
DarshitChanpura: why not use PLUGIN_PREFIX directly instead of declaring this method.

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack, doing it once.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack!

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack!

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack!

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Including `_meta` creates a duplicate entry for the field.

divyanshi-0402: Currently we have rollback functionality for same versions of config types.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Yes, async updates cache and triggers update for version index as well, sync doesn't.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Not required, removed.

src/integrationTest/java/org/opensearch/security/api/RollbackVersionApiIntegrationTest.java
divyanshi-0402: Ack!

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: It's essentially rollback, providing `versionID` allows users to rollback to any particular version. 
Yes, we are separately working on a dryRun API for previewing the patch applied and it's impact before actually applying it.

src/main/java/org/opensearch/security/dlic/rest/api/Endpoint.java
divyanshi-0402: View API is just GET request, while rollback API is a POST request, hence kept them separate. But will look into this.

src/main/java/org/opensearch/security/dlic/rest/api/RestApiAdminPrivilegesEvaluator.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack will check this out.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
divyanshi-0402: Ack will check this out.

src/test/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiActionValidationTest.java
divyanshi-0402: Ack

src/test/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiTest.java
divyanshi-0402: Ack

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
divyanshi-0402: ValidationResult.success only takes a SecurityConfiguration as content object, so in order to build a custom JSON I had to use this.

DarshitChanpura: .error() is not the correct usage. You can simply create response class that extends ToXContentObject and then return an instance of that object here.
You can refer this file: https://github.com/opensearch-project/security/pull/5389/files#diff-157c9f69cf6c49e618db3748a7803470432b527ce78fe5251228d093daa5483e

src/main/java/org/opensearch/security/dlic/rest/api/Endpoint.java
nagarajg17: Just version may not sound similar to for ex: say roles because GET `/role/{}`, PUT `/role/{}` is intuitive to understand but GET `/version/{}`, PUT `/version/{}` is not that intuitive. PUT/POST `/version/{}` is not updating its entity, here it is doing rolling back which is not intuitive to understand

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
nagarajg17: We cannot use `ValidationResult.error`, it is very misleading and wrong pattern

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
```diff
@@ -0,0 +1,192 @@
...
+            XContentBuilder inner = buildVersionsJsonBuilder(doc.getVersions());
+            builder.copyCurrentStructure(JsonXContent.jsonXContent.createParser(null, null, BytesReference.bytes(inner).streamInput()));
+            return builder;
+        });
+    }
+
+    private ValidationResult<SecurityConfiguration> viewSpecificVersion(String versionId) throws IOException {
+        SecurityConfigVersionDocument doc = versionsLoader.loadFullDocument();
+
+        var maybeVer = doc.getVersions().stream().filter(v -> versionId.equals(v.getVersion_id())).findFirst();
```
nagarajg17, line 109: nit: what does maybeVer mean?

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
nagarajg17: This might make more sense to keep in rollbackCommon() since you would need to check this anyway for any caller of this function

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
nagarajg17: > Including _meta creates a duplicate entry for the field

Why? which other code is adding `_meta` field?

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
nagarajg17: Resolved - doing only once now

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
nagarajg17: Resolved - variable names are updated

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
nagarajg17: Created new generic method to handle json content. Closing this thread 
```
public RequestHandlersBuilder onJsonContentGetRequest(
                final CheckedFunction<RestRequest, ValidationResult<ToXContent>, IOException> mapper
        ) {
            Objects.requireNonNull(mapper, "onGetRequest request handler can't be null");
            add(
                    RestRequest.Method.GET,
                    (channel, request, client) -> mapper.apply(request)
                            .valid(toXContent -> ok(channel, toXContent))
                            .error((status, toXContent) -> response(channel, status, toXContent))
            );
            return this;
        }
```

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
nagarajg17: Resolved - renamed

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
nagarajg17: Resolved - Created new method in Responses.java to handle this

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
nagarajg17: Resolved - Have moved common code parts to caller function. Its fine to be 2 separate functions now

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
nagarajg17: Resolved - Overwriting instead of excluding

nagarajg17:
We can discard this PR. This PR changes was carried forward in this PR and closed
https://github.com/opensearch-project/security/pull/5641

Issue #4590, opened 2024-07-23
[FEATURE] Add support for configuring BCrypt password hashing


dancristiancecoi:
**What solution would you like?**
Currently password hashing with BCrypt is not configurable.  It will be useful for cluster administrators to be able to configure properties such as:

- number of rounds
- version of the BCrypt algorithm

**Do you have any additional context?**
The work for this was completed as part of https://github.com/opensearch-project/security/pull/4524. Raising this issue so we can tag it for 2.16

scrawfor99:
[Triage] Hi @dancristiancecoi, thanks for filing this issue. This sounds like a good addition to the existing BCrypt support. Will mark as triaged.

PR #5441, opened 2025-06-25
Adds Argon2 support for password hashing

### LINKED ISSUE #4592: [FEATURE] Add support for Argon2 password hashing algorithm
It will be useful for some cluster administrators to be able to use Argon2 as the password hashing algorithm and be able to configure properties such as:

- memory
- iterations
- length
- parallelism
- type
- version

aidenlindsay:
### Description
This change adds support for Argon2 as another alternative password hashing algorithm. Which builds on the added support for configuration of parameters, by building support for the following parameters:

## Argon2:
- Memory (defaults to 64MiB/65536)
- Iterations (defaults to 3)
- Length (defaults to 32)
- Parallelism (defaults to 1)
- Type (defaults to Argon2ID)
- Version (defaults to 19)

This includes adding the Argon2PasswordHasher class as well as amending all current classes that use the hashing algorithms to support the new algorithm (PasswordHasherFactory.java, Hasher.java, test scripts, etc.)

Previously there was only support for BCrypt and PBKDF2 with BCrypt being the default, this allows the option to use another alternative algorithm in password hashing

### Issues Resolved
#4592 
### Related Issues
#4590 
#4524 

### Testing
Added both integration tests and unit tests to cover the new code as well as some manual testing. 

### Check List


src/test/java/org/opensearch/security/hasher/Argon2PasswordHasherTests.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Test
+    public void shouldGenerateValidHashesFromParameters() {
+        PasswordHasher hasher = new Argon2PasswordHasher(47104, 1, 2, 32, "argon2id", 19);
+        String hash = hasher.hash(password.toCharArray());
+        assertThat(hasher.check(password.toCharArray(), hash), is(true));
+        assertThat(hasher.check(wrongPassword.toCharArray(), hash), is(false));
+
+        hasher = new Argon2PasswordHasher(19456, 2, 1, 64, "argon2i", 19);
```
cwperks: This looks like a good contender for using the `@Parameterized` pattern to pass in different parameters for a list of test scenarios. See https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/auditlog/impl/AuditCategoryTest.java#L33-L79

This is also fine as is, but I wanted to point out that this repo has examples of passing a list of parameters to run for the same set of assertions.

src/main/java/org/opensearch/security/hasher/Argon2PasswordHasher.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Override
+    @SuppressWarnings("removal")
+    public boolean check(char[] password, String hash) {
+        checkPasswordNotNullOrEmpty(password);
+        checkHashNotNullOrEmpty(hash);
+        CharBuffer passwordBuffer = CharBuffer.wrap(password);
+        try {
+            SecurityManager securityManager = System.getSecurityManager();
```
cwperks: Let's remove all JSM related code here. This will always be null in 3.0.0.

Technically you can leave the AccessController.doPrivileged in but I think its unnecessary now.

src/main/java/org/opensearch/security/hasher/Argon2PasswordHasher.java
```diff
@@ -0,0 +1,121 @@
...
+        );
+    }
+
+    @Override
+    @SuppressWarnings("removal")
+    public String hash(char[] password) {
+        checkPasswordNotNullOrEmpty(password);
+        CharBuffer passwordBuffer = CharBuffer.wrap(password);
+        try {
+            SecurityManager securityManager = System.getSecurityManager();
```
cwperks: same here

src/main/java/org/opensearch/security/hasher/Argon2PasswordHasher.java
```diff
@@ -0,0 +1,121 @@
...
+    @SuppressWarnings("removal")
+    Argon2PasswordHasher(int memory, int iterations, int parallelism, int length, String type, int version) {
+        this.iterations = iterations;
+        this.memory = memory;
+        this.parallelization = parallelism;
+        this.length = length;
+        this.typeArgon2 = parseType(type);
+        this.version = version;
+
+        SecurityManager securityManager = System.getSecurityManager();
```
cwperks: same here

cwperks (commented):
Thank you @aidenlindsay ! The PR mostly looks good to me. Please address the JSM related comments and fix the integration tests.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1436,6 +1436,55 @@ public List<Setting<?>> getSettings() {
...
+            Setting.simpleString(
+                ConfigConstants.SECURITY_PASSWORD_HASHING_ARGON2_TYPE,
+                ConfigConstants.SECURITY_PASSWORD_HASHING_ARGON2_TYPE_DEFAULT,
+                Property.NodeScope,
+                Property.Final
+            )
+        );
+        settings.add(
+            Setting.intSetting(
+                ConfigConstants.SECURITY_PASSWORD_HASHING_ARGON2_VERSION,
```
DarshitChanpura, line 1482: out of curiosity, should we allow configuring version?

src/test/java/org/opensearch/security/auditlog/compliance/RestApiComplianceAuditlogTest.java
```diff
@@ -377,6 +377,31 @@ public void testPBKDF2HashRedaction() {
...
+        rh.sendAdminCertificate = true;
+        rh.keystore = "kirk-keystore.jks";
+
+        // read internal users and verify no Argon2 hash is present in audit logs
+        final AuditMessage message1 = TestAuditlogImpl.doThenWaitForMessage(() -> {
+            rh.executeGetRequest("/_opendistro/_security/api/internalusers");
+        });
+
+        Assert.assertFalse(message1.toString().contains("$argon2id$"));
+        Assert.assertTrue(message1.toString().contains("__HASH__"));
```
DarshitChanpura, line 402: is `__HASH__` a standard redaction string? If it's possible can we check exact instead of contains, that way we will not need assertion on line 401

src/test/java/org/opensearch/security/hasher/Argon2PasswordHasherTests.java
```diff
@@ -0,0 +1,79 @@
...
+            ConfigConstants.SECURITY_PASSWORD_HASHING_ARGON2_ITERATIONS_DEFAULT,
+            ConfigConstants.SECURITY_PASSWORD_HASHING_ARGON2_PARALLELISM_DEFAULT,
+            ConfigConstants.SECURITY_PASSWORD_HASHING_ARGON2_LENGTH_DEFAULT,
+            ConfigConstants.SECURITY_PASSWORD_HASHING_ARGON2_TYPE_DEFAULT,
+            ConfigConstants.SECURITY_PASSWORD_HASHING_ARGON2_VERSION_DEFAULT
+        );
+    }
+
+    @Test
+    public void shouldGenerateValidHashesFromParameters() {
```
DarshitChanpura, line 74: can we add a test for passing invalid hash? and then checking

src/test/java/org/opensearch/security/tools/HasherTests.java
```diff
@@ -142,4 +158,179 @@ public void testWithPBKDF2AllArguments() {
         assertEquals("should return a valid PBKDF2 hash with the default value for \"iterations\"", pbkdf2Function.getIterations(), 150000);
         assertEquals("should return a valid PBKDF2 hash with the default value for \"length\"", pbkdf2Function.getLength(), 250);
     }
+
+    // ARGON2
+    @Test
+    public void testWithArgon2DefaultArguments() {
+        Hasher.main(new String[] { "-p", "password", "-a", "Argon2" });
+        Argon2Function argon2Function = Argon2Function.getInstanceFromHash(out.toString().trim());
+        assertEquals("should return a valid Argon2 hash with the default value for \"memory\"", argon2Function.getMemory(), 65536);
```
DarshitChanpura: instead of magic numbers here, we should use default values from ConfigConstants file

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Features
 
 * Introduced new experimental versioned security configuration management feature ([#5357] (https://github.com/opensearch-project/security/pull/5357))
+* Introduced support for the Argon2 Password Hashing Algorithm ([#5441] (https://github.com/opensearch-project/security/pull/5441))
```
DarshitChanpura: nit: present tense. 
We probs need to go and update the changelog for other entries on our end.

src/main/java/org/opensearch/security/hasher/Argon2PasswordHasher.java
aidenlindsay: Okay I have fixed these, and at the same time will go back and make the same changes to the other password hashing algorithms that use the same legacy JSM code

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
aidenlindsay: @DarshitChanpura I am happy to remove this, but it was one of the parameters referenced in the original issue, and in online documentation the version is taken into account. Tbf I don't see a reason why v16 would ever be used over v19 so let me know and I can remove this

src/test/java/org/opensearch/security/tools/HasherTests.java
aidenlindsay: Yea I was just using the values copied from the ConfigConstants file as the defaults, but yes I don't know why I was hardcoding this... fixed now

src/test/java/org/opensearch/security/auditlog/compliance/RestApiComplianceAuditlogTest.java
aidenlindsay: For the 401 assert I can be more specific with the hash data instead of just saying .contains, but I don't really know what you mean by removing it? __HASH__ is not a standard but it does seem to be the convention used in the security plugin for other algorithms to indicate the hash has been redacted. I think that for its use case, checking that this appears at all is sufficient to ensure hashes are being redacted as intended, but I'm open to other suggestions

aidenlindsay:
will fix the test failures after lunch

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: I'm okay with keeping this. I see that you've added checks to only allow 16 and 19

src/test/java/org/opensearch/security/auditlog/compliance/RestApiComplianceAuditlogTest.java
DarshitChanpura: I see. I meant to say whether we should compare with equalTo instead of contains.

DarshitChanpura (commented):
The changes look good to me. I'll wait for CI to be green before my final review.

src/test/java/org/opensearch/security/auditlog/compliance/RestApiComplianceAuditlogTest.java
aidenlindsay: I think in practice audit messages are usually JSON objects or logs, and so they will contain more than just the string __HASH__, and therefore it would be unrealistic to check it is exactly equal, and more helpful to check that the term appears to show redaction

aidenlindsay:
Added the relevant documentation for this PR:

https://github.com/opensearch-project/documentation-website/pull/10163

cwperks (approved):

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1437,6 +1437,55 @@ public List<Setting<?>> getSettings() {
...
+        settings.add(
+            Setting.intSetting(
+                ConfigConstants.SECURITY_PASSWORD_HASHING_ARGON2_ITERATIONS,
+                ConfigConstants.SECURITY_PASSWORD_HASHING_ARGON2_ITERATIONS_DEFAULT,
+                Property.NodeScope,
+                Property.Final
+            )
+        );
+        settings.add(
+            Setting.intSetting(
```
cwperks, line 1449: FYI you can also specify min for intSettings.

See https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/SecuritySettings.java#L31-L37

I see that there is separate validation logic further in the PR, but it can be directly done in the declaration in the setting.

cwperks:
@DarshitChanpura @willyborankin @shikharj05 Can this get another maintainer to review?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
willyborankin: +1

willyborankin (approved):

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
cwperks: Marking as resolved since there is validation logic separately in this PR.

PR #5459, opened 2025-07-03
[Resource Sharing] Adds a Share API to fetch and update sharing information

DarshitChanpura:
### Description

Adds a new `/share` API to support sharing of resources by dashboard component. It supports three methods:
1. GET `/_plugins/_security/api/resource/share` - requires two query params `?resource_id=<id>&resource_type=<name>` to be present.
    1. fetches the current sharing information for the requested resource
2. PUT `/_plugins/_security/api/resource/share` 
```json
{
  "resource_id": "%s",
  "resource_type": "%s",
  "share_with": {
    "%s" : {
        "users": ["%s"]
    }
  }
}
```
   1. Creates or update (overrides) any existing sharing information.
3. PATCH `/_plugins/_security/api/resource/share`
```json
{
  "resource_id": "%s",
  "resource_type": "%s",
  "add": {
    %s
  },
  "revoke": {
    %s
  }
}
``` 
  1. Patches existing sharing info:
     1. `add` block contains information of entities and levels at which to share
     2. `revoke` block contains information of entities and levels at which to revoke

The structure for PUT and PATCH request content' sharing block must look like:
```json
 {
    "READ_ONLY": {
      "users": ["user1", "user2"],
      "roles": ["viewer_role"],
    },
    "READ_WRITE": {
      "users": ["admin_user"],
      "roles": ["editor_role"],
    }
  }
```

* Category - Enhancement
   *  Sets up base for future dashboard component to directly invoke share apis when loading and updating sharing info.

### Issues Resolved
TBD

TBD

### Testing
Integration testing

### Check List

### DIFF
--- a/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/TestUtils.java
+++ b/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/TestUtils.java
@@ -212,7 +212,7 @@ public final class TestUtils {
         return """
             {
               "resource_id": "%s",
-              "resource_index": "%s",
+              "resource_type": "%s",
               "share_with": {
                 "%s" : {
                     "users": ["%s"]
@@ -280,7 +280,7 @@ public final class TestUtils {
             return """
                 {
                   "resource_id": "%s",
-                  "resource_index": "%s",
+                  "resource_type": "%s",
                   "patch":{
                       "add": {
                         %s
--- a/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/TestUtils.java
+++ b/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/TestUtils.java
@@ -281,13 +281,11 @@ public final class TestUtils {
                 {
                   "resource_id": "%s",
                   "resource_type": "%s",
-                  "patch":{
-                      "add": {
-                        %s
-                      },
-                      "revoke": {
-                        %s
-                      }
+                  "add": {
+                    %s
+                  },
+                  "revoke": {
+                    %s
                   }
                 }
                 """.formatted(resourceId, resourceIndex, allShares, allRevokes);
--- a/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/TestUtils.java
+++ b/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/TestUtils.java
@@ -8,7 +8,10 @@
 
 package org.opensearch.sample.resource;
 
+import java.io.IOException;
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -16,10 +19,14 @@ import org.apache.http.HttpStatus;
 import org.awaitility.Awaitility;
 
 import org.opensearch.Version;
+import org.opensearch.common.xcontent.XContentFactory;
+import org.opensearch.core.xcontent.ToXContent;
+import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.painless.PainlessModulePlugin;
 import org.opensearch.plugins.PluginInfo;
 import org.opensearch.sample.SampleResourcePlugin;
 import org.opensearch.security.OpenSearchSecurityPlugin;
+import org.opensearch.security.spi.resources.sharing.Recipients;
 import org.opensearch.test.framework.TestSecurityConfig;
 import org.opensearch.test.framework.certificate.CertificateData;
 import org.opensearch.test.framework.cluster.ClusterManager;
@@ -72,7 +79,8 @@ public final class TestUtils {
         "sample_plugin_index_all_access",
         TestSecurityConfig.ActionGroup.Type.INDEX,
         "indices:*",
-        "cluster:admin/sample-resource-plugin/*"
+        "cluster:admin/sample-resource-plugin/*",
+        "cluster:admin/security/resource/share"
     );
 
     public static final String SAMPLE_RESOURCE_CREATE_ENDPOINT = SAMPLE_RESOURCE_PLUGIN_PREFIX + "/create";
@@ -83,6 +91,7 @@ public final class TestUtils {
     public static final String SAMPLE_RESOURCE_REVOKE_ENDPOINT = SAMPLE_RESOURCE_PLUGIN_PREFIX + "/revoke";
 
     static final String RESOURCE_SHARING_MIGRATION_ENDPOINT = "_plugins/_security/api/resources/migrate";
+    static final String SECURITY_SHARE_ENDPOINT = "_plugins/_security/api/resource/share";
 
     public static LocalCluster newCluster(boolean featureEnabled, boolean systemIndexEnabled) {
         return new LocalCluster.Builder().clusterManager(ClusterManager.THREE_CLUSTER_MANAGERS_COORDINATOR)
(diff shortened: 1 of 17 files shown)


DarshitChanpura:
This PR has changes from #5408 and hence that PR must be reviewed prior to reviewing this one.

cwperks:
In the PATCH version of the API can we replace `share_with` with `add`?

src/main/java/org/opensearch/security/resources/api/share/ShareRequest.java
```diff
@@ -0,0 +1,219 @@
...
+            throw arv;
+        }
+
+        // no further check needed in case of GET
+        if (method == RestRequest.Method.GET) {
+            return null;
+        }
+        // either of shareWith or patch must be present in the request
+        if (shareWith == null && (patch == null || patch.isEmpty())) {
+            String message = method == RestRequest.Method.PATCH ? "Patch" : "ShareWith";
```
cwperks: These should correspond to the actual keys.

src/main/java/org/opensearch/security/resources/api/share/ShareRequest.java
```diff
@@ -0,0 +1,219 @@
...
+            arv.addValidationError(message + " is required");
+            throw arv;
+        }
```
cwperks, line 97: if `patch` is present do you then also verify that the map contains `add` and `revoke` somewhere?

src/main/java/org/opensearch/security/resources/api/share/ShareRestAction.java
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * This class implements the share REST API for resource access management.
+ * It provides endpoints for sharing a resource.
+ *
+ */
+public class ShareRestAction extends BaseRestHandler {
+    private static final Logger LOGGER = LogManager.getLogger(ShareRestAction.class);
+
+    private final static Set<String> allowedPatchOperations = Set.of("share_with", "revoke");
```
cwperks: Use UPPER_CAMEL_CASE for constants.

src/main/java/org/opensearch/security/resources/api/share/ShareRestAction.java
```diff
@@ -0,0 +1,122 @@
...
+            client.executeLocally(ShareAction.INSTANCE, sharingInfoUpdateRequest, new ActionListener<>() {
+
+                @Override
+                public void onResponse(ShareResponse response) {
+                    ok(channel, response::toXContent);
+                }
+
+                @Override
+                public void onFailure(Exception e) {
+                    LOGGER.debug(e.getMessage(), e);
```
cwperks: I think this can be a higher level than debug.

spi/src/main/java/org/opensearch/security/spi/resources/sharing/Recipients.java
```diff
@@ -60,16 +60,22 @@ public Map<Recipient, Set<String>> getRecipients() {
     public void share(Recipients target) {
         Map<Recipient, Set<String>> targetRecipients = target.getRecipients();
         for (Recipient recipientType : targetRecipients.keySet()) {
-            Set<String> updatedRecipients = recipients.get(recipientType);
-            updatedRecipients.addAll(targetRecipients.get(recipientType));
+            if (!recipients.containsKey(recipientType)) { // will not be present in case of very first share
```
cwperks: nit: This can be simplified to:

```
recipients
    .computeIfAbsent(recipientType, k -> new HashSet<>())
    .addAll(targetRecipients.getOrDefault(recipientType, Collections.emptySet()));
```

spi/src/main/java/org/opensearch/security/spi/resources/sharing/Recipients.java
```diff
@@ -60,16 +60,22 @@ public Map<Recipient, Set<String>> getRecipients() {
...
+            }
         }
     }
 
     public void revoke(Recipients target) {
         Map<Recipient, Set<String>> targetRecipients = target.getRecipients();
         for (Recipient recipientType : targetRecipients.keySet()) {
-            Set<String> updatedRecipients = recipients.get(recipientType);
-            updatedRecipients.removeAll(targetRecipients.get(recipientType));
+            if (recipients.containsKey(recipientType)) { // no need to revoke if access was not granted in first place
```
cwperks: likewise we could use computeIfPresent here:

```
recipients.computeIfPresent(recipientType, (k, s) -> {
    s.removeAll(targetRecipients.getOrDefault(recipientType, Collections.emptySet()));
    return s;
});
```

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -567,6 +573,151 @@ public void revoke(String resourceId, String resourceIndex, ShareWith revokeAcce
...
+        String resourceSharingIndex = getSharingIndex(resourceIndex);
+
+        // Fetch the current ResourceSharing document
+        fetchSharingInfo(resourceIndex, resourceId, sharingInfoListener);
+
+        // Apply patch and update the document
+        sharingInfoListener.whenComplete(resourceSharing -> {
+            // parse and apply the patch
+            Map<String, ShareWith> patches = patchParser(patchContent);
+            ShareWith updatedShareWith = applyPatch(resourceSharing.getShareWith(), patches);
```
cwperks: Why not make these as methods?

i.e. `resourceSharing.getShareWith().add(<recipients_to_add>).revoke(<recipients_to_revoke>)`

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -567,6 +573,151 @@ public void revoke(String resourceId, String resourceIndex, ShareWith revokeAcce
         }, listener::onFailure);
     }
 
+    /**
+     * Helper method to build shareWith object from the supplied patch to then be used to apply the patch
+     * @param patch source content
+     * @return the parsed content map of entities to share and to revoke
+     */
+    @SuppressWarnings("unchecked")
```
cwperks: IMO I think we should do XContent parsing like is done with ResourceSharing here: https://github.com/opensearch-project/security/blob/main/spi/src/main/java/org/opensearch/security/spi/resources/sharing/ResourceSharing.java#L165-L198

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: done.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: done. I've refactored request parser completely.

src/main/java/org/opensearch/security/resources/api/share/ShareRequest.java
DarshitChanpura: yes, I added a validation that either add or revoke must be present.

DarshitChanpura:
> @DarshitChanpura for public facing APIs I think we should use resource_type instead of resource_index as resource_type is a proxy for the index and meant to be a human readable name for that type of sharable resource. We can support resource_index, but let's make sure to extend this and support type as well.

This API will mostly be called by dashboards and REST clients. We have thus far used notion of resource-index everywhere. I'm not too opinionated on resource-index but I think it makes more sense as it strictly ties with the index where the resource is stored.

spi/src/test/java/org/opensearch/security/spi/resources/ShareWithTests.java
```diff
@@ -220,4 +222,57 @@ public void test_writeSharedWithScopesToStream() throws IOException {
 
         verify(mockStreamOutput, times(1)).writeMap(eq(map), any(), any());
     }
+
+    @Test
```
cwperks, line 227: Why all the mocking in here? It should be possible to test this with actual values.

src/main/java/org/opensearch/security/resources/api/share/ShareRequest.java
```diff
@@ -0,0 +1,225 @@
...
+        this.shareWith = in.readOptionalWriteable(ShareWith::new);
+        this.patch = in.readMap(StreamInput::readString, ShareWith::new);
+        this.method = in.readEnum(RestRequest.Method.class);
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        out.writeString(resourceId);
+        out.writeString(resourceIndex);
+        if (shareWith != null) {
```
cwperks: We always need to write a value even if its not present for serialization. Generally for optional variables, there are optional equivalents for writing to StreamOutput to indices to anyone deserializing this that the value is not present.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/TestUtils.java
```diff
@@ -199,6 +208,92 @@ static String migrationPayload_missingBackendRoles() {
...
+        }
+
+        public String build() {
+            String allShares = buildJsonString(share);
+            String allRevokes = buildJsonString(revoke);
+            return """
+                {
+                  "resource_id": "%s",
+                  "resource_type": "%s",
+                  "patch":{
```
cwperks: wdyt about removing `patch` and then check for either `add` and/or `revoke` at the top-level?

spi/src/test/java/org/opensearch/security/spi/resources/ShareWithTests.java
DarshitChanpura: This is unit test of ShareWith object. most of non-mocked items are covered in integ tests.

src/main/java/org/opensearch/security/resources/api/share/ShareRequest.java
DarshitChanpura: since this request is combined for patch and put, i've added this block. share_with with only be null in PATCH request and hence we do not want to write it.

cwperks: I can see its overloaded, but this either 1) always needs to be serialized using one of the OptionalWriteable methods or 2) remove the overloading in ShareRequest and split into GetSharingRequest and ShareRequest.

spi/src/test/java/org/opensearch/security/spi/resources/ShareWithTests.java
cwperks: I think we should cover them in unit tests as well to keep the tests minimal and only testing the unit being tested

src/main/java/org/opensearch/security/resources/api/share/ShareRequest.java
cwperks: Replace with `out.writeOptionalWriteable(shareWith);` and likewise for the others below

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -567,6 +569,73 @@ public void revoke(String resourceId, String resourceIndex, ShareWith revokeAcce
...
+
+        StepListener<ResourceSharing> sharingInfoListener = new StepListener<>();
+        String resourceSharingIndex = getSharingIndex(resourceIndex);
+
+        // Fetch the current ResourceSharing document
+        fetchSharingInfo(resourceIndex, resourceId, sharingInfoListener);
+
+        // Apply patch and update the document
+        sharingInfoListener.whenComplete(resourceSharing -> {
+            ShareWith updatedShareWith = resourceSharing.getShareWith();
```
cwperks, line 600: Can we always have this return empty if not present and move null handling to within `add()` and `revoke()`.

src/main/java/org/opensearch/security/resources/api/share/ShareRequest.java
```diff
@@ -0,0 +1,214 @@
...
+        this.method = builder.method;
+    }
+
+    public ShareRequest(StreamInput in) throws IOException {
+        super(in);
+        this.method = in.readEnum(RestRequest.Method.class);
+        this.resourceId = in.readString();
+        this.resourceIndex = in.readString();
+        this.shareWith = in.readOptionalWriteable(ShareWith::new);
+        this.add = new ShareWith(in);
```
cwperks: likewise change this to readOptionalWriteable like above

spi/src/test/java/org/opensearch/security/spi/resources/ShareWithTests.java
DarshitChanpura: the only mocked item is Recipients here which is not what we are testing. So it should be fine as is.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
cwperks: I think you could also avoid re-assigning this variable and have the updates occur in-place.

DarshitChanpura: It should be fine here. No point in creating and returning empty object if not shareWith exists.

spi/src/test/java/org/opensearch/security/spi/resources/ShareWithTests.java
cwperks: I'd like to see more unit tests added here in the future.

cwperks (approved):

cwperks (approved):

Issue #5464, opened 2025-07-08
[BUG] ActionPrivileges initialization can take a long time in clusters with a large number of roles with repeated index patterns


cwperks:
**What is the bug?**

With [Optimized Privilege Evaluation](https://github.com/opensearch-project/security/issues/3870), much of the work of privilege evaluation is now done upfront by creating optimized data structures on node boostrap or if an index is created/deleted (change to cluster state)

An issue is observed, where the initialization of the ActionPrivileges takes a long time in clusters with a large number of roles that have the same repeated index patterns across role definitions.

Below is a snapshot of the hot threads output:

```
69.6% (348.1ms out of 500ms) cpu usage by thread 'opensearch[f97f9501b3e26a987f62a225d50caa46][generic][T#8]'
     5/10 snapshots sharing following 30 elements
       org.opensearch.security.support.WildcardMatcher$SimpleMatcher.test(WildcardMatcher.java:487)
       org.opensearch.security.support.WildcardMatcher$SimpleMatcher.test(WildcardMatcher.java:461)
       org.opensearch.security.support.WildcardMatcher$MatcherCombiner.lambda$test$0(WildcardMatcher.java:530)
       org.opensearch.security.support.WildcardMatcher$MatcherCombiner$$Lambda/0x000000020343a9c0.test(Unknown Source)
       java.base@21.0.6/java.util.stream.MatchOps$1MatchSink.accept(MatchOps.java:90)
       java.base@21.0.6/java.util.Spliterators$ArraySpliterator.tryAdvance(Spliterators.java:1034)
       java.base@21.0.6/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:129)
       java.base@21.0.6/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:527)
       java.base@21.0.6/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513)
       java.base@21.0.6/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
       java.base@21.0.6/java.util.stream.MatchOps$MatchOp.evaluateSequential(MatchOps.java:230)
       java.base@21.0.6/java.util.stream.MatchOps$MatchOp.evaluateSequential(MatchOps.java:196)
       java.base@21.0.6/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
[...]

PR #5465, opened 2025-07-09
Introduced setting for disabling denormalized privilege data structures

nibix:
### Description

This introduces the setting `plugins.security.privileges_evaluation.precomputed_privileges.enabled` which can be used to disable the creation of denormalized privilege data structures. It is enabled by default to provide the best action throughput. It can make sense to disable the setting when it is seen that the initialisation process takes so much time/resources that it negatively affects the cluster performance (like observed in #5464) . This come at the price of a reduced action throughput.

* Category: Enhancement
* Why these changes are required? Operational safety

### Testing

- Unit testing

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
+++ b/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java
@@ -774,7 +774,7 @@ public class RoleBasedActionPrivileges extends RuntimeOptimizedActionPrivileges 
             this.indices = ImmutableMap.copyOf(indices);
             this.metadataVersion = metadataVersion;
 
-            long duration = System.currentTimeMillis();
+            long duration = System.currentTimeMillis() - startTime;
 
             if (duration > 30000) {
                 log.warn("Creation of StatefulIndexPrivileges took {} ms", duration);


src/integrationTest/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivilegesTest.java
```diff
@@ -1015,6 +1015,25 @@ public void aliasesOnDataStreamBackingIndices() throws Exception {
             );
             assertThat(resultForIndexNotCoveredByAlias, isForbidden());
         }
+
+        @Test
+        public void statefulDisabled() throws Exception {
+            SecurityDynamicConfiguration<RoleV7> roles = SecurityDynamicConfiguration.fromYaml(
+                "test_role:\n" + "  index_permissions:\n" + "  - index_patterns: ['test_*']\n" + "    allowed_actions: ['indices:*']",
```
cwperks, line 1022: nit: let's start using Text Blocks since minimum compatibility is now JDK21.

nibix:
> What would be the effort required to run more of the automated tests with plugins.security.privileges_evaluation.precomputed_privileges.enabled set to false?

We already run all unit tests for RoleBasedActionPrivileges and the AbstractRuleBasedPrivileges family. Due to slight implementation differences, it is explicitly implemented in AbstractRuleBasedPrivileges. In RoleBasedActionPrivileges, it is just implicitly tested via the test param NON_STATEFUL:

https://github.com/opensearch-project/security/blob/da7d66a808266f053120e66fb21fdbf1eda6d3a2/src/integrationTest/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivilegesTest.java#L792-L796

shikharj05:
Thanks @nibix !

> This come at the price of a reduced action throughput.

Do you think an async method should be implemented as a follow-up to still allow building of pre-computed privileges and keep most of the benefits from https://github.com/opensearch-project/security/issues/3870 ? I see your comment [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java#L101 ) - did you see any challenges while trying an async method?

nibix:
> Do you think an async method should be implemented as a follow-up to still allow building of pre-computed privileges and keep most of the benefits from #3870 ? I see your comment [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java#L101) - did you see any challenges while trying an async method or is it just a pending TODO?

That's not a TODO, async processing is already in place. The only exception is currently the security config update process. I am still actively looking for the cause of the issue in #5464 but so far I could not really pinpoint anything.

shikharj05 (approved):

shikharj05:
> > Do you think an async method should be implemented as a follow-up to still allow building of pre-computed privileges and keep most of the benefits from #3870 ? I see your comment [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java#L101) - did you see any challenges while trying an async method or is it just a pending TODO?
> 
> That's not a TODO, async processing is already in place. The only exception is currently the security config update process. However, from the test results so far, I can not really imagine that the CPU time is a real issue here. My tests showed so far maybe 1000 ms of processing time, which is of course not fast, but which also should not lead to timeout issues. I am still actively looking for the cause of the issue in #5464 but so far I could not really pinpoint anything.

we can continue to discuss on the open issue https://github.com/opensearch-project/security/issues/5464

pranu2502:
Hey @nibix, I noticed that the backport attempt was unsuccessful. This appears to be related to PR #5374 , which hasn't yet been merged into the 3.1 branch.

Would you mind verifying if this PR was inadvertently missed? If so, can we backport it now?

cc: @cwperks

nibix:
> Do you mean https://github.com/opensearch-project/security/pull/5374 was not meant to be for 3.1 while the dependent https://github.com/opensearch-project/security/pull/5483 was?

Sorry, that was a mid-air collision regarding comments ;-) This was in response to the request for backports, this is the new backport PR

PR #5470, opened 2025-07-12
Optimizations and simplifications for WildcardMatcher

nibix:
### Description

While researching #5464 we found that the string matching algorithm used by WildcardMatcher does significantly more comparisons than necessary for simple cases. 

This PR does not change the matching algorithm, but it introduces special cases for the most common patterns: `"prefix*"` and `"*contains*"`. In a simple micro benchmark, we found that these optimizations can reduce the runtime required for matching large amounts of strings by 30% (see comments for micro benchmark code).

Additionally, this PR removes unused methods from WildcardMatcher and improves test coverage.

* Category: Enhancement
* Why these changes are required? For clusters with many indices and roles, runtime of wildcard matching gets very significant.

### Testing

- New unit tests in `WilldcardMatcherTests`

### Check List


nibix:
This is the (very simplicistic) micro benchmark:

```
public class ManyRolesTest {
    @Test
    public void test() throws IOException, InterruptedException {

        FlattenedActionGroups actionGroups = new FlattenedActionGroups(
            SecurityDynamicConfiguration.fromNode(
                DefaultObjectMapper.YAML_MAPPER.readTree(
                    DynamicConfigFactory.class.getResourceAsStream("/static_config/static_action_groups.yml")
                ),
                CType.ACTIONGROUPS,
                2,
                0,
                0
            )
        );

        MockIndexMetadataBuilder mockIndexMetadataBuilder = new MockIndexMetadataBuilder();

        for (int i = 0; i < 2500; i++) {
            mockIndexMetadataBuilder.alias(".kibana_t_" + i).of(".kibana_t_" + i + "_001");
        }

        for (int i = 0; i < 20; i++) {
            mockIndexMetadataBuilder.dataStream("my_example_ds_" + i, 10);
        }

        Metadata metadata = mockIndexMetadataBuilder.build();

        StringBuilder rolesString = new StringBuilder();

        for (int i = 0; i < 2000; i++) {
            rolesString.append(
                "role"
                    + i
                    + ":\n"
                    + "  index_permissions:\n"
                    + "  - index_patterns: ['*example*', '.*example*']\n"
                    + "    allowed_actions: ['crud']\n"
                    + "  tenant_permissions:\n"
                    + "  - tenant_patterns: ['tenant_"
                    + i
                    + "']\n"
                    + "    allowed_actions: ['foo']\n\n"
            );
        }
        SecurityDynamicConfiguration<RoleV7> roles = SecurityDynamicConfiguration.fromYaml(rolesString.toString(), CType.ROLES);

        for (int i = 0; i < 20; i++) {
            this.execute(roles, actionGroups, metadata);
        }

    }

    private void execute(SecurityDynamicConfiguration<RoleV7> roles, FlattenedActionGroups actionGroups, Metadata metadata) {
        long s1 = System.currentTimeMillis();
        RoleBasedActionPrivileges subject = new RoleBasedActionPrivileges(roles, actionGroups, Settings.EMPTY);
        System.out.println("Static init took " + (System.currentTimeMillis() - s1) + " ms");
        long s2 = System.currentTimeMillis();
        subject.updateStatefulIndexPrivileges(metadata.getIndicesLookup(), 1);
        System.out.println("Stateful init took " + (System.currentTimeMillis() - s2) + " ms");
    }
}
```

Without optimizations, we observed a runtime of about 7900 ms per loop. With the optimization, the runtime dropped to 5490 ms.

src/test/java/org/opensearch/security/UtilTests.java
```diff
@@ -49,7 +49,7 @@ static private WildcardMatcher wc(String pattern) {
     }
 
     static private WildcardMatcher iwc(String pattern) {
-        return WildcardMatcher.from(pattern, false);
+        return WildcardMatcher.from(pattern).ignoreCase();
```
willyborankin, line 52: :+1:

willyborankin (approved):

src/main/java/org/opensearch/security/rest/SecurityWhoAmIAction.java
```diff
@@ -111,7 +111,7 @@ public void accept(RestChannel channel) throws Exception {
 
                         final String dn = sslInfo.getPrincipal();
                         final boolean isAdmin = adminDns.isAdminDN(dn);
-                        final boolean isNodeCertificateRequest = dn != null && WildcardMatcher.from(nodesDn, true).matchAny(dn);
+                        final boolean isNodeCertificateRequest = dn != null && WildcardMatcher.from(nodesDn).ignoreCase().matchAny(dn);
```
shikharj05, line 114: nit: this moves from case-sensitive to ignoreCase?

nibix: Oops, good catch!

nibix: Yes, that's interesting: If you look at the effective nodesDn matching in DefaultInterClusterRequestEvaluator, this is also done ignoring the case. Thus, this change aligns the behavior of the two string matching operations.

(Sorry, I forgot to explicity point this out before)

shikharj05 (approved):

src/main/java/org/opensearch/security/support/WildcardMatcher.java
```diff
@@ -144,37 +120,50 @@ public boolean test(String candidate) {
...
+                return new Exact(pattern, true);
+            } else if (star == pattern.length() - 1 && questionmark == -1) {
+                // Simple prefix pattern: "a*"
+                return new PrefixMatcher(pattern, true);
+            } else if (pattern.length() > 1 && questionmark == -1 && star == 0 && pattern.indexOf('*', 1) == pattern.length() - 1) {
+                // Simple contains pattern: "*a*"
+                return new ContainsMatcher(pattern, true);
+            } else {
+                return new SimpleMatcher(pattern);
+            }
```
nagarajg17, line 156: If pattern by implicitly is not regex pattern `pattern.startsWith("/") && pattern.endsWith("/")` then if we construct regex pattern explicitly (Ex: `a*`, `*a*`) which is possible, then we can use Java's `pattern.matcher` which would definitely be optimized. Would that be faster than this current approach? I feel code also becomes much simpler

src/main/java/org/opensearch/security/support/WildcardMatcher.java
```diff
@@ -332,8 +310,34 @@ private void init() {
...
+
+    /**
+     * Returns a list with the elements from the given candidates that match this pattern.
+     <p>
+     * This function can iterate through any type of object. The toStringFunction() parameter will be internally used
+     * to convert the object into a string on which the matching can be performed. This is typically a getName() function
+     * or something similar.
+     */
+    public <E> List<E> matching(Collection<E> candidates, Function<E, String> toStringFunction) {
+        List<E> result = new ArrayList<>(Math.min(candidates.size(), 20));
```
nagarajg17, line 328: What if collection size > 20? Would caller of this function handle it?

nagarajg17:
> it introduces special cases for the most common patterns: "prefix*" and "*contains*"
Could you explain more how is this reducing runtime?

src/main/java/org/opensearch/security/support/WildcardMatcher.java
nibix: The present approach has the advantage that it requires zero heap allocations. On the other hand, the JDK `Pattern` and `Matcher` classes require the creation of new objects. Also, the JDK `Pattern` and `Matcher` are generally more heavy due to the support of capturing groups, etc.

I did a quick test and changed `ContainsMatcher` so that internally a `Pattern` and `Matcher` class is used:

```
    static final class ContainsMatcher extends AbstractSimpleWildcardMatcher {
        private final Pattern string;
        private final boolean caseSensitive;

        ContainsMatcher(String pattern, boolean caseSensitive) {
            super(pattern);
            assert pattern.endsWith("*") && pattern.startsWith("*");
            this.string = Pattern.compile(".*" + pattern.substring(1, pattern.length() - 1) + ".*");
            this.caseSensitive = caseSensitive;
        }

        @Override
        public boolean test(String s) {
            Matcher m = this.string.matcher(s);
            return m.matches();
        }

        @Override
        public WildcardMatcher ignoreCase() {
            return new ContainsMatcher(this.pattern, false);
        }
    }
```

The micro benchmark which I listed above now requires 11250 ms per loop, which is nearly twice the runtime which we achieve with the new implementation.

src/main/java/org/opensearch/security/support/WildcardMatcher.java
nibix: The value passed to the constructor of `ArrayList` is just the initial capacity. It will still auto expand, if the initial capacity is used up.

nibix:
> Could you explain more how is this reducing runtime?

See the comments at https://github.com/opensearch-project/security/issues/5464#issuecomment-3055414928 and https://github.com/opensearch-project/security/issues/5464#issuecomment-3055639743 : 

In the CPU profile, it can be seen that the WildcardMatcher did in average about 64 `charAt()` calls per `test()` call. As the test index names are considerably shorter, this indicates that the algorithm does not perform optimally.

Using `contains()` or `startsWith()` on strings, we limit the number of character comparisons to the length of the string. Thus, as we do fewer comparisons, we will be faster.

src/main/java/org/opensearch/security/support/WildcardMatcher.java
nagarajg17: Oh I see. Thanks for checking!

src/main/java/org/opensearch/security/support/WildcardMatcher.java
```diff
@@ -255,18 +233,6 @@ public boolean matchAny(String... candidates) {
...
-        return candidates.allMatch(this);
-    }
-
-    public boolean matchAll(Collection<String> candidates) {
-        return matchAll(candidates.stream());
-    }
-
-    public boolean matchAll(String[] candidates) {
-        return matchAll(Arrays.stream(candidates));
-    }
```
kumargu, line 268: @nibix do you remember why was this removed? AWS had a dependency on this method (as a patch) which now seems broken. is it okay to bring it back as it is?

kumargu: also do we know why this code is commented? should we remove it?

https://github.com/opensearch-project/security/blob/2.19/src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java#L566-L571

kumargu: nevermind. i found why it was removed. 

reference https://github.com/opensearch-project/security/commit/c34e772e90875ac92418726919e582615b9e606a

nibix: I removed these because these were not used any more. 

In the 2.19 branch, one of them is still used in one obsolete part of the code. However, these can be easily replaced by `Stream.of(actions).allMatch(wildcardMatcher)`.

nibix: Regarding the commented code in https://github.com/opensearch-project/security/blob/2.19/src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java#L566-L571 

My guess would be that this code is this way since the inception of OpenSearch (or OpenDistro). It stems from old ES 6 times, when there were still doc types. It is totally safe to remove that code. I guess there are a couple of more instances like this.

kumargu: > My guess would be that this code is this way since the inception of OpenSearch (or OpenDistro).

Got it. 

>  I guess there are a couple of more instances like this.

Yeah found some more commented too. They just add confusion during audits.

nibix: > Yeah some more commented too. They just add confusion during audits.

I just checked a bit of the code: Good news is that we could remove quite a lot of that with the optimized privilege evaluation PR.

kumargu: Super nice then.

Issue #5468, opened 2025-07-11
[FEATURE] Security Permissions Simulation


nagarajg17:
### Problem

Customers enable FGAC to have better security posture and implement principle of least privileges. For the Authorization part particularly they use Roles and RolesMapping to achieve it. While configuring various permissions in roles, they would need sometimes to verify if permission they have configured is good enough or over permissive. Currently there is no good way other than configuring and asking the user to test if they are able to perform all the required operations. This also makes it difficult to know if they have undesired/over permissions. Another problem is when they want to update role permissions and if they do it incorrectly it can have negative experience for the users as they may not be able to perform the actions they were able to. Currently its difficult to test/troubleshoot changes to permissions without any impact

### Benefits

- Risk-free permission testing
- Reduced user disruption
- Faster configuration validation
- Better security posture

### Proposal

Create a Security Permissions Simulation API that allows testing permission configurations before actually making a change and need of real testing by users. This API will mimic the actual privilege evaluation flow. This will resolve the indexes to concrete indices and has all capabilities that privilege evaluations offers today. Simulation can be performed on roles or users or both. Further it can be done on a existing role/new role that would be added/existing role that would be modified to and same with user new user/existing user

#### API Design:
- Endpoint: POST /_plugins/_security/api/simulation
- Access Control: Admin/Security Manager permissions required
- Request body parameters: 
    - user
        - type: string
        - required: optional
    - role_name
        - type: string
        - required: optional
    - action
        - type: string
        - required: yes
    - index: 
        - type: string
        - required: yes
     - roles:
        - type: json
        - required: optional
     - roles_mapping:
        - type: json
        - required: optional
     - actions_groups:
        - type: json
        - required: optional
- Response format:
     - accessAllowed
        - type: boolean
     - missingPrivileges
         - type: array

shikharj05:
Thanks for the proposal @nagarajg17 

Instead of a new API - does it make sense to add a new Query-param in the any API request to support this universally across APIs? Security plugin can short-circuit the request if this new parameter is present in the request.

shikharj05:
> This approach will work for only permissions already present in security index but if user wants to validate for new roles/ permissions that they would want to update thinking how we can use this approach?

While making changes, I think a separate functionality in security APIs - roles, users, etc should have a dry_run flag. Thoughts?

nagarajg17:
Yeah this param will certainly help in not updating the security index in security APIs. Trying to understand how we can validate permissions. Could you explain your idea more?

PR #5473, opened 2025-07-14
Added Simulation API for security configuration permissions

pranathz:
### Description
**Permission Simulation API for Security Configuration**

- Category: New feature
- Why these changes are required?
  Previously, there was no way to preview whether permission access would be allowed or denied before applying new security configurations. This change allows users to simulate the permissions of proposed security configuration changes without applying them. and It also supports simulation using the current configuration.
- What is the old behavior before changes and new behavior after changes?
 **Old Behavior:** No simulation support for evaluating permission changes before applying.
**New Behavior:** Introduces a Simulation API that allows users to preview the effect (Allow/Deny) using either current or proposed configuration

### Supported Simulation Modes
1. Role Based Simulation
2. User Based Simulation
Each Mode Supports both proposed and Existing configuration

### API Design:
Endpoint: `POST /_plugins/_security/api/simulation`

**Request Format:
Role-Based Simulation (Proposed Configuration)**
```
curl -X POST "https://localhost:9200/_plugins/_security/api/simulation?pretty" \
 -H "Content-Type: application/json" \
 -u 'admin:<password>' \
 --insecure \
 -d '{
  "role_name": "test_role",
  "action": "indices:data/read/get",
  "index": "logs-test",
  "roles": {
   "test_role": {
    "cluster_permissions": [],
    "index_permissions": [
     {
      "index_patterns": ["logs-*"],
      "allowed_actions": ["read"]           
     }
    ]
   }
  },
  "action_groups": {
   "read": {
    "allowed_actions": ["indices:data/read/search", "indices:data/read/get"]
   }
  }
 }'
```
**Role-Based Simulation (Existing Configuration)**
```
curl -X POST "https://localhost:9200/_plugins/_security/api/simulation?pretty" \
 -H "Content-Type: application/json" \
 -u 'admin:<password>' \
 --insecure \
 -d '{
  "role_name": "kibana_read_only",
  "action": "indices:data/read/search",
  "index": "kibana_sample_data_logs"
 }'
```
**User-Based Simulation (Proposed Configuration)**
```
curl -X POST "https://localhost:9200/_plugins/_security/api/simulation?pretty" \
 -H "Content-Type: application/json" \
 -u 'admin:<password>' \
 --insecure \
 -d '{
  "user": "testuser",
  "action": "indices:data/read/search",
  "index": "logs-test",
  "roles": {
   "custom_reader": {
    "cluster_permissions": [],
    "index_permissions": [
     {
      "index_patterns": ["logs-*"],
      "allowed_actions": ["indices:data/read/search"]
     }
    ]
   }
  },
  "roles_mapping": {
   "custom_reader": {
    "users": ["testuser"]
   }
  }
 }'
```
**User-Based Simulation (Existing Configuration)**
```
curl -X POST "https://localhost:9200/_plugins/_security/api/simulation?pretty" \
-H "Content-Type: application/json" \
-u 'admin:<password>' \
--insecure \
-d '{
  "user": "testuser",
  "action": "indices:data/read/search",
  "index": "logs-test"
}'
```

Response format if action is allowed
```
{
  "accessAllowed" : true,
  "missingPrivileges" : [ ]
}
```
Response format if action is not allowed
```
{
  "accessAllowed" : false,
  "missingPrivileges" : ["indices:data/read/msearch"]
}
```

**Validation Responses**

Empty request body
```
{
  "status": "BAD_REQUEST",
  "message": "Request body cannot be empty"
}
```
Missing action field
```
{
  "status": "BAD_REQUEST",
  "message": "Missing required field: 'action'. Action field is required for permission simulation"
}
```
Missing both role_name and user
```
{
  "status": "BAD_REQUEST",
  "message": "Either 'role_name' or 'user' must be provided to simulate permissions."
}
```
**Issues Resolved**
Related to [5468](https://github.com/opensearch-project/security/issues/5468)

**Related Documentation**
[Simulation API documentation PR](https://github.com/opensearch-project/documentation-website/pull/10250)

### Testing
Unit Tests and Integration Tests included

### Check List


pranathz:
The current API design is focused only for admins to validate security configuration permission changes at index and cluster level before applying them. its not part of dashboard UI

nibix:
Just throwing in an idea to address the concern from @cwperks : 

How about not introducing a new API, but rather a new query param or maybe a prefix that is added to the existing REST APIs.

The query param (or prefix) could be consumed by the security plugin in its RestFilter. Then, in the ActionFilter, the security plugin would perform the normal privilege evaluation process. However, when the new param is present, we won't continue on the ActionFilterChain to execute the actual transport action. Rather, we would just display the simulation result.

This has two advantages:

- You will be automatically reusing the logic of each action which parses REST requests into ActionRequests, which are the basis for the actual privilege evaluation.
- I would believe that the code will be actually quite simple.

nagarajg17:
@nibix 
@pranathz is trying the approach you mentioned. Wanted to understand more - one of the simulation usecase is before user makes a change to security index for new/updated role permissions, they would want to confirm if permissions works as expected and then commit the change to security index. In this approach would that be feasible?

cwperks:
I'd prefer to use a query param then update the routes because that would lead to an explosion in the number of routes.

FYI there is one (unused) use-case where we do duplicate all registered routes in core and that is in the [opensearch-dashboards module](https://github.com/opensearch-project/OpenSearch/tree/main/modules/opensearch-dashboards)

nagarajg17:
@cwperks She is trying with query param only and not with API path for same reason you mentioned

nibix:
@nagarajg17 

> Wanted to understand more - one of the simulation usecase is before user makes a change to security index for new/updated role permissions, they would want to confirm if permissions works as expected and then commit the change to security index. In this approach would that be feasible?

In order to tell, I'd need to understand the use cases a bit better. How would the user decide which APIs to test? They would then test one API after the other? 

Generally, could you elaborate on the test process you have in mind?

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -164,6 +164,11 @@ public void handleRequest(RestRequest request, RestChannel channel, NodeClient c
                 return;
             }
 
+            boolean isSimulation = request.paramAsBoolean(ConfigConstants.SECURITY_SIMULATE_AUTHZ_PARAM, false);
+            if (isSimulation) {
+                threadContext.putTransient(ConfigConstants.SECURITY_SIMULATE_AUTHZ_PARAM, Boolean.TRUE);
```
nagarajg17: I think we need to use `putHeader` else this will be discarded when transport action is executed on other nodes (than the node which received rest req) and may result in real API execution instead of simulation.

Can we test with multi node this feature?

nagarajg17:
@nibix Thought process was - 
Admins or security managers when they configure roles and permissions for the cluster for all the users they might need to test if it works as expected or if they have under/over provisioned access. So before they actually make updates to security index, if they can test their configurations (which they want to update in security index) will help them. So in earlier approach we were taking roles, roles mapping etc for same purpose and action was used to know what scenario they wanted to test

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
pranathz: Makes sense, I’ll update it to use putHeader.

cwperks:
@pranathz Can you please resolve the conflicts?

DarshitChanpura:
Closing in lieu of #5496

Issue #5472, opened 2025-07-13
[BUG] Security configuration update after snapshot restore reloads security configuration incompletely


nibix:
**What is the bug?**

In #5307, a functionality was introduced to reload the security configuration from the security index after the index was restored from a snapshot.

The main logic is located here:

https://github.com/opensearch-project/security/blob/deb19f9a8fbd7cc4be0e1c8018c295285f9320f6/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java#L678-L695

You can see that line 685 restricts the configuration update to the primary shard of the index.

By default, the security index has the following index settings:

```
index.number_of_shards: 1
index.auto_expand_replicas: 0-all
```

These are defined here: https://github.com/opensearch-project/security/blob/deb19f9a8fbd7cc4be0e1c8018c295285f9320f6/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java#L325-L328

Thus, by default, the security index will have a single primary shard. During creation, OpenSearch will choose an arbitrary data node on which the primary shard will be located. Each other node will get a replica shard; this is indicated by the `auto_expand_replicas` setting.

This means that the reload logic quoted above will only trigger the reload on the OpenSearch node which hosts the primary shard. On all other nodes, reloading will be skipped. Thus, afterwards, you will have a cluster with an inconsistent security configuration across nodes.

**How can one reproduce the bug?**
1. On a cluster with more than one node, restore a snapshot containing the security index
2. Observe the logs on all nodes
3. You will see that only one node will reload the security config

**What is the expected behavior?**

All nodes should reload the security configuration

cwperks:
Thank you for filing this issue Nils. Let's remove the check and ensure the logic runs on all nodes.

cwperks:
The security index is set to auto_expand 0-all so all nodes of the cluster should have a shard. I suppose there could be an issue with node drop, but wouldn't a node recover on bootup after its been replaced or comes back online?

nibix:
> The security index is set to auto_expand 0-all so all nodes of the cluster should have a shard. 

I think that will be only the case if all nodes in the cluster have a `data` role. Pure coordinator or cluster manager roles won't get any shards, IMHO:

https://docs.opensearch.org/docs/latest/tuning-your-cluster/

sandervandegeijn:
There is another problem with this 0-all designation: if one node is above the low watermark for disk usage (not a critical error in itself) this will lead to a yellow cluster because a replica can't be assigned.

PR #5478, opened 2025-07-15
Bug fix: Fixing partial cache update post snapshot restore

nagarajg17:
### Description
* Category: Bug Fix

In [#5307](https://github.com/opensearch-project/security/pull/5307), a functionality was introduced to update the cache from the security index after the index was restored from a snapshot. It was done on nodes having only primary shard. This PR fixes to reload cache on all nodex
   
Cache update was only done for nodes having primary shard. Now it will be done for all nodes

### Issues Resolved
[5472](https://github.com/opensearch-project/security/issues/5472)

### Testing
Tested with 2 node setup on single machine. 
1. Added  myuser1 with permissions to read index `my-index-000001`
2. Took the snapshot of security index
3. Added myuser2 with permissions to read index `my-index-000001`
4. Verified myuser2 was able to read index from both nodes
5. Deleted security index
6. Restored from snapshot
7. Verified cache getting updating on both nodes
8. Verified myuser2 was not able to read index from both nodes

```
# Node 1
cluster.name: test-cluster
node.name: node-1
node.roles: [ master, data ]
network.host: 0.0.0.0
http.port: 9200
transport.port: 9300
discovery.seed_hosts: ["localhost:9300", "localhost:9301"]
cluster.initial_master_nodes: ["node-1"]
```

```
# Node 2
cluster.name: test-cluster
node.name: node-3
node.roles: [ data ]
network.host: 0.0.0.0
http.port: 9201
transport.port: 9301
discovery.seed_hosts: ["localhost:9300", "localhost:9301"]
cluster.initial_master_nodes: ["node-1"]
```

### Check List


src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -681,11 +681,10 @@ public void afterIndexShardStarted(IndexShard indexShard) {
 
         // Check if this is a security index shard
         if (securityIndex.equals(index.getName())) {
-            // Only trigger on primary shard to avoid multiple reloads
-            if (indexShard.routingEntry() != null && indexShard.routingEntry().primary()) {
+            if (indexShard.routingEntry() != null) {
                 threadPool.generic().execute(() -> {
```
cwperks, line 653: Let's make sure this get's updated when we have a dedicated threadpool for security updates

cwperks (approved):

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nagarajg17: Ack

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -681,11 +681,10 @@ public void afterIndexShardStarted(IndexShard indexShard) {
 
         // Check if this is a security index shard
         if (securityIndex.equals(index.getName())) {
-            // Only trigger on primary shard to avoid multiple reloads
-            if (indexShard.routingEntry() != null && indexShard.routingEntry().primary()) {
+            if (indexShard.routingEntry() != null) {
```
nibix: See my comments at https://github.com/opensearch-project/security/issues/5472#issuecomment-3073753962 ... I have doubts that this is sufficient. OpenSearch nodes can be configured not to have any shards at all. IMHO, this will be also the case despite the `index.auto_expand_replicas: 0-all` setting.

nibix (changes requested):
I have doubts that this will fix the issue. 

Of course, I might be wrong here. However, in any case, a feature that relies on such specific cluster behavior should *always* come with an integration test which tests it on a real cluster with nodes assuming different roles.

Just having a unit test which mocks dependencies won't be sufficient, as these mocks encode just the assumptions of the developer - which might be just wrong.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
cwperks: Doesn't this condition imply that the node running this code contains a shard?

cwperks: agree on integration test for the behavior

nibix: > Doesn't this condition imply that the node running this code contains a shard?

I have not tested this, but IMHO, there is the realistic chance that on nodes without any shards, the methods will not be called at all.

willyborankin:
@nagarajg17, @cwperks, and @nibix — there's a `SnapshotRestoreHelper` class that manages and provides information about the restoration process. I think it makes sense to use this class and reload the configuration once the restoration is complete.

willyborankin:
> @willyborankin The SnapshotRestoreHelper is being using here to determine if the security index is being restored from a snapshot. See https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/SnapshotRestoreHelper.java#L96-L112

Ahhh did not notice static import :-) .

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nagarajg17: I'll add Integ tests for this

nagarajg17: Discussed with Craig and removed the conditions. Added Integration test

CHANGELOG.md
```diff
@@ -18,6 +18,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 * Allow any plugin system request when `plugins.security.system_indices.enabled` is set to `false` ([#5579](https://github.com/opensearch-project/security/pull/5579))
+* Fix compilation issue after change to Subject interface in core and bump to 3.2.0 ([#5423](https://github.com/opensearch-project/security/pull/5423))
+* Provide SecureHttpTransportParameters to complement SecureTransportParameters counterpart ([#5432](https://github.com/opensearch-project/security/pull/5432))
+* Use isClusterPerm instead of requestedResolved.isLocalAll() to determine if action is a cluster action ([#5445](https://github.com/opensearch-project/security/pull/5445))
+* Fix config update with deprecated config types failing in mixed clusters ([#5456](https://github.com/opensearch-project/security/pull/5456))
+* Fix usage of jwt_clock_skew_tolerance_seconds in HTTPJwtAuthenticator ([#5506](https://github.com/opensearch-project/security/pull/5506))
+* Fix partial cache update post snapshot restore[#5478](https://github.com/opensearch-project/security/pull/5478)
```
cwperks: @nagarajg17 can you fix the CHANGELOG here? Otherwise this looks good to me.

nibix:
Thank you for the integration test!

However, apologies once again; and I am not really feeling good with always being the one who sees issues here :-( 

I am still not convinced that this works the way it is. I just tried the integration test. Additionally, I added some more logs to `afterIndexShardStarted()` in `ConfigurationRepository`: 

```
    @Override
    public void afterIndexShardStarted(IndexShard indexShard) {
        final ShardId shardId = indexShard.shardId();
        final Index index = shardId.getIndex();

        // Check if this is a security index shard
        if (securityIndex.equals(index.getName())) {
            threadPool.generic().execute(() -> {
                LOGGER.info("Shard started on node " + clusterService.localNode().getName());
                if (isSecurityIndexRestoredFromSnapshot(clusterService, index, securityIndex)) {
                    LOGGER.info("Security index shard {} started - config reloading for snapshot restore; node: {}", shardId, clusterService.localNode().getName());
                    reloadConfiguration(CType.values());
                }
            });
        }
    }
```

The output of the test is as follows:

```
2025-09-05 18:47:30 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] INFO  RepositoriesService:252 - put repository [test-snapshot-repository]
2025-09-05 18:47:30 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] INFO  PluginsService:343 - PluginService:onIndexModule index:[my_index_001/vnGUzqpjR6yMbTm5l1WWBQ]
2025-09-05 18:47:30 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] DEBUG SecurityFlsDlsIndexSearcherWrapper:103 - FLS/DLS org.opensearch.security.configuration.SecurityFlsDlsIndexSearcherWrapper@28ebbc7e enabled for index my_index_001
2025-09-05 18:47:30 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] INFO  MetadataCreateIndexService:563 - [my_index_001] creating index, cause [api], templates [], shards [1]/[1]
2025-09-05 18:47:30 opensearch[data_0][clusterApplierService#updateTask][T#1] INFO  PluginsService:343 - PluginService:onIndexModule index:[my_index_001/vnGUzqpjR6yMbTm5l1WWBQ]
2025-09-05 18:47:30 opensearch[data_0][clusterApplierService#updateTask][T#1] DEBUG SecurityFlsDlsIndexSearcherWrapper:103 - FLS/DLS org.opensearch.security.configuration.SecurityFlsDlsIndexSearcherWrapper@3bff4c3c enabled for index my_index_001
2025-09-05 18:47:30 opensearch[data_1][clusterApplierService#updateTask][T#1] INFO  PluginsService:343 - PluginService:onIndexModule index:[my_index_001/vnGUzqpjR6yMbTm5l1WWBQ]
2025-09-05 18:47:30 opensearch[data_1][clusterApplierService#updateTask][T#1] DEBUG SecurityFlsDlsIndexSearcherWrapper:103 - FLS/DLS org.opensearch.security.configuration.SecurityFlsDlsIndexSearcherWrapper@3e9d578f enabled for index my_index_001
2025-09-05 18:47:30 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] INFO  PluginsService:343 - PluginService:onIndexModule index:[my_index_001/vnGUzqpjR6yMbTm5l1WWBQ]
2025-09-05 18:47:30 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] INFO  MetadataMappingService:335 - [my_index_001/vnGUzqpjR6yMbTm5l1WWBQ] create_mapping
2025-09-05 18:47:30 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] INFO  PluginsService:343 - PluginService:onIndexModule index:[my_index_001/vnGUzqpjR6yMbTm5l1WWBQ]
2025-09-05 18:47:31 opensearch[data_0][generic][T#5] INFO  RecoverySourceHandler:916 - finalizing recovery took [62.3ms]
2025-09-05 18:47:31 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] INFO  AllocationService:577 - Cluster health status changed from [YELLOW] to [GREEN] (reason: [shards started [[my_index_001][0]]]).
2025-09-05 18:47:31 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] INFO  SnapshotsService:443 - snapshot [test-snapshot-repository:test-snap/JhJTMIK6T1CXyEndM73Zjg] started
2025-09-05 18:47:31 opensearch[cluster_manager_1][snapshot][T#3] INFO  SnapshotsService:2175 - snapshot [test-snapshot-repository:test-snap/JhJTMIK6T1CXyEndM73Zjg] completed with state [SUCCESS]
2025-09-05 18:47:34 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] INFO  MetadataDeleteIndexService:166 - [.opendistro_security/ZDkcCrfES56szgkkYqcs6Q] deleting index
2025-09-05 18:47:35 opensearch[data_0][clusterApplierService#updateTask][T#1] INFO  PluginsService:343 - PluginService:onIndexModule index:[.opendistro_security/K0nnLcTYRrqlA4kl12KhKg]
2025-09-05 18:47:35 opensearch[data_0][clusterApplierService#updateTask][T#1] DEBUG SecurityFlsDlsIndexSearcherWrapper:103 - FLS/DLS org.opensearch.security.configuration.SecurityFlsDlsIndexSearcherWrapper@2125eceb enabled for index .opendistro_security
2025-09-05 18:47:35 opensearch[data_0][generic][T#2] INFO  ConfigurationRepository:685 - Shard started on node data_0
2025-09-05 18:47:35 opensearch[data_0][generic][T#2] INFO  ConfigurationRepository:687 - Security index shard [.opendistro_security][0] started - config reloading for snapshot restore; node: data_0
2025-09-05 18:47:35 opensearch[data_1][clusterApplierService#updateTask][T#1] INFO  PluginsService:343 - PluginService:onIndexModule index:[.opendistro_security/K0nnLcTYRrqlA4kl12KhKg]
2025-09-05 18:47:35 opensearch[data_1][clusterApplierService#updateTask][T#1] DEBUG SecurityFlsDlsIndexSearcherWrapper:103 - FLS/DLS org.opensearch.security.configuration.SecurityFlsDlsIndexSearcherWrapper@2755ec7f enabled for index .opendistro_security
2025-09-05 18:47:35 opensearch[data_0][generic][T#3] INFO  RecoverySourceHandler:916 - finalizing recovery took [9.3ms]
2025-09-05 18:47:35 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] INFO  AllocationService:577 - Cluster health status changed from [YELLOW] to [GREEN] (reason: [shards started [[.opendistro_security][0]]]).
2025-09-05 18:47:35 opensearch[data_1][generic][T#5] INFO  ConfigurationRepository:685 - Shard started on node data_1
2025-09-05 18:47:35 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] INFO  MetadataDeleteIndexService:166 - [my_index_001/vnGUzqpjR6yMbTm5l1WWBQ] deleting index
2025-09-05 18:47:35 opensearch[cluster_manager_1][clusterManagerService#updateTask][T#1] INFO  RepositoriesService:364 - delete repository [test-snapshot-repository]
2025-09-05 18:47:36 ForkJoinPool.commonPool-worker-6 INFO  LocalOpenSearchCluster:505 - Stopping cluster_manager_0 RUNNING [47320, 47220]
```

You can see that the log message for `afterIndexShardStarted()` is only emitted for the nodes `data_0` and `data_1`. The cluster is however configured as `ClusterManager.THREE_CLUSTER_MANAGERS`:

```
    THREE_CLUSTER_MANAGERS(
        new NodeSettings(NodeRole.CLUSTER_MANAGER),
        new NodeSettings(NodeRole.CLUSTER_MANAGER),
        new NodeSettings(NodeRole.CLUSTER_MANAGER),
        new NodeSettings(NodeRole.DATA),
        new NodeSettings(NodeRole.DATA)
    ),
```

So, we have two data nodes and three further cluster manager nodes. The cluster manager nodes do not get any calls to `afterIndexShardStarted()`. 

Generally, the int test should verify that each node has loaded the valid configuration. With the current test, you only verify that a single - undetermined - node has reloaded the configuration.

cwperks:
Got it, I suppose we can yank the code from AbstractApiAction to trigger a ConfigUpdateAction from the node with a primary shard: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/dlic/rest/api/AbstractApiAction.java#L502-L503

nibix:
yip, I think that's the way to go

cwperks:
Actually the important section is the ConfigUpdatingActionListener here: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/dlic/rest/api/AbstractApiAction.java#L551-L567

The index has already been restored from a snapshot, so we need to trigger `ConfigUpdateAction` using the node client for all ctypes.

nibix:
Yes. Still, I think, we also have to keep timing in mind: When the primary shard has been restored, it's possible that not all shards are up to date. If we are having bad luck, the update request might be faster than the restoring of the shards and trigger the reloading too early.

I am not sure how to take care of this: Possibly a fixed delay? That would be a bit hacky and fragile as well.

src/integrationTest/java/org/opensearch/security/SecurityIndexSnapshotRestoreTests.java
```diff
@@ -0,0 +1,189 @@
...
+            assertEquals(RestStatus.OK, restoreResponse.status());
+        }
+
+        // 7. Read data in custom index with LIMITED_READ_USER_1 because it was in snapshot
+        try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(LIMITED_READ_USER_1)) {
+            GetResponse response = restHighLevelClient.get(new GetRequest(TEST_INDEX_NAME, DOC_ID), RequestOptions.DEFAULT);
+            assertThat(response, containDocument(TEST_INDEX_NAME, DOC_ID));
+        }
+
+        // 8. Should get 401 error to read custom index with LIMITED_READ_USER_2 because it was not in snapshot
```
nibix, line 191: I think it would be cool if we could test the property that the correct configuration was restored in a direct fashion, and not just using an indirect effect like done here. This would also have the advantage that we could verify the fact that the configuration was actuall reloaded for all nodes. 

You cant verify that property with index operations.

Thus, I would propose the following; after restoring the config snapshot, do something like:

```
for (LocalOpenSearchCluster.Node node : cluster.nodes()) {
   ConfigurationRepository configRepository = node.getInjectable(ConfigurationRepository.class);
   // We will get the node-specific configRepository instance here
   // TODO do some assertion on the SecurityDynamicConfiguration instances stored in the ConfigurationRepository instance
}
```

src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
```diff
@@ -583,6 +595,72 @@ public void getConfigurationsFromIndex_SecurityIndexNotInitiallyReady() throws I
         assertThat(result.size(), is(CType.values().size()));
     }
 
+    @Test
+    public void afterIndexShardStarted_whenSecurityIndexRestoredFromSnapshot() throws InterruptedException, TimeoutException {
```
nibix, line 599: Does this test something that goes beyond what we are testing in the integration test?

src/integrationTest/java/org/opensearch/security/SecurityIndexSnapshotRestoreTests.java
```diff
@@ -0,0 +1,189 @@
...
+            client.assignRoleToUser(LIMITED_READ_USER_2.getName(), "limited-reader_2").assertStatusCode(200);
+        }
+
+        // 4. Read data in custom index with LIMITED_READ_USER_2
+        try (RestHighLevelClient restHighLevelClient = cluster.getRestHighLevelClient(LIMITED_READ_USER_2)) {
+            GetResponse response = restHighLevelClient.get(new GetRequest(TEST_INDEX_NAME, DOC_ID), RequestOptions.DEFAULT);
+            assertThat(response, containDocument(TEST_INDEX_NAME, DOC_ID));
+        }
+
+        // 5. Delete security index
```
cwperks, line 167: I'm also wondering whether we should test a scenario where we take a snapshot, then make modifications to security config (i.e. change a role). Verify that changes take effect, then restore the index from a snapshot and verify it reverts back to the config when the snapshot was taken

nagarajg17: Yes. In step 6 we are restoring from snapshot and checking permissions are working as expected but yeah direct config match I'm not doing though

src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
nagarajg17: No same only with mocking. Do we not write UT when we've Integ?

src/integrationTest/java/org/opensearch/security/SecurityIndexSnapshotRestoreTests.java
nagarajg17: Ack

src/integrationTest/java/org/opensearch/security/SecurityIndexSnapshotRestoreTests.java
cwperks: not that though. i.e. Take a snapshot, then add perms (like `indices:data/read/search*` on `test-*` index. Then verify the search succeeds after modifying the role. Afterwards restore from a snapshot (the snapshot was from before perms were added) and then search again. Search should fail.

src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
nibix: I don't think we need to test the same thing twice; IMHO, integration tests and unit tests should complement each other, i.e. unit tests can fill coverage gaps which are difficult to achive with integration tests. But, if we do not need to add further coverage, this seems to be redundant.

One of my mottos is: The best code is code that does not exist ;-)

src/integrationTest/java/org/opensearch/security/SecurityIndexSnapshotRestoreTests.java
nagarajg17: Yeah this would be ideal test

src/test/java/org/opensearch/security/configuration/ConfigurationRepositoryTest.java
nagarajg17: I think UT and Integ both serve different aspects in test pyramid and both are imp. I think should be okay to keep both of them

nibix: When writing code, we also always have to consider that someone needs to maintain the code later (that's then the maintainers). Especially unit tests with heavy mocking usage are extremely difficult to maintain as they react very fragile to internal changes. 

I do not want to block this, but I still maintain my opinion that we should only have either int test or unit test if they have exactly the same coverage and do not complement each other in any way, i.e. if one test does not add value which can be concretely named.

PR #5484, opened 2025-07-17
[3.1 backport] Introduced setting for disabling denormalized privilege data structures

nibix:
### Description

Backport from https://github.com/opensearch-project/security/pull/5465

This introduces the setting `plugins.security.privileges_evaluation.precomputed_privileges.enabled` which can be used to disable the creation of denormalized privilege data structures. It is enabled by default to provide the best action throughput. It can make sense to disable the setting when it is seen that the initialisation process takes so much time/resources that it negatively affects the cluster performance (like observed in #5464) . This come at the price of a reduced action throughput.

* Category: Enhancement
* Why these changes are required? Operational safety

### Testing

- Unit testing

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased 3.1.x]
 ### Added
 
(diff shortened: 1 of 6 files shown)


nibix:
This replaces #5483 as the CI there seems to be broken without any obvious way to fix

nibix:
The CI seems to be weird here as well 🤔 possibly I need to do a dummy commit to get a different commit hash.

PR #5486, opened 2025-07-17
update role for query insights

ansjcy:
### Description
* Category: Maintenance
* Why these changes are required: We need to update the role for query insights access to allow addition APIs we introduced in 3.x

### Issues Resolved

### Testing

### Check List


RyanL1997 (approved):

cwperks (approved):
Approving but can you please explain the change inded pattern here? Would any users need to perform a migration from the older naming to newer naming?

ansjcy:
There are 2 changes in this PR:
1. We added permission to access live queries API
2. Fixed the index permission to the right one. There's no top_queries_by index, It's a bug that could cause users not able to access the correct top_queries-* index. So there's no migration needed.

config/roles.yml
```diff
@@ -451,10 +451,11 @@ flow_framework_read_access:
 query_insights_full_access:
   reserved: true
   cluster_permissions:
+    - 'cluster:admin/opensearch/insights/live_queries/*'
     - 'cluster:admin/opensearch/insights/top_queries/*'
   index_permissions:
     - index_patterns:
-        - 'top_queries_by_*'
```
shikharj05, line 457: this was never used?

shikharj05:
can you also add a change-log entry?

config/roles.yml
DarshitChanpura: also, is suffix -* correct or should it be _*?

config/roles.yml
```diff
@@ -451,10 +451,11 @@ flow_framework_read_access:
...
+        - 'top_queries-*'
```
cwperks, line 458: +1 to @DarshitChanpura 's comment above. Should we mix _ and - here?

other question I had is whether these indices should be directly queryable by a user? If these are not system indices then any user with requisite permissions on these indices can write to them directly or delete them. Can the query insights plugin handle such cases gracefully?

config/roles.yml
ansjcy: yes, it should be `-`. This is defined in query insights plugin.

dzane17 (approved):

config/roles.yml
ansjcy: > should be directly queryable by a user

When we design this feature we want it to be similar to audit log and it should be queryable by a user. So this is expected.

PR #5491, opened 2025-07-18
Add user custom attributes to thread context

markdboyd:
### Description

* Category: Bug fix

In https://github.com/opensearch-project/alerting/issues/1829#issuecomment-3019268162, I was informed that the the user serialization into the threadcontext does not handle custom attributes, which breaks DLS for alert monitors. 

Without these changes, the serialized user in the thread context **does not include** custom attribute names. With these changes, the serialized user in the thread context **does include** custom attribute names.

### Related Issues

Related to https://github.com/opensearch-project/alerting/issues/1829

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ ] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/common-utils/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### Testing

I have not done any testing because I was unsure how to implement it or run it

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
@@ -116,16 +116,14 @@ import static org.opensearch.security.support.SecurityUtils.escapePipe;
 public class PrivilegesEvaluator {
 
     static final WildcardMatcher DNFOF_MATCHER = WildcardMatcher.from(
-        ImmutableList.of(
-            "indices:data/read/*",
-            "indices:admin/mappings/fields/get*",
-            "indices:admin/shards/search_shards",
-            "indices:admin/resolve/index",
-            "indices:monitor/settings/get",
-            "indices:monitor/stats",
-            "indices:admin/aliases/get"
-        )
-    );
+            ImmutableList.of(
+                    "indices:data/read/*",
+                    "indices:admin/mappings/fields/get*",
+                    "indices:admin/shards/search_shards",
+                    "indices:admin/resolve/index",
+                    "indices:monitor/settings/get",
+                    "indices:monitor/stats",
+                    "indices:admin/aliases/get"));
 
     private static final WildcardMatcher ACTION_MATCHER = WildcardMatcher.from("indices:data/read/*search*");
 
@@ -159,19 +157,18 @@ public class PrivilegesEvaluator {
     private final AtomicReference<ActionPrivileges> actionPrivileges = new AtomicReference<>();
 
     public PrivilegesEvaluator(
-        final ClusterService clusterService,
-        Supplier<ClusterState> clusterStateSupplier,
-        ThreadPool threadPool,
-        final ThreadContext threadContext,
-        final ConfigurationRepository configurationRepository,
-        final IndexNameExpressionResolver resolver,
-        AuditLog auditLog,
-        final Settings settings,
-        final PrivilegesInterceptor privilegesInterceptor,
-        final ClusterInfoHolder clusterInfoHolder,
-        final IndexResolverReplacer irr,
-        NamedXContentRegistry namedXContentRegistry
-    ) {
+            final ClusterService clusterService,
+            Supplier<ClusterState> clusterStateSupplier,
+            ThreadPool threadPool,
+            final ThreadContext threadContext,
+            final ConfigurationRepository configurationRepository,
+            final IndexNameExpressionResolver resolver,
+            AuditLog auditLog,
+            final Settings settings,
+            final PrivilegesInterceptor privilegesInterceptor,
+            final ClusterInfoHolder clusterInfoHolder,
+            final IndexResolverReplacer irr,
+            NamedXContentRegistry namedXContentRegistry) {
 
         super();
         this.resolver = resolver;
@@ -184,9 +181,8 @@ public class PrivilegesEvaluator {
         this.settings = settings;
 
         this.checkSnapshotRestoreWritePrivileges = settings.getAsBoolean(
-            ConfigConstants.SECURITY_CHECK_SNAPSHOT_RESTORE_WRITE_PRIVILEGES,
-            ConfigConstants.SECURITY_DEFAULT_CHECK_SNAPSHOT_RESTORE_WRITE_PRIVILEGES
-        );
+                ConfigConstants.SECURITY_CHECK_SNAPSHOT_RESTORE_WRITE_PRIVILEGES,
+                ConfigConstants.SECURITY_DEFAULT_CHECK_SNAPSHOT_RESTORE_WRITE_PRIVILEGES);
 
         this.clusterInfoHolder = clusterInfoHolder;
         this.irr = irr;
@@ -201,10 +197,11 @@ public class PrivilegesEvaluator {
         if (configurationRepository != null) {
             configurationRepository.subscribeOnChange(configMap -> {
                 try {
-                    SecurityDynamicConfiguration<ActionGroupsV7> actionGroupsConfiguration = configurationRepository.getConfiguration(
-                        CType.ACTIONGROUPS
-                    );
-                    SecurityDynamicConfiguration<RoleV7> rolesConfiguration = configurationRepository.getConfiguration(CType.ROLES);
+                    SecurityDynamicConfiguration<ActionGroupsV7> actionGroupsConfiguration = configurationRepository
+                            .getConfiguration(
+                                    CType.ACTIONGROUPS);
+                    SecurityDynamicConfiguration<RoleV7> rolesConfiguration = configurationRepository
+                            .getConfiguration(CType.ROLES);
 
                     this.updateConfiguration(actionGroupsConfiguration, rolesConfiguration);
                 } catch (Exception e) {
@@ -225,21 +222,19 @@ public class PrivilegesEvaluator {
     }
 
     void updateConfiguration(
-        SecurityDynamicConfiguration<ActionGroupsV7> actionGroupsConfiguration,
-        SecurityDynamicConfiguration<RoleV7> rolesConfiguration
-    ) {
+            SecurityDynamicConfiguration<ActionGroupsV7> actionGroupsConfiguration,
+            SecurityDynamicConfiguration<RoleV7> rolesConfiguration) {
         if (rolesConfiguration != null) {
             SecurityDynamicConfiguration<ActionGroupsV7> actionGroupsWithStatics = actionGroupsConfiguration != null
-                ? DynamicConfigFactory.addStatics(actionGroupsConfiguration.clone())
-                : DynamicConfigFactory.addStatics(SecurityDynamicConfiguration.empty(CType.ACTIONGROUPS));
+                    ? DynamicConfigFactory.addStatics(actionGroupsConfiguration.clone())
+                    : DynamicConfigFactory.addStatics(SecurityDynamicConfiguration.empty(CType.ACTIONGROUPS));
             FlattenedActionGroups flattenedActionGroups = new FlattenedActionGroups(actionGroupsWithStatics);
             ActionPrivileges actionPrivileges = new ActionPrivileges(
-                DynamicConfigFactory.addStatics(rolesConfiguration.clone()),
-                flattenedActionGroups,
-                () -> clusterStateSupplier.get().metadata().getIndicesLookup(),
-                settings,
-                pluginToClusterActions
-            );
+                    DynamicConfigFactory.addStatics(rolesConfiguration.clone()),
+                    flattenedActionGroups,
+                    () -> clusterStateSupplier.get().metadata().getIndicesLookup(),
+                    settings,
+                    pluginToClusterActions);
             Metadata metadata = clusterStateSupplier.get().metadata();
             actionPrivileges.updateStatefulIndexPrivileges(metadata.getIndicesLookup(), metadata.version());
             ActionPrivileges oldInstance = this.actionPrivileges.getAndSet(actionPrivileges);
@@ -264,7 +259,8 @@ public class PrivilegesEvaluator {
         return this.actionPrivileges.get();
     }
 
-    public boolean hasRestAdminPermissions(final User user, final TransportAddress remoteAddress, final String permission) {
+    public boolean hasRestAdminPermissions(final User user, final TransportAddress remoteAddress,
+            final String permission) {
         PrivilegesEvaluationContext context = createContext(user, permission);
         return this.actionPrivileges.get().hasExplicitClusterPrivilege(context, permission).isAllowed();
     }
@@ -285,6 +281,19 @@ public class PrivilegesEvaluator {
             if (!Strings.isNullOrEmpty(requestedTenant)) {
                 joiner.add(escapePipe(requestedTenant));
             }
+
+            Set<String> customAttributes = new HashSet<String>();
+            for (Map.Entry<String, String> entry : user.getCustomAttributesMap().entrySet()) {
+                if (entry.getKey() == null || entry.getValue() == null) {
+                    continue;
+                }
+                customAttributes.add(entry.getKey());
+            }
+            if (customAttributes.size() > 0) {
+                joiner.add(escapePipe(String.join(",", customAttributes)));
+            }
+
+            log.debug("set user info in thread context: {}", joiner.toString());
             threadContext.putTransient(OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT, joiner.toString());
         }
     }
@@ -294,20 +303,21 @@ public class PrivilegesEvaluator {
     }
 
     public PrivilegesEvaluationContext createContext(
-        User user,
-        String action0,
-        ActionRequest request,
-        Task task,
-        Set<String> injectedRoles
-    ) {
+            User user,
+            String action0,
+            ActionRequest request,
+            Task task,
+            Set<String> injectedRoles) {
         if (!isInitialized()) {
             throw new OpenSearchSecurityException("OpenSearch Security is not initialized.");
         }
 
         TransportAddress caller = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_REMOTE_ADDRESS);
-        ImmutableSet<String> mappedRoles = ImmutableSet.copyOf((injectedRoles == null) ? mapRoles(user, caller) : injectedRoles);
+        ImmutableSet<String> mappedRoles = ImmutableSet
+                .copyOf((injectedRoles == null) ? mapRoles(user, caller) : injectedRoles);
 
-        return new PrivilegesEvaluationContext(user, mappedRoles, action0, request, task, irr, resolver, clusterStateSupplier);
+        return new PrivilegesEvaluationContext(user, mappedRoles, action0, request, task, irr, resolver,
+                clusterStateSupplier);
     }
 
     public PrivilegesEvaluatorResponse evaluate(PrivilegesEvaluationContext context) {
@@ -337,10 +347,10 @@ public class PrivilegesEvaluator {
         PrivilegesEvaluatorResponse presponse = new PrivilegesEvaluatorResponse();
 
         final String injectedRolesValidationString = threadContext.getTransient(
-            ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROLES_VALIDATION
-        );
+                ConfigConstants.OPENDISTRO_SECURITY_INJECTED_ROLES_VALIDATION);
         if (injectedRolesValidationString != null) {
-            HashSet<String> injectedRolesValidationSet = new HashSet<>(Arrays.asList(injectedRolesValidationString.split(",")));
+            HashSet<String> injectedRolesValidationSet = new HashSet<>(
+                    Arrays.asList(injectedRolesValidationString.split(",")));
             if (!mappedRoles.containsAll(injectedRolesValidationSet)) {
                 presponse.allowed = false;
                 presponse.missingSecurityRoles.addAll(injectedRolesValidationSet);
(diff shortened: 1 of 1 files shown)


src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -351,40 +361,44 @@ public PrivilegesEvaluatorResponse evaluate(PrivilegesEvaluationContext context)
...
-        // Add the security roles for this user so that they can be used for DLS parameter substitution.
+        // Add the security roles for this user so that they can be used for DLS
+        // parameter substitution.
         user.addSecurityRoles(mappedRoles);
         setUserInfoInThreadContext(user);
 
         final boolean isDebugEnabled = log.isDebugEnabled();
         if (isDebugEnabled) {
-            log.debug("Evaluate permissions for {}", user);
+            log.debug("Evaluate permissions for {}", user.toStringWithAttributes());
```
markdboyd: Not sure if this is necessary but it is helpful in debugging

cwperks:
@markdboyd I think your IDE may have some conflicts with the spotless rules in this repo. Make sure to run `./gradlew spotlessApply` to apply the spotless rules prior to push.

cwperks:
@markdboyd FYI for serialization, I was thinking something like this: https://github.com/cwperks/security/pull/59

i.e. let's base64 encode it to serialize.

I'm thinking that its safe to base64 encode as its done elsewhere ([example](https://github.com/opensearch-project/security/blob/2.19/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java#L541)) and won't clash with the pipe-delimited format.

markdboyd:
@cwperks OK, I applied your feedback on base64 encoding and ran `./gradlew spotlessApply`. Let me know if I need to do anything else.

cwperks:
Thank you @markdboyd1 Can you also make sure to sign the commits? (See [DCO check failure](https://github.com/opensearch-project/security/pull/5491/checks?check_run_id=46392705572)) Each repo has a DCO check that requires contributors to sign commits as a Developer Certificate of Origin: https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin

I think we should also add an entry in the [CHANGELOG](https://github.com/opensearch-project/security/blob/main/CHANGELOG.md) for this PR.

CHANGELOG.md
```diff
@@ -26,6 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Provide SecureHttpTransportParameters to complement SecureTransportParameters counterpart ([#5432](https://github.com/opensearch-project/security/pull/5432))
 * Use isClusterPerm instead of requestedResolved.isLocalAll() to determine if action is a cluster action ([#5445](https://github.com/opensearch-project/security/pull/5445))
 * Fix config update with deprecated config types failing in mixed clusters ([#5456](https://github.com/opensearch-project/security/pull/5456))
+* Add serialized user custom attributes to the the thread context ([#5491](https://github.com/opensearch-project/security/pull/5491))
```
cwperks, line 39: Can you remove the other changes in the CHANGELOG outside of adding this line? Otherwise, this PR LGTM!

cwperks:
Other test error is bc `ImmutableMap` is not Serializable:

> »  org.opensearch.OpenSearchException: Instance {} of class class com.google.common.collect.RegularImmutableMap is not serializable
»  	at org.opensearch.security.support.Base64Helper.serializeObject(Base64Helper.java:84) ~[?:?]
»  	at org.opensearch.security.privileges.PrivilegesEvaluator.setUserInfoInThreadContext(PrivilegesEvaluator.java:298) ~[?:?]
»  	at org.opensearch.security.privileges.PrivilegesEvaluator.evaluate(PrivilegesEvaluator.java:397) ~[?:?]

markdboyd:
@cwperks 

OK. I'll fix the CHANGELOG.

Any suggestions re: Serializable and ImmutableMap? `Base64Helper.serializeObject` wouldn't work on `user.getCustomAttributesMap()` without `Serializable`. [I took the idea from this code](https://github.com/opensearch-project/security/blob/2.19/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java#L541).

Also the common-utils PR is up and ready for review as well if that needs to go first

CHANGELOG.md
markdboyd: Done: https://github.com/opensearch-project/security/pull/5491/commits/d4887ffc6d221f0eb5ccd91ed394db5a857714ef

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -289,7 +291,12 @@ private void setUserInfoInThreadContext(User user, Set<String> mappedRoles) {
             String requestedTenant = user.getRequestedTenant();
             if (!Strings.isNullOrEmpty(requestedTenant)) {
                 joiner.add(escapePipe(requestedTenant));
+            } else {
+                joiner.add("null");
```
shikharj05: is this handled in alerting plugin?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -289,7 +291,12 @@ private void setUserInfoInThreadContext(User user, Set<String> mappedRoles) {
...
             }
+
+            joiner.add(Base64Helper.serializeObject((Serializable) user.getCustomAttributesMap()));
```
shikharj05: This can lead to performance issues depending on size of custom attributes - should this be made selective using a role configuration or authz configuration?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
markdboyd: I'm not sure. But I'll just remove this change for simplicity.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
markdboyd: Sure. Can you provide any guidance as to how that would work? I'm a novice Java developer with minimal overall knowledge of the OpenSearch codebase

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
markdboyd: https://github.com/opensearch-project/security/pull/5491/commits/125ea6bf74b036c7cb8a724a6c415fe76e8d2aa2

cwperks: It has to be handled in common-utils [here](https://github.com/opensearch-project/common-utils/blob/main/src/main/java/org/opensearch/commons/authuser/User.java#L188-L190). i.e. in addition to `String.isNullOrEmpty, it would also need to check `"null".equals(strs[3])` and the change in common-utils and this repo would need to be released in the same minor version.

cwperks: We need that change. basically, the 3rd entry is now optional and we want to add a 4th, so we have to make the 3rd non-optional. We always need to serialize userRequestedTenant and use the String "null" as a placeholder meaning null value.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
cwperks: @shikharj05 would a feature flag be useful here? IMO all attributes should be serialized bc user role definitions can change at any time.

cwperks:
@markdboyd for an example of a feature flag see: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L1365-L1373

Essentially, we'd add a setting that an admin can place in opensearch.yml to enable/disable serialization of user attributes.

i.e.

```
plugins.security.user_attribute_serialization.enabled: true
```

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
shikharj05: @cwperks do you think it's possible to look at role definition, identify the dynamic parameter substitution required and just serilalize this specific parameter?

cwperks: I think the feature flag approach would be simpler. We could also add another setting in the future for additional controls for an allowlist of which attributes to specifically include when serializing instead of serializing all.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -290,7 +302,12 @@ private void setUserInfoInThreadContext(User user, Set<String> mappedRoles) {
             if (!Strings.isNullOrEmpty(requestedTenant)) {
                 joiner.add(escapePipe(requestedTenant));
             }
-            threadContext.putTransient(OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT, joiner.toString());
+
+            joiner.add(Base64Helper.serializeObject((Serializable) user.getCustomAttributesMap()));
+
+            if (this.isUserAttributeSerializationEnabled()) {
+                threadContext.putTransient(OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT, joiner.toString());
+            }
```
cwperks:
```suggestion
            if (this.isUserAttributeSerializationEnabled()) {
                joiner.add(Base64Helper.serializeObject((Serializable) user.getCustomAttributesMap()));
            }
            threadContext.putTransient(OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT, joiner.toString());
```

cwperks: changed this to add the attribute map within the block where `isUserAttributeSerializationEnabled` and always populate the serialized user info header outside this if block.

markdboyd: https://github.com/opensearch-project/security/pull/5491/commits/a68f55cfaa733c049224b93605abc02fe2b5d627

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -290,6 +298,11 @@ private void setUserInfoInThreadContext(User user, Set<String> mappedRoles) {
             if (!Strings.isNullOrEmpty(requestedTenant)) {
                 joiner.add(escapePipe(requestedTenant));
             }
```
cwperks: We need to add the else block back in here and explicitly put the string "null" if requestedTenant is null or empty and then handle it appropriately on the deserialization end.

Currently, requestedTenant is optionally serialized so the serialized user either has 3 parts or 4 parts depending on if requestedTenant is present. We need to remove the optionality so that it is _always_ serialized since we are adding a 5th element to the serialization for the user attributes. Otherwise there will be ambiguity that cannot be resolved.

markdboyd: Makes sense. I re-added it: https://github.com/opensearch-project/security/pull/5491/commits/90a680c4efdfe0ab5c3e149ee9077bd27d4d16cd

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/SampleResourcePluginTests.java
```diff
@@ -97,6 +107,34 @@ public void testPluginInstalledCorrectly() {
         }
     }
 
+    @Test
```
DarshitChanpura: @markdboyd I apologize, but would you mind moving these tests to a separate class? I originally added this base test class for a feature that i was working on, but have since refactored it completely here (https://github.com/opensearch-project/security/pull/5408) and I would want to avoid any merge conflicts in future:

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/CreateResourceTransportAction.java
```diff
@@ -56,6 +56,7 @@ protected void doExecute(Task task, CreateResourceRequest request, ActionListene
         ThreadContext threadContext = transportService.getThreadPool().getThreadContext();
         String userStr = threadContext.getTransient(ConfigConstants.OPENSEARCH_SECURITY_USER_INFO_THREAD_CONTEXT);
         User user = User.parse(userStr);
+        System.out.println("Parsed User: " + user);
```
DarshitChanpura: should be removed.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -2222,6 +2222,15 @@ public List<Setting<?>> getSettings() {
                     Property.Final
                 )
             );
+
+            settings.add(
```
DarshitChanpura, line 2234: I don't think we need to add a feature flag for this feature. The only reason seems to be performance concern where a user could potentially have thousands of attributes. But in that case we can may be use something like Deflater and Inflater to compress attributes. Thoughts?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
DarshitChanpura: in that case we can simply call joiner.add(escapePipe(requestedTenant)). escapePipe returns empty string if input is null. https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/SecurityUtils.java#L146

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -289,7 +297,14 @@ private void setUserInfoInThreadContext(User user, Set<String> mappedRoles) {
             String requestedTenant = user.getRequestedTenant();
             if (!Strings.isNullOrEmpty(requestedTenant)) {
                 joiner.add(escapePipe(requestedTenant));
+            } else {
+                joiner.add("null");
             }
+
+            if (this.isUserAttributeSerializationEnabled()) {
+                joiner.add(Base64Helper.serializeObject((Serializable) user.getCustomAttributesMap()));
```
DarshitChanpura: Base64 encoding might still raise performance concerns if the user has thousands of attributes. Not a blocker for this PR, but we should test that scenario and add a fix in a follow-up PR.

cwperks:
Note: integ tests for the sample plugin will fail until https://github.com/opensearch-project/common-utils/pull/827 is merged.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
cwperks: yes, this works too and is neater. @markdboyd can we apply this change and also remove that special handling in the common-utils PR.

markdboyd: OK, done here: https://github.com/opensearch-project/security/pull/5491/commits/bb74e416217860804a5a12b44692b8163ba75257

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/SampleResourcePluginTests.java
markdboyd: Sure. How about this? https://github.com/opensearch-project/security/pull/5491/commits/a69b5c18b26ae613f6adeb70d7be5431e2464b24

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
markdboyd: You all understand the performance implications of these changes much better than I do, so I'll defer to you and @cwperks

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/CreateResourceTransportAction.java
markdboyd: https://github.com/opensearch-project/security/pull/5491/commits/8971760f9e2b6a31f9d71172bf4313c397118bdb

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
markdboyd: @cwperks common-utils updated here: https://github.com/opensearch-project/common-utils/pull/827/commits/ad0c98df39418fc6a5c0fcff86563f980bb877e2

DarshitChanpura (dismissed):
Thanks @markdboyd ! LGTM. lets keep this feature behind feature flag for now.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: Let's keep it behind feature flag for now.

CHANGELOG.md
```diff
@@ -55,6 +57,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+* Bump `com.google.errorprone:error_prone_annotations` from 2.38.0 to 2.40.0 ([#5447](https://github.com/opensearch-project/security/pull/5447), [#5477](https://github.com/opensearch-project/security/pull/5477))
+* Bump `io.dropwizard.metrics:metrics-core` from 4.2.32 to 4.2.33 ([#5428](https://github.com/opensearch-project/security/pull/5428))
+* Bump `org.junit.jupiter:junit-jupiter` from 5.13.2 to 5.13.3 ([#5460](https://github.com/opensearch-project/security/pull/5460))
+* Bump `org.checkerframework:checker-qual` from 3.49.4 to 3.49.5 ([#5462](https://github.com/opensearch-project/security/pull/5462))
+* Bump `com.google.googlejavaformat:google-java-format` from 1.27.0 to 1.28.0 ([#5475](https://github.com/opensearch-project/security/pull/5475))
+* Bump `commons-validator:commons-validator` from 1.9.0 to 1.10.0 ([#5476](https://github.com/opensearch-project/security/pull/5476))
+* Bumps checkstyle to 10.26.1 that fixes CVE-2025-48734 ([#5485](https://github.com/opensearch-project/security/pull/5485))
+* Bump `commons-io:commons-io` from 2.19.0 to 2.20.0 ([#5494](https://github.com/opensearch-project/security/pull/5494))
+
```
shikharj05: nit: you might need to rebase your branch?

markdboyd: Fixed

markdboyd:
@shikharj05 Thanks for the review! I rebased the PR to fix the bad CHANGELOG updates so this should be good to go.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
nibix: TBH, i am not really happy with this. Just recently, we took great effort to minimize the use of Java serialization because it can have a *huge* performance hit:

https://github.com/opensearch-project/security/pull/5212

Now doing again serialization on *every* request hop sounds very inefficient, especially for LDAP we might have quite large amounts of user attributes.

Have you considered other possible representations?

cwperks: @nibix any suggestions on serialization format?

nibix: Having thought about this a bit more, I am realizing that this is quite a tricky one 😅 

We have a number of challenges:

- I have only realized now that this is set as a transient thread context value, i.e. it won't leave the current node. (correct?) This however means that we cannot employ the object reuse strategy that is used for the user object.
- Besides performance, Java serialization actually has another issue: If I understand it correctly, this is for exchanging the user attribute information between different plugins. Is that correct? Then, the consuming plugin needs to have a dependency on Guava (that's where the serialized classes come from; it is not provided by core). Additionally, you will have the challenge that both plugins must make sure that they are always on the same Guava version . Guava does not guarantee serialization compatibiltiy between different versions. In OpenSearch, it would be otherwise technically totally possible that plugin A is on one Guava version while plugin B has its own Guava version. This could lead to unexpected breaks.

How about using plain JSON for the user attributes? That would fix all interoperability concerns.

Regarding the performance: How about keeping a user object to `user|object|represenation` cache similar to the one we use for the serialized user object:

https://github.com/opensearch-project/security/blob/2dc126f2d1482bf90374ae377d4511c1c462568d/src/main/java/org/opensearch/security/user/UserFactory.java#L85-L96

This would remove the need to rebuild the potentially use string for each and every request.

nibix: Instead of a cache, we could also have a transient attribute in the `User` object. It should be transient because otherwise the serialized size of the User object would grow to more than twice the size. 

But, it could lazily construct the string representation and cache it locally.

cwperks: > Besides performance, Java serialization actually has another issue: If I understand it correctly, this is for exchanging the user attribute information between different plugins. Is that correct? Then, the consuming plugin needs to have a dependency on Guava (that's where the serialized classes come from; it is not provided by core).

Yes, it would be consumed by plugins through common-utils.

See the corresponding common-utils PR: https://github.com/opensearch-project/common-utils/pull/827

plugins would only need a dependency on common-utils, not on guava and any plugin consuming this change already has a dep on common-utils.

cwperks: FYI this code is executed at authorization time, so typically a plugin will have an API like `POST _plugins/_alerting/monitors` and as part of handling the request we authorize the corresponding transport action, serialize the userinfo in the threadcontext where alerting can then pick it up by calling [readUserFromThreadContext](https://github.com/opensearch-project/alerting/blob/ae6436aa914b743bb3c7a1269f9301aec5995cb7/alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt#L46-L50) and then persist the XContent representation of the user along with the metadata for monitor.

A lot of the community is probably aware at this point that I'm not a fan of storing user copies, but in lieu of a replacement I support extending this existing pattern for new use cases like users creating monitors where the user that creates the monitor is mapped to a role with DLS restrictions utilizing user attribute substitution.

Ultimately, for a replacement to roles injection I'd prefer that the user creating a monitor pass a token that has only the permissions necessary for the monitor to run and then security enforces that on every instance that a background job runs.

nibix: ok, so i would follow the assessment of @DarshitChanpura : 

> Not a blocker for this PR, but we should test that scenario and add a fix in a follow-up PR.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -292,10 +300,20 @@ private void setUserInfoInThreadContext(PrivilegesEvaluationContext context) {
             joiner.add(escapePipe(String.join(",", context.getMappedRoles())));
 
             String requestedTenant = context.getUser().getRequestedTenant();
-            joiner.add(requestedTenant);
```
cwperks, line 295: Can we keep this as-is?

markdboyd: https://github.com/opensearch-project/security/pull/5491/commits/47dbec76e46c9ef85e5f3b526eed34f88b7405af

cwperks (approved):

PR #5492, opened 2025-07-19
Restricting star tree index for users with DLS/FLS/FieldMasking restrictions

bharath-techie:
### Description
Restricts usage of star tree index when DLS/FLS/FieldMasking restrictions are applied for user for a particular index.

As part of this PR , we set the `StarTreeQueryContext` to null when DLS/FLS/FieldMasking restrictions are present, so that the query falls back to the default aggregations behavior. We'll just skip using star tree for optimizations.

### Issues Resolved

### Testing

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
+++ b/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
@@ -75,7 +75,6 @@ import org.opensearch.security.privileges.dlsfls.DlsFlsLegacyHeaders;
 import org.opensearch.security.privileges.dlsfls.DlsFlsProcessedConfig;
 import org.opensearch.security.privileges.dlsfls.DlsRestriction;
 import org.opensearch.security.privileges.dlsfls.FieldMasking;
-import org.opensearch.security.privileges.dlsfls.FieldPrivileges;
 import org.opensearch.security.privileges.dlsfls.IndexToRuleMap;
 import org.opensearch.security.resolver.IndexResolverReplacer;
 import org.opensearch.security.securityconf.DynamicConfigFactory;
@@ -409,9 +408,9 @@ public class DlsFlsValveImpl implements DlsFlsRequestValve {
 
             // Restrict access for star tree index if there are any DLS/FLS/field masking restrictions
             if (searchContext.getQueryShardContext().getStarTreeQueryContext() != null) {
-                FieldPrivileges.FlsRule fls = config.getFieldPrivileges().getRestriction(privilegesEvaluationContext, index);
-                FieldMasking.FieldMaskingRule rule = config.getFieldMasking().getRestriction(privilegesEvaluationContext, index);
-                if (!dlsRestriction.isUnrestricted() || !fls.isUnrestricted() || !rule.isUnrestricted()) {
+                boolean flsUnrestricted = config.getFieldPrivileges().isUnrestricted(privilegesEvaluationContext, index);
+                boolean fieldMaskingUnrestricted = config.getFieldMasking().isUnrestricted(privilegesEvaluationContext, index);
+                if (!dlsRestriction.isUnrestricted() || !flsUnrestricted || !fieldMaskingUnrestricted) {
                     searchContext.getQueryShardContext().setStarTreeQueryContext(null);
                 }
             }
--- a/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
+++ b/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
@@ -75,6 +75,7 @@ import org.opensearch.security.privileges.dlsfls.DlsFlsLegacyHeaders;
 import org.opensearch.security.privileges.dlsfls.DlsFlsProcessedConfig;
 import org.opensearch.security.privileges.dlsfls.DlsRestriction;
 import org.opensearch.security.privileges.dlsfls.FieldMasking;
+import org.opensearch.security.privileges.dlsfls.FieldPrivileges;
 import org.opensearch.security.privileges.dlsfls.IndexToRuleMap;
 import org.opensearch.security.resolver.IndexResolverReplacer;
 import org.opensearch.security.securityconf.DynamicConfigFactory;
@@ -385,7 +386,6 @@ public class DlsFlsValveImpl implements DlsFlsRequestValve {
             if (privilegesEvaluationContext == null) {
                 return;
             }
-
             DlsFlsProcessedConfig config = this.dlsFlsProcessedConfig.get();
 
             DlsRestriction dlsRestriction = config.getDocumentPrivileges().getRestriction(privilegesEvaluationContext, index);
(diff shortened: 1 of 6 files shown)


bharath-techie:
> > Restricts usage of star tree index when DLS/FLS/FieldMasking restrictions are applied for user for a particular index
> 
> Can you expand on the intended behavior in the PR description? i.e. will it reject the request outright (like we do with write operations when a user has DLS/FLS/FieldMasking restrictions) or will it disable star tree and behave like other aggs?
> 
> I know when we talked on slack it was the latter, but the PR description should reflect that as well.

Sure, updated the description to reflect the behavior @cwperks

mgodwan (approved):

src/test/java/org/opensearch/security/dlic/dlsfls/StarTreeDlsFlsTest.java
```diff
@@ -0,0 +1,476 @@
...
+                        Settings.builder()
+                            .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1)
+                            .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 0)
+                            .put("index.composite_index", true)
+                            .put("index.append_only.enabled", true)
+                            .put("index.search.star_tree_index.enabled", true)
+                            .build()
+                    )
+                        .mapping(
+                            XContentFactory.jsonBuilder()
```
cwperks: nit: any chance we can convert this to raw json and use a text block for readability?

src/test/java/org/opensearch/security/dlic/dlsfls/StarTreeDlsFlsTest.java
```diff
@@ -0,0 +1,476 @@
...
+                .source(
+                    "{\"department\": \"engineering\", \"region\": \"us-west\", \"sales_amount\": 1000.0, \"employee_id\": \"emp1\", \"sensitive_data\": \"confidential info 1\"}",
+                    XContentType.JSON
+                )
+        ).actionGet();
+
+        tc.index(
+            new IndexRequest(STARTREE_INDEX_NAME).setRefreshPolicy(RefreshPolicy.IMMEDIATE)
+                .source(
+                    "{\"department\": \"engineering\", \"region\": \"us-east\", \"sales_amount\": 1500.0, \"employee_id\": \"emp2\", \"sensitive_data\": \"confidential info 2\"}",
```
cwperks: nit: prefer to use text blocks for readability

cwperks: It would also be nice to use `_bulk` here

cwperks (approved):

src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
```diff
@@ -407,6 +407,15 @@ public void handleSearchContext(SearchContext searchContext, ThreadPool threadPo
                 }
             }
 
+            // Restrict access for star tree index if there are any DLS/FLS/field masking restrictions
+            if (searchContext.getQueryShardContext().getStarTreeQueryContext() != null) {
+                FieldPrivileges.FlsRule fls = config.getFieldPrivileges().getRestriction(privilegesEvaluationContext, index);
+                FieldMasking.FieldMaskingRule rule = config.getFieldMasking().getRestriction(privilegesEvaluationContext, index);
+                if (!dlsRestriction.isUnrestricted() || !fls.isUnrestricted() || !rule.isUnrestricted()) {
+                    searchContext.getQueryShardContext().setStarTreeQueryContext(null);
```
nagarajg17, line 414: Is `StarTreeQueryContext` used anywhere else post DLS/FLS check? Asking because we recently had bug in OS 1.3 for doc corruption issue which was trying to fix some problem inline (very similar to this) which resulted in another problem. @mgodwan is aware of it if you need any context

sandeshkr419:
Thanks @nagarajg17 for taking a look.

> Why do we need to fall back to the default aggregations behaviour when DLS/FLS/FieldMasking restrictions are present? What is the current behaviour when we don't?

The default aggregations take into account and DLS/FLS/Fieldmasking as they iterate through doc values. Any pre-computations, specifically with star-tree index do not preserve doc values of individual documents and the resulting composite index structure is an aggregated index structure aggregated based on fields, which could not incorporate any filters based on document (doc values) alone.

In future, we might be able to do some work-arounds to make star-tree aggregation possible for FLS & fieldmasking, but not for DLS for sure since star-tree lose document identities during its creation. As a safety net for now, so that DLS/FLS/Fieldmasking can apply as intended, this code change directs the flow back to default by setting `starTreeContext` to `null` - in that way any aggregation realizes that they are not supposed to use star-tree for aggregations for such requests.

src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
sandeshkr419: `searchContext.getQueryShardContext().setStarTreeQueryContext(null)` is a safe change in a way that star-tree is used only for aggregations. StarTreeQueryContext is a way to indicate aggregations that they can utilize star-tree to pre-compute the aggregations. StarTreeQueryContext as null would imply that star-tree cannot be used for star-tree pre-computation and the aggregation will have to utilize doc values to finally compute the aggragtions.

src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
```diff
@@ -407,6 +407,15 @@ public void handleSearchContext(SearchContext searchContext, ThreadPool threadPo
                 }
             }
 
+            // Restrict access for star tree index if there are any DLS/FLS/field masking restrictions
+            if (searchContext.getQueryShardContext().getStarTreeQueryContext() != null) {
+                FieldPrivileges.FlsRule fls = config.getFieldPrivileges().getRestriction(privilegesEvaluationContext, index);
+                FieldMasking.FieldMaskingRule rule = config.getFieldMasking().getRestriction(privilegesEvaluationContext, index);
+                if (!dlsRestriction.isUnrestricted() || !fls.isUnrestricted() || !rule.isUnrestricted()) {
```
nibix: Nit: As you only need the information whether something is restricted or not, you can also use the method `isUnrestricted()`: https://github.com/opensearch-project/security/blob/7262ec00696886e22e81a26e6e2a1450e2de8bcd/src/main/java/org/opensearch/security/privileges/dlsfls/AbstractRuleBasedPrivileges.java#L188

As this won't need to collect rules, this might have a small performance advantage.

bharath-techie: Thanks have done this

src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
bharath-techie: Resolving this as @mgodwan is already aware of this change.

bharath-techie: Just to explain further, this null set happens during pre query phase.

There is a check for StarTreeQueryContext != null  in the actual query phase , only if its non null, star tree gets used in the query. Otherwise fall back to default aggregations.

mgodwan: LGTM

bharath-techie:
All conditions are evaluated via IT. So codecoverage is a bit misleading

src/test/java/org/opensearch/security/dlic/dlsfls/StarTreeDlsFlsTest.java
```diff
@@ -0,0 +1,453 @@
...
+              }
+            }
+            """;
+
+        HttpResponse response = rh.executePostRequest(
+            "/" + STARTREE_INDEX_NAME + "/_search?pretty",
+            aggregationQuery,
+            encodeBasicHeader("startree_dls_user", "password")
+        );
+        assertThat(response.getStatusCode(), is(HttpStatus.SC_OK));
```
DarshitChanpura, line 307: can you add the aggregation response assertion. Explicitly what numbers are expected?

DarshitChanpura (commented):
@bharath-techie I have 1 comment around strengthening test assertion to assert what is expected aggregation response when star-tree optimization is disabled. PR looks good otherwise. Also, is there a relevant issue for this fix?

nagarajg17 (approved):

bharath-techie:
> @bharath-techie I have 1 comment around strengthening test assertion to assert what is expected aggregation response when star-tree optimization is disabled. PR looks good otherwise. Also, is there a relevant issue for this fix?

I have addressed the comment. 
Since star tree is a pre-computed data structure separate from the original Lucene fields / docs , we are restricting the usage of star tree index for queries for the users who has index restrictions.
https://github.com/opensearch-project/OpenSearch/issues/12498 is the issue for star tree index.

DarshitChanpura (approved):

PR #5496, opened 2025-07-21
Add url query param (has_permission_check=true) to check if a user has access to call an API

pranathz:
### Description

This PR Adds support to simulate permission checks on existing REST APIs using a query parameter.
* Category (New feature)
 Previously, there was no way to preview whether a specific action request would be allowed or denied without actually executing it. This change introduces a safe way to check permissions in advance by simply adding a query parameter (simulate_authz=true) to the existing APIs.
   Old Behavior: No support to preview permission results; requests were either executed or blocked.
   New Behavior: When simulate_authz=true is present, the request does not execute but instead returns whether the action would be allowed and lists any missing privileges.

### Issues Resolved
Related to [5468](https://github.com/opensearch-project/security/issues/5468)

**API Design:**
You can add the simulate_authz parameter to any existing REST endpoint.
For e.g.,    `GET /_cluster/health?simulate_authz=true
`

Response format if action is allowed
```
{
  "accessAllowed" : true,
  "missingPrivileges" : [ ]
}
```
Response format if action is not allowed

```
{
  "accessAllowed" : false,
  "missingPrivileges" : ["cluster:monitor/health"]
}
```
### Testing
Integration test included 

### Check List


src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -388,6 +389,21 @@ private <Request extends ActionRequest, Response extends ActionResponse> void ap
                 log.debug(pres.toString());
             }
 
+            String isSimulation = threadContext.getHeader(ConfigConstants.SECURITY_SIMULATE_AUTHZ_PARAM);
+            if ("true".equals(isSimulation)) {
```
nagarajg17: nit: We can just convert isSimulation to Boolean

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -159,6 +159,11 @@ public class ConfigConstants {
     public static final String SECURITY_AUTHCZ_IMPERSONATION_DN = SECURITY_SETTINGS_PREFIX + "authcz.impersonation_dn";
     public static final String SECURITY_AUTHCZ_REST_IMPERSONATION_USERS = SECURITY_SETTINGS_PREFIX + "authcz.rest_impersonation_user";
 
+    public static final String SECURITY_SIMULATION_API_ENABLED = SECURITY_SETTINGS_PREFIX + "simulation_api.enabled";
+    public static final boolean SECURITY_SIMULATION_API_ENABLED_DEFAULT = true;
```
nagarajg17: Is it used anywhere?

src/test/java/org/opensearch/security/filter/SecurityRestFilterTests.java
```diff
@@ -300,4 +302,38 @@ public void testAllowlistedApiWithoutExtraSlash() throws Exception {
...
+            nonAdminCredsHeader
+        );
+
+        rh.sendAdminCertificate = false;
+        response = rh.executeGetRequest("_cluster/health?" + ConfigConstants.SECURITY_SIMULATE_AUTHZ_PARAM + "=true", nonAdminCredsHeader);
+        assertThat(response.getBody(), response.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertTrue(response.getBody().contains("\"accessAllowed\":"));
+        assertTrue(response.getBody().contains("\"missingPrivileges\":"));
+
+        response = rh.executeGetRequest("_search?" + ConfigConstants.SECURITY_SIMULATE_AUTHZ_PARAM + "=true", nonAdminCredsHeader);
```
nagarajg17: `_search` is not allowlisted like above for `_cluster/health`?

src/test/java/org/opensearch/security/filter/SecurityRestFilterTests.java
```diff
@@ -300,4 +302,38 @@ public void testAllowlistedApiWithoutExtraSlash() throws Exception {
...
+
+    /**
+     * Tests that the simulate_authz param works correctly.
+     * When simulate_authz=true is added to a request, returns
+     * whether the request would be allowed or Denied, without actually executing the request.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testSimulateAuthzParam() throws Exception {
```
nagarajg17: - We can add one test with value not as true?
- One more test - without this query param this new code flow shouldn't be invoked (to test there is no regression)

src/main/java/org/opensearch/security/filter/SecurityFilter.java
shikharj05: nice! maybe extract all this handling out to a different Function.

src/main/java/org/opensearch/security/support/ConfigConstants.java
pranathz: Not used, will remove in the next commit.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
pranathz: Ack.

pranathz: Ack.

src/main/java/org/opensearch/security/support/ConfigConstants.java
cwperks: Agree, I don't think feature flag is needed here.

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -159,6 +159,11 @@ public class ConfigConstants {
...
+
+    public static final String SECURITY_SIMULATE_AUTHZ_PARAM = "simulate_authz";
```
cwperks: IMO I'd prefer to name this `has_permission` or `can_perform` since it is not a simulation, but rather asking whether the user has permission without actually performing the action.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
cwperks: This looks good for the general case, but there may be a few edge cases that we need to test out or consider as well:

1. I think this will be ignored for superuser because their requests never hit here. We can either choose to exit early [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/filter/SecurityRestFilter.java#L170-L175) or have a log message saying that the param is ignored for superuser
2. For users with DLS/FLS/Field Masking restrictions, should we also consider returning what the restrictions may be if the simulate request would apply the restrictions? i.e. user has DLS restrictions on an index so they are forbidden from writing even if they have an invalid role that says they can write to an index.

src/test/java/org/opensearch/security/filter/SecurityRestFilterTests.java
pranathz: correct, I intentionally didn’t add `_search` to the allowlist to ensure it returns a forbidden error. This validates the negative scenario where the action is not allowlisted

src/test/java/org/opensearch/security/filter/SecurityRestFilterTests.java
pranathz: Added both tests.

src/main/java/org/opensearch/security/support/ConfigConstants.java
nagarajg17: `has_permission=true` doesn’t imply the intent i.e checking permissions, instead gives meaning like permission is there/true. How about `has_permissions_check` or `has_permissions_validate`? Intent is clear here without ambiguity

src/main/java/org/opensearch/security/filter/SecurityFilter.java
DarshitChanpura: Based on SecurityRestFilter this header will only be present if simulate_authz param is present in request. We can simplify the if block to `!Strings.isNullOrEmpty(isSimulation)`.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -164,6 +164,10 @@ public void handleRequest(RestRequest request, RestChannel channel, NodeClient c
                 return;
             }
 
+            boolean isSimulation = request.paramAsBoolean(ConfigConstants.SECURITY_SIMULATE_AUTHZ_PARAM, false);
+            if (isSimulation) {
+                threadContext.putHeader(ConfigConstants.SECURITY_SIMULATE_AUTHZ_PARAM, Boolean.TRUE.toString());
+            }
```
DarshitChanpura, line 174: Have we tested the behavior in multi-node scenario?

src/main/java/org/opensearch/security/support/ConfigConstants.java
DarshitChanpura: how about `dry_run`?

src/main/java/org/opensearch/security/support/ConfigConstants.java
DarshitChanpura: this feature flag is not needed. We can remove this

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -164,6 +164,10 @@ public void handleRequest(RestRequest request, RestChannel channel, NodeClient c
                 return;
             }
 
+            boolean isSimulation = request.paramAsBoolean(ConfigConstants.SECURITY_SIMULATE_AUTHZ_PARAM, false);
```
DarshitChanpura: no need to check paramAsBoolean. We can simply consume param and if the param exists, we know its simulation, or a dry_run, and save some network bytes :D

pranathz: okay, I'll update

src/main/java/org/opensearch/security/filter/SecurityFilter.java
pranathz: > For users with DLS/FLS/Field Masking restrictions, should we also consider returning what the restrictions may be if the simulate request would apply the restrictions? i.e. user has DLS restrictions on an index so they are forbidden from writing even if they have an invalid role that says they can write to an index.

DLS is a read-time control.it does not apply to write operations. So, even if a user has DLS restrictions on an index, they are not forbidden from writing if their role grants write permissions. For your example, even if we wanted to return what the restrictions would be, it wouldn’t make sense for write operations, since DLS has no effect on writes.

src/main/java/org/opensearch/security/support/ConfigConstants.java
pranathz: I agree with Nagaraj’s point — has_permission=true sounds more like stating a fact rather than performing a check. IMO, `has_permissions_check` feels the more clearer and unambiguous, it clearly conveys that we are checking permissions without actually performing the action.

src/test/java/org/opensearch/security/filter/SecurityRestFilterTests.java
```diff
@@ -300,4 +303,79 @@ public void testAllowlistedApiWithoutExtraSlash() throws Exception {
...
+     * @throws Exception
+     */
+    @Test
+    public void testHasPermissionCheckParam() throws Exception {
+        setup();
+
+        // ALLOWLIST GET /_cluster/health
+        rh.keystore = "restapi/kirk-keystore.jks";
+        rh.sendAdminCertificate = true;
+        response = rh.executePutRequest(
```
cwperks: Does this test do anything with the first response?

src/test/java/org/opensearch/security/filter/SecurityRestFilterTests.java
```diff
@@ -300,4 +303,79 @@ public void testAllowlistedApiWithoutExtraSlash() throws Exception {
...
+        );
+
+        rh.sendAdminCertificate = false;
+        response = rh.executeGetRequest("_search?" + ConfigConstants.SECURITY_HAS_PERMISSION_CHECK_PARAM + "=false", nonAdminCredsHeader);
+        assertThat(response.getBody(), response.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertFalse(response.getBody().contains("\"accessAllowed\":"));
+        assertFalse(response.getBody().contains("\"missingPrivileges\":"));
+
+        response = rh.executeGetRequest("_cat/nodes?" + ConfigConstants.SECURITY_HAS_PERMISSION_CHECK_PARAM + "=true", nonAdminCredsHeader);
+        assertThat(response.getBody(), response.getStatusCode(), equalTo(HttpStatus.SC_FORBIDDEN));
```
cwperks: Why is Forbidden expected here instead of the response saying allowed is false with the list of missing permissions?

src/main/java/org/opensearch/security/action/simulate/PermissionCheckResponse.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public PermissionCheckResponse(final StreamInput in) throws IOException {
+        this.accessAllowed = in.readBoolean();
+        this.missingPrivileges = in.readSet(StreamInput::readString);
+    }
+
+    @Override
+    public RestStatus status() {
+        return accessAllowed ? RestStatus.OK : RestStatus.FORBIDDEN;
```
DarshitChanpura: Status should always be OK since the request was successful. The response will indicate whether action is allowed.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -507,6 +511,31 @@ private void attachSourceFieldContext(ActionRequest request) {
...
+        String isSimulation = threadContext.getHeader(ConfigConstants.SECURITY_HAS_PERMISSION_CHECK_PARAM);
+        if (Boolean.parseBoolean(isSimulation)) {
+
+            try {
+
+                @SuppressWarnings("unchecked")
+                Response response = (Response) new PermissionCheckResponse(pres.isAllowed(), pres.getMissingPrivileges());
+                listener.onResponse(response);
+
+                log.info("accessAllowed {}", pres.isAllowed());
```
DarshitChanpura: these logs should be changed to debug level and should contain information about what action was simulation run on.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -164,12 +167,21 @@ public void handleRequest(RestRequest request, RestChannel channel, NodeClient c
...
+                threadContext.putHeader(ConfigConstants.SECURITY_HAS_PERMISSION_CHECK_PARAM, Boolean.TRUE.toString());
+            }
             // Authorize Request
             final User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
             String intiatingUser = threadContext.getTransient(OPENDISTRO_SECURITY_INITIATING_USER);
             if (userIsSuperAdmin(user, adminDNs)) {
                 // Super admins are always authorized
                 auditLog.logSucceededLogin(user.getName(), true, intiatingUser, requestChannel);
+                if (hasPermissionCheck) {
+                    log.info("Permission check skipped: Super admin has full access");
```
DarshitChanpura: log level should be changed to debug.

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -159,6 +159,8 @@ public class ConfigConstants {
     public static final String SECURITY_AUTHCZ_IMPERSONATION_DN = SECURITY_SETTINGS_PREFIX + "authcz.impersonation_dn";
     public static final String SECURITY_AUTHCZ_REST_IMPERSONATION_USERS = SECURITY_SETTINGS_PREFIX + "authcz.rest_impersonation_user";
 
+    public static final String SECURITY_HAS_PERMISSION_CHECK_PARAM = "has_permission_check";
```
DarshitChanpura, line 162: this is not a configurable constant. it should be moved to the handler class

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -507,6 +511,31 @@ private void attachSourceFieldContext(ActionRequest request) {
...
+                log.info("missing privileges {}", pres.getMissingPrivileges());
+
+            } catch (Exception e) {
+                log.error("Error creating Permission Check response", e);
+                listener.onFailure(new OpenSearchException("Error creating Permission Check response", e));
```
DarshitChanpura: i don't think catch block is needed as we are not expecting any checked exceptions here.

src/main/java/org/opensearch/security/action/simulate/PermissionCheckResponse.java
pranathz: Got it, I'll update

src/main/java/org/opensearch/security/filter/SecurityFilter.java
pranathz: Okay, will change it to debug level

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
pranathz: Sure, will update it to debug level

src/main/java/org/opensearch/security/support/ConfigConstants.java
pranathz: Okay, will move it to the handler class

src/test/java/org/opensearch/security/filter/SecurityRestFilterTests.java
pranathz: No, it doesn’t use the first response. I’ll update the test

src/main/java/org/opensearch/security/filter/SecurityFilter.java
pranathz: We still need the catch block because a `ClassCastException` (an unchecked exception) can occur when casting to Response

DarshitChanpura: we only reach inside if block if its a simulation request and if its a simulation request then the expected response is PermissionsCheckResponse and so we would never trigger this for that response.

src/test/java/org/opensearch/security/filter/SecurityRestFilterTests.java
```diff
@@ -300,4 +303,67 @@ public void testAllowlistedApiWithoutExtraSlash() throws Exception {
...
+
+        rh.keystore = "restapi/kirk-keystore.jks";
+        rh.sendAdminCertificate = true;
+        response = rh.executeGetRequest(
+            "_cluster/health?has_permission_check=true",
+            nonAdminCredsHeader
+        );
+        rh.sendAdminCertificate = false;
+        assertThat(response.getBody(), response.getStatusCode(), equalTo(HttpStatus.SC_OK));
+        assertTrue(response.getBody().contains("\"accessAllowed\":"));
```
DarshitChanpura: can you check the actual value as well to see if contains true or false, try a different endpoint where it is explicitly false and contains missingPrivileges.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
DarshitChanpura: Tests use DEFAULT configuration which 1 m, 2d nodes. It satisfies multi-node scenario. resolving comment

src/test/java/org/opensearch/security/filter/SecurityRestFilterTests.java
```diff
@@ -300,4 +303,67 @@ public void testAllowlistedApiWithoutExtraSlash() throws Exception {
...
+        assertTrue(response.getBody().contains("\"missingPrivileges\":"));
+
+
+        rh.sendAdminCertificate = true;
```
cwperks, line 352: Is the following block related to the feature? IMO I don't think its necessary to call a separate API and assert that it doesn't contain the `accessAllowed` or `missingPrivileges` in the body. There are many other tests that cover general behavior.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
pranathz: got it, will change

src/test/java/org/opensearch/security/filter/SecurityRestFilterTests.java
pranathz: Added to verify has_permission_check=false triggers normal flow and doesn’t affect existing behavior.

src/test/java/org/opensearch/security/filter/SecurityRestFilterTests.java
pranathz: Screenshot 2025-07-25 at 12 01 14 PM

nagarajg17 (approved):

cwperks (dismissed):

cwperks (approved):

PR #5510, opened 2025-07-24
Update ppl_full_access with new required permissions

vamsimanohar:
### Description
* Category : Enhancement
PPL started using create/delete PIT APIs for execution and ppl_full_access is not reflecting those permissions. This PR address the gap.

### Testing
Manual testing done with above permissions and requests are working fine.

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Introduced new experimental versioned security configuration management feature ([#5357] (https://github.com/opensearch-project/security/pull/5357))
 * [Resource Sharing] Adds migrate API to move resource-sharing info to security plugin ([#5389](https://github.com/opensearch-project/security/pull/5389))
 * Introduces support for the Argon2 Password Hashing Algorithm ([#5441] (https://github.com/opensearch-project/security/pull/5441))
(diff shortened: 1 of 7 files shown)


cwperks (approved):
@vamsimanohar I'm approving this PR, but I think we should hide the internal implementation from a user by executing these actions in the plugin's system context.

i.e. stash the threadContext then execute the actions

```
try (ThreadContext.StoredContext ctx = threadContext.stashContext()) {
    ...run actions here...
}
```

DarshitChanpura (commented):
@vamsimanohar qq, shouldn't `manage_point_in_time` action-group cover these? Or do we not want readll and segments permission added as part of refresh?

Also, would you mind adding a CHANGELOG entry for this PR under maintenance section.

RyanL1997 (approved):

vamsimanohar:
@cwperks  SQL Plugin has few use cases where PIT management within plugin is not feasible right now.

* We have support for cross cluster search queries where we are creating PIT across cluster where plugin permissions are not working.
* We are additionally working on making sql plugin as a library which works on top of RestClient without plugin environment.

These two use cases are making the changes much more involved. So, we still want to continue with existing permission model and revisit at a later time.



@DarshitChanpura  `ppl_full_access` is a default role not an action group?. I didn't get your concern actually? Are you suggesting to use point_in_time action group inside the default role?

shikharj05:
@vamsimanohar is this required?

PR #5530, opened 2025-08-02
Integration of security with stream transport

rishabhmaurya:
### Description
With introduction of stream transport in OpenSearch (https://github.com/opensearch-project/OpenSearch/pull/18722), there could be multiple responses for a transport request, thus security plugin need to ensure the headers are injected for each response. Currently, stream transport is only available for node-node communication. 

The new flight based stream transport doesn't use DirectChannel for local node, thus this PR also addresses bypassing optimization done for local node. 

### Testing

- Manually tested using scripts here -  https://gist.github.com/rishabhmaurya/f714a87ac775a70b0632c5e9e9ea15f4 
- Unit tests
### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
+++ b/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
@@ -162,10 +162,9 @@ public class SecurityInterceptor {
         );
 
         final boolean isDebugEnabled = log.isDebugEnabled();
-
-        final boolean isSameNodeRequest = localNode != null
-            && localNode.equals(connection.getNode())
-            && (options.type() == null || !options.type().equals(TransportRequestOptions.Type.STREAM));
+        final boolean isStreamChannel = options != null && TransportRequestOptions.Type.STREAM.equals(options.type());
+        // skip the same node optimization for stream transport which doesn't use DirectChannel and thus ser/de is needed
+        final boolean isSameNodeRequest = localNode != null && localNode.equals(connection.getNode()) && !isStreamChannel;
 
         try (ThreadContext.StoredContext stashedContext = getThreadContext().stashContext()) {
             final TransportResponseHandler<T> restoringHandler = new RestoringTransportResponseHandler<T>(handler, stashedContext);
--- a/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
+++ b/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
@@ -73,6 +73,7 @@ import org.opensearch.transport.TransportRequest;
 import org.opensearch.transport.TransportRequestHandler;
 import org.opensearch.transport.TransportRequestOptions;
 import org.opensearch.transport.TransportResponseHandler;
+import org.opensearch.transport.stream.StreamTransportResponse;
 
 public class SecurityInterceptor {
 
@@ -162,7 +163,9 @@ public class SecurityInterceptor {
 
         final boolean isDebugEnabled = log.isDebugEnabled();
 
-        final boolean isSameNodeRequest = localNode != null && localNode.equals(connection.getNode());
+        final boolean isSameNodeRequest = localNode != null
+            && localNode.equals(connection.getNode())
+            && (options.type() == null || !options.type().equals(TransportRequestOptions.Type.STREAM));
 
         try (ThreadContext.StoredContext stashedContext = getThreadContext().stashContext()) {
             final TransportResponseHandler<T> restoringHandler = new RestoringTransportResponseHandler<T>(handler, stashedContext);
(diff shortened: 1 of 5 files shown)


rishabhmaurya:
@andrross @finnegancarroll @cwperks

src/main/java/org/opensearch/security/transport/SecureStreamTransportResponse.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public T nextResponse() {
+        T response = delegate.nextResponse();
+        // TODO: validate if it can be restored multiple times
+        storedContext.restore();
+        securityInterceptor.processResponseHeaders(response, threadPool.getThreadContext());
+        return response;
+    }
```
nibix: The processing of headers like `OPENDISTRO_SECURITY_DLS_QUERY_HEADER` is only necessary for cross cluster search if the remote cluster is on OpenSearch 2.19.x or earlier. It is not necessary if we are communicating with a node which is on OpenSearch 3.0 or newer.

If there are no plans to backport the stream transport to OpenSearch 2.19.x, I would guess that we actually do not need this handling for stream transport. Wdyt?

nibix (commented):
Thank you for this! It looks pretty good - still I am wondering whether we actually need most of it. See the inline comment below.

src/main/java/org/opensearch/security/transport/SecureStreamTransportResponse.java
```diff
@@ -0,0 +1,84 @@
+/*
```
cwperks: We can just use the OpenSearch license headers for new files.

src/main/java/org/opensearch/security/transport/SecureStreamTransportResponse.java
rishabhmaurya: this makes sense, we have no plans to backport it to 2.x. Removed!

src/main/java/org/opensearch/security/transport/SecureStreamTransportResponse.java
rishabhmaurya: Removed this class

DarshitChanpura (dismissed):

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -162,7 +163,9 @@ public <T extends TransportResponse> void sendRequestDecorate(
 
         final boolean isDebugEnabled = log.isDebugEnabled();
 
-        final boolean isSameNodeRequest = localNode != null && localNode.equals(connection.getNode());
+        final boolean isSameNodeRequest = localNode != null
+            && localNode.equals(connection.getNode())
+            && (options.type() == null || !options.type().equals(TransportRequestOptions.Type.STREAM));
```
cwperks: Why is this change needed? Wouldn't behavior for sameNodeRequest be the same for any transport? i.e. retain the existing threadcontext headers and avoid ser/de?

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
```diff
@@ -130,7 +130,7 @@ protected void messageReceivedDecorate(
 
             String channelType = transportChannel.getChannelType();
 
-            if (!channelType.equals("direct") && !channelType.equals("transport")) {
+            if (!channelType.equals("direct") && !channelType.equals("transport") && !channelType.equals("stream-transport")) {
```
cwperks: Let's create a set of these values and then check that the channelType is not within the set.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
rishabhmaurya: we do not have direct channel support (to avoid ser/de) yet in stream transport, once we have it, then we can enable it here.

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
rishabhmaurya: fixed

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
nibix: I would recommend to add a comment explaining this, or possibly slightly refactoring this. This is a bit confusing, as the intialized variable is called `isSameNodeRequest` - and the fact that this request comes from the same node does not change. Just the channel changes.

DarshitChanpura: Can you separate the clauses out please:
```java
final boolean isSameNodeRequest = localNode != null
            && localNode.equals(connection.getNode());
final boolean shouldSkipIfNotStreamChannel =  options.type() == null || !options.type().equals(TransportRequestOptions.Type.STREAM));     
 
```            
and then use these?

rishabhmaurya: done

DarshitChanpura (approved):

PR #5504, opened 2025-07-23
Catches the null pointer exception that occurs when sorting fields on geo point.

ajleong623:
### Description
This fixes the bug where a null pointer exception occurs when a sort is run on geo point fields. 

### Issues Resolved
#5503 

### Check List
- [ ] New functionality includes testing
- [ ] New functionality has been documented
- [ ] New Roles/Permissions have a corresponding security dashboards plugin PR
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md)
- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/security/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #5537, opened 2025-08-04
Handle cases gracefully where null field name is passed to LeafReader methods

nibix:
### Description

In a couple of cases, OpenSearch calls `LeafReader` methods with null `fieldName` parameters. The semantics of this are a bit unclear; in most cases, the underlying `LeafReader` methods will just also return `null`. However, I cannot prove that this will be always the case. 

Thus, the change in this PR does not try to put semantics into this; it just handles the `null` case gracefully by using a NPE safe replacement value. That has the effect that users with high restrictions will also get an restriction enforced on the result returned in the `null` cases. Users without restrictions will be still passed through to the underlying `LeafReader`.

See also #5504 

* Category: Bug fix
* Why these changes are required? Diverse NPE issues

### Issues Resolved

Fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9873

### Testing

- Unit test
- Manual test

### Check List


cwperks:
Thank you for the quick fix @nibix !

cwperks (approved):

RyanL1997 (dismissed):

cwperks (dismissed):

cwperks (approved):

DarshitChanpura (approved):

PR #5541, opened 2025-08-04
[Resource Sharing] Restores client accessor pattern to fix compilation issues when security plugin is not installed

DarshitChanpura:
### Description
#5408 removed client accessor pattern to allow fetch client by Dependency Injection of the ResourceSharingExtension instance. However, that required that SPI be present on the classpath. That might not be the case in plugins which are not dependent on security.
See this example run from AD plugin PR that fails with ClassNotFoundException:
https://github.com/opensearch-project/anomaly-detection/actions/runs/16732487298/job/47363684918?pr=1533#step:6:635

This was not caught in our e2e test framework because security plugin is always present in the test cluster even if it is disabled.

* Category: Bug fix
* To unblock compilation of plugins

* Without this change, plugins will not compile when security is not present.

### Testing
automated testing

### Check List

### DIFF
--- a/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/SecurityDisabledTests.java
+++ b/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/SecurityDisabledTests.java
@@ -51,7 +51,7 @@ import static org.opensearch.test.framework.TestSecurityConfig.User.USER_ADMIN;
 public class SecurityDisabledTests {
 
     @ClassRule
-    public static LocalCluster cluster = new LocalCluster.Builder().clusterManager(ClusterManager.SINGLENODE)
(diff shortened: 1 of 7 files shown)


cwperks:
I'm getting convinced that OpenSearch has a bug in dependency injection. We know for a fact that it works with constructors, but no matter what I try I cannot get the field or method annotation to work.

This is the annotation: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/inject/Inject.java

It clearly shows what we'd need is supported, but I just cannot get it to work. Fields and methods are not being injected when the injector is created here: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/node/Node.java#L1680

That optional flag is exactly what's needed for the use case

derek-ho (approved):

PR #5543, opened 2025-08-05
[Backport 2.19] Optimizations and simplifications for WildcardMatcher

nibix:
### Description

Backport from https://github.com/opensearch-project/security/pull/5470

While researching https://github.com/opensearch-project/security/issues/5464 we found that the string matching algorithm used by WildcardMatcher does significantly more comparisons than necessary for simple cases.

This PR does not change the matching algorithm, but it introduces special cases for the most common patterns: "prefix*" and "*contains*". In a simple micro benchmark, we found that these optimizations can reduce the runtime required for matching large amounts of strings by 30% (see comments for micro benchmark code).

Additionally, this PR removes unused methods from WildcardMatcher and improves test coverage.

* Category: Enhancement
* Why these changes are required? For clusters with many indices and roles, runtime of wildcard matching gets very significant.

### Testing

- unit tests
- int tests

### Check List


src/main/java/org/opensearch/security/rest/SecurityWhoAmIAction.java
```diff
@@ -111,7 +111,7 @@ public void accept(RestChannel channel) throws Exception {
 
                         final String dn = sslInfo.getPrincipal();
                         final boolean isAdmin = adminDns.isAdminDN(dn);
-                        final boolean isNodeCertificateRequest = dn != null && WildcardMatcher.from(nodesDn, true).matchAny(dn);
+                        final boolean isNodeCertificateRequest = dn != null && WildcardMatcher.from(nodesDn).ignoreCase().matchAny(dn);
```
devardee, line 114: @nibix we don't have to call `.ignoreCase()` here right ?

nibix: At this place, it is possibly not so important, as this code serves no security critical function. The security critical checks are performed in DefaultInterClusterRequestEvaluator where they are done with `ignoreCase()` as well. This would indicate that in the action it should be done in the same fashion.

src/main/java/org/opensearch/security/securityconf/EvaluatedDlsFlsConfig.java
```diff
@@ -119,7 +119,7 @@ private Map<String, Set<String>> filter(Map<String, Set<String>> map, Set<String
         HashMap<String, Set<String>> result = new HashMap<>(map.size());
 
         for (Map.Entry<String, Set<String>> entry : map.entrySet()) {
-            if (WildcardMatcher.from(entry.getKey(), false).matchAny(allIndices)) {
+            if (WildcardMatcher.from(entry.getKey()).matchAny(allIndices)) {
```
devardee, line 122: Thanks @nibix , can you confirm here as well

nibix: In OpenSearch, index names are always lower case:

https://docs.opensearch.org/latest/api-reference/index-apis/create-index/

devardee: makes sense, thanks

PR #5557, opened 2025-08-11
[Resource Sharing] Adds `/search` API to sample plugins

DarshitChanpura:
### Description
Adds `/search` API to sample plugin and tests its behavior with resource sharing feature.

### Testing
- automated testsing

### Check List

### DIFF
--- a/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/TestUtils.java
+++ b/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/TestUtils.java
@@ -35,7 +35,9 @@ import org.opensearch.test.framework.cluster.TestRestClient;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.greaterThanOrEqualTo;
+import static org.hamcrest.Matchers.is;
 import static org.opensearch.sample.utils.Constants.RESOURCE_INDEX_NAME;
 import static org.opensearch.sample.utils.Constants.SAMPLE_RESOURCE_PLUGIN_PREFIX;
 import static org.opensearch.security.resources.ResourceSharingIndexHandler.getSharingIndex;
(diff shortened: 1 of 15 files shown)


cwperks (commented):
Thank you for this PR @DarshitChanpura ! I also wanted to capture the conversation we had about a proposal to make the search use-case more like DLS in the long-term to automatically filter results behind the scenes for resource plugins to work seamlessly so the plugin doesn't need to insert the filter clause themselves and that work is completely offloaded to security.

Currently, we have all sharing information stored in a separate sharing index. i.e. `.plugins-ml-model-group-sharing` where we have records like this:

```
"model-group-123": {
   "created_by": {
      "user": "darshit"
   },
   "share_with": {
      "ml_model_group_read_only": {
         "users": ["user1", "user2"],
         "roles": ["viewer_role"],
      },
      "ml_model_group_full_access": {
         "users": ["admin_user"],
         "roles": ["editor_role"],
      }
   }
}
```

in order to make the search use-case seamless for resource plugins we would also need to store a new field on the resource document itself to list the principals for which the resource is visible. The sharing index will be the source of truth for access levels, but this new field will be used to implement the filtering so that users see only what they should see and nothing more.

For example the field could be something like `"principals": ["user:craig", "role:role1", "backend_role:br1"]`

Behind the scenes security would populate this field on resource creation or a share/revoke operation. Security would then also use this field to inject a DLS-like clause behind the scenes when plugins search on the resource index so that the results return to the plugin will be limited to the resources that are visible to the user.

As discussed, we have problem similar to the one described [here](https://github.com/opensearch-project/index-management/issues/1471) where we would need a way to ensure that existing system indices mappings can be added to. Once a solution to that problem is solved, then there's a clear path towards a full solution that is seamless to plugins...then we can think deeper into the hierarchy and keeping track of resource parent ids.

sample-resource-plugin/src/main/java/org/opensearch/sample/resource/actions/transport/ShareResourceTransportAction.java
```diff
@@ -33,12 +33,13 @@
  */
 public class ShareResourceTransportAction extends HandledTransportAction<ShareResourceRequest, ShareResourceResponse> {
     private static final Logger log = LogManager.getLogger(ShareResourceTransportAction.class);
-    private final ResourceSharingClient resourceSharingClient;
+
+    @Inject(optional = true)
+    public SampleResourceExtension sampleResourceExtension;
```
cwperks, line 38: Nice to see that field injection is working! Its interesting how in OpenSearch' fork of Guice from 10 years ago that fields must be public. In newer versions of Guice, that's not the case but until OpenSearch removes the fork of guice and adds a dependency on Guice instead, we are stuck with our fork.

DarshitChanpura: Yes, if I have some time on my hands I'll look into raising a PR to remove Guice implementation and declare dependency on the package..

RyanL1997 (approved):

PR #5558, opened 2025-08-12
Remove `commons-io` and `commons-lang3` maven metadata from being shaded in opensaml jar

DarshitChanpura:
Unblocks 3.0 build by addressing CVE-2024-47554

https://github.com/opensearch-project/opensearch-build/issues/5595#issuecomment-3177822911

### Check List

### DIFF
--- a/libs/opensaml/build.gradle
+++ b/libs/opensaml/build.gradle
@@ -70,6 +70,7 @@ tasks.shadowJar {
     exclude 'org/slf4j/**'
     exclude 'javax/**'
     exclude 'META-INF/maven/commons-io/commons-io/**'
+    exclude 'META-INF/maven/org.apache.commons/commons-lang3/**'
     exclude 'META-INF/versions/**/org/bouncycastle/**'
     exclude 'META-INF/services/org.opensaml.security.crypto.ec.NamedCurve'
 }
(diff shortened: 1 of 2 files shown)


libs/opensaml/build.gradle
```diff
@@ -69,6 +69,7 @@ tasks.shadowJar {
     exclude 'org/publicsuffix/**'
     exclude 'org/slf4j/**'
     exclude 'javax/**'
+    exclude 'META-INF/maven/commons-io/commons-io/**'
```
cwperks, line 72: ref: https://github.com/opensearch-project/opensearch-build/issues/5595#issuecomment-3177822911

`org.apache.commons:commons-lang3 (opensaml-3.2.0.0-all.jar)`

Should we also exclude these files?

```
META-INF/maven/org.apache.commons/
META-INF/maven/org.apache.commons/commons-lang3/
META-INF/maven/org.apache.commons/commons-lang3/pom.xml
META-INF/maven/org.apache.commons/commons-lang3/pom.properties
```

gaiksaya: Yeah looks like that is being reported as well. Thanks for catching this @cwperks

cwperks: @gaiksaya FYI the class files are not being included in this shadow jar. I think excluded the META-INF files will make it fall off the report, but any code from affected jars would not be executable since the class files are not present.

DarshitChanpura: not sure if its related to the CVE but I've removed it.

PR #5560, opened 2025-08-13
Inject user custom attributes

markdboyd:
### Description

These changes are a follow-up to #5491. They are part of ongoing work to address #5491. These changes specifically inject user custom attributes to the security plugin context so that DLS/FLS replacement on queries will work properly.

These changes work in concert with these changes in other plugins:

https://github.com/opensearch-project/common-utils/pull/827
https://github.com/opensearch-project/alerting/pull/1917

  * Enhancement/bug fix
  * To ensure that user custom attributes are available in the security plugin so that DLS/FLS queries will work properly. See #5491 
  * User custom attributes are now available on the user object in the security plugin

### Issues Resolved

Addresses #5491

### Testing

I have been doing manual testing in Docker by compiling the jar files and mounting the jar files into the containers.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
@@ -307,7 +307,7 @@ public class PrivilegesEvaluator {
             log.debug(joiner);
 
             if (this.isUserAttributeSerializationEnabled()) {
-                joiner.add(Base64Helper.serializeObject((Serializable) context.getUser().getCustomAttributesMap()));
+                joiner.add(Base64Helper.serializeObject(new HashMap<>(context.getUser().getCustomAttributesMap())));
             }
 
             threadContext.putTransient(OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT, joiner.toString());
--- a/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
@@ -26,7 +26,6 @@
 
 package org.opensearch.security.privileges;
 
-import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
--- a/src/main/java/org/opensearch/security/auth/RolesInjector.java
+++ b/src/main/java/org/opensearch/security/auth/RolesInjector.java
@@ -72,8 +72,8 @@ final public class RolesInjector {
         }
         Set<String> roles = ImmutableSet.copyOf(strs[1].split(","));
 
-
-        Map<String, String> customAttributes = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER_CUSTOM_ATTRIBUTES);
+        Map<String, String> customAttributes = threadPool.getThreadContext()
+            .getTransient(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER_CUSTOM_ATTRIBUTES);
 
         User user = new User(strs[0]).withSecurityRoles(roles).withAttributes(customAttributes);
 
--- a/src/main/java/org/opensearch/security/auth/RolesInjector.java
+++ b/src/main/java/org/opensearch/security/auth/RolesInjector.java
@@ -15,6 +15,7 @@
 
 package org.opensearch.security.auth;
 
+import java.util.Map;
 import java.util.Set;
 
 import com.google.common.collect.ImmutableSet;
@@ -70,7 +71,11 @@ final public class RolesInjector {
             return null;
         }
         Set<String> roles = ImmutableSet.copyOf(strs[1].split(","));
-        User user = new User(strs[0]).withSecurityRoles(roles);
+
+
+        Map<String, String> customAttributes = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER_CUSTOM_ATTRIBUTES);
+
+        User user = new User(strs[0]).withSecurityRoles(roles).withAttributes(customAttributes);
 
         addUser(user, threadPool);
         return roles;
--- a/src/main/java/org/opensearch/security/support/ConfigConstants.java
+++ b/src/main/java/org/opensearch/security/support/ConfigConstants.java
@@ -131,6 +131,7 @@ public class ConfigConstants {
 
     public static final String OPENDISTRO_SECURITY_INJECTED_USER = "injected_user";
     public static final String OPENDISTRO_SECURITY_INJECTED_USER_HEADER = "injected_user_header";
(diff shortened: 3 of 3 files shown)


src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -307,7 +306,7 @@ private void setUserInfoInThreadContext(PrivilegesEvaluationContext context) {
             log.debug(joiner);
 
             if (this.isUserAttributeSerializationEnabled()) {
-                joiner.add(Base64Helper.serializeObject((Serializable) context.getUser().getCustomAttributesMap()));
+                joiner.add(Base64Helper.serializeObject(new HashMap<>(context.getUser().getCustomAttributesMap())));
```
cwperks, line 309: In the future we should consider using immutable collections directly from the JDK instead of using Guava to avoid having to copy this map on every call when user attribute serialization is enabled.

src/main/java/org/opensearch/security/privileges/dlsfls/DocumentPrivileges.java
```diff
@@ -174,6 +174,12 @@ static class Dynamic extends DlsQuery {
             @Override
             RenderedDlsQuery evaluate(PrivilegesEvaluationContext context) throws PrivilegesEvaluationException {
                 String effectiveQueryString = UserAttributes.replaceProperties(this.queryString, context);
+                if (effectiveQueryString.contains("${")) {
```
cwperks: This logic may be overly simple. If performance is a concern then I think this is a good check, but it may make sense to abstract this logic to a helper function so that we can write unit test cases for this. `public static boolean isAttributeSubstitutionIncomplete(String queryString)`

markdboyd: I added this method, but I actually refactored the code in several places where it was doing the same check to use the new method: 9f475ed033530b3e07c5a8c40b19df2a99d4949b. Let me know if that seems unnecessary or incorrrect.

I also named the method `needsAttributeSubstitution` because it seemed to fit better for the various places in the code where this method is now used, but happy to change the name at your discretion

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
markdboyd: OK, is there any action to take now?

cwperks: Not in this PR. btw Java has not made it easy to distinguish between a Mutable List and Immutable List when using JDK collections: https://chatgpt.com/share/68adc4df-d9e8-8008-b70e-72ca2545fa8c

src/integrationTest/java/org/opensearch/security/privileges/dlsfls/DocumentPrivilegesTest.java
```diff
@@ -377,10 +384,7 @@ public void queryPatternTemplate() throws Exception {
                 // It would be probably better if an error would be raised in that case.
```
cwperks: nit: looks like this comment may not be needed now

src/test/java/org/opensearch/security/privileges/UserAttributesUnitTest.java
```diff
@@ -0,0 +1,74 @@
...
+            null,
+            null,
+            null,
+            null,
+            null,
+            null,
+            null
+        );
+
+        String stringWithPlacholders = """
```
cwperks: nit: stringWithPlaceholders

src/test/java/org/opensearch/security/privileges/UserAttributesUnitTest.java
```diff
@@ -0,0 +1,74 @@
...
+            {
+                \"name\": \"${user.name}\",
+                \"name2\": \"${user_name}\",
+                \"bar\": \"${attr.proxy.attr1}\",
```
cwperks: Are all of these escapes actually necessary?

cwperks (approved):
Thank you @markdboyd! The changes LGTM. left a few minor comments, but nothing major.

src/main/java/org/opensearch/security/privileges/dlsfls/DocumentPrivileges.java
```diff
@@ -174,6 +174,12 @@ static class Dynamic extends DlsQuery {
             @Override
             RenderedDlsQuery evaluate(PrivilegesEvaluationContext context) throws PrivilegesEvaluationException {
                 String effectiveQueryString = UserAttributes.replaceProperties(this.queryString, context);
+                if (UserAttributes.needsAttributeSubstitution(effectiveQueryString)) {
```
DarshitChanpura, line 177: won't above method guarantee that query string is properly substituted? Is it expected to contain any other pattern than the ones mentioned in replaceProperties, replaceSecurityRoles and replaceRoles? or is it added to prevent this method from failing silently?

DarshitChanpura (approved):
thanks @markdboyd ! The PR looks good to me. left one clarifying question

src/main/java/org/opensearch/security/privileges/dlsfls/DocumentPrivileges.java
markdboyd: @DarshitChanpura `UserAttributes.replaceProperties` may not successfully replace all of the placeholders, especially in the case where the pattern string contains a property reference like `${attr.proxy.attr1}` but somehow that property is not available when the substitution occurs.

Granted, the whole point of this PR is to make sure that user attributes are injected at the right time so that property substitution can succeed, but it seems like there could be edge-case scenarios where substitution may not be fully successful.

cwperks: It may still fail to substitution all attributes.

Take this GPT created example:

```
// Original query string with user attribute placeholders
String queryString = """
{
  "term": {
    "owner": "${attr.owner}"
  },
  "range": {
    "created_at": {
      "gte": "${attr.start_date}",
      "lte": "${attr.end_date}"
    }
  },
  "term": {
    "region": "${attr.region}"
  }
}
""";

// Context simulates attributes available for the current user
PrivilegesEvaluationContext context = ...; 
// Suppose context.userAttributes() returns:
// {
//   "owner": "alice",
//   "region": "us-east-1"
// }
// Notice "start_date" and "end_date" are missing!
```

The result of `String effectiveQueryString = UserAttributes.replaceProperties(queryString, context);` would then be: 

```
{
  "term": {
    "owner": "alice"
  },
  "range": {
    "created_at": {
      "gte": "${attr.start_date}",
      "lte": "${attr.end_date}"
    }
  },
  "term": {
    "region": "us-east-1"
  }
}
```

DarshitChanpura: I see, thank you for the explanation.

PR #5574, opened 2025-08-18
[Resource Sharing] Fixes accessible resource ids search by marking created_by.user field as keyword search instead of text

DarshitChanpura:
### Description
Category: Bug Fix

While adding tests for AD plugin's search functionality to consume Resource sharing I discovered a bug in current getAccessibleResources() handler. At present, when searching for resource-ids that the owner it uses `created_by.user` as term query which behind the scenes uses standard text analyzer which defaults to lowercase exact term match.

This cause unexpected results when searching for forecasters in a test. This was not caught in original testing because all test user's had names in lower-case.

This PR changes the owner match clause to use `created_by.user.keyword` to allow exact case-sensitive match.

### Testing
- automated integ tests

### Check List
- [s] New functionality includes testing
- [s] Commits are signed per the DCO using --signoff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * Integration with stream transport ([#5530](https://github.com/opensearch-project/security/pull/5530))
 * [Resource Sharing] Adds `search` API to sample plugin and makes ResourceExtension an injectable pattern ([#5557](https://github.com/opensearch-project/security/pull/5557))
 * Update api permissions for query_insights_full_access ([#5554](https://github.com/opensearch-project/security/pull/5554))
+* [Resource Sharing] Restores client accessor pattern to fix compilation issues when security plugin is not installed ([#5541](https://github.com/opensearch-project/security/pull/5541))
 
 ### Bug Fixes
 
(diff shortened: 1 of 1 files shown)


sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/feature/enabled/ApiAccessTests.java
```diff
@@ -203,12 +203,12 @@ public void testApiAccess_allAccessUser() {
             api.assertApiRevoke(adminResId, FULL_ACCESS_USER, USER_ADMIN, sampleReadOnlyAG.name(), HttpStatus.SC_FORBIDDEN);
 
             // can share or revoke own resource
-            api.assertApiGet(userResId, USER_ADMIN, HttpStatus.SC_FORBIDDEN, "");
-            api.assertApiShare(userResId, FULL_ACCESS_USER, USER_ADMIN, sampleReadOnlyAG.name(), HttpStatus.SC_OK);
-            api.assertApiGet(userResId, USER_ADMIN, HttpStatus.SC_OK, "sampleUpdateUser");
-            api.assertApiPostSearch(searchByNamePayload("sampleUpdateUser"), USER_ADMIN, HttpStatus.SC_OK, 1, "sampleUpdateUser");
-            api.assertApiRevoke(userResId, FULL_ACCESS_USER, USER_ADMIN, sampleReadOnlyAG.name(), HttpStatus.SC_OK);
-            api.assertApiGet(userResId, USER_ADMIN, HttpStatus.SC_FORBIDDEN, "");
+            api.assertApiGet(userResId, LIMITED_ACCESS_USER, HttpStatus.SC_FORBIDDEN, "");
```
RyanL1997, line 206: just for my knowledge, why would like to transfer this into non-admin user?

DarshitChanpura: ah, so admin/non-admin doesn't matter here since access is evaluated on whether either of the user have access to FULL_USER's resource.  

The change is because USER_ADMIN user is named as "admin" which is all lower-case whereas LIMITED_ACCESS_USER is named "resource_sharing_user_Limited_Perms" which has mixed-case which is what exactly this PR is resolving.

So this change guarantees that my PR change in ResourceAccessHandler works as expected.

PR #5576, opened 2025-08-19
[Resource Sharing] Reverts @Inject pattern usage for ResourceSharingExtension to client accessor pattern.

DarshitChanpura:
### Description
After trying to implement `@Inject ` pattern to inject the extension and use the Guice injection to retrieve client from extension in plugin's transport actions, we discovered that cluster bootstrap would always fails with `NoClasDefFoundError` whenever SPI is not present in the classpath (i.e. security is not present).

We experimented with this annotation as discussed in comments of https://github.com/opensearch-project/security/pull/5541, but were not successful.
Hence, we are reverting to client accessor singleton pattern.

Category: Bug fix

### Issues Resolved
Unblocks plugin CIs and allows clusters without security plugin to bootstrap.

### Testing
Automated tests

### Check List


DarshitChanpura:
This changes reverts back to the changes made here: https://github.com/opensearch-project/security/pull/5541

The only other alternative is to have plugins implement their own version of:
<details>
<summary>ResourceExtensionWrapper.java</summary>

```java
/*
 * SPDX-License-Identifier: Apache-2.0
 *
 * The OpenSearch Contributors require contributions made to
 * this file be licensed under the Apache-2.0 license or a
 * compatible open source license.
 *
 * Modifications Copyright OpenSearch Contributors. See
 * GitHub history for details.
 */

package org.opensearch.sample;

import org.apache.logging.log4j.LogManager;
import org.apache.logging.log4j.Logger;
import org.opensearch.common.inject.Inject;
import org.opensearch.common.inject.Injector;

import java.lang.reflect.Method;

public final class ResourceExtensionWrapper {
    private static final Logger log = LogManager.getLogger(ResourceExtensionWrapper.class);

    // Class names kept as strings to avoid compile-time coupling
    private static final String SPI_IFACE =
            "org.opensearch.security.spi.resources.ResourceSharingExtension";
    private static final String IMPL_CLASS =
            "org.opensearch.sample.SampleResourceExtension";

    @Inject
    public Injector injector;

    // Init state
    private volatile boolean initialized;          // set to true when we have a definitive state
    private volatile boolean spiAbsentForever;     // true iff SPI not on classpath
    private volatile Object extension;             // the extension instance or null
    private volatile Method getClientMethod;       // cached "getResourceSharingClient" Method

    /** Visible for tests: allow explicit instance injection. */
    public void setExtension(Object extension) {
        this.extension = extension;
        // cache method if possible
        if (extension != null) {
            try {
                this.getClientMethod = extension.getClass().getMethod("getResourceSharingClient");
            } catch (ReflectiveOperationException e) {
                log.debug("Timeseries RS extension present but no getResourceSharingClient()", e);
                this.getClientMethod = null;
            }
        } else {
            this.getClientMethod = null;
        }
        this.initialized = true;
    }

    /** Returns true if SPI is on the classpath AND an extension instance is available. */
    public boolean isPresent() {
        ensureInitialized();
        return extension != null && getClientMethod != null;
    }

    /**
     * Returns the ResourceSharingClient (as Object) or null if SPI/extension is not available.
     * Intentionally returns Object to avoid compile-time coupling to SPI types.
     */
    public Object getResourceSharingClient() {
        ensureInitialized();
        Object ext = extension;
        Method m = getClientMethod;
        if (ext == null || m == null) return null;
        try {
            return m.invoke(ext);
        } catch (ReflectiveOperationException e) {
            // treat as absent at runtime; don't flip initialized so future retries could still happen if state changes
            log.debug("Invoking getResourceSharingClient() failed; treating as absent.", e);
            return null;
        }
    }

    private void ensureInitialized() {
        if (initialized) return;

        synchronized (this) {
            if (initialized) return;

            // 0) If we previously discovered SPI is missing, we can finalize immediately.
            if (spiAbsentForever) {
                initialized = true;
                return;
            }

            // 1) Check SPI presence using TCCL first, then our loader.
            ClassLoader tccl = Thread.currentThread().getContextClassLoader();
            ClassLoader self = getClass().getClassLoader();
            if (classAbsent(tccl) && classAbsent(self)) {
                // SPI truly not installed — no need to ever retry
                spiAbsentForever = true;
                extension = null;
                getClientMethod = null;
                initialized = true;
                log.debug("Security SPI not present on classpath; RS extension unavailable.");
                return;
            }

            // 2) SPI present; try to obtain our implementation instance from Guice.
            //    If this fails due to binding timing, leave 'initialized'=false so we retry on next call.
            if (injector == null) {
                // This shouldn't happen with ctor injection, but be defensive if refactored.
                log.debug("Injector not available yet; deferring RS extension initialization.");
                return; // DO NOT set initialized — we want to retry later
            }

            try {
                Class<?> implClazz = Class.forName(IMPL_CLASS, /* initialize */ false, self);
                Object instance = injector.getInstance((Class<?>) implClazz);
                Method m = implClazz.getMethod("getResourceSharingClient");

                this.extension = instance;
                this.getClientMethod = m;
                this.initialized = true; // success — we can finalize
                log.debug("Initialized Timeseries RS extension via Guice: {}", IMPL_CLASS);
            } catch (ClassNotFoundException e) {
                // Our impl isn't on the classpath (shouldn't happen since it's our own class) — finalize as absent
                log.debug("RS extension impl class not found; treating as unavailable.", e);
                this.extension = null;
                this.getClientMethod = null;
                this.initialized = true;
            } catch (Throwable t) {
                // Provisioning/binding not ready yet or other runtime issue — DO NOT finalize; allow retrial.
                log.debug("RS extension not ready yet (will retry on next call).", t);
                // leave initialized=false to allow a later retry
            }
        }
    }

    private static boolean classAbsent(ClassLoader cl) {
        try {
            if (cl == null) return true;
            Class.forName(ResourceExtensionWrapper.SPI_IFACE, /* initialize */ false, cl);
            return false;
        } catch (ClassNotFoundException e) {
            return true;
        } catch (LinkageError e) {
            // present but broken — treat as absent to be safe
            return true;
        }
    }
}

```
</details>
and then inject it in their createComponents like:
```java
 try {
            // This loader will throw class not found exception if SPI doesn't exist (feature disabled or security not installed)
            Class.forName("org.opensearch.security.spi.resources.ResourceSharingExtension");
            // if no exception was thrown we make the wrapper an injectable component
            components.add(new ResourceExtensionWrapper());
        } catch (ClassNotFoundException e) {
            // spi is not loaded into classpath, i.e. the feature is disabled or security plugin is absent, do nothing
        }
```

I'm not in favor of this approach.

cwperks:
Thank you for trying @DarshitChanpura . Unfortunately, it seems like our fork of Guice from 10 years ago doesn't have the optional bindings necessary for such a use-case.

RyanL1997 (approved):

PR #5597, opened 2025-08-28
[Resource Sharing] Adds API to provide dashboards support for resource access management 

### LINKED ISSUE #5607: [FEATURE] Add APIs to support resource access management dashboard
This feature requests addition of resource-access-management APIs to be consumed by dashboards. Specifically, 
- `/types` API that lists available resource types
- `/list?resource_type=<type>` API that lists available resource for the selected type.

### LINKED ISSUE #5608: [FEATURE] Consume action-groups registered by resource plugins
At present, resource access evaluator relies on action-groups declared across security plugin which is too broad. Sharing with a "read" action group may result in evaluation of all actions associated with that action-group. In addition, it becomes difficult to scope down the action-groups pertinent to the resource on the front-end.

Enable plugins to register to supply action-groups pertinent to the resource via `resource-action-groups.yml` file. 
Format:
```yml
resource_types:
   <fully-qualified-resource-class-name-1>:
      <action-group-1>:
          - <action1>
          - <action2>
      <action-group-2>:
          - <action1>
          - <action2>
       
   <fully-qualified-resource-class-name-2>:
      <action-group-1>:
          - <action1>
          - <action2>
      <action-group-2>:
          - <action1>
          - <action2>
    
   # ...
```
Here is an example yml for anomaly-detecion plugin that provides two resources:
```yml
resource_types:
  org.opensearch.ad.model.AnomalyDetector:
    ad_read_only:
      - "cluster:admin/opendistro/ad/detectors/get"
      - "cluster:admin/opendistro/ad/detectors/search"
      - "indices:data/read*"

    ad_read_write:
      - "cluster:admin/opendistro/ad/*"
      - "indices:*"

    ad_full_access:
      - "cluster:admin/opendistro/ad/*"
      - "indices:*"
      - "cluster:admin/security/resource/share"

  org.opensearch.forecast.model.Forecaster:
    forecast_read_only:
      - "cluster:admin/plugin/forecast/forecasters/get"
      - "cluster:admin/plugin/forecast/forecasters/search"
      - "indices:data/read*"

    forecast_read_write:
      - "cluster:admin/plugin/forecast/*"
      - "indices:*"

    forecast_full_access:
      - "cluster:admin/plugin/forecast/*"
      - "indices:*"
      - "cluster:admin/security/resource/share"
```

These static action-groups will enable a narrowed down list of action-groups to be displayed on front-end and permission will be specific to the resource-type.

DarshitChanpura:
### Description
Adds support for dashboard page to manage resource access. Enables plugins to supply static per-resource action-groups through yml config.

* Category : Enhancement

### Issues Resolved
- resolves #5607
- resolves #5608 

### Testing
Automated + manual

- Apis added:

### 1. `GET /_plugins/_security/api/resource/types`

**Description:**
Retrieves the current sharing configuration for a given resource.

**Example Request:**

```
GET /_plugins/_security/api/resource/types
```

**Response:**

```json
{
  "types": [
    {
      "type": "org.opensearch.sample.SampleResource",
      "index": ".sample_resource",
      "action_groups": ["sample_read_only", "sample_read_write", "sample_full_access"]
    }
  ]
}
```
NOTE: `action_groups` are fetched from `resource-action-groups.yml` supplied by resource plugin.

### 2. `GET /_plugins/_security/api/resource/list?resource_type=<resource-index-name>`

**Description:**
Retrieves sharing information for all records accessible to requesting user for the given resource_index.

**Example Request:**
as user `darshit`
```
GET /_plugins/_security/api/resource/list?resource_type=.sample_resource
```

**Response:**

```json
{
  "resources": [
    {
      "resource_id": "1",
      "created_by":  {
        "user": "darshit",
        "tenant": "some-tenant"
      },
      "share_with": {
        "sample_read_only": {
          "users": ["craig"]
        }
      },
      "can_share": true
    }
  ]
}
```

NOTE:
- `share_with` may not be present if resource has not been shared yet
- if same request is made as user `craig`, `can_share` value for resource_id `1` will be `false` since `craig` does not have share permission.

### Check List

### DIFF
--- a/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
+++ b/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
@@ -108,6 +108,7 @@ opensearchplugin {
 
       # ...
     ```
+
 #### **Example resource-action-groups yml**
 ```yml
 resource_types:
--- a/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
+++ b/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
@@ -93,19 +93,23 @@ opensearchplugin {
     resource_types:
       <resource-type-1>:
           <action-group-1>:
-          - <action1>
-          - <action2>
+              allowed_actions:
+                  - <action1>
+                  - <action2>
           <action-group-2>:
-          - <action1>
-          - <action2>
+              allowed_actions:
+                  - <action1>
+                  - <action2>
 
       <resource-type-2>:
           <action-group-1>:
-          - <action1>
-          - <action2>
+              allowed_actions:
+                  - <action1>
+                  - <action2>
           <action-group-2>:
-          - <action1>
-          - <action2>
+              allowed_actions:
+                  - <action1>
+                  - <action2>
 
       # ...
     ```
--- a/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
+++ b/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
@@ -113,7 +113,7 @@ opensearchplugin {
 
       # ...
     ```
-NOTE: The resource-type supplied here must match the ones supplied in each of the resource-providers declared in the ResourceSharingExtension implementation.
+**NOTE**: The resource-type supplied here must match the ones supplied in each of the resource-providers declared in the ResourceSharingExtension implementation.
 
 #### **Example resource-action-groups yml**
 ```yml
@@ -568,7 +568,7 @@ Creates or replaces sharing settings for a resource.
 ```json
 {
   "resource_id": "resource-123",
-  "resource_type": "my-resource-index",
+  "resource_type": "my-type",
   "share_with": {
     "read_only": {
       "users": ["alice"],
@@ -613,7 +613,7 @@ Updates sharing settings by **adding** or **removing** recipients at any access 
 ```json
 {
   "resource_id": "resource-123",
-  "resource_type": "my-resource-index",
+  "resource_type": "my-type",
   "add": {
     "read_only": {
       "users": ["charlie"]
@@ -658,7 +658,7 @@ Updates sharing settings by **adding** or **removing** recipients at any access 
 - `"revoke"` – Removes recipients
 
 
-### 3. `GET /_plugins/_security/api/resource/share?resource_id=<id>&resource_type=<index>`
+### 3. `GET /_plugins/_security/api/resource/share?resource_id=<id>&resource_type=<type>`
 
 **Description:**
 Retrieves the current sharing configuration for a given resource.
@@ -666,7 +666,7 @@ Retrieves the current sharing configuration for a given resource.
 **Example Request:**
 
 ```
-GET /_plugins/_security/api/resource/share?resource_id=resource-123&resource_type=my-resource-index
+GET /_plugins/_security/api/resource/share?resource_id=resource-123&resource_type=my-type
 ```
 
 **Response:**
@@ -705,8 +705,7 @@ GET /_plugins/_security/api/resource/types
 {
   "types": [
     {
-      "type": "org.opensearch.sample.SampleResource",
-      "index": ".sample_resource",
+      "type": "sample-resource",
       "action_groups": ["sample_read_only", "sample_read_write", "sample_full_access"]
     }
   ]
@@ -714,7 +713,7 @@ GET /_plugins/_security/api/resource/types
 ```
 NOTE: `action_groups` are fetched from `resource-action-groups.yml` supplied by resource plugin.
 
-### 5. `GET /_plugins/_security/api/resource/list?resource_type=<resource-index-name>`
+### 5. `GET /_plugins/_security/api/resource/list?resource_type=<type>`
 
 **Description:**
 Retrieves sharing information for all records accessible to requesting user for the given resource_index.
@@ -722,7 +721,7 @@ Retrieves sharing information for all records accessible to requesting user for 
 **Example Request:**
 as user `darshit`
 ```
-GET /_plugins/_security/api/resource/list?resource_type=.sample_resource
+GET /_plugins/_security/api/resource/list?resource_type=sample-resource
 ```
 
 **Response:**
--- a/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
+++ b/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
@@ -373,14 +373,14 @@ By default, all `shareableResources` are private — visible only to their **own
 
 ### **Example: Publicly Shared Resource**
 
-To make a resource accessible to everyone, share it with all entities using the wildcard `*`:
+To make a resource accessible to everyone, share it with `users` entity using the wildcard `*`:
 
 ```json
 {
   "share_with": {
     "default": {
-      "backend_roles": ["*"],
-      "roles": ["*"],
+      "backend_roles": ["some_backend_role"],
+      "roles": ["some_role"],
       "users": ["*"]
     }
   }
--- a/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
+++ b/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
@@ -83,7 +83,48 @@ opensearchplugin {
     ```
     org.opensearch.sample.SampleResourceSharingExtension
     ```
+- **Declare** action-groups **per resource** in `resource-action-groups.yml` file:
+  ```
+  src/main/resources/resource-action-groups.yml
+  ```
+  - This file must be structured in a following way:
+    ```yml
+    resource_types:
+      <fully-qualified-resource-class-name-1>:
+          <action-group-1>:
+          - <action1>
+          - <action2>
+          <action-group-2>:
+          - <action1>
+          - <action2>
 
+      <fully-qualified-resource-class-name-2>:
+          <action-group-1>:
+          - <action1>
+          - <action2>
+          <action-group-2>:
+          - <action1>
+          - <action2>
+
+      # ...
+    ```
+#### **Example resource-action-groups yml**
+```yml
+resource_types:
+  org.opensearch.sample.SampleResource:
+      sample_read_only:
+        - "cluster:admin/sample-resource-plugin/get"
+        - "indices:data/read*"
+
+      sample_read_write:
+        - "cluster:admin/sample-resource-plugin/*"
+        - "indices:*"
+
+      sample_full_access:
+        - "cluster:admin/sample-resource-plugin/*"
+        - "indices:*"
+        - "cluster:admin/security/resource/share"
+```
 
 ## **3. Resource Sharing API Design**
 
@@ -104,15 +145,17 @@ NOTE: **action-groups** and **access-levels** are used inter-changeably througho
 
 ### This object contains details about the **creator** of the resource.
 
-|**Field**  |**Type** |Description  |
-|---  |---  |---  |
-| user|String |The username of the creator. |
+| **Field** |**Type** | Description                                                  |
+|-----------|---  |--------------------------------------------------------------|
+| user      |String | The username of the creator.                                 |
+| tenant    |String | The tenant where resource sits. If multi-tenancy is enabled. |
 
 **Example:**
 
 ```
 "created_by": {
-   "user": "darshit"
+   "user": "darshit",
+   "tenant": "some_tenant"
 }
 ```
 
@@ -163,14 +206,14 @@ Each **action-group** entry contains the following access definitions:
 }
 ```
 
-
 ### **Example Resource-Sharing Document**
 
 ```
 {
    "resource_id": "model-group-123",
    "created_by": {
-      "user": "darshit"
+      "user": "darshit",
+      "tenant": "some-tenant"
    },
    "share_with": {
       "action-group1": {
@@ -641,25 +684,93 @@ GET /_plugins/_security/api/resource/share?resource_id=resource-123&resource_typ
 }
 ```
 
+### 4. `GET /_plugins/_security/api/resource/types`
+
+**Description:**
+Retrieves the current sharing configuration for a given resource.
+
+**Example Request:**
+
+```
+GET /_plugins/_security/api/resource/types
+```
+
+**Response:**
+
+```json
+{
+  "types": [
+    {
+      "type": "org.opensearch.sample.SampleResource",
+      "index": ".sample_resource",
+      "action_groups": ["sample_read_only", "sample_read_write", "sample_full_access"]
+    }
+  ]
+}
+```
+NOTE: `action_groups` are fetched from `resource-action-groups.yml` supplied by resource plugin.
+
+### 5. `GET /_plugins/_security/api/resource/list?resource_type=<resource-index-name>`
+
+**Description:**
+Retrieves sharing information for all records accessible to requesting user for the given resource_index.
+
+**Example Request:**
+as user `darshit`
+```
+GET /_plugins/_security/api/resource/list?resource_type=.sample_resource
+```
+
+**Response:**
+
+```json
+{
+  "resources": [
+    {
+      "resource_id": "1",
+      "created_by":  {
+        "user": "darshit",
+        "tenant": "some-tenant"
+      },
+      "share_with": {
+        "sample_read_only": {
+          "users": ["craig"]
+        }
+      },
+      "can_share": true
+    }
+  ]
+}
+```
+
+NOTE:
+- `share_with` may not be present if resource has not been shared yet
+- if same request is made as user `craig`, `can_share` value for resource_id `1` will be `false` since `craig` does not have share permission.
+
+
 
 ## Who Can Use This?
 
-| API                                            | Permission Required               | Intended User     |
-|------------------------------------------------|-----------------------------------|-------------------|
+| API                                             | Permission Required               | Intended User     |
+|-------------------------------------------------|-----------------------------------|-------------------|
 | `POST /_plugins/_security/api/resources/migrate` | REST admin or Super admin         | Cluster admin     |
-| `PUT /_plugins/_security/api/resource/share`     | Resource Owner                    | Dashboards / REST |
-| `PATCH /_plugins/_security/api/resource/share`    | Resource Owner / share permission | Dashboards / REST |
-| `GET /_plugins/_security/api/resource/share`      | Resource Owner / read permission  | Dashboards / REST |
+| `PUT /_plugins/_security/api/resource/share`    | Resource Owner                    | Dashboards / REST |
+| `PATCH /_plugins/_security/api/resource/share`  | Resource Owner / share permission | Dashboards / REST |
+| `GET /_plugins/_security/api/resource/share`    | Resource Owner / read permission  | Dashboards / REST |
+| `GET /_plugins/_security/api/resource/types`    | Dashboard access                  | Dashboards |
+| `GET /_plugins/_security/api/resource/list`     | Dashboard access                  | Dashboards |
 
 
 ## When to Use
 
-| Use Case                                            | API                        |
-|-----------------------------------------------------|----------------------------|
-| Migrating existing plugin-specific sharing configs  | `POST /_plugins/_security/api/resources/migrate` |
-| Sharing a document with another user or role        | `PUT /_plugins/_security/api/resource/share`   |
-| Granting/revoking access without affecting others   | `PATCH /_plugins/_security/api/resource/share` |
-| Fetching the current sharing status of a resource   | `GET /_plugins/_security/api/resource/share`   |
+| Use Case                                                    | API                                              |
+|-------------------------------------------------------------|--------------------------------------------------|
+| Migrating existing plugin-specific sharing configs          | `POST /_plugins/_security/api/resources/migrate` |
+| Sharing a document with another user or role                | `PUT /_plugins/_security/api/resource/share`     |
+| Granting/revoking access without affecting others           | `PATCH /_plugins/_security/api/resource/share`   |
+| Fetching the current sharing status of a resource           | `GET /_plugins/_security/api/resource/share`     |
+| Listing resource type. Encouraged only for dashboard access | `GET /_plugins/_security/api/resource/types`     |
+| Listing accessible resources in given resource index.       | `GET /_plugins/_security/api/resource/list`      |
 
 
 ## **5. Best Practices For Users & Admins**
--- a/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
+++ b/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
@@ -44,6 +44,7 @@ This feature introduces **one primary component** for plugin developers:
 ### **Plugin Implementation Requirements:**
 
 - **Resource indices must be declared as system indices** to prevent unauthorized direct access.
+  NOTE: If system-index protection is disabled, requests will be evaluated as normal index requests.
 - **Declare a `compileOnly` dependency** on `opensearch-security-spi` package and opensearch-security plugin zip:
 ```build.gradle
 configurations {
@@ -90,7 +91,7 @@ opensearchplugin {
   - This file must be structured in a following way:
     ```yml
     resource_types:
-      <fully-qualified-resource-class-name-1>:
+      <resource-type-1>:
           <action-group-1>:
           - <action1>
           - <action2>
@@ -98,7 +99,7 @@ opensearchplugin {
           - <action1>
           - <action2>
 
-      <fully-qualified-resource-class-name-2>:
+      <resource-type-2>:
           <action-group-1>:
           - <action1>
           - <action2>
@@ -108,22 +109,20 @@ opensearchplugin {
 
       # ...
     ```
+NOTE: The resource-type supplied here must match the ones supplied in each of the resource-providers declared in the ResourceSharingExtension implementation.
 
 #### **Example resource-action-groups yml**
 ```yml
 resource_types:
-  org.opensearch.sample.SampleResource:
+  sample-resource:
       sample_read_only:
         - "cluster:admin/sample-resource-plugin/get"
-        - "indices:data/read*"
 
       sample_read_write:
         - "cluster:admin/sample-resource-plugin/*"
-        - "indices:*"
 
       sample_full_access:
         - "cluster:admin/sample-resource-plugin/*"
-        - "indices:*"
         - "cluster:admin/security/resource/share"
 ```
 
--- a/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/SecurityDisabledTests.java
+++ b/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/SecurityDisabledTests.java
@@ -29,6 +29,7 @@ import org.opensearch.test.framework.matcher.RestMatchers;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.opensearch.sample.resource.TestUtils.SAMPLE_READ_ONLY_RESOURCE_AG;
 import static org.opensearch.sample.resource.TestUtils.SAMPLE_RESOURCE_CREATE_ENDPOINT;
 import static org.opensearch.sample.resource.TestUtils.SAMPLE_RESOURCE_DELETE_ENDPOINT;
 import static org.opensearch.sample.resource.TestUtils.SAMPLE_RESOURCE_GET_ENDPOINT;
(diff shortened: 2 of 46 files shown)


RESOURCE_SHARING_AND_ACCESS_CONTROL.md
```diff
@@ -83,7 +83,49 @@ opensearchplugin {
...
     org.opensearch.sample.SampleResourceSharingExtension
     ```
+- **Declare** action-groups **per resource** in `resource-action-groups.yml` file:
+  ```
+  src/main/resources/resource-action-groups.yml
+  ```
+  - This file must be structured in a following way:
+    ```yml
+    resource_types:
+      <fully-qualified-resource-class-name-1>:
```
cwperks: Why classname? Can this be the human-readable `resource-type` instead?

RESOURCE_SHARING_AND_ACCESS_CONTROL.md
```diff
@@ -83,7 +83,49 @@ opensearchplugin {
...
+      # ...
+    ```
 
+#### **Example resource-action-groups yml**
+```yml
+resource_types:
+  org.opensearch.sample.SampleResource:
+      sample_read_only:
+        - "cluster:admin/sample-resource-plugin/get"
+        - "indices:data/read*"
```
cwperks: Is this accurate? `indices:...` perms don't have any effect here.

This should support `cluster:` initially, but I'd ultimately like to see a new category of perms going forward for these to separate these further from `cluster:` and `indices:`

RESOURCE_SHARING_AND_ACCESS_CONTROL.md
```diff
@@ -104,15 +146,17 @@ NOTE: **action-groups** and **access-levels** are used inter-changeably througho
 
 ### This object contains details about the **creator** of the resource.
 
-|**Field**  |**Type** |Description  |
-|---  |---  |---  |
-| user|String |The username of the creator. |
+| **Field** |**Type** | Description                                                  |
+|-----------|---  |--------------------------------------------------------------|
+| user      |String | The username of the creator.                                 |
+| tenant    |String | The tenant where resource sits. If multi-tenancy is enabled. |
```
cwperks, line 155: Ty for adding this :).

RESOURCE_SHARING_AND_ACCESS_CONTROL.md
```diff
@@ -641,25 +685,93 @@ GET /_plugins/_security/api/resource/share?resource_id=resource-123&resource_typ
...
+      "type": "org.opensearch.sample.SampleResource",
+      "index": ".sample_resource",
+      "action_groups": ["sample_read_only", "sample_read_write", "sample_full_access"]
+    }
+  ]
+}
+```
+NOTE: `action_groups` are fetched from `resource-action-groups.yml` supplied by resource plugin.
+
+### 5. `GET /_plugins/_security/api/resource/list?resource_type=<resource-index-name>`
```
cwperks: I think we should have a human readable `resource-type` for each resource sharing extension.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/SecurityDisabledTests.java
```diff
@@ -37,7 +37,7 @@
 import static org.opensearch.sample.resource.TestUtils.SAMPLE_RESOURCE_SHARE_ENDPOINT;
 import static org.opensearch.sample.resource.TestUtils.SAMPLE_RESOURCE_UPDATE_ENDPOINT;
 import static org.opensearch.sample.resource.TestUtils.revokeAccessPayload;
-import static org.opensearch.sample.resource.TestUtils.sampleReadOnlyAG;
+import static org.opensearch.sample.resource.TestUtils.sampleReadOnlyResourceAG;
```
cwperks: Since this is a constant, should we take the opportunity to rename it using UPPER_SNAKE_CASE?

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/AccessibleResourcesApiTests.java
```diff
@@ -0,0 +1,190 @@
...
+            List<Object> resources = (List<Object>) response.bodyAsMap().get("resources");
+            assertThat(resources.size(), equalTo(1));
+            Map<String, Object> resource = (Map<String, Object>) resources.getFirst();
+            assertThat(resource.get("resource_id"), equalTo(adminResId));
+            assertThat(resource.get("created_by"), equalTo(Map.of("user", USER_ADMIN.getName())));
+            Map<String, Object> shareWith = Map.of(
+                sampleReadOnlyResourceAG,
+                Map.of(Recipient.USERS.getName(), List.of(user.getName())),
+                sampleFullAccessResourceAG,
+                Map.of(Recipient.USERS.getName(), List.of(user.getName()))
```
cwperks, line 150: We should update sharing so that the same principal cannot be in multiple access levels simultaneously.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/AccessibleResourcesApiTests.java
```diff
@@ -0,0 +1,190 @@
...
+            assertThat(resource.get("resource_id"), equalTo(adminResId));
+            assertThat(resource.get("created_by"), equalTo(Map.of("user", USER_ADMIN.getName())));
+            assertThat(
+                resource.get("share_with"),
+                equalTo(Map.of(sampleReadOnlyResourceAG, Map.of(Recipient.USERS.getName(), List.of(user.getName()))))
+            );
+            assertThat(resource.get("can_share"), equalTo(Boolean.FALSE));
+        }
+
+        // share at read-only, can_share should say false
```
cwperks: I think this comment is meant for the block above?

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/AccessibleResourcesApiTests.java
```diff
@@ -0,0 +1,190 @@
...
+import static org.opensearch.sample.utils.Constants.RESOURCE_INDEX_NAME;
+import static org.opensearch.test.framework.TestSecurityConfig.User.USER_ADMIN;
+
+/**
+ * This test file tests the list API that lists current user's accessible resources
+ * Sharing behaviour is tested in ShareApiTests, this class simply tests that the shared resources are visible after sharing through list api
+ */
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class AccessibleResourcesApiTests {
```
cwperks, line 55: I think we should expand on this suite by adding multiple resources with different creators. Either in this suite or a separate one.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/TypesApiTests.java
```diff
@@ -0,0 +1,65 @@
...
+import static org.opensearch.sample.resource.TestUtils.SECURITY_TYPES_ENDPOINT;
+import static org.opensearch.sample.resource.TestUtils.newCluster;
+import static org.opensearch.sample.utils.Constants.RESOURCE_INDEX_NAME;
+
+/**
+ * This test file tests the types API that lists available resource types
+ */
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class TypesApiTests {
```
cwperks: nit: I'm wondering whether we should call this `SharableResourceInfoTests` to be consistent with `DashboardsInfo` and `TenantInfo` apis.

spi/src/main/java/org/opensearch/security/spi/resources/sharing/ShareWith.java
```diff
@@ -169,4 +169,34 @@ public ShareWith revoke(ShareWith other) {
...
+        Map<Recipient, Set<String>> src = r.getRecipients();
+        if (src == null || src.isEmpty()) return null;
+
+        Map<Recipient, Set<String>> cleaned = new EnumMap<>(Recipient.class);
+        for (Map.Entry<Recipient, Set<String>> e : src.entrySet()) {
+            Set<String> vals = e.getValue();
+            if (vals == null) continue;
+            Set<String> filtered = vals.stream()
+                .filter(s -> s != null && !s.isBlank())
+                .collect(java.util.stream.Collectors.toCollection(java.util.LinkedHashSet::new));
```
cwperks: nit: use imports for these

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1217,6 +1231,8 @@ public Collection<Object> createComponents(
             // CS-ENFORCE-SINGLE
         }
 
+        resourceAccessEvaluator = new ResourceAccessEvaluator(resourcePluginInfo.getResourceIndices(), settings, resourceAccessHandler);
```
cwperks, line 1237: why this change?

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -2428,17 +2443,128 @@ private void tryAddSecurityProvider() {
...
-            resourcePluginInfo.setResourceSharingExtensions(resourceSharingExtensions);
+            return;
+        }
+
+        // discover & register extensions and their types
+        Set<ResourceSharingExtension> exts = new HashSet<>(loader.loadExtensions(ResourceSharingExtension.class));
+        resourcePluginInfo.setResourceSharingExtensions(exts);
+
+        for (var ext : exts) {
+            URL url = ext.getClass().getClassLoader().getResource("resource-action-groups.yml");
```
cwperks: Should we create a class like [ConfigHelper](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/ConfigHelper.java) for this?

RESOURCE_SHARING_AND_ACCESS_CONTROL.md
DarshitChanpura: At present for resource-type we promote usage of `ResourceClass.getClass().getCanonicalName()` to avoid collisions with another similar named resources.  These classes are displayed as human readable on front-end tho. 

We can implement an uniqueness enforcing logic in another PR, perhaps? 
Should we fail cluster bootstrap if there are two or more resources with same resource-type

RESOURCE_SHARING_AND_ACCESS_CONTROL.md
DarshitChanpura: it is for controlling direct index access requests when system index is disabled.

RESOURCE_SHARING_AND_ACCESS_CONTROL.md
DarshitChanpura: there is currently no way to enforce uniqueness. We can enforce uniqueness in a follow-up PR.

Also, this particular request is for dashboards only, we don't intend actual user to make this request. Dashboard code will handle this.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/AccessibleResourcesApiTests.java
DarshitChanpura: It is already tested in ShareApiTests. We only test whether resources are visible after sharing and can user share them further.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/AccessibleResourcesApiTests.java
DarshitChanpura: updated it.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/AccessibleResourcesApiTests.java
DarshitChanpura: this is a cleanup that we have to implement. However, from dashboard this should be easy on user's end as they will see who this is shared_with and can only share_with a user at single level.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/TypesApiTests.java
DarshitChanpura: done.

spi/src/main/java/org/opensearch/security/spi/resources/sharing/ShareWith.java
DarshitChanpura: done.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: done

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: originally added because I was thinking about storing action-groups info in the evaluator, but now kept it to make SecurityFilter definition more readable.

RESOURCE_SHARING_AND_ACCESS_CONTROL.md
cwperks: > Should we fail cluster bootstrap if there are two or more resources with same resource-type

yes, let's enforce uniqueness. 

I'd rather enforce uniqueness constraint for `resource-type` then using fully-qualified class name. Unless there's a mechanism in place, these static files would break with refactor. I suppose same could be said about `resource-type` as well, but then again we could also enforce that for every `resource-type` there needs to be at least one access level which ought to be enforced.

RESOURCE_SHARING_AND_ACCESS_CONTROL.md
cwperks: Get Document requests should not be authorized like sharable resource requests. Let's remove this from here. This is similar to what's being discussed here: https://github.com/opensearch-project/security/pull/5600#discussion_r2325920820

DarshitChanpura: I'm confused. Are you saying that they should be handled through the new DLS approach we are introducing in #5600 or are you saying that we don't need to handle it all?

cwperks: All of the actions that go in a resource access level are actions that pertain to a single resource. For example, `Edit Report Definition`. As part of handling the report definition, there may be a system-level IndexRequest or UpdateRequest to change the metadata, but the first gate is answering the question "Does the authenticated user have permission to Edit the Report Definition" which would be authorized as a Resource Request.

Index perms should be authorized the same regardless if its a GET on a regular index, alias, data stream, non-resource index system index or a resource system index.

If system index protection is disabled and a user tries to get a doc from a resource index, then we should treat that as getting a document from a regular index.

We want to encourage plugins to create wrappers around getting a document from a system index.

Take for examples [GET Task API](https://docs.opensearch.org/latest/api-reference/tasks/get-tasks/).

```
GET /_tasks/{task_id}
```

As part of handling the request it will perform a transport request similar to `GET .tasks/_doc/{task_id}`, but that's part of the internal handling of the request.

DarshitChanpura: okay, so we are saying that resource-access-control will not work if system index protection is disabled.

What about custom indexes that we may support in the future?

cwperks: It would still work. Consider how authz would work for getting a reporting definition.

Remember the 2 gates:

1. **Authenticated user gate** - Resource authz is applicable here. The user calls the rest API `GET _plugins/_reporting/definition/{reportId}` and the first action that gets authorized would be a subclass of `DocRequest` where resource authz is performed

2. **Plugin System gate** - Assuming the first gate is passed, the plugin then tries to perform a GetRequest to get a doc from the `.opendistro-report-definitions` index. This GetRequest would be authorized as a plugin system-level request.

RESOURCE_SHARING_AND_ACCESS_CONTROL.md
cwperks: We should move over to resource-type everywhere. I think this needs to be built in such a way where multiple different sharable resource types can co-exist in the same resource index (thinking forward to saved objects in the .kibana indices)

src/main/java/org/opensearch/security/resources/ResourceActionGroupsHelper.java
```diff
@@ -0,0 +1,159 @@
...
+public class ResourceActionGroupsHelper {
+    public static final Logger log = LogManager.getLogger(ResourceActionGroupsHelper.class);
+
+    /**
+     * Loads action-groups config from the {@code resource-action-groups.yml} file from each resource sharing extension
+     * @param resourcePluginInfo will store the loaded action-groups config
+     *
+     * Sample yml file:
+     *   resource_types:
+     *   org.opensearch.sample.SampleResource:
```
cwperks: nit: let's update this javadoc

src/main/java/org/opensearch/security/resources/ResourceActionGroupsHelper.java
```diff
@@ -0,0 +1,159 @@
...
+import org.apache.logging.log4j.Logger;
+
+import org.opensearch.security.securityconf.FlattenedActionGroups;
+import org.opensearch.security.securityconf.impl.CType;
+import org.opensearch.security.securityconf.impl.SecurityDynamicConfiguration;
+import org.opensearch.security.securityconf.impl.v7.ActionGroupsV7;
+
+import org.yaml.snakeyaml.Yaml;
+
+// CS-SUPPRESS-SINGLE: RegexpSingleline get Resource Sharing Extensions
```
cwperks, line 25: We should probably remove the check to keep out the word extension soon. Given that the extensions effort is not under active development I don't think it makes sense to keep it, especially as we make modifications for extensible/extending plugin.

src/main/java/org/opensearch/security/resources/ResourceActionGroupsHelper.java
```diff
@@ -0,0 +1,159 @@
...
+     * @param resourcePluginInfo will store the loaded action-groups config
+     *
+     * Sample yml file:
+     *   resource_types:
+     *   org.opensearch.sample.SampleResource:
+     *       sample_read_only:
+     *         - "cluster:admin/sample-resource-plugin/get"
+     *         - "indices:data/read*"
+     */
+    public static void loadActionGroupsConfig(ResourcePluginInfo resourcePluginInfo) {
```
cwperks, line 44: I think actionGroup makes sense for the backend, but on the UI I'd like to refer to these as Access Levels. They are not mutually exclusive and different levels should progressively give more access.

i.e. in the Quip use-case the lowest level of access would be read, then read + comment, then read + comment, edit and the highest-level of access would be full access which is read, comment, edit + share

cwperks: I wonder if we can even assemble the higher-level access levels from the lower-level ones. i.e. use the action group name for `read` within the action group name for `full-access`.

src/main/java/org/opensearch/security/resources/ResourceActionGroupsHelper.java
```diff
@@ -0,0 +1,159 @@
...
+                        .removeIf(
+                            e -> e.getValue() == null
+                                || e.getValue().getAllowed_actions() == null
+                                || e.getValue().getAllowed_actions().isEmpty()
+                        );
+
+                    FlattenedActionGroups flattened = new FlattenedActionGroups(cfg);
+
+                    // Publish to ResourcePluginInfo → used by UI and authZ
+                    resourcePluginInfo.registerActionGroupNames(typeFqn, cfg.getCEntries().keySet());
```
cwperks: nice! It makes a lot of sense to include this in resourcePluginInfo.

src/main/java/org/opensearch/security/resources/ResourceActionGroupsHelper.java
```diff
@@ -0,0 +1,159 @@
...
+                    String typeFqn = rp.resourceType();
+
+                    Object typeCfgNode = byType.get(typeFqn);
+                    if (!(typeCfgNode instanceof Map<?, ?> typeMapRaw)) {
+                        log.info("No per-type block for {} in {}", typeFqn, ext.getClass().getName());
+                        continue; // no fallback
+                    }
+
+                    // buildActionGroupsYaml() accepts only lists-of-strings; anything else becomes allowed_actions: []
+                    String perTypeAgYaml = buildActionGroupsYaml(typeMapRaw);
```
cwperks: Would it be cleaner to use [SecurityDynamicConfiguration.fromMap](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/securityconf/impl/SecurityDynamicConfiguration.java#L121-L131)? We can update the javadoc to exclude for testing only.

src/main/java/org/opensearch/security/resources/ResourceActionGroupsHelper.java
```diff
@@ -0,0 +1,159 @@
...
+
+                Object rtNode = root.get("resource_types");
+                if (!(rtNode instanceof Map<?, ?> byType)) {
+                    log.warn("'resource_types' missing or invalid in {}", ext.getClass().getName());
+                    continue;
+                }
+
+                // For each type this extension provides, read its OWN map directly as groups (no wrapper)
+                for (var rp : ext.getResourceProviders()) {
+                    String typeFqn = rp.resourceType();
```
cwperks: nit: can we rename the var to `resourceType`?

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -202,11 +207,19 @@ public void hasPermission(
                 return;
             }
 
-            Set<String> allowedActions = roleBasedActionPrivileges.flattenedActionGroups().resolve(accessLevels);
-            WildcardMatcher matcher = WildcardMatcher.from(allowedActions);
+            // Fetch the static action-groups registered by plugins on bootstrap and check whether any match
+            final String resourceType = resourcePluginInfo.typeByIndex(resourceIndex);
```
cwperks, line 208: FYI raised https://github.com/opensearch-project/OpenSearch/pull/19313 in core so that we can do this from the action request itself. 

Currently this is assuming single resource-type per resource index. Thinking forward to dashboards saved objects, it should be built in a way that multiple resource types can co-exist in the same system index.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -388,12 +401,66 @@ public void revoke(
...
+    private Set<String> getFlatPrincipals(User user) {
+        // 1) collect all entities we’ll match against share_with arrays
+        // for users:
+        Set<String> users = new HashSet<>();
+        users.add(user.getName());
+        users.add("*"); // for matching against publicly shared resource
+
+        // for roles:
+        Set<String> roles = new HashSet<>(user.getSecurityRoles());
+        roles.add("*"); // for matching against publicly shared resource
```
cwperks: Let's use the convention of `*` only in users for publicly visible. This could even be a separate field to indicate if the resource is sharable which would improve searchability of these. 

In https://github.com/opensearch-project/security/pull/5600, there's a `all_shared_principals` field added to the resource doc itself (outside sharing index) and it would contain `user:*` as the convention for meaning the resource was publicly shared.

src/main/java/org/opensearch/security/resources/ResourceActionGroupsHelper.java
DarshitChanpura: yes, separate PR?

src/main/java/org/opensearch/security/resources/ResourceActionGroupsHelper.java
DarshitChanpura: > I wonder if we can even assemble the higher-level access levels from the lower-level ones. i.e. use the action group name for read within the action group name for full-access.

we definitely should.. In next version of access-levels for this feature?

src/main/java/org/opensearch/security/resources/ResourceActionGroupsHelper.java
DarshitChanpura: The problem there is the ConfigType, while we call this an action-group it doesn't exactly correlate with ActionGroupV7.java as CType.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
DarshitChanpura: yes, should we address this in a separate PR?

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
DarshitChanpura: okay, in that case we should update the public documentation.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -891,10 +936,190 @@ private void executeSearchRequest(
...
+     * Step-2:
+     *  - Use the terms query to search resource-sharing index to fetch matching records.
+     *
+     * @param resourceIndex the index for which records are to be searched
+     * @param user the user that is requesting the records
+     * @param flatPrincipals user's name, roles, backend_roles to be used for matching.
+     * @param listener to collect and return accessible sharing records
+     */
+    @SuppressWarnings("unchecked")
+    public void fetchAccessibleResourceSharingRecords(
```
cwperks, line 1006: Can this be re-written assuming that we will be able to obtain which resources are visible to a user given https://github.com/opensearch-project/security/pull/5600?

With that, you can directly search on the resource index (per page) get ids and then do an mget request for the respective sharing records.

DarshitChanpura: While I'm not against or for, I;ve addressed this approach in the latest commit.

DarshitChanpura: we will have to leave fetchAllResourceSharingRecords as is anyway since were simply need to fetch all records and mget requires explicit ids which we don't need for fetch all.

cwperks: Makes sense. The mGet in a batch is an interesting approach. I had in mind that for each scroll it would issue an mGet so the batch for the mGet is the scroll window size. In any event, I don't expect existing dashboards use-cases to have thousands of "resoures" so what exists here is pretty defensive. In the future, it would be interesting to see what people build on top of dashboards with a better plugin developer experience.

cwperks: As discussed, the central page being built into dashboards is primarily to help with migration from backend-role based visbility check to full-block resource sharing and authz. This code path is the exceptional case and not the norm. 

When the `Share` button is built into existing dashboards plugins, then the primary operation on the resource sharing indices are GET operations given the resource type and id.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -962,4 +1218,49 @@ private void clearScroll(String scrollId, ActionListener<Void> listener) {
...
+            if (!groupAllows(resourceType, level, ShareAction.NAME)) continue;
+
+            var recips = sw.atAccessLevel(level);
+            if (recips == null) continue;
+
+            var u = recips.getRecipients().getOrDefault(Recipient.USERS, Set.of());
+            var r = recips.getRecipients().getOrDefault(Recipient.ROLES, Set.of());
+            var b = recips.getRecipients().getOrDefault(Recipient.BACKEND_ROLES, Set.of());
+
+            boolean matches = u.contains("*")
```
cwperks: FYI I think we should move to a high-level flag in the resource sharing doc to indicate if something has been made public. It would be good for consistency and to remove checks like this.

RESOURCE_SHARING_AND_ACCESS_CONTROL.md
```diff
@@ -83,7 +84,47 @@ opensearchplugin {
...
+- **Declare** action-groups **per resource** in `resource-action-groups.yml` file:
+  ```
+  src/main/resources/resource-action-groups.yml
+  ```
+  - This file must be structured in a following way:
+    ```yml
+    resource_types:
+      <resource-type-1>:
+          <action-group-1>:
+          - <action1>
```
cwperks: why don't we add in `allowed_actions` here to avoid the normalizing step?

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: refactored and cleaned it more.

RESOURCE_SHARING_AND_ACCESS_CONTROL.md
DarshitChanpura: done.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: yes, central page is a sink for management of all resources. Soon, each dashboard component will have a pluggable share button where user can share the resource directly without having to navigate to the central dashboard.

RESOURCE_SHARING_AND_ACCESS_CONTROL.md
DarshitChanpura: yes, future iteration should have this along with the hierarchical access model like linux file system.

RyanL1997 (approved):

Issue #5603, opened 2025-09-03
Security Initialization Gaps/Issues


nagarajg17:
When the node boots up, security plugin creates security index, indexes all security docs into security index and initializes various security modules like BackendRegistry, PrivelegeEvaluation based on the index data as part of its initialization process. There are certain issues we've observed

1. Initialization failing silently without any errors thrown
2. Initialization process stuck for long time without exiting

These issues lead to customers receiving 5xx errors since security is not initialized yet which is leading to complete outage/availability issues

The code issues and improvements we need -
1. During indexation of default security docs into security index, all the errors are caught and silently ignored ([ref](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java#L275-L277)).  Due to this ES process comes up without security being initialized leading to customer receiving 5xx errors. There are certain expected errors like doc already exists. Apart from them all other errors should be thrown and initialization flow should fail. We need to add catch blocks for exceptions which can be ignored and which needs to be thrown back as error. This will fail ES process to come up and avoid customer receiving 5xx errors
2. We’ve a infinite while loop which waits for security initialization to be completed ([ref](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java#L280-L295)). We need to make this time bound and fail security initialization if it doesn’t complete within a time limit. Without time bound, it will be stuck in this while loop forever and also makes debugging very hard
3. No timeouts are configured for [index creation](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java#L325-L336) and [doc indexing operations](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/ConfigHelper.java#L95-L99). We’ve noticed in certain cases where this operations are being stuck. Due to no timeout, these operations do not end. This leaves initialization process in incomplete state and also makes debugging hard. We need to add timeouts

nibix:
Just my 5 cent:

At the moment, the code that tries intializing the security plugin by reading the configuration from the security index is very simplicistic:

- As soon as we get the call to `onNoteStarted()`, we try to read from the index. Loop until it works.

Because the indices are usually not yet ready during this time, it is just clear that this will hit errors in the beginning. As these errors would confuse cluster admins, one could argue that it is justify to silently ignore these errors.

[...]

PR #5625, opened 2025-09-14
Security init improvements

nagarajg17:
### Description

### Issues Resolved

### Testing

### Check List


src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -273,26 +277,33 @@ private void initalizeClusterConfiguration(final boolean installDefaultConfig) {
...
-            while (!dynamicConfigFactory.isInitialized()) {
+            for (int retryCount = 0; retryCount < MAX_RETRIES && !dynamicConfigFactory.isInitialized(); retryCount++) {
                 try {
-                    LOGGER.debug("Try to load config ...");
+                    LOGGER.info("Try to load config ...");
                     reloadConfiguration(CType.values(), true);
                     break;
                 } catch (Exception e) {
-                    LOGGER.debug("Unable to load configuration due to {}", String.valueOf(ExceptionUtils.getRootCause(e)));
+                    LOGGER.error("Unable to load configuration due to {}", String.valueOf(ExceptionUtils.getRootCause(e)));
```
kumjiten, line 291: Thanks, @nagarajg17, for the improvement.

please keep the logging format consistent for logging error, i.e., 
```
LOGGER.error("Unable to load configuration due to", e);
```

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -360,9 +372,12 @@ private void waitForSecurityIndexToBeAtLeastYellow() {
...
             }
             try {
-                response = client.admin().cluster().health(new ClusterHealthRequest(securityIndex).waitForYellowStatus()).actionGet();
+                response = client.admin()
+                    .cluster()
+                    .health(new ClusterHealthRequest(securityIndex).waitForYellowStatus().timeout(DEFAULT_TIMEOUT))
+                    .actionGet();
             } catch (Exception e) {
-                LOGGER.debug("Caught again a {} but we just try again ...", e.toString());
+                LOGGER.error("Caught again a {} but we just try again ...", e.toString());
```
kumjiten, line 380: please keep the logging format consistent for logging error, i.e., 
```
 LOGGER.error("Caught again but we just try again ...", e);
```

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -96,11 +96,15 @@
 import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.client.Client;
 
+import static org.opensearch.security.support.ConfigConstants.DEFAULT_TIMEOUT;
 import static org.opensearch.security.support.ConfigConstants.SECURITY_ALLOW_DEFAULT_INIT_USE_CLUSTER_STATE;
 import static org.opensearch.security.support.SnapshotRestoreHelper.isSecurityIndexRestoredFromSnapshot;
 
 public class ConfigurationRepository implements ClusterStateListener, IndexEventListener {
     private static final Logger LOGGER = LogManager.getLogger(ConfigurationRepository.class);
+    private static final int INITIALIZATION_RETRY_INTERVAL_MS = 3000;
```
kumjiten, line 105: good to move these constants in  src/main/java/org/opensearch/security/support/ConfigConstants.java

src/main/java/org/opensearch/security/support/ConfigHelper.java
```diff
@@ -95,6 +96,7 @@ public static void uploadFile(
                 final IndexRequest indexRequest = new IndexRequest(index).id(configType)
                     .opType(OpType.CREATE)
                     .setRefreshPolicy(RefreshPolicy.IMMEDIATE)
+                    .timeout(DEFAULT_TIMEOUT)
```
kumjiten, line 99: What happens when a cluster is large and overloaded, taking more than 30 seconds for indexing? Shouldn't the timeout value be configurable like a normal indexing API?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -341,14 +353,14 @@ private void waitForSecurityIndexToBeAtLeastYellow() {
         try {
             response = client.admin()
                 .cluster()
-                .health(new ClusterHealthRequest(securityIndex).waitForActiveShards(1).waitForYellowStatus())
+                .health(new ClusterHealthRequest(securityIndex).waitForActiveShards(1).waitForYellowStatus().timeout(DEFAULT_TIMEOUT))
                 .actionGet();
         } catch (Exception e) {
-            LOGGER.debug("Caught a {} but we just try again ...", e.toString());
+            LOGGER.error("Caught a {} but we just try again ...", e.toString());
```
kumjiten, line 359: please keep the logging format consistent for logging error, i.e.,

```
LOGGER.error("Caught exception but we just try again ...", e);
```

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -273,26 +277,33 @@ private void initalizeClusterConfiguration(final boolean installDefaultConfig) {
                         LOGGER.error("{} does not exist", confFile.getAbsolutePath());
                     }
                 } catch (Exception e) {
-                    LOGGER.error("Cannot apply default config (this is maybe not an error!)", e);
+                    LOGGER.error("Cannot apply default config", e);
+                    throw e;
```
nibix, line 281: Could you elaborate on why it is ok to abort the process here by throwing an exception?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -273,26 +277,33 @@ private void initalizeClusterConfiguration(final boolean installDefaultConfig) {
...
-                        LOGGER.debug("Thread was interrupted so we cancel initialization");
+                        LOGGER.error("Thread was interrupted so we cancel initialization");
                         break;
                     }
                 }
             }
+
+            if (!dynamicConfigFactory.isInitialized()) {
+                LOGGER.error("Node '{}' failed to initialize", clusterService.localNode().getName());
+                throw new IllegalStateException(String.format("Node '%s' failed to initialize", clusterService.localNode().getName()));
```
nibix, line 304: This will be caught and logged in line 311:

```
        } catch (Exception e) {
            LOGGER.error("Unexpected exception while initializing node " + e, e);
        }
```

Thus, basically you will get the log message `"Node '{}' failed to initialize` twice in a row.

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -426,6 +427,8 @@ public enum RolesMappingResolution {
     public static final String OPENSEARCH_RESOURCE_SHARING_ENABLED = "plugins.security.experimental.resource_sharing.enabled";
     public static final boolean OPENSEARCH_RESOURCE_SHARING_ENABLED_DEFAULT = false;
 
+    public static final TimeValue DEFAULT_TIMEOUT = TimeValue.timeValueSeconds(30);
```
nibix, line 430: I think we should not have a general purpose globall default timeout. 

Timeout handling is tricky, thus we should look on a case by case basis what timeouts are suitable. Initialization processes might have a much longer timeout than for example API calls triggered by users.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
```diff
@@ -341,14 +353,14 @@ private void waitForSecurityIndexToBeAtLeastYellow() {
         try {
             response = client.admin()
                 .cluster()
-                .health(new ClusterHealthRequest(securityIndex).waitForActiveShards(1).waitForYellowStatus())
+                .health(new ClusterHealthRequest(securityIndex).waitForActiveShards(1).waitForYellowStatus().timeout(DEFAULT_TIMEOUT))
```
nibix, line 356: ClusterHealthRequest and CreateIndexRequest already use 30s as a default value for timeouts. 

Could you explain the rationale for setting the timeout value explicitly?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nibix: We are here in a retry loop

```
        while (response == null || response.isTimedOut() || response.getStatus() == ClusterHealthStatus.RED) {
```

That means that potentially we will log error messages repeatedly. Are we sure this won't be spamming the logs with bogus error messages in "good" cases?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nibix: We are here in a retry loop, which is potentially executed every 3 seconds. Thus, we might log an error message every 3 seconds. Are we sure this won't spam the logs in some cases?

nibix (commented):
Thank you for the work!

On a high level, could you explain the approach? It is mostly just making the init procedure in case of an absent security index time bound, correct?

Additional note: These changes only apply for the cases that the security index does not exist yet, i.e. a fresh cluster or a cluster where security is enabled at some point.

From the description of #5603, it was not exactly clear to me that this was only about this case. Are we also going to look at the case that the index does already exist or is this not a concern here?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nagarajg17: The problem we see is since it is set to debug logs, when we try to investigate during initialization failures there won't be any debug logs and becomes hard to debug. Since it is in catch block, do you see it can get logged every 3 secs? Ideally we don't expect code flow to come here right?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nagarajg17: Ack, will remove line 303 where we are logging first time

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nagarajg17: We have seen good number of times where these operations are stuck like below. In this case  during indexing of `'rolesmapping'` doc. Are we sure it is having default value of 30 secs? We have seen many such tickets  

```
3603:[2025-06-09T14:32:01,165][INFO ][o.o.s.s.ConfigHelper     ] [bd87625b46a927da71043d943ba615a0] Will update 'config' with /apollo/_env/OS_2_19AMI-ES2-p001-d0db_OS_219/opensearch-security/config.yml and populate it with empty doc if file missing and populateEmptyIfFileMissing=false
4031:[2025-06-09T14:32:01,620][INFO ][o.o.s.s.ConfigHelper     ] [bd87625b46a927da71043d943ba615a0] Will update 'roles' with /apollo/_env/OS_2_19AMI-ES2-p001-d0db_OS_219/opensearch-security/roles.yml and populate it with empty doc if file missing and populateEmptyIfFileMissing=false
4038:[2025-06-09T14:32:01,759][INFO ][o.o.s.s.ConfigHelper     ] [bd87625b46a927da71043d943ba615a0] Will update 'rolesmapping' with /apollo/_env/OS_2_19AMI-ES2-p001-d0db_OS_219/opensearch-security/roles_mapping.yml and populate it with empty doc if file missing and populateEmptyIfFileMissing=false


4030:[2025-06-09T14:32:01,620][INFO ][o.o.s.s.ConfigHelper     ] [bd87625b46a927da71043d943ba615a0] Index .opendistro_security already contains doc with id config, skipping update.
4037:[2025-06-09T14:32:01,759][INFO ][o.o.s.s.ConfigHelper     ] [bd87625b46a927da71043d943ba615a0] Index .opendistro_security already contains doc with id roles, skipping update
```

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nagarajg17: Should we use warn level? Since its infinite while loop, how do you suggest to get visibility if code is stuck in this block?

src/main/java/org/opensearch/security/support/ConfigConstants.java
nagarajg17: Agree, can remove this. Timeout currently used is for index creation and default docs indexation, do you see any scenario taking more than 30 sec? What is your recommendation for ideal timeout value here?

src/main/java/org/opensearch/security/support/ConfigHelper.java
nagarajg17: What is the timeout for normal indexing API?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nagarajg17: The error is due to security index creation/default docs indexation failure (Excluding doc already exists error). So it might be partially indexed or not indexed at all depending on failure point. So there is no good benefit to allow the code to process further when indexing itself is not complete and more than that it can leave system in partially initialized state and customer API calls can receive 4xx/5xx errors which isn't desired behaviour

nagarajg17:
@nibix 
> On a high level, could you explain the approach? It is mostly just making the init procedure in case of an absent security index time bound, correct?

Yes making init procedure time bound

> Additional note: These changes only apply for the cases that the security index does not exist yet, i.e. a fresh cluster or a cluster where security is enabled at some point.

Yes for new cluster/security enabled at some point

> From the description of https://github.com/opensearch-project/security/issues/5603, it was not exactly clear to me that this was only about this case. Are we also going to look at the case that the index does already exist or is this not a concern here?
 
`installDefaultConfig` will be still true for a non fresh cluster. When say ex: ES process is restarted on node, it goes through same flow but since security index, docs are already present, its a no-op

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nibix: From my experience, during the early node startup phase, OpenSearch can throw sometimes exceptions you do not see otherwise (something like cluster state not intialized or similar).

I cannot give you any concrete examples, but I would recommend to thoroughly test this whether this might trip over such exceptions.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nibix: I cannot really say if we expect control flow to reach this place. I guess this needs to be found out by testing.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nibix: The default timeout value is here:

https://github.com/opensearch-project/OpenSearch/blob/6fdb010ef91b075f4060a6d87da9beae681c2c37/server/src/main/java/org/opensearch/action/admin/cluster/health/ClusterHealthRequest.java#L65

I am not sure if I understand the logs correctly. Are these from two different nodes? Nothing follows after these logs?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nibix: Maybe: Only start logging errors after being in this loop for more than 2 minutes? I think one has to experiment a bit (with big clusters) to find a good value for this.

If you had issues with stuck threads: Did you collect thread dumps? That is usually a good way to see what's going on.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nagarajg17: No, logs are from same domain. Node id - `bd87625b46a927da71043d943ba615a0`

This is for ClusterHealthRequest only?
https://github.com/opensearch-project/OpenSearch/blob/6fdb010ef91b075f4060a6d87da9beae681c2c37/server/src/main/java/org/opensearch/action/admin/cluster/health/ClusterHealthRequest.java#L65

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nagarajg17: Got it. We can bake this change for a week and see if there is any unexpected regression?

nibix: With "bake" do you mean comitting it to the repo and see if regressions show up in the nightlies of this or other repos? IMHO, that would be too little exposure to get an reliable impression. Or did you think of something else?

nibix:
> installDefaultConfig will be still true for a non fresh cluster. 

I guess this is the case for the managed service? From a general POV, this cannot be assumed IMHO.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
nibix: yes, 30 seconds is the default for ClusterHealthRequest. For indexing documents, the default is 1 minute: 

https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/replication/ReplicationRequest.java

src/main/java/org/opensearch/security/support/ConfigConstants.java
nibix: In the loop inside `waitForSecurityIndexToBeAtLeastYellow()` it does not matter so much IMHO, as it is retrying indefinitely. For the IndexRequest, one might consider increasing it to something more than the default of 1 minute, as reaching the timeout will be a fatal failure. Yet, this is all only an "IMHO", experiences have to be collected by testing.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
DarshitChanpura: if it's not a configurable constant then its fine to leave it here.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
DarshitChanpura: What type of exceptions do we expect here? Instead of aborting on a generic exception can we isolate and  abort on a specific expected type of exception?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
DarshitChanpura: +1, can we log an error message only retryCount exceeded?

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
DarshitChanpura: instead of throwing an exception we can simply log the message and return early.

src/main/java/org/opensearch/security/configuration/ConfigurationRepository.java
DarshitChanpura: +1 to Nils comment, error should only be thrown when its definitive and we have exhausted retry mechanism

src/main/java/org/opensearch/security/support/ConfigConstants.java
DarshitChanpura: Can we remove this global timeout and keep it closer to the actual requests. Also this file is for configurable constants and so we should not any non-configurable settings here.

DarshitChanpura (commented):
@nagarajg17 Do you still intend to work on this PR?

PR #5632, opened 2025-09-16
Thorough integration tests for index API authorization

nibix:
### Description

This introduces new integration tests for authorization against index APIs.

In order to be as thorough as possible, the test suites use parameters to define several test dimensions:

- Users with different privileges
- Cluster configurations (at the moment system index privilege disabled vs enabled)
- The final dimension are different operations with different options; this is manifested by the test methods

This is mostly a preparation for https://github.com/opensearch-project/security/pull/5399; it will give us a good overview over the changed behaviors and allows us to judge whether we want to change a behavior or not.

* Category: Tests
* What is the old behavior before changes and new behavior after changes? No behavioral changes

### Check List


src/integrationTest/java/org/opensearch/security/privileges/int_tests/ClusterConfig.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.function.Supplier;
+
+import org.opensearch.test.framework.cluster.LocalCluster;
+
+/**
+ * This is one of the test parameter dimensions used by the *Authorization*IntTests test suites.
+ * The test suites run on different cluster configurations; the possible cluster configurations are defined here.
+ */
+public enum ClusterConfig {
+    LEGACY_PRIVILEGES_EVALUATION(
```
cwperks, line 25: To make sure I understand correctly, `legacy` here is related to how DNFOF relates to system index authorization?

cwperks:
Thank you @nibix ! This is quite a large PR (even if only test files). It may make sense to break this up into smaller PRs for sake of review similar to what was done to the `integrationTest` framework. Would there be any good opportunities for splitting this PR up into smaller ones?

src/integrationTest/java/org/opensearch/security/privileges/int_tests/ClusterConfig.java
nibix: Good point. 

The naming is already in anticipation of the enhanced privilege evaluation PR which introduces one further element to the enum; thus, the enum looks like this:

https://github.com/opensearch-project/security/blob/b3e2cca6b2417da218152649dfa283e3fea6d06a/src/integrationTest/java/org/opensearch/security/privileges/int_tests/ClusterConfig.java#L20-L38

Thus, currently, this is not really legacy. But maybe we can keep it this way, as it is only temporary and avoids lots of renaming?

But, generally, I am also not sure about a good way to name these things. Also I am not too happy about this "next_gen" name.

nibix:
@cwperks 

> Would there be any good opportunities for splitting this PR up into smaller ones?

Well, we have 5 main test classes:

- `IndexAuthorizationReadOnlyIntTests`
- `IndexAuthorizationReadWriteIntTests`
- `DataStreamAuthorizationReadOnlyIntTests`
- `DataStreamAuthorizationReadWriteIntTests`
- `SnapshotAuthorizationIntTests`

Thus, one could break it down into 5 PRs:

- `IndexAuthorizationReadOnlyIntTests` and all the framework stuff
- One PR for the remaining classes

Still, I am not sure if this will be really helpful, as this does not really change the grouping of the changes. Already now, you could review class by class.

Also, the test cases themselves are then quite repetetive; if you reviewed the first few test cases, you will likely just skim over the remaining ones. In a few cases, these document some oddities; but in most cases, these will be addressed in the enhanced index resolution PR. In the end, this PR is only a preparation for the next one.

build.gradle
```diff
@@ -565,13 +564,17 @@ allprojects {
...
-        integrationTestImplementation "org.apache.commons:commons-collections4:${versions.commonscollections4}"
+        integrationTestImplementation 'org.apache.commons:commons-collections4:4.5.0'
         integrationTestImplementation('com.flipkart.zjsonpatch:zjsonpatch:0.4.16'){
             exclude(group:'com.fasterxml.jackson.core')
             exclude(group: 'org.apache.commons', module: 'commons-collections4')
         }
         integrationTestImplementation 'org.slf4j:slf4j-api:2.0.12'
         integrationTestImplementation 'com.selectivem.collections:special-collections-complete:1.4.0'
+
+        integrationTestImplementation ('com.jayway.jsonpath:json-path:2.9.0') {
```
DarshitChanpura, line 575: why do we need this? would defaultObjectMapper not work?

src/integrationTest/java/org/opensearch/test/framework/TestSecurityConfig.java
```diff
@@ -538,6 +541,20 @@ public Set<String> getRoleNames() {
...
+            return description;
+        }
+
+        @Override
+        public boolean isAdminCertUser() {
+            return adminCertUser;
+        }
+
+        public User adminCertUser() {
+            this.adminCertUser = true;
```
DarshitChanpura, line 554: i'm confused, how does this validate that "this" user is an adminCert user? or is the point not to validate?

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
```diff
@@ -0,0 +1,764 @@
...
+            return containsExactly(Arrays.asList(testIndices));
+        }
+
+        public static OnResponseIndexMatcher containsExactly(Collection<TestIndexOrAliasOrDatastream> testIndices) {
+            Map<String, TestIndexOrAliasOrDatastream> indexNameMap = new HashMap<>();
+            boolean containsOpenSearchIndices = false;
+
+            for (TestIndexOrAliasOrDatastream testIndex : testIndices) {
+                if (testIndex == TestIndex.openSearchSecurityConfigIndex()) {
+                    containsOpenSearchIndices = true;
```
DarshitChanpura: should this be renamed to containsSecurityIndex?

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
```diff
@@ -0,0 +1,764 @@
...
+                    if (response.getStatusCode() != this.statusCodeWhenEmpty.statusCode()) {
+                        mismatchDescription.appendText("Status was: ")
+                            .appendValue(response.getStatusCode() + " " + response.getStatusReason())
+                            .appendText("\n\n")
+                            .appendText(formatResponse(response));
+                        return false;
+                    }
+
+                    if (response.getStatusCode() != 200) {
+                        return true;
```
DarshitChanpura, line 278: why return true on non 200 code? is it because of empty list of indices and non-existent status code entry ?

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
```diff
@@ -0,0 +1,764 @@
...
+        public boolean isEmpty() {
+            return expectedIndices.isEmpty();
+        }
+
+        @Override
+        public int size() {
+            if (!containsOpenSearchSecurityIndex) {
+                return expectedIndices.size();
+            } else {
+                throw new RuntimeException("Size cannot be exactly specified because containsOpenSearchIndices is true");
```
DarshitChanpura: what is the purpose of this?

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
```diff
@@ -0,0 +1,764 @@
...
+            TestRestClient.HttpResponse response
+        ) {
+            ImmutableSet<String> expectedIndices = this.getExpectedIndices();
+            ImmutableSet.Builder<String> seenIndicesBuilder = ImmutableSet.builderWithExpectedSize(expectedIndices.size());
+            ImmutableSet.Builder<String> seenOpenSearchIndicesBuilder = new ImmutableSet.Builder<>();
+
+            for (Object object : collection) {
+                String index = object.toString();
+
+                if (containsOpenSearchSecurityIndex && (index.startsWith(".opendistro"))) {
```
DarshitChanpura: why only .opendistro prefix?

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
```diff
@@ -0,0 +1,764 @@
...
+                    seenOpenSearchIndicesBuilder.add(index);
+                } else if (index.startsWith(".ds-")) {
+                    // We do a special treatment for data stream backing indices. We convert these to the normal data streams if expected
+                    // indices contains these.
+                    java.util.regex.Matcher matcher = DS_BACKING_INDEX_PATTERN.matcher(index);
```
DarshitChanpura, line 463: nit: inline import

src/integrationTest/java/org/opensearch/test/framework/cluster/TestRestClient.java
```diff
@@ -495,4 +525,14 @@ public void close() {
         // TODO: Is there anything to clean up here?
     }
 
+    public static HttpEntity json(Object... attributes) {
+        Map<String, Object> map = new HashMap<>();
+
+        for (int i = 0; i < attributes.length - 1; i += 2) {
```
DarshitChanpura, line 536: what does this logic do?

src/integrationTest/java/org/opensearch/test/framework/data/TestIndexOrAliasOrDatastream.java
```diff
@@ -0,0 +1,58 @@
...
+    Set<String> documentIds();
+
+    Map<String, TestData.TestDocument> documents();
+
+    void create(Client client);
+
+    void delete(Client client);
+
+    default TestIndexOrAliasOrDatastream intersection(TestIndexOrAliasOrDatastream other) {
+        if (other == this) {
```
DarshitChanpura: probly unnecessary check

src/integrationTest/java/org/opensearch/test/framework/data/TestAlias.java
```diff
@@ -0,0 +1,153 @@
...
+
+    TestAlias(String name, ImmutableSet<TestIndexOrAliasOrDatastream> indices, TestIndexOrAliasOrDatastream writeIndex, boolean hidden) {
+        this.name = name;
+        this.indices = indices;
+        this.writeIndex = writeIndex;
+        this.hidden = hidden;
+    }
+
+    public TestAlias on(TestIndexOrAliasOrDatastream... indices) {
+        return new TestAlias(this.name, ImmutableSet.copyOf(indices), this.writeIndex, this.hidden);
```
DarshitChanpura, line 52: why create a new object here instead of returning "this"?

src/integrationTest/java/org/opensearch/test/framework/data/TestData.java
```diff
@@ -220,6 +209,53 @@ public void createIndex(Client client, String name, Settings settings) {
...
+                    log.debug("refresh at " + i + " " + g + " " + (g * 0.5 + 1));
+                }
+
+                if (i > nextRollover) {
+                    client.admin().indices().rolloverIndex(new RolloverRequest(name, null));
+
+                    nextRollover += rolloverAfter;
+                }
+
+                i++;
```
DarshitChanpura, line 244: what does the refresh and rollover logic do here?

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DataStreamAuthorizationReadOnlyIntTests.java
```diff
@@ -0,0 +1,882 @@
...
+    @Test
+    public void search_noPattern_allowNoIndicesFalse() throws Exception {
+        try (TestRestClient restClient = cluster.getRestClient(user)) {
+            TestRestClient.HttpResponse httpResponse = restClient.get("_search?size=1000&allow_no_indices=false");
+
+            assertThat(
+                httpResponse,
+                containsExactly(ALL_INDICES_EXCEPT_SYSTEM_INDICES).at("hits.hits[*]._index")
+                    .reducedBy(user.reference(READ))
+                    .whenEmpty(clusterConfig.allowsEmptyResultSets ? isNotFound() : isForbidden())
```
DarshitChanpura, line 285: i'm confused reading this test. Will admin cert see 403 if allowsEmptyResultSets is set to false?

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DataStreamAuthorizationReadOnlyIntTests.java
```diff
@@ -0,0 +1,882 @@
...
+            );
+        }
+    }
+
+    @Test
+    public void search_indexPattern_minus() throws Exception {
+        try (TestRestClient restClient = cluster.getRestClient(user)) {
+            TestRestClient.HttpResponse httpResponse = restClient.get("ds_a*,ds_b*,-ds_b2,-ds_b3/_search?size=1000");
+            if (user == SUPER_UNLIMITED_USER || user == UNLIMITED_USER) {
+                // does not handle the expression ds_a*,ds_b*,-ds_b2,-ds_b3 in a way that excludes the data streams. See
```
DarshitChanpura, line 393: can you explain why is that the case ?

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DataStreamAuthorizationReadOnlyIntTests.java
```diff
@@ -0,0 +1,882 @@
...
+    public void search_indexPattern_noWildcards() throws Exception {
+        try (TestRestClient restClient = cluster.getRestClient(user)) {
+            TestRestClient.HttpResponse httpResponse = restClient.get(
+                "ds_a*,ds_b*/_search?size=1000&expand_wildcards=none&ignore_unavailable=true"
+            );
+            if (user == SUPER_UNLIMITED_USER || user == UNLIMITED_USER) {
+                assertThat(httpResponse, isOk());
+                assertThat(httpResponse, containsExactly().at("hits.hits[*]._index"));
+            } else {
+                // dnfof makes the expand_wildcards=none option ineffective
```
DarshitChanpura, line 484: so unlimited user will see no results but other users may see results?

src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexAuthorizationReadOnlyIntTests.java
```diff
@@ -0,0 +1,1910 @@
...
+            if (user == SUPER_UNLIMITED_USER || user == UNLIMITED_USER) {
+                // Users with full privileges get an empty result, like expected due to the expand_wildcards=none option
+                assertThat(httpResponse, isOk());
+                assertThat(httpResponse, containsExactly().at("hits.hits[*]._index"));
+            } else {
+                // The dnfof implementation has the effect that the expand_wildcards=none option is disregarded
+                // Additionally, the dnfof implementation has the effect that hidden indices might be included even though not requested
+                assertThat(
+                    httpResponse,
+                    containsExactly(clusterConfig.systemIndexPrivilegeEnabled ? ALL_INDICES : ALL_INDICES_EXCEPT_SYSTEM_INDICES).at(
```
DarshitChanpura, line 448: will unlimited user see less results than limited users?

src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexAuthorizationReadOnlyIntTests.java
```diff
@@ -0,0 +1,1910 @@
...
+        try (TestRestClient restClient = cluster.getRestClient(user)) {
+            TestRestClient.HttpResponse httpResponse = restClient.get(
+                "index_a*,index_b*,xxx_non_existing/_search?size=1000&ignore_unavailable=true"
+            );
+
+            // The presence of a non existing index has the effect that the other patterns are not resolved by IndexResolverReplacer
+            // This causes a few more 403 errors where the granted index patterns do not use wildcards
+
+            if (user == LIMITED_USER_B1 || user == LIMITED_USER_ALIAS_AB1) {
+                assertThat(httpResponse, isForbidden());
```
DarshitChanpura, line 752: should it be forbidden for B1 user with dnfof enabled?

src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexAuthorizationReadOnlyIntTests.java
```diff
@@ -0,0 +1,1910 @@
...
+                    // only requested index. If also non system indices are requested in the same request, it will require
+                    // the presence of the system index privilege for all indices. As this is not the case, the request
+                    // will be denied with a 403 error.
+                    assertThat(httpResponse, isForbidden());
+                } else {
+                    // The other users do not have privileges for the system index. The dnfof feature promises to filter
+                    // out indices without authorization from eligible requests. However, the SystemIndexAccessEvaluator
+                    // is not aware of this and just denies all these requests
+                    // See also https://github.com/opensearch-project/security/issues/5546
+                    assertThat(httpResponse, isForbidden());
```
DarshitChanpura, line 848: both blocks are checking same 403 response. Can be clubbed together?

src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexAuthorizationReadOnlyIntTests.java
```diff
@@ -0,0 +1,1910 @@
...
+        }
+    }
+
+    @Test
+    public void search_alias_pattern_negation() throws Exception {
+        try (TestRestClient restClient = cluster.getRestClient(user)) {
+            TestRestClient.HttpResponse httpResponse = restClient.get("alias_*,-alias_ab1/_search?size=1000");
+            assertThat(
+                httpResponse,
+                containsExactly(index_a1, index_a2, index_a3, index_b1, index_c1).at("hits.hits[*]._index")
```
DarshitChanpura, line 892: what effect did the negation have here?

src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexAuthorizationReadOnlyIntTests.java
```diff
@@ -0,0 +1,1910 @@
...
+
+    @Test
+    public void analyze_noIndex() throws Exception {
+        try (TestRestClient restClient = cluster.getRestClient(user)) {
+            TestRestClient.HttpResponse httpResponse = restClient.postJson("_analyze", "{\"text\": \"sample text\"}");
+
+            // _analyze without index is different from most other operations:
+            // Usually, the absence of an index means "all indices". For analyze, however, it means: "no index".
+            // However, the IndexResolverReplacer does not get this right; it assumes that all indices are requested.
+            // Thus, we get only through to this operation with full privileges for all indices
```
DarshitChanpura, line 1582: should we fix this?

build.gradle
nibix: JSON path allows you to use expressions like "$.data_streams[*].name" to extract information from JSON documents in a very versatile manner. This is not supported by Jackson (ObjectMapper).

src/integrationTest/java/org/opensearch/test/framework/TestSecurityConfig.java
nibix: This is evaluated here and has the effect that a admin client certificate is used to create the connection:

https://github.com/opensearch-project/security/blob/2c321d613c97dbddbd2b41b88123441493179a95/src/integrationTest/java/org/opensearch/test/framework/cluster/OpenSearchClientProvider.java#L107-L112

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
nibix: This is the case that an empty set of indices is expected.

If we have a 200 response, we have a valid response and can check this in the response body. But if we do not have 200 response, we must/can just assume that we did not get any indices and stop without further checks.

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
nibix: we also have ` org.hamcrest.Matcher` here, thus the inline import.

src/integrationTest/java/org/opensearch/test/framework/cluster/TestRestClient.java
nibix: A very simple helper to construct dynamic JSON objects; will add a comment.

src/integrationTest/java/org/opensearch/test/framework/data/TestAlias.java
nibix: These objects are using the immutable object pattern; this adds a bit of safety to the code.

src/integrationTest/java/org/opensearch/test/framework/data/TestData.java
nibix: For data streams, we want to have several generations of backing indices; this creates these generations. Will add a comment.

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DataStreamAuthorizationReadOnlyIntTests.java
nibix: Yes, on OpenSearch with dnfof enabled, but do_not_fail_on_forbidden_empty is false (default), you will get a 403 error if there are no indices with privileges. With do_not_fail_on_forbidden_empty, youll get an empty result instead.

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DataStreamAuthorizationReadOnlyIntTests.java
nibix: In the end, it is a bug. 

At the moment, this test asserts *a lot of* oddities. It is the goal of https://github.com/opensearch-project/security/pull/5399 to fix these.

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DataStreamAuthorizationReadOnlyIntTests.java
nibix: Yes, also this will be fixed by https://github.com/opensearch-project/security/pull/5399

src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexAuthorizationReadOnlyIntTests.java
nibix: yes (still, both will only see authorized indices), the limited users will just see more than they requested (the request shall request no indices).

src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexAuthorizationReadOnlyIntTests.java
nibix: That's also a bug, to be fixed in the follow up PR

src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexAuthorizationReadOnlyIntTests.java
nibix: For the first branch, this is a bug, for the else branch, this is expected behavior. That's why i kept it separate with the comments explaining the behavior.

src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexAuthorizationReadOnlyIntTests.java
nibix: This will be fixed by https://github.com/opensearch-project/security/pull/5399

src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexAuthorizationReadOnlyIntTests.java
nibix: Actually, it does have no effect. This is actually a bug in core. I have added a comment and will file an issue.

src/integrationTest/java/org/opensearch/test/framework/data/TestIndexOrAliasOrDatastream.java
nibix: The whole method was kind of unnecessary; I have removed it completely.

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
nibix: changed

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
nibix: that was a leftover from a situation where there might be more than one index. This is currently not the case, removed.

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
nibix: done

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
DarshitChanpura: we do not use inline imports IIRC. Can this be removed, maybe in a follow-up?

src/integrationTest/java/org/opensearch/test/framework/data/TestAlias.java
DarshitChanpura: okay, this should be fine since these are test classes.

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DataStreamAuthorizationReadOnlyIntTests.java
DarshitChanpura: shouldn't admin cert not see a 403 on any operation?

src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexAuthorizationReadOnlyIntTests.java
DarshitChanpura: this behavior is confusing to me. How would admin cert user end up seeing less results than a limited user.

nibix: When the do_not_fail_on_forbidden mode is active, the security plugin does the following:

- Try to gather from a request which indices are requested. It gets this wrong by always including hidden indices, evenn though they are not requested
- Reduce the indices to the ones where the user is authorized. This may include hidden indices.
- Update the request with the reduced indices. Becausee the reduced indices are based on the set of indices including hidden indices, the hidden indices are now suddenly included in the request, even though originally the were not.

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
nibix: Hmm, how shall we handle the name conflict betwen the two Matcher classes then?

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DataStreamAuthorizationReadOnlyIntTests.java
nibix: ah, i got the question wrong. clusterConfig.allowsEmptyResultSets refers to the special mode of do_not_fail_on_forbidden. As admin users won't get an empty result set in this test, they will never get a 404 or 403 error.

src/integrationTest/java/org/opensearch/test/framework/matcher/RestIndexMatchers.java
DarshitChanpura: I see. Let's leave it as is.

src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexAuthorizationReadOnlyIntTests.java
DarshitChanpura: i understand that but admin cert should always see same or more number of results than regular user, correct?

nibix: yes, that observed behavior is certainly a bug, which we will also fix with the improved index resolution

PR #5641, opened 2025-09-17
Add View API and Rollback API for Versioned Security Configurations

nagarajg17:
### Description
This PR is to address outstanding comments of [PR - 5431](https://github.com/opensearch-project/security/pull/5431) for adding View API and Rollback API for Versioned Security Configurations authored by @divyanshi-0402 

Read this PR for full description and details https://github.com/opensearch-project/security/pull/5431

### Issues Resolved
[5095](https://github.com/opensearch-project/security/issues/5093)

### Testing
Did same testing performed in this PR https://github.com/opensearch-project/security/pull/5431

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/dlic/rest/api/RestApiAdminPrivilegesEvaluator.java
+++ b/src/main/java/org/opensearch/security/dlic/rest/api/RestApiAdminPrivilegesEvaluator.java
@@ -70,9 +70,9 @@ public class RestApiAdminPrivilegesEvaluator {
         .put(Endpoint.ROLES, action -> buildEndpointPermission(Endpoint.ROLES))
         .put(Endpoint.ROLESMAPPING, action -> buildEndpointPermission(Endpoint.ROLESMAPPING))
         .put(Endpoint.TENANTS, action -> buildEndpointPermission(Endpoint.TENANTS))
-        .put(Endpoint.SSL, action -> buildEndpointActionPermission(Endpoint.SSL, action))
         .put(Endpoint.VIEW_VERSION, action -> buildEndpointPermission(Endpoint.VIEW_VERSION))
         .put(Endpoint.ROLLBACK_VERSION, action -> buildEndpointPermission(Endpoint.ROLLBACK_VERSION))
+        .put(Endpoint.SSL, action -> buildEndpointActionPermission(Endpoint.SSL, action))
         .build();
 
     private final ThreadContext threadContext;


src/integrationTest/java/org/opensearch/security/api/RollbackVersionApiIntegrationTest.java
```diff
@@ -0,0 +1,160 @@
...
+    public void setupConfigVersionsIndex() throws Exception {
+        try (TestRestClient client = localCluster.getRestClient(ADMIN_USER_NAME, DEFAULT_PASSWORD)) {
+            client.get("/_cluster/health?wait_for_status=yellow&timeout=30s");
+            client.delete("/.opensearch_security_config_versions");
+            client.put("/.opensearch_security_config_versions");
+            client.post("/_refresh");
+            client.get("/_cluster/health/.opensearch_security_config_versions?wait_for_status=yellow&timeout=5s");
+
+            String bulkPayload =
+                "{ \"index\": { \"_index\": \".opensearch_security_config_versions\", \"_id\": \"opensearch_security_config_versions\" } }\n"
```
cwperks: nit: now that java min compatibility is 21, let's use [Text Blocks](https://docs.oracle.com/en/java/javase/15/text-blocks/index.html)

src/integrationTest/java/org/opensearch/security/api/RollbackVersionApiIntegrationTest.java
```diff
@@ -0,0 +1,160 @@
...
+                + "        \"key1\": { \"dummy\": \"value1\" }"
+                + "      }"
+                + "    }"
+                + "  }"
+                + "} ] }\n";
+
+            var bulkResponse = client.postJson("/_bulk?refresh=true", bulkPayload);
+            assertThat("Bulk insert failed: " + bulkResponse.getBody(), bulkResponse.getStatusCode(), is(200));
+
+            client.post("/_refresh");
```
cwperks: specifying `refresh=true` and then doing an explicit refresh here is redundant. Let's just do this on the bulk operation.

src/integrationTest/java/org/opensearch/security/api/RollbackVersionApiIntegrationTest.java
```diff
@@ -0,0 +1,160 @@
...
+            var response = client.post(RollbackVersion("does-not-exist"));
+            assertThat(response.getStatusCode(), is(HttpStatus.SC_NOT_FOUND));
+            assertThat(response.getTextFromJsonBody("/message"), containsString("not found"));
+        });
+    }
+
+    @Test
+    public void testRollbackWhenOnlyOneVersion_shouldFail() throws Exception {
+        withUser(ADMIN_USER_NAME, DEFAULT_PASSWORD, client -> {
+            client.delete("/.opensearch_security_config_versions");
```
cwperks: A few issues with this:

- I know `.opensearch_security_config_versions` is not formally designated a system index, but if it is in the future then this test will start failing
- On fresh cluster wouldn't there only be one version? Why do we need to delete and insert a dummy value?

src/integrationTest/java/org/opensearch/security/api/ViewVersionApiIntegrationTest.java
```diff
@@ -0,0 +1,136 @@
...
+    protected Map<String, Object> getClusterSettings() {
+        Map<String, Object> settings = super.getClusterSettings();
+        settings.put(EXPERIMENTAL_SECURITY_CONFIGURATIONS_VERSIONS_ENABLED, true);
+        return settings;
+    }
+
+    @Before
+    public void setupIndexAndCerts() throws Exception {
+        try (TestRestClient client = localCluster.getRestClient(ADMIN_USER_NAME, DEFAULT_PASSWORD)) {
+            client.get("/_cluster/health?wait_for_status=yellow&timeout=30s");
```
cwperks: Can we use more realistic setup for this? When the cluster spins up, shouldn't v1 already exist?

src/integrationTest/java/org/opensearch/security/api/RollbackVersionApiIntegrationTest.java
nagarajg17: Not sure why it was written this way, have removing it in all places. V1 will be same as what is stored in security index, V2..N will be based on further updates to security index

src/integrationTest/java/org/opensearch/security/api/ViewVersionApiIntegrationTest.java
nagarajg17: Removed it

src/main/java/org/opensearch/security/dlic/rest/api/ViewVersionApiAction.java
```diff
@@ -0,0 +1,179 @@
...
+        builder.startObject();
+        builder.startArray("versions");
+        for (SecurityConfigVersionDocument.Version<?> ver : versions) {
+            builder.startObject();
+            builder.field("version_id", ver.getVersion_id());
+            builder.field("timestamp", ver.getTimestamp());
+            builder.field("modified_by", ver.getModified_by());
+            builder.field("security_configs");
+            Map<String, Object> plainConfigs = new LinkedHashMap<>();
+            for (Map.Entry<String, SecurityConfigVersionDocument.HistoricSecurityConfig<?>> entry : ver.getSecurity_configs().entrySet()) {
```
DarshitChanpura, line 127: IMO, we should not be storing entire document if only a small detail has changed. Instead can we store a git style diff in the snapshot then apply them in sequence?

if user wants to restore from the 4th latest snapshot we compute the diff from 1,2,3 and 4, and apply all of those on the fly. 

Storing entire object may not be worthwhile if snapshots are taken frequently and for users with large configs.

nagarajg17: Yes this was discussed during the design phase itself. It was a conscious decision to do this. Feature is experimental, so should be fine.  This decision will be revisited when removing it as experimental feature

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
```diff
@@ -0,0 +1,395 @@
...
+                    continue;
+                }
+
+                SecurityDynamicConfiguration<?> currentConfig = currentConfigs.get(cType);
+                if (currentConfig == null) {
+                    throw new IllegalArgumentException("Rollback aborted: Could not fetch current config for cType '" + cTypeName + "'");
+                }
+
+                SecurityDynamicConfiguration<?> rollBackConfig = SecurityDynamicConfiguration.empty(cType);
+                rollBackConfig.setSeqNo(currentConfig.getSeqNo());
```
cwperks, line 222: Are we allowed to overwrite this? My understanding is that this number only increases as modifications are made to a doc.

src/integrationTest/java/org/opensearch/security/api/RollbackVersionApiIntegrationTest.java
```diff
@@ -0,0 +1,122 @@
...
+        withUser(ADMIN_USER_NAME, DEFAULT_PASSWORD, client -> {
+            // To perform below test, delete all entries in .opensearch_security_config_versions index
+            String deleteQuery = """
+                    {
+                        "query": {
+                            "match_all": {}
+                        }
+                    }
+                """;
+            client.postJson("/.opensearch_security_config_versions/_delete_by_query", deleteQuery);
```
cwperks, line 104: Can this be moved to a method with `@Before` and done during test setup?

src/integrationTest/java/org/opensearch/security/api/RollbackVersionApiIntegrationTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+            // Insert 1 record in .opensearch_security_config_versions index
+            String bulkPayload =
+                """
+                       {"index":{"_index":".opensearch_security_config_versions","_id":"opensearch_security_config_versions"}}
+                       {"versions":[{"version_id":"v1","timestamp":"2025-04-03T00:00:00Z","modified_by":"admin","security_configs":{"config_type_1":{"lastUpdated":"2025-04-03T00:00:00Z","configData":{"key1":{"dummy":"value1"}}}}}]}
+                    """;
```
cwperks, line 111: nit: prefer using text blocks here for readability

cwperks:
@nagarajg17 we need to declare `.opensearch_security_config_versions` as a system index since it can contain secrets. Many of the tests in this PR will need to be changed after declaring it a system index. I know its marked experimental initially so I will delegate to @shikharj05 for input as well but I would really like to see that addressed prior to merge.

src/main/java/org/opensearch/security/dlic/rest/api/RollbackVersionApiAction.java
nagarajg17: This is for handling race conditions (optimistic locking). It was added intentionally to handle concurrent writes/updates to security index

```
Dummy example -

// Admin A reads role
GetResponse responseA = client.get("analyst_role");
long seqNoA = responseA.getSeqNo();        // 5
long primaryTermA = responseA.getPrimaryTerm(); // 2

// Admin B reads role
GetResponse responseB = client.get("analyst_role");
long seqNoB = responseB.getSeqNo();        // 5
long primaryTermB = responseB.getPrimaryTerm(); // 2

// Admin B updates first
IndexRequest requestB = new IndexRequest()
    .id("analyst_role")
    .setIfSeqNo(seqNoB)          // 5
    .setIfPrimaryTerm(primaryTermB)   // 2
    .source({"permissions": ["read_data", "write_logs"]});
// Update successful, SeqNo now 6

// Admin A tries to update
IndexRequest requestA = new IndexRequest()
    .id("analyst_role")
    .setIfSeqNo(seqNoA)          // 5
    .setIfPrimaryTerm(primaryTermA)   // 2
    .source({"permissions": ["read_data", "delete_index"]});
// Throws VersionConflictEngineException!
// Because seqNo is now 6, not 5
```

src/integrationTest/java/org/opensearch/security/api/RollbackVersionApiIntegrationTest.java
nagarajg17: This is only needed for this test case specifically where we try to rollback to `-1` version which won't exist. Its clearing all the present data in version index. This is for negative testing

src/integrationTest/java/org/opensearch/security/api/RollbackVersionApiIntegrationTest.java
nagarajg17: Had specified in multi line using text block. Bulk request expect in this format else it was failing. So had to keep this way. Compared to earlier messy way of string concatenation and new line chars over multiple lines, this looks much better

nagarajg17:
@cwperks Made it system index given the security risk

src/integrationTest/java/org/opensearch/security/api/RollbackVersionApiIntegrationTest.java
cwperks: Does this call actually succeed? I suspect its failing now that `.opensearch_security_config_versions` is declared as a system index.

DarshitChanpura: seems to be working based on the 404 condition check below and Ci is passing.

src/integrationTest/java/org/opensearch/security/api/ViewVersionApiIntegrationTest.java
```diff
@@ -0,0 +1,115 @@
...
+            assertThat(json.get("status").asText(), equalTo("NOT_FOUND"));
+
+            assertThat(json.has("message"), is(true));
+            assertThat(json.get("message").asText(), containsString("not found"));
+        });
+    }
+
+    @Test
+    public void testViewAllVersions_forbiddenWithoutAdminCert() throws Exception {
+        withUser("limitedUser", "limitedPass", client -> {
```
DarshitChanpura, line 110: tests above do not seem to be adding admin cert either. It simply sends admin username and password. Are we sure that we are testing the behavior as intended?

DarshitChanpura: or is the test misnamed?

src/main/java/org/opensearch/security/configuration/SecurityConfigVersionHandler.java
```diff
@@ -101,25 +101,29 @@ public void onChange(ConfigurationMap typeToConfig) {
...
-            String nextVersionId = fetchNextVersionId();
-            User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
-            String userinfo = (user != null) ? user.getName() : "system";
+                waitForOpendistroSecurityConfigVersionsIndexToBeAtLeastYellow();
 
-            Version<?> version = buildVersionFromSecurityIndex(nextVersionId, userinfo);
-            saveCurrentVersionToSystemIndex(version);
+                String nextVersionId = fetchNextVersionId();
+                User user = threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+                String userinfo = (user != null) ? user.getName() : "system";
```
DarshitChanpura, line 118: why the magic variable "system"?

src/main/java/org/opensearch/security/dlic/rest/api/SecurityRestApiActions.java
```diff
@@ -68,46 +71,71 @@ public static Collection<RestHandler> getHandler(
...
-                clusterService,
-                threadPool,
-                sslSettingsManager,
-                certificatesReloadEnabled,
-                securityApiDependencies
-            ),
-            new CertificatesApiAction(clusterService, threadPool, securityApiDependencies),
-            new MigrateResourceSharingInfoApiAction(clusterService, threadPool, securityApiDependencies, resourceSharingIndexHandler)
+        List<RestHandler> handler = new ArrayList<>(
+            List.of(
```
DarshitChanpura, line 75: nit: why are we making a copy of immutable list? we can directly add the entries in handler list.

src/test/java/org/opensearch/security/dlic/rest/api/ViewVersionApiActionValidationTest.java
```diff
@@ -0,0 +1,83 @@
...
+            .createEndpointValidator();
+
+        var result = validator.onConfigLoad(SecurityConfiguration.of("some_id", configuration));
+
+        assertTrue(result.isValid());
+        assertThat(result.status(), is(RestStatus.OK));
+    }
+
+    @Test
+    public void forbidsDeleteRequest() throws IOException {
```
DarshitChanpura, line 63: is this true for admin certificate?

nagarajg17: Its not about access. API doesn't implement DELETE functionality

src/main/java/org/opensearch/security/dlic/rest/api/SecurityRestApiActions.java
nagarajg17: Ack, will club with other changes Craig mentioned

src/main/java/org/opensearch/security/configuration/SecurityConfigVersionHandler.java
nagarajg17: This is from different PR. This is when user is null, just to handle that case. It can be true when we use admin certs I think. I'm not exactly sure when it can be null, instead storing null as user who modified it, we are using system. Open to suggestion if you have any which we pick along with other changes

src/integrationTest/java/org/opensearch/security/api/ViewVersionApiIntegrationTest.java
nagarajg17: The API is only available for admin user. This test checks with non admin user. Agree test name can be improved

nagarajg17: > seems to be working based on the 404 condition check below and Ci is passing

404 is for rollback API call not this

src/integrationTest/java/org/opensearch/security/api/RollbackVersionApiIntegrationTest.java
nagarajg17: We get 4xx, will remove this delete in next PR

nagarajg17: 404 is for rollback API call (negative testing check). That is expected resp

src/main/java/org/opensearch/security/configuration/SecurityConfigVersionHandler.java
DarshitChanpura: yea, let's revisit this in next iteration.

DarshitChanpura (approved):
LGTM from experimental feature release perspective. There are few action-items we need to revisit before marking this feature as experimental:
- git style diff instead of complete doc
- better naming of tests
- auditlogging the config version index changes.

PR #5647, opened 2025-09-19
Fixed TLS endpoint identification by SAN

nibix:
### Description

This fixes TLS endpoint identification

* Category: Bugfix

### Testing

- Integration test

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/TlsHostnameVerificationTests.java
+++ b/src/integrationTest/java/org/opensearch/security/TlsHostnameVerificationTests.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ */
+package org.opensearch.security;
+
+import java.util.Map;
+
+import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import org.opensearch.security.ssl.util.SSLConfigConstants;
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.test.framework.certificate.TestCertificates;
+import org.opensearch.test.framework.cluster.ClusterManager;
+import org.opensearch.test.framework.cluster.LocalCluster;
+import org.opensearch.test.framework.log.LogsRule;
+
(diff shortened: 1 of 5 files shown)


src/integrationTest/java/org/opensearch/security/TlsHostnameVerificationTests.java
```diff
@@ -0,0 +1,64 @@
...
+import org.opensearch.test.framework.log.LogsRule;
+
+@RunWith(com.carrotsearch.randomizedtesting.RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class TlsHostnameVerificationTests {
+
+    @Rule
+    public LogsRule logsRule = new LogsRule("org.opensearch.transport.netty4.ssl.SecureNetty4Transport");
+
+    public LocalCluster.Builder clusterBuilder = new LocalCluster.Builder().clusterManager(ClusterManager.THREE_CLUSTER_MANAGERS)
```
cwperks, line 34: Is it important to test with a multi-node cluster here?

src/integrationTest/java/org/opensearch/test/framework/certificate/TestCertificates.java
```diff
@@ -142,12 +148,12 @@ private void isCorrectNodeNumber(int node) {
...
     }
 
-    private CertificateData createNodeCertificate(Integer node) {
+    private CertificateData createNodeCertificate(Integer node, String nodeSanIp) {
         final var subject = String.format(NODE_SUBJECT_PATTERN, node);
         String domain = String.format("node-%d.example.com", node);
         CertificateMetadata metadata = CertificateMetadata.basicMetadata(subject, CERTIFICATE_VALIDITY_DAYS)
             .withKeyUsage(false, DIGITAL_SIGNATURE, NON_REPUDIATION, KEY_ENCIPHERMENT, CLIENT_AUTH, SERVER_AUTH)
-            .withSubjectAlternativeName("1.2.3.4.5.5", List.of(domain, "localhost"), "127.0.0.1");
+            .withSubjectAlternativeName("1.2.3.4.5.5", List.of(domain, "localhost"), nodeSanIp);
```
cwperks, line 156: I know its not introduced in this PR, but at some point I think we need to remove this OID from the default value [here](https://github.com/opensearch-project/security/blob/b8c76eed6f409ac76235c9594493665396dd1047/src/main/java/org/opensearch/security/transport/OIDClusterRequestEvaluator.java#L47).

src/integrationTest/java/org/opensearch/security/TlsHostnameVerificationTests.java
nibix: actually yes, because the int test verifies the cluster formation. If we only have one node, thats not really testable :)

PR #5657, opened 2025-09-26
Make configuration setting for user attribute serialization dynamic

markdboyd:
### Description

This PR makes the configuration setting for `plugins.security.user_attribute_serialization.enabled`dynamic, so it can be updated via requests to `PUT cluster/_settings`.

* Category - Enhancement
* Why these changes are required? - Makes the configuration of this setting more dynamic, which is especially useful in testing contexts
* What is the old behavior before changes and new behavior after changes? - Before: The setting could only be configured statically in `opensearch.yml`. After: The setting can be dynamically updated with requests to `PUT cluster/_settings`.

### Issues Resolved

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -2288,8 +2288,18 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
             return settingsFilter;
         }
         settingsFilter.add("opendistro_security.*");
-        settingsFilter.add("plugins.security.http.*");
-        settingsFilter.add("plugins.security.transport.*");
+        settingsFilter.add("plugins.security.transport_user_cache.*");
+        settingsFilter.add("plugins.security.nodes_dn.*");
+        settingsFilter.add("plugins.security.restapi.*");
+        settingsFilter.add("plugins.security.ssl.*");
+        settingsFilter.add("plugins.security.config_version.*");
+        settingsFilter.add("plugins.security.nodes_dn_dynamic_config_enabled.*");
+        settingsFilter.add("plugins.security.privileges_evaluation.*");
+        settingsFilter.add("plugins.security.authcz.*");
+        settingsFilter.add("plugins.security.password.*");
+        settingsFilter.add("plugins.security.unsupported.*");
+        settingsFilter.add("plugins.security.audit.*");
+        settingsFilter.add("plugins.security.compliance.*");
         return settingsFilter;
     }
 
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -2272,7 +2272,8 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
                     ConfigConstants.USER_ATTRIBUTE_SERIALIZATION_ENABLED,
                     ConfigConstants.USER_ATTRIBUTE_SERIALIZATION_ENABLED_DEFAULT,
                     Property.NodeScope,
-                    Property.Filtered
+                    Property.Filtered,
+                    Property.Dynamic
                 )
             );
         }
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -2272,7 +2272,6 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
                     ConfigConstants.USER_ATTRIBUTE_SERIALIZATION_ENABLED,
                     ConfigConstants.USER_ATTRIBUTE_SERIALIZATION_ENABLED_DEFAULT,
                     Property.NodeScope,
-                    Property.Filtered,
                     Property.Dynamic
                 )
             );
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -2267,14 +2267,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
                 )
             );
 
-            settings.add(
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -2272,7 +2272,8 @@ public List<Setting<?>> getSettings() {
                     ConfigConstants.USER_ATTRIBUTE_SERIALIZATION_ENABLED,
                     ConfigConstants.USER_ATTRIBUTE_SERIALIZATION_ENABLED_DEFAULT,
                     Property.NodeScope,
-                    Property.Filtered
+                    Property.Filtered,
+                    Property.Dynamic
```
kumjiten: Thanks, @markdboyd, for the dynamic implementation.

Quick question: Could toggling this property at runtime cause issues with in-flight requests or active scroll queries? The user context format change might lead to deserialization failures or inconsistent behavior for operations that span the configuration change.

cwperks: @kumjiten I don't think scroll would be affected. The user attributes are only needed for a subset of actions where the plugin persists these values in their system index. I suppose one enhancement could be to start collecting a list of those action names, but one problem with that is that it more tightly couples security with other plugins. Another enhancement could be to configure a list of attributes names that should be serialized and only serialize those attributes without serializing all attributes.

I think for anyone wanting to toggle this setting we should allow it. They can toggle back if they don't use DLS with user attributes substitution.

kumjiten: Agreed, we should allow toggling. 

If the consumer of https://github.com/opensearch-project/security/blob/eb7153d772e9e00d49d9cb5ffafb33b5f02399fc/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L309-L312 `OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT` if any, needs an update to deserialize the attributes, then toggling can result in failures. However, that is outside of the scope for this change.

cwperks: See https://github.com/opensearch-project/common-utils/pull/878#issuecomment-3340230255

nibix: Has this been tested? IMHO, it is not sufficient to add `Property.Dynamic` to a `Setting` object to make it dynamic; you also have to listen to settings updates. The way it is implemented currently, `PrivilegesEvaluator` will still read from the immutable `Settings` object that will never change.

See here for a minimal implementation:

https://github.com/opensearch-project/OpenSearch/blob/5e6ee972a7f0eb7b457bec50ed49f70ed52a7f72/server/src/main/java/org/opensearch/action/support/AutoCreateIndex.java#L76-L79

nibix (changes requested):

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
markdboyd: @nibix @kumjiten Doing manual testing, I found that I could not view or update this setting with the cluster settings API. After discussing with @cwperks, we found that:

- I needed to remove `Property.FILTERED` for my setting: f8169f81312dab359440923ef7d129646cdd90b6
- There are two instances of `getSettingsFilter` which seem to exclude all security plugin settings (`plugins.security.*`) from management by the cluster settings API by default. We updated these methods to change the default filters to something less restrictive: c0d8e2ad271d87d8ad3e33f665867ce2dfce2de2, since properties can also be filtered by using `Property.FILTERED`

markdboyd: Updated the default security settings filters: 1dbcbd97b67119190f5363c108c8e9bb1fe8fb9b

markdboyd: @nibix @cwperks @kumjiten I updated PrivilegesEvaluator to subscribe to updates on `plugins.security.user_attribute_serialization.enabled`:

7645558525384258007b0cbc983c522dd10f20b5
4c6c607fc813958eda02fce065f228b107c8600d

With these changes, in manual testing, I can see that when the default of `true` for the setting is applied, then DLS/FLS user attribute substitution works as expected. But when I use the the cluster settings API to change the setting to `false`, then I see failures in DLS/FLS attribute substitution as expected.

Are there any tests I need to add here? It seems like I could add a test to `src/integrationTest/java/org/opensearch/security/SecuritySettingsTests.java`, but I'm not sure if it it necessary.

nibix:
@markdboyd 

thanks for the changes!

Unfortunately, we have now unit test failures, can you please check? It's likely connected missing mocking or so:

```
  2> java.lang.NullPointerException: Cannot invoke "org.opensearch.common.settings.ClusterSettings.addSettingsUpdateConsumer(org.opensearch.common.settings.Setting, java.util.function.Consumer)" because "clusterSettings" is null
        at org.opensearch.security.privileges.PrivilegesEvaluator.registerClusterSettingsChangeListener(PrivilegesEvaluator.java:299)
        at org.opensearch.security.privileges.PrivilegesEvaluator.<init>(PrivilegesEvaluator.java:247)
        at org.opensearch.security.privileges.RestLayerPrivilegesEvaluatorTest.createPrivilegesEvaluator(RestLayerPrivilegesEvaluatorTest.java:163)
        at org.opensearch.security.privileges.RestLayerPrivilegesEvaluatorTest.testEvaluate_Successful_LegacyPermission(RestLayerPrivilegesEvaluatorTest.java:141)
```

cwperks:
after here https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluatorTest.java#L80

I think we need to add the following:

```
when(clusterService.getClusterSettings()).thenReturn(Settings.EMPTY);
```

cwperks:
actually need this:

```
when(clusterService.getClusterSettings()).thenReturn(new ClusterSettings(Settings.EMPTY, Set.of(USER_ATTRIBUTE_SERIALIZATION_ENABLED_SETTING)));
```

markdboyd:
The changes in this PR were merged as part of #5673. So closing this PR

PR #5658, opened 2025-09-26
[Bug Fix] Handles resource updates which otherwise may wipe out all_shared_principals

DarshitChanpura:
### Description
AD saw test failure during CI run which was result of `all_shared_principals` field being removed from the document on update request. This PR fixes it by prevent such scenarios and explicitly updating `all_shared_principals` on index-requests masked as updates.
https://github.com/opensearch-project/anomaly-detection/pull/1546

* Category: Bug fix
Search Request after updates would fail.

### Testing
Manual + Automated

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/resources/ResourceIndexListener.java
+++ b/src/main/java/org/opensearch/security/resources/ResourceIndexListener.java
@@ -54,8 +54,20 @@ public class ResourceIndexListener implements IndexingOperationListener {
         String resourceId = index.id();
 
         // Only proceed if this was a create operation and for primary shard
-        if (!result.isCreated() || !index.origin().equals(Engine.Operation.Origin.PRIMARY)) {
-            log.debug("Skipping resource sharing entry creation as this was an update operation for resource {}", resourceId);
+        if (!index.origin().equals(Engine.Operation.Origin.PRIMARY)) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/security/resources/ResourceIndexListener.java
```diff
@@ -54,8 +54,20 @@ public void postIndex(ShardId shardId, Engine.Index index, Engine.IndexResult re
...
 
         // Only proceed if this was a create operation and for primary shard
-        if (!result.isCreated() || !index.origin().equals(Engine.Operation.Origin.PRIMARY)) {
-            log.debug("Skipping resource sharing entry creation as this was an update operation for resource {}", resourceId);
+        if (!index.origin().equals(Engine.Operation.Origin.PRIMARY)) {
+            log.debug("Skipping resource sharing entry creation for {} as this operation was on a replica shard", resourceId);
+            return;
+        }
+
+        if (!result.isCreated()) {
```
cwperks, line 62: ~~Can we also add the primary shard check here as well?~~

cwperks: nvm, the will not be reached unless its on the primary shard.

cwperks (approved):

PR #5671, opened 2025-10-01
Adds a list setting to explicitly specify resources to be protected

DarshitChanpura:
### Description
Adds an opensearch setting that allow cluster-admins to specify which resources must be marked as protected.
```yml
plugins.security.experimental.resource_sharing.protected_types: []
```

* Category : Enhancement
* To allow individual control over which resources should be access-protected
* Without this change, by default, all resources onboarded to this feature will receive protection, meaning cluster-admins cannot opt out of resource-protection for specific types. With this change, such specification will be allowed.

### Issues Resolved
TBD

### Testing
Automated + manual

### Check List

### DIFF
--- a/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/feature/enabled/ExcludedResourceTypeTests.java
+++ b/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/feature/enabled/ExcludedResourceTypeTests.java
@@ -1,0 +1,89 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.sample.resource.feature.enabled;
+
+import java.util.List;
+
+import com.carrotsearch.randomizedtesting.RandomizedRunner;
+import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
+import org.apache.http.HttpStatus;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import org.opensearch.sample.resource.TestUtils;
+import org.opensearch.test.framework.cluster.LocalCluster;
+import org.opensearch.test.framework.cluster.TestRestClient;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.not;
+import static org.opensearch.sample.resource.TestUtils.FULL_ACCESS_USER;
+import static org.opensearch.sample.resource.TestUtils.LIMITED_ACCESS_USER;
+import static org.opensearch.sample.resource.TestUtils.NO_ACCESS_USER;
+import static org.opensearch.sample.resource.TestUtils.RESOURCE_SHARING_INDEX;
+import static org.opensearch.sample.resource.TestUtils.newCluster;
+import static org.opensearch.test.framework.TestSecurityConfig.User.USER_ADMIN;
+
+/**
+ * Test resource access when sample-resource is not marked as protected resource, even-though resource sharing protection is enabled.
+ */
+@RunWith(RandomizedRunner.class)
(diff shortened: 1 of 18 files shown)


src/main/java/org/opensearch/security/resources/ResourcePluginInfo.java
```diff
@@ -48,14 +49,18 @@ public class ResourcePluginInfo {
...
         resourceSharingExtensions.clear();
         typeToIndex.clear();
         indexToType.clear();
         // Enforce resource-type unique-ness
         Set<String> resourceTypes = new HashSet<>();
         for (ResourceSharingExtension extension : extensions) {
             for (var rp : extension.getResourceProviders()) {
+                // exclude resource types not mentioned in the explicit list. defaults to no resource marked as protected resources
+                if (protectedTypes.isEmpty() || !protectedTypes.contains(rp.resourceType())) {
+                    return;
```
RyanL1997:
```suggestion
                    continue;
```
I think for this one, you want to register only the resource types that are in the `protectedTypes` list. If the list is empty, you want to register nothing (no resources should be protected).

However for the following scenario of Multiple resource types, only some protected:

- `protectedTypes = ["dashboard"]`
- First iteration: `rp.resourceType() = "sample-resource"`
- Condition: `isEmpty()` is false, `contains("sample-resource")` is false → `!contains()` is true → entire OR is true
- Action: `return` → exits method immediately
- Result: "dashboard" never gets registered. The method exits before processing it.

RyanL1997 (commented):
Hi @DarshitChanpura , thanks for the change, and I just left a comment.

src/main/java/org/opensearch/security/resources/ResourcePluginInfo.java
RyanL1997: Not sure if the above is a valid scenario, but you can take a look to see if the above is necessary or not.

DarshitChanpura: yep. changed it.

RyanL1997 (dismissed):

DarshitChanpura:
@cwperks These two settings act independently. We do not necessarily negate filter_by_backend_role in security plugin, however, the plugins implementing this feature will have precedence over filter_by setting.

DarshitChanpura:
the tricky part however is that an extension will be assigned resource-sharing client even if no associated types were marked as protected. Which means, plugins using conditions like:
`ResourceSharingClientAccessor.getInstance().getResoiurceSharingClient()` to decide the code-path may end up using it incorrectly if resource is marked as not protected. I'm thinking of a way to mitigate this.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/feature/enabled/ExcludedResourceTypeTests.java
```diff
@@ -0,0 +1,89 @@
...
+import static org.opensearch.test.framework.TestSecurityConfig.User.USER_ADMIN;
+
+/**
+ * Test resource access when sample-resource is not marked as protected resource, even-though resource sharing protection is enabled.
+ */
+@RunWith(RandomizedRunner.class)
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class ExcludedResourceTypeTests {
+
+    // do not include sample resource as protected resource, should behave as if feature was disable for that resource
```
cwperks, line 43: Makes sense. I think this javadoc can be expanded in a follow-up to mention how the sample plugin doesn't do any filtering by backend role like other plugins so behavior would be expected that there is no filtering when resource sharing is disabled for `sample-resource` and anyone that can list/search would be able to see all.

src/main/java/org/opensearch/security/resources/api/share/ShareRequest.java
```diff
@@ -186,6 +201,7 @@ public void parseContent(XContentParser xContentParser, ResourcePluginInfo resou
                                 this.resourceId(parser.text());
                                 break;
                             case "resource_type":
+                                this.resourceType(parser.text());
```
cwperks, line 204: Nice, I started making similar change on https://github.com/cwperks/security/pull/66 to change the restriction of resource type <-> resource index being 1-to-1 to become many <-> 1

cwperks (dismissed):

RyanL1997 (approved):

RyanL1997 (approved):

PR #5677, opened 2025-10-02
Makes resource settings dynamic

DarshitChanpura:
### Description
Makes resource sharing settings dynamic. Also adds capability for plugin to chose codepath based on protected resource-types
* Category : Enhancement
* To allow settings to be update-able by customers through cluster-settings API

### Issues Resolved
TBD

### Testing
Manual + Automated Testing

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
+++ b/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
@@ -83,6 +83,7 @@ import org.opensearch.security.resources.ResourceSharingDlsUtils;
 import org.opensearch.security.securityconf.DynamicConfigFactory;
 import org.opensearch.security.securityconf.impl.SecurityDynamicConfiguration;
 import org.opensearch.security.securityconf.impl.v7.RoleV7;
+import org.opensearch.security.setting.OpensearchDynamicSetting;
 import org.opensearch.security.support.ConfigConstants;
 import org.opensearch.security.support.HeaderHelper;
 import org.opensearch.security.support.WildcardMatcher;
@@ -107,7 +108,7 @@ public class DlsFlsValveImpl implements DlsFlsRequestValve {
     private final FieldMasking.Config fieldMaskingConfig;
     private final Settings settings;
     private final AdminDNs adminDNs;
-    private boolean isResourceSharingFeatureEnabled = false;
+    private final OpensearchDynamicSetting<Boolean> resourceSharingEnabledSetting;
     private final WildcardMatcher resourceIndicesMatcher;
 
     public DlsFlsValveImpl(
@@ -119,7 +120,8 @@ public class DlsFlsValveImpl implements DlsFlsRequestValve {
         ThreadPool threadPool,
         DlsFlsBaseContext dlsFlsBaseContext,
         AdminDNs adminDNs,
-        Set<String> resourceIndices
+        Set<String> resourceIndices,
+        OpensearchDynamicSetting<Boolean> resourceSharingEnabledSetting
     ) {
         super();
         this.nodeClient = nodeClient;
@@ -141,10 +143,7 @@ public class DlsFlsValveImpl implements DlsFlsRequestValve {
                 config.updateClusterStateMetadataAsync(clusterService, threadPool);
             }
         });
-        this.isResourceSharingFeatureEnabled = settings.getAsBoolean(
-            ConfigConstants.OPENSEARCH_RESOURCE_SHARING_ENABLED,
-            ConfigConstants.OPENSEARCH_RESOURCE_SHARING_ENABLED_DEFAULT
-        );
+        this.resourceSharingEnabledSetting = resourceSharingEnabledSetting;
     }
 
     /**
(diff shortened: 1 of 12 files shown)


src/main/java/org/opensearch/security/resources/ResourceIndexListener.java
```diff
@@ -37,18 +39,44 @@ public class ResourceIndexListener implements IndexingOperationListener {
...
     public void postIndex(ShardId shardId, Engine.Index index, Engine.IndexResult result) {
+
+        if (!resourceSharingEnabledSetting.getDynamicSettingValue()) {
+            // feature is disabled
+            return;
+        }
         String resourceIndex = shardId.getIndexName();
+
+        List<String> protectedResourceTypes = protectedResourceTypesSetting.getDynamicSettingValue();
+        if (!protectedResourceTypes.contains(resourcePluginInfo.typeByIndex(resourceIndex))) {
```
cwperks, line 75: Can we check `resourcePluginInfo.getResourceIndices()` here? I'll be refactoring to remove `typeByIndex` to make it so that resource type <-> index is a many to one relationship.

DarshitChanpura: We only get index name in the postIndex() listener. And the protected_types setting accepts types only. Hence we need typeByIndex(resourceIndex) here to look up type based on the index name.

we can handle this change separately in many<->one PR. But for this PR we need this.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/feature/FeatureFlagSettingTests.java
```diff
@@ -0,0 +1,437 @@
...
+    private void awaitResourceSharingEnabled(boolean expected) {
+        // Wait for cluster setting to reflect desired value
+        await().atMost(30, TimeUnit.SECONDS).pollInterval(200, TimeUnit.MILLISECONDS).until(() -> readSettingEquals(expected));
+    }
+
+    private boolean readSettingEquals(boolean expected) {
+        try (var client = cluster.getRestClient(cluster.getAdminCertificate())) {
+            var r = client.get("_cluster/settings?include_defaults=true&flat_settings=true");
+            r.assertStatusCode(200);
+            String key = "\"" + ConfigConstants.OPENSEARCH_RESOURCE_SHARING_ENABLED + "\":\"" + expected + "\"";
```
cwperks, line 294: We should consider make a general setting matcher where we pass it the expected key and value to re-use in tests.

DarshitChanpura: agreed however this is the only place this setting is used so left it as is

src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
```diff
@@ -164,26 +167,31 @@ public boolean invoke(PrivilegesEvaluationContext context, final ActionListener<
...
-                return DlsFilterLevelActionHandler.handle(
-                    context,
-                    sharedResourceMap,
-                    listener,
-                    nodeClient,
-                    clusterService,
-                    OpenSearchSecurityPlugin.GuiceHolder.getIndicesService(),
-                    resolver,
-                    threadContext
+                Set<String> protectedIndices = resourcePluginInfo.getResourceIndicesForProtectedTypes(
```
cwperks: Can we pass the `resourceSharingProtectedTypesSetting` to `resourcePluginInfo` and encapsulate this setting fully within there? It would be convenient to just be able to get the list of indices with sharable resources with a single call.

cwperks: Marking as resolved as the latest commit addresses this.

src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
```diff
@@ -119,7 +122,9 @@ public DlsFlsValveImpl(
         ThreadPool threadPool,
         DlsFlsBaseContext dlsFlsBaseContext,
         AdminDNs adminDNs,
-        Set<String> resourceIndices
+        ResourcePluginInfo resourcePluginInfo,
+        OpensearchDynamicSetting<Boolean> resourceSharingEnabledSetting,
+        OpensearchDynamicSetting<List<String>> resourceSharingProtectedTypesSetting
```
cwperks: This isn't used in this class any longer and can be removed.

src/main/java/org/opensearch/security/resources/ResourceIndexListener.java
cwperks: FYI We can entirely get this with `resourcePluginInfo.getResourceIndicesForProtectedTypes().contains(resourceIndex)` now.

src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
DarshitChanpura: done.

Issue #5643, opened 2025-09-17
[FEATURE] Add support for X509v3 extensions for authentication


asifabashar:
**Is your feature request related to a problem?**
Currently there is no option to use X509v3 extensions fields for authentication which allows for enhanced certificate authentication

**What solution would you like?**
We need functionality to be added to security plugin in OpenSearch , to be able to read  "X509v3 Subject Alternative Name: " field from certificate content.
This will  use  username_attribute field to configure this attribute in config.yaml in opensearch security config. This value in certificate will be  compared with role mapped username in Opensearch. This is a request to add "X509v3 Subject Alternative Name: " field as additional possible  username_attribute  field along with existing possible attributes such as CN and DN.  

```
X509v3 extensions:
            X509v3 Basic Constraints: critical
                CA:FALSE
            X509v3 Authority Key Identifier: 
                keyid:<generated value >

            Authority Information Access: 
                CA Issuers - URI:<generated value>
                OCSP - URI:<generated value>

            X509v3 Subject Alternative Name: 
                <generated value: this is the field , that can be used for authentication>
            X509v3 Extended Key Usage: 
                TLS Web Client Authentication
            X509v3 CRL Distribution Points: 

                Full Name:
                  URI:<generated value>

            X509v3 Subject Key Identifier: 
                 <generated value: this is another field , that can be used for authentication>
            X509v3 Key Usage: critical
                Digital Signature, Key Encipherment

```
Below config block is from config.yaml where, username attribute need to be set to "X509v3 Subject Alternative Name" , then code need to be enhanced in security plugin to use this attribute to compare user along with exiting possible values such as  dn or cn.
```
    clientcert_auth_domain:
        authentication_backend:
          type: "noop"
        description: "Authenticate via SSL client certificates"
        http_authenticator:
          challenge: false
          config:
            username_attribute: null
          type: "clientcert"
        http_enabled: true
        order: 2
        transport_enabled: false


```
After change, config will also have capability to configure as below username_attribute option.

```
    clientcert_auth_domain:
        authentication_backend:
          type: "noop"
        description: "Authenticate via SSL client certificates"
        http_authenticator:
          challenge: false
          config:
            username_attribute: "X509v3 Subject Alternative Name"
          type: "clientcert"
        http_enabled: true
        order: 2
        transport_enabled: false


```

the code should ensure backward compatibility of configuring cn or dn attribute in config.yaml remains intact.

**What alternatives have you considered?**
The enhanced functionality is not available.

[...]

PR #5701, opened 2025-10-12
[FEATURE] Add support for X509v3 extensions for authentication

laminelam:
Resolves [#5643](https://github.com/opensearch-project/security/issues/5643) & [#4209](https://github.com/opensearch-project/security/issues/4209)

  

### Testing

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Adding Alerting V2 roles to roles.yml ([#5747](https://github.com/opensearch-project/security/pull/5747))
 - add suggest api to ad read access role ([#5754](https://github.com/opensearch-project/security/pull/5754))
 - Get list of headersToCopy from core and use getHeader(String headerName) instead of getHeaders() ([#5769](https://github.com/opensearch-project/security/pull/5769))
+- Add support for X509 v3 extensions (SAN) for authentication  ([#5701](https://github.com/opensearch-project/security/pull/5701))
 
 ### Bug Fixes
 - Create a WildcardMatcher.NONE when creating a WildcardMatcher with an empty string ([#5694](https://github.com/opensearch-project/security/pull/5694))


src/main/java/org/opensearch/security/http/HTTPClientCertAuthenticator.java
```diff
@@ -54,66 +63,126 @@ public class HTTPClientCertAuthenticator implements HTTPAuthenticator {
...
+            return null;
+        }
+
+        try {
+            final String username = extractUsername(threadContext, principal);
+            final String[] roles = extractRoles(threadContext, principal);
+            return new AuthCredentials(username, roles).markComplete();
+        } catch (InvalidNameException e) {
+            log.error("Client cert had no properly formed DN");
+            log.debug("Client cert had no properly formed DN (was: {})", principal);
```
cwperks: nit: this would be repetitive with logs at debug level. Considering only logging the DN with this log line. i.e. `log.debug("DN (was: {})", principal);`

mishail:
@reta, maybe you could give a second review, thanks!

src/main/java/org/opensearch/security/http/HTTPClientCertAuthenticator.java
```diff
@@ -54,66 +63,126 @@ public class HTTPClientCertAuthenticator implements HTTPAuthenticator {
...
-
-                if (usernameAttribute != null && usernameAttribute.length() > 0) {
-                    final List<String> usernames = getDnAttribute(rfc2253dn, usernameAttribute);
-                    if (usernames.isEmpty() == false) {
-                        username = usernames.get(0);
-                    }
+            Pattern pattern = null;
+            if (!Strings.isNullOrEmpty(regex)) {
+                try {
+                    pattern = Pattern.compile(regex, Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE);
```
reta: I believe (at least per RFC [1]) that the pattern dialect being used is not compatible with Java pattern spec: like `*` is equivalent of Java's `.*` , or I am missing something here? thank you


[1] https://www.rfc-editor.org/rfc/rfc6125#section-6.4.3

laminelam: Right now I treat the value as a Java regex and it should clearly be stated in the documentation.
As for the RFC, the wildcard spec is only for the case of SAN:DNS hostnames (e.g., *.example.com glob). We can still support both regex and glob but I'am afraid this would make things more complex and creates confusion for the user. I see three options:

A) Keep regex only and document clearly.

B) For DNS, support RFC-6125 glob by default with an explicit regex: opt-in.

C) Allow both for all SAN types via small prefixes (?glob) / (?regex) with auto-detect for simple globs.

What are your thoughts?

reta: >  We can still support bother regex and glob but I'am afraid this would make things more complex and creates confusion for the user. I see three options:

Certainly don't intent to complicate things, basically my question would be - when if ever the pattern will be (and should be) Java compatible? (yes, we could document etc, but from practical standpoint). This looks to me as restriction that is artificial (due to the fact we use Java standard library), may be we could completely disregards the pattern handing?

laminelam: The reason of using java regex is that we need to support groups capturing. Why? Because we want to be able to extract specific information to use it as a _username_ or back-end _role_. For example, let's say the cert passes such info in the following format:
`SAN:URN:googleservice/groupid/123/dept `
We want our regex to not only much this format but also extract the group id =123 and use it in roles.

To be honest I was very hesitant to add regex support, the first version of this PR didn't have it. So I am willing to remove it for now and restrict the validation to exact much, and may in a futur PR extend it to support regex.

reta: >To be honest I was very hesitant to add regex support, the first version of this PR didn't have it. So I am willing to remove it for now and restrict the validation to exact much, and may in a futur PR extend it to support regex.

Certainly +1 to that, thank you @laminelam

src/main/java/org/opensearch/security/http/HTTPClientCertAuthenticator.java
```diff
@@ -129,4 +192,152 @@ private List<String> getDnAttribute(LdapName rfc2253dn, String attribute) {
...
+                .limit(MAX_SAN_MATCHES)
+                .collect(Collectors.toList());
+        } catch (CertificateParsingException e) {
+            log.error("Error parsing X509 certificate", e);
+            return Collections.emptyList();
+        }
+    }
+
+    // Minimal case-insensitive '*' glob
+    private static boolean matchesGlobCI(String value, String pattern) {
```
cwperks: Let's use the existing [WildcardMatcher](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/WildcardMatcher.java) class for this.

cwperks (dismissed):
Thank you for the latest update @laminelam ! Can we please refactor the latest commit to use WildcardMatcher?

src/main/java/org/opensearch/security/http/HTTPClientCertAuthenticator.java
laminelam: Thanks for the suggestion @cwperks , didn't know about WildcardMatcher.
Done

cwperks: Glob matching is done many ways in OpenSearch. Core also has a class called [Regex](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/regex/Regex.java) (misnomer) for glob matching to. I would say let's use that, but its marked for core internal use.

cwperks (approved):

src/main/java/org/opensearch/security/http/HTTPClientCertAuthenticator.java
```diff
@@ -54,66 +63,130 @@ public class HTTPClientCertAuthenticator implements HTTPAuthenticator {
...
-                final LdapName rfc2253dn = new LdapName(principal);
-                String username = principal.trim();
-                String[] backendRoles = null;
-
-                if (usernameAttribute != null && usernameAttribute.length() > 0) {
-                    final List<String> usernames = getDnAttribute(rfc2253dn, usernameAttribute);
-                    if (usernames.isEmpty() == false) {
-                        username = usernames.get(0);
-                    }
+            WildcardMatcher matcher = null; // null means pass-through
```
cwperks: nit: if you'd like to remove the special null handling, we can instantiate this as WildcardMatcher.ANY.

laminelam: done.
Thanks for the review @cwperks

cwperks (approved):

reta (approved):
Thank you @laminelam !

PR #5749, opened 2025-10-28
Add security provider earlier in bootstrap process

terryquigleysas:
### Description
* Category 
Bug fix, Refactoring
To aid the use of BCFKS using more JDKs the loading of the BCFIPS provider is moved earlier in the bootstrap process to ensure it is present before loading SSL settings
Using BCFKS as the default keystore can fail 

### Related Issues
https://github.com/opensearch-project/security/issues/3420
https://github.com/opensearch-project/documentation-website/pull/11412

### Testing
Ran the Bulk Integration Test action - all tests passing.
Tested custom build of Security plugin with the change on local deployment.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -432,8 +432,6 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
         demoCertHashes.add("ba9c5a61065f7f6115188128ffbdaa18fca34562b78b811f082439e2bef1d282"); // esnode-key
         demoCertHashes.add("bcd708e8dc707ae065f7ad8582979764b497f062e273d478054ab2f49c5469c6"); // root-ca
 
-        tryAddSecurityProvider();
-
         final String advancedModulesEnabledKey = ConfigConstants.SECURITY_ADVANCED_MODULES_ENABLED;
         if (settings.hasValue(advancedModulesEnabledKey)) {
             deprecationLogger.deprecate("Setting {} is ignored.", advancedModulesEnabledKey);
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -2442,17 +2440,6 @@ public Optional<SecureSettingsFactory> getSecureSettingFactory(Settings settings
         );
     }
 
-    @SuppressWarnings("removal")
-    private void tryAddSecurityProvider() {
```
cwperks, line 2446: Are we intentionally keeping the second call to `tryAddSecurityProvider` in this class? Shouldn't we only call that once?

cwperks: Otherwise, this PR LGTM.

terryquigleysas: Yes, I was being ultra-cautious leaving that in, in case there was some code path I hadn't considered, and that the additional call wasn't too onerous. The second call checks if it has anything to do and otherwise continues without any further action. I can remove it if you prefer?

cwperks: Where you have it now will always be called because the constructor to OpenSearchSecurityPlugin calls super on the first line: https://github.com/terryquigleysas/security/blob/7d21db99b4cf0a8432c4870660feb1d65f5349cf/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L360

The only time it would not be called is if security is installed, but intentionally disabled (`plugins.security.disabled: true`). Would we want the logic to run in that case?

terryquigleysas: Yes, on my deployment I know it is called and takes no action the second time. I will remove it as you confirm there should be no other path. I have a meeting soon but I will do it after that.

terryquigleysas:
@cwperks Is it possible to have this backported to the next 3.3.x release?

peterzhuamazon:
Need manual backport and this will be the last PR to catch 3.3.2 release train as we need time to integTests.

PR #5753, opened 2025-10-28
New integration tests for Dashboards multi-tenancy

nibix:
### Description

This introduces integration tests for Dashboards multi-tenancy based on the concept that is also used by the other newly authorization integration tests. 

This is in preparation for the "Improved index resolution" project #5399 and allows us to test the different privilege evaluation implementations and compare their behavior.

* Category: Maintenance

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/privileges/int_tests/DashboardMultiTenancyIntTests.java
+++ b/src/integrationTest/java/org/opensearch/security/privileges/int_tests/DashboardMultiTenancyIntTests.java
@@ -40,6 +40,8 @@ import static org.opensearch.test.framework.matcher.RestIndexMatchers.OnUserInde
 import static org.opensearch.test.framework.matcher.RestMatchers.isCreated;
 import static org.opensearch.test.framework.matcher.RestMatchers.isForbidden;
 import static org.opensearch.test.framework.matcher.RestMatchers.isOk;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 /**
  * An integration test matrix for Dashboards multi-tenancy. Verifies both read and write operations
@@ -58,6 +60,10 @@ public class DashboardMultiTenancyIntTests {
 
     static final TestSecurityConfig.Tenant TENANT_BUSINESS_INTELLIGENCE = new TestSecurityConfig.Tenant("business_intelligence")
         .description("Business Intelligence Department Tenant");
+
+    static final TestSecurityConfig.Tenant TENANT_FINANCE = new TestSecurityConfig.Tenant("finance").description(
+        "Finance Department Tenant"
+    );
 
     // -------------------------------------------------------------------------------------------------------
     // Test indices and aliases
@@ -201,6 +207,8 @@ public class DashboardMultiTenancyIntTests {
             new TestSecurityConfig.Role("bi_analyst_role").clusterPermissions("cluster_composite_ops")
                 .tenantPermissions("kibana_all_write")
                 .on("business_intelligence")
+                .tenantPermissions("kibana_all_write")
+                .on("finance")
         )
         .reference(
             READ,
@@ -330,7 +338,7 @@ public class DashboardMultiTenancyIntTests {
         return new LocalCluster.Builder().clusterManager(ClusterManager.THREE_CLUSTER_MANAGERS)
             .authc(AUTHC_HTTPBASIC_INTERNAL)
             .users(USERS)
-            .tenants(TENANT_HUMAN_RESOURCES, TENANT_BUSINESS_INTELLIGENCE)
+            .tenants(TENANT_HUMAN_RESOURCES, TENANT_BUSINESS_INTELLIGENCE, TENANT_FINANCE)
             .indices(
                 dashboards_index_global,
                 dashboards_index_human_resources,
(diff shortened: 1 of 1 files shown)


src/integrationTest/java/org/opensearch/security/privileges/int_tests/DashboardMultiTenancyIntTests.java
```diff
@@ -0,0 +1,794 @@
...
+                """.formatted(testDocId1, testDocId2);
+
+            TestRestClient.HttpResponse response = restClient.postJson(
+                "_bulk?pretty",
+                bulkBody,
+                new BasicHeader("securitytenant", "human_resources")
+            );
+
+            if (user.reference(WRITE).covers(dashboards_index_human_resources)) {
+                if (user == GLOBAL_TENANT_READ_WRITE_USER || user == WILDCARD_TENANT_USER) {
```
DarshitChanpura, line 663: why use == comparison instead of .equals? I've seen this pattern used in your other PRs and so curious to know.

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DashboardMultiTenancyIntTests.java
```diff
@@ -0,0 +1,794 @@
...
+        this.cluster = clusterConfig.cluster(DashboardMultiTenancyIntTests::clusterBuilder);
+        this.clusterConfig = clusterConfig;
+    }
+
+    private void delete(String... paths) {
+        try (TestRestClient adminRestClient = cluster.getAdminCertRestClient()) {
+            for (String path : paths) {
+                TestRestClient.HttpResponse response = adminRestClient.delete(path);
+                if (response.getStatusCode() != 200 && response.getStatusCode() != 404) {
+                    throw new RuntimeException("Error while deleting " + path + "\n" + response.getBody());
```
DarshitChanpura, line 789: should failure during cleanup fail entire test even though the test passed?

DarshitChanpura (commented):
thanks @nibix . Changes lgtm. Left a couple of clarifying comments.

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DashboardMultiTenancyIntTests.java
cwperks: The parameter factory is passing the same memory reference of the user: https://github.com/opensearch-project/security/pull/5753/files#diff-fdb669ef5d02fe976ad660adc622d504a1f0aa2e4239bd5f64fa09d2fc052132R763-R772

Since its the same object (in memory), == is a valid way to compare. If a separate User object was instantiated with exactly the same attributes then a deep comparison would be needed.

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DashboardMultiTenancyIntTests.java
```diff
@@ -0,0 +1,794 @@
...
+                    assertThat(
+                        response,
+                        containsExactly(dashboards_index_human_resources).at("items[*].update[?(@.result == 'updated')]._index")
+                            .reducedBy(user.reference(WRITE))
+                            .whenEmpty(isOk())
+                    );
+                } else {
+                    // At the moment, we get here a nested error:
+                    // Update is not supported when FLS or DLS or Fieldmasking is activated
+                    // This is a bug; even though it does not seem to have an impact on Dashboards functionality
```
cwperks, line 674: I've seen this come up on a separate repo here: https://github.com/opensearch-project/observability/issues/1951

Could this be related?

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DashboardMultiTenancyIntTests.java
DarshitChanpura: i understand. I wanted to know the reason but using reference comparison for objects where we usually use it for primitives.

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DashboardMultiTenancyIntTests.java
DarshitChanpura: I leveraged LLM to summarize what could cause this error even though nothing signification has changed in DlsFlsValveImpl.java between 2.19 (where error is not seen) and main, and here is what it suggests:
<details>

```
Conditions that must be true for the exception to fire

In invoke(...), we hit that UpdateRequest block only if all of these are true:

- The request is an UpdateRequest (or a BulkShardRequest containing an UpdateRequest).
That’s the direct instanceof check.

- The request did not get short-circuited earlier:

Not a cluster-level permission (or it is MultiGetAction, which is explicitly excluded from the cluster-perm fast-path):

if (isClusterPerm(context.getAction()) && !MultiGetAction.NAME.equals(context.getAction())) {
    return true;
}


- The user is not an admin DN:

if (userSubject != null && adminDNs.isAdmin(userSubject.getUser())) {
    return true;
}


- It is not considered an internal/plugin request (or if it is, we return early):

if (HeaderHelper.isInternalOrPluginRequest(threadContext)) {
    // ... possible resource-sharing handling ...
    return true;
}


- The “filter level DLS already done” header is not set:

if (threadContext.getHeader(ConfigConstants.OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE) != null) {
    // log ..
    return true;
}


- At least one of DLS/FLS/field masking is active for this request:

boolean hasDlsRestrictions = !config.getDocumentPrivileges().isUnrestricted(context, resolved);
boolean hasFlsRestrictions = !config.getFieldPrivileges().isUnrestricted(context, resolved);
boolean hasFieldMasking = !config.getFieldMasking().isUnrestricted(context, resolved);

if (!hasDlsRestrictions && !hasFlsRestrictions && !hasFieldMasking) {
    return true; // no restrictions -> no exception later
}


If any of those early returns happens, the UpdateRequest check is never hit, and you don’t see the error.
```
</details>

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DashboardMultiTenancyIntTests.java
nibix: IMHO, yes, as a non-well cleaned up cluster has the potential to cause unexpected effects in consecutive tests. Thus, a fail-fast mechanism is better than having to reason about test failures due to wrong preconditions.

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DashboardMultiTenancyIntTests.java
nibix: Hmm, if this is about an update on the .kibana index, then this might be actually the case. This is likely caused by the deny-by-default behavior of DLS/FLS introduced in OpenSearch 3. We haven't noticed it so far, as it seems that Dashboards only rarely uses an update operation. index/delete operations however pass through, as the DlsFlsValve has this special behavior of only blocking updates.

src/integrationTest/java/org/opensearch/security/privileges/int_tests/DashboardMultiTenancyIntTests.java
nibix: You'd also use `==` for enum objects. Even though `TestUser` is not an enum, it acts in this context quite similarly to an enum, especially in the way that all possible values are known upfront and available as static fields.

PR #5789, opened 2025-11-13
[Resource Sharing] Requires `default_owner` for `resource/migrate` API

DarshitChanpura:
### Description
For plugins which do not store user info in their resources, migrate API would fail to migrate those resources. Instead of failing, we now require API callers to pass a `defauil_owner` param which would act as the owner of the resources without any owner.

### Testing
Manual + Automated

### Check List

### DIFF
--- a/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/feature/disabled/DirectIndexAccessTests.java
+++ b/sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/feature/disabled/DirectIndexAccessTests.java
@@ -31,6 +31,7 @@ import static org.opensearch.sample.resource.TestUtils.NO_ACCESS_USER;
 import static org.opensearch.sample.resource.TestUtils.SAMPLE_RESOURCE_SEARCH_ENDPOINT;
 import static org.opensearch.sample.resource.TestUtils.newCluster;
 import static org.opensearch.sample.utils.Constants.RESOURCE_INDEX_NAME;
+import static org.opensearch.sample.utils.Constants.RESOURCE_TYPE;
 import static org.opensearch.security.api.AbstractApiIntegrationTest.forbidden;
 import static org.opensearch.test.framework.TestSecurityConfig.User.USER_ADMIN;
 
@@ -66,7 +67,7 @@ public class DirectIndexAccessTests {
 
             // cannot access any raw request
             try (TestRestClient client = cluster.getRestClient(NO_ACCESS_USER)) {
-                String sample = "{\"name\":\"sampleUser\"}";
+                String sample = "{\"name\":\"sampleUser\",\"resource_type\":\"" + RESOURCE_TYPE + "\"}";
                 HttpResponse resp = client.postJson(RESOURCE_INDEX_NAME + "/_doc", sample);
                 resp.assertStatusCode(HttpStatus.SC_FORBIDDEN);
             }
@@ -89,7 +90,7 @@ public class DirectIndexAccessTests {
 
             // cannot create a resource since user doesn't have indices:data/write/index permission
             try (TestRestClient client = cluster.getRestClient(LIMITED_ACCESS_USER)) {
-                String sample = "{\"name\":\"sampleUser\"}";
+                String sample = "{\"name\":\"sampleUser\",\"resource_type\":\"" + RESOURCE_TYPE + "\"}";
                 HttpResponse resp = client.postJson(RESOURCE_INDEX_NAME + "/_doc", sample);
                 resp.assertStatusCode(HttpStatus.SC_FORBIDDEN);
             }
@@ -114,7 +115,7 @@ public class DirectIndexAccessTests {
 
             // cannot create a resource directly since system index protection (SIP) is enabled
             try (TestRestClient client = cluster.getRestClient(FULL_ACCESS_USER)) {
-                String sample = "{\"name\":\"sampleUser\"}";
+                String sample = "{\"name\":\"sampleUser\",\"resource_type\":\"" + RESOURCE_TYPE + "\"}";
                 HttpResponse resp = client.postJson(RESOURCE_INDEX_NAME + "/_doc", sample);
                 resp.assertStatusCode(HttpStatus.SC_FORBIDDEN);
             }
@@ -179,7 +180,7 @@ public class DirectIndexAccessTests {
 
             // cannot access any raw request
             try (TestRestClient client = cluster.getRestClient(NO_ACCESS_USER)) {
-                String sample = "{\"name\":\"sampleUser\"}";
+                String sample = "{\"name\":\"sampleUser\",\"resource_type\":\"" + RESOURCE_TYPE + "\"}";
                 TestRestClient.HttpResponse resp = client.postJson(RESOURCE_INDEX_NAME + "/_doc", sample);
                 resp.assertStatusCode(HttpStatus.SC_FORBIDDEN);
             }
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -159,8 +159,6 @@ public void hasPermission(
             listener.onResponse(true);
             return;
         }
-        Set<String> userRoles = new HashSet<>(user.getSecurityRoles());
```
cwperks, line 158: Thanks for removing these.

src/main/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -310,7 +310,7 @@ private ValidationResult<MigrationStats> createNewSharingRecords(Triple<String,
                         "Successfully migrated a resource sharing entry {} for resource {} within index {}",
                         entry,
                         resourceId,
-                        sourceInfo.getLeft()
```
cwperks, line 309: nice. Much more readable this way.

src/main/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -245,22 +243,23 @@ private ValidationResult<Triple<String, Map<String, String>, List<SourceDoc>>> l
...
      *      1. Parses existing sharing info to a new ResourceSharing records
      *      2. Indexes the new record into corresponding resource-sharing index
      */
-    private ValidationResult<MigrationStats> createNewSharingRecords(Triple<String, Map<String, String>, List<SourceDoc>> sourceInfo)
-        throws IOException {
+    private ValidationResult<MigrationStats> createNewSharingRecords(ValidationResultArg sourceInfo) throws IOException {
         AtomicInteger migratedCount = new AtomicInteger();
-        AtomicReference<Set<String>> skippedNoUser = new AtomicReference<>();
-        skippedNoUser.set(new HashSet<>());
+        AtomicReference<Set<String>> skippedNoType = new AtomicReference<>();
```
cwperks: nit: consider encapsulating this within MigrationStats and call a fn with a name like `addNoType(String resourceId)`. We can also encapsulate the message that's created.

src/main/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -410,20 +417,42 @@ public SourceDoc(String id, String username, List<String> backendRoles, String t
         }
     }
 
+    static class ValidationResultArg {
```
cwperks: nit: can be turned into a record which would automatically give this getters.

src/main/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -137,17 +137,15 @@ private void handleMigrate(RestChannel channel, RestRequest request, Client clie
      *      1. Pulls "source_index" from request body
      *      2. Does a match_all search (up to 10k) in the "source_index"
      *      3. Create a SourceDoc for each raw doc
-     *      4. Returns a triple of the source index name, the default access level and the list of source docs.
+     *      4. Returns an object of the source index name, the default owner name, the default access level and the list of source docs.
```
cwperks: default access level may be a bit misleading in this comment. It would be the mapping of type -> default access level

cwperks:
One high-level comment. What happens if the API is called twice in a row with a different param for `default_owner`? Should the subsequent request be a no-op or should it overwrite existing resource sharing entries?

IMO I think the behavior should be the latter in the event that the user calls the API with a typo initially.

src/main/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiAction.java
DarshitChanpura: should be fine as the scope if limited to this class. not worried about encapsulation as no other class will be able to reference MigrationStats directly.

DarshitChanpura: i've refactored this for safety.

DarshitChanpura:
>One high-level comment. What happens if the API is called twice in a row with a different param for default_owner? Should the subsequent request be a no-op or should it overwrite existing resource sharing entries?

>IMO I think the behavior should be the latter in the event that the user calls the API with a typo initially.

At present, it doesn't over-write existing record. We should not allow over-writing of records, instead we can add a security-admin style management that allows transfer of ownership.

cwperks:
> At present, it doesn't over-write existing record. We should not allow over-writing of records, instead we can add a security-admin style management that allows transfer of ownership.

Agreed, and maybe 2 different flavors here:

1. Admin level that allows overwrite of a sharing record
2. One similar to Change Password API where a user can transfer ownership of their own resources

src/integrationTest/java/org/opensearch/security/api/AbstractApiIntegrationTest.java
```diff
@@ -205,10 +206,12 @@ protected static String[] allRestAdminPermissions() {
         var counter = 0;
         for (final var e : ENDPOINTS_WITH_PERMISSIONS.entrySet()) {
             if (e.getKey() == Endpoint.SSL) {
-                permissions[counter] = e.getValue().build(CERTS_INFO_ACTION);
```
cwperks, line 208: This block has always bugged me bc of the brittle logic. We should find a way to make this more robust if a new security endpoint is added to restapi: permissions.

src/main/java/org/opensearch/security/dlic/rest/api/RestApiAdminPrivilegesEvaluator.java
```diff
@@ -74,6 +76,7 @@ default String build() {
         .put(Endpoint.VIEW_VERSION, action -> buildEndpointPermission(Endpoint.VIEW_VERSION))
         .put(Endpoint.ROLLBACK_VERSION, action -> buildEndpointPermission(Endpoint.ROLLBACK_VERSION))
         .put(Endpoint.SSL, action -> buildEndpointActionPermission(Endpoint.SSL, action))
+        .put(Endpoint.RESOURCE_SHARING, action -> buildEndpointActionPermission(Endpoint.RESOURCE_SHARING, action))
```
cwperks, line 79: If you add this above line 78 then the changes in AbstractApiIntegrationTest are not necessary.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
```diff
@@ -283,7 +288,11 @@ public void indexResourceSharing(String resourceIndex, ResourceSharing sharingIn
                 if (ExceptionsHelper.unwrapCause(e) instanceof VersionConflictEngineException) {
                     // already exists → skipping
                     LOGGER.debug("Entry for [{}] already exists in [{}], skipping", resourceId, resourceSharingIndex);
-                    listener.onResponse(sharingInfo);
+                    if (isMigrationCall) {
```
cwperks: Why distinguish here?

src/main/java/org/opensearch/security/dlic/rest/api/RestApiAdminPrivilegesEvaluator.java
DarshitChanpura: correct. But let;s leave it as it as the above logic is brittle.

src/integrationTest/java/org/opensearch/security/api/AbstractApiIntegrationTest.java
DarshitChanpura: i agree.

src/main/java/org/opensearch/security/resources/ResourceSharingIndexHandler.java
DarshitChanpura: Just for stats purposes. We return the original object on non-migration calls (mostly for logging).

DarshitChanpura: I've removed it and default returned to null

cwperks (approved):

PR #5791, opened 2025-11-16
Modularized PrivilegesEvaluator

nibix:
### Description

Currently the class `PrivilegesEvaluator` has many more tasks than evaluating privileges: It also encapsulates configuration for role mapping, multi tenancy, it manages thread context user info data, etc.

Goal of this PR is to cut the class `PrivilegesEvaluator` into several modules with clearly dedicated scopes.  Additionally, this PR introduces an interface for `PrivilegesEvaluator` which has the final goal to make privileges evaluator implementations pluggable (see #5399).

The main modules introduced by this PR are:

- `PrivilegesConfiguration`:  This now acts as an umbrella over all privileges-related configuration classes. It is centrally instantiated in the `OpenSearchSecurityPlugin` and exists only once per node runtime.
- Interface `PrivilegesEvaluator` with implementation `PrivilegesEvaluatorImpl` and `PrivilegesEvaluator.NotInitialized`: This is reduced to the core privilege evaluation. `PrivilegesEvaluatorImpl`  now contains the logic formerly located in  `PrivilegesEvaluator`. Instances of `PrivilegesEvaluator` are managed by `PrivilegesConfiguration`, which can re-instantiate `PrivilegeEvaluator` on demand. Thus, `PrivilegesEvaluatorImpl`  has no longer an unitialized state in case config has not been loaded (yet); the unitialized state is now represented by the class `PrivilegesEvaluator.NotInitialized`.
- Interface `RoleMapper` and classes `ConfigurableRoleMapper` and `InjectedRoleMapper`. This encapsulates role mapping logic which was formerly sprayed around in `PrivilegesEvaluator`, `SecurityFilter` and `ConfigModel`. Instances of `RoleMapper` are managed by `OpenSearchSecurityPlugin`. They are not managed by  `PrivilegesConfiguration` because role mapping is not really related to privileges.
- `ThreadContextUserInfo`: Encapsulates logic related to writing a user info string into the thread context.
- `DashboardsMultiTenancyConfiguration`: A lightweight container for configuration for OpenSearch Dashboards multi tenancy.

* Category: Refactoring
* Why these changes are required? Making PrivilegesEvaluator pluggable as preparation for #5399.

### Testing

- Existing int tests

### Check List

### DIFF
--- a/src/test/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluatorTest.java
+++ b/src/test/java/org/opensearch/security/privileges/RestLayerPrivilegesEvaluatorTest.java
@@ -89,7 +89,7 @@ public class RestLayerPrivilegesEvaluatorTest {
             OpenSearchSecurityException.class,
             () -> restPrivilegesEvaluator.evaluate(TEST_USER, "route_name", null)
         );
-        assertThat(exception.getMessage(), equalTo("OpenSearch Security is not initialized: PrivilegesEvaluator not initialized"));
+        assertThat(exception.getMessage(), equalTo("OpenSearch Security not initialized: PrivilegesEvaluator not initialized"));
     }
 
     @Test


src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -379,25 +390,15 @@ private <Request extends ActionRequest, Response extends ActionResponse> void ap
                     }
             }
 
-            final PrivilegesEvaluator eval = evalp;
-
-            if (!eval.isInitialized()) {
-                StringBuilder error = new StringBuilder("OpenSearch Security not initialized for ");
```
nibix, line 385: Note: These checks and handling are now in privilegesEvaluator.createContext()

See: https://github.com/opensearch-project/security/blob/65375fd992c7711efee8abf2c468b90cac719ea0/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L147-L158

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluatorImpl.java
```diff
@@ -0,0 +1,758 @@
...
+import org.opensearch.security.user.User;
+import org.opensearch.tasks.Task;
+import org.opensearch.threadpool.ThreadPool;
+
+import static org.opensearch.security.OpenSearchSecurityPlugin.traceAction;
+
+/**
+ * The current implementation for action privilege evaluation.
+ * <p>
+ * Moved from https://github.com/opensearch-project/security/blob/062ea716d10240cc50d01735f457523a61393a59/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```
nibix, line 108: Note: The core logic is unchanged from the old PrivilegesEvaluator class (even though it shows up as an addition here).

src/main/java/org/opensearch/security/user/ThreadContextUserInfo.java
```diff
@@ -0,0 +1,101 @@
...
+            joiner.add(escapePipe(context.getUser().getName()));
+            joiner.add(escapePipe(String.join(",", context.getUser().getRoles())));
+            joiner.add(escapePipe(String.join(",", context.getMappedRoles())));
+
+            String requestedTenant = context.getUser().getRequestedTenant();
+            joiner.add(requestedTenant);
+
+            String tenantAccessToCheck = getTenancyAccess(context);
+            joiner.add(tenantAccessToCheck);
+
```
shikharj05, line 88: nit: missing logger from original method- https://github.com/nibix/security/blob/main/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java#L324

The log line has been helpful to debug configuration/operational issues.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluatorImpl.java
```diff
@@ -0,0 +1,758 @@
...
+            task,
+            irr,
+            resolver,
+            clusterStateSupplier,
+            actionPrivileges
+        );
+    }
+
+    @Override
+    public PrivilegesEvaluatorResponse evaluate(PrivilegesEvaluationContext context) {
```
shikharj05, line 291: as a follow-up, should this be broken down into Cluster/Index/Doc/Field-level evaluate methods?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluatorImpl.java
```diff
@@ -0,0 +1,758 @@
...
+    }
+
+    private static Map<String, SubjectBasedActionPrivileges> createActionPrivileges(
+        Map<String, RoleV7> pluginIdToRolePrivileges,
+        FlattenedActionGroups staticActionGroups
+    ) {
+        Map<String, SubjectBasedActionPrivileges> result = new HashMap<>(pluginIdToRolePrivileges.size());
+
+        for (Map.Entry<String, RoleV7> entry : pluginIdToRolePrivileges.entrySet()) {
+            result.put(entry.getKey(), new SubjectBasedActionPrivileges(entry.getValue(), staticActionGroups));
```
shikharj05, line 739: does this require to be thread-safe? also, can it lead to ``ConcurrentModificaitonException``?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluatorImpl.java
nibix: PrivilegeEvaluator is only for actions, so we could only break it into cluster vs index. 

Still, I am not sure of the use-case; could you elaborate on that?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluatorImpl.java
nibix: This is only called in the constructor, so it is strictly single threaded:

https://github.com/opensearch-project/security/blob/65375fd992c7711efee8abf2c468b90cac719ea0/src/main/java/org/opensearch/security/privileges/PrivilegesEvaluatorImpl.java#L155-L202

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluatorImpl.java
shikharj05: No specific use-case - just to make the code more modular.

nibix: I think in #5399 we will be achieving already a quite good and clear structure:

https://github.com/opensearch-project/security/blob/8e209b4f25df54a35b6e2e8fb599e9d3ffbfe915/src/main/java/org/opensearch/security/privileges/actionlevel/nextgen/PrivilegesEvaluator.java#L279-L356

Yet, it does not define separate interface methods for cluster vs index actions, as the task of deciding what is what is actually a task of PrivilegeEvaluator. If we would want to have such interface methods, we would kind of have to define sub-interfaces which might complicate the whole structure a bit too much.

shikharj05: Yes, I didn't mean new interfaces - just restructuring the code. 

> I think in https://github.com/opensearch-project/security/pull/5399 we will be achieving already a quite good and clear structure:

Thanks! works for me :)

src/main/java/org/opensearch/security/user/ThreadContextUserInfo.java
nibix: I have restored the log message.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluator.java
```diff
@@ -26,911 +26,136 @@
...
-        String requestedTenant = context.getUser().getRequestedTenant();
-        final String tenant = Strings.isNullOrEmpty(requestedTenant) ? GLOBAL_TENANT : requestedTenant;
-        if (tenantPrivileges.get().hasTenantPrivilege(context, tenant, TenantPrivileges.ActionType.WRITE)) {
-            return WRITE_ACCESS;
-        } else if (tenantPrivileges.get().hasTenantPrivilege(context, tenant, TenantPrivileges.ActionType.READ)) {
-            return READ_ACCESS;
-        } else {
-            return NO_ACCESS;
-        }
+    default PrivilegesEvaluationContext createContext(User user, String action) {
```
cwperks, line 48: @nibix Is this a method that can be moved up to PrivilegesConfiguration?

nibix: That's a good question!

Looking at https://github.com/opensearch-project/security/pull/5399 the implementation is actually the same in the old and the new version. *However*, the implementations operate on differently constructed ActionPrivilege data structures which are governed  by the PrivilegeEvaluator implementation. This makes moving this method to PrivilegeConfiguration difficult, as this would make the necessary interface between the two significantly wider.

src/main/java/org/opensearch/security/user/ThreadContextUserInfo.java
```diff
@@ -0,0 +1,109 @@
...
+        ClusterSettings clusterSettings,
+        Settings settings
+    ) {
+        this.threadContext = threadContext;
+        this.userAttributeSerializationEnabled = settings.getAsBoolean(
+            USER_ATTRIBUTE_SERIALIZATION_ENABLED,
+            USER_ATTRIBUTE_SERIALIZATION_ENABLED_DEFAULT
+        );
+        this.privilegesConfiguration = privilegesConfiguration;
+        if (clusterSettings != null) {
```
cwperks, line 64: In what cases is this null?

cwperks (approved):

src/main/java/org/opensearch/security/user/ThreadContextUserInfo.java
nibix: Only in mocked test environments. Was considering mocking ClusterSettings as well, but this seemed to be a bottomless pit (IMHO, we are mocking a bit too often in the unit tests, that does not make maintenance easier in cases).

Issue #3905, opened 2024-01-02
[RFC] Refined index authorization (replacement for do_not_fail_on_forbidden)


nibix:
# Background

The `do_not_fail_on_forbidden` setting switches globally between two different modes how OpenSearch reacts to cases when users have no privileges for indices.

By default, OpenSearch will resolve any patterns specified in action requests against all indices on the cluster. If then the current user does not have access privileges for any of these indices, OpenSearch will yield a 403 error.

If `do_not_fail_on_forbidden` is enabled, the behavior is changed: The resolution is limited to indices to which the user has access. Thus, the operation will succeed. 

A very common example for the difference is a `_search` request to all indices (`GET /_search`, or `GET /*/_search`). For users which have not complete read-access to all indices on the cluster, this request will always fail with a 403 error if `do_not_fail_on_forbidden` disabled. If `do_not_fail_on_forbidden` is enabled, these users will get the search result limited to the indices they have access to.

Historically, the `do_not_fail_on_forbidden` mode was implemented to make working in Dashboards more smooth. However, this is a very minimal implementation. This leads to confusion regarding the behavior of operations which are not supported (cf https://github.com/opensearch-project/security/issues/1815, https://github.com/opensearch-project/security/issues/2257, https://github.com/opensearch-project/security/issues/2113).

Additionally, the OpenSearch documentation is quite sparse on the option. All documentation regarding the option is located in sections related to multi-tenancy (cf https://opensearch.org/docs/latest/security/multi-tenancy/multi-tenancy-config/). It just states:

> If `true`, the security plugin removes any content that a user is not allowed to see from search results. If `false`, the plugin returns a security exception. Default is `false`.

This is not really helpful because:

- The documentation does not describe use-cases for this option and does not explain scenarios in which it should be enabled and in which it should be disabled.
- Even though it is only mentioned in the multi-tenancy chapters of the documentation, the setting might be also useful for users without multi-tenancy enabled, or even for users which do not utilize Dashboards.
- There are nuances in the behavior which are left unclear. This pertains for example to behavior when using aggregations or when referring to concrete indices (like  `GET /_search/index_i_have_no_privileges_for`).

To make matters more complicated, there is also the `do_not_fail_on_forbidden_empty` setting which is undocumented. If it is `false`, `_search` operations will fail with a 403 error if the given index expression matches no indices the user has access to. If it is `true`, `_search` operations will yield an empty result set  if the given index expression matches no indices the user has access to.

[...]

PR #5797, opened 2025-11-18
Fix IllegalArgumentException when resolved indices are empty

### LINKED ISSUE #5794: [BUG] java.lang.IllegalArgumentException: Must contain at least one column and at least one row (got []/[indices:monitor/recovery])
**What is the bug?**

If the security plugin 2.19.3.0 is enabled, information about ongoing and completed shard recoveries cannot be obtained via the `/_cat/recovery` endpoint because all indices are closed (have been closed manually).

The request fails with:
`java.lang.IllegalArgumentException: Must contain at least one column and at least one row (got []/[indices:monitor/recovery])`

On 2.17 the same works fine, and setting `plugins.security.privileges_evaluation.use_legacy_impl=true` solves the issue. 

- Start 3-node cluster with a security plugin 2.19.3.0 enabled
- Close all indices
- curl `/_cat/recovery`

The resolution for such a request yelds an empty list:
```
        Set<String> resolvedIndicesSet = resolvedIndices.getAllIndicesResolved(
            context.getClusterStateSupplier(),
            context.getIndexNameExpressionResolver());

        // resolvedIndicesSet - empty list
```

The request fails with the stacktrace in logs:

```
Unexpected exception java.lang.IllegalArgumentException: Must contain at least one column and at least one row (got []/[indices:monitor/recovery])
java.lang.IllegalArgumentException: Must contain at least one column and at least one row (got []/[indices:monitor/recovery])
        at com.selectivem.collections.CheckTableImpl.create(CheckTableImpl.java:81) ~[special-collections-complete-1.4.0.jar:?]
        at com.selectivem.collections.CheckTable.create(CheckTable.java:75) ~[special-collections-complete-1.4.0.jar:?]
        at org.opensearch.security.privileges.ActionPrivileges.hasIndexPrivilege(ActionPrivileges.java:210) ~[opensearch-security-2.19.3.0.jar:2.19.3.0]
        at org.opensearch.security.privileges.PrivilegesEvaluatorImpl.evaluate(PrivilegesEvaluatorImpl.java:603) ~[opensearch-security-2.19.3.0.jar:2.19.3.0]
        at org.opensearch.security.filter.SecurityFilter.apply0(SecurityFilter.java:377) [opensearch-security-2.19.3.0.jar:2.19.3.0]
        at org.opensearch.security.filter.SecurityFilter.apply(SecurityFilter.java:166) [opensearch-security-2.19.3.0.jar:2.19.3.0]
        at org.opensearch.action.support.TransportAction$RequestFilterChain.proceed(TransportAction.java:218) [opensearch-2.19.3.jar:2.19.3]
        at org.opensearch.action.support.TransportAction.execute(TransportAction.java:190) [opensearch-2.19.3.jar:2.19.3]
        at org.opensearch.action.support.TransportAction.execute(TransportAction.java:109) [opensearch-2.19.3.jar:2.19.3]

```

The request completes fine.

 - OS: Linux dev40 6.8.5-301.fc40.aarch64
 - OpenSearch 2.19.3
 - opensearch security plugin 2.19.3.0

**Possible solution**

The problem is probably in the following [condition in the ActionPrivileges](https://github.com/opensearch-project/security/blob/2.19/src/main/java/org/opensearch/security/privileges/ActionPrivileges.java#L185) , when is actully not applied when `isLocalAll == true` (request containing `"*"`).

The solution could be as simple as checking the resolved indices before passing them to the `CheckTable`:

```
    if (resolvedLocalIndices.isEmpty()) {
        return PrivilegesEvaluatorResponse.ok();
    }
```

Please, let me know what do you think and if I need prepare a fix for that.

Mmuzaf:
## Description

This PR fixes an `IllegalArgumentException` that occured when resolved indices are empty in the `PrivilegesEvaluator`. This issue occurs when all indices are closed (e.g., during cluster maintenance) and operations like `_cat/recovery` are executed. 

## Category

Bug fix

## Why these changes are required?

When all indices in a cluster are closed, certain operations like `_cat/recovery` can still be valid and should be authorized. However, when `getAllIndicesResolved()` returns an empty set, the code was attempting to create a `CheckTable` with an empty collection, which throws an `IllegalArgumentException`. 

This prevents valid operations from being authorized when indices are closed.

## What is the old behavior before changes and new behavior after changes?

**Old behavior:**
- When all indices were closed and `getAllIndicesResolved()` returned an empty set, the code would attempt to create a `CheckTable` with an empty collection and throw an exception.

**New behavior:**
- When `getAllIndicesResolved()` returns an empty set allow operations to proceed
- Operations like `_cat/recovery` now succeed when users have the appropriate permissions, even when all indices are closed

## Issues Resolved

Resolves #5794

## Is this a backport?

No

## Do these changes introduce new permission(s) to be displayed in the static dropdown on the front-end?

No

## Testing

- Added unit test `hasIndexPrivilegeEmptyResolvedIndices()`
- Added integration test `IndexAuthorizationWithClosedIndicesIntTests`

### Manual Testing

- Verified that `_cat/recovery` and similar operations work correctly when all indices are closed

## Check List

- [x] New functionality has been documented (CHANGELOG.md updated)
- [x] New Roles/Permissions have a corresponding security dashboards plugin PR (N/A - no new permissions)
- [x] API changes companion pull request created (N/A - no API changes)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
           arguments: |
             ${{ matrix.gradle_task }} -Dbuild.snapshot=false
 
-      - uses: actions/upload-artifact@v5
+      - uses: actions/upload-artifact@v6
         if: always()
         with:
           name: ${{ matrix.platform }}-JDK${{ matrix.jdk }}-${{ matrix.gradle_task }}-reports
@@ -108,7 +108,7 @@ jobs:
           arguments: |
             ${{ matrix.gradle_task }} -Dbuild.snapshot=false
 
-      - uses: actions/upload-artifact@v5
+      - uses: actions/upload-artifact@v6
         if: always()
         with:
           name: ${{ matrix.platform }}-JDK${{ matrix.jdk }}-${{ matrix.gradle_task }}-reports
@@ -120,7 +120,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-      - uses: actions/download-artifact@v6
+      - uses: actions/download-artifact@v7
         with:
           path: downloaded-artifacts
 
@@ -165,7 +165,7 @@ jobs:
           arguments: |
             :integrationTest -Dbuild.snapshot=false
 
-      - uses: actions/upload-artifact@v5
+      - uses: actions/upload-artifact@v6
         if: always()
         with:
           name: integration-${{ matrix.platform }}-JDK${{ matrix.jdk }}-reports
@@ -208,7 +208,7 @@ jobs:
           arguments: |
             :integrationTest -Dbuild.snapshot=false
 
-      - uses: actions/upload-artifact@v5
+      - uses: actions/upload-artifact@v6
         if: always()
         with:
           name: integration-${{ matrix.platform }}-JDK${{ matrix.jdk }}-reports
@@ -250,7 +250,7 @@ jobs:
           arguments: |
             :opensearch-sample-resource-plugin:integrationTest -Dbuild.snapshot=false
 
-      - uses: actions/upload-artifact@v5
+      - uses: actions/upload-artifact@v6
         if: always()
         with:
           name: sample-plugin-integration-${{ matrix.platform }}-JDK${{ matrix.jdk }}-reports
@@ -282,7 +282,7 @@ jobs:
           arguments: |
             :opensearch-sample-resource-plugin:integrationTest -Dbuild.snapshot=false
 
-      - uses: actions/upload-artifact@v5
+      - uses: actions/upload-artifact@v6
         if: always()
         with:
           name: sample-plugin-integration-${{ matrix.platform }}-JDK${{ matrix.jdk }}-reports
@@ -384,7 +384,7 @@ jobs:
           java-version: 21
       - run: ./gradlew clean buildHealth -Dorg.gradle.jvmargs="-Xms1g -Xmx4g -XX:+UseG1GC -XX:MaxMetaspaceSize=512m"
 
-      - uses: actions/upload-artifact@v5
+      - uses: actions/upload-artifact@v6
         if: always()
         with:
           name: dependency-analysis
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -24,7 +24,7 @@ jobs:
 
     - run: ./gradlew test
 
-    - uses: actions/upload-artifact@v5
+    - uses: actions/upload-artifact@v6
       if: always()
       with:
         name: ${{ matrix.jdk }}-${{ matrix.test-run }}-reports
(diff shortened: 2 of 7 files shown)


nibix:
@Mmuzaf 

Thanks for the update! On Friday, I will be quite busy with other stuff, so I will only have time to take a closer look on Monday. 

> if there are no indices, and users do not have any indices priveleges for the action at all, such requests should be also forbidden.

We need to take a closer look at this:

- I *believe* that this behavior is the case also for all other index actions. If this is a fundamentally exposed behavior, we will likely have to mantain backwards compatiblity
- For remote index operations, such a behavior is actually required. It is even possible that there are coordinator clusters with no local indices at all. In such cases it is expected that no local privileges are required.

Of course, this is a discussable topic - at the moment, we are working on some fundamental behavior changes (see #3905, #5367 and #5399 ; I also need to write an updated RFC soon). Feel invited to discuss there.

nibix:
@Mmuzaf 

Sorry for the late reply, was just too busy :-(

That looks generally very good to me! Yet, I think we need to briefly reflect on this:

>  if there are no indices, and users do not have any indices priveleges for the action at all, such requests should be also forbidden

If I see this correctly, this would have the following effect:

- Doing a `GET /a*,remote_cluster:b/_search` would be allowed even if the local user has no local privileges for search if `a*` does not match any indices. That's due to the condition here: https://github.com/opensearch-project/security/blob/91a70978ae4f01e53bb2e4540a81464756cfe978/src/main/java/org/opensearch/security/privileges/actionlevel/RuntimeOptimizedActionPrivileges.java#L94
- Doing a `GET /*,remote_cluster:b/_search` would be then forbidden however as this goes through the new gate `this.index.hasAnyIndexPrivilegeForAction(context, actions)`

This is a kind of inconsistent behavior which we should avoid IMHO (even though it might be a case that seldomly happens in the real word). 

For now, I would propose not to do the `hasAnyIndexPrivilegeForAction()` check to stay in sync with the existing behavior. 

At the moment, we are busy on revising the index authorization behavior, see #5814. We could use that chance and implement the behavior change in that go. Feel invited to have a look there and comment :)

@cwperks wdyt?

src/integrationTest/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivilegesTest.java
```diff
@@ -1034,6 +1035,30 @@ public void statefulDisabled() throws Exception {
...
         }
+
+        /**
+         * Tests the behavior of hasIndexPrivilege when the resolved indices are empty.
+         * @throws Exception If failed.
+         */
+        @Test
+        public void hasIndexPrivilegeEmptyResolvedIndices() throws Exception {
+            SecurityDynamicConfiguration<RoleV7> roles = SecurityDynamicConfiguration.fromYaml(
+                "test_role:\n"
```
cwperks, line 1046: nit: I think we can use java text blocks to define multi-line strings.

cwperks:
When all indices are closed is it possible to call `_cat/recovery?expand_wildcards=all`?

I'm also wondering if it makes sense to respond with a warning telling the caller that it could not resolve to indices.

cwperks:
@Mmuzaf @nibix @willyborankin do you think we should respond with a warning when we cannot resolve to indices? I believe https://github.com/opensearch-project/OpenSearch/blob/799fb9bf46b1eb823f26f72e368f70a5675bf88f/server/src/main/java/org/opensearch/cluster/metadata/MetadataIndexTemplateService.java#L651-L662 shows an example of how it can be done.

nibix (dismissed):

nibix (approved):
re-approved

PR #5804, opened 2025-11-20
Cleaned up use of PrivilegesEvaluatorResponse

nibix:
### Description

This PR is in preparation of #5399 

The class `PrivilegesEvaluatorResponse` was using some less-than-optimal implementation patterns, especially exposing public attribute members which were directly manipulated by other classes. Additionally, the `isComplete()` pattern was never really used in a meaningful way.

This PR does the following changes:

- All member attributes of `PrivilegeEvaluatorResponse` are private
- Instances of the class `PrivilegeEvaluatorResponse` are now immutable
- The `isComplete()` pattern is given up. Instead, methods now return a non-null `PrivilegesEvaluatorResponse` if they can decide on the authorization status and null otherwise.
- Some unused code is removed
- 
* Category: Refactoring

### Testing

- existing tests

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/privileges/ProtectedIndexAccessEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/ProtectedIndexAccessEvaluator.java
@@ -86,13 +86,13 @@ public class ProtectedIndexAccessEvaluator {
             && !allowedRolesMatcher.matchAny(mappedRoles)) {
             auditLog.logMissingPrivileges(action, request, task);
             log.warn("{} for '{}' index/indices is not allowed for a regular user", action, indexMatcher);
-            return PrivilegesEvaluatorResponse.insufficient(action);
+            return PrivilegesEvaluatorResponse.insufficient("");
         }
 
         if (requestedResolved.isLocalAll() && deniedActionMatcher.test(action) && !allowedRolesMatcher.matchAny(mappedRoles)) {
             auditLog.logMissingPrivileges(action, request, task);
             log.warn("{} for '_all' indices is not allowed for a regular user", action);
-            return PrivilegesEvaluatorResponse.insufficient(action);
+            return PrivilegesEvaluatorResponse.insufficient("");
         }
         if ((requestedResolved.isLocalAll() || indexMatcher.matchAny(requestedResolved.getAllIndices()))
             && !allowedRolesMatcher.matchAny(mappedRoles)) {
--- a/src/main/java/org/opensearch/security/privileges/SnapshotRestoreEvaluator.java
+++ b/src/main/java/org/opensearch/security/privileges/SnapshotRestoreEvaluator.java
@@ -107,7 +107,7 @@ public class SnapshotRestoreEvaluator {
         if (rs != null && (rs.contains(securityIndex) || rs.contains("_all") || rs.contains("*"))) {
             auditLog.logSecurityIndexAttempt(request, action, task);
             log.warn("{} for '{}' as source index is not allowed", action, securityIndex);
-            return PrivilegesEvaluatorResponse.insufficient(action).reason(securityIndex + " as source index is not allowed");
+            return PrivilegesEvaluatorResponse.insufficient("").reason(securityIndex + " as source index is not allowed");
         }
         return null;
     }


src/main/java/org/opensearch/security/privileges/ProtectedIndexAccessEvaluator.java
```diff
@@ -67,32 +67,32 @@ public ProtectedIndexAccessEvaluator(final Settings settings, AuditLog auditLog)
...
-            return presponse.markComplete();
+            return PrivilegesEvaluatorResponse.insufficient("");
         }
 
         if (requestedResolved.isLocalAll() && deniedActionMatcher.test(action) && !allowedRolesMatcher.matchAny(mappedRoles)) {
             auditLog.logMissingPrivileges(action, request, task);
             log.warn("{} for '_all' indices is not allowed for a regular user", action);
-            presponse.allowed = false;
-            return presponse.markComplete();
+            return PrivilegesEvaluatorResponse.insufficient("");
```
DarshitChanpura, line 95: should this be `return PrivilegesEvaluatorResponse.insufficient(action);`?

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
```diff
@@ -123,17 +123,30 @@ private static List<String> deniedActionPatterns() {
...
+        PrivilegesEvaluatorResponse presponse = evaluateSystemIndicesAccess(
+            action,
+            requestedResolved,
+            request,
+            task,
+            context,
+            actionPrivileges,
+            user
+        );
+        if (presponse != null && !presponse.isAllowed()) {
```
DarshitChanpura, line 147: won't this prevent block below from executing when presponse is not null?

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
```diff
@@ -284,9 +293,7 @@ private void evaluateSystemIndicesAccess(
                         String.join(", ", getAllProtectedSystemIndices(requestedResolved))
                     );
                 }
-                presponse.allowed = false;
-                presponse.markComplete();
-                return;
+                return PrivilegesEvaluatorResponse.insufficient("");
```
DarshitChanpura, line 296: this should be `return PrivilegesEvaluatorResponse.insufficient(action);`

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
```diff
@@ -298,9 +305,7 @@ private void evaluateSystemIndicesAccess(
                             String.join(", ", getAllSystemIndices(requestedResolved))
                         );
                     }
-                    presponse.allowed = false;
-                    presponse.markComplete();
-                    return;
+                    return PrivilegesEvaluatorResponse.insufficient("");
```
DarshitChanpura, line 308: this should be `return PrivilegesEvaluatorResponse.insufficient(action);`

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
```diff
@@ -329,17 +332,12 @@ private void evaluateSystemIndicesAccess(
                                 matchingPluginIndices
                             );
                         }
-                        presponse.allowed = false;
-                        presponse.getMissingPrivileges();
-                        presponse.markComplete();
-                        return;
+                        return PrivilegesEvaluatorResponse.insufficient("");
```
DarshitChanpura, line 335: this should be `return PrivilegesEvaluatorResponse.insufficient(action);`

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
```diff
@@ -359,8 +357,7 @@ private void evaluateSystemIndicesAccess(
                 } else {
                     auditLog.logSecurityIndexAttempt(request, action, task);
                     log.warn("{} for '_all' indices is not allowed for a regular user", action);
-                    presponse.allowed = false;
-                    presponse.markComplete();
+                    return PrivilegesEvaluatorResponse.insufficient("");
```
DarshitChanpura, line 360: this should be `return PrivilegesEvaluatorResponse.insufficient(action);`

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
```diff
@@ -373,9 +370,7 @@ else if (containsSystemIndex && !isSystemIndexPermissionEnabled) {
                         if (log.isDebugEnabled()) {
                             log.debug("Filtered '{}' but resulting list is empty", securityIndex);
                         }
-                        presponse.allowed = false;
-                        presponse.markComplete();
-                        return;
+                        return PrivilegesEvaluatorResponse.insufficient("");
```
DarshitChanpura, line 373: this should be `return PrivilegesEvaluatorResponse.insufficient(action);`

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
```diff
@@ -385,10 +380,10 @@ else if (containsSystemIndex && !isSystemIndexPermissionEnabled) {
                     auditLog.logSecurityIndexAttempt(request, action, task);
                     final String foundSystemIndexes = String.join(", ", getAllSystemIndices(requestedResolved));
                     log.warn("{} for '{}' index is not allowed for a regular user", action, foundSystemIndexes);
-                    presponse.allowed = false;
-                    presponse.markComplete();
+                    return PrivilegesEvaluatorResponse.insufficient("");
```
DarshitChanpura, line 383: this should be `return PrivilegesEvaluatorResponse.insufficient(action);`

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluatorResponse.java
```diff
@@ -26,33 +26,61 @@
...
-    private String reason;
+    private final String reason;
 
     /**
      * Contains issues that were encountered during privilege evaluation. Can be used for logging.
      */
-    private List<PrivilegesEvaluationException> evaluationExceptions = new ArrayList<>();
+    private ImmutableList<PrivilegesEvaluationException> evaluationExceptions;
+
+    public PrivilegesEvaluatorResponse(
```
DarshitChanpura, line 53: out of curiosity, why not implement a complete Builder pattern over here? we already use chaining for method calls for this class.

src/main/java/org/opensearch/security/privileges/ProtectedIndexAccessEvaluator.java
nibix: Actually, in the first iteration of this PR the line was like this.

Unfortunately, we have hundreds of tests (yes, really, see https://github.com/opensearch-project/security/tree/main/src/test/java/org/opensearch/security/system_indices ) with asserts that assert on error messages like `No Permissions for []`. 

Changing the line above to `PrivilegesEvaluatorResponse.insufficient(action)` would change that error message, which both constitutes a breaking change and triggers the need to adjusting huge amounts of test asserts. I started that, but believe me, this is really incredible how massive that test code is.

As all this code will be going to moved to legacy state by #5399 anyway, such an investment is not really worthwhile IMHO.

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
nibix: when it is not null, it will be insufficient. That means that the request is going to be rejected anyway; thus, any changes to the request objects are unnecessary.

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
nibix: See https://github.com/opensearch-project/security/pull/5804#discussion_r2558512894

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
nibix: See https://github.com/opensearch-project/security/pull/5804#discussion_r2558512894

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
nibix: See https://github.com/opensearch-project/security/pull/5804#discussion_r2558512894

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
nibix: See https://github.com/opensearch-project/security/pull/5804#discussion_r2558512894

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
nibix: See https://github.com/opensearch-project/security/pull/5804#discussion_r2558512894

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluatorResponse.java
nibix: Could you elaborate on what you'd like to change for that?

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluatorImpl.java
```diff
@@ -347,23 +347,26 @@ public PrivilegesEvaluatorResponse evaluate(PrivilegesEvaluationContext context)
         // check snapshot/restore requests
         // NOTE: Has to go first as restore request could be for protected and/or system indices and the request may
         // fail with 403 if system index or protected index evaluators are triggered first
-        if (snapshotRestoreEvaluator.evaluate(request, task, action0, presponse).isComplete()) {
+        presponse = snapshotRestoreEvaluator.evaluate(request, task, action0);
```
cwperks, line 350: I think this PR looks good, but one thing I'd like to see is a separation of eligibility check and authz check. Similar to https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/filter/SecurityFilter.java#L424

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluatorResponse.java
cwperks: javadoc for the constructor args would be good here. Particularly around `createIndexRequestBuilder`. For newcomers to the codebase, its not obvious that sometimes the .kibana tenant indices can be created in this flow.

nibix: I agree that this the handling of `createIndexRequestBuilder` is a quite horrible thing. But IMHO, this should be done in a separate PR.

src/main/java/org/opensearch/security/privileges/ProtectedIndexAccessEvaluator.java
DarshitChanpura: oh wow.. okay. Maybe in a separate PR then.

src/main/java/org/opensearch/security/privileges/SystemIndexAccessEvaluator.java
DarshitChanpura: i see.

src/main/java/org/opensearch/security/privileges/PrivilegesEvaluatorImpl.java
DarshitChanpura: resolving to merge.

PR #5827, opened 2025-12-02
Improved index resolution and revised index authorization

### LINKED ISSUE #5814: [RFC] Revised index action authorization scheme
# Introduction

This is a follow-up and extension of the RFC on Refined index authorization from #3905 and #5367.

Based on that, we already worked on an implementation, which current state can be found in #5399. 

During implementation we found quite a few issues and oddities with the current index action authorization implementation. Many of these are documented inside the integration tests at https://github.com/opensearch-project/security/tree/main/src/integrationTest/java/org/opensearch/security/privileges/int_tests

The changes in #5399 have the goal to fix these. However, this sometimes requires quite fundamental, breaking changes. 

This RFC shall document these changes and open a forum to discuss the approaches.

# Feature flag

In order to keep backwards compatability, OpenSearch shall *not* use  the new behavior by default initially. A feature flag will be provided to switch on the new behavior. 

The feature flag will be named `privileges_evaluation_type` and located in the `config.yml` security plugin configuration under the `dynamic` attribute; it supports the values `legacy` (default) and `next_gen`. Setting it to `next_gen` enables the new functionality globally (note: I am not a big fan of this terminology, suggestions for improved terminology welcome!)

Example:

```yaml
_meta:
  type: "config"
  config_version: 2

config:
  dynamic:
     privileges_evaluation_type: next_gen
```

# `do_not_fail_on_forbidden` mode

The `do_not_fail_on_forbidden` mode will be enabled by default in the `next_gen` mode. However, the mode will employ new, refined semantics. In some regards, these semantics are deviating quite significantly from the old mode.

The following semantics have changed:

- In the old implementation, the security plugin could split aliases or data streams into member indices in case there were only privileges for some of the member indices present. This had the effect that filtered aliases functionality could be silently deactivated because the security plugin would replace a filtered alias by one of its (unfiltered) member indices. The new mode will no longer try to split up datastreams or filtered aliases.
- In the old implementation, the "do_not_fail_on_forbidden" mode would be only applied for a predefined list of actions. In the new mode, it will be applied to all requests that implement `IndicesRequest.Replaceable`. Thus, also 3rd party plugins which are not known to the security plugin can use the functionality.
- In the old implementation, all requests were subject to the index reduction employed by the  "do_not_fail_on_forbidden" mode. The new mode gives a request-level fine-grained control over the behavior: Requests which enumerate concrete indices without wildcards (like `index_a,index_b/_search`) are no longer subject to the index reduction performed by the "do_not_fail_on_forbidden" mode. Rather, if a privilege is missing to execute the request, the whole request will be denied. However, the user has the option to specifiy `index_a,index_b/_search?ignore_unavailable=true` in order to make sure that the request does not fail in case indices without privileges are present. If an index expression utilized wildcards, the "do_not_fail_on_forbidden" mode will be always active in the sense that the wildcard will only match indices with privileges.
- System indices and protected indices now integrate with the  "do_not_fail_on_forbidden" mode; so, any index for which such a special protection applies will be automatically filtered from the request.

# Alias privilege handling

Privilege handling for aliases has been refined in order to fix logic issues with alias specific actions.

- The `indices:admin/aliases` must be now explicitly given to any alias name this action shall be applied to.
- The same holds for any other action which can implicitly create an alias, like `indices:admin/create` with an alias option.
- In the old mode, reading from an alias or datastream was also possible if a user just had privileges for all members of the alias or datastream, but not for the alias or datastream itself. In the new mode, the user must have privileges on the alias or datastream itself in order to access it directly. Access via backing indices remains possible, though.

# Analyze action without index

The `_analyze` action can be used with and without indices. If it is used without indices, the user now needs a privilege for `indices:admin/analyze` on an arbitrary index.

# Point in time actions

The point in time actions `/_cat/pit_segments` and `/_search/point_in_time` have a special `_all` mode that returns segments for all open point in time sessions. To be able to use these, you now need the dedicated clusters privileges `indices:data/read/point_in_time/readall` or `cluster:monitor/point_in_time/segments/_all`, respectively. 

# Discontinued configuration options

When the new mode is active, a number of configuration options are no longer respected. This has several sources of motivation:

- The configuration option was originally introduced to provide backwards compatability and to avoid a breaking change. As we have a breaking change here anyway, it is fine to abandon the configuration option.
- The configuration option is no longer necessary with the new concepts.
- Reducing code and test complexity; currently, the authz part of the security plugin just has too many configuration options to make it properly maintaineable. See also #5624 

The dropped config options are documented in the following sections:

## config.yml

- `filtered_alias_mode`: As the new mode no longer splits up aliases, there is no need to warn about the use of filtered aliases. This flag is undocumented.
- `respect_request_indices_options`: As the index resolution is now done in core, this flag has no relevance any more. This flag is undocumented.
- `do_not_fail_on_forbidden_empty`: This flag has been provided for keeping backwards compatibility. 

## opensearch.yml

- `plugins.security.filter_securityindex_from_all_requests`: The security plugin will always filter the security index from requests
- `plugins.security.system_indices.enabled`: System index privilege handling will be always enabled
- `plugins.security.system_indices.permission.enabled`: System index permissions handling will be always enabled
- `plugins.security.unsupported.restore.securityindex.enabled`: This flag is undocumented and unsupported
- `plugins.security.enable_snapshot_restore_privilege`: This is replaced by the general purpose config option `plugins.security.privileges_evaluation.actions.universally_denied_actions`.
- `plugins.security.check_snapshot_restore_write_privileges`: This option also exists for backwards compatiblity. In the new mode, write privileges will be always checked for restore operations. 

# Additional new configuration attributes

Besides the already mentioned feature flag, we introduce a few config options which provide us more flexibility to influence privileges evaluator behavior. This is especially useful when mitigating bugs. These options should be considered as advanced options which are usually not used by normal users. 

- `plugins.security.privileges_evaluation.actions.force_as_cluster_actions`: Allows to treat actions that are usually considered index privileges, explicitly as cluster privileges instead.
- `plugins.security.privileges_evaluation.actions.universally_denied_actions`: Denies all requests of normal users for these actions. Only super admins can use these actions.
- `plugins.security.privileges_evaluation.actions.map_action_names`: Allows remapping of action names to privilege names.

nibix:
### Description

This introduces all the necessary changes for the improved index resolution and revised index authorization described in #5814. 

This has several advantages:

- The old index authorization and index resolution system had many inconsistencies and oddities. These are fixed by the new approach.
- The responsibility for index resolution is moved from the security plugin to the individual transport actions. This makes the whole authorization process much more robust and the code less messy and fragile.
- As each transport action knows its indices it is operating on very well, it can do the resolution in a more efficient manner (often actually no resolution takes place at all, as all indices are already determined as constant values). This reduces the performance overhead induced by the security plugin.
- The security plugin no longer needs to resolve data streams to its individual backing indices to check privileges; this brings further performance improvements.

* Category: Enhancement
  * Fundamental improvements for index authorization were due
  * By default, no behavioral changes are introduced; if the new index authorization system is enabled in config.yml, quite fundamental behavioral changes take place. These are described in #5814.

### Issues Resolved

#5814

### Testing

- extensive integration tests

### Check List


src/main/java/org/opensearch/security/privileges/actionlevel/legacy/PrivilegesEvaluatorImpl.java
```diff
@@ -148,63 +155,68 @@ public class PrivilegesEvaluatorImpl implements PrivilegesEvaluator {
...
-        final Settings settings,
-        final PrivilegesInterceptor privilegesInterceptor,
-        final IndexResolverReplacer irr,
-        FlattenedActionGroups actionGroups,
-        FlattenedActionGroups staticActionGroups,
-        SecurityDynamicConfiguration<RoleV7> rolesConfiguration,
-        ConfigV7 generalConfiguration,
-        Map<String, RoleV7> pluginIdToRolePrivileges
-    ) {
+    public PrivilegesEvaluatorImpl(CoreDependencies coreDependencies, DynamicDependencies dynamicDependencies) {
```
cwperks, line 162: Much cleaner like this!

config/config.yml
```diff
@@ -60,11 +60,10 @@ _meta:
...
   dynamic:
-    # Set filtered_alias_mode to 'disallow' to forbid more than 2 filtered aliases per index
-    # Set filtered_alias_mode to 'warn' to allow more than 2 filtered aliases per index but warns about it (default)
-    # Set filtered_alias_mode to 'nowarn' to allow more than 2 filtered aliases per index silently
-    #filtered_alias_mode: warn
-    #do_not_fail_on_forbidden: false
+    # Set privileges_evaluation_type to "next_gen" to enable the new privileges evaluation engine.
+    # This should give you a more seamless experience, especially for users with restricted privileges
+    # working in OpenSearch Dashboards.
+    # privileges_evaluation_type: next_gen
```
cwperks: I'm not crazy about the name next_gen for this. Do you think there can be another more descriptive name that is also succinct?

nibix: I am also struggling with this one. I agree that `next_gen` is weird. But all other ideas I have, are also weird. I asked a fellow artificial intelligence for ideas. Their suggestions:

- `improved`: What do we do when release another improvement `improved_improved`?
- `correct`: That's a bold claim and suffers from the same issues as `improved`.
- `precise`: Also weird.
- `revised`: Same isues as `improved`. 

I could imagine that one uses an OpenSearch major release number to suggest that this is going to be fully activated in that release. So:

- `v4` 

But that might be also confusing?

cwperks: v4 would be more descriptive so I would be supportive of that.

shikharj05: +1, v4 works better than any other names

src/main/java/org/opensearch/security/privileges/actionlevel/nextgen/PrivilegesEvaluatorImpl.java
```diff
@@ -0,0 +1,846 @@
...
+        if (request.includeGlobalState()) {
+            return PrivilegesEvaluatorResponse.insufficient(context.getAction())
+                .reason("Restoring snapshot with 'include_global_state' enabled is not allowed");
+        }
+
+        if (!clusterStateSupplier.get().nodes().isLocalNodeElectedClusterManager()) {
+            // We need to return ok here, because we can only retrieve the snapshot info on a cluster manager node.
+            // This is fine, as the next thing the TransportAction implementation will do, is forwarding the request to a
+            // cluster manager node.
+            return PrivilegesEvaluatorResponse.ok();
```
shikharj05, line 641: will the cluster manager re-run this code to evaluate actual permissions?

nibix: Yes, this is verified by the SnapshotAuthorizationIntTest, which creates a multi-node cluster and usually targets a non-cm node:

https://github.com/opensearch-project/security/blob/1d6fb4bdc0be681289b4b29c966adfaa81959494/src/integrationTest/java/org/opensearch/security/privileges/int_tests/SnapshotAuthorizationIntTests.java#L188-L194

Generally, an action request which is retargeted to another node (possibly a cm node), is always going through the ActionFilter on each node. A node which receives an action request has no way to know whether that request hit it directly or whether that request was retargeted.

This approach is not new; in the old privilege evaluation implementation, it was just not commented and not made that explicit. The logic which has the same effect is this:

https://github.com/opensearch-project/security/blob/c050f12f83189dbbf09cd5dcb856eeb00c83fdb9/src/main/java/org/opensearch/security/resolver/IndexResolverReplacer.java#L711-L739

cwperks (approved):

shikharj05 (approved):

config/config.yml
```diff
@@ -60,11 +60,10 @@ _meta:
...
   dynamic:
-    # Set filtered_alias_mode to 'disallow' to forbid more than 2 filtered aliases per index
-    # Set filtered_alias_mode to 'warn' to allow more than 2 filtered aliases per index but warns about it (default)
-    # Set filtered_alias_mode to 'nowarn' to allow more than 2 filtered aliases per index silently
-    #filtered_alias_mode: warn
-    #do_not_fail_on_forbidden: false
+    # Set privileges_evaluation_type to "v4" to enable the new privileges evaluation engine.
+    # This should give you a more seamless experience, especially for users with restricted privileges
+    # working in OpenSearch Dashboards.
+    # privileges_evaluation_type: "v4"
```
cwperks, line 66: wdyt about making this a dynamic setting so that the default can more easily change from one major version to the next?

cwperks (approved):

config/config.yml
cwperks: I suppose the same problem exists with dynamic cluster settings. Can we just confirm that this is not persisted with the default value in a fresh cluster or on upgrade?

nibix: The default value is null, and the attr is annotated with`@JsonInclude(JsonInclude.Include.NON_NULL)`, so we do not persist the attribute by default.

nibix:
@shikharj05 @DarshitChanpura @derek-ho @RyanL1997 @reta @willyborankin 

I kind this is a big PR; is there something I can facilitate the approval? It is already sitting around for a while.

PR #5831, opened 2025-12-03
Hardens input validation for resource sharing APIs

DarshitChanpura:
### Description
Prior to this change users could input practically infinite value in strings of input parameters in resource-sharing related REST APIs. This PR hardens those inputs by introducing limits on those parameters.

* Category:  Enhancement

### Testing
integration testing

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Features
 
-### Enhancements
+### Enh
+- Make security plugin aware of FIPS build param (-Pcrypto.standard=FIPS-140-3) ([#5952](https://github.com/opensearch-project/security/pull/5952))
 - Hardens input validation for resource sharing APIs ([#5831](https://github.com/opensearch-project/security/pull/5831)
 
 ### Bug Fixes
-
+- Fix the issue of unprocessed X-Request-Id ([#5954](https://github.com/opensearch-project/security/pull/5954))
 ### Refactoring
 
 ### Maintenance
@@ -25,6 +26,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Bump `aws-actions/configure-aws-credentials` from 5 to 6 ([#5946](https://github.com/opensearch-project/security/pull/5946))
 - Bump `ch.qos.logback:logback-classic` from 1.5.26 to 1.5.28 ([#5948](https://github.com/opensearch-project/security/pull/5948))
 - Bump `com.github.seancfoley:ipaddress` from 5.5.1 to 5.6.1 ([#5949](https://github.com/opensearch-project/security/pull/5949))
+- Bump `spring_version` from 7.0.3 to 7.0.4 ([#5957](https://github.com/opensearch-project/security/pull/5957))
+- Bump `org.junit.jupiter:junit-jupiter-api` from 5.14.2 to 5.14.3 ([#5956](https://github.com/opensearch-project/security/pull/5956))
+- Bump `org.checkerframework:checker-qual` from 3.53.0 to 3.53.1 ([#5955](https://github.com/opensearch-project/security/pull/5955))
 
 ### Removed
 
--- a/build.gradle
+++ b/build.gradle
@@ -11,6 +11,7 @@
 
 
 import com.diffplug.gradle.spotless.JavaExtension
+import org.opensearch.gradle.info.FipsBuildParams
 import org.opensearch.gradle.test.RestIntegTestTask
 import groovy.json.JsonBuilder
 
@@ -33,7 +34,7 @@ buildscript {
         jjwt_version = '0.13.0'
         guava_version = '33.5.0-jre'
         jaxb_version = '2.3.9'
-        spring_version = '7.0.3'
+        spring_version = '7.0.4'
 
         if (buildVersionQualifier) {
             opensearch_build += "-${buildVersionQualifier}"
@@ -510,7 +511,7 @@ configurations {
             force "org.apache.httpcomponents:httpclient:4.5.14"
             force "org.apache.httpcomponents:httpcore:4.4.16"
             force "com.google.errorprone:error_prone_annotations:${versions.error_prone_annotations}"
-            force "org.checkerframework:checker-qual:3.53.0"
+            force "org.checkerframework:checker-qual:3.53.1"
             force "ch.qos.logback:logback-classic:1.5.28"
             force "commons-io:commons-io:2.21.0"
             force "com.carrotsearch.randomizedtesting:randomizedtesting-runner:2.8.3"
@@ -677,9 +678,16 @@ dependencies {
     implementation "com.google.guava:guava:${guava_version}"
     implementation 'org.greenrobot:eventbus-java:3.3.1'
     implementation 'commons-cli:commons-cli:1.11.0'
-    implementation "org.bouncycastle:bc-fips:${versions.bouncycastle_jce}"
-    implementation "org.bouncycastle:bcpkix-fips:${versions.bouncycastle_pkix}"
-    implementation "org.bouncycastle:bcutil-fips:${versions.bouncycastle_util}"
+    // When building with -Pcrypto.standard=FIPS-140-3, bcFips jars are provided by OpenSearch
+    if (FipsBuildParams.isInFipsMode()) {
+        compileOnly "org.bouncycastle:bc-fips:${versions.bouncycastle_jce}"
+        compileOnly "org.bouncycastle:bcpkix-fips:${versions.bouncycastle_pkix}"
+        compileOnly "org.bouncycastle:bcutil-fips:${versions.bouncycastle_util}"
+    } else {
+        implementation "org.bouncycastle:bc-fips:${versions.bouncycastle_jce}"
+        implementation "org.bouncycastle:bcpkix-fips:${versions.bouncycastle_pkix}"
+        implementation "org.bouncycastle:bcutil-fips:${versions.bouncycastle_util}"
+    }
     implementation 'org.ldaptive:ldaptive:1.2.3'
     implementation 'com.nimbusds:nimbus-jose-jwt:10.7'
     implementation 'com.rfksystems:blake2b:2.0.0'
@@ -743,7 +751,7 @@ dependencies {
     runtimeOnly 'org.apache.ws.xmlschema:xmlschema-core:2.3.2'
     runtimeOnly 'org.apache.santuario:xmlsec:2.3.5'
     runtimeOnly "com.github.luben:zstd-jni:${versions.zstd}"
-    runtimeOnly 'org.checkerframework:checker-qual:3.53.0'
+    runtimeOnly 'org.checkerframework:checker-qual:3.53.1'
     runtimeOnly 'org.scala-lang.modules:scala-java8-compat_3:1.0.2'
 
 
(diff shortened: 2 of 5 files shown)


src/main/java/org/opensearch/security/resources/utils/InputValidation.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.regex.Pattern;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+import org.opensearch.core.common.Strings;
+
+/**
+ * Validates user inputs supplied to resource-sharing REST APIs
+ */
+public final class InputValidation {
```
cwperks: Let's make this more generic and part of the [RequestContentValidator](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java).

cwperks: Specifically, for String type we should allow setting params like maxLength and possibly a callback function to validate similar to how OpenSearch Settings are validated. Look at the Validator param of https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/settings/Setting.java#L252-L269

DarshitChanpura: done.

src/test/java/org/opensearch/security/multitenancy/test/TenancyMultitenancyEnabledTests.java
```diff
@@ -92,7 +92,7 @@ public void testMultitenancyDisabled_endToEndTest() throws Exception {
 
         final HttpResponse updateMutlitenancyToDisabled = nonSslRestHelper().executePutRequest(
             "/_plugins/_security/api/tenancy/config",
-            "{\"multitenancy_enabled\": \"false\"}",
+            "{\"multitenancy_enabled\": false}",
```
DarshitChanpura: these tests were supplying a string type but the multi-tenancy api validator expected a boolean type. my changes fixed it by introducing an additional case for handling BOOLEANs and hence this test failed for string type. CHanging it to boolean fixed it, which is correct way for this param.

RyanL1997 (dismissed):

src/test/java/org/opensearch/security/multitenancy/test/TenancyMultitenancyEnabledTests.java
cwperks: What would it take to support both? Looks like its breaking the current API if the current API supports supplying both ways.

src/main/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -153,44 +154,68 @@ private void handleMigrate(RestChannel channel, RestRequest request, Client clie
...
+        // Raw JSON for default_access_level
+        JsonNode defaultAccessNode = body.get("default_access_level");
+
+        // Business logic validation (after framework validation)
+        try {
+            RequestContentValidator.validateJsonPath("username_path", userNamePath);
+            RequestContentValidator.validateJsonPath("backend_roles_path", backendRolesPath);
+            RequestContentValidator.validateDefaultOwner(defaultOwner);
+            RequestContentValidator.validateSourceIndex(sourceIndex, resourcePluginInfo.getResourceIndicesForProtectedTypes());
+            RequestContentValidator.validateDefaultAccessLevelNode(defaultAccessNode);
```
cwperks: Can we keep RequestContentValidator generic? I think methods like these sound too specific. 

For the RequestContentValidator I believe the rules should be defined on the route itself and used as part of validating API payloads.

src/test/java/org/opensearch/security/multitenancy/test/TenancyMultitenancyEnabledTests.java
DarshitChanpura: the data-type is marked as boolean, but yeah I can add a Boolean.parseValue wrapper if its a string

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
```diff
@@ -393,4 +535,242 @@ public String message() {
         }
 
     }
+
+    /* ========================================================================
+     * Input Validation Utilities (Generic)
+     * ======================================================================== */
+
+    public static final int MAX_ID_LENGTH = 256;
+    public static final int MAX_TYPE_LENGTH = 256;
```
cwperks: Can we move these closer to the applicable routes? i.e. in the subclass of AbstractApiAction that defines a route with these fields in the payload. Some of the high-level values like max length for list params can be at the top-level RequestContentValidator.

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
```diff
@@ -393,4 +535,242 @@ public String message() {
...
+    public static final int MAX_LEVEL_LENGTH = 256;
+    public static final int MAX_PRINCIPAL_LENGTH = 256;
+    public static final int MAX_PATH_LENGTH = 256;
+    public static final int MAX_NAME_LENGTH = 256;
+    public static final int MAX_ARRAY_SIZE = 100_000;
+
+    // Alphanumeric + _ - : OR : * - "*" is only allowed as standalone
+    private static final Pattern SAFE_VALUE = Pattern.compile("^(\\*|[A-Za-z0-9_:-]+)$");
```
cwperks, line 555: Where is the safe criteria coming from and in what instances is this used?

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
```diff
@@ -393,4 +535,242 @@ public String message() {
...
+            if (value instanceof String strValue) {
+                validateSafeValue(fieldName, strValue, MAX_PRINCIPAL_LENGTH, allowWildcard);
+            }
+        };
+    }
+
+    /**
+     * Validator for principal values (no wildcard support)
+     */
+    public static final FieldValidator PRINCIPAL_VALIDATOR = principalValidator(false);
```
cwperks: Let's move these to the applicable subclass of AbstractApiAction

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
DarshitChanpura: This is generic pattern where roles can have : or - or _ in it. "*" is only allowed by itself for now.

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
DarshitChanpura: no subclasses using this value implement AbstractApiAction, and it should be fine here for now. We already have validatePassword in here that is specific to users API action.

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
DarshitChanpura: These are used in 3 classes two of which do not implement AbstractApiAction.

DarshitChanpura: I've combined these into 1 generic constant: MAX_STRING_LENGTH

src/test/java/org/opensearch/security/resources/sharing/ShareWithTests.java
```diff
@@ -106,7 +109,13 @@ public void testFromXContentWithStartObject() throws IOException {
 
         parser.nextToken();
 
-        ShareWith shareWith = ShareWith.fromXContent(parser);
+        ShareWith shareWith = ShareWith.fromXContent(
+            parser,
+            org.opensearch.security.dlic.rest.validation.RequestContentValidator.allowedValuesValidator(
```
cwperks: nit: use class level import

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
```diff
@@ -393,4 +543,227 @@ public String message() {
...
+            if (count > MAX_ARRAY_SIZE) {
+                throw new IllegalArgumentException("Array field [" + fieldName + "] exceeds maximum size of " + MAX_ARRAY_SIZE);
+            }
+        }
+    };
+
+    /**
+     * Creates a validator that checks if a string value is in an allowed set
+     */
+    public static FieldValidator allowedValuesValidator(Set<String> allowedValues, String errorMessage) {
```
cwperks, line 723: Can we overload this method to avoid having to pass `null` in as the second param?

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
```diff
@@ -393,4 +543,227 @@ public String message() {
...
+        }
+        if (!allowedValues.contains(value)) {
+            throw new IllegalArgumentException("Invalid " + fieldName + " [" + value + "]. Allowed " + errorContext + ": " + allowedValues);
+        }
+    }
+
+    /**
+     * Validates a JSON object node with non-empty string values
+     */
+    public static void validateObjectWithStringValues(String fieldName, JsonNode node) {
```
cwperks: Can we inline this if its just used in the single context of the `default_access_level` param? I think the name of this method is a bit of a misnomer if its simply validating that the key-value pair is non-empty.

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
```diff
@@ -393,4 +543,227 @@ public String message() {
...
+        String value = node.asText();
+        requireNonEmpty(fieldName, value);
+        validateMaxLength(fieldName, value, maxLength);
+        return value;
+    }
+
+    /**
+     * Extracts an optional text field from JSON with length validation
+     */
+    public static String getOptionalText(JsonNode body, String fieldName, int maxLength) {
```
cwperks: If this is unused, can we remove it?

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
```diff
@@ -393,4 +543,227 @@ public String message() {
...
+            );
+        }
+    }
+
+    /* ---------------------- JSON extraction helpers ---------------------- */
+
+    /**
+     * Extracts a required text field from JSON with length validation
+     */
+    public static String getRequiredText(JsonNode body, String fieldName, int maxLength) {
```
cwperks: Let's remove all unused methods.

src/main/java/org/opensearch/security/resources/sharing/Recipients.java
```diff
@@ -96,19 +97,64 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
...
+                        arraySizeValidator.validate(fieldName, count);
+                    }
+
+                    String value = parser.text();
+
+                    // Validate element value if validator provided
+                    // For "users" field, allow wildcard; for others, don't
+                    if (elementValidator != null) {
+                        // Check if wildcard should be allowed based on recipient type
+                        boolean allowWildcard = (recipient == Recipient.USERS);
```
cwperks, line 147: This is ok, but in the longer-term let's add a top-level flag about whether a resource is publicly visible and at what access level users have on the resource if it hasn't been explicitly shared with them otherwise.

DarshitChanpura: this is to simply check whether wildcard is allowed when parsing the request.

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
DarshitChanpura: sure.

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
DarshitChanpura: Let's leave it here I will change the name to validateNonEmptyValuesInAnObject

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
DarshitChanpura: done

cwperks (approved):

cwperks (dismissed):

RyanL1997 (dismissed):

RyanL1997 (approved):

cwperks (approved):

PR #5858, opened 2025-12-15
Skip hasExplicitIndexPrivilege check for plugin users accessing their own system indices

llilyy:
### Description

* Category

  * Bug fix

  * When `isSystemIndexPermissionEnabled` is set to `true`, plugin users are unable to access their own registered system indices. This happens because the `hasExplicitIndexPrivilege` check is performed before the plugin user's own system index access logic, causing the request to be rejected even though the plugin should have access to its own system indices.

  * This breaks the expected behavior where plugins can manage their own system indices without requiring explicit `system_index` permission.

  * Old behavior:
    * When `isSystemIndexPermissionEnabled = true`, all system index access (including plugin's own system indices) goes through `hasExplicitIndexPrivilege` check
    * Plugin users are rejected with "no permissions" error when trying to access their own registered system indices
    * The plugin user special handling logic at the bottom of `evaluateSystemIndicesAccess()` is never reached because the request is already rejected

  * New behavior:
    * Plugin users can access their own registered system indices for non-cluster actions without requiring `hasExplicitIndexPrivilege` check
    * The `isPluginIndexAction` condition is added to skip the permission check when:
      * The user is a plugin user
      * The action is not a cluster-level action
      * The request only targets the plugin's own registered system indices
    * Cluster-level actions are still properly restricted for plugin users
    * A new helper method `getMatchingPluginIndices()` is added to calculate plugin's own system indices once and reuse it, avoiding code duplication

### Issues Resolved

https://github.com/opensearch-project/job-scheduler/issues/865

### Testing

**New Integration Tests Added:**
- Created `SystemIndexPermissionEnabledTests.java` with `isSystemIndexPermissionEnabled = true` configuration
- Tests verify plugin users can access their own system indices when system index permission is enabled:
  - `testPluginShouldBeAbleToIndexDocumentIntoItsSystemIndex`
  - `testPluginShouldBeAbleToBulkIndexDocumentIntoItsSystemIndex`
  - `testPluginShouldBeAbleSearchOnItsSystemIndex`
  - `testPluginShouldBeAbleGetOnItsSystemIndex`
  - `testPluginShouldBeAbleUpdateOnItsSystemIndex`
- Tests verify plugin users are still properly restricted:
  - `testPluginShouldNotBeAbleToIndexDocumentIntoSystemIndexRegisteredByOtherPlugin`
  - `testPluginShouldNotBeAbleToRunClusterActions`
  - `testPluginShouldNotBeAbleToBulkIndexDocumentIntoMixOfSystemIndexWhereAtLeastOneDoesNotBelongToPlugin`

**Existing Tests Verified:**
- `SystemIndexTests` - All tests pass (no regression with `isSystemIndexPermissionEnabled = false`)
- `SystemIndexAccessEvaluatorTest` - All unit tests pass (no impact on regular user behavior)

**Test Execution:**
```bash
./gradlew :integrationTest --tests "org.opensearch.security.systemindex.SystemIndexTests" --tests "org.opensearch.security.systemindex.SystemIndexPermissionEnabledTests"
# BUILD SUCCESSFUL

./gradlew :test --tests "org.opensearch.security.privileges.SystemIndexAccessEvaluatorTest"
# BUILD SUCCESSFUL
```

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,15 +13,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Enhancements
 
 ### Bug Fixes
-- Fix IllegalArgumentException when resolved indices are empty in PrivilegesEvaluator ([#5770](https://github.com/opensearch-project/security/pull/5797)
-- Fixes an issue where recursive LDAP role search would fail with a NullPointerException ([#5861](https://github.com/opensearch-project/security/pull/5861))
-- Create a WildcardMatcher.NONE when creating a WildcardMatcher with an empty string ([#5694](https://github.com/opensearch-project/security/pull/5694))
-- Improve array validator to also check for blank string in addition to null ([#5714](https://github.com/opensearch-project/security/pull/5714))
-- Use RestRequestFilter.getFilteredRequest to declare sensitive API params ([#5710](https://github.com/opensearch-project/security/pull/5710))
-- Fix deprecated SSL transport settings in demo certificates ([#5723](https://github.com/opensearch-project/security/pull/5723))
-- Updates DlsFlsValveImpl condition to return true if request is internal and not a protected resource request ([#5721](https://github.com/opensearch-project/security/pull/5721))
-- [Performance] Call AdminDns.isAdmin once per request ([#5752](https://github.com/opensearch-project/security/pull/5752))
-- Update operations on `.kibana` system index now work correctly with Dashboards multi tenancy enabled. ([#5778](https://github.com/opensearch-project/security/pull/5778))
 - Fix IllegalArgumentException when resolved indices are empty in PrivilegesEvaluator ([#5770](https://github.com/opensearch-project/security/pull/5797))
 - Fixes an issue where recursive LDAP role search would fail with a NullPointerException ([#5861](https://github.com/opensearch-project/security/pull/5861))
 - Skip hasExplicitIndexPrivilege check for plugin users accessing their own system indices ([#5858](https://github.com/opensearch-project/security/pull/5858))


src/integrationTest/java/org/opensearch/security/systemindex/SystemIndexPermissionEnabledTests.java
```diff
@@ -0,0 +1,193 @@
...
+    @Before
+    public void setup() {
+        try (TestRestClient client = cluster.getRestClient(cluster.getAdminCertificate())) {
+            client.delete(SYSTEM_INDEX_1);
+        }
+    }
+
+    @Test
+    public void testPluginShouldBeAbleToIndexDocumentIntoItsSystemIndex() {
+        try (TestRestClient client = cluster.getRestClient(USER_ADMIN)) {
```
nibix: I am a bit confused why the admin user is used for these tests. Could you elaborate on that?

llilyy: I referenced the existing `SystemIndexTests.java`.
The `USER_ADMIN` is only used to trigger the plugin's REST endpoint.
The actual system index access is performed by the plugin user.
This seems to replicate the real-world scenario where a plugin receives a user request and performs operations using its own plugin identity.

llilyy:
Thanks for the review, @nibix!
I looked into it, and `IndexAuthorizationReadOnlyIntTests` uses `@ParametersFactory` to test various user combinations, but plugin user tests always run as the plugin user internally regardless of which user calls the REST endpoint. This doesn't fit well with the matrix structure.

I can either:
1. Move the tests to a separate class in the same directory while reusing the `ClusterConfig` enum
2. Merge them into `IndexAuthorizationReadOnlyIntTests` if you prefer

Which approach would you prefer?

src/integrationTest/java/org/opensearch/security/systemindex/SystemIndexPermissionEnabledTests.java
nibix: ah, understood, thanks!

llilyy:
@nibix 

This test structure is not using inheritance for ad-hoc code reuse. It defines an explicit privileges test contract that is executed across multiple cluster configurations.

The abstract base class expresses the complete set of privilege rules that must hold, while the concrete subclasses provide different environments in which the same contract is validated. These tests are actually executed polymorphically: the same test methods are run with different `LocalCluster` implementations, and behavior is dynamically bound via `getCluster()` at runtime.

In privilege-focused tests, duplicating the same privilege rules and expectations across multiple test classes is particularly error-prone. When privilege semantics change, duplicated logic is easy to update inconsistently, which can lead to missed cases, diverging behavior, or even false positives. Centralizing the privilege rules in a single abstract test suite intentionally enforces a single source of truth for the privileges contract, while cleanly separating environmental differences such as cluster configuration or feature flags.

From the perspective of privilege correctness and long-term test stability, this is a deliberate and justified design choice rather than an inheritance anti-pattern. To address the readability concern, I've added comprehensive Javadoc to the abstract base class and the concrete implementations explaining that the actual test cases are defined in the base class to enforce consistency across configurations.

nibix:
@llilyy 

given the current test implementation, we will have to re-implement the tests before we can merge #5827 in order to also have test coverage for the new privilege evaluation mode. This will then add friction to merging #5827. 

Regarding this:

> In privilege-focused tests, duplicating the same privilege rules and expectations across multiple test classes is particularly error-prone. When privilege semantics change, duplicated logic is easy to update inconsistently, which can lead to missed cases, diverging behavior, or even false positives. Centralizing the privilege rules in a single abstract test suite intentionally enforces a single source of truth for the privileges contract, while cleanly separating environmental differences such as cluster configuration or feature flags.

As far as I understand, the approach with a parameterized test using the `ClusterConfig` enum would provide you the same benefits. No code duplication would occur. Or do I get something wrong?

> From the perspective of privilege correctness and long-term test stability, this is a deliberate and justified design choice rather than an inheritance anti-pattern

For this exact purpose, parameterized tests do exist.

Issue #5859, opened 2025-12-15
[BUG] Unexpected exception java.util.ConcurrentModificationException


liupengkd:
**What is the bug?**
when  I fixed the ConcurrentModificationException like [#5084 ](https://github.com/opensearch-project/security/pull/5084) the ConcurrentModificationException still occurred

[error]
[2025-12-08T00:59:14,663][ERROR][o.o.s.f.SecurityFilter ] [testsearch] Unexpected exception java.util.ConcurrentModificationException
java.util.ConcurrentModificationException: null
at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1605) ~[?:?]
at java.base/java.util.HashMap$KeyIterator.next(HashMap.java:1628) ~[?:?]
at java.base/java.util.Collections$UnmodifiableCollection$1.next(Collections.java:1078) ~[?:?]
at java.base/java.util.AbstractCollection.addAll(AbstractCollection.java:337) ~[?:?]
at org.opensearch.security.securityconf.ConfigModelV7$RoleMappingHolder.map(ConfigModelV7.java:1215) ~[opensearch-security-2.18.0.0.jar:2.18.0.0]
at org.opensearch.security.securityconf.ConfigModelV7.mapSecurityRoles(ConfigModelV7.java:1280) ~[opensearch-security-2.18.0.0.jar:2.18.0.0]
at org.opensearch.security.privileges.PrivilegesEvaluator.mapRoles(PrivilegesEvaluator.java:564) ~[opensearch-security-2.18.0.0.jar:2.18.0.0]
at org.opensearch.security.privileges.PrivilegesEvaluator.createContext(PrivilegesEvaluator.java:236) ~[opensearch-security-2.18.0.0.jar:2.18.0.0]

**How can one reproduce the bug?**
We can pull out the access logic for User.java and ConfigModelV7.java and write a concurrency test script to reproduce the problem
1. MyUserTest.java:

```
import java.util.Collection;
import java.util.Collections;
import java.util.HashSet;
import java.util.Set;

class MyUserTest {

    private final Set<String> securityRoles = Collections.synchronizedSet(new HashSet<String>());

    public final void addSecurityRoles(final Collection<String> securityRoles) {
        if (securityRoles != null && this.securityRoles != null) {
            this.securityRoles.addAll(securityRoles);
        }
    }

    public final Set<String> getSecurityRoles() {
        return this.securityRoles == null
            ? Collections.synchronizedSet(Collections.emptySet())
[...]

PR #5860, opened 2025-12-16
fix ConcurrentModificationException for SecurityRoles for 2.x

liupengkd:
### Description

fix ConcurrentModificationException for SecurityRoles for 2.x
[#5859 ](https://github.com/opensearch-project/security/issues/5859)

Change `user.getSecurityRoles()` to `user.getRawSecurityRoles()` to ensure the effectiveness of synchronization.

### Issues Resolved
[#5859 ](https://github.com/opensearch-project/security/issues/5859)

### Testing

tested locally as [#5859 ](https://github.com/opensearch-project/security/issues/5859)

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/user/User.java
+++ b/src/main/java/org/opensearch/security/user/User.java
@@ -282,9 +282,7 @@ public class User implements Serializable, Writeable, CustomAttributesAware {
     }
 
     public final Set<String> getSecurityRoles() {
-        return this.securityRoles == null
-            ? Collections.synchronizedSet(Collections.emptySet())
-            : this.securityRoles;
+        return this.securityRoles == null ? Collections.synchronizedSet(Collections.emptySet()) : this.securityRoles;
     }
 
     /**


src/main/java/org/opensearch/security/user/User.java
```diff
@@ -281,6 +281,10 @@ public final void addSecurityRoles(final Collection<String> securityRoles) {
         }
     }
 
+    public final Set<String> getRawSecurityRoles() {
+        return this.securityRoles
```
nibix: Interesting; what do you think about just removing the ` Collections.unmodifiableSet()` in line 291? That should fix it as well; and it should be fine as we are exposing the writeable object now anyway.

liupengkd: My consideration here is to minimize the impact, so that only the code path I modified will be affected. This prevents other places calling the original interface from having dependencies on the read-only protection returned here.

liupengkd:
> Thank you for the contribution! You'd need to additionally sign off your commit. Please review https://github.com/opensearch-project/.github/blob/main/CONTRIBUTING.md for more information on that.

Thank you for the reminder. I have resubmitted the commit with the signature.

src/main/java/org/opensearch/security/user/User.java
nibix: TBH, for me it feels quite confusing to have two methods, where one returns a modifiable version and the other returns an unmodifiable version. if we are going to have two methods, we should have quite good code comments explaining the purposes.

I would still vote for just removing the `Collections.unmodifiableSet()` ...

liupengkd: Thank you for your suggestion. I finally removed `Collections.unmodifiableSet()` according to your proposal.

PR #5866, opened 2025-12-18
Run the Security plugin under FIPS approved-only mode with BC providers

### LINKED ISSUE #4254: [FEATURE] FIPS-140 Compliance Roadmap for OpenSearch
# Introduction

The Federal Information Processing Standard (FIPS), is a required standard for any software that will be used in the Government of the United States of America or the Government of Canada. It defines the security mechanisms that need to be implemented in software and hardware when working with information that will be used by state entities.

One key component of this standard is the 140 series, it specifies requirements for cryptographic modules in order to guarantee that the information is as secure as possible and algorithms that are known to be weak are not used for encryption.

Even though this standard is not required to be implemented in software used by private companies, many do increasingly prefer the FIPS compliant version of the software programs they use.

OpenSearch is a software that manages substantial volumes of diverse information. Therefore, it is crucial for user confidence that this information is handled as securely as possible. To this end there has been increasing interest in the community for implementing a mechanism that enables users of OpenSearch to operate a FIPS-compliant version of the software.

**This RFC is related to and based on [this issue](https://github.com/opensearch-project/security/issues/3420) and [this issue](https://github.com/opensearch-project/security/issues/1497) and is intended to help facilitate a community discussion and provide a structured, iterative roadmap towards implementation**.

## Scope of this RFC

All work related to FIPS 140 must consider the current version of the standard, which is FIPS 140-3.

FIPS consists of hardware and software related security aspects, however, in this RFC only software related aspects are considered.

FIPS establishes different levels of security from 1 to 4, we currently only aim at level 1, to implement the starting point for further improvements in the future.

For OpenSearch to be FIPS compliant at level 1 as a software cryptographic module it is required that:

- OpenSearch needs to be able to run in a FIPS compliant environment with:
  - Operating system configured in FIPS mode
  - Java Runtime Environment configured in FIPS mode
- OpenSearch needs to be able to be configured to run in FIPS mode
- In FIPS mode, all cryptographic algorithms used in OpenSearch need to be algorithms that are approved by FIPS as listed under [this link](https://csrc.nist.gov/Projects/cryptographic-module-validation-program/sp-800-140-series-supplemental-information/sp800-140c).
- It is recommended that all used security providers for algorithms are also software modules which have undergone FIPS validation and are listed under [this link](https://csrc.nist.gov/projects/cryptographic-module-validation-program/validated-modules/search?SearchMode=Basic&CertificateStatus=Active&ValidationYear=0).
- Enabling FIPS mode at runtime, rather than at compile time prevents having to maintain two distinct code versions and is therefore the preferred solution.

To accomplish this, it is essential to utilize the Java Cryptographic Architecture (JCA), Java Cryptographic Extension (JCE), and Cryptographic Service Providers (CSP) in lieu of direct, hard coded implementations of specific security providers.

## Implementation Summary

The following steps are proposed to be implemented sequentially and iteratively to reach the goal of OpenSearch FIPS compliance:

1.	Verify currently used algorithms and if needed define FIPS compliant alternatives
2.	Define the default security providers to be used
3.	Adapt the policy and security files
4.	Test the functionality by running OpenSearch in a FIPS compliant environment. If something fails, go back to step 1 and repeat
5.	Implement a central security provider
6.	Adapt the Gradle file(s)
7.	Adapt and create new tests

## Detailed description

1. **Verify currently used algorithms and if needed define FIPS compliant alternatives.**
The initial step involves identifying all instances within the OpenSearch (core and plugins) code where cryptographic algorithms are utilized.
A big part of this work has already been done by searching the codebase of OpenSearch and OpenSearch Security Plugin for uses of engine classes, these classes are part of the JCA / JSE and define algorithms for specific cryptographic tasks like hashing, encryption, random number generation and so on. The result of this search can be shared on request.
Some security providers may have been hardcoded, requiring more detailed analysis for identification. Hence, our strategy involves pinpointing them during step 4, given the challenge of examining the entire codebase. Alternatively, we encourage the community to document any known instances of hardcoded security providers.
After identifying all the algorithms in use, it is essential to check this list against the FIPS-defined list of approved algorithms as defined [here](https://csrc.nist.gov/Projects/cryptographic-module-validation-program/sp-800-140-series-supplemental-information/sp800-140c)
If the algorithm is absent, an alternative must be specified when OpenSearch operates in FIPS mode.
Following this step, a comprehensive list will emerge, detailing the cryptographic algorithms utilized in both normal and FIPS modes, which will be an important piece of documentation for future development and to define guidelines for security.

2. **Define the default security providers.**
Having identified all algorithms that will be used in FIPS mode, the next step is to define a default security provider for each algorithm, therefore, it is recommended to use a software module which has undergone FIPS verification and is listed in the [Cryptographic Module Validation Program (CMVP) of FIPS](https://csrc.nist.gov/projects/cryptographic-module-validation-program/validated-modules/search?SearchMode=Basic&CertificateStatus=Active&ValidationYear=0) \
As a result of this step, a list of ordered security providers will be defined for use in normal mode and in FIPS mode, both of which can define default configuration options OpenSearch will be shipped with.
If users instead desire to use their own security providers, they will be able to do so because the list of algorithms defined in step 1 will allow them to verify that their custom providers cover all required algorithms.

3. **Adapt the policy and security files.**
The third step is to update the policy and security files of the project, to be able to use the defined security providers in step 2 and allow for all the algorithms defined in step 1 to be run correctly.
As a result, base security and policy files will be created for use in either normal or FIPS mode, which will enable users to employ these files as parameters should they need to integrate alternative providers.

4. **Test the functionality by running OpenSearch in a FIPS compliant environment.**
Before implementing significant changes or modifications in the code, it is necessary to test the configuration environment as defined in steps 1 to 3. For this test the following steps are required:
- Configure a FIPS compliant environment in FIPS mode, including operating system and JRE
  - Install the default security providers as defined in step 2
  -	Configure the JRE with the FIPS enabled security and policy files as per step 3
- Replace non-FIPS compliant algorithms in the code with their FIPS compliant counterparts directly, while minimizing structural code changes
  -	All places in the code that need to be changed are already identified in step 1.
As a result of steps 1 to 3 OpenSearch should be able to run (and end to end tested) on a FIPS compliant environment, and thus assert that the algorithms, providers, policy and security configurations are implemented correctly. If non FIPS compliant providers were hardcoded or omitted, the test will fail and these cases will be identified, requiring the need to return to steps 1 to 3. For these cases, it will be required that FIPS compliant algorithms are implemented. As a result of this step, a FIPS-compliant version of OpenSearch will be operational.

5. **Implement a central security provider.**
To avoid handling the specifics of FIPS and non FIPS compliant algorithms across the code base, a central security provider should be implemented.
This provider should be able to offer any required cryptographic functionality declaratively, like hashing functions, symmetric and asymmetric encryption and decryption, random number generators, etc.
The security provider should not only allow the definition of a specific algorithm, but also, based on the intended OpenSearch configuration, automatically define the appropriate algorithm to comply with FIPS if necessary or to keep the existing non FIPS algorithms.\
Having a centralized security provider in place would also ease the development towards future security standards.\
As a part of this task, all engine classes or direct security providers should be replaced with the central security manager. All places in the code where this would need to be changed are already defined in step 1.
Following this step, OpenSearch will be capable of operating in either FIPS or non-FIPS mode, accommodating any security provider chosen by the user. Additionally, it will provide a stable set of default values to ensure the software runs smoothly in any mode, minimizing user inconvenience.

6. **Adapt Gradle file(s).**
It is necessary to adapt the Gradle files to build the project correctly.

7. **Adapt Tests.**
All the tests currently only support the non FIPS mode, if FIPS mode is enabled the tests are skipped.
All relevant tests have to be adapted to allow FIPS and non FIPS mode. New tests must be implemented.

beanuwave:
## Description

**Category:** Enhancement, New feature, Bug fix

This branch makes the Security plugin run correctly under FIPS approved-only mode (BCFIPS as the sole crypto provider). Every change traces to one of: an algorithm/format/usage that FIPS disallows (whether or not BCFIPS rejects it at runtime), a weakness surfaced during the audit, or build/packaging plumbing to make FIPS the always-available baseline.

At a glance:

- **Single source of truth:** `FipsMode.isEnabled()` (env `OPENSEARCH_FIPS_MODE=true`) replaces `CryptoServicesRegistrar.isInApprovedOnlyMode()` everywhere. Intent is decoupled from provider state and cross-checked at startup.
- **Build & FIPS activation:** BCFIPS is always built (`compileOnly`, shipped by core); the compile-time `FipsBuildParams` fork is gone. FIPS engages purely through the `java.security` the JVM loads (no provider is registered in code) - the runtime launcher merges in `fips_java.security` on `OPENSEARCH_FIPS_MODE=true`, while tests swap it wholesale (`-D...properties==<file>`).
- **OBO/JWT tokens:** encryption rewritten from default-mode AES (ECB) -> AES-GCM + HKDF (fixes a FIPS/NIST violation and a deterministic-crypto weakness), plus an entropy floor, lazy fail-closed init, an optional BCFKS keystore for the keys, and secret redaction.
- **Keystores / TLS:** JKS/PKCS12 out, BCFKS in (PKCS11 opt-in, signed via SunJSSE); `PemKeyReader` rebuilt on BouncyCastle; TLSv1.1 dropped in FIPS mode.
- **LDAP:** SNI re-implemented for BCTLS, and ldaptive's non-FIPS internal PKCS12 key-copy bypassed.
- **Auth hardening:** Kerberos cleanup, timing-safe dummy hash, a 14-char password floor in FIPS, approved-DRBG randomness, key-material zeroing.
- **CLI:** `securityadmin` accepts BCFKS/PKCS11 and now launches via core's `opensearch-cli`.

### OBO / JWT tokens

The OBO key path was audited end-to-end; the encryption rewrite is the FIPS trigger, the rest are weaknesses found alongside it.

- **Encryption rewrite.** `EncryptionDecryptionUtil`: built its cipher with the bare `Cipher.getInstance("AES")` and a key forced to 16 bytes via `Arrays.copyOf` (silently truncating or zero-padding) -> now AES-256-GCM (random 12-byte IV, 128-bit tag), key derived via HKDF-SHA256. A bare `"AES"` transformation resolves to the JCE provider's default mode - ECB - which for data confidentiality violates NIST SP 800-38A<sup>[2](#references)</sup> (ECB is approved only for key wrapping, SP 800-38F<sup>[3](#references)</sup>). The mode is never spelled out or configurable, and BC FIPS permits the ECB primitive at runtime - so this had to be caught by review, not the provider.
- **Entropy floor (SP 800-133r2<sup>[5](#references)</sup>).** HKDF can't create entropy, so a short `encryption_key` would produce a nominal AES-256 key with sub-112-bit strength. A FIPS guard rejects input keying material < 32 bytes and zeroes the IKM after derivation.
- **Signing-key length check.** Validates the *decoded* key bytes; measuring the Base64 string length would over-count by ~4/3 and let a 384-bit key pass a 512-bit gate.
- **Lazy, fail-closed init.** `OnBehalfOfAuthenticator` initializes lazily and atomically: the constructor is inert, the first token request triggers init, a failure logs once and declines - a bad key can't throw out of the constructor, propagate through config reload, and retry forever. Mirrors `ApiTokenAuthenticator`.
- **Keystore support.** The OBO signing and encryption keys can now be loaded from a BCFKS keystore (`KeyUtils.loadKeyFromKeystore`) instead of inline config, keeping the secrets out of cluster state. Relative `*_keystore_path` values resolve against the node config dir (consistent with other security file settings). This is an SP 800-57<sup>[6](#references)</sup> key-at-rest hardening (protecting keying material / limiting exposure), not a FIPS 140-3<sup>[1](#references)</sup> requirement.
- **Secret redaction (CWE-532).** `OnBehalfOfSettings.toString()` redacts the `signing_key`/`encryption_key` values so they no longer leak into logs.

### Keystores / TLS

- `PemKeyReader` rewritten onto BouncyCastle (`PEMParser` / `JcaPEMKeyConverter` / PKCS8 decryptor) instead of raw JCE; adds BCFKS + PKCS11 and store-type auto-detection.
- `SSLConfigConstants`: both defaults are now FIPS-conditional - default store type is forced to BCFKS in FIPS, and `ALLOWED_SSL_PROTOCOLS` drops TLSv1.1 in FIPS. Both LDAP backends now reuse `ALLOWED_SSL_PROTOCOLS` for their default `enabled_ssl_protocols`.
- **PKCS#11 keys are signed via SunJSSE, not BCJSSE.** A PKCS#11/HSM private key is non-exportable and BCJSSE can't sign with it (`no encoding for key`); SunJSSE delegates the handshake signature to the key's own provider (SunPKCS11). So when the keystore is PKCS#11, `SslConfiguration` builds the `SSLContext` against SunJSSE (`SslContextBuilder.sslContextProvider(SunJSSE)`).
- **The BC FIPS provider is declared, not instantiated.** `main` self-registered it at plugin load (`OpenSearchSecuritySSLPlugin.tryAddSecurityProvider()` -> `Security.addProvider(new BouncyCastleFipsProvider())`); that method is removed. Providers now come solely from the active `java.security` file (JCA lazy-loads them), so FIPS vs non-FIPS is a launch-time provider swap (BCJSSE vs SunJSSE) with no code branch - the security files are a core/distribution concern.

### LDAP

- SNI for BCTLS - two sequential (not duplicate) concerns: `SNISettingTLSSocketFactory` sets the ClientHello SNI before the handshake so a multi-cert server serves the right cert; `HostnameVerifyingTrustManager` checks the returned cert after. For IP targets the SNI factory early-returns (no SNI) and the trust manager is the only hostname check; for DNS it sets SNI + endpointIdentification (the trust-manager hostname check is then redundant). `HostnameAwareConnectionFactory` threads the target hostname through so SNI works.
- Avoiding ldaptive's internal PKCS12 copy (both authentication and authorization paths) - ldaptive's `create*CredentialConfig` with key aliases routes through `KeyStoreSSLContextInitializer.getKeyManagers()`, which copies the private key into a fresh in-memory PKCS12 store; SunJCE protects it with `PBEWithHmacSHA256AndAES_256`, not available in FIPS. Fix: build BCFKS keystores from PEM via `PemKeyReader.toTruststore/toKeystore` and pass null key aliases, so `kmf.init(keystore, password)` is called directly and the PKCS12-copy branch is bypassed. (`LDAPAuthorizationBackend` previously used `createX509CredentialConfig`, which hit the same path unconditionally.)
- Extracted the default backend's `Java9CL` into a shared `SocketFactoryClassLoader` (resolves the JNDI socket-factory classes by name) and set it on `ldap2`'s JNDI provider config, so `PrivilegedProvider`'s thread-context swap resolves the socket factory on reconnect. Fixes an ldap2 LDAPS reconnect `ClassNotFoundException`.

### Auth hardening (found mid-audit)

- `HTTPSpnegoAuthenticator`: no longer mutates global `System.setProperty` debug flags; stops logging the acceptor principal; proper `LoginContext.logout()` + decoded-header zeroing in `finally`.
- `InternalAuthenticationBackend` + `PasswordHasher.getDummyHash()`: not-found timing path now uses the configured hasher (closes user-enumeration side-channel under PBKDF2).
- **Password-length floor.** PBKDF2 keys are derived from the password itself, and BC FIPS rejects key material under 112 bits (< 14 ASCII chars) at hashing time. `PasswordValidator.FIPS_MIN_PASSWORD_LENGTH` (14) now anchors both ends: FIPS raises an unset `restapi.password_min_length` to 14, and startup rejects a lower explicit value - otherwise the REST API accepts passwords the hasher then refuses.
- **Randomness from core.** `Randomness.createSecure()` replaces `new SecureRandom()` for OBO encryption, api-tokens, and user passwords - it resolves to the approved SP 800-90A<sup>[4](#references)</sup> DRBG in FIPS. `UserService` generates 20-27 chars in FIPS (>=119 bits over the 62-char alphabet), 8-15 otherwise; `char[]` zeroed in `finally`.

### CLI

- `SecurityAdmin`: accepts BCFKS/PKCS11; PKCS#11 PIN prompt. Launcher scripts now delegate to core's shared `opensearch-cli`. Standalone bundle ships BCFIPS jars under `deps/`.
- `SecurityAdmin.buildPkcs11SslContext` routes PKCS#11 keys through SunJSSE client-side (same as the server TLS layer - see Keystores/TLS).

> [!NOTE]
> **Reviewer call-outs**
>
> 1. **OBO token wire-format break** - GCM-encrypted tokens are not interchangeable with old AES-ECB tokens across the upgrade boundary. Benign in practice (OBO TTL < 10 min, self-heals shortly after rollout completes), but worth a release note for hot/rolling deployments.
> 2. **`securityadmin` now launches via core's `opensearch-cli`** - this targets the in-distribution path; the standalone bundle is no longer self-launching and is effectively deprecated.
> 3. **Test strategy diverges from core's convention** - this branch runs one dual-mode suite (skip incompatible cases at runtime), whereas the core project forks the suite: a FIPS test class extends the non-FIPS one.
> 4. **Dropped the auto `System.setProperty(disableEndpointIdentification, true)` from the default `ldap` backend** - now warn-only. Two reasons: (a) it aligns both backends on the same logic (`ldap2` never set it); (b) mutating a global JVM property from application code is bad practice - it's process-wide and load-order-dependent, so one auth-domain's `verify_hostnames: false` silently reconfigured hostname checking for the *whole* JVM. Operators who need it must now set the `-D` flag deliberately.
> 5. **Why are `verify_hostnames` and `trust_all` coupled to the same verifier?** - `verifyHostnames = !trustAll && <setting>`, so `trust_all: true` forces `AllowAnyHostnameVerifier` on top of `AllowAnyTrustManager`. Chain validation and hostname matching are orthogonal concerns; collapsing them onto one code path means you can't relax one without the other and hides which layer a config change actually touches. Worth untangling so each parameter maps to exactly one verification layer.

## Core / distribution follow-ups

Not plugin changes - each fix belongs in core or the distribution, tracked here so it can be routed:

- **`bctls-fips` has no socket `connect` grant.** Core's base `security.policy` grants `bc-fips`/`bcpkix-fips` but not `bctls-fips`, so under the agent all outbound BCJSSE TLS (LDAPS, audit sinks, remote reindex/snapshot over https) is denied - and plugin policies can't grant a core `lib/` jar. Verified against a live LDAPS-over-FIPS cluster.
- **JUL->log4j bridge not active for BC FIPS at rollout.** `LogConfigurator` sets `java.util.logging.manager` at *runtime*, too late for the BouncyCastle FIPS JSSE provider (which initialises JUL during bootstrap); its per-handshake INFO traces then leak to `System.err` as `[WARN][stderr]` spam. Fix in the distribution: set `-Djava.util.logging.manager=org.apache.logging.log4j.jul.LogManager` as a launch-time `-D` in `jvm.options`, and default `org.bouncycastle.jsse` to `warn` in the shipped `log4j2.properties`.
- **Core's `org.opensearch.fips.FipsMode` still detects FIPS via `CryptoServicesRegistrar.isInApprovedOnlyMode()`** - the provider-state probe this branch replaced with the env var; it should follow.
- **`BCFipsEntropyDaemonFilter` should move into core's test framework.** The local thread-filter accommodation for the "BC FIPS Entropy Daemon" thread belongs in core's `BouncyCastleThreadFilter`, not this plugin's test sources.

## Additional notes

- JKS can't hold `SecretKey` entries (`engineSetKeyEntry` requires `PrivateKey`) - relevant to the JWT-signing-key path; BCFKS is the only FIPS-approved store that holds secret keys.
- Provider list is a complete override; `SunJGSS` is deliberately retained for Kerberos/SPNEGO.
- Operational: keystore passwords must be >=14 chars (BCFIPS); user passwords likewise (the 112-bit PBKDF2 floor, now enforced - see *Auth hardening*); the "BC FIPS Entropy Daemon" thread required a test thread-filter accommodation.
- SAML is scoped out - both SAML stacks it uses (OneLogin java-saml and OpenSAML/Shibboleth) are not FIPS-compliant (tests skip in FIPS mode).
- HTTP/3 exclusion is a build-level constraint, not a categorical ban - a FIPS-certified BoringSSL substituted at the OS level re-enables it.
- The SNI plumbing (`SNISettingTLSSocketFactory` + the `SniAwareConnection` decorator + hostname ThreadLocal + `HostnameAwareConnectionFactory`) works around the JNDI LDAP provider resolving hostnames to IPs before socket creation ([bcgit/bc-java#460](https://github.com/bcgit/bc-java/issues/460)); ldaptive 2.x's native (Netty) transport opens sockets with the real hostname, which would let this entire stack be deleted (out of scope here).

## Testing

The suite runs in non-FIPS mode by default. To exercise the FIPS code paths, set the environment variable before invoking Gradle:

```bash
OPENSEARCH_FIPS_MODE=true ./gradlew test integrationTest
```

When set, the build swaps in the FIPS `java.security` policy (BCFIPS-only providers), enables `-Dorg.bouncycastle.fips.approved_only=true`, and points the JVM at the BCFKS truststore. FIPS-incompatible tests (BCrypt, Argon2, SAML, SSLv3, JKS/PKCS12, weak/short passwords) are auto-skipped via JUnit assumptions. Static bcrypt fixtures and their short demo passwords are rewritten to PBKDF2 and padded past the 14-char floor by `FipsHashAdapter` (a no-op outside FIPS), and a few timing-sensitive integ tests scale down under FIPS, where PBKDF2 logins and BCTLS handshakes are markedly slower.

For a running cluster, select the FIPS-approved password hasher in `opensearch.yml` (BCrypt/Argon2 are not available in approved-only mode):

```yaml
plugins.security.password.hashing.algorithm: pbkdf2
```

The demo hashes in `config/opensearch-security/internal_users.yml` are BCrypt, which won't verify under PBKDF2 - regenerate the hash for each test account (e.g. with `tools/hash.sh`) and replace it before applying the security config.

<details>
<summary>Test securityadmin.sh with BCFKS + PKCS#11 keystores (SoftHSM)</summary>

Exercises a token-resident node TLS key (signed via SunJSSE) and `securityadmin` authenticating with a PKCS#11 client key. This example is FIPS-specific, but adjusts easily to non-FIPS by registering a SunPKCS11 provider via `OPENSEARCH_JAVA_OPTS="-Djava.security.properties=$OPENSEARCH_HOME/config/java.security"` instead of `OPENSEARCH_FIPS_MODE=true`. All paths below are relative to `$OPENSEARCH_HOME`.

```bash
# 1. Set up a test cluster.
...
cd $OPENSEARCH_HOME
sh bin/opensearch-keystore create --password
sh plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s

# 2. Init token - the --pin becomes the keystore/truststore password.
softhsm2-util --init-token \
  --free \
  --label opensearch \
  --so-pin 4321 \
  --pin 1234

# 3. Register the provider in config/fips_java.security:
      security.provider.<n>=SunPKCS11 /path/to/config/softhsm-pkcs11.cfg
#    with softhsm-pkcs11.cfg in the same dir:
      name = SoftHSM
      library = /usr/lib/softhsm/libsofthsm2.so
      slotListIndex = 0

# 4. Import node + admin keys WITH chains (PKCS#12 -> token).
#    Repeat the keytool step for kirk (swap -name / -srcalias / -destalias).
openssl pkcs12 -export \
  -inkey config/esnode-key.pem \
  -in config/esnode.pem \
  -certfile config/root-ca.pem \
  -name esnode-cert \
  -out /tmp/esnode.p12 \
  -passout pass:1234

jdk/bin/keytool \
  -importkeystore \
  -srckeystore /tmp/esnode.p12 \
  -srcstoretype PKCS12 \
  -srcstorepass 1234 \
  -srcalias esnode-cert \
  -destkeystore NONE \
  -deststoretype PKCS11 \
  -deststorepass 1234 \
  -destalias esnode-cert \
  -addprovider SunPKCS11 \
  -providerarg config/softhsm-pkcs11.cfg

# 5. Trust anchor -> BCFKS file.
jdk/bin/keytool -importcert -noprompt \
  -alias root-ca \
  -file config/root-ca.pem \
  -keystore config/root-ca.bcfks \
  -storetype BCFKS \
  -storepass changeit \
  -providerClass org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider \
  -providerPath lib/bc-fips-2.1.2.jar

# 6. In opensearch.yml: comment the demo *.pem*_filepath lines, keep admin_dn: CN=kirk,...,
#    and add for both transport and http:

# --- Node identity from the PKCS#11 token ---
plugins.security.ssl.transport.keystore_type: PKCS11
plugins.security.ssl.transport.keystore_alias: esnode-cert
plugins.security.ssl.transport.keystore_password: "1234"      # SoftHSM PIN
plugins.security.ssl.http.keystore_type: PKCS11
plugins.security.ssl.http.keystore_alias: esnode-cert
plugins.security.ssl.http.keystore_password: "1234"

# --- Trust anchor from a BCFKS file ---
plugins.security.ssl.transport.truststore_type: BCFKS
plugins.security.ssl.transport.truststore_filepath: root-ca.bcfks
plugins.security.ssl.transport.truststore_password: "changeit"
plugins.security.ssl.http.truststore_type: BCFKS
plugins.security.ssl.http.truststore_filepath: root-ca.bcfks
plugins.security.ssl.http.truststore_password: "changeit"

# 7. Run the cluster and apply security config with the PKCS#11 admin key.
#    Pass = 'Connected as "CN=kirk,..."' followed by 'Done with success'.
OPENSEARCH_FIPS_MODE=true sh plugins/opensearch-security/tools/securityadmin.sh \
  -cd config/opensearch-security/ \
  -icl \
  -nhnv \
  -cacert config/root-ca.pem \
  -kst PKCS11 \
  -kspass 1234 \
  -ksalias kirk
```

</details>

<details>
<summary>Test LDAP authentication over LDAPS (SNI, hostname verification, mTLS)</summary>

Self-contained manual tests for the LDAP TLS changes: SNI / hostname verification, mutual TLS, and the TLS protocol floor. **Authentication only** - these changes don't touch authz code, so role resolution is out of scope (verify that against a real directory).

Run them against any LDAPS directory that supports **mTLS** (a client cert is required). The walkthrough uses a local UnboundID in-memory stand-in only because it's repeatable and trivial to set up - a convenience, not a requirement; substitute your own server anywhere it appears. Its setup lives in [`LDAP_UNBOUNDID_STANDIN_GUIDE.md`](https://github.com/sternadsoftware/security/commit/0d35598261643192a567a36b9147e6de6cadab24).

TLS material is the OpenSearch install's own demo certs in `$OPENSEARCH_HOME/config/` - `esnode` (server, SAN includes `localhost`), `root-ca.pem` (trust anchor), `kirk` (client). Run the node with `OPENSEARCH_FIPS_MODE=true` (omit for non-FIPS; the only observable difference is the TLS protocol floor).

> **FIPS agent-build gotchas (both are *Core / distribution follow-ups*):** (1) core's base `security.policy` grants `bc-fips`/`bcpkix-fips` but not `bctls-fips`, so every LDAPS bind is *denied* under the agent; (2) the JUL->log4j bridge isn't active for BC FIPS JSSE at handshake time, so each bind's INFO traces leak to stderr as `[WARN][stderr]` spam (noise, not a failure).

```bash
cd $OPENSEARCH_HOME
export LDAP_USER="testuser"; export LDAP_PASS="testpassword"   # cn=Test User,ou=people,o=TEST

# Optional trace logging for the "what to look for" lines (config/log4j2.properties):
#   logger.ldap.name=org.opensearch.security.auth.ldap
#   logger.ldap.level=trace
#   logger.ldap2.name=org.opensearch.security.auth.ldap2
#   logger.ldap2.level=trace
#   logger.ldaptive.name=org.ldaptive
#   logger.ldaptive.level=debug

# === Baseline config (authc.ldap.authentication_backend.config; o=TEST, no authz block) ======
# Edit this first, then apply below. Bare *_filepath resolve against config/; kirk + root-ca are the
# shipped demo certs, so as shipped this IS scenario 1a / 2a -> 200. Run every scenario once per
# backend (flip the type: line).
#   type: ldap                  # DEFAULT | LDAP2: org.opensearch.security.auth.ldap2.LDAPAuthenticationBackend2
#   config:
#     enable_ssl: true
#     enable_ssl_client_auth: true          # mTLS - client cert required
#     verify_hostnames: true
#     hosts: [localhost:8636]
#     pemtrustedcas_filepath: root-ca.pem
#     pemcert_filepath: kirk.pem
#     pemkey_filepath: kirk-key.pem
#     bind_dn: "cn=opensearch-bind,ou=people,o=TEST"
#     password: "bindpassword"
#     userbase: "ou=people,o=TEST"
#     usersearch: '(uid={0})'
#     username_attribute: uid

# === Apply / authenticate (verify loop - re-run after each config.yml edit) ==
# apply: push the authc.ldap block (-t config, live; TLS-material/host edits need a node restart).
sh ./plugins/opensearch-security/tools/securityadmin.sh \
  -f ./config/opensearch-security/config.yml \
  -t config \
  -icl \
  -nhnv \
  -cacert config/root-ca.pem \
  -cert config/kirk.pem \
  -key config/kirk-key.pem \
  -h localhost \
  -p 9200

# authenticate: 200 + user_name=testuser (backend_roles empty - no authz) = LDAPS + mTLS bind OK.
curl -sk -u "$LDAP_USER:$LDAP_PASS" https://localhost:9200/_plugins/_security/authinfo?pretty

# What to look for (baseline 200):
#   Configuring SNI for hostname: localhost ...            # SNI server_name set (the fix)
#   checkServerTrusted ... succeeded                       # esnode chains to root-ca
#   verifyDNS found hostname match: localhost              # hostname layer 1 pass
#   Opened a connection, total count is now 1              # DEFAULT | LDAP2: Authenticated username testuser
```

**Test matrix.** Run every scenario in all four cells - flip the backend on `type:`; for FIPS set `OPENSEARCH_FIPS_MODE=true` (launcher loads `fips_java.security` -> BCJSSE), for non-FIPS set `OPENSEARCH_JAVA_OPTS="-Djava.security.properties=$OPENSEARCH_HOME/config/java.security"` (BCFIPS stays declared - the FIPS installer converts the node stores to BCFKS - but TLS runs on SunJSSE). Launch-time provider swap, no code path (see *Keystores / TLS*). Outcomes are identical; only the protocol floor (`[TLSv1.3, TLSv1.2]` FIPS vs `+ TLSv1.1` non-FIPS) and the provider differ, so the `Prov*`/`TlsFatalAlert` class names in the excerpts are BCJSSE-only.

| | DEFAULT `ldap` | LDAP2 (`...ldap2.LDAPAuthenticationBackend2`) |
|---|---|---|
| **FIPS** | yes | yes |
| **non-FIPS** | yes | yes |

**Scenario 1 - hostname verification.** Same trusted `esnode` cert throughout; 1b-1d dial a name not in its SAN (`echo "127.0.0.1  ldap-wrong.example.com" | sudo tee -a /etc/hosts`, then set `hosts: [ldap-wrong.example.com:8636]`), so the only thing that can object is one of the two hostname guards: (1) ldaptive's verifier (`verify_hostnames`), (2) JNDI endpoint-id (`-Dcom.sun.jndi.ldap.object.disableEndpointIdentification=true` in `config/jvm.options`; the plugin no longer sets it, only warns). Chain trust is valid throughout, so this isolates *hostname* checking - the untrusted-cert case is 2c. Apply + restart per row.

| | `verify_hostnames` | JNDI endpoint-id | Result / what it proves |
|---|---|---|---|
| **1a** | `true`  | on  | correct name -> 200 (baseline; SNI fix works) |
| **1b** | `true`  | on  | rejected, layer 1 - ldaptive `DefaultHostnameVerifier` |
| **1c** | `false` | on  | rejected, layer 2 - JNDI/BC endpoint-id (`verify_hostnames: false` alone isn't enough) |
| **1d** | `false` | off | accepted (200) - hostname unenforced only when both guards are off |

```text
1b  HostnameVerifyingTrustManager ... hostnames=[ldap-wrong.example.com] failed         # layer 1 (ldaptive)
    CertificateException: Hostname '[ldap-wrong.example.com]' does not match 'CN=node-0.example.com...'
    (for ldap2 the reject fires inside SniAwareConnection.open() = no fail-open)
1c  AllowAnyHostnameVerifier ... succeeded                                              # layer 1 off
    CertificateException: No subject alternative name found matching domain name ldap-wrong.example.com  # layer 2
        at org.bouncycastle.jsse.provider.ProvX509TrustManager.checkEndpointID(...)
1d  (no hostname reject) -> Opened a connection / Authenticated username testuser
```

Cleanup: remove the `/etc/hosts` line + the `jvm.options` flag, restore `hosts:` + `verify_hostnames: true`. Never set `disableEndpointIdentification=true` in production (it's process-wide).

**Scenario 2 - mTLS client authentication.** Vary only the client cert / trust anchor; apply + restart per row, and restore `pemtrustedcas_filepath: root-ca.pem` after 2c. First generate the two "bad" credentials once - both self-signed, so neither chains to the demo `root-ca` - into `config/`:

```bash
cd $OPENSEARCH_HOME/config
# 2b: untrusted client cert + key (self-signed; does NOT chain to root-ca)
openssl req -x509 -newkey rsa:2048 -nodes -days 30 -subj "/CN=untrusted-client" \
  -keyout untrusted-client.key -out untrusted-client.pem
# 2c: untrusted CA - a trust anchor that did NOT sign esnode
openssl req -x509 -newkey rsa:2048 -nodes -days 30 -subj "/CN=untrusted-ca" \
  -keyout untrusted-ca.key -out untrusted-ca.pem
```

| | change (in `config.yml`) | Result / what it proves |
|---|---|---|
| **2a** | `pemcert/pemkey_filepath: kirk.*` (baseline) | accepted - PEM client key loads + signs the handshake under BC |
| **2b** | `pemcert/pemkey_filepath: untrusted-client.*` | rejected - BC withholds the client cert -> server aborts mandatory mTLS |
| **2c** | `pemtrustedcas_filepath: untrusted-ca.pem` | fails - server cert no longer chains to the trust anchor |

```text
2b  checkServerTrusted ... succeeded                        # server side fine
    received fatal(2) certificate_required(116) alert       # client cert withheld
    must NOT log: ClassNotFoundException ...SNISettingTLSSocketFactory  (ldap2 reconnect bug, fixed here)
2c  checkServerTrusted ... failed
    CertPathBuilderException: No issuer certificate for certificate in certification path found.
    TlsFatalAlert: certificate_unknown(46)  ->  Authentication finally failed
```

</details>

<details>
<summary>Test OnBehalfOf (OBO) token</summary>

Exercises OBO token issuance and verification against an already-running cluster, in three modes: keys inline in the dynamic config (Scenario A), held in a BCFKS keystore out of cluster state (Scenario B), or in a PKCS#12 keystore for non-FIPS builds (Scenario C). No restart needed - `-t config` pushes only the dynamic `on_behalf_of` block, picked up live. Pick ONE scenario, edit `config.yml`, then run the apply / issue / use steps. All paths are relative to `$OPENSEARCH_HOME`.

```bash
cd $OPENSEARCH_HOME
export ADMIN_AUTH="admin:<admin-password>"

# === Apply / issue / use (the verify loop - run after editing config.yml) ====
# apply: push the dynamic config (-t config = on_behalf_of block only, live reload).
sh ./plugins/opensearch-security/tools/securityadmin.sh \
  -f ./config/opensearch-security/config.yml \
  -t config \
  -icl \
  -nhnv \
  -cacert config/root-ca.pem \
  -cert config/kirk.pem \
  -key config/kirk-key.pem \
  -h localhost \
  -p 9200

# issue: generate a token.
export OBO_TOKEN=$(curl -sk \
  -u "$ADMIN_AUTH" \
  -X POST \
  -H 'Content-Type: application/json' \
  https://localhost:9200/_plugins/_security/api/generateonbehalfoftoken \
  -d '{
        "description": "obo test",
        "service": "test-service",
        "durationSeconds": "300"
      }' | jq -r '.authenticationToken')
echo "$OBO_TOKEN"

# use: a populated user_name / roles proves the verify side loaded the key,
#      checked the signature, and decrypted the roles.
curl -sk \
  -H "Authorization: Bearer $OBO_TOKEN" \
  https://localhost:9200/_plugins/_security/authinfo?pretty

# === Scenario A - inline keys ================================================
# signing_key >= 512 bits (64 bytes) for HS512; encryption_key >= 256 bits
# (32 bytes) for the FIPS IKM floor.
export OBO_SIGNING_KEY=$(openssl rand 64 | base64 -w0)
export OBO_ENCRYPTION_KEY=$(openssl rand 32 | base64 -w0)

# Put under config.dynamic.on_behalf_of in config/opensearch-security/config.yml:
#   on_behalf_of:
#     enabled: true
#     signing_key: "<value of $OBO_SIGNING_KEY>"
#     encryption_key: "<value of $OBO_ENCRYPTION_KEY>"
#
# Negative checks: a too-short encryption_key (e.g. ZW5jcnlwdGlvbktleQ==, 13 bytes)
# is declined in FIPS mode ("encryption_key is not strong enough for FIPS mode");
# tampering with the token's last segment makes the `use` step return no credentials.

# === Scenario B - BCFKS keystore =========
keytool \
  -genseckey \
  -alias obo-signing \
  -keyalg HmacSHA512 \
  -keysize 512 \
  -storetype BCFKS \
  -providername BCFIPS \
  -keystore config/obo.bcfks \
  -storepass kspass \
  -keypass keypass \
  -providerClass org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider \
  -providerPath lib/bc-fips-2.1.2.jar

keytool \
  -genseckey \
  -alias obo-enc \
  -keyalg AES \
  -keysize 256 \
  -storetype BCFKS \
  -providername BCFIPS \
  -keystore config/obo.bcfks \
  -storepass kspass \
  -keypass keypass \
  -providerClass org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider \
  -providerPath lib/bc-fips-2.1.2.jar

# Reference the keystore in config.yml (no inline keys). <key>_keystore_path is
# config-dir-relative.
#   on_behalf_of:
#     enabled: true
#     signing_key_keystore_path: "obo.bcfks"
#     signing_key_keystore_type: "BCFKS"
#     signing_key_keystore_alias: "obo-signing"
#     signing_key_keystore_password: "kspass"
#     signing_key_keystore_key_password: "keypass"
#     encryption_key_keystore_path: "obo.bcfks"
#     encryption_key_keystore_type: "BCFKS"
#     encryption_key_keystore_alias: "obo-enc"
#     encryption_key_keystore_password: "kspass"
#     encryption_key_keystore_key_password: "keypass"
#
# Same result as Scenario A, but no key material in the config index. GET
# /_plugins/_security/api/securityconfig shows only keystore references.

# === Scenario C - PKCS#12 keystore (NON-FIPS builds only) ====================
# A FIPS build rejects PKCS#12 (no FIPS-approved PKCS#12 in BC FIPS). PKCS#12 has
# no per-entry passwords, so use one value for -storepass and -keypass.
keytool \
  -genseckey \
  -alias obo-signing \
  -keyalg HmacSHA512 \
  -keysize 512 \
  -storetype PKCS12 \
  -keystore config/obo.p12 \
  -storepass kspass \
  -keypass kspass

keytool \
  -genseckey \
  -alias obo-enc \
  -keyalg AES \
  -keysize 256 \
  -storetype PKCS12 \
  -keystore config/obo.p12 \
  -storepass kspass \
  -keypass kspass

# Reference it in config.yml as in Scenario B with the PKCS#12 path/type (the
# loader falls back to the keystore password when _keystore_key_password is absent).
#   on_behalf_of:
#     enabled: true
#     signing_key_keystore_path: "obo.p12"
#     signing_key_keystore_type: "PKCS12"
#     signing_key_keystore_alias: "obo-signing"
#     signing_key_keystore_password: "kspass"
#     encryption_key_keystore_path: "obo.p12"
#     encryption_key_keystore_type: "PKCS12"
#     encryption_key_keystore_alias: "obo-enc"
#     encryption_key_keystore_password: "kspass"
```

</details>

## Issues Resolved

Resolves [RFC](https://github.com/opensearch-project/security/issues/4254)

Related to the series of FIPS PRs in the security plugin:
- https://github.com/opensearch-project/security/pull/6126
- https://github.com/opensearch-project/security/pull/6059
- https://github.com/opensearch-project/security/pull/6087
- https://github.com/opensearch-project/security/pull/6042
- https://github.com/opensearch-project/security/pull/5952
- https://github.com/opensearch-project/security/pull/6004
- https://github.com/opensearch-project/security/pull/5439
- https://github.com/opensearch-project/security/pull/5404
- https://github.com/opensearch-project/security/pull/4275
- https://github.com/opensearch-project/security/pull/4524

## Check List

## References

1. [FIPS 140-3 - Security Requirements for Cryptographic Modules](https://csrc.nist.gov/pubs/fips/140-3/final)
2. [NIST SP 800-38A - Block Cipher Modes of Operation (ECB/CBC/CFB/OFB/CTR)](https://csrc.nist.gov/pubs/sp/800/38/a/final)
3. [NIST SP 800-38F - Methods for Key Wrapping](https://csrc.nist.gov/pubs/sp/800/38/f/final)
4. [NIST SP 800-90A Rev. 1 - Random Number Generation Using DRBGs](https://csrc.nist.gov/pubs/sp/800/90/a/r1/final)
5. [NIST SP 800-133 Rev. 2 - Recommendation for Cryptographic Key Generation](https://csrc.nist.gov/pubs/sp/800/133/r2/final)
6. [NIST SP 800-57 Part 1 Rev. 5 - Recommendation for Key Management](https://csrc.nist.gov/pubs/sp/800/57/pt1/r5/final)


src/integrationTest/java/org/opensearch/security/hash/PBKDF2DefaultConfigHashingTests.java
```diff
@@ -28,15 +29,20 @@
...
-        .hash(
-            generatePBKDF2Hash(
-                "secret",
-                ConfigConstants.SECURITY_PASSWORD_HASHING_PBKDF2_FUNCTION_DEFAULT,
-                ConfigConstants.SECURITY_PASSWORD_HASHING_PBKDF2_ITERATIONS_DEFAULT,
-                ConfigConstants.SECURITY_PASSWORD_HASHING_PBKDF2_LENGTH_DEFAULT
+    private static final String password = CryptoServicesRegistrar.isInApprovedOnlyMode() ? "dtekVF0vEAA9FNvm#KMkTwMN" : "secret";
+
+    static {
+        ADMIN_USER = new TestSecurityConfig.User("admin").roles(ALL_ACCESS)
```
nibix: This modifies an attribute that was formely static final but was changed to just static in this PR.

I think this needs to be done in a different way, this is very prone to concurrency issues; and it is modifying static members of other classes is very surprising behavior which is usually not expected.

src/main/java/org/opensearch/security/auth/ldap/backend/LDAPAuthorizationBackend.java
```diff
@@ -201,6 +202,20 @@ private static void checkConnection0(
             config.setConnectionInitializer(new BindConnectionInitializer(bindDn, new Credential(password)));
 
             DefaultConnectionFactory connFactory = new DefaultConnectionFactory(config);
+
+            // Register custom socket factory for SNI hostname verification with BouncyCastle FIPS if SSL is enabled
+            if (config.getLdapUrl() != null && config.getLdapUrl().startsWith("ldaps:")) {
```
nibix, line 212: I am a bit confused by this; according to the PR this, this is only about the test code; yet, this is prod code. 

Could you elaborate which this change is necessary here? Possibly, we can do it in a separate PR?

src/integrationTest/java/org/opensearch/security/hash/PBKDF2DefaultConfigHashingTests.java
beanuwave: I agree - this is just a dirty hack to make it work. Thanks for pointing it out, it should be reworked before moving out of WIP

src/main/java/org/opensearch/security/auth/ldap/backend/LDAPAuthorizationBackend.java
beanuwave: I expect the amount of code to double or triple once this moves out of WIP. I also see potential to split this PR by authentication mechanism: LDAP, SAML, ... and possibly hashing-related logic, plus the rest.

src/integrationTest/java/org/opensearch/security/hash/Argon2DefaultConfigHashingTests.java
```diff
@@ -56,6 +59,14 @@ public class Argon2DefaultConfigHashingTests extends HashingTests {
         )
         .build();
 
+    @ClassRule
+    public static final RuleChain rules = RuleChain.outerRule(new ExternalResource() {
```
reta: Interesting, why not simply:

```
@Before
public void setUp() {
    Assume.assumeFalse("Skipping Argon2 hashing tests: Argon2 is (yet) not FIPS-compliant", FipsMode.isEnabled());
}
```

> A failing assumption in a @Before or @BeforeClass method will have the same effect as a failing assumption in each @Test method of the class. - https://github.com/junit-team/junit4/wiki/assumptions-with-assume

reta: Or same as https://github.com/opensearch-project/security/pull/5866/changes?w=1#diff-7169bc656600ff5de39321d16b584cd7da8f988b345b38d54c05a859ff7f8d77R59

src/integrationTest/java/org/opensearch/security/privileges/ApiTokenTest.java
```diff
@@ -56,10 +56,7 @@ public class ApiTokenTest {
     private static final String OBO_DESCRIPTION = "{\"description\":\"Testing\", \"service\":\"self-issued\"}";
     private static final String signingKey = Base64.getEncoder()
         .encodeToString("jwt signing key for an on behalf of token authentication backend for testing".getBytes(StandardCharsets.UTF_8));
-    private static final String encryptionKey = Base64.getEncoder().encodeToString("encryptionKey".getBytes(StandardCharsets.UTF_8));
-    public static final String ADMIN_USER_NAME = "admin";
-    public static final String REGULAR_USER_NAME = "regular_user";
-    public static final String DEFAULT_PASSWORD = "secret";
+    private static final String encryptionKey = Base64.getEncoder().encodeToString("encryptionKey!!".getBytes(StandardCharsets.UTF_8));
```
reta: We don't care about FIPS compatibility here (as in https://github.com/opensearch-project/security/pull/5866/changes?w=1#diff-f34c84b94aa1e1326e55ce390191ae380ce278f997f4ce1c2849310a662331c4R78)?

src/integrationTest/java/org/opensearch/security/ResourceFocusedTests.java
```diff
@@ -111,8 +111,8 @@ public void testUnauthenticatedManyMedium() {
         // Tweaks:
         final RequestBodySize size = RequestBodySize.Medium;
         final String requestPath = "/*/_search";
-        final int parallelism = 20;
-        final int totalNumberOfRequests = 10_000;
+        final int parallelism = FipsMode.isEnabled() ? 8 : 20;
```
reta: This looks odd, why do we have to change these settings? No relation to FIPS I believe

src/integrationTest/java/org/opensearch/test/framework/cluster/ReactorHttpClient.java
```diff
@@ -150,22 +173,47 @@ private HttpClient createClient(final InetSocketAddress remoteAddress, final Eve
 
         if (secure) {
             if (protocol == HttpProtocol.HTTP11) {
+                Consumer<SslContextBuilder> http11Configure = s -> {
+                    if (FipsMode.isEnabled()) {
+                        s.sslProvider(SslProvider.JDK);
```
reta: I believe we used to have openssl tests (and dependencies) but not anymore, how come non-JDK provider could be selected? We shouldn't have any on classpath

src/integrationTest/java/org/opensearch/security/http/OnBehalfOfJwtAuthenticationTest.java
```diff
@@ -72,11 +73,18 @@ public class OnBehalfOfJwtAuthenticationTest {
             "alternativeSigningKeyalternativeSigningKeyalternativeSigningKeyalternativeSigningKey".getBytes(StandardCharsets.UTF_8)
         );
 
-    private static final String encryptionKey = Base64.getEncoder().encodeToString("encryptionKey".getBytes(StandardCharsets.UTF_8));
+    private static final String encryptionKey = fipsCompatibleEncryptionKey();
+
+    private static String fipsCompatibleEncryptionKey() {
+        final byte[] keyMaterial = new byte[32];
+        new SecureRandom().nextBytes(keyMaterial);
```
reta:
```suggestion
        Randomness.createSecure().nextBytes(keyMaterial) 
```?

File: src/integrationTest/java/org/opensearch/test/framework/TestSecurityConfig.java [hunk 8]
```diff
@@ -95,10 +98,23 @@
 */
 public class TestSecurityConfig {
 
+    public static final String DEFAULT_TEST_PASSWORD;
+
+    static {
+        byte[] bytes = new byte[16]; // satisfies BC FIPS 112-bit minimum
+        new SecureRandom().nextBytes(bytes);
```
reta:
```suggestion
        Randomness.createSecure().nextBytes(bytes);
```?

File: src/integrationTest/java/org/opensearch/security/UserBruteForceAttacksPreventionTests.java [hunk 9]
```diff
@@ -38,7 +39,8 @@ public class UserBruteForceAttacksPreventionTests {
     private static final User USER_5 = new User("simple-user-5").roles(ALL_ACCESS);
 
     public static final int ALLOWED_TRIES = 3;
-    public static final int TIME_WINDOW_SECONDS = 3;
+    public static final int TIME_WINDOW_SECONDS = 30;
```
reta, line 42: Hm, 10x increase, why is that?

src/integrationTest/java/org/opensearch/test/framework/cluster/ReactorHttpClient.java
beanuwave: true, it's dead code on arrival. We can revert this change or replace it with `assert !OpenSsl.isAvailable()`

src/integrationTest/java/org/opensearch/security/ResourceFocusedTests.java
beanuwave: The generic client fires all requests at once `ParallelFlux.from(monos)` - for the test with 15k unbounded requests in FIPS-mode (executed in local env)  ~80% of exceptions are connect/handshake timeouts at the node: the server can't `accept()` + handshake fast enough, so SYNs/handshakes expire. FIPS trips it first because the BCTLS handshake is ~x4 costlier.

IMO it comes down to two solutions:
- Explicitly limit parallel requests - this code (introduces in fd9ed3036dee6174acc70bfec26dd470ab7abd03) 
- Proper saturation via backpressure (`parallelism` = demand). Let Reactor apply backpressure so `parallelism` **is** the concurrency limit:

```java
Flux.fromArray(monos).flatMap(mono -> mono, parallelism).collectList().block(...);
```

src/integrationTest/java/org/opensearch/security/hash/Argon2DefaultConfigHashingTests.java
beanuwave: The advantage is a single self-contained lifecycle: the `@ClassRule` starts the cluster once and tears it down automatically, and the outer `assumeFalse` gates that startup so the cluster never boots under FIPS. Moving the assumption into `@Before`/`@BeforeClass` still works but it's wasteful  - a `@ClassRule` runs first and the cluster would boot even in FIPS mode where no tests run. And the parameterized 3-hook pattern from `*CustomConfigHashingTests` (`@BeforeClass` skip + `@Before` build/start + `@After` close) would be misplaced here: it rebuilds the cluster per parameter row, and the `*DefaultConfigHashingTests` classes aren't parameterized.

src/integrationTest/java/org/opensearch/security/privileges/ApiTokenTest.java
beanuwave: "encryptionKey!!" at 120 bits satisfies cryptographic minimum the provider cares about >= 112bit. But we could reuse `TestSecurityConfig.DEFAULT_TEST_PASSWORD` here - so it future-proof.

src/integrationTest/java/org/opensearch/security/UserBruteForceAttacksPreventionTests.java
beanuwave: Switching form Bcrypt -> BPKDF2 (600k sequential HMAC-SHA256 iterations) roughly doubles per-request auth latency.  That matters because the attempts run back-to-back, so the window acts as an execution-time budget rather than a policy knob. Widening `TIME_WINDOW_SECONDS` is safe because no assertion inside this class depends on the window ever elapsing, and widening it cannot weaken anything.

Alternative aproach: lowering `plugins.security.password.hashing.pbkdf2.iterations` for the test cluster could shrink FIPS auth cost enough to fit the original 3s window.

Side note: most of the `Awaitility` usage in this class can go.

src/integrationTest/java/org/opensearch/security/ResourceFocusedTests.java
reta: What about third option to increase connect timeouts?

src/integrationTest/java/org/opensearch/security/privileges/ApiTokenTest.java
reta: Would be good to reuse, thanks

src/integrationTest/java/org/opensearch/security/UserBruteForceAttacksPreventionTests.java
reta: > Widening TIME_WINDOW_SECONDS is safe because no assertion inside this class depends on the window ever elapsing, and widening it cannot weaken anything

👍

src/main/java/org/opensearch/security/auth/ldap2/HostnameAwareConnectionFactory.java
```diff
@@ -0,0 +1,47 @@
...
+ * <p>Extending {@link DefaultConnectionFactory} (rateher than merely implementing
+ * {@code ConnectionFactory}) lets the same class back a connection pool, which requires a
+ * concrete {@link DefaultConnectionFactory}, as well as serve non-pooled connections directly.
+ * The pool creates each physical connection via {@link #getConnection()} and then calls
+ * {@code open()} on it, so the SNI wrapping applies uniformly to pooled and non-pooled paths.
+ *
+ * <p>This is necessary because JNDI LDAP resolves hostnames to IP addresses before creating
+ * SSL sockets, making the hostname unavailable for SNI configuration.
+ */
+public class HostnameAwareConnectionFactory extends DefaultConnectionFactory {
```
reta, line 33: The LDAP related changes (SNI, etc) seems to be unrelated to FIPS. could we extract them into separate feature + pull request?

beanuwave: I wouldn't quite call the LDAP changes FIPS-unrelated. The SNI plumbing and the PKCS12 workaround only become necessary because of running LDAPS through the BCTLS/BCFIPS provider stack.

That said, I do agree they're a fairly substantial, self-contained part of this PR. Splitting them out would make this diff smaller and keep LDAP-related chunk in one place.

Btw. as mentioned in the PR description, updating ldaptive for version 2.x would likely make the SNI plumbing unnecessary anyway.

reta: Thanks @beanuwave 

> That said, I do agree they're a fairly substantial, self-contained part of this PR. Splitting them out would make this diff smaller and keep LDAP-related chunk in one place.

👍 , this is exactly the reasons

> Btw. as mentioned in the PR description, updating ldaptive for version 2.x would likely make the SNI plumbing unnecessary anyway.

If we could do that (as separate change) so we could get rid of LDAP workarounds - even better

src/integrationTest/java/org/opensearch/security/ResourceFocusedTests.java
beanuwave: generic-client storm (15k requests, parallelism 100, FIPS)

| Metric                  |   30s / 30s |  120s / 120s |
|-------------------------|------------:|-------------:|
| ok                      | 7,920 (52.8%) | 13,525 (90.2%) |
| failed                  |       7,080 |        1,475 |
| `TCP_CONNECT_TIMEOUT`   |       5,144 |            0 |
| `TLS_HANDSHAKE_TIMEOUT` |         839 |            0 |
| `TCP_CONNECTION_RESET`  |         738 |        1,281 |
| `PREMATURE_CLOSE`       |         359 |          194 |
| elapsed time              |       50.1s |        81.9s |

**Takeaway:** raising the connect/handshake timeout (30s -> 120s) eliminates the`*_TIMEOUT` failures but cannot make the run green - `TCP_CONNECTION_RESET` (accept-queue overflow) grows and wall-clock rises.
2nd option is a better solution, because it keeps the accept queue shallow so nothing overflows and nothing waits near a timeout.

src/main/java/org/opensearch/security/auth/ldap2/HostnameAwareConnectionFactory.java
beanuwave: sound like a plan - I'll prepare a new LDAP PR!

src/integrationTest/java/org/opensearch/security/ResourceFocusedTests.java
reta: I see, thank you, let's do 2nd option

src/main/java/org/opensearch/security/authtoken/jwt/EncryptionDecryptionUtil.java
```diff
@@ -12,60 +12,124 @@
...
-        return Base64.getEncoder().encodeToString(encryptedBytes);
+    public EncryptionDecryptionUtil(final String encodedSecret) {
+        this(decodeBase64(encodedSecret));
     }
 
-    public String decrypt(final String encryptedString) {
-        byte[] decodedBytes = Base64.getDecoder().decode(encryptedString);
-        return new String(processWithCipher(decodedBytes, decryptCipher), StandardCharsets.UTF_8);
+    public EncryptionDecryptionUtil(final byte[] secretBytes) {
+        this.aesKey = deriveKey(secretBytes);
```
reta, line 70: I think with this change, the upgrade path (rolling upgrade primarily) for existing clusters will not be possible, right?

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
```diff
@@ -38,42 +41,92 @@
...
+    }
 
-    private final EncryptionDecryptionUtil encryptionUtil;
+    /**
+     * Builds the JWT parser and encryption helper on first use. Initialization is attempted exactly once.
+     *
+     * @return {@code true} if OBO authentication is usable, {@code false} if it is misconfigured
+     */
+    private synchronized boolean ensureInitialized() {
+        if (!initialized) {
```
reta, line 80: Probably `AtomicBoolean` + CAS will be safer: `if (initialized.compareAndSet(false, true)) ...`, than we don't really need `synchronized`

src/main/java/org/opensearch/security/ssl/config/KeyStoreConfiguration.java
```diff
@@ -29,13 +29,22 @@
...
 
     List<Path> files();
 
     List<Certificate> loadCertificates();
 
+    /**
+     * @return true when the key material is held in a PKCS#11 token. Such keys are non-exportable, so the
+     * TLS engine must delegate signing to the token's provider (SunPKCS11) rather than BouncyCastle FIPS.
+     */
+    default boolean isPkcs11() {
```
reta: Could we instead of checking PKCS11 in many places as true/false,  represent that as data classes? For `Pkcs11KeyStoreConfiguration` and `BcfksKeyStoreConfiguration` fe?

reta:
@beanuwave this is tremendous effort, thank you, a few high level comments:
 - the change is very large and difficult to review (it took me at least full day, split over the week), I strongly believe it will help to split it into smaller chunks (fe LDAP we already identified, there are probably other subareas)
 - the logic is primarily driven by `if`s (fips mode, store types, etc..), if we could project that to data classes / abstractions, it will be much simpler to read and reason about
 - the upgrade path (for existing clusters) in unclear (to me): could non FIPS node be updated in place? could FIPS node join non-FIPS cluster? (or vice versa)  

Hope it make sense, thank you!

beanuwave:
@reta Thank you for pushing this through the review process - that's great!

If you plan to do a second review, I'd really appreciate it if you could also take the "Reviewer call-outs" into account.

> the logic is primarily driven by ifs (fips mode, store types, etc..), if we could project that to data classes / abstractions, it will be much simpler to read and reason about

The FIPS test class abstraction is a no-brainer, since it would then follow the same style as the core project. However, I'm not sure the same general rules can be applied to production code - it's more of a case-by-case assessment.

> the upgrade path (for existing clusters) in unclear (to me): could non FIPS node be updated in place? could FIPS node join non-FIPS cluster? (or vice versa)

Do you have a specific scenario in mind where a heterogeneous cluster would be advantageous and provide real-world value? Assuming it would be possible for a non-FIPS node to join a FIPS cluster, IMO that would introduce a weak link in the chain and undermine the FIPS security guarantees of the entire cluster.

reta:
> If you plan to do a second review, I'd really appreciate it if you could also take the "Reviewer call-outs" into account.
> 
> >

👍 

> Do you have a specific scenario in mind where a heterogeneous cluster would be advantageous and provide real-world value? Assuming it would be possible for a non-FIPS node to join a FIPS cluster, IMO that would introduce a weak link in the chain and undermine the FIPS security guarantees of the entire cluster.

It is probably could be summarized like that: how would we recommend to start with FIPS support. Only brand new clusters or gradual migrations. 

> The FIPS test class abstraction is a no-brainer, since it would then follow the same style as the core project. However, I'm not sure the same general rules can be applied to production code - it's more of a case-by-case assessment

Correct, tests are indeed no-brainer, the one which really clicked for me for PCKS11, I think we could have cleaner implementation, but certainly - case by case.

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
beanuwave: A lock makes the loser wait until the winner has finished the work. CAS only makes the flag flip atomic - the loser is released immediately, before the winner's initialization completes, and reads uninitialized state.

src/main/java/org/opensearch/security/authtoken/jwt/EncryptionDecryptionUtil.java
beanuwave: Yes, all existing OBO connections will be interrupted for up to 10 minutes after the rollout. If this maintenance window is inconvenient, I can work out a migration plan to avoid the disruption.

src/main/java/org/opensearch/security/ssl/config/KeyStoreConfiguration.java
beanuwave: Certainly, this is a much cleaner design! Originally I rushed to get this through with a low-footprint design. However using polymorphism introduces quite a few code changes. Would you prefer the `TrustStoreConfiguration` implementations to remain as inner classes?

reta: Thanks @beanuwave 

> Would you prefer the TrustStoreConfiguration implementations to remain as inner classes?

Yeah, I think we could keep it there (and may be make them `sealed` if that makes sense). thanks!

src/main/java/org/opensearch/security/http/OnBehalfOfAuthenticator.java
reta: Agree, indeed the state is important

src/main/java/org/opensearch/security/ssl/config/KeyStoreConfiguration.java
beanuwave: 3ac26d19f3ab945357be9a886133484ecb8efd39 wdyt

src/main/java/org/opensearch/security/ssl/config/KeyStoreConfiguration.java
```diff
@@ -29,8 +25,9 @@
 
 import org.opensearch.OpenSearchException;
 import org.opensearch.common.collect.Tuple;
+import org.opensearch.security.support.PemKeyReader;
 
-public interface KeyStoreConfiguration {
+public sealed interface KeyStoreConfiguration {
```
reta, line 35: 👍

src/main/java/org/opensearch/security/ssl/SslConfiguration.java
```diff
@@ -138,8 +157,9 @@ SslContext buildClientSslContext(final boolean validateCertificates) {
                     .sslProvider(sslParameters.provider())
                     .keyManager(kmFactory)
                     .trustManager(trustStoreConfiguration.createTrustManagerFactory(validateCertificates, issuerDns))
-                    .endpointIdentificationAlgorithm(null)
-                    .build();
+                    .endpointIdentificationAlgorithm(null);
+                routePkcs11ThroughSunJsse(builder);
```
reta: May be add a method `configure` to  `KeyStoreConfiguration` (noop for JDK,  routePkcs11ThroughSunJsse for PKCS11): `keyStoreConfiguration.configure(SslContextBuilder builder)`

src/main/java/org/opensearch/security/ssl/config/SslCertificatesLoader.java
```diff
@@ -61,25 +62,26 @@ public SslCertificatesLoader(final String sslConfigSuffix, final String extended
     public Tuple<TrustStoreConfiguration, KeyStoreConfiguration> loadConfiguration(final Environment environment) {
         final var settings = environment.settings();
         final var sslConfigSettings = settings.getByPrefix(fullSslConfigSuffix);
-        if (settings.hasValue(sslConfigSuffix + KEYSTORE_FILEPATH)) {
-            final var keyStorePassword = resolvePassword(sslConfigSuffix + KEYSTORE_PASSWORD, settings, DEFAULT_STORE_PASSWORD);
+        final boolean isPkcs11Keystore = PemKeyReader.PKCS11.equalsIgnoreCase(environment.settings().get(sslConfigSuffix + KEYSTORE_TYPE));
```
reta: Could we moved those to factory methods? `TrustStoreConfiguration::buildTrustStoreConfiguration` and `KeyStoreConfiguration::buildKeyStoreConfiguration` ?

DarshitChanpura:
@beanuwave Are you still working on this PR? If so could you please resolve conflicts?

beanuwave:
@DarshitChanpura  I think this PR is now mostly useful as a historical archive, so we can probably close it.

@reta To make the changes easier to review, I've split it into a stack of x5 PRs, each focused on a major topic. PRs #2, #3, and #4 are interchangeable and can be reviewed in any order.

The only unresolved comment is related to *OBO token wire-format break*, which I'll try to fix soon. Everything else in the stack can already be reviewed.

```
main
 └─ 1  fips-split/1-build-tooling
     └─ 2  fips-split/2-ssl-store-config
         └─ 3  fips-split/3-ldap-sni
             └─ 4  fips-split/4-obo-keystore
                 └─ 5  fips-split/5-fips-enforcement
```

- https://github.com/opensearch-project/security/pull/6394
- https://github.com/opensearch-project/security/pull/6395
- https://github.com/opensearch-project/security/pull/6396
- https://github.com/opensearch-project/security/pull/6397
- https://github.com/opensearch-project/security/pull/6398

PR #5865, opened 2025-12-18
New tests for protected indices feature

nibix:
### Description

Another PR for preparing the improved index resolution feature #5827 ; this adds new tests for the protected index feature. This allows us to easily test the new implementation in the future.

* Category: Maintenance
* Why these changes are required? Preparation for #5827
* What is the old behavior before changes and new behavior after changes? no changes

### Check List
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/security/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #5869, opened 2025-12-19
Removed com.carrotsearch.randomizedtesting.RandomizedRunner from tests in src/integrationTest

nibix:
### Description

This is in reaction to the issues observed in https://github.com/opensearch-project/security/pull/5865#issuecomment-3671427643 

This PR removes the com.carrotsearch.randomizedtesting.RandomizedRunner test runner from the tests in src/integrationTest/ ; IMHO, this randomized testing approach is not really applicable to security plugin testing anyway. 

* Category: Maintenance
* Why these changes are required? Fixing issues with com.carrotsearch.randomizedtesting.RandomizedRunner
* What is the old behavior before changes and new behavior after changes? no changes

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/api/InternalUsersRestApiIntegrationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/api/InternalUsersRestApiIntegrationTest.java
@@ -73,7 +73,7 @@ public class InternalUsersRestApiIntegrationTest extends AbstractConfigEntityApi
 
             @Override
             public ToXContentObject entityPayload(Boolean hidden, Boolean reserved, Boolean _static) {
-                return internalUser(hidden, reserved, _static, "randomPassword", null, null, null);
+                return internalUser(hidden, reserved, _static, "rv2fR924a149ava32", null, null, null);
             }
 
             @Override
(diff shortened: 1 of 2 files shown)


nibix:
@cwperks @DarshitChanpura @shikharj05 what do you think about this?

This would fix the test issues in #5865

src/integrationTest/java/org/opensearch/security/api/InternalUsersRestApiIntegrationTest.java
```diff
@@ -72,7 +73,7 @@ public InternalUsersRestApiIntegrationTest() {
 
             @Override
             public ToXContentObject entityPayload(Boolean hidden, Boolean reserved, Boolean _static) {
-                return internalUser(hidden, reserved, _static, randomAsciiAlphanumOfLength(10), null, null, null);
+                return internalUser(hidden, reserved, _static, "rv2fR924a149ava32", null, null, null);
```
cwperks, line 76: Changes look ok to me, but why replace some of the calls like this?

nibix: This has two reasons:

- The randomAsciiAlphanumOfLength stems from the randomization infrastructure; this is not available any more.
- At some places, I added substitutes. However, for passwords, if we use a random value, there is a slight chance that the returned random value does not meet the password requirements (e.g. an upper case character is missing). This can lead to flakyness. Thus, a hard coded value which certainly passes password requirements seemed better to me.

cwperks (approved):

RyanL1997 (approved):

PR #5894, opened 2026-01-02
implement buildSecureClientTransportEngine with serverName parameter

orages:
### Description
This PR add support for `SecureTransportSettingsProvider.buildSecureClientTransportEngine(Settings settings, String serverName, String hostname, int port)` introduced in https://github.com/opensearch-project/OpenSearch/pull/20321  

This change add the possibility to create SSL Engine with the TLS Server Name Indication (SNI) extension populated with a given `serverName`.

### Issues Resolved
https://github.com/opensearch-project/OpenSearch/issues/17316

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/ssl/SslContextHandler.java
+++ b/src/main/java/org/opensearch/security/ssl/SslContextHandler.java
@@ -79,7 +79,7 @@ public class SslContextHandler {
             if (serverName != null) {
                 sslParams.setServerNames(List.of(new SNIHostName(serverName)));
             }
-            sslEngine.setSSLParameters(params);
+            sslEngine.setSSLParameters(sslParams);
         }
         return sslEngine;
     }


src/main/java/org/opensearch/security/ssl/SslContextHandler.java
```diff
@@ -71,6 +74,21 @@ public SSLEngine createClientSSLEngine(final String hostname, final int port) {
         return sslEngine;
     }
 
+    /**
+     * Creates a SSL engine for usage as a client with a specified Server Name Indication (SNI).
+     */
+    public SSLEngine createClientSSLEngine(final String hostname, final int port, final String serverName) {
```
reta, line 72: I think it would be better to move the  `createClientSSLEngine(final String hostname, final int port, final String serverName)` implementation here (and just delegate from it to this new method `return createClientSSLEngine(hostname, port, null);`)

src/main/java/org/opensearch/security/ssl/SslContextHandler.java
```diff
@@ -62,12 +65,26 @@ public SSLEngine createSSLEngine() {
...
     public SSLEngine createClientSSLEngine(final String hostname, final int port) {
+        return createClientSSLEngine(hostname, port, null);
+    }
+
+    /**
+     * Creates a SSL engine for usage as a client with a specified Server Name Indication (SNI).
+     */
+    public SSLEngine createClientSSLEngine(final String hostname, final int port, final String serverName) {
         SSLEngine sslEngine = sslContext.newEngine(NettyAllocator.getAllocator(), hostname, port);
         if (hostname != null) {
```
reta: A bit more concise I think:
```java
if (hostname != null || serverName != null) {
     SSLParameters sslParams = new SSLParameters();
    if (hostname != null) {
        sslParams.setEndpointIdentificationAlgorithm("HTTPS");
    }
    if (serverName != null) {  
            params.setServerNames(List.of(new SNIHostName(serverName));
    }
    sslEngine.setSSLParameters(params);
}
```

reta:
> Oh I see now, we are overloading `buildSecureClientTransportEngine` in core. Any reason why we are overloading it instead of modifying the existing signature? I'm not terribly concerned about bwc bc the security plugin would be the only consumer.

I think we could definitely use the new method signature only, the weak motivation behind keeping 2 options is that in 90% cases, the server name would not be needed

Issue #5867, opened 2025-12-19
[FEATURE] Support configurable timezone for audit log index naming instead of hardcoded UTC


salme4:
**Is your feature request related to a problem?**

Yes. When using daily audit log indices with the OpenSearch Security plugin, the index name and related rollover logic are fixed to UTC.
Currently, the code in AuditLogSink.java uses DateTime.now(DateTimeZone.UTC), making it impossible to change the timezone (e.g., to local timezones such as Asia/Seoul). This causes audit log indices to roll over based on UTC dates rather than the local date context.

**What solution would you like?**

Provide a configuration option to specify the timezone used for audit log index naming and related time calculations instead of always using UTC.
For example:
	•	Allow setting a timezone via audit configuration (e.g., plugins.security.audit.config.timezone)
	•	Update the implementation in AuditLogSink.java to respect this configured timezone when formatting index names and generating timestamps

This will allow users in local timezones to align audit log index dates with their local date boundaries instead of UTC.

**What alternatives have you considered?**

- Relying on custom index patterns only changes the format but does not alter the underlying UTC timestamp logic in the code.
- External scripts to rename indices after creation are possible workarounds but are manual and error prone.

**Do you have any additional context?**
In the current code, the timezone is hardcoded in:
```
protected String getExpandedIndexName(DateTimeFormatter indexPattern, String index) {
    if (indexPattern == null) {
        return index;
    }
    return indexPattern.print(DateTime.now(DateTimeZone.UTC));
}
```
As this method always uses UTC, there is no way to configure it for other timezones (e.g., Asia/Seoul). A configuration option or improvement to respect non-UTC timezones would be beneficial.

salme4:
Great point! Instead of hardcoding UTC, how about using the JVM’s default time zone (DateTimeZone.getDefault())?

We could follow a sensible fallback hierarchy:

1. Explicit configuration (if provided) - highest priority
2. JVM default timezone (system timezone) - sensible default

Would this approach align with OpenSearch's design philosophy? I'd be happy to prepare a PR if this sounds reasonable.

PR #5898, opened 2026-01-05
 Allow configuring the timezone for audit log #5867

pCastq:
### Description

### Issues Resolved

- Improves usability for users in non-UTC timezones.
- Fixes mismatch between local date expectations and audit log index rollover

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Enhancements
 - Support nested JWT claims in role DLS queries ([#5687](https://github.com/opensearch-project/security/issues/5687))
+- Support creation of client SSL engine with a given SNI ([#5894](https://github.com/opensearch-project/security/pull/5894))
+
 ### Bug Fixes
 - Fix IllegalArgumentException when resolved indices are empty in PrivilegesEvaluator ([#5770](https://github.com/opensearch-project/security/pull/5797))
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/security/auditlog/sink/AuditLogSink.java
```diff
@@ -36,6 +36,7 @@ public abstract class AuditLogSink {
     protected final AuditLogSink fallbackSink;
     private final int retryCount;
     private final long delayMs;
+    private DateTimeZone auditTimeZone;
```
nibix: this should be final

src/main/java/org/opensearch/security/auditlog/sink/AuditLogSink.java
```diff
@@ -45,6 +46,20 @@ protected AuditLogSink(String name, Settings settings, String settingsPrefix, Au
...
+        String timezone = settings.get(ConfigConstants.SECURITY_AUDIT_CONFIG_TIMEZONE, "UTC");
+        DateTimeZone parsedZone;
+        try {
+            parsedZone = DateTimeZone.forID(timezone);
+        } catch (IllegalArgumentException e) {
+            log.warn("Invalid timezone '{}', falling back to UTC", timezone);
+            parsedZone = DateTimeZone.UTC;
+        }
+        auditTimeZone = parsedZone;
+        return auditTimeZone;
```
nibix: It seems that you are assigning the object variable `auditTimeZone` here; afterwards, you return its value and assign it again in line 49. This is a bit confusing; I would suggest to not assign `auditTimeZone` in this method but just return `parsedZone`.  Then, you could make the method even static.

PR #5899, opened 2026-01-08
Allow configuring the timezone for audit log - Feature #5867

pCastq:
### Description

### Issues Resolved
Improves usability for users in non-UTC timezones.
Fixes mismatch between local date expectations and audit log index rollover

### Testing
Unit tests were implemented to verify the behavior of the expandedIndexName method in AuditLogSink regarding timezone (DateTimeZone) handling:

- Ensures a valid timezone set via settings is correctly applied to the generated timestamp.
- Checks that an invalid timezone falls back to UTC.
- Verifies the default timezone (when none is specified) is UTC.
- Handles null or unsupported patterns, returning the original index name without errors.
- Covers non-standard offsets (+3:30, -3:30) to ensure proper conversion and timestamp formatting.
- Uses a fixed-time setup (DateTimeUtils.setCurrentMillisFixed) for deterministic and reproducible tests.

In short, these tests stress the timezone handling logic, ensuring expected behavior for all valid, invalid, and edge-case configurations.

### Check List


src/main/java/org/opensearch/security/auditlog/sink/TestAuditLogSink.java
```diff
@@ -0,0 +1,35 @@
...
+    public TestAuditLogSink(String name, Settings settings, String settingsPrefix, AuditLogSink fallbackSink) {
+        super(name, settings, settingsPrefix, fallbackSink);
+    }
+
+    @Override
+    protected boolean doStore(AuditMessage msg) {
+        return true;
+    }
+
+    public String expandedIndexName(String pattern, String index) {
```
cwperks, line 31: @pCastq I think we should check an actual message to see what is stored in the `@timestamp` field. See an example message here: https://github.com/opensearch-project/security/blob/main/src/integrationTest/java/org/opensearch/security/rest/WhoAmITests.java#L221

pCastq: @cwperks Thanks a lot for your input. That inspired me to examine the issue more thoroughly (and also understand better something :) )

Hmm… I’m not entirely sure. What you are pointing out is an output of AuditMessage [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auditlog/impl/AuditMessage.java)
, so it’s a separate thing, among other things, that has _DateTimeZone.UTC_ hardcoded. But that’s correct because it’s a message used by _central audit_ (it centralizes information from nodes in different time zones, which prevents confusing data by time, right??).

I believe the test is correctly targeting the right part of the code because I’m just setting a DateTimeZone in AuditLogSink. What matters is that the correct value is set. The unit test I wrote focuses specifically on this, so I’m confident it verifies the intended behavior.

To clarify, I am not modifying the AuditMessage itself—its @timestamp remains in UTC. What I am controlling with this feature is the timezone used by the sink to name or partition local audit storage, for example:

```
file1-search-2026-01-20-15-30
file2-search-01-21-16-30
file3-search-2026-01-22-01-10
```

Regarding this, It might be a better choice to rename the key to
`security.audit.sink.timezone`  instead of  `audit.config.timezone` 
and add a small Javadoc:
```
/**
 * Timezone used for naming-audit indices or files.
 * Does NOT affect the @timestamp of audit messages, which is always in UTC.
 */
```
This should reduce confusion.

It would also be better to rename this:
 
`     private DateTimeZone auditTimeZone;`
into 
` private DateTimeZone  sinkDateTimeZone`
 
I think it would be valuable to add more Javadoc here to help others quickly understand these points and see what they do and how.


For example, for the field` sinkDateTimeZone`
 
```
  /**
 * Timezone used to calculate index names or filenames for audit messages.
 *
 * <p>This does <b>not</b> affect the @timestamp of the AuditMessage,
 * which remains in UTC. It is only used for formatting or grouping
 * events in local time for storage purposes.
 */
```
 
And for the class:
 
```
 /**
 * Base class for audit log outputs ("sinks").
 *
 * <p>An AuditLogSink handles storing AuditMessage events. It decides
 * how and where events are written (index name, file, etc.), 
 * including retry logic and fallback sinks. It does <b>not</b> modify
 * the timestamp of the events themselves, which are always in UTC.
 */
```
 
 
Let me know if I should apply these changes, or add anything else if needed.

Issue #5878, opened 2025-12-22
[BUG] audit log doesn't work with alias


rave189:
### Describe the bug

I was using OpenSearch version 2.19.3.
At that time, I was able to manage the audit log settings using an alias name and a rollover policy.
However, this setting is not applied in OpenSearch versions 3.2 and 3.3.2.
Opensearch throws an error log saying that it cannot insert values ​​into an index that already exists.

### Related component

Indexing

### Expected behavior

When saving the audit log, if there is an index that already exists, please modify it to insert it into that index.

### Additional Details

**Plugins**
Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**

**Additional context**

tender-barbarian:
@cwperks I've encountered this issue as well. It actually prevents me from updating to v3. Are there plans on fixing this any time soon please?

cwperks:
@tender-barbarian yes, I believe @pCastq has expressed interest. Once a PR is raised and merged I will update this issue with a tag for the release that a fix will be included in. I think v3.5.0 (next release may be possible), if not v3.5.0 then I would certainly expect it in v3.6.0.

cwperks:
Closing as this will be released in 3.6.0.

PR #5900, opened 2026-01-12
Fix audit log writing errors for rollover-enabled alias indices

pCastq:
### Description
[Bug fix, Enhancement , Test fix]

the audit log cannot be written to an index that already exists when using an alias with a rollover policy. This causes errors and prevents audit events from being stored.  
* Old behavior:  
  When an audit log index already exists under an alias, OpenSearch throws an error and does not insert audit events.  
* New behavior:  
  The audit log now detects if the target index already exists and inserts new events into it instead of failing. 

### Issues Resolved
[#5878 + integration test]

### Testing

* Integration tests confirming that rollover policies are respected and no errors occur when writing to an existing index.  

### Check List


src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,402 @@
...
+     * exists after event generation (e.g., {@code security-auditlog-2025.01.11}).</p>
+     */
+    @Test
+    public void testCreatesAuditIndexAutomatically() throws Exception {
+        try (Client client = cluster.getInternalNodeClient()) {
+            long eventCountBefore = countAuditEvents(client);
+            
+            generateAuditEvent("_cluster/health");
+
+            Thread.sleep(1500);
```
nibix: Thread.sleep() in tests is not good for two reasons:

- There might times when the runners are very busy, and this is actually not long enough. This leads to flakey tests.
- Otherwise, often its much faster, so this unnecessarily slows down tests

Please use something like https://github.com/awaitility/awaitility to poll the assertion condition.

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,402 @@
...
+     */
+    @Test
+    public void testPersistsAuditEventsToIndex() throws Exception {
+        try (Client client = cluster.getInternalNodeClient()) {
+            long eventCountBefore = countAuditEvents(client);
+
+            generateAuditEvent("_cluster/health");
+            generateAuditEvent("_cluster/stats");
+
+            Thread.sleep(1500);
```
nibix: same as above

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,402 @@
...
+    @ClassRule
+    public static final LocalCluster cluster = new LocalCluster.Builder().clusterManager(ClusterManager.SINGLENODE)
+        .anonymousAuth(true)
+        .internalAudit(new AuditConfiguration(true).filters(new AuditFilters().enabledRest(true).enabledTransport(false)))
+        .build();
+
+    @BeforeClass
+    public static void waitForCluster() throws Exception {
+        // Allow cluster initialization and plugin loading
+        Thread.sleep(2000);
```
nibix: This should be unnecessary

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,402 @@
...
+
+    private static final String AUDIT_INDEX_PREFIX = "security-auditlog-";
+
+    /**
+     * Shared cluster for all tests in this class.
+     * Uses default audit configuration without custom index override.
+     */
+    @ClassRule
+    public static final LocalCluster cluster = new LocalCluster.Builder().clusterManager(ClusterManager.SINGLENODE)
+        .anonymousAuth(true)
```
nibix: do we need that here?

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,402 @@
...
+ *
+ * <h3>Why Separate from Alias Tests?</h3>
+ * <p>This class tests the regular index creation path, while
+ * {@code InternalOpenSearchSinkIntegrationTest_AuditAlias} tests the alias path
+ * ({@code metadata.hasAlias(indexName)}). These are mutually exclusive configurations
+ * requiring separate cluster setups.</p>
+ *
+ * @see InternalOpenSearchSinkIntegrationTest_AuditAlias
+ */
+public class InternalOpenSearchSinkIntegrationTest {
```
nibix: We already have https://github.com/opensearch-project/security/blob/51039a5b2c1fdda8dd01c6eaaa63b521e113e324/src/integrationTest/java/org/opensearch/security/InternalAuditLogTest.java#L29-L52 ... for code hygiene, i guess we should combine these, otherwise it might be confusing.

nibix (commented):
Thank you for this. I left a couple of comments.

Could you also please:

- Check the CI errors regarding code hygiene? maybe it is just a missing spotlessApply
- Write a changelog entry
- Change the name of this PR to something more descriptive 

Thank you :)

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
pCastq: I'll try to take a closer look, ty !

pCastq: @nibix I replied there [resp](https://github.com/opensearch-project/security/pull/5900#issuecomment-3757532492)

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,404 @@
...
+     */
+    @Test
+    public void testCreatesAuditIndexAutomatically() {
+        try (Client client = cluster.getInternalNodeClient()) {
+            long eventCountBefore = countAuditEvents(client);
+
+            generateAuditEvent("_cluster/health");
+
+            await().atMost(3, SECONDS).pollInterval(100, MILLISECONDS).untilAsserted(() -> {
+                refreshAuditIndices(client); // refresh prima di verificare
```
nibix: si prega di preferire l'uso dell'inglese :-)

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,404 @@
...
+     * exists after event generation (e.g., {@code security-auditlog-2025.01.11}).</p>
+     */
+    @Test
+    public void testCreatesAuditIndexAutomatically() {
+        try (Client client = cluster.getInternalNodeClient()) {
+            long eventCountBefore = countAuditEvents(client);
+
+            generateAuditEvent("_cluster/health");
+
+            await().atMost(3, SECONDS).pollInterval(100, MILLISECONDS).untilAsserted(() -> {
```
nibix: i would be a bit more generous regarding the timeout to be sure that this tests also succeeds in case of special situations, like a concurrent GC sweep, etc.

I think 10 seconds would be safer.

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,404 @@
...
+            generateAuditEvent("_cluster/health");
+            generateAuditEvent("_cluster/stats");
+            generateAuditEvent("_nodes/stats");
+
+            await().atMost(3, SECONDS).pollInterval(100, MILLISECONDS).untilAsserted(() -> {
+                refreshAuditIndices(client);
+                assertThat(
+                    "All generated events must be persisted without errors",
+                    countAuditEvents(client) - eventCountBefore,
+                    greaterThanOrEqualTo(3L)
```
nibix: I do not really understand the difference to `testPersistsAuditEventsToIndex()`: In one case, we generate two audit events and in the other three audit events: How does this change the index recreation behavior? IMHO, this should be already covered by two tests?

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,404 @@
...
+     *
+     * <p><b>Purpose:</b> Confirms that the audit sink captures events across
+     * different REST API categories, not just specific endpoints.</p>
+     *
+     * <p><b>Test Strategy:</b> Uses delta-based counting to ensure this test
+     * generates at least 3 new events, independent of prior test executions
+     * and midnight rollover timing.</p>
+     */
+    @Test
+    public void testMultipleRequestTypesGenerateAuditEvents() {
```
nibix: Could you elaborate why this is something related to the OpenSearch sink? Isnt this a behavior that occurs at a higher level?

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,404 @@
...
+                assertThat("At least one audit index must exist", indicesResponse.indices().length, greaterThan(0));
+
+                boolean foundDateBasedIndex = false;
+                for (String indexName : indicesResponse.indices()) {
+                    if (indexName.matches("^security-auditlog-\\d{4}\\.\\d{2}\\.\\d{2}$")) {
+                        foundDateBasedIndex = true;
+                        break;
+                    }
+                }
+                assertThat("At least one index must follow pattern: security-auditlog-YYYY.MM.dd", foundDateBasedIndex, is(true));
```
nibix: This condition would actually hold for all indices matching `AUDIT_INDEX_PREFIX + "*"`, no?

If so, we could just assert this condition for all index names. I guess this would simplify the assertion code.

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTestAuditAlias.java
```diff
@@ -0,0 +1,321 @@
...
+     * <ul>
+     *   <li>{@code audit_transport_request_type} - Must NOT exist for REST events</li>
+     * </ul>
+     *
+     * <p><b>Test Strategy:</b> Generates a fresh event and validates the most recent
+     * document to avoid contamination from previous tests. Uses delta to ensure this
+     * test created a new event.</p>
+     */
+    @Test
+    public void testAuditDocumentsViaAliasContainMandatoryFields() {
```
nibix, line 136: Is this a test which covers the alias logic? It seems to be this covers logic from a higher level.

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,404 @@
...
+ *
+ * <h3>Why Separate from Alias Tests?</h3>
+ * <p>This class tests the regular index creation path, while
+ * {@code InternalOpenSearchSinkIntegrationTestAuditAlias} tests the alias path
+ * ({@code metadata.hasAlias(indexName)}). These are mutually exclusive configurations
+ * requiring separate cluster setups.</p>
+ *
+ * @see InternalOpenSearchSinkIntegrationTestAuditAlias
+ */
+public class InternalOpenSearchSinkIntegrationTest {
```
cwperks: I'm generally in favor of expanded coverage (esp if there are gaps), but this file looks unrelated to the functional change in this PR. I think we should keep the tests in this PR specific to the change being introduced.

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
pCastq: maybe redundant,I'll removing it

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
pCastq: ok I'll try to clean the integration test , ty !

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -252,22 +150,18 @@ public void testReusesExistingIndexWithoutRecreation() {
...
+     *   <li>{@code audit_rest_request_method} - HTTP method</li>
+     *   <li>{@code audit_rest_request_path} - Request path</li>
      * </ul>
-     *
-     * <p><b>Test Strategy:</b> Generates a fresh event and validates the most
-     * recent document (sorted by timestamp DESC) to avoid contamination from
-     * previous tests. Uses delta assertion to ensure a new event was created.</p>
      */
     @Test
     public void testAuditDocumentContainsMandatoryFields() {
```
pCastq: This test is intentionally at the integration level because it verifies the real effect of the original bug: before the fix, writing via an alias failed and no document was created.
Even though it doesn’t call metadata.hasAlias() directly, it confirms that alias writes behave like direct index writes, including all required fields, and it protects against future regressions, for example if the audit log writing logic is refactored.
Removing it would remove this critical check.

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
nibix: I am still not quite sure if we need anonymous auth?

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,249 @@
...
+     * without manual index creation.</p>
+     */
+    @Test
+    public void testCreatesAuditIndexAutomatically() {
+        try (Client client = cluster.getInternalNodeClient()) {
+            long eventCountBefore = countAuditEvents(client);
+
+            generateAuditEvent("_cluster/health");
+
+            await().atMost(10, SECONDS).pollInterval(100, MILLISECONDS).untilAsserted(() -> {
```
nibix: 10 seconds is actually the default timeout constraint, so this can be dropped. Same holds for the poll interval.

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTestAuditAlias.java
```diff
@@ -0,0 +1,208 @@
...
+     * <p><b>Test coverage:</b> Confirms that documents routed via aliases maintain
+     * expected audit fields and values (REST events only).</p>
+     */
+    @Test
+    public void testAuditDocumentsViaAliasContainMandatoryFields() {
+        try (Client client = cluster.getInternalNodeClient()) {
+            long before = countAuditDocs(client);
+            generateAuditEvent("_cluster/health");
+
+            await().atMost(3, SECONDS).pollInterval(100, MILLISECONDS).until(() -> countAuditDocs(client) > before);
```
nibix: can we also use default timeout and poll interval here?

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,249 @@
...
+                greaterThan(0L)
+            );
+
+            Map<String, Object> auditDoc = response.getHits().getAt(0).getSourceAsMap();
+
+            assertThat("Missing mandatory field: audit_category", auditDoc.containsKey("audit_category"), is(true));
+            assertThat("Missing mandatory field: audit_request_origin", auditDoc.containsKey("audit_request_origin"), is(true));
+            assertThat("Missing mandatory field: @timestamp", auditDoc.containsKey("@timestamp"), is(true));
+            assertThat("Missing REST field: audit_rest_request_method", auditDoc.containsKey("audit_rest_request_method"), is(true));
+            assertThat("Missing REST field: audit_rest_request_path", auditDoc.containsKey("audit_rest_request_path"), is(true));
```
nibix: The hamcrest way of doing such assertions would be 

```
assertThat(auditDoc, hasKey("@timestamp"));
assertThat(auditDoc, hasKey("audit_rest_request_method"));
assertThat(auditDoc, hasKey("audit_rest_request_path"));
```

This will yield much richer assertion errors in the case of test failures.

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTestAuditAlias.java
```diff
@@ -0,0 +1,208 @@
...
+                .actionGet();
+        }
+    }
+
+    private long countAuditDocs(Client client) {
+        SearchResponse response = client.search(
+            new SearchRequest(AUDIT_ALIAS).source(new SearchSourceBuilder().query(QueryBuilders.matchAllQuery()).size(0))
+        ).actionGet();
+
+        return Objects.requireNonNull(response.getHits().getTotalHits()).value();
```
nibix: This `Objects.requireNonNull` is redundant, with our without you will get a NPE in case getTotalHits() returns null.

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTestAuditAlias.java
pCastq: This test validates the end-to-end write path through the alias branch of createIndexIfAbsent(). The other two tests verify that the alias is recognized (testRecognizesAuditTargetAsWriteAlias) and that documents are counted (testWritesEventsToAliasSuccessfully), but neither inspects the document content. This test closes the loop by confirming that a document written through the alias pipeline — createIndexIfAbsent() → prepareIndex(aliasName) → execute() — arrives in the backing index structurally intact with all mandatory fields and correct values. Without it, we'd know documents are delivered but not that they're correct.

CHANGELOG.md
```diff
@@ -14,6 +14,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Make security plugin aware of FIPS build param (-Pcrypto.standard=FIPS-140-3) ([#5952](https://github.com/opensearch-project/security/pull/5952))
 
 ### Bug Fixes
+- Fix IllegalArgumentException when resolved indices are empty in PrivilegesEvaluator ([#5770](https://github.com/opensearch-project/security/pull/5797))
+- Fixes an issue where recursive LDAP role search would fail with a NullPointerException ([#5861](https://github.com/opensearch-project/security/pull/5861))
+- Serialize Search Request object in DLS Filter Level Handler only when debug mode is enabled ([#5883](https://github.com/opensearch-project/security/pull/5883))
+- Skip hasExplicitIndexPrivilege check for plugin users accessing their own system indices ([#5858](https://github.com/opensearch-project/security/pull/5858))
+- Fix test failure related to change in core to add content-encoding to response headers ([#5897](https://github.com/opensearch-project/security/pull/5897))
+- Fix audit log writing errors for rollover-enabled alias indices ([#5878](https://github.com/opensearch-project/security/pull/5878)
```
cwperks:
```suggestion
- Fix audit log writing errors for rollover-enabled alias indices ([#5878](https://github.com/opensearch-project/security/pull/5878)
```

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTest.java
```diff
@@ -0,0 +1,243 @@
...
+ * daily indices (e.g., {@code security-auditlog-2025.01.11}).</p>
+ *
+ * <h5>Tested Code Path:</h5>
+ * <p>These tests exercise the regular index flow in {@code createIndexIfAbsent()}:
+ * both the {@code CreateIndexRequest} branch (when the index does not yet exist)
+ * and the {@code metadata.hasIndex(indexName)} early-return branch (when it already exists).</p>
+ *
+ * @see InternalOpenSearchSinkIntegrationTestAuditAlias for alias-specific tests
+ */
+public class InternalOpenSearchSinkIntegrationTest {
```
cwperks: Can we add this class in a separate PR and keep the tests in this PR limited to the functional change? I don't see how the tests in here relate to the alias fix.

src/integrationTest/java/org/opensearch/security/auditlog/sink/InternalOpenSearchSinkIntegrationTestAuditAlias.java
```diff
@@ -0,0 +1,202 @@
...
+
+    @ClassRule
+    public static final LocalCluster cluster = new LocalCluster.Builder().clusterManager(ClusterManager.SINGLENODE)
+        .nodeSettings(Map.of("plugins.security.audit.config.index", AUDIT_ALIAS))
+        .internalAudit(new AuditConfiguration(true).filters(new AuditFilters().enabledRest(true).enabledTransport(false)))
+        .build();
+
+    @BeforeClass
+    public static void setupAuditAlias() {
+        try (Client client = cluster.getInternalNodeClient()) {
```
cwperks: nit: prefer to use RestClient here instead of internalNodeClient

cwperks (approved):

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Make security plugin aware of FIPS build param (-Pcrypto.standard=FIPS-140-3) ([#5952](https://github.com/opensearch-project/security/pull/5952))
 
 ### Bug Fixes
+- Fix audit log writing errors for rollover-enabled alias indices ([#5878](https://github.com/opensearch-project/security/pull/5878)
+
```
DarshitChanpura, line 18: nit: empty line

RyanL1997 (approved):

PR #5901, opened 2026-01-12
Allow configuring the timezone for audit log - Feature #5867 

pCastq:
Description

Issues Resolved
Improves usability for users in non-UTC timezones.
Fixes mismatch between local date expectations and audit log index rollover

Testing
Unit tests were implemented to verify the behavior of the expandedIndexName method in AuditLogSink regarding timezone (DateTimeZone) handling:

Ensures a valid timezone set via settings is correctly applied to the generated timestamp.
Checks that an invalid timezone falls back to UTC.
Verifies the default timezone (when none is specified) is UTC.
Handles null or unsupported patterns, returning the original index name without errors.
Covers non-standard offsets (+3:30, -3:30) to ensure proper conversion and timestamp formatting.
Uses a fixed-time setup (DateTimeUtils.setCurrentMillisFixed) for deterministic and reproducible tests.
In short, these tests stress the timezone handling logic, ensuring expected behavior for all valid, invalid, and edge-case

### Check List


pCastq:
A few points to clarify ...

3 components can cause confusion :


1) The audit message `(AuditMessage) `-> rappresent the message like
` ....  " ... " ,  "@timestamp":"2023-08-16T17:35:53.621+00:00",  " ... " , ...`

and has forced internally to DateTimeZone.UTC

```
private String currentTime() {
    DateTime dt = new DateTime(DateTimeZone.UTC);
    return DEFAULT_FORMAT.print(dt);
}
```
so that audit events from different nodes and timezones can be reliably ordered, compared, and aggregated.

and this object is used by others !

2) The Audit Implementation `(AuditLogImpl)` -> (write all message from different node that use save(final AuditMessage msg))

3)The sink audit `(AuditLogSink)` -> ( have a personal DateTimezone (new feature developed) that use for store local file or indices ) and have a method store(AuditMessage msg), take a AuditMessage that have, as we look, the fixing UTC timezone internally).

So we just develop the change to store for example locally file with name like this (_### example_) :

file1-search-2026-01-20-15-30
file2-search-01-21-16-30
file3-search-2026-01-22-01-10
in the right timezone. For customers not in utc zone , is very helpfull


@cwperks in previous PR (where i made a mistake with a branch -> then I closed),
You were looking at the output of AuditMessage, not the actual problem.
It’s a separate thing, among other things, that has DateTimeZone.UTC hardcoded. But that’s correct because it’s a message used by central audit (it centralizes information from nodes in different time zones, which prevents confusing data by time, for have a sort data information, right??(need conferm ,but I think yes)).

I believe the test is correctly targeting the right part of the code because I’m just setting a DateTimeZone in AuditLogSink. What matters is that the correct value is set. The unit test I wrote focuses specifically on this, so I’m confident it verifies the intended behavior.

To clarify, I am not modifying the AuditMessage itself—its @timestamp remains in UTC. What I am controlling with this feature is the timezone used by the sink to name or partition local audit storage, for example:

file1-search-2026-01-20-15-30
file2-search-01-21-16-30
file3-search-2026-01-22-01-10
Regarding this, It might be a better choice to rename the key to
`security.audit.sink.timezone`   instead of   `audit.config.timezone`
and add a small Javadoc:

/**
 * Timezone used for naming-audit indices or files.
 * Does NOT affect the @timestamp of audit messages, which is always in UTC.
 */
This should reduce confusion.

It would also be better to rename this:

` private DateTimeZone auditTimeZone;`
into
` private DateTimeZone  sinkDateTimeZone`

I think it would be valuable to add more Javadoc here to help others quickly understand these points and see what they do and how.

For example, for the field sinkDateTimeZone

  /**
 * Timezone used to calculate index names or filenames for audit messages.
 *
 * <p>This does <b>not</b> affect the @timestamp of the AuditMessage,
 * which remains in UTC. It is only used for formatting or grouping
 * events in local time for storage purposes.
 */
And for the class:

 /**
 * Base class for audit log outputs ("sinks").
 *
 * <p>An AuditLogSink handles storing AuditMessage events. It decides
 * how and where events are written (index name, file, etc.), 
 * including retry logic and fallback sinks. It does <b>not</b> modify
 * the timestamp of the events themselves, which are always in UTC.
 */
 
Let me know if I should apply these changes, or add anything else if needed.

src/main/java/org/opensearch/security/auditlog/sink/TestAuditLogSink.java
```diff
@@ -0,0 +1,35 @@
...
+    public TestAuditLogSink(String name, Settings settings, String settingsPrefix, AuditLogSink fallbackSink) {
+        super(name, settings, settingsPrefix, fallbackSink);
+    }
+
+    @Override
+    protected boolean doStore(AuditMessage msg) {
+        return true;
+    }
+
+    public String expandedIndexName(String pattern, String index) {
```
cwperks: @pCastq The changes mostly look good to me, but I have one question about this new test audit log.

I see we have an existing one called [LoggingSink](https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/auditlog/helper/LoggingSink.java), could we perhaps re-use that?

Another question is if it may be possible to use `@VisibleForTesting` or change the access modifier in `AuditLogSink` of `getExpandedIndexName` so that its visible in a test and we don't need this wrapper.

pCastq: Ty for u input. 

I don't know the main reason anyway AuditLogSink is an abstract class , it is supposed to have some implementation in order to work.  I can use a LoggingSink and the impact of this change is small, and I think it’s a good solution!

Do you think is good to change name of key ?? 
It would also be better to rename this:

`private DateTimeZone auditTimeZone;`
into
` private DateTimeZone  sinkDateTimeZone`
 
maybe made less confusion.

cwperks: IMO I think `auditTimeZone` is more of an apt name. I believe there can only be a single sink configured for the audit log, but if multiple sinks were supported then I would think the latter would be more descriptive.

cwperks (dismissed):

cwperks (approved):

nibix (approved):

PR #5914, opened 2026-01-19
Enable audit logging of document contents for DELETE operations

### LINKED ISSUE #5872: [FEATURE] Audit log does not capture deleted document contents for delete_by_query when write diff logging is enabled
### Problem

When documents are deleted via `delete_by_query`, the OpenSearch Security audit log currently does **not** capture the contents of the deleted documents—even when `write_log_diffs` is enabled.

This is inconsistent with existing behavior for write operations (`create`, `index`, `update`), where the audit log can diffs when configured to do so.

As a result, users who rely on audit logs for compliance lack visibility into **what data was deleted**, even though equivalent visibility exists for document writes.

---

### Expected Behavior

When write diff logging is enabled, audit log entries corresponding to document deletion (including `delete_by_query`) should include the document contents (or diffs, where applicable), similar to how document writes are handled today.

---

### Current Behavior

* `delete_by_query` operations generate audit log entries
* The deleted document contents are **not logged** (metadata like index name and doc id are logged)
* This differs from `create`, `index`, and `update` operations, which do log document contents/diffs when enabled

---

### Proposed Change

Extend audit logging for delete operations so that document contents are logged when write diff logging is enabled.

At a high level, this can be implemented by reusing existing logic already present for write operations:

1. **Hook into delete handling**

   * Leverage `postDelete` in `ComplianceIndexingOperationListenerImpl`
     [https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java#L70-L81](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java#L70-L81)
   * This currently invokes `auditLog.logDocumentDeleted(...)`

2. **Enhance delete audit logging**

   * Extend `auditLog.logDocumentDeleted` to include document content
   * Reuse logic from `auditLog.logDocumentWritten`, which already:

     * Handles write diff logging
     * Captures document contents and diffs when enabled
       [https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java#L542-L636](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java#L542-L636)

3. **Behavior parity**

   * Deletion audit logging should follow the same configuration and semantics as write operations
   * No behavior change when write diff logging is disabled

Example queries for testing

```
PUT /movies
{
  "mappings": {
    "properties": {
      "title": {
        "type": "text"
      },
      "director": {
        "type": "keyword"
      },
      "year": {
        "type": "integer"
      },
      "genre": {
        "type": "keyword"
      }
    }
  }
}
```

```
POST /_bulk
{ "index": { "_index": "movies", "_id": "1" } }
{ "title": "Jaws", "director": "Steven Spielberg", "year": 1975, "genre": "Thriller" }
{ "index": { "_index": "movies", "_id": "2" } }
{ "title": "E.T. the Extra-Terrestrial", "director": "Steven Spielberg", "year": 1982, "genre": "Sci-Fi" }
{ "index": { "_index": "movies", "_id": "3" } }
{ "title": "Jurassic Park", "director": "Steven Spielberg", "year": 1993, "genre": "Adventure" }
{ "index": { "_index": "movies", "_id": "4" } }
{ "title": "Inception", "director": "Christopher Nolan", "year": 2010, "genre": "Sci-Fi" }
{ "index": { "_index": "movies", "_id": "5" } }
{ "title": "The Dark Knight", "director": "Christopher Nolan", "year": 2008, "genre": "Action" }
```

```
POST /movies/_delete_by_query
{
  "query": {
    "term": {
      "director": "Steven Spielberg"
    }
  }
}
```

```
GET /movies/_search
{
  "query": {
    "match_all": {}
  }
}
```

```
DELETE movies
```

```
POST security-auditlog-2025.12.17/_search
{
  "query": {
    "match_all": {}
  },
  "sort": [
    {
      "@timestamp": {
        "order": "desc"
      }
    }
  ]
}
```

Rishav9852Kumar:
### Description
This change enables audit logging to capture the contents of deleted documents when compliance write diff logging is enabled, providing parity with CREATE and UPDATE operations.
* Category: Enhancement
  * Currently, when documents are deleted (including via delete_by_query), the audit log does not capture the contents of the deleted documents, even when `write_log_diffs` is enabled
  * This creates an inconsistency with write operations (CREATE, UPDATE) which do log document contents/diffs when configured
  * For compliance purposes, users need visibility into what data was deleted, not just metadata like index name and document ID
 * __What is the old behavior before changes and new behavior after changes?__
 
   * Old Behavior:
     * DELETE operations only logged metadata (index, document ID, operation type)
     * `audit_compliance_diff_content` field was not present in DELETE audit logs
     * No visibility into what document content was removed
   * New Behavior:
     * When `write_log_diffs` is enabled, DELETE operations now include the full document content in audit logs
     * `audit_compliance_diff_content` field contains JSON Patch format showing removed fields
     * Consistent behavior across all write operations (CREATE, UPDATE, DELETE)
     * Works for all delete types: single document delete, delete_by_query, and bulk delete operations

### Issues Resolved
Fixes #5872

Is this a backport? No

### Testing
 Added new integration test `testDeleteLogDiffs()` in `ComplianceAuditlogTest.java`
   * Test validates DELETE operations generate audit logs with document content/diffs
   * Verifies `audit_compliance_diff_content` contains expected remove operations
   * Verifies all document fields are captured in the diff

#### Test Case 1: Single Document Delete [✅ DELETE /<index>/_doc/<id>]
```
curl -XDELETE "https://localhost:9200/movies/_doc/1?pretty" \
  -u 'admin:Rishav@123' \
  --insecure
```
Audit Log Entry:
``` json
{
        "_index": "security-auditlog-2026.01.19",
        "_id": "UsaS1psBqZax7JAdhsVy",
        "_score": null,
        "_source": {
          "audit_compliance_operation": "DELETE",
          "audit_cluster_name": "opensearch",
          "audit_node_name": "c889f3ea09d1",
          "audit_category": "COMPLIANCE_DOC_WRITE",
          "audit_request_origin": "REST",
          "audit_compliance_doc_version": 6,
          "audit_node_id": "5oKZgVZwSeun0L0mlMTolg",
          "@timestamp": "2026-01-19T14:04:45.543+00:00",
          "audit_compliance_diff_is_noop": false,
          "audit_format_version": 4,
          "audit_request_remote_address": "::1",
          "audit_trace_doc_id": "1",
          "audit_compliance_diff_content": "[{\"op\":\"remove\",\"path\":\"/title\",\"value\":\"Jaws\"},{\"op\":\"remove\",\"path\":\"/director\",\"value\":\"Steven Spielberg\"},{\"op\":\"remove\",\"path\":\"/year\",\"value\":1975},{\"op\":\"remove\",\"path\":\"/genre\",\"value\":\"Thriller\"}]",
          "audit_node_host_address": "127.0.0.1",
          "audit_request_effective_user": "admin",
          "audit_trace_shard_id": 0,
          "audit_trace_indices": [
            "movies"
          ],
          "audit_trace_resolved_indices": [
            "movies"
          ],
          "audit_node_host_name": "127.0.0.1"
        },
        "sort": [
          1768831485543
        ]
      }
```
#### Test Case 2: Delete by Query [✅ POST /<index>/_delete_by_query]
```
# Delete all Christopher Nolan movies
curl -XPOST "https://localhost:9200/movies/_delete_by_query?pretty" \
  -u 'admin:Rishav@123' \
  --insecure \
  -H 'Content-Type: application/json' \
  -d '{
  "query": {
    "term": {
      "director": "Christopher Nolan"
    }
  }
}'
```
Audit Log Entry:
``` json
{
        "_index": "security-auditlog-2026.01.19",
        "_id": "V8aV1psBqZax7JAd5sWF",
        "_score": null,
        "_source": {
          "audit_compliance_operation": "DELETE",
          "audit_cluster_name": "opensearch",
          "audit_node_name": "c889f3ea09d1",
          "audit_category": "COMPLIANCE_DOC_WRITE",
          "audit_request_origin": "REST",
          "audit_compliance_doc_version": 2,
          "audit_node_id": "5oKZgVZwSeun0L0mlMTolg",
          "@timestamp": "2026-01-19T14:08:26.754+00:00",
          "audit_compliance_diff_is_noop": false,
          "audit_format_version": 4,
          "audit_request_remote_address": "::1",
          "audit_trace_doc_id": "9",
          "audit_compliance_diff_content": "[{\"op\":\"remove\",\"path\":\"/title\",\"value\":\"Tenet\"},{\"op\":\"remove\",\"path\":\"/director\",\"value\":\"Christopher Nolan\"},{\"op\":\"remove\",\"path\":\"/year\",\"value\":2020},{\"op\":\"remove\",\"path\":\"/genre\",\"value\":\"Sci-Fi\"}]",
          "audit_node_host_address": "127.0.0.1",
          "audit_request_effective_user": "admin",
          "audit_trace_shard_id": 0,
          "audit_trace_indices": [
            "movies"
          ],
          "audit_trace_resolved_indices": [
            "movies"
          ],
          "audit_node_host_name": "127.0.0.1"
        },
        "sort": [
          1768831706754
        ]
      }
```
 #### Test Case 3: Bulk Delete [✅ POST /_bulk with delete actions]
```
# Bulk delete remaining documents (IDs: 2, 3, 8) 
curl -XPOST "https://localhost:9200/_bulk?pretty" \
  -u 'admin:Rishav@123' \
  --insecure \
  -H 'Content-Type: application/json' \
  -d '
{"delete": {"_index": "movies", "_id": "2"}}
{"delete": {"_index": "movies", "_id": "3"}}
{"delete": {"_index": "movies", "_id": "8"}}
'
```
Audit Log Entry:
``` json
 {
        "_index": "security-auditlog-2026.01.19",
        "_id": "WsaX1psBqZax7JAdS8V0",
        "_score": null,
        "_source": {
          "audit_compliance_operation": "DELETE",
          "audit_cluster_name": "opensearch",
          "audit_node_name": "c889f3ea09d1",
          "audit_category": "COMPLIANCE_DOC_WRITE",
          "audit_request_origin": "REST",
          "audit_compliance_doc_version": 2,
          "audit_node_id": "5oKZgVZwSeun0L0mlMTolg",
          "@timestamp": "2026-01-19T14:09:58.129+00:00",
          "audit_compliance_diff_is_noop": false,
          "audit_format_version": 4,
          "audit_request_remote_address": "::1",
          "audit_trace_doc_id": "8",
          "audit_compliance_diff_content": "[{\"op\":\"remove\",\"path\":\"/title\",\"value\":\"Schindlers List\"},{\"op\":\"remove\",\"path\":\"/director\",\"value\":\"Steven Spielberg\"},{\"op\":\"remove\",\"path\":\"/year\",\"value\":1993},{\"op\":\"remove\",\"path\":\"/genre\",\"value\":\"Drama\"}]",
          "audit_node_host_address": "127.0.0.1",
          "audit_request_effective_user": "admin",
          "audit_trace_shard_id": 0,
          "audit_trace_indices": [
            "movies"
          ],
          "audit_trace_resolved_indices": [
            "movies"
          ],
          "audit_node_host_name": "127.0.0.1"
        },
        "sort": [
          1768831798129
        ]
      }
```
#### Test Case 4: Delete Movies Index
```
curl -XDELETE "https://localhost:9200/movies?pretty" \
  -u 'admin:Rishav@123' \
  --insecure
```
Audit Log Entry:
``` json
 {
        "_index" : "security-auditlog-2026.01.19",
        "_id" : "Ou621psBU4Aun40RY1-c",
        "_score" : null,
        "_source" : {
          "audit_cluster_name" : "opensearch",
          "audit_node_name" : "c889f3ea09d1",
          "audit_trace_task_id" : "5oKZgVZwSeun0L0mlMTolg:123",
          "audit_transport_request_type" : "DeleteIndexRequest",
          "audit_category" : "INDEX_EVENT",
          "audit_request_origin" : "REST",
          "audit_node_id" : "5oKZgVZwSeun0L0mlMTolg",
          "audit_request_layer" : "TRANSPORT",
          "@timestamp" : "2026-01-19T14:43:55.931+00:00",
          "audit_format_version" : 4,
          "audit_request_remote_address" : "::1",
          "audit_request_privilege" : "indices:admin/delete",
          "audit_node_host_address" : "127.0.0.1",
          "audit_request_effective_user" : "admin",
          "audit_trace_indices" : [
            "movies"
          ],
          "audit_trace_resolved_indices" : [
            "movies"
          ],
          "audit_node_host_name" : "127.0.0.1"
        },
        "sort" : [
          1768833835931
        ]
      }
```

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
+++ b/src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
@@ -67,44 +67,70 @@ public final class ComplianceIndexingOperationListenerImpl extends ComplianceInd
 
     private static final ThreadLocal<Context> threadContext = new ThreadLocal<Context>();
 
-    @Override
-    public Delete preDelete(final ShardId shardId, final Delete delete) {
-        if (isLoggingWriteDiffEnabled(auditlog.getComplianceConfig(), shardId.getIndexName())) {
-            Objects.requireNonNull(is);
-
-            final IndexShard shard;
-
-            if (delete.origin() != org.opensearch.index.engine.Engine.Operation.Origin.PRIMARY) {
-                return delete;
-            }
-
-            if ((shard = is.getShardOrNull(shardId.getId())) == null) {
-                return delete;
-            }
-
-            if (shard.isReadAllowed()) {
-                try (ThreadContext.StoredContext ctx = threadPool.getThreadContext().stashContext()) {
-                    threadPool.getThreadContext().putHeader(OPENDISTRO_SECURITY_CONF_REQUEST_HEADER, "true");
-                    // Use getForUpdate to retrieve the document before deletion
-                    final GetResult getResult = shard.getService()
-                        .getForUpdate(delete.id(), delete.getIfSeqNo(), delete.getIfPrimaryTerm());
-                    if (getResult.isExists()) {
-                        threadContext.set(new Context(getResult));
-                    } else {
-                        threadContext.set(new Context(null));
-                    }
-                } catch (Exception e) {
-                    if (log.isDebugEnabled()) {
-                        log.debug("Cannot retrieve original document due to {}", e.toString());
-                    }
-                }
-            } else {
-                if (log.isDebugEnabled()) {
-                    log.debug("Cannot read from shard {}", shardId);
-                }
-            }
+    /**
+     * Attempts to retrieve the original document from the shard before a write operation.
+     * This is a best-effort operation for diff logging - if retrieval fails, the operation proceeds normally.
+     *
+     * @param shardId The shard ID
+     * @param documentId The document ID
+     * @param ifSeqNo The sequence number for optimistic concurrency control
+     * @param ifPrimaryTerm The primary term for optimistic concurrency control
+     * @param origin The operation origin (must be PRIMARY for logging)
+     */
+    private void retrieveOriginalDocumentForDiff(
+        final ShardId shardId,
+        final String documentId,
+        final long ifSeqNo,
+        final long ifPrimaryTerm,
+        final org.opensearch.index.engine.Engine.Operation.Origin origin
+    ) {
+        if (!isLoggingWriteDiffEnabled(auditlog.getComplianceConfig(), shardId.getIndexName())) {
+            return;
         }
 
+        Objects.requireNonNull(is);
+
+        if (origin != org.opensearch.index.engine.Engine.Operation.Origin.PRIMARY) {
+            return;
+        }
+
+        final IndexShard shard = is.getShardOrNull(shardId.getId());
+        if (shard == null) {
+            return;
+        }
+
+        if (shard.isReadAllowed()) {
+            try (ThreadContext.StoredContext ctx = threadPool.getThreadContext().stashContext()) {
+                threadPool.getThreadContext().putHeader(OPENDISTRO_SECURITY_CONF_REQUEST_HEADER, "true");
+                final GetResult getResult = shard.getService().getForUpdate(documentId, ifSeqNo, ifPrimaryTerm);
+                threadContext.set(new Context(getResult.isExists() ? getResult : null));
+            } catch (Exception e) {
+                if (log.isDebugEnabled()) {
+                    log.debug("Cannot retrieve original document due to {}", e.toString());
+                }
+            }
+        } else {
+            if (log.isDebugEnabled()) {
+                log.debug("Cannot read from shard {}", shardId);
+            }
+        }
+    }
+
+    /**
+     * Retrieves the previous document content from thread context and cleans up.
+     *
+     * @return The previous document content, or null if not available
+     */
+    private GetResult retrieveAndCleanupContext() {
+        final Context context = threadContext.get();
+        final GetResult previousContent = context == null ? null : context.getGetResult();
+        threadContext.remove();
+        return previousContent;
+    }
+
+    @Override
+    public Delete preDelete(final ShardId shardId, final Delete delete) {
+        retrieveOriginalDocumentForDiff(shardId, delete.id(), delete.getIfSeqNo(), delete.getIfPrimaryTerm(), delete.origin());
         return delete;
     }
 
@@ -118,9 +144,7 @@ public final class ComplianceIndexingOperationListenerImpl extends ComplianceInd
                 && delete.origin() == org.opensearch.index.engine.Engine.Operation.Origin.PRIMARY) {
 
                 if (complianceConfig.shouldLogDiffsForWrite()) {
-                    final Context context = threadContext.get();
-                    final GetResult previousContent = context == null ? null : context.getGetResult();
-                    threadContext.remove();
+                    final GetResult previousContent = retrieveAndCleanupContext();
                     auditlog.logDocumentDeleted(shardId, delete, result, previousContent);
                 } else {
                     auditlog.logDocumentDeleted(shardId, delete, result, null);
@@ -136,40 +160,7 @@ public final class ComplianceIndexingOperationListenerImpl extends ComplianceInd
 
     @Override
     public Index preIndex(final ShardId shardId, final Index index) {
-        if (isLoggingWriteDiffEnabled(auditlog.getComplianceConfig(), shardId.getIndexName())) {
-            Objects.requireNonNull(is);
-
-            final IndexShard shard;
-
-            if (index.origin() != org.opensearch.index.engine.Engine.Operation.Origin.PRIMARY) {
-                return index;
-            }
-
-            if ((shard = is.getShardOrNull(shardId.getId())) == null) {
-                return index;
-            }
-
-            if (shard.isReadAllowed()) {
-                try (ThreadContext.StoredContext ctx = threadPool.getThreadContext().stashContext()) {
-                    threadPool.getThreadContext().putHeader(OPENDISTRO_SECURITY_CONF_REQUEST_HEADER, "true");
-                    final GetResult getResult = shard.getService().getForUpdate(index.id(), index.getIfSeqNo(), index.getIfPrimaryTerm());
-                    if (getResult.isExists()) {
-                        threadContext.set(new Context(getResult));
-                    } else {
-                        threadContext.set(new Context(null));
-                    }
-                } catch (Exception e) {
-                    if (log.isDebugEnabled()) {
-                        log.debug("Cannot retrieve original document due to {}", e.toString());
-                    }
-                }
-            } else {
-                if (log.isDebugEnabled()) {
-                    log.debug("Cannot read from shard {}", shardId);
-                }
-            }
-        }
-
+        retrieveOriginalDocumentForDiff(shardId, index.id(), index.getIfSeqNo(), index.getIfPrimaryTerm(), index.origin());
         return index;
     }
 
@@ -188,9 +179,7 @@ public final class ComplianceIndexingOperationListenerImpl extends ComplianceInd
         }
 
         if (complianceConfig.shouldLogDiffsForWrite()) {
-            final Context context = threadContext.get();
-            final GetResult previousContent = context == null ? null : context.getGetResult();
-            threadContext.remove();
+            final GetResult previousContent = retrieveAndCleanupContext();
             Objects.requireNonNull(is);
 
             if (result.getFailure() != null || index.origin() != org.opensearch.index.engine.Engine.Operation.Origin.PRIMARY) {


src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
```diff
@@ -67,6 +67,47 @@ public GetResult getGetResult() {
...
+                        threadContext.set(new Context(null));
+                    }
+                } catch (Exception e) {
+                    if (log.isDebugEnabled()) {
+                        log.debug("Cannot retrieve original document due to {}", e.toString());
+                    }
+                }
+            } else {
+                if (log.isDebugEnabled()) {
+                    log.debug("Cannot read from shard {}", shardId);
```
nibix: When does this happen? Is this an error condition? If it is an error condition, we possibly should raise the log level.

src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
```diff
@@ -67,6 +67,47 @@ public GetResult getGetResult() {
...
+                    final GetResult getResult = shard.getService()
+                        .getForUpdate(delete.id(), delete.getIfSeqNo(), delete.getIfPrimaryTerm());
+                    if (getResult.isExists()) {
+                        threadContext.set(new Context(getResult));
+                    } else {
+                        threadContext.set(new Context(null));
+                    }
+                } catch (Exception e) {
+                    if (log.isDebugEnabled()) {
+                        log.debug("Cannot retrieve original document due to {}", e.toString());
```
nibix: Is this expected to happen? If it is not expected to happen, we should use a higher log level.

src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
Rishav9852Kumar: This occurs when `shard.isReadAllowed()` returns `false`, which happens during normal shard lifecycle events

it's an expected scenario during normal cluster operations. The compliance logging feature attempts to retrieve the original document for diff logging, but if the shard isn't ready for reads, it gracefully skips this step. 

this same pattern is used consistently in preIndex() method: ref https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java#L114

nibix: ah, ok, that means in such cases I do not get a diff stored in the audit log, but just the complete new doc?

src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
Rishav9852Kumar: Yes, this is expected to happen in several scenarios.
In `preDelete()`, we're attempting to retrieve the document's content __before__ it gets deleted so we can log the diff (what was removed). The DEBUG level here is appropriate because the exception does __NOT__ prevent the delete operation from proceeding.

this same pattern is used consistently in preIndex() method: ref https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java#L109

src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
Rishav9852Kumar: it depends on the operation (write or delete) but here  in case of delete yes,
No diff is logged, but the audit trail still shows that the delete happened and who did it, just without the detail of what was deleted.
something like how the originally delete action was logged 
```json
{
  "_index": "security-auditlog-2026.01.15",
  "_id": "gKgbw5sBaJ38aiaWiz2U",
  "_source": {
    "audit_compliance_operation": "DELETE",
    "audit_cluster_name": "opensearch",
    "audit_node_name": "c889f3ea09d1",
    "audit_category": "COMPLIANCE_DOC_WRITE",
    "audit_request_origin": "REST",
    "audit_compliance_doc_version": 2,
    "audit_node_id": "5oKZgVZwSeun0L0mlMTolg",
    "@timestamp": "2026-01-15T19:22:00.976+00:00",
    "audit_format_version": 4,
    "audit_request_remote_address": "::1",
    "audit_trace_doc_id": "1",
    "audit_node_host_address": "127.0.0.1",
    "audit_request_effective_user": "admin",
    "audit_trace_shard_id": 0,
    "audit_trace_indices": ["movies"],
    "audit_trace_resolved_indices": ["movies"],
    "audit_node_host_name": "127.0.0.1"
  }
}
```

similarly in write operations we get the operation metadata but lose the diff showing what changed (but just the complete new doc).
However, this is a very edge case - it only occurs when the shard is not in a readable state

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -691,16 +693,99 @@ public void logDocumentDeleted(ShardId shardId, Delete delete, DeleteResult resu
...
+        msg.addId(id);
         msg.addShardId(shardId);
         msg.addComplianceDocVersion(result.getVersion());
         msg.addComplianceOperation(Operation.DELETE);
+
+        if (complianceConfig.shouldLogDiffsForWrite() && originalResult != null && originalResult.isExists()) {
+            try {
+                String originalSource = null;
+
+                if (securityIndex.equals(shardId.getIndexName())) {
```
cwperks, line 712: I know this is the same as `logDocumentWritten`, but curious why the logic for parsing is so different when I would imagine the only difference should be `addSecurityConfigWriteDiffSource` vs `addComplianceWriteDiffSource`

src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
```diff
@@ -67,6 +67,47 @@ public GetResult getGetResult() {
...
+
+            if (delete.origin() != org.opensearch.index.engine.Engine.Operation.Origin.PRIMARY) {
+                return delete;
+            }
+
+            if ((shard = is.getShardOrNull(shardId.getId())) == null) {
+                return delete;
+            }
+
+            if (shard.isReadAllowed()) {
```
cwperks: Can we abstract some of the common logic to re-usable functions?

Rishav9852Kumar: sure , initialy i wanted to keep the new changes seperate from existing code,  but yes can create a helper method to reuse the code for both write and delete.

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
Rishav9852Kumar: Yes the parsing logic should be identical. The difference is largely historical. 
i can raise a follow up cr to extract the common parsing logic into helper methods (for both write and delete) or add them in same pr?

Personally, I'd prefer to address it separately to keep this PR focused on the delete diff logging feature

cwperks: Yes, that's fine for this PR. This is the first time I really looked at the code in detail and wonder if treating the security index separately is really necessary apart from the event type in the audit log where we do treat the security index as a special case.

src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
nibix: For me to understand better: In which scenarios does this happen?

Rishav9852Kumar: - __New document creation__ - No previous version exists to retrieve

- __Version conflicts__ - Document modified between operation start and retrieval 

- __Race conditions__ - Document deleted by concurrent operation

- __Shard state transitions__ - During relocation, failover, or recovery

These are normal conditions that don't prevent the write operation from succeeding, but : - 
- The audit logging still works (captures metadata like who, what, when)

- Only the detailed "diff" part (showing actual content changes) is affected

src/test/java/org/opensearch/security/auditlog/compliance/ComplianceAuditlogTest.java
```diff
@@ -517,4 +517,54 @@ public void testWriteLogDiffsEnabledAndLogRequestBodyDisabled() throws Exception
...
+            ComplianceConfig.from(
+                ImmutableMap.of("enabled", true, "write_watched_indices", Collections.singletonList("movies"), "write_log_diffs", true),
+                additionalSettings
+            )
+        );
+        updateAuditConfig(AuditTestUtils.createAuditPayload(auditConfig));
+
+        // Create index and document
+        try (Client tc = getClient()) {
+            rh.executePutRequest("movies", "{\"settings\": {\"index\": {\"number_of_shards\": 1, \"number_of_replicas\": 0}}}");
```
cwperks, line 542: nit: prefer to use text blocks for JSON

src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
```diff
@@ -67,6 +67,73 @@ public GetResult getGetResult() {
...
+     * @param ifSeqNo The sequence number for optimistic concurrency control
+     * @param ifPrimaryTerm The primary term for optimistic concurrency control
+     * @param origin The operation origin (must be PRIMARY for logging)
+     */
+    private void retrieveOriginalDocumentForDiff(
+        final ShardId shardId,
+        final String documentId,
+        final long ifSeqNo,
+        final long ifPrimaryTerm,
+        final org.opensearch.index.engine.Engine.Operation.Origin origin
```
RajatGupta02, line 85: nit: can we avoid classnames and use imports instead

src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
```diff
@@ -67,6 +67,73 @@ public GetResult getGetResult() {
...
+    ) {
+        if (!isLoggingWriteDiffEnabled(auditlog.getComplianceConfig(), shardId.getIndexName())) {
+            return;
+        }
+
+        Objects.requireNonNull(is);
+
+        if (origin != org.opensearch.index.engine.Engine.Operation.Origin.PRIMARY) {
```
RajatGupta02, line 93: same as above, can use import

Rishav9852Kumar: I agree these are good improvements for code readability. However, since these patterns (fully qualified class names) were part of the existing/traditional code in this file, 
ref : - https://github.com/opensearch-project/security/blob/0c67b1939f845e27d5764cd6436c025019d27046/src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java#L90

and the current PR focuses on [Enable audit logging of document contents for DELETE operations]

I'll create a follow-up PR specifically for these style improvements:

src/main/java/org/opensearch/security/compliance/ComplianceIndexingOperationListenerImpl.java
Rishav9852Kumar: will cover in follow-up PR specifically for these style improvements

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -691,16 +693,99 @@ public void logDocumentDeleted(ShardId shardId, Delete delete, DeleteResult resu
...
+                    );
+                    msg.addComplianceWriteDiffSource(diffnode.size() == 0 ? "" : diffnode.toString());
+                }
+            } catch (Exception e) {
+                log.error("Unable to generate diff for delete operation {}", msg.toPrettyString(), e);
+            }
+        }
+
+        if (!complianceConfig.shouldLogWriteMetadataOnly()
+            && !complianceConfig.shouldLogDiffsForWrite()
```
shikharj05, line 757: can you help me understand this case?  do we need to log diffs even if ``write_log_diffs`` is disabled?

let me know if I have misread

Rishav9852Kumar: No, diffs are NOT logged when write_log_diffs is disabled.

When both write_log_diffs and write_metadata_only are disabled, the code logs the full document content instead of diffs.

It follows the same pattern as in `logDocumentWritten` 

https://github.com/opensearch-project/security/blob/0c67b1939f845e27d5764cd6436c025019d27046/src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java#L638

### Three Logging Modes for Write Operations:

```
// MODE 1: Log DIFFS (when write_log_diffs is ENABLED) [by default it is disabled]
if (complianceConfig.shouldLogDiffsForWrite()) {
    Compute and log diffs
}

// MODE 2: Log FULL CONTENT (when BOTH flags are DISABLED)
if (!complianceConfig.shouldLogWriteMetadataOnly() 
    && !complianceConfig.shouldLogDiffsForWrite()) {
    Log full document content
}

// MODE 3: Log METADATA ONLY (when write_metadata_only is ENABLED)[by default it is enabled]
// Neither block executes, only basic audit fields are logged

```

To conclude 

- When write_log_diffs is disabled, diffs are NOT logged. 
- When write_log_diffs is enabled write_metadata_only is disabled, diffs are logged.
- When write_log_diffs is enabled write_metadata_only is enabled, full document content is logged

** hope it was clear

cwperks: > When write_log_diffs is enabled write_metadata_only is enabled, full document content is logged

Note: This is only the case when `log_request_body` is set to true.

If its set to false, then there's no trace of the request body and instead the auditlog has high level info like HTTP Verb, route, etc.

PR #5929, opened 2026-01-29
[Bugfix] Make gRPC JWT header keys case insensitive

finnegancarroll:
### Description

Follow up on some pending items form the initial JWT authentication PR:
https://github.com/opensearch-project/security/pull/5916

1. While testing the full JWT happy path on a local single node cluster I notice the authorization header is case sensitive over the gRPC transport. The integration tests miss this case as they configure a lowercase "jwt-auth" header. Resolved this by moving the `GrpcRequestChannel` implementation to match the `TreeMap<>(String.CASE_INSENSITIVE_ORDER)` header map used by [NettyRequest](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/filter/NettyRequest.java#L50-L54).

2. Reverted the backend registry license header which I accidentally changed to the updated header used in new files.

3. Remove a small check which skips super user auth for gRPC in the backend registry. There is no need for this since we never attempt to extract `SSLInfo` from gRPC requests.

### Issues Resolved
Follow up for post merge issues here:
https://github.com/opensearch-project/security/pull/5916

### Testing

### Check List
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md)

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/grpc/GrpcAnonymousAuthTest.java
+++ b/src/integrationTest/java/org/opensearch/security/grpc/GrpcAnonymousAuthTest.java
@@ -112,7 +112,7 @@ public class GrpcAnonymousAuthTest {
                 doBulk(channelWithAuth, "test-invalid-auth", 2);
                 fail("Expected authentication failure - invalid auth header");
             } catch (StatusRuntimeException e) {
-                assertEquals(Status.Code.INVALID_ARGUMENT, e.getStatus().getCode());
+                assertEquals(Status.Code.UNAUTHENTICATED, e.getStatus().getCode());
             }
         } finally {
             channel.shutdown();


src/integrationTest/java/org/opensearch/security/grpc/JWTGrpcDefaultAuthHeaderTest.java
```diff
@@ -0,0 +1,144 @@
...
+        String jwtToken = createValidJwtToken("grpc_user", "grpc_index_role");
+        ManagedChannel channel = secureChannel(getSecureGrpcEndpoint(cluster));
+
+        try {
+            ClientInterceptor authInterceptor = createHeaderInterceptor(Map.of("authorization", "Bearer " + jwtToken));
+            Channel channelWithAuth = io.grpc.ClientInterceptors.intercept(channel, authInterceptor);
+            var bulkResp = doBulk(channelWithAuth, "test-grpc-index-lower", 2);
+            assertNotNull(bulkResp);
+            assertFalse(bulkResp.getErrors());
+            assertEquals(2, bulkResp.getItemsCount());
```
cwperks, line 125: Do we test out the case where one item of the bulk succeeds and the other fails?

Security treats this case in 2 different ways:

1. `dnfof: false` - This is the default and if one of the items in the bulk is unauthorized then the whole bulk fails
2. `dnfof: true` - In this case, the bulk item that's permitted succeeds and the other items (silently) fail.

cwperks (approved):

PR #5934, opened 2026-01-31
Refactored config REST API tests to avoid unsafe patterns causing memory leaks

### LINKED ISSUE #5915: [Tests] Integration test issues
It seems that the integration tests got a bit more unreliable recently. In some cases, they time out after 360 minutes. 

We need to research the reason.

nibix:
### Description

In #5915, we discovered that the config REST API integration tests cause memory leaks as they use static references across test class boundaries in an unsafe manner.

This PR refactores these tests so that these no longer share static cluster or mutable static configuration references across test classes. IMHO, one should strive for abolishing the abstract super classes for these tests alltogether, but this is outside of my time availability.

* Category: Test fix
  * Fixes memory leaks in integration tests
  * No behavioral changes

### Issues Resolved

Fixes #5915

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/api/AbstractApiIntegrationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/api/AbstractApiIntegrationTest.java
@@ -11,35 +11,21 @@
 
 package org.opensearch.security.api;
 
-import java.io.IOException;
-import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.StringJoiner;
 
-import org.apache.commons.io.FileUtils;
 import org.apache.http.HttpStatus;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-import org.awaitility.Awaitility;
-import org.junit.AfterClass;
-import org.junit.Before;
 
-import org.opensearch.common.CheckedConsumer;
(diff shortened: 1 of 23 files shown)


src/integrationTest/java/org/opensearch/security/api/AbstractApiIntegrationTest.java
```diff
@@ -112,121 +95,28 @@ private static String randomAsciiAlphanumOfLength(final int length) {
...
-                .until(() -> client.securityHealth().getTextFromJsonBody("/status"), equalTo("UP"));
-        }
-        testClass = this.getClass();
+            .authc(TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL)
+            .users(ADMIN_USER, REST_ADMIN_USER, NEW_USER)
+            .roles(EXAMPLE_ROLE, REST_ADMIN_REST_API_ACCESS_ROLE);
     }
 
-    protected Map<String, Object> getClusterSettings() {
+    protected static Map<String, Object> getClusterSettings() {
```
cwperks, line 106: nit: better as a static constant than a method?

nibix: yeah, i guess one could even inline it. Still, I think there's quite a few stuff around that could be polished more. But due to the volume, I actually think that we need to do it incrementally.

Issue #5812, opened 2025-11-24
[FEATURE] JWT support for auxiliary transports (gRPC)


finnegancarroll:
**Is your feature request related to a problem?**
Provide support for the gRPC transport module, implemented as an auxiliary transport in core, to authenticate users with Java Web Tokens.

**What solution would you like?**
Security plugin should inject a gRPC interceptor into the gRPC transport through the extensible plugin framework.

Requests passing through this interceptor should have JWT tokens read from HTTP headers, and authenticated with the back end provider.

**What alternatives have you considered?**
Not supporting JWT for auxiliary transports.

**Do you have any additional context?**
Upstream issue: https://github.com/opensearch-project/security/issues/5785

PR #5940, opened 2026-02-03
Set gRPC security interceptor to highest priority

aparajita31pandey:
### Description
Prioritize security gRPC interceptor with minimum order value
* Conceptually  gRPCInterceptor follows similar  pattern like ActionFilter
* SecurityFilter which implements ActionFilter in opensearchSecurity has ordering on **Integer.MIN_VALUE** -https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/filter/SecurityFilter.java#L173  following similar in `SecurityGrpcFilter`

### Issues Resolved
[#5812]

### Testing

 Unit Test Added 
### Check List


cwperks (dismissed):

cwperks (approved):

finnegancarroll:
> For HTTP, it is important to be able to do authn before decompression is applied to avoid unauthenticated DDOS attacks. I do not know so much about gRPC, but that might be a similar consideration there, right?

I think my compression example was not great but maybe some other use cases which would need to execute before security plugin could be:
- Custom rate limiting (Security plugin has limited failed login support)
- Observability use cases where we want information about every request, not just those authenticated
- User injection use cases where we authenticate a user with a proprietary/unsupported auth backend and want to take advantage of security plugin FGAC.

By making the gRPC filter max prio we remove the above use cases. Is the idea to protect future plugin developers from accidentally registering an insecure intercepter or doing unnecessary work ahead of the grpc security filter?

DarshitChanpura:
@finnegancarroll @aparajita31pandey Can we address integ test failures if this PR is being actively worked on? If not this PR will be closed during next cleanup.

PR #5971, opened 2026-02-25
Add buildSecureTransportContext() to SecureTransportSettingsProvider for Arrow Flight TLS cert reload

rishabhmaurya:
### Description
Exposes the live in-memory javax.SSLContext from SslContextHandler via a new buildSecureTransportContext() method on SecureTransportSettingsProvider, enabling Arrow Flight transport to wrap it in JdkSslContext without reading cert files from disk after reloadcerts

### Issues Resolved

### Testing

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java
+++ b/src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java
@@ -122,6 +122,11 @@ public class OpenSearchSecureSettingsFactory implements SecureSettingsFactory {
             }
 
             @Override
+            public Optional<SSLContext> buildSecureTransportContext(Settings settings) {
+                return sslSettingsManager.sslContextHandler(CertType.TRANSPORT).map(SslContextHandler::tryFetchSSLContext);
+            }
+
+            @Override
             public Optional<SSLEngine> buildSecureServerTransportEngine(Settings settings, Transport transport) throws SSLException {
                 return sslSettingsManager.sslContextHandler(CertType.TRANSPORT).map(SslContextHandler::createSSLEngine);
             }


src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java
```diff
@@ -121,6 +121,11 @@ public Optional<TrustManagerFactory> trustManagerFactory() {
                 });
             }
 
+            @Override
+            public Optional<SSLContext> buildSecureTransportContext(Settings settings) {
```
reta, line 125: @rishabhmaurya for SSLContext, please use this hook `Optional<SecureTransportParameters> parameters(Settings settings)` - in this case you have full control over the `SSLContext`, examples are here (for HTTP transport but the idea is the same): 
 - https://github.com/opensearch-project/OpenSearch/blob/main/modules/transport-netty4/src/main/java/org/opensearch/http/netty4/Netty4Http3ServerTransport.java#L281 
 - https://github.com/opensearch-project/OpenSearch/blob/main/plugins/transport-reactor-netty4/src/main/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpServerTransport.java#L429

reta (changes requested):

src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java
rishabhmaurya: @reta thanks for feedback. We were using  `Optional<SecureTransportParameters> parameters(Settings settings)` in arrow-flight-rpc plugin - https://github.com/opensearch-project/OpenSearch/blob/main/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/bootstrap/tls/DefaultSslContextProvider.java#L76 but this method has a fundamental issue. It assumes certificate files are always persisted, so in case of new connection and reloaded certificates, if the certs are deleted from node (which can happen for security reasons) once security plugins applies them, then this approach doesn't work. 

So we are exposing a similar builder here as what [SecureAuxTransportSettingsProvider](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/plugins/SecureAuxTransportSettingsProvider.java#L32) does

DarshitChanpura: If we are planning to merge this to unblock 3.5, can we add a TODO here and follow-up with a fix that @reta suggested?

rishabhmaurya: I agree, if he thinks this is the not correct way to address it.

DarshitChanpura (approved):
Approving it to unblock 3.5, as feature itself is experimental

src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java
reta: Got it thanks @rishabhmaurya , I think if we could limit the amount of APIs (`SSLEngine` for out of the box engine, and parameters for custom SSLContext), that would be beneficial, because we now have 3 different ways to configure same SSL.

reta (approved):

src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java
rishabhmaurya: you are right. The main reason for having 3 different APIs is because each transport netty/grpc/flight configure ssl differently. Let me revisit this API on how it can be improved.

Issue #1310, opened 2021-07-13
Add ability to change default DLS parameter substitution behavior


rofleksey:
Currently, if DLS contains a parameter that is not quoted and is not defined anywhere, i get this error:
`Unrecognized token '$': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')`

This is happening, because [replaceParameters](https://github.com/opensearch-project/security/blob/2b69d9c84d807fc68b3ad9770d8bb75f6b99e725/src/main/java/org/opensearch/security/securityconf/ConfigModelV7.java#L906) method replaces only defined parameters, but does nothing to at least replace undefined ones with ''. This doesn't result in error if all the parameters are quoted, because even though they are not being replaced, the JSON structure still stays syntactically correct. But if there is at least one unquoted parameter, I get this error.

For example, if I have this role configuration:

```yaml
{
  "index_permissions": [
    {
      ...
      "dls": "{\"somefield\":{\"arr\":[${attr.jwt.array}]}}",
      ...
  ],
  ...
}
```

And JWT payload looks like: 

```yaml
{
  "array": "\"1\", \"2\", \"3\"",
}
```

Then everything works correctly. But if for some reason I don't define this `array` field in JWT payload, I get the error described above.

I think there should be an option to tweak this behavior:

1. do nothing to undefined parameters (default)
2. replace undefined parameters with ''
3. throw an exception if undefined parameters are detected

andy840314:
I think offering options to tweak this is confusing for the users. However, this error message 
`Unrecognized token '$': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')` 
is not really reflecting the root cause. Throwing an exception should be a better solution.

scrawfor99:
[Triage] This request remains relevant. Error/exception message should be swapped to be more relevant. Tagging as "help wanted" and "good first issue."

PR #5975, opened 2026-03-01
fixed throw exception for DLS user attributes

ThyTran1402:
### Description
When a DLS query references a user attribute ${attr.jwt.array} that is not present in the authenticated user's JWT/custom attributes, UserAttributes.replaceProperties() leaves the placeholder in the query string. The guard in DocumentPrivileges.Dynamic.evaluate() correctly detects this and throws a PrivilegesEvaluationException, but the previous error message ("Invalid DLS query: <full query string>") gave operators no indication of which attribute was missing or what attributes were available — forcing manual inspection of both the query and the user's token.
- Before: PrivilegesEvaluationException: Invalid DLS query: {"terms":{"arr":["${attr.jwt.array}"]}}
- Fixed: PrivilegesEvaluationException: DLS query references undefined user attributes: [attr.jwt.array]. Available attributes are: [attr.jwt.roles, attr.proxy.department]

### Issues Resolved
- #1310 

### Testing

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Fix the issue of unprocessed X-Request-Id ([#5954](https://github.com/opensearch-project/security/pull/5954))
 - Improve DLS error message to identify undefined user attributes when query substitution fails ([#5975](https://github.com/opensearch-project/security/pull/5975))
-
 - Fix the issue of unprocessed X-Request-Id ([#5976](https://github.com/opensearch-project/security/pull/5976))
 ### Refactoring
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fix audit log writing errors for rollover-enabled alias indices ([#5878](https://github.com/opensearch-project/security/pull/5878)
 
 - Fix the issue of unprocessed X-Request-Id ([#5954](https://github.com/opensearch-project/security/pull/5954))
+- Improve DLS error message to identify undefined user attributes when query substitution fails ([#5975](https://github.com/opensearch-project/security/pull/5975))
+
+- Fix the issue of unprocessed X-Request-Id ([#5976](https://github.com/opensearch-project/security/pull/5976))
 ### Refactoring
 
 ### Maintenance
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Fix the issue of unprocessed X-Request-Id ([#5954](https://github.com/opensearch-project/security/pull/5954))
 - Improve DLS error message to identify undefined user attributes when query substitution fails ([#5975](https://github.com/opensearch-project/security/pull/5975))
-- Fix the issue of unprocessed X-Request-Id ([#5976](https://github.com/opensearch-project/security/pull/5976))
 ### Refactoring
 
 ### Maintenance
--- a/src/main/java/org/opensearch/security/privileges/UserAttributes.java
+++ b/src/main/java/org/opensearch/security/privileges/UserAttributes.java
@@ -10,8 +10,13 @@
  */
 package org.opensearch.security.privileges;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.Iterables;
@@ -25,8 +30,23 @@ import org.opensearch.security.user.User;
  * This code was moved over from ConfigModelV7.
  */
 public class UserAttributes {
+    private static final Pattern UNRESOLVED_ATTRIBUTE_PATTERN = Pattern.compile("\\$\\{([^}]+)\\}");
+
     public static boolean needsAttributeSubstitution(String patternString) {
         return patternString.contains("${");
+    }
+
+    /**
+     * Returns the names of all unresolved ${...} attribute references remaining
+     * in {@code s} after substitution has been performed.
+     */
+    public static List<String> findUnresolvedAttributes(String s) {
+        List<String> result = new ArrayList<>();
+        Matcher matcher = UNRESOLVED_ATTRIBUTE_PATTERN.matcher(s);
+        while (matcher.find()) {
+            result.add(matcher.group(1));
+        }
+        return Collections.unmodifiableList(result);
     }
 
     public static String replaceProperties(String orig, PrivilegesEvaluationContext context) {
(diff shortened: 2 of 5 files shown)


src/main/java/org/opensearch/security/privileges/dlsfls/DocumentPrivileges.java
```diff
@@ -176,8 +177,11 @@ static class Dynamic extends DlsQuery {
             RenderedDlsQuery evaluate(PrivilegesEvaluationContext context) throws PrivilegesEvaluationException {
                 String effectiveQueryString = UserAttributes.replaceProperties(this.queryString, context);
                 if (UserAttributes.needsAttributeSubstitution(effectiveQueryString)) {
+                    List<String> unresolved = UserAttributes.findUnresolvedAttributes(effectiveQueryString);
+                    Set<String> available = context.getUser().getCustomAttributesMap().keySet();
                     throw new PrivilegesEvaluationException(
-                        "Invalid DLS query: " + effectiveQueryString,
+                        "DLS query references undefined user attributes: " + unresolved
+                            + ". Available attributes are: " + available,
```
cwperks: Should we consider ellipsizing `available` after a certain amount of characters because I believe this can be quite large in some instances.

ThyTran1402: Yup. That's a good point. We can limit the number of attributes shown rather than truncating by character count. That way keeps the message more readable. Do you have any suggestions on the available limit?

ThyTran1402: I think I can set the limit to 10 for now. Please let me know if that works.

cwperks: yea 10 is reasonable. If anyone wants to support customization in the future then we could entertain a setting but I think 10 is a good default.

src/test/java/org/opensearch/security/privileges/UserAttributesUnitTest.java
```diff
@@ -71,4 +71,32 @@ public void testReplaceProperties() {
...
+            UserAttributes.findUnresolvedAttributes("{\"terms\":{\"arr\":[${attr.jwt.array}]}}")
+        );
+    }
+
+    @Test
+    public void testFindUnresolvedAttributes_multipleTokens() {
+        assertEquals(
+            List.of("attr.jwt.dept", "attr.proxy.region"),
+            UserAttributes.findUnresolvedAttributes(
+                "{\"bool\":{\"must\":[{\"term\":{\"dept\":\"${attr.jwt.dept}\"}},{\"term\":{\"region\":\"${attr.proxy.region}\"}}]}}"
```
cwperks: nit: since the min java version supported is 21 we can use TextBlocks. For JSON blocks I think TextBlocks make the json much more readable.

cwperks (dismissed):

cwperks:
@ThyTran1402 one other thing, can you run `./gradlew spotlessApply` and commit and push the formatting fixes? I see Code Hygiene check is failing.

src/main/java/org/opensearch/security/privileges/dlsfls/DocumentPrivileges.java
```diff
@@ -176,8 +178,14 @@ static class Dynamic extends DlsQuery {
             RenderedDlsQuery evaluate(PrivilegesEvaluationContext context) throws PrivilegesEvaluationException {
                 String effectiveQueryString = UserAttributes.replaceProperties(this.queryString, context);
                 if (UserAttributes.needsAttributeSubstitution(effectiveQueryString)) {
+                    List<String> unresolved = UserAttributes.findUnresolvedAttributes(effectiveQueryString);
+                    Set<String> available = context.getUser().getCustomAttributesMap().keySet();
+                    String availableStr = available.size() > 10
```
DarshitChanpura: We should not be hard coding limits inline. Can you declare a class-level constant ?

CHANGELOG.md
```diff
@@ -15,7 +15,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Hardens input validation for resource sharing APIs ([#5831](https://github.com/opensearch-project/security/pull/5831)
 
 ### Bug Fixes
-- Fix the issue of unprocessed X-Request-Id ([#5954](https://github.com/opensearch-project/security/pull/5954))
+- Improve DLS error message to identify undefined user attributes when query substitution fails ([#5975](https://github.com/opensearch-project/security/pull/5975))
+
```
DarshitChanpura, line 19: nit: remove extra line

src/test/java/org/opensearch/security/privileges/UserAttributesUnitTest.java
ThyTran1402: hmm interesting I haven't thought about it. That makes lot of sense to me. I can update to make it more readable.

src/main/java/org/opensearch/security/privileges/dlsfls/DocumentPrivileges.java
ThyTran1402: That's a good point. I can declare a class and replace the hard coding 10 with constant.

cwperks (dismissed):

nibix (dismissed):

RyanL1997 (dismissed):

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fix audit log writing errors for rollover-enabled alias indices ([#5878](https://github.com/opensearch-project/security/pull/5878)
 
 - Fix the issue of unprocessed X-Request-Id ([#5954](https://github.com/opensearch-project/security/pull/5954))
+- Improve DLS error message to identify undefined user attributes when query substitution fails ([#5975](https://github.com/opensearch-project/security/pull/5975))
+- Fix the issue of unprocessed X-Request-Id ([#5976](https://github.com/opensearch-project/security/pull/5976))
```
cwperks:
```suggestion
```

cwperks (approved):

RyanL1997 (approved):

src/main/java/org/opensearch/security/privileges/UserAttributes.java
```diff
@@ -25,10 +30,25 @@
...
     }
 
+    /**
+     * Returns the names of all unresolved ${...} attribute references remaining
+     * in {@code s} after substitution has been performed.
+     */
+    public static List<String> findUnresolvedAttributes(String s) {
+        List<String> result = new ArrayList<>();
+        Matcher matcher = UNRESOLVED_ATTRIBUTE_PATTERN.matcher(s);
+        while (matcher.find()) {
```
rursprung, line 46: this whole method can be simplified to `return UNRESOLVED_ATTRIBUTE_PATTERN.matcher(s).results().map(m -> m.group(1)).toList();`
see [`Matcher#results`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/regex/Matcher.html#results())

(stumbled over it by chance while reviewing a bug)

ThyTran1402: hmmm that's interesting. It seems to be cleaner with Matcher#results(). 
I wonder should I open another PR for this suggestion since it already merged?

rursprung: i raised #6122

Issue #5976, opened 2026-03-01
[FEATURE] Add a way to split Log4j audit log messages when resolve_indices is enabled for clusters with many indices to prevent audit_trace_indices/audit_trace_resolved_indices fields making messages too big


parislarkins:
**What is the bug?**
For clusters with large numbers of indices (e.g. ~11,000 in the case I first observed this issue) with the `plugins.security.audit.config.resolve_indices` OpenSearch setting and `appender.rolling_audit.layout.maxMessageLength=0` log4j setting enabled (to disable log message truncation), the `audit_trace_indices` and `audit_trace_resolved_indices` fields in audit logs that relate to all indices can be extremely long. For example, 11,000 indexes with 100 character long names requires 1,100,000 characters to just to represent the index names alone (plus even more characters for the surrounding quotes and commas between each of them). 

These messages can get so large as to cause problems for downstream parts of your logging pipeline (for example, the default Apache Kafka maximum message is 1mb). In these cases, it is usually recommended to split large messages into smaller ones, as they are able to be handled more efficiently than giant messages.

It would be ideal if OpenSearch could be configured to split audit messages with huge numbers of index names (such as by specifying a maximum number of index name characters per log message) into multiple smaller messages, keeping all other audit message fields the same apart from `audit_trace_indices` and `audit_trace_resolved_indices`.  For a simple example, if we could set the maximum index name characters per message to `18`, this original message:
```json
{
  "audit_trace_indices": [
    "index*"
  ],
  "audit_trace_resolved_indices": [
    "index1",
    "index2",
    "index3",
    "index4",
    "index5",
    "index6",
    "index7",
    "index8"
  ],
  "audit_category": "AUTHENTICATED"
}
```
Would then be split into the following 3 messages:
```json
{
  "audit_trace_indices": [
    "index*"
  ]
  "audit_trace_resolved_indices": [
    "index2",
    "index3"
  ],
  "audit_category": "AUTHENTICATED"
}
```
```json
{
  "audit_trace_resolved_indices": [
    "index4",
    "index5",
    "index6"
  ],
  "audit_category": "AUTHENTICATED"
}
```
```json
{
  "audit_trace_resolved_indices": [
    "index7",
    "index8"
  ],
  "audit_category": "AUTHENTICATED"
}
```
These 3 split messages contain the exact same information as the source message, so no information is lost (although it is more effort to re-construct the original event).

**How can one reproduce the bug?**
[...]

PR #5977, opened 2026-03-02
Add new setting to control splitting log4j audit messages to prevent audit_trace_indices/audit_trace_resolved_indices from making messages too big

parislarkins:
Add new setting `plugins.security.audit.config.log4j.maximum_index_characters_per_message`, when non-zero split audit log messages audit_trace_indices and audit_trace_resolved_indices into multiple messages that are under the maximum limit.

### Description
**Enhancement/New feature?**

For clusters with large numbers of indices (e.g. ~11,000 in the case I first observed this issue) with the `plugins.security.audit.config.resolve_indices` OpenSearch setting and `appender.rolling_audit.layout.maxMessageLength=0` log4j setting enabled (to disable log message truncation), the `audit_trace_indices` and `audit_trace_resolved_indices` fields in audit logs that relate to all indices can be extremely long. For example, 11,000 indexes with 100 character long names requires 1,100,000 characters to just to represent the index names alone (plus even more characters for the surrounding quotes and commas between each of them). 

These messages can get so large as to cause problems for downstream parts of your logging pipeline (for example, the default Apache Kafka maximum message is 1mb). In these cases, it is usually recommended to split large messages into smaller ones, as they are able to be handled more efficiently than giant messages.

**Old behaviour**
Log4j audit messages log an unlimited number of index names in the `audit_trace_indices` and `audit_trace_resolved_indices` fields, leading to giant log messages for clusters with lots of indexes.

**New behaviour**
A simple example: If we set the new setting `plugins.security.audit.config.log4j.maximum_index_characters_per_message` to`18`, this original message:
```json
{
  "audit_trace_indices": [
    "index*"
  ],
  "audit_trace_resolved_indices": [
    "index1",
    "index2",
    "index3",
    "index4",
    "index5",
    "index6",
    "index7",
    "index8"
  ],
  "audit_category": "AUTHENTICATED"
}
```
Would then be split into the following 3 messages:
```json
{
  "audit_trace_indices": [
    "index*"
  ]
  "audit_trace_resolved_indices": [
    "index2",
    "index3"
  ],
  "audit_category": "AUTHENTICATED"
}
```
```json
{
  "audit_trace_resolved_indices": [
    "index4",
    "index5",
    "index6"
  ],
  "audit_category": "AUTHENTICATED"
}
```
```json
{
  "audit_trace_resolved_indices": [
    "index7",
    "index8"
  ],
  "audit_category": "AUTHENTICATED"
}
```
These 3 split messages contain the exact same information as the source message, so no information is lost (although it is more effort to re-construct the original event).

### Issues Resolved
https://github.com/opensearch-project/security/issues/5976

### Testing
Unit tests added to AuditMessageTest class.

I have been running my fork of the plugin using this setting on live clusters successfully to reduce the maximum size of audit log messages so they can fit in the 1mb maximum Apache Kafka message size in my logging pipeline. 

### Check List
- [ ] New functionality has been documented - raised https://github.com/opensearch-project/documentation-website/pull/12185


src/main/java/org/opensearch/security/auditlog/impl/AuditMessage.java
```diff
@@ -510,6 +517,100 @@ public String toString() {
         }
     }
 
+    public List<String> toJsonSplitIndices(final int maximumIndexCharsPerMessage) {
```
cwperks, line 520: @parislarkins I think we should be a bit more deterministic about this even if it means being less efficient. What I'm thinking is that [MAX_INDEX_NAME_LENGTH is 255 in OpenSearch](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java#L176). Can we figure out how many indices can fit into a messages assuming they are all max length and then separate the audit log messages into chunks of that size?

parislarkins: @cwperks this was my second approach (the first being a more expensive method which found the maximum number of messages possible), which was definitely a simpler implementation and much nicer to read! I went with the implementation here as a middle ground as in our deployment the longest index names we had were 118 chars long, so assuming max length was quite wasteful for us.

With the current logic and a `maximumIndexCharsPerMessage` value of `120000` (what we've used so far) this lets each message contain ~1000 index names compared to slightly less than 500 if we split assuming each index name was maximum length. In one of the clusters we deployed this to we had ~10,000 indexes, so the difference in splitting a message into ~10 vs 20 parts was pretty desirable!

Happy to defer to you on this though, or maybe there's a way we can allow the user to choose what sort of behaviour they'd like?

cwperks: Can we skip the logic to find the `longestIndexName` if maximumIndexCharsPerMessage == Integer.MAX_VALUE.

I want to avoid iterating all the indices if there will be no split (current behavior).

cwperks: otherwise I think what you have is reasonable and more efficient in terms of number of audit messages generated.

parislarkins: great suggestion, thanks! Have added an early-exit to immediately skip further logic when the setting is at its default value of MAX_VALUE

Issue #5969, opened 2026-02-25
[FEATURE] Make `opensearch_security.multitenancy.tenants.preferred` dynamic via Security config (similar to 2.7 multitenancy)


cwperks:
## Summary

Today, the `opensearch_security.multitenancy.tenants.preferred` setting is configured statically in `opensearch_dashboards.yml`. In order to change the value of this setting requires OSD node reboots unlike other dynamic security dashboards settings.

In OpenSearch 2.7, we made a change to make **multi-tenancy** dynamic, allowing it to be managed via the Security config and updated without requiring a Dashboards restart:

* Backend: [https://github.com/opensearch-project/security/pull/2607](https://github.com/opensearch-project/security/pull/2607)
* Frontend: [https://github.com/opensearch-project/security-dashboards-plugin/pull/1394](https://github.com/opensearch-project/security-dashboards-plugin/pull/1394)

This issue proposes making `opensearch_security.multitenancy.tenants.preferred` configurable dynamically via the Security configuration API in OpenSearch, instead of being solely controlled via `opensearch_dashboards.yml`.

---

## Problem Statement

Currently:

```yaml
opensearch_security.multitenancy.tenants.preferred: ["Private", "Global"]
```

* This setting is defined statically in `opensearch_dashboards.yml`
* Changes require a Dashboards restart

---

#### 1. Extend Security Config

* Extend `securityconfig` to include a `preferred_tenants` setting under kibana.

Example (illustrative):

```json
{
  "dynamic": {
    "kibana": {
      ...
      "multitenancy_enabled": true,
      "preferred_tenants": ["Private", "Global"]
    }
  }
}
```

#### 2. Expose via `/_plugins/_security/dashboardsinfo`

The value should be returned by the `/_plugins/_security/dashboardsinfo` endpoint.

This keeps Dashboards configuration fetch centralized and consistent.

Example (illustrative):

```json
{
  ...
  "multitenancy_enabled": true,
  "preferred_tenants": ["Private", "Global"]
}
```

### Frontend (Security Dashboards Plugin)

* Similar pattern to:

  * [https://github.com/opensearch-project/security-dashboards-plugin/pull/1394](https://github.com/opensearch-project/security-dashboards-plugin/pull/1394)
* Fetch preferred tenants from the Security config instead of only reading from `opensearch_dashboards.yml`.
* Maintain backward compatibility:

  * If dynamic config is absent → fallback to `opensearch_dashboards.yml`
[...]

PR #5986, opened 2026-03-03
Make `opensearch_security.multitenancy.tenants.preferred` configurable dynamically via security config api

itsmevichu:
### Description
* Category (Enhancement)
This PR proposes making `opensearch_security.multitenancy.tenants.preferred` configurable dynamically via the Security configuration API in OpenSearch, instead of being solely controlled via opensearch_dashboards.yml
  **Before:** `preferred_tenants` was not part of config APIs or `/dashboardsinfo`.
  **After:** `preferred_tenants` can be set via tenancy config API, is persisted in `ConfigV7.Kibana`, returned in `/dashboardsinfo`.

### Issues Resolved
https://github.com/opensearch-project/security/issues/5969

### Testing
- Added/updated unit tests in `MultiTenancyConfigApiTest` and `ConfigV7Test` for `preferred_tenants`. :

### Check List


src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
```diff
@@ -379,7 +380,15 @@ private ValidationResult<JsonNode> validateDataType(final JsonNode jsonContent)
     }
 
     private ValidationResult<JsonNode> nullValuesInArrayValidator(final JsonNode jsonContent) {
-        for (final Map.Entry<String, DataType> allowedKey : validationContext.allowedKeys().entrySet()) {
```
cwperks, line 382: nit: I think key name is all that matters here so idk if we need this change. That being said, is it enforced that `allowedKeyWithConfig` is strictly a subset of `allowedKeys`?

cwperks (dismissed):

src/main/java/org/opensearch/security/dlic/rest/validation/RequestContentValidator.java
itsmevichu: I understand, we don’t need the DataType; just the key name is sufficient, and I’ll update the changes.

And `allowedKeysWithConfig` isn’t required to be a strict subset of `allowedKeys`. If `allowedKeysWithConfig` is present, it takes precedence; otherwise, we fallback to `allowedKeys`. Going forward, for any new API we define, we can start using `allowedKeysWithConfig` directly, which will make it easier when we eventually remove allowedKeys completely.

itsmevichu:
>Only question I have is if preferred_tenants will be populated in the index when blank or if its only persisted when its set?

@cwperks Its persisted only if its set.

cwperks (dismissed):

niravpi (approved):

cwperks (approved):

RyanL1997 (approved):

PR #5988, opened 2026-03-05
Optimized string matching for RoleBasedActionPrivileges

### LINKED ISSUE #5998: Optimize `RoleBasedActionPrivileges` construction by sorting concrete indices before pattern evaluation
## Problem

`RoleBasedActionPrivileges` builds precomputed privilege structures for index authorization. In clusters with many concrete indices/aliases and many distinct index patterns across role index-permissions, construction can spend a lot of CPU repeatedly matching patterns against the full concrete name set. ([[GitHub](https://raw.githubusercontent.com/opensearch-project/security/main/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java)][1])

## Proposed improvement

Sort the concrete indices / aliases before evaluating index patterns during construction.

With a sorted concrete name set, many wildcard patterns—especially prefix-like patterns such as `logs-*`, `metrics-api-*`, etc.—can be evaluated more efficiently because matching can:

* skip irrelevant names sooner, and
* terminate once names move beyond the possible match range.

That should reduce repeated comparisons and improve CPU efficiency during privilege precomputation.

## Why this helps

For a large set of concrete names, the current effective behavior can resemble:

* for each unique pattern
* scan many or all concrete indices / aliases

Sorting first creates an opportunity for eager stop behavior and more locality in matching.

## Notes

This seems especially valuable for large clusters with:

* many unique patterns across roles,
* many aliases,
* and frequent privilege rebuilds.

`RoleBasedActionPrivileges` already notes that stateful index privilege updates can take around 1000 ms on large clusters, so this looks like a good place for a targeted optimization. ([[GitHub](https://raw.githubusercontent.com/opensearch-project/security/main/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java)][1])

[1]: https://raw.githubusercontent.com/opensearch-project/security/main/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java "raw.githubusercontent.com"

nibix:
### Description

Fixes #5998

This PR provides two optimizations:

- All compilation of index patterns, static DLS queries, etc. is decoupled from building `RoleBasedActionPrivileges` and `AbstractRuleBasedPrivileges` into a new class `CompiledRoles`. This will guarantee that each index pattern and static DLS query is only compiled once even if it is later used in different components (e.g. stateful vs non-stateful classes or index privileges vs DLS) and for different indices.
- The `IndexPattern` class is extended to provide optimized matching for prefix type index patterns (like `index_*`) and for exact index patterns (like `index_a`). For prefix type index patterns, we leverage the fact that the cluster state contains a `TreeSet` of all indices, aliases and data streams. Performing a prefix pattern match is then only a *O(log n)* operation, while it was an *O(n)* operation before.

These optimizations reduces the runtime of  stateful index privileges  in one micro benchmark test case by a factor of about 0.46.

Before:

```
[constructionPerformance_sharedPatterns] Configuration: 3000 roles, 6000 indices, 9000 aliases
[constructionPerformance_sharedPatterns] RoleBasedActionPrivileges construction: 148ms
[constructionPerformance_sharedPatterns] StatefulIndexPrivileges update: 2169ms
[constructionPerformance_sharedPatterns] Total: 2317ms
```

After:

```
[constructionPerformance_sharedPatterns] Configuration: 3000 roles, 6000 indices, 9000 aliases
[constructionPerformance_sharedPatterns] RoleBasedActionPrivileges construction: 42ms
[constructionPerformance_sharedPatterns] StatefulIndexPrivileges update: 971ms
[constructionPerformance_sharedPatterns] Total: 1013ms
```

* Category
  - Enhancement
  - In clusters with many roles and indices, the update of role configuration could have a latency of several seconds. This will now the reduce the latency by about 50%.
  - No behavioral changes.
  
### Testing
- Existing integration tests

### Check List

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       # Drafts the next Release notes as Pull Requests are merged into "main"
-      - uses: release-drafter/release-drafter@v6
+      - uses: release-drafter/release-drafter@v7
         with:
           config-name: release-notes-drafter-config.yml
         env:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,25 +15,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Hardens input validation for resource sharing APIs ([#5831](https://github.com/opensearch-project/security/pull/5831)
 - Optimize getFieldFilter to only return a predicate when index has FLS restrictions for user ([#5777](https://github.com/opensearch-project/security/pull/5777))
 - Performance optimizations for building internal authorization data structures upon config updates ([#5988](https://github.com/opensearch-project/security/pull/5988))
+- Make encryption_key optional for obo token authenticator ([#6017](https://github.com/opensearch-project/security/pull/6017)
+- Enable basic authentication for gRPC transport ([#6005](https://github.com/opensearch-project/security/pull/6005))
 
 ### Bug Fixes
 - Fix audit log writing errors for rollover-enabled alias indices ([#5878](https://github.com/opensearch-project/security/pull/5878)
-
 - Fix the issue of unprocessed X-Request-Id ([#5954](https://github.com/opensearch-project/security/pull/5954))
 - Improve DLS error message to identify undefined user attributes when query substitution fails ([#5975](https://github.com/opensearch-project/security/pull/5975))
+
 ### Refactoring
 
 ### Maintenance
 - Bump `commons-codec:commons-codec` from 1.20.0 to 1.21.0 ([#5937](https://github.com/opensearch-project/security/pull/5937))
 - Bump `at.yawk.lz4:lz4-java` from 1.10.2 to 1.10.4 ([#5938](https://github.com/opensearch-project/security/pull/5938), [#5994](https://github.com/opensearch-project/security/pull/5994))
 - Bump `open_saml_shib_version` from 9.1.6 to 9.2.1 ([#5936](https://github.com/opensearch-project/security/pull/5936), [#5982](https://github.com/opensearch-project/security/pull/5982))
-- Bump `com.google.googlejavaformat:google-java-format` from 1.33.0 to 1.34.1 ([#5947](https://github.com/opensearch-project/security/pull/5947))
+- Bump `com.google.googlejavaformat:google-java-format` from 1.33.0 to 1.35.0 ([#5947](https://github.com/opensearch-project/security/pull/5947), [#6011](https://github.com/opensearch-project/security/pull/6011))
 - Bump `aws-actions/configure-aws-credentials` from 5 to 6 ([#5946](https://github.com/opensearch-project/security/pull/5946))
 - Bump `ch.qos.logback:logback-classic` from 1.5.26 to 1.5.32 ([#5948](https://github.com/opensearch-project/security/pull/5948), [#5995](https://github.com/opensearch-project/security/pull/5995))
-- Bump `com.github.seancfoley:ipaddress` from 5.5.1 to 5.6.1 ([#5949](https://github.com/opensearch-project/security/pull/5949))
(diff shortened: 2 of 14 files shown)


nibix:
There seem to be test issues, taking back to draft

src/main/java/org/opensearch/security/privileges/actionlevel/WellKnownActions.java
```diff
@@ -74,10 +70,6 @@ public class WellKnownActions {
         UpdateByQueryAction.NAME,
         DeleteByQueryAction.NAME,
         TransportShardBulkAction.ACTION_NAME,
-        PutMappingAction.NAME,
-        AutoPutMappingAction.NAME,
-        AnalyzeAction.NAME,
-        AutoCreateAction.NAME,
```
nibix, line 80: As an additional optimization, I have removed a few action names from the list of actions for which the denormalized data structures are created.

Reasoning:

- AutoCreate, AutoPut are likely never used because they work on non-existing indices. However the denormalized data structures by principle can only be created for existing indices
- Analyze, PutMapping: This is maybe too seldomly used to justify the overhead

src/integrationTest/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivilegesTest.java
```diff
@@ -74,14 +78,26 @@
     RoleBasedActionPrivilegesTest.Misc.class,
     RoleBasedActionPrivilegesTest.StatefulIndexPrivilegesHeapSize.class })
 public class RoleBasedActionPrivilegesTest {
+
+    static RoleBasedActionPrivileges createSubject(
```
cwperks, line 82: Why `createSubject`?

src/integrationTest/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivilegesTest.java
```diff
@@ -74,14 +78,26 @@
...
+        return new RoleBasedActionPrivileges(
+            new CompiledRoles(roles, actionGroups, NamedXContentRegistry.EMPTY, FieldMasking.Config.DEFAULT),
+            settings
+        );
+    }
+
     public static class ClusterPrivileges {
         @Test
         public void wellKnown() throws Exception {
             SecurityDynamicConfiguration<RoleV7> roles = SecurityDynamicConfiguration.fromYaml("test_role:\n" + //
```
cwperks, line 96: nit: I know its not changed in this PR, but this would be a good contender for changing to Text Block.

src/main/java/org/opensearch/security/privileges/CompiledRoles.java
```diff
@@ -0,0 +1,263 @@
...
+         * The compiled index permission entries for this role.
+         */
+        public final ImmutableList<Index> indexPermissions;
+
+        public Role(
+            String roleName,
+            RoleV7 base,
+            FlattenedActionGroups actionGroups,
+            NamedXContentRegistry xContentRegistry,
+            FieldMasking.Config fieldMaskingConfig
```
cwperks, line 111: Why only the field masking config here?

src/integrationTest/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivilegesTest.java
nibix: subject as in "test subject". `RoleBasedActionPrivileges` is the subject of this test.

src/integrationTest/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivilegesTest.java
nibix: yeah. maybe we should do one bulk PR where we convert all these contenders in one go? I can give this a shot later this week.

src/main/java/org/opensearch/security/privileges/CompiledRoles.java
nibix: We need to know the default hashing algorithm when creating the field masking rule objects.

Issue #2673, opened 2023-04-12
Audit log setting parameters not working


pratikshavyas:
**What is the bug?**
I am using Opensearch v2.2.1 and have enabled audit logging. I observed that few setting parameters are not taking any effect and not working. Below is the configuration I have added in opensearch.yml file after following documentation: [Audit logs - OpenSearch documentation](https://opensearch.org/docs/2.2/security/audit-logs/index/)

**How can one reproduce the bug?**
Add below configuration in the opensearch.yml file and install opensearch 
plugins.security.audit.type: log4j
plugins.security.audit.config.log4j.logger_name: audit
plugins.security.audit.config.log4j.level: INFO
plugins.security.audit.config.disabled_rest_categories: NONE
plugins.security.audit.config.disabled_transport_categories: NONE
plugins.security.audit.ignore_users: NONE

**What is the expected behavior?**
All audit event categories should get logged and no user should get ignored and should emit logs for even kibanaserver and authenticated and granted_privileges category.

**What is your host/environment?**
 - Opensearch v2.2.1

**Do you have any additional context?**
For plugins.security.audit.config.disabled_rest_categories and plugins.security.audit.config.disabled_transport_categories it is not throwing any error, however it is not having any impact.
For plugins.security.audit.ignore_users setting it throws below error:
org.opensearch.bootstrap.StartupException: java.lang.IllegalArgumentException: unknown setting [plugins.security.audit.ignore_users] did you mean [plugins.security.audit.config.username]?

This same error about unknown setting is coming for many other settings mentioned in the documentation: https://opensearch.org/docs/2.2/security/audit-logs/index/

jmnstewart:
We have been experiencing this issue since we moved from elasticsearch to opensearch.

**Versions (relevant - OpenSearch):**

v2.7.0

Describe the issue:

When applying audit log settings they do not take effect. See screenshot of the configuration not being applied.   Logs of install attached, it seems to go into a loop regarding the audit config.

Configuration:

plugins.security.audit.config.index: "security-auditlog-'yyyy.mm.dd'"
plugins.security.audit.type: internal_opensearch
plugins.security.allow_unsafe_democertificates: false
plugins.security.allow_default_init_securityindex: false
plugins.security.audit.config.enable_rest: true
plugins.security.audit.config.enable_transport: true
plugins.security.audit.config.log_request_body: true
plugins.security.audit.config.resolve_indices: true
plugins.security.audit.config.disabled_rest_categories: NONE
[...]

PR #6021, opened 2026-03-18
fixed plugins unknown setting and disabled_rest_categories

ThyTran1402:
### Description
- Fixed: Audit log NONE sentinel value not respected in dynamic configuration, and misleading "unknown setting" error for ignore_users
- New behavior: 
AuditCategory.parse() now treats a single "NONE" value (case-insensitive) as an empty set, consistent with the existing behavior in the static opensearch.yml path. This fixes disabled_rest_categories and disabled_transport_categories for both config paths.
Filter.from(Map) (the dynamic/REST path) now applies the same "NONE" → empty set conversion for ignore_users that fromSettingStringSet() already applied for the static path.
The correct key for opensearch.yml is plugins.security.audit.config.ignore_users (note the .config. segment); using the wrong key plugins.security.audit.ignore_users is a documentation/usage issue — no code change required

### Issues Resolved
- #2673 

### Testing
- Added testNoneViaMap() unit test to AuditConfigFilterTest covering the previously untested Filter.from(Map) code path, verifying that "NONE", "None", and "none" all correctly clear disabled categories and ignored users when set via the REST API.
- Existing testNone() test continues to cover the static opensearch.yml path.

### Check List


cwperks (approved):

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
```diff
@@ -256,9 +256,10 @@ public static Filter from(Map<String, Object> properties) throws JsonProcessingE
                     ConfigConstants.OPENDISTRO_SECURITY_AUDIT_DISABLED_CATEGORIES_DEFAULT
                 )
             );
-            final Set<String> ignoredAuditUsers = ImmutableSet.copyOf(
-                getOrDefault(properties, FilterEntries.IGNORE_USERS.getKey(), DEFAULT_IGNORED_USERS)
-            );
+            final List<String> rawIgnoredUsers = getOrDefault(properties, FilterEntries.IGNORE_USERS.getKey(), DEFAULT_IGNORED_USERS);
+            final Set<String> ignoredAuditUsers = rawIgnoredUsers.size() == 1 && "NONE".equalsIgnoreCase(rawIgnoredUsers.get(0))
```
DarshitChanpura, line 260: why not filter out NONE from rawIngoredUsers?

ThyTran1402: hmmm I think NONE is the sentinel config value, not actual username. It means don't ignore any users. So, rawIngoredUsers holds exactly what's in the config, keeping the input separate from the output logic. If we filtered it, we could lose the information that the user explicitly configured rather than just providing an empty list.

ThyTran1402: So, NONE is like special handling not regular filter.

PR #6037, opened 2026-03-26
Only update PrivilegesEvaluator configuration when the base config objects have actually changed.

### LINKED ISSUE #5999: Avoid reconstructing `RoleBasedActionPrivileges` when security configs other than roles and actiongroups change
## Is your feature request related to a problem?

`RoleBasedActionPrivileges` is a precomputed, optimized privilege structure derived from role and action group configuration. The class documentation explicitly says its lifecycle corresponds to the lifecycle of the role and action group configuration, and that a new instance needs to be built when role or action group configuration changes. It also notes that stateful index privilege updates on large clusters can take on the order of ~1000 ms. ([[raw.githubusercontent.com](https://raw.githubusercontent.com/opensearch-project/security/main/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java)][1])

However, it appears that `RoleBasedActionPrivileges` may currently be reconstructed for broader security configuration changes than necessary, including updates to config types that do not change the role/action-group-derived privilege model itself.

Examples of config changes that should not require rebuilding `RoleBasedActionPrivileges` include:

* `internalusers`
* `rolesmapping`
* other unrelated security config documents

From a dependency perspective, this structure should only need to be rebuilt when the inputs it is derived from change:

* `roles`
* `actiongroups`

The Security plugin APIs and configuration model treat users, roles, role mappings, and action groups as separate config domains, which suggests these changes can be invalidated more selectively. ([[OpenSearch Docs](https://docs.opensearch.org/latest/security/access-control/api/?utm_source=chatgpt.com)][2])

## What solution would you like?

Tighten the invalidation / rebuild logic so that `RoleBasedActionPrivileges` is reconstructed **only** when:

* role definitions change, or
* action group definitions change.

Changes to unrelated security config types should not trigger reconstruction.

For example:

* changing an internal user password should **not** rebuild `RoleBasedActionPrivileges`
* changing backend role mappings for a user should **not** rebuild `RoleBasedActionPrivileges`
* changing unrelated security plugin settings/documents should **not** rebuild `RoleBasedActionPrivileges`

Those changes may require refreshing other security state, but not this particular precomputed privilege structure.

## Why this helps

Rebuilding `RoleBasedActionPrivileges` is more expensive than necessary if it is triggered by config changes that do not affect its actual inputs. Since the class is specifically derived from `roles` and `actionGroups`, broader rebuild triggers create avoidable CPU work, object churn, and latency spikes during security config updates. ([[GitHub](https://raw.githubusercontent.com/opensearch-project/security/main/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java)][1])

This is especially relevant in clusters where:

* security configs are updated frequently,
* user and role mapping data changes more often than role definitions,
* and role/action-group structures are large enough that reconstruction is noticeable.

## Additional context

`RoleBasedActionPrivileges` is already designed as a long-lived precomputed structure whose lifecycle matches role/action-group configuration, so narrowing reconstruction triggers would be consistent with the class’ stated design intent. ([[GitHub](https://raw.githubusercontent.com/opensearch-project/security/main/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java)][1])

[1]: https://raw.githubusercontent.com/opensearch-project/security/main/src/main/java/org/opensearch/security/privileges/actionlevel/RoleBasedActionPrivileges.java "raw.githubusercontent.com"
[2]: https://docs.opensearch.org/latest/security/access-control/api/?utm_source=chatgpt.com "API - OpenSearch Documentation"

nibix:
### Description

This changes PrivilegesConfiguration config change listening logic to only update the PrivilegesEvaluator and DLS/FLS configuration if the underlying configuration objects actually have been changed since the last time. 

This can provide a small performance benefit, as for clusters with many roles and indices, the config updates can be costly (a few seconds for very large environments).

* Category: Performance Enchancements

### Issues Resolved

Fixes #5999

### Testing

- Existing int tests

### Check List

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,9 +4,12 @@ on:
   push:
     branches:
       - main
-      - 1.*
       - 2.*
   pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 env:
   GRADLE_OPTS: -Dhttp.keepAlive=false
--- a/.github/workflows/maven-publish.yml
+++ b/.github/workflows/maven-publish.yml
@@ -23,7 +23,7 @@ jobs:
           java-version: 21
       - uses: actions/checkout@v6
       - name: Load secret
-        uses: 1password/load-secrets-action@v3
+        uses: 1password/load-secrets-action@v4
         with:
           # Export loaded secrets as environment variables
           export-env: true
--- a/.github/workflows/plugin_install.yml
+++ b/.github/workflows/plugin_install.yml
@@ -2,6 +2,10 @@ name: Plugin Install
 
 on: [push, pull_request, workflow_dispatch]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
 env:
   OPENSEARCH_VERSION: 3.6.0
   PLUGIN_NAME: opensearch-security
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Make security plugin aware of FIPS build param (-Pcrypto.standard=FIPS-140-3) ([#5952](https://github.com/opensearch-project/security/pull/5952))
 - Hardens input validation for resource sharing APIs ([#5831](https://github.com/opensearch-project/security/pull/5831)
 - Optimize getFieldFilter to only return a predicate when index has FLS restrictions for user ([#5777](https://github.com/opensearch-project/security/pull/5777))
+- Make `plugins.security.dfm_empty_overrides_all` dynamically toggleable ([#6016](https://github.com/opensearch-project/security/pull/6016)
 - Performance optimizations for building internal authorization data structures upon config updates ([#5988](https://github.com/opensearch-project/security/pull/5988))
 - Make encryption_key optional for obo token authenticator ([#6017](https://github.com/opensearch-project/security/pull/6017)
 - [Resource Sharing] Using custom action prefixes for sample resource plugin ([#6020](https://github.com/opensearch-project/security/pull/6020)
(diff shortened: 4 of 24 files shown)


src/main/java/org/opensearch/security/privileges/PrivilegesConfiguration.java
```diff
@@ -227,23 +248,6 @@ public PrivilegesEvaluator privilegesEvaluator() {
...
-
-    /**
-     * Returns the current compiled roles. These are built once per configuration update and shared between the
-     * action privilege layer and the DLS/FLS layer to avoid redundant pattern compilation.
-     * Important: Do not store the references to the instances returned here; these will change after configuration updates.
-     */
-    public CompiledRoles compiledRoles() {
-        return this.compiledRoles.get();
-    }
-
```
nibix, line 246: Code was already unused before, removed to facilitate changes.

src/main/java/org/opensearch/security/privileges/PrivilegesConfiguration.java
```diff
@@ -139,58 +137,81 @@ public PrivilegesConfiguration(
...
-                            settings,
-                            privilegesInterceptor,
-                            indexResolverReplacer,
-                            flattenedActionGroups,
-                            staticActionGroups,
-                            newCompiledRoles,
-                            generalConfiguration,
-                            pluginIdToRolePrivileges
-                        )
+                boolean privilegesChanged = oldRawConfiguration == null
```
cwperks, line 140: Should we switch the ordering of the last 2 checkis here? I imagine `oldRawConfiguration.equals(rawConfiguration)` is the most expensive to compute.

src/main/java/org/opensearch/security/privileges/PrivilegesConfiguration.java
```diff
@@ -139,58 +137,81 @@ public PrivilegesConfiguration(
...
+                    || !oldRawConfiguration.equals(rawConfiguration)
+                    || currentType != targetType;
+                if (privilegesChanged) {
+                    log.debug("Privileges for PrivilegesEvaluator and DLS/FLS did not change; updating.");
```
cwperks: Is this debug statement accurate? The condition above is checking `if (privilegedChanged)` whereas this log line is saying did not change.

cwperks (dismissed):

src/main/java/org/opensearch/security/privileges/PrivilegesConfiguration.java
nibix: yeah, this is wrong

src/main/java/org/opensearch/security/privileges/PrivilegesConfiguration.java
nibix: I am not sure if it relevant to consider the computation time here. comparing will certainly not take longer than constructing the data structure. and looking into SecurityDynamicConfiguration, we make very liberal use of cloning. So, I would guess that is max a few ms.

src/main/java/org/opensearch/security/privileges/PrivilegesConfiguration.java
nibix: fixed

PR #6044, opened 2026-03-27
Cache hasFlsOrFieldMasking in PrivilegesEvaluationContext

nibix:
### Description

Replaces #6001

This adds a caching for the result of `DlsFlsValveImpl.hasFlsOrFieldMasking()` to `PrivilegesEvaluationContext`. The method can be called multiple times if the query is complex and there a serveral shards of the same index on the same node. This reduced the amount of calls per request and node to a single one.

* Category: Performance enhancements
* What is the old behavior before changes and new behavior after changes? No behavioral changes

### Testing

- existing int tests

### Check List


nibix:
@finnegancarroll 

> For security plugin, we need to evaluate if this index is subject to any fls/masking for each clause since the lucene query cache is not aware of these policies and is liable to erroneously return a doc matching the clause, when actually the matching field is masked or unavailable to the user?

Yes, caching (both query and search caching) and fine grained access controls are not the best friends. If this is active, we have to make sure that data is not served from the cache which might have been populated by other users.

PR #6059, opened 2026-04-01
Facilitate FIPS-compliant store resolution

beanuwave:
### Description

Adds BCFKS keystore files alongside all existing JKS/PKCS12 test keystores in `src/test/resources/` (excluding saml/ and on-behalf-token/), and updates test infrastructure to resolve the correct format at runtime.

`RestHelper.java` and `AbstractSecurityUnitTest.java` use `resolveStorePath` for all keystore/truststore loading. `inferStoreType(resolvedPath)` derives the KeyStore type from the resolved path extension, keeping type and format always consistent.

`inferStoreType(resolvedPath)`  is a clone from core-projects test-framework https://github.com/opensearch-project/OpenSearch/blob/5b95e609b5ee1bcf5a5452a152a861bbe8971f31/test/framework/src/main/java/org/opensearch/test/KeyStoreUtils.java#L37-L48

Tests that explicitly exercise JDK store formats as part of their logic (`SslSettingsManagerReloadListenerTest.java`, `JdkSslCertificatesLoaderTest.java`) are out of scope and will be handled separately in https://github.com/opensearch-project/security/pull/5866

Other changes:
- _DEFAULT_STORE_TYPE_ resolves to _BCFKS_ in FIPS-mode, replacing hardcoded _JKS_ and `KeyStore.getDefaultType() `across keystore creation
  sites
- New keystores created during alias extraction (`KeyStoreUtils`) now inherit the source store's type, preventing _BCFKS_ key material from being re-encrypted into _PKCS12_
- `PemKeyReader.detectStoreType` sniffs file bytes to distinguish _JKS/PKCS12/BCFKS_
- Fixed `parseOtherName` in `Certificate` and `DefaultSecurityKeyStore` to handle both JDK X.509 & BC X.509 encoding classes.
- Extract SAN parsing logic from `Certificate.subjectAlternativeNames()` and
  `DefaultSecurityKeyStore.getSubjectAlternativeNames()` into a shared `SanParser` utility class.

### Check List


src/main/java/org/opensearch/security/support/PemKeyReader.java
```diff
@@ -347,5 +354,56 @@ public static char[] randomChars(int len) {
         return ret;
     }
 
+    public static String extractStoreType(String storePath, String storeType) {
+        if (null == storeType) {
+            storeType = detectStoreType(storePath);
+        }
+        if (CryptoServicesRegistrar.isInApprovedOnlyMode() && !PemKeyReader.BCFKS.equalsIgnoreCase(storeType)) {
+            throw new IllegalArgumentException(
+                storeType.toUpperCase(Locale.ROOT) + " truststores are not supported in FIPS mode - use BCFKS."
```
reta: I believe it is not just about truststore but keystore as well:
 
```suggestion
                storeType.toUpperCase(Locale.ROOT) + " keystores / truststores are not supported in FIPS mode - use BCFKS."
```

src/test/java/org/opensearch/security/test/helper/file/FileHelper.java
```diff
@@ -61,39 +63,77 @@ public class FileHelper {
...
+     *       {@code .bcfks} (Bouncy Castle FIPS keystore)</li>
+     *   <li>Non-FIPS → {@code .jks} if a JKS variant exists on the classpath,
+     *       otherwise {@code .p12}</li>
+     * </ul>
+     *
+     * @param baseName classpath-relative base name without extension, e.g. {@code "ssl/truststore"}
+     * @return absolute path to the resolved keystore file
+     * @throws IllegalStateException if no matching file is found on the classpath
+     */
+    public static Path resolveStorePath(final String baseName) {
```
reta: Minor suggestion, it seems like in majority of usage scenarios we do resolveStorePath + resolveStorePath , may be we could make it a bit simpler with (pseudo code):

```java
record TypedStore(Path path, String type) {};

TypedStore resolveStore(final String baseName) {
  ...
}
```

src/test/java/org/opensearch/security/test/DynamicSecurityConfig.java
```diff
@@ -152,7 +152,7 @@ public List<IndexRequest> getDynamicConfig(String folder) {
             );
         }
 
-        if (null != FileHelper.getAbsoluteFilePathFromClassPath(prefix + securityNodesDn)) {
+        if (DynamicSecurityConfig.class.getClassLoader().getResource(prefix + securityNodesDn) != null) {
```
reta: Hm ... why not delegate to `FileHelper`?

reta:
Thanks @beanuwave , the change makes sense to me, a few minor comments

src/test/java/org/opensearch/security/test/DynamicSecurityConfig.java
beanuwave: Sure, let's do that!

Background: This change updates `FileHelper.getAbsoluteFilePathFromClassPath(...)` to throw a runtime exception when the file path cannot be resolved, instead of returning `null`. This cleanly separates resource availability checks in tests from `String` → `Path` conversion.

src/test/java/org/opensearch/security/test/helper/file/FileHelper.java
beanuwave: Could not agree more!

reta (approved):

src/main/java/org/opensearch/security/ssl/config/SanParser.java
```diff
@@ -0,0 +1,78 @@
...
+import org.bouncycastle.cert.jcajce.JcaX509CertificateHolder;
+import org.bouncycastle.crypto.CryptoServicesRegistrar;
+
+public class SanParser {
+
+    private static final Logger LOGGER = LogManager.getLogger(SanParser.class);
+
+    private SanParser() {}
+
+    public static String parse(X509Certificate certificate) {
```
cwperks, line 42: Nice to see this abstracted to another class where it can be unit tested.

cwperks (commented):
@beanuwave the changes LGTM, but can we add a README somewhere with instructions on how some of the trust stores were created? I believe we have a README outlining how some of the various test certs were generated.

cwperks (approved):
@beanuwave The change looks good to me and thanks for the thoroughness.

A couple comments and lmk what you think:

1. I'd love to have it documented somewhere in the repo how the key stores are generated through a README or other markdown file. Since we're checking expiring certs into the repo, someone in the future will need to replace them and know how they were generated.
2. I think we should use this opportunity to add a new CI check in the security repo to run the security plugin's `integrationTest` against a cluster running in FIPS enforced mode since the security repo has a lot of coverage on different actions used on a cluster.

Issue #6051, opened 2026-03-30
[FEATURE] Introduce `plugins.security.superadmin.secret` for superadmin access when https is disabled


cwperks:
### What are you proposing?

Introduce a new statically-defined superadmin secret for OpenSearch Security that can be used when HTTPS is disabled, and potentially when HTTPS is enabled as well.

Today, superadmin access is tied to `plugins.security.authcz.admin_dn`, which depends on client certificate authentication over HTTPS. In deployments where HTTPS is disabled, there is no equivalent mechanism to access the security index with superadmin privileges. This creates a recovery and operability gap, especially if the security index is deleted, corrupted, or otherwise unavailable.

The proposal is to add a new setting such as:

```yaml
plugins.security.superadmin.secret: ${SUPERADMIN_SECRET}
```

This secret would be defined statically in `opensearch.yml` and intentionally remain outside the security index, similar in spirit to `admin_dn`. A caller that presents the correct secret through a dedicated HTTP header would be treated as a superadmin for privileged security operations, including access to the security index.

This mechanism could also be supported by `securityadmin.sh` so that seeding or reapplying security configuration from YAML files works in HTTP-only environments.

### What users have asked for this feature?

A recurring problem is that users running with HTTP only do not have a true equivalent to certificate-based superadmin access.

There is also a related operational issue where users want an internal user with `superuser`-like permissions to act as a fallback administrator. That does not fully solve the problem, because if the security index is deleted and caches are refreshed, authentication can fail entirely since internal users are themselves defined in the security index.

In that situation, there is no independent recovery path unless the deployment uses HTTPS with configured admin certificates.

### What problems are you trying to solve?

When a deployment runs without HTTPS, an administrator wants a built-in, out-of-band recovery mechanism so they can still access and repair the security index.

When the security index is deleted, unavailable, or misconfigured, an operator wants a way to authenticate independently of that index so they can restore configuration and recover the cluster.

When using `securityadmin.sh` in HTTP-only environments, an operator wants a supported way to seed or reapply security configuration without requiring certificate-based superadmin authentication.

More broadly, this provides HTTP users with an operational escape hatch equivalent to what `admin_dn` provides for HTTPS users today.

### What is the developer experience going to be?

This would introduce a new static setting outside the security index, for example:

```yaml
plugins.security.superadmin.secret: ${SUPERADMIN_SECRET}
```

Proposed behavior:

[...]

PR #6073, opened 2026-04-06
Introduces `plugins.security.superadmin.secret` for superadmin access when https is disabled

itsmevichu:
### Description
* New feature
    - These changes add support for a superadmin secret authentication path, allowing superadmin access without relying on admin DN or the security index. This creates a recovery and operability gap, especially if the security index is deleted, corrupted, or otherwise unavailable.
#### Old Behavior
- No dedicated superadmin secret existed.
- Superadmin access was only possible via:
    - admin DN client certificate authentication, or
    - existing security index-based authentication flows.
- In HTTP-only mode, there was no equivalent recovery or superadmin bypass path, because DN-based superadmin only works for HTTPS/client-cert auth.

#### New Behaviour
- Introduces a new setting `plugins.security.superadmin.secret` for plaintext secret configuration. 
- Adds a corresponding secure variant- `plugins.security.superadmin.secret_secure`  for keystore-based storage, which is the recommended approach:
     - ./bin/opensearch-keystore add plugins.security.superadmin.secret_secure
- At runtime, the secret is validated from the incoming request header, when the secret is valid, the request is authenticated as the superadmin user and receives full superadmin capabilities.
     - `X-OpenSearch-Superadmin-Secret`
- superadmin.sh cnow supports the secret via configurable options: 
     - `-sas or --superadmin-secret`

### Issues Resolved
https://github.com/opensearch-project/security/issues/6051

## To do:
- Unit tests
- Docs

### Testing
- Manual testing

### Check List


src/main/java/org/opensearch/security/support/SecuritySettings.java
```diff
@@ -57,4 +63,40 @@ public class SecuritySettings {
...
+        ConfigConstants.SECURITY_SUPERADMIN_SECRET_SECURE,
+        SECURITY_SUPERADMIN_SECRET_INSECURE_SETTING
+    );
+
+    /**
+     * Alternative to InsecureStringSetting, which doesn't raise an exception if allow_insecure_settings is false, but
+     * instead logs a warning. This is to allow insecure settings for container-based deployments while recommending
+     * secure keystore usage.
+     */
+    private static class InsecureFallbackStringSetting extends Setting<SecureString> {
```
nibix: Could you elaborate on the usecase here? If  allow_insecure_setting is configured to be false, then IMHO insecure settings should not be allowed.

src/main/java/org/opensearch/security/support/SecuritySettings.java
```diff
@@ -57,4 +63,40 @@ public class SecuritySettings {
...
     );
+
+    public static final Setting<SecureString> SECURITY_SUPERADMIN_SECRET_INSECURE_SETTING = new InsecureFallbackStringSetting(
+        ConfigConstants.SECURITY_SUPERADMIN_SECRET
+    );
+
+    public static final Setting<SecureString> SECURITY_SUPERADMIN_SECRET_SETTING = SecureSetting.secureString(
+        ConfigConstants.SECURITY_SUPERADMIN_SECRET_SECURE,
+        SECURITY_SUPERADMIN_SECRET_INSECURE_SETTING
+    );
```
nibix, line 76: IMHO, this should be accompanied with a very prominent note that using such a setting is inherently insecure and should be only if there is no other way. It should be noted that using TLS client cert auth should be always preferred.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -278,6 +283,24 @@ public boolean authenticate(final SecurityRequestChannel request) {
             return true;
         }
 
+        /*
+        Authenticates superuser based on superadmin secret. The secret is read from thread context
```
finnegancarroll: "secret is read from thread context" -> "from header"

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -271,7 +271,7 @@ public RestHandler wrap(RestHandler original, AdminDNs adminDNs, Set<RestHeaderD
      * Checks if a given user is a SuperAdmin
      */
     boolean userIsSuperAdmin(User user, AdminDNs adminDNs) {
-        return user != null && adminDNs.isAdmin(user);
+        return user != null && (adminDNs.isAdmin(user) || ConfigConstants.SECURITY_SUPERADMIN_SECRET_USER.equals(user.getName()));
```
finnegancarroll: What happens if I configure a user on my cluster with the same name as this hard coded `SECURITY_SUPERADMIN_SECRET_USER`? Is there a path to escalate privileges here?

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -278,6 +283,24 @@ public boolean authenticate(final SecurityRequestChannel request) {
...
+        and compared against the configured superadmin secret. If superuser is authenticated here we skip the remaining
+        authentication flow. This mechanism is independent of the security index and serves as an out-of-band recovery
+        path for HTTP deployments.
+         */
+        if (!gRPC) {
```
finnegancarroll: Just as a note I can't think of any reason this would not work the same over gRPC but makes sense to keep gRPC out of scope for these changes as well.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -278,6 +283,24 @@ public boolean authenticate(final SecurityRequestChannel request) {
...
+            final String providedSecret = request.header(ConfigConstants.SECURITY_SUPERADMIN_SECRET_HEADER);
+            if (isSuperadminSecretValid(providedSecret)) {
```
finnegancarroll: Do we want to audit log on a failed check since this is similar to a failed basic auth attempt?

finnegancarroll (changes requested):

src/main/java/org/opensearch/security/support/SecuritySettings.java
itsmevichu: The existing adminDN flow already had similar behavior, so I initially followed the same approach here. But I agree that insecure settings should not be allowed when `allow_insecure_settings` is false. This was something I had already noted and have now addressed. Thanks for pointing it out.

itsmevichu:
Apologies for the delay, I was completely occupied last month. I've now pushed my changes addressing the review comments. Please take a look when you have a chance.

DarshitChanpura:
@itsmevichu Thank you for this contribution. Could you please resolve the conflicts?

DarshitChanpura:
@itsmevichu Are you actively working on this PR? If not, it will be closed as stale during next cleanup.

PR #6076, opened 2026-04-07
Add indices:admin/mappings/get to ml_full_access role for Agentic Search

owaiskazi19:
### Description

Adds `indices:admin/mappings/get` to the `ml_full_access` role to support Agentic Search (`QueryPlanningTool`). QPT needs to retrieve index mappings to generate accurate OpenSearch DSL queries from natural language input.

### Dependency

This PR depends on [opensearch-project/ml-commons#4785](https://github.com/opensearch-project/ml-commons/pull/4785), which changes `QueryPlanningTool` to use `GetMappingsRequest` instead of `GetIndexRequest`. Without that ml-commons change, QPT would require the broader `indices:admin/get` permission. With it, only `indices:admin/mappings/get` is needed.

### Security implications

**This permission does not expose system index mappings.** System indices are protected by multiple layers:

1. **System index permission layer** — When `plugins.security.system_indices.permissions.enabled` is `true` (default), accessing system indices requires the `system:admin/system_index` permission, which `ml_full_access` does not have.
2. **Security config index protection** — The `.opendistro_security` index has additional hardcoded protection that filters it from wildcard requests when `plugins.security.filter_securityindex_from_all_requests` is enabled.
3. **Scoped access** — QPT calls `getMappings` with a specific user-provided index name, not `*`.

**This is a read-only metadata permission** — it only exposes field names and types, not document data.

**Consistent with existing roles** — `indices:admin/mappings/get` is already used by `knn_full_access`, `query_assistant_access`, and `reports_read_access` with the same `index_patterns: ['*']` scope.

### Related Issue

opensearch-project/ml-commons#4775

### Check List
  - N/A (config-only change to a predefined role)
  - N/A
- [x] Commits are signed off as per the DCO using `--signoff`


config/roles.yml
```diff
@@ -332,6 +332,7 @@ ml_full_access:
     - index_patterns:
         - '*'
       allowed_actions:
+        - 'indices:admin/mappings/get'
```
rithinpullela, line 335: we also need `indices:data/read/search*` because agentic search also samples documents from the index, should we add that too?

owaiskazi19: Good call. QPT's async chain has two steps: (1) `getMappings` for index structure, and (2) a search to sample a document. `indices:data/read/search*` would be needed for it

rithinpullela (approved):

DarshitChanpura:
@owaiskazi19 Is this PR ready for review? If not, this will be marked as stale and closed during next cleanup.

PR #6078, opened 2026-04-10
Support Jackson 3.x release line

reta:
### Description
Support Jackson 3.x release line

### Issues Resolved
Closes https://github.com/opensearch-project/security/issues/6075

### Testing
Covered by existing test suites

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -483,7 +483,11 @@ configurations {
             force 'commons-logging:commons-logging:1.3.6'
             force 'org.slf4j:slf4j-api:1.7.36'
             force 'org.scala-lang:scala-library:2.13.18'
-            force "com.fasterxml.jackson:jackson-bom:${versions.jackson}"
+            force "tools.jackson:jackson-bom:${versions.jackson3}"
+            force "tools.jackson.core:jackson-core:${versions.jackson3}"
+            force "tools.jackson.datatype:jackson-datatype-jdk8:${versions.jackson3}"
+            force "tools.jackson.core:jackson-databind:${versions.jackson3_databind}"
+            // Keeping Jackson 2.x since Kafka clients and etc still need it
             force "com.fasterxml.jackson.core:jackson-core:${versions.jackson}"
             force "com.fasterxml.jackson.datatype:jackson-datatype-jdk8:${versions.jackson}"
             force "com.fasterxml.jackson.core:jackson-databind:${versions.jackson_databind}"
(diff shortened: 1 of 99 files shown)


src/main/java/org/opensearch/security/tools/SecurityAdmin.java
```diff
@@ -1261,15 +1262,17 @@ private static int issueWarnings(RestHighLevelClient restHighLevelClient) throws
 
             Version maxVersion = Version.fromString(
```
reta, line 1263: This is very interesting, I honestly don't know how this particular check was working before, but calling `asText()` for `ObjectNode` always returns `""` for Jackson 2.x (in Jackson 3.x it is more strict and raises an exception). The `version` property should have been extracted instead.

src/main/java/org/opensearch/security/DefaultObjectMapper.java
```diff
@@ -61,48 +67,76 @@ public ConfigMapSerializer() {
...
         }
         gen.writeEndObject();
     }
 }
 
 public class DefaultObjectMapper {
-    public static final ObjectMapper objectMapper = new ObjectMapper();
-    public final static ObjectMapper YAML_MAPPER = new ObjectMapper(new YAMLFactory());
-    private static final ObjectMapper defaulOmittingObjectMapper = new ObjectMapper();
+    public static volatile ObjectMapper objectMapper;
```
reta: @cwperks this is probably the most important change: security plugin heavily relies on mutability of `ObjectMapper` for injectable values support. In Jackson 3.x, `ObjectMapper` is immutable and the only way to perform the dynamic injections is to rebuild the mappers on demand, hence they won't be `final` anymore.

build.gradle
```diff
@@ -706,8 +710,11 @@ dependencies {
     implementation "io.jsonwebtoken:jjwt-api:${jjwt_version}"
     implementation "io.jsonwebtoken:jjwt-impl:${jjwt_version}"
     implementation "io.jsonwebtoken:jjwt-jackson:${jjwt_version}"
+    implementation "com.fasterxml.jackson.core:jackson-core:${versions.jackson}"
```
reta, line 713: Keeping Jackson 2.x for `jjwt-jackson` since it does not support Jackson 3.x yet

src/main/java/org/opensearch/security/DefaultObjectMapper.java
cwperks: Do you know what dynamic injections security performs?

src/main/java/org/opensearch/security/tools/SecurityAdmin.java
cwperks: I suspect this check was not working. Do we have any automated tests around it?

Its definitely a good idea to enforce that securityadmin cannot be run in a mixed cluster which is the spirit of this check and should be fixed if it is broken!

src/main/java/org/opensearch/security/DefaultObjectMapper.java
reta: I don't know the exact reasons regarding the need to inject values, but we have two places where `Settings` are being injected:
 - `SecurityAdmin::execute`
 - `OpenSearchSecuritySSLPlugin`

src/main/java/org/opensearch/security/tools/SecurityAdmin.java
reta: > I suspect this check was not working. Do we have any automated tests around it?

We do (some tests were failing because Jackson now throws an exception), however I don't think we have tests that actually hit mixed versions.

cwperks (approved):
ty @reta ! The changes LGTM. Thanks for the help with the upgrade.

reta:
@willyborankin @finnegancarroll @DarshitChanpura folks looking for second look please, thank you

willyborankin (approved):

PR #6111, opened 2026-04-23
support fallback values in DLS/FLS variables

### LINKED ISSUE #6110: [FEATURE] support fallback values in DLS/FLS variables
currently DLS/FLS variables need to be present and there is no way to define a fallback.

e.g. if a DLS query contains `${attr.jwt.foo}` then the claim `foo` must be present in the JWT, otherwise the request fails.

it should be possible to define fallback values (which themselves can be variables again), e.g. `${attr.jwt.foo:-baz}` (replacing it with "baz" if the `foo` claim is not present) or `$attr.jwt.foo:-${attr.jwt.bar}}` (replacing it with the value of the `bar` claim if the `foo` claim is not present; if both are not present it'll still fail; otherwise it'd need `${attr.jwt.foo:-${attr.jwt.bar:-}}` to default to an empty string)

assigning different roles with different DLS/FLS queries

in 2.x undefined attributes were silently ignored.
see also #1310 & #5975 

the current implementation is in `UserAttributes#replaceProperties`:
https://github.com/opensearch-project/security/blob/dc82fb7c11b119db54007fc69eb2c347fdf948fc/src/main/java/org/opensearch/security/privileges/UserAttributes.java#L52-L66

apache commons text provides [`StringSubstitutor` which does exactly what we want](https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/StringSubstitutor.html#providing-default-values-heading).

rursprung:
### Description
* Category: Enhancement

see the individual commit messages for details. basically this uses [`StringSubstitutor`](https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/StringSubstitutor.html) from apache-commons text to re-create the existing behaviour (first commit) and also support fallbacks in the variables.

if a user attribute is not defined then the system fails with an exception. in 2.x this was silently ignored, allowing use-cases where attributes were optional.

with this we now support fallbacks, thus it's now possible to e.g. write `${attr.proxy.attr2:-foo}` so it prints "foo" if `attr2` is not defined or even `${attr.proxy.attr2:-${attr.proxy.attr1}}` to fall back to the value of `attr1` if `attr2` is not defined.

### Issues Resolved
resolves #6110

### Testing
unit testing

### Check List
- [ ] New functionality has been documented - docs PR will be raised once this PR is approved
- [ ] ~~New Roles/Permissions have a corresponding security dashboards plugin PR~~
- [ ] ~~API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md)~~


src/main/java/org/opensearch/security/privileges/UserAttributes.java
```diff
@@ -50,39 +49,24 @@ public static List<String> findUnresolvedAttributes(String s) {
...
+        replacementsWithDots.forEach((k, v) -> replacements.put(k.replace(".", "_"), v));
 
-    private static String replaceRoles(final String orig, final User user) {
-        String retVal = orig;
-        if (orig.contains("${user.roles}") || orig.contains("${user_roles}")) {
-            final String commaSeparatedRoles = toQuotedCommaSeparatedString(user.getRoles());
-            retVal = orig.replace("${user.roles}", commaSeparatedRoles).replace("${user_roles}", commaSeparatedRoles);
-        }
-        return retVal;
+        final var stringSubstitutor = new StringSubstitutor(replacements).setEnableSubstitutionInVariables(true);
```
Ganesh-RB, line 59: Is there a specific use case for setting `setEnableSubstitutionInVariables` to true? It can lead to cyclic references and throw an `IllegalStateException`.

Also it introduces a risk of variable injection in DLS queries.

rursprung: yes, enabling this is what this PR is all about: it gives the possibility to write `${attr.proxy.attr2:-${attr.proxy.attr1}}` which is exactly what i need in my setup (i can't influence the IdP setup and i have no role to identify whether i might get `attr2` or not (otherwise i'd use two separate roles and separate role mappings); i do however know that `attr1` is always present and that `attr2` takes precedence if it is present)

rursprung:
the AI-analysis is wrong: without `.setEnableSubstitutionInVariables(true)` the test case with `${attr.proxy.attr2:-${attr.proxy.attr1}}` does not work:
```
org.junit.ComparisonFailure: expected:<...alue1",
    "baz": "[value1]",
    "roles": ["ro...> but was:<...alue1",
    "baz": "[${attr.proxy.attr1}]",
    "roles": ["ro...>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at org.opensearch.security.privileges.UserAttributesUnitTest.testReplaceProperties(UserAttributesUnitTest.java:78)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
```

this is why i did it in the first place.

security-wise i think this change is ok, we'll just have to document that `${${attr.something}}` is a really bad idea (i'm not sure why anyone would do this, anyway). we could even forcibly check for `${${` and throw an exception if we see that to prevent it completely?

cwperks:
> security-wise i think this change is ok, we'll just have to document that ${${attr.something}} is a really bad idea (i'm not sure why anyone would do this, anyway). we could even forcibly check for ${${ and throw an exception if we see that to prevent it completely?

Given that the input is from admin and just not any API caller its fine. We couldn't certainly add better validation on Security APIs but not in the scope of this PR.

rursprung:
rebased & incorporated @cwperks' slack feedback (`attr2` -> `notexists` in the test to make it more obvious)

nibix (approved):
Look good to me!

IMHO, the whole attribute substitution is in a bad need of a complete redesign, but this is a good enhancement! 👍

rursprung:
thanks! btw, the CI failure is likely a false reject (it worked fine before and i only rebased).

src/main/java/org/opensearch/security/privileges/UserAttributes.java
```diff
@@ -42,39 +41,24 @@ public static List<String> findUnresolvedAttributes(String s) {
...
```
reta, line 59: I think there are two concerns from my side:
 - new failure mode with cycles (already brought here https://github.com/opensearch-project/security/pull/6111/changes#r3133408405)
 - bringing a whole new library for simple string substitution (if we want to keep it simple) looks excessive to me

Do we want a full power of `StringSubstitutor`? May be we do, from other perspective keeping it very simple, non recursive, would be safer I believe.

rursprung: as said, i need the option "if A is not set fall back to B" where both A and B are attributes (in my case on the JWT).
in 2.x we solved this with two queries in a `bool` which worked because it just ignored the non-existent variable and this part then didn't match.
having the proper fallback will also be good for our performance (this is actually a TLQ we're doing, so doing it once instead of twice would be a big win)

as for bringing in a new library: it was already present as `runtimeOnly` - the only difference is that now i can also use it in this code base rather than it just being around at runtime for some other dependencies. so i don't see this is a big problem?

reta: > as said, i need the option "if A is not set fall back to B" where both A and B are attributes (in my case on the JWT).

This part is fine, probably was not clearly articulated, +1 to have fallback, but StringSubstitutor will throw an exception on recursive substitution, whereas we could not push that as far by extending the existing manual substitution.  

>  so i don't see this is a big problem?

you are actually very right, we bundle it even now, non issue

reta (approved):

src/main/java/org/opensearch/security/privileges/UserAttributes.java
rursprung: > but StringSubstitutor will throw an exception on recursive substitution, whereas we could not push that as far by extending the existing manual substitution.

ah, ok. this isn't a regression as this wasn't possible so far. and i think also a custom implementation would have to judge what to do after the first round of replacement. in the current implementation (with no fallback) it also raises an exception on finding a placeholder after having replaced the known ones:
https://github.com/opensearch-project/security/blob/18b1a03629285089391e1b31521b77e63fa7a5e8/src/main/java/org/opensearch/security/privileges/dlsfls/DocumentPrivileges.java#L177-L179

and i think/hope that people will not write such complex replacement code in a DLS (then again: who knows? i'm to blame for even having TLQ support in DLS 😅) and if they do that they'll test it.

rursprung:
btw, talking of exceptions: there's [`StringSubstitutor#setEnableUndefinedVariableException`](https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/StringSubstitutor.html#setEnableUndefinedVariableException(boolean)) which could be used as a replacement for the manual check written by @ThyTran1402 (the `IllegalArgumentException` would then still have to be converted to the current exception to preserve having the "nice" output if you want to keep that)

but i'd strongly suggest that this is a follow-up cleanup

PR #6112, opened 2026-04-23
Fix NPE in LDAPAuthorizationBackend when rolesearch disabled

### LINKED ISSUE #5832: [BUG] LDAP auth broke after upgrading from opensearch:2 to opensearch:3
**What is the bug?**
There are two methods to integrate LDAP auth documented here: https://docs.opensearch.org/latest/security/authentication-backends/ldap/#approach-1-query-the-role-subtree
https://docs.opensearch.org/latest/security/authentication-backends/ldap/#approach-2-use-a-users-attribute-as-the-role-name

Approach 2 no longer works after upgrading. 
The LDAP server(FreeIPA) has all group memberships under the user as memberOf attribute. 

Use a config similar to: 
```
      authz:
        ldap_roles:
          description: "Authorize using LDAP"
          http_enabled: true
          transport_enabled: true
          authorization_backend:
            type: ldap
            config:
              enable_ssl: false
              enable_start_tls: false
              enable_ssl_client_auth: false
              verify_hostnames: true
              hosts:
                - "freeipa-1:389"
              bind_dn: uid=opensearch,cn=users,cn=accounts,dc=example,dc=com
              password: "${password}"
              userbase: cn=users,cn=accounts,dc=example,dc=com
              usersearch: (uid={0})
              username_attribute: uid
              skip_users:
                - admin
                - kibanaserver
              userroleattribute: memberOf
              rolesearch_enabled: false
              rolebase: cn=groups,cn=accounts,dc=example,dc=com
              rolename: cn
```

Roles to be resolved as they were in opensearch:2

 - OS: Linux (docker)
 - Version: 9
 - Plugins (default in the image)

Related Forum posts: 
https://forum.opensearch.org/t/ldap-login-does-not-work-after-upgrade-from-2-19-1-to-3-1-0/25603
https://forum.opensearch.org/t/solved-unable-to-log-in-with-ldap-after-upgrade-to-3-3/27360/4

tronboto:
### Description

* Bug fix

When `rolesearch_enabled: false` is configured and the user authenticated via ldap, an NPE is thrown:

```
Cannot invoke "org.ldaptive.Connection.getProviderConnection()" because
the return value of "org.ldaptive.SearchOperation.getConnection()" is null
```

This happens because a connection to ldap is never opened in the user lookup block as the user's DN is already known. The code then falls through to the else branch that calls `getRoleFromEntry()` to resolve role names, but connection is still null at this point, causing the NPE.

The rolesearch block (and the nested roles block) already have null guards which explains why enabling rolesearch worked as a workaround.

### Issues Resolved

Fixes #5832

Is this a backport? No

Do these changes introduce new permission(s) to be displayed in the static dropdown on the front-end? No

### Testing

Added `testLdapAuthorizationRolesearchDisabledWithLdapAuthContext` which authenticates the user with an `LDAPAuthenticationBackend` then calls `addRoles` with `rolesearch_enabled: false`.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/auth/ldap/backend/LDAPAuthorizationBackend.java
+++ b/src/main/java/org/opensearch/security/auth/ldap/backend/LDAPAuthorizationBackend.java
@@ -936,6 +936,10 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
 
             } else {
                 // DN roles, extract rolename according to config
+                if (connection == null) {
+                    connection = getConnection(settings, configPath);
+                }
+
                 for (final LdapName roleLdapName : ldapRoles) {
                     final String role = getRoleFromEntry(connection, roleLdapName, roleName);
 
--- a/src/main/java/org/opensearch/security/auth/ldap/backend/LDAPAuthorizationBackend.java
+++ b/src/main/java/org/opensearch/security/auth/ldap/backend/LDAPAuthorizationBackend.java
@@ -832,15 +832,15 @@ public class LDAPAuthorizationBackend implements AuthorizationBackend {
                 userRoleAttributeValue = Utils.getSingleStringValue(userRoleAttribute);
             }
 
+            if (connection == null) {
+                connection = getConnection(settings, configPath);
+            }
+
             if (rolesearchEnabled) {
                 String escapedDn = dn;
 
                 if (isDebugEnabled) {
                     log.debug("DBGTRACE (8): escapedDn" + escapedDn);
-                }
-
-                if (connection == null) {
-                    connection = getConnection(settings, configPath);
                 }
 
                 for (Map.Entry<String, Settings> roleSearchSettingsEntry : roleBaseSettings) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/security/auth/ldap/backend/LDAPAuthorizationBackend.java
```diff
@@ -936,6 +936,10 @@ public User addRoles(User user, AuthenticationContext context) throws OpenSearch
 
             } else {
                 // DN roles, extract rolename according to config
+                if (connection == null) {
```
cwperks: nit: Should we assign the connection before the if block if its being assigned in both paths?

cwperks (dismissed):

src/main/java/org/opensearch/security/auth/ldap/backend/LDAPAuthorizationBackend.java
tronboto: Makes sense - updated @cwperks

cwperks: it can be before `if (rolesearchEnabled) {` and then L842-844 could be deleted as redundant.

cwperks (dismissed):

src/main/java/org/opensearch/security/auth/ldap/backend/LDAPAuthorizationBackend.java
tronboto: Done 👍

cwperks (approved):

willyborankin (approved):

PR #6123, opened 2026-04-28
[FIX] Preserve response headers across context restore in SecurityInterceptor

KishoreKicha14:
### Description

Use newRestorableContext(true) instead of stashContext() for context restoration in RestoringTransportResponseHandler to preserve response headers (including TASK_RESOURCE_USAGE) across transport round-trips.

Category: Bug fix

Why these changes are required?

When the security plugin is enabled, `SecurityInterceptor.sendRequestDecorate()` stashes the ThreadContext via stashContext() before sending transport requests. On response, RestoringTransportResponseHandler.handleResponse() calls StoredContext.restore(), which blindly overwrites the thread-local context — discarding all response headers added during shard-level execution on data nodes. This causes shard-level TASK_RESOURCE_USAGE entries to be silently dropped, so the coordinator only sees its own resource usage.

Old behavior: RestoringTransportResponseHandler used ThreadContext.StoredContext from stashContext(), whose restore() discards response headers added after stashing. Shard-level task resource usage data was lost when security plugin was enabled.

New behavior: RestoringTransportResponseHandler uses Supplier<ThreadContext.StoredContext> from newRestorableContext(true), whose get() merges current response headers back into the restored context — matching the pattern used by OpenSearch core's ContextRestoreResponseHandler. All response headers including TASK_RESOURCE_USAGE now survive context restore.

### Issues Resolved

### Testing

1. Unit tests verifying TASK_RESOURCE_USAGE response header survives context restore in handleResponse()
2. Unit tests verifying multiple response headers survive context restore
3. Preservation tests for DLS/FLS/masked-field transient header propagation on ClusterSearchShardsResponse
4. Preservation tests for handleException() context restore and exception propagation
5. Preservation tests for handleStreamResponse() direct delegation
6. Combinatorial tests covering all 8 subsets of {DLS, FLS, MaskedField} headers
7. Randomized TransportException variant tests for exception handling robustness

Manual integration test — built OpenSearch 3.7.0-SNAPSHOT with security + query-insights plugins, created single-shard index, ran search, confirmed _insights/top_queries returns both entries in task_resource_usages:

```
"task_resource_usages": [
  {
    "action": "indices:data/read/search[phase/query]",
    "taskId": 156,
    "parentTaskId": 155,
    "nodeId": "nJn3W-S9TwqMHeWPcsK9gw",
    "taskResourceUsage": {
      "cpu_time_in_nanos": 27630000,
      "memory_in_bytes": 8088664
    }
  },
  {
    "action": "indices:data/read/search",
    "taskId": 155,
    "parentTaskId": -1,
    "nodeId": "nJn3W-S9TwqMHeWPcsK9gw",
    "taskResourceUsage": {
      "cpu_time_in_nanos": 3343000,
      "memory_in_bytes": 112632
    }
  }
]
```

### Check List
- [X ] Commits are signed per the DCO using --signoff

### DIFF
--- a/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
+++ b/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
@@ -178,8 +178,12 @@ public class SecurityInterceptor {
             requestHeadersToCopy.removeAll(Task.REQUEST_HEADERS); // Special case where this header is preserved during stashContext.
         }
 
+        final Supplier<ThreadContext.StoredContext> restorableContextSupplier = getThreadContext().newRestorableContext(true);
         try (ThreadContext.StoredContext stashedContext = getThreadContext().stashContext()) {
-            final TransportResponseHandler<T> restoringHandler = new RestoringTransportResponseHandler<T>(handler, stashedContext);
+            final TransportResponseHandler<T> restoringHandler = new RestoringTransportResponseHandler<T>(
+                handler,
+                restorableContextSupplier
+            );
             getThreadContext().putHeader("_opendistro_security_remotecn", cs.getClusterName().value());
 
             final Map<String, String> headerMap = new HashMap<>(
@@ -377,10 +381,13 @@ public class SecurityInterceptor {
     // which is private scoped
     private class RestoringTransportResponseHandler<T extends TransportResponse> implements TransportResponseHandler<T> {
 
-        private final ThreadContext.StoredContext contextToRestore;
+        private final Supplier<ThreadContext.StoredContext> contextToRestore;
         private final TransportResponseHandler<T> innerHandler;
 
-        private RestoringTransportResponseHandler(TransportResponseHandler<T> innerHandler, ThreadContext.StoredContext contextToRestore) {
+        private RestoringTransportResponseHandler(
+            TransportResponseHandler<T> innerHandler,
+            Supplier<ThreadContext.StoredContext> contextToRestore
+        ) {
             this.contextToRestore = contextToRestore;
             this.innerHandler = innerHandler;
         }
@@ -407,7 +414,7 @@ public class SecurityInterceptor {
             List<String> dlsResponseHeader = responseHeaders.get(ConfigConstants.OPENDISTRO_SECURITY_DLS_QUERY_HEADER);
             List<String> maskedFieldsResponseHeader = responseHeaders.get(ConfigConstants.OPENDISTRO_SECURITY_MASKED_FIELD_HEADER);
 
-            contextToRestore.restore();
+            contextToRestore.get();
 
             final boolean isDebugEnabled = log.isDebugEnabled();
             if (response instanceof ClusterSearchShardsResponse) {
@@ -438,8 +445,9 @@ public class SecurityInterceptor {
 
         @Override
         public void handleException(TransportException e) {
-            contextToRestore.restore();
-            innerHandler.handleException(e);
+            try (ThreadContext.StoredContext ignore = contextToRestore.get()) {
+                innerHandler.handleException(e);
+            }
         }
 
         @Override
--- a/src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
+++ b/src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
@@ -464,9 +464,11 @@ public class SecurityInterceptorTests {
     @Test
     public void testTaskResourceUsageResponseHeaderSurvivesContextRestore() {
         final String TASK_RESOURCE_USAGE = "TASK_RESOURCE_USAGE";
-        final String resourceUsageValue = "{\"action\":\"indices:data/read/search[phase/query]\","
-            + "\"taskId\":1,\"parentTaskId\":2,\"nodeId\":\"dataNode1\","
-            + "\"taskResourceUsage\":{\"cpu_time_in_nanos\":123,\"memory_in_bytes\":456}}";
+        final String resourceUsageValue = """
+            {"action":"indices:data/read/search[phase/query]",\
+            "taskId":1,"parentTaskId":2,"nodeId":"dataNode1",\
+            "taskResourceUsage":{"cpu_time_in_nanos":123,"memory_in_bytes":456}}\
+            """;
 
         final AtomicReference<Map<String, List<String>>> responseHeadersAfterRestore = new AtomicReference<>();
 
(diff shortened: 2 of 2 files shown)


cwperks:
@KishoreKicha14 can you please take a look at the unit test failures? Looks like it might be related to the change in this PR.

The change looks sensible to me and I'm surprised it hasn't been brought up before. How exactly does QI use these response headers in a way that other plugins do not?

Edit: appears to be due to flakiness

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -449,4 +457,601 @@ public void testStreamRequestType() {
...
+     * Property-based style test: Generate random TransportException instances and verify
+     * handleException() restores context and delegates to inner handler for each.
+     *
+     * Tests with various exception messages and causes to ensure robust exception handling.
+     *
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    @Test
+    public void testPreservation_RandomTransportExceptions_HandleExceptionRestoresAndDelegates() {
+        Random random = new Random(42); // Fixed seed for reproducibility
```
dzane17: Variable is unused

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -438,7 +445,7 @@ public void handleResponse(T response) {
 
         @Override
         public void handleException(TransportException e) {
-            contextToRestore.restore();
+            contextToRestore.get();
             innerHandler.handleException(e);
         }
```
dzane17, line 451: Nit: `contextToRestore.get()` returns a `Closeable`. You could close it by wrapping the call in a try-with-resources, like:

```java
@Override
public void handleException(TransportException e) {
    try (ThreadContext.StoredContext ignore = contextToRestore.get()) {
        innerHandler.handleException(e);
    }
}
```

Reference: [TransportService.java#L1606-L1608](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/transport/TransportService.java#L1606-L1608)

cwperks (dismissed):

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -449,4 +456,599 @@ public void testStreamRequestType() {
...
 
+    /**
+     * Verifies that TASK_RESOURCE_USAGE response header survives context restore
+     * in RestoringTransportResponseHandler.handleResponse().
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    @Test
+    public void testTaskResourceUsageResponseHeaderSurvivesContextRestore() {
+        final String TASK_RESOURCE_USAGE = "TASK_RESOURCE_USAGE";
+        final String resourceUsageValue = "{\"action\":\"indices:data/read/search[phase/query]\","
```
cwperks: nit: prefer Text Blocks for large JSON strings

cwperks: @KishoreKicha14 can there also be an integ test for this in QI repo?

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -449,4 +456,599 @@ public void testStreamRequestType() {
...
+
+    /**
+     * Verifies that ALL response headers (TASK_RESOURCE_USAGE + arbitrary custom headers)
+     * survive context restore in RestoringTransportResponseHandler.handleResponse().
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    @Test
+    public void testMultipleResponseHeadersSurviveContextRestore() {
+        final String TASK_RESOURCE_USAGE = "TASK_RESOURCE_USAGE";
+        final String resourceUsageValue = "{\"action\":\"indices:data/read/search[phase/query]\","
```
cwperks: same as above. prefer Text Blocks.

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
```diff
@@ -178,8 +178,12 @@ public <T extends TransportResponse> void sendRequestDecorate(
             requestHeadersToCopy.removeAll(Task.REQUEST_HEADERS); // Special case where this header is preserved during stashContext.
         }
 
+        final Supplier<ThreadContext.StoredContext> restorableContextSupplier = getThreadContext().newRestorableContext(true);
```
cwperks, line 181: Wow, TIL that ThreadContext has a notion of:

1. Request Headers
2. Response Headers

They're loosely analogous to HTTP headers, but since these are internal transport node-to-node they are not HTTP Headers.

The difference is as follows:

1. Request Headers are a Map<String, String> data structure
2. Response Headers are Map<String, Set<String>> data structure
3. Transient headers are Map<String, Object> data structure


super confusing IMO

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
KishoreKicha14: Thanks. I will create a issue for this in qi repo and add integ tests

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
cwperks: Would it make sense to make this change in core similar to https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/util/concurrent/ThreadContext.java#L157-L165?

cwperks (approved):

src/main/java/org/opensearch/security/transport/SecurityInterceptor.java
DarshitChanpura: IMO it should be find here as it simply handles the header.

DarshitChanpura (approved):

src/test/java/org/opensearch/security/transport/SecurityInterceptorTests.java
```diff
@@ -449,4 +456,603 @@ public void testStreamRequestType() {
         completableRequestDecorate(jdkSerializedSender, connection1, action, request, streamOptions, handler, localNode);
     }
 
+    /**
+     * Verifies that TASK_RESOURCE_USAGE response header survives context restore
+     * in RestoringTransportResponseHandler.handleResponse().
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    @Test
+    public void testTaskResourceUsageResponseHeaderSurvivesContextRestore() {
```
DarshitChanpura, line 465: i see that you added a manual after fix view in PR description. Can you add a before-fix view as well for comparison?

PR #6147, opened 2026-05-18
fix(config)!: remove own_index default roles mapping

robertpaschedag:
### Description
Remove the `own_index` default roles mapping from `config/roles_mapping.yml`.

* Category: Maintenance
* Why these changes are required? The `own_index` roles mapping maps all users (`*`) to a role that grants full access to an index named like the username. This is an overly permissive default that weakens the security posture out of the box.
* Old behavior: All users automatically received full access to an index matching their username via the default roles mapping. New behavior: No default mapping for `own_index` exists; administrators must explicitly configure this mapping if needed.

We want to use the [opensearch-k8s-operator](https://github.com/opensearch-project/opensearch-k8s-operator) to setup our clusters. The overall defaults (`roles` and `roles_mapping`) are very good (and we want to keep them). But with additional authentication backends (like OIDC / SAML / LDAP) this would allow *any* authenticated user to potentially create an index with their name (which is not what we want in a production cluster).

Still...the old behaviour can be enabled again by providing this roles mapping manually again (if this is needed). As I do not know, if truly users exist, that rely on this roles mapping....this could be a BREAKING CHANGE.

This `own_index` roles mapping makes perfectly sense for some "demo / gym / playground" cluster to "play around".

### Issues Resolved
N/A

### Testing
No new testing required — this is a configuration-only change removing a default mapping. Existing tests pass.

### Check List

### DIFF
--- a/config/roles_mapping.yml
+++ b/config/roles_mapping.yml
@@ -16,12 +16,6 @@ all_access:
   - "admin"
   description: "Maps admin to all_access"
 
-own_index:
-  reserved: false
-  users:
-  - "*"
-  description: "Allow full access to an index named like the username"
-
 logstash:
   reserved: false
   backend_roles:


cwperks:
Thank you for this PR @robertpaschedag . This is also something I'd like to do, but I think it should be done on a major release given that it could be considered a Breaking Change. Is there a workaround that would be suitable for your setup before a major release?

Edit: On second thought, I would be ok with removing this in a minor version. @nibix thoughts?

robertpaschedag:
> Thank you for this PR @robertpaschedag . This is also something I'd like to do, but I think it should be done on a major release given that it could be considered a Breaking Change. Is there a workaround that would be suitable for your setup before a major release?
> 
> Edit: On second thought, I would be ok with removing this in a minor version. @nibix thoughts?

@cwperks I think, the only true workaround would be to pass a copy of the `roles_mapping.yml` (with `own_index` removed) into the **secret** (used by the operator). Even if the *operator* would support to make sure, that some "object" is `absent`, this would only last as long as the `securityconfig` job does not run again.

nibix:
@cwperks @robertpaschedag 

I think it is totally fine to remove this role mapping in a minor version, for the following reasons:

- It only affects newly set up clusters, existing clusters are not affected
- The role mapping itself seems to be undocumented, so the awareness is low (this is actually a good reason in support of removing it)

We should however document this in the CHANGELOG.md file to provide minimum awareness just in case people observe unexpected changes.

cwperks:
> We should however document this in the CHANGELOG.md file to provide minimum awareness just in case people observe unexpected changes.

@nibix the CHANGELOG process has been supplanted by a groomed git log + LLMs to help craft the release notes text. For 3.7.0 release there will be a release notes PR which we can edit from there.

cwperks (approved):

robertpaschedag:
should I update the branch again (merge latest code) or is this something you will take care of? Thanks.

cwperks:
> should I update the branch again (merge latest code) or is this something you will take care of? Thanks.

no need, that failing check is flaky

PR #6151, opened 2026-05-19
Introduce an AGENTS.MD file

terryquigleysas:
This is very much a proposal / first draft for both issues addressed. I will defer to those with more extensive knowledge on both counts.

### Description
* Category (Enhancement, New feature, Maintenance, Documentation)
Introduce an AGENTS.MD file to provide instructions to agents to improve the experience.
Uses https://github.com/terryquigleysas/OpenSearch/blob/main/AGENTS.md as a guide.
Also update the backporting instructions due to recent changes.

### Issues Resolved
Improve the agentic experience.

### Testing
We have found this to improve the agentic experience on our own applicatons.

### Check List


AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+tools/              # Shell scripts: securityadmin.sh, install_demo_configuration.sh
+scripts/            # CI/CD helper scripts
+bwc-test/           # Backwards-compatibility tests
+checkstyle/         # Checkstyle rule configuration
+formatter/          # Eclipse JDT formatter configuration
+```
+
+## Build
+
+**Minimum JDK: 21.** `JAVA_HOME` must be set.
```
cwperks, line 71: Any idea what happens when Agents.md file becomes outdated? Does the agent automatically understand that and update accordingly? I know where I'm at we talk about dynamic knowledge bases frequently to ensure that agents can work with up-to-date information.

AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+./gradlew precommit               # precommit checks only (run before every commit)
+./gradlew spotlessApply           # auto-fix Java formatting
+```
+
+## Testing
+
+### Unit Tests
+
+- Located in `src/test/`
+- Run with: `./gradlew test`
```
cwperks, line 91: Is this accurate given the test split we have in build.gradle? 

This would actually be of great use to help agents understand the correct test task to run depending on the high-level package that a test suite is in.

AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+- HTML results: `build/reports/tests/test/index.html`
+- Run a specific test:
+  ```bash
+  ./gradlew test --tests "org.opensearch.security.auth.BackendRegistryTests"
+  ```
+- Repeat a test for reliability using the `@Repeat` annotation:
```
cwperks, line 97: 🙌 . Great idea to include this...particularly if asking an agent to investigate test flakiness.

AGENTS.md
cwperks: oh, its done down below :)

AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+- **SPI changes** (`spi/`) are public API consumed by external plugins. Apply `@PublicApi`, `@InternalApi`, or `@ExperimentalApi` annotations and maintain backwards compatibility accordingly.
+- **Resource sharing** changes should be validated against the sample plugin in `sample-resource-plugin/`.
+- **FIPS compliance** — the build supports FIPS-140-3 mode (`gradle.properties`). Avoid cryptographic primitives that are not FIPS-approved.
+
+## Adding Dependencies
+
+When adding or removing a non-test scope dependency in `build.gradle`:
+
+1. Copy the library's `LICENSE.txt` and `NOTICE.txt` into `licenses/` as `<artifact>-LICENSE.txt` and `<artifact>-NOTICE.txt`.
+2. Run `./gradlew updateSHAs` to regenerate SHA checksum files.
```
cwperks: This is actually disabled on this repo.

We certainly could re-enable it, but we'd be at the whim of upstream opensearch and constantly updating what's in `licenses/` depending on changes in core's version catalog.

I'd be amenable to enabling this if we can likewise update the logic to exclude licenses that overlap from the core. 

I know the performance-analyzer repo gets into a state where checks fail constantly and its very annoying.

AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+3. Verify with `./gradlew check`.
+
+## Commits
+
+Run `./gradlew precommit` before every commit.
```
cwperks: not relevant in this repo. 

I'm definitely not opposed to adding a precommit hook though

AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+Write commit titles focused on **user impact**, not implementation details:
+
+- ✅ `Enforce minimum TLS 1.2 on HTTP transport by default`
+- ❌ `Add tlsMinVersion field to SSLConfig`
+
+Commit title ≤ 50 characters. Leave a blank line before the body; wrap body at 72 characters. All commits must be signed off (DCO).
+
+## Pull Requests
+
+- Always push to your personal fork. Never push directly to `opensearch-project/security` or to `main`.
```
cwperks, line 195: yes and this is actually enforced now regardless of branch name! No maintainer can push to upstream.

AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+## Backports
+
+The automated backport workflow has been retired. Backports must be performed manually using your preferred Git workflow — the example below uses `git cherry-pick`, but other approaches (e.g. patch files or GUI tools) are equally valid.
+
+**Example using `git cherry-pick`:**
+
+```bash
+# Check out a new branch from the target release branch
+git fetch upstream
+git checkout -b backport/my-fix-2.x upstream/2.x
```
cwperks: Let's not use 2.x here as that is a dead branch. Let's mention something about 2.19 being in LTS.

DEVELOPER_GUIDE.md
```diff
@@ -316,8 +316,23 @@ See [CONTRIBUTING](CONTRIBUTING.md).
...
-merged to main, the workflow will create a backport PR to the `1.x` branch.
+The automated backport workflow has been retired. Backports must now be performed manually using your preferred Git workflow — the example below uses `git cherry-pick`, but other approaches (e.g. patch files or GUI tools) are equally valid.
+
+**Example using `git cherry-pick`** — to backport a merged PR to a release branch (e.g. `2.x`):
+
+1. Identify the commit SHA(s) from the merged PR on `main`.
+2. Check out the target branch locally:
+   ```bash
+   git fetch upstream
+   git checkout -b backport/my-fix-2.x upstream/2.x
```
cwperks: Same here

AGENTS.md
terryquigleysas: On my own application I added a section to explicitly direct agents to keep it up to date. I can do the same here.

I'm pretty sure some agents already know to do this but I will add something.

AGENTS.md
terryquigleysas: Thanks, I'll remove it.

AGENTS.md
terryquigleysas: Thanks, I'll remove it.

AGENTS.md
terryquigleysas: Thanks, I will change this.

DEVELOPER_GUIDE.md
```diff
@@ -316,8 +316,25 @@ See [CONTRIBUTING](CONTRIBUTING.md).
...
 ## Backports
 
-The Github workflow in [`backport.yml`](.github/workflows/backport.yml) creates backport PRs automatically when the
-original PR with an appropriate label `backport <backport-branch-name>` is merged to main with the backport workflow
-run successfully on the PR. For example, if a PR on main needs to be backported to `1.x` branch, add a label
-`backport 1.x` to the PR and make sure the backport workflow runs on the PR along with other checks. Once this PR is
-merged to main, the workflow will create a backport PR to the `1.x` branch.
+The automated backport workflow has been retired. Backports must now be performed manually using your preferred Git workflow — the example below uses `git cherry-pick`, but other approaches (e.g. patch files or GUI tools) are equally valid.
+
+The most common backport targets are the `2.19` and `3.6` branches, which are the current Long-Term Support (LTS) releases.
```
RyanL1997: For backport, I'd say the back port strategy used to target to all the LTS versions. I think it is good to mention the latest versions in 2.x line but I think it is not necessary to mention 3.6 as a specific version for 3.x line otherwise this needs to be updated very frequently as long as we have a new versions.

My suggestion is that we make this as generic as much as possible.

RyanL1997 (commented):
Hi @terryquigleysas , thanks for this change. Just left a comment regarding to the backport strategy.

DEVELOPER_GUIDE.md
RyanL1997:
```diff
-The most common backport targets are the `2.19` and `3.6` branches, which are the current Long-Term Support (LTS) releases.
+The most common backport targets are the current Long-Term Support (LTS) release branches. Check the project's release documentation for the active LTS versions.
```

terryquigleysas: @RyanL1997 Thank you . I will make the changes.

terryquigleysas:
@cwperks @RyanL1997 Apologies for the hassle but my branch got into a bad state and my changes, addressing your review comments, can now be found at https://github.com/opensearch-project/security/pull/6156

PR #6158, opened 2026-05-22
Allow audit config PUT when document does not exist

### LINKED ISSUE #5862: [BUG / FEATURE] Audit log REST API unusable if audit.yml not part of initial deployment
**What is the bug?**
If audit.yml is not deployed at initial cluster creation then none of the Audit Log REST API calls (GET,PUT,PATCH) will work.
See https://docs.opensearch.org/latest/security/access-control/api/#audit-logs
(Note the docs use the legacy url - we will update this separately)

Reloading all the security files again using securityadmin.sh is not a viable option for us on existing deployments.

I would expect there to be a way of calling the REST API to create the internal document, whether that be 
a) a new API method or a change to the existing calls 
OR 
b) to create the document with defaults as is done with other types

1. Deploy without audit.yml
2. Call GET /_opendistro/_security/api/audit
3. See returned message: `{"status":"NOT_FOUND","message":"Method GET not supported for this action."}`
4. Same is true for PUT and PATCH

The existing behaviour is likely by design based on the documentation, however I don't think it is correct that it means this part of the API is then unusable.

 - N/A

We have to support upgrades where this file has not been part of the initial deployment.
Reloading all the security files again using securityadmin.sh is not a viable option for us on existing deployments.

I'm pretty sure the code preventing this is a chain of method calls:
AuditApiAction.withEnabledAuditApi() 
-> ConfigurationRepository.isAuditHotReloadingEnabled() 
-> ConfigurationLoaderSecurity7.isAuditConfigDocPresentInIndex()

terryquigleysas:
### Description

* Category (Bug fix)

Previously, all Audit Log REST API operations (GET, PUT, PATCH) were blocked when no audit.yml was deployed at initial cluster creation. This created a chicken-and-egg problem where operators could not bootstrap the audit configuration via the API.
 
PUT now uses a new withCreateOrUpdateAuditApi() validation that allows bootstrapping when the document is absent and normal updates when it exists. GET and PATCH remain gated behind the existing check.

If audit.yml is not deployed at initial cluster creation then none of the Audit Log REST API calls (GET,PUT,PATCH) will work.

### Issues Resolved
Resolves https://github.com/opensearch-project/security/issues/5862

### Testing
New supporting tests added.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/dlic/rest/api/AuditApiAction.java
+++ b/src/main/java/org/opensearch/security/dlic/rest/api/AuditApiAction.java
@@ -267,10 +267,22 @@ public class AuditApiAction extends AbstractApiAction {
             .onChangeRequest(RestRequest.Method.PATCH, request -> withEnabledAuditApi(request).map(this::processPatchRequest))
             .onChangeRequest(
                 RestRequest.Method.PUT,
-                request -> withEnabledAuditApi(request).map(ignore -> processPutRequest("config", request))
+                request -> withCreateOrUpdateAuditApi(request).map(ignore -> processPutRequest("config", request))
             )
             .override(RestRequest.Method.POST, methodNotImplementedHandler)
             .override(RestRequest.Method.DELETE, methodNotImplementedHandler);
+    }
+
+    /**
+     * Allows PUT to bootstrap the audit config when the document does not exist,
+     * or to update it through normal validation when it does.
+     */
+    ValidationResult<RestRequest> withCreateOrUpdateAuditApi(final RestRequest request) {
+        if (!securityApiDependencies.configurationRepository().isAuditHotReloadingEnabled()) {
+            // Document does not exist — allow PUT to bootstrap the audit config
+            return ValidationResult.success(request);
+        }
+        return withEnabledAuditApi(request);
     }
 
     ValidationResult<RestRequest> withEnabledAuditApi(final RestRequest request) {
--- a/src/test/java/org/opensearch/security/dlic/rest/api/AuditApiActionValidationTest.java
+++ b/src/test/java/org/opensearch/security/dlic/rest/api/AuditApiActionValidationTest.java
@@ -12,11 +12,13 @@
 package org.opensearch.security.dlic.rest.api;
 
 import java.util.List;
+import java.util.Set;
 
 import org.junit.Test;
 
 import org.opensearch.common.settings.Settings;
 import org.opensearch.core.rest.RestStatus;
+import org.opensearch.rest.RestRequest;
 import org.opensearch.security.auditlog.config.AuditConfig;
 import org.opensearch.security.securityconf.impl.CType;
 import org.opensearch.security.securityconf.impl.SecurityDynamicConfiguration;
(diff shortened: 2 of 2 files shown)


terryquigleysas:
> ## PR Reviewer Guide 🔍
> Here are some key observations to aid the review process:
> 
> 🧪 **PR contains tests**
> 🔒 **Security concerns**
> 
> **Authorization bypass:** The new `withCreateOrUpdateAuditApi()` method allows PUT requests to proceed when `isAuditHotReloadingEnabled()` returns false, which was previously blocked. This creates a potential authorization bypass where audit configuration can be modified even when audit hot reloading is disabled. An attacker or misconfigured client could exploit this to bootstrap or modify audit settings in environments where such changes should be restricted. The code assumes that a missing document always means bootstrapping is safe, but this may not align with the security policy enforced by the `isAuditHotReloadingEnabled()` check.
> 
> ✅ **No TODO sections**
> 🔀 **No multiple PR themes**
> ⚡ **Recommended focus areas for review**
> 
> [**Security Concern**](https://github.com/opensearch-project/security/pull/6158/files#diff-dfab7f54c14aa1838f764895e7a5dfe24268709bd91a13e2c6f5af97092195ebR280-R286)
> Allowing PUT when `isAuditHotReloadingEnabled()` returns false bypasses the intended validation. If the audit config document does not exist, PUT will succeed without verifying that audit hot reloading is actually enabled. This could allow unauthorized or unintended audit configuration changes in environments where audit hot reloading should be disabled. The bootstrap scenario should verify that the system is in a state where bootstrapping is appropriate, not just that the document is absent.
> 
> ```java
> ValidationResult<RestRequest> withCreateOrUpdateAuditApi(final RestRequest request) {
>     if (!securityApiDependencies.configurationRepository().isAuditHotReloadingEnabled()) {
>         // Document does not exist — allow PUT to bootstrap the audit config
>         return ValidationResult.success(request);
>     }
>     return withEnabledAuditApi(request);
> }
> ```

 The isAuditHotReloadingEnabled() check was never an authorization gate. It's a feature availability gate — "does the audit config doc exist in the index?" The actual authentication and
  authorization (credentials, admin privileges) are enforced upstream in the request handler framework via verifyAccessForAllMethods() and the accessHandler predicate. A caller must already
  have security API admin privileges to even reach withCreateOrUpdateAuditApi(). Our fix doesn't bypass any auth — it removes a chicken-and-egg feature gate

terryquigleysas:
> ## PR Code Suggestions ✨
> Explore these optional code suggestions:
> 
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> Possible issue	
> Verify audit config existence logic
> **The logic appears inverted. When `isAuditHotReloadingEnabled()` returns `false`, it typically means audit is disabled or the document doesn't exist. However, the method name suggests it should handle both create (when document doesn't exist) and update (when it does exist) cases. Consider verifying if the condition should check for document existence rather than hot reloading status.**
> 
> [src/main/java/org/opensearch/security/dlic/rest/api/AuditApiAction.java [280-286]](https://github.com/opensearch-project/security/pull/6158/files#diff-dfab7f54c14aa1838f764895e7a5dfe24268709bd91a13e2c6f5af97092195ebR280-R286)
> 
> ```diff
>  ValidationResult<RestRequest> withCreateOrUpdateAuditApi(final RestRequest request) {
>      if (!securityApiDependencies.configurationRepository().isAuditHotReloadingEnabled()) {
>          // Document does not exist — allow PUT to bootstrap the audit config
>          return ValidationResult.success(request);
>      }
> +    // Document exists — validate through normal audit API checks
>      return withEnabledAuditApi(request);
>  }
> ```
> 
> Suggestion importance[1-10]: 3
> __
> 
> Why: The suggestion asks to verify the logic but doesn't identify an actual bug. The test cases in the PR show the logic is intentional: when `isAuditHotReloadingEnabled()` is false (document doesn't exist), PUT is allowed for bootstrapping. The 'improved_code' only adds a comment without changing functionality, making this a low-impact verification request.
> 
> Low

A note on what isAuditHotReloadingEnabled() checks. It returns false specifically when the audit config document doesn't exist (via ConfigurationLoaderSecurity7.noData() → isAuditConfigDocPresentInIndex.set(false)). There is no separate "document existence" check to use — this is the document-existence check. The method name is somewhat misleading, but the semantics appear to be correct.

src/main/java/org/opensearch/security/dlic/rest/api/AuditApiAction.java
```diff
@@ -267,12 +267,24 @@ private void auditApiRequestHandlers(RequestHandler.RequestHandlersBuilder reque
             .onChangeRequest(RestRequest.Method.PATCH, request -> withEnabledAuditApi(request).map(this::processPatchRequest))
             .onChangeRequest(
                 RestRequest.Method.PUT,
-                request -> withEnabledAuditApi(request).map(ignore -> processPutRequest("config", request))
+                request -> withCreateOrUpdateAuditApi(request).map(ignore -> processPutRequest("config", request))
```
cwperks, line 270: Is it worth getting rid of `withEnabledAuditApi` entirely? 

I see that there's this block which [checks for doc existing in `.opendistro_security`](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/dlic/rest/api/AuditApiAction.java#L277-L279). 

I agree that hot reload being disabled should be intentional and not proxied by checking for the auditlog entry in the security index.

terryquigleysas:
@cwperks As discussed in the SIG call, this may require more work. I will need to park it for now and revisit at a later time as I will be out of the office again for an extended period.

DarshitChanpura:
@terryquigleysas Converting this PR to draft until you resume work on this.

PR #6179, opened 2026-06-04
Fix spurious MISSING_PRIVILEGES audits in SSL exception handler

### LINKED ISSUE #6150: [BUG] MISSING_PRIVILEGES audits incorrectly emitted for transient exceptions caught in SecuritySSLRequestHandler
**What is the bug?**

AuditLogSslExceptionHandler.logError at line 71-73 misclassifies any OpenSearchException as a privilege denial:
https://github.com/opensearch-project/security/blob/e408f5cdf52f0ed53de7c73b0230faceaa2af684/src/main/java/org/opensearch/security/auditlog/AuditLogSslExceptionHandler.java#L68-L83

`OpenSearchException` is the base class of essentially every domain exception in the project - `ShardNotFoundException`, `IndexNotFoundException`, `IllegalIndexShardStateException`, `RetentionLeaseNotFoundException`, `AlreadyClosedException`, `TaskCancelledException`, etc. Only a small subset are actual permission denials. The check should narrow to `OpenSearchSecurityException`.

The bug surfaces because the caller - the generic `catch (Exception e)` at `SecuritySSLRequestHandler.java:164` - funnels every exception thrown by the wrapped transport handler through `errorHandler.logError(e, request, action, task, 0)`:

https://github.com/opensearch-project/security/blob/e408f5cdf52f0ed53de7c73b0230faceaa2af684/src/main/java/org/opensearch/security/ssl/transport/SecuritySSLRequestHandler.java#L163-L166

Combined effect: any transient `OpenSearchException` thrown anywhere downstream of the security wrapper produces a phantom `MISSING_PRIVILEGES` audit, even though no privilege check failed and the exception may be silently absorbed.

Steps:

  1. Stand up a cluster with the Security plugin installed and audit logging configured to record MISSING_PRIVILEGES events.
  2. Create an ISM policy with a rollover action on an alias / data stream (any cadence that produces frequent rollovers — e.g., min_doc_count: 1 on a stream with continuous ingest).
  3. As a user with full privileges (all_access), continuously run a search that targets the alias / data stream:
  POST /<alias>/_search
  { "query": { "match_all": {} }, "size": 0 }
  4. Wait for ISM to roll over. The rollover transitions the new write index's shards through not-yet-instantiated → RECOVERING → POST_RECOVERY → STARTED over a window of typically a few hundred milliseconds on
  each data node.
  5. Observe the audit log during the rollover window.

  What you'll see: a burst of MISSING_PRIVILEGES events, all attributed to the search user, all on the new write index, with actions like `indices:data/read/search[can_match][n]` and `indices:data/read/search[phase/query]`. Many will be repeated on the same shard.

`MISSING_PRIVILEGES` should be emitted only when a privilege check actually failed.

 - OS: Linux (any)
 - OpenSearch Version: 2.19.5
 - Plugins: security (this issue), index-management (used to trigger rollover in repro, any rollover trigger would do).
 - Audit logging: MISSING_PRIVILEGES category enabled on transport layer.

N/A. A repesentative stack trace from the data node at the moment `logMissingPrivileges` is invoked:
```
ShardNotFoundException: [my-index-2026.05.15-001623][3] no such shard
      at org.opensearch.index.IndexService.getShard(IndexService.java:533)
      at org.opensearch.search.SearchService.canMatch(SearchService.java:1684)
      at org.opensearch.search.SearchService.canMatch(SearchService.java:1667)
      at org.opensearch.search.SearchService.canMatch(SearchService.java:1655)
      at o.o.action.search.SearchTransportService.lambda$registerRequestHandler$14(SearchTransportService.java:670)
      at o.o.security.ssl.transport.SecuritySSLRequestHandler.messageReceivedDecorate(SecuritySSLRequestHandler.java:207)
      at o.o.security.transport.SecurityRequestHandler.messageReceivedDecorate(SecurityRequestHandler.java:320)
      at o.o.security.ssl.transport.SecuritySSLRequestHandler.messageReceived(SecuritySSLRequestHandler.java:155)
      ... (caught at SecuritySSLRequestHandler.java:164)
      -> AuditLogSslExceptionHandler.logError(...)         
      -> AuditLogImpl.logMissingPrivileges(...)               
      (rethrown; coordinator absorbs via TransportActions.isShardNotAvailableException; user-visible search succeeds)
```

shiv0408:
### Description

Fix spurious `MISSING_PRIVILEGES` audits emitted by `AuditLogSslExceptionHandler.logError(type=0)` for non-security exceptions reaching the generic catch in `SecuritySSLRequestHandler.messageReceived`. Surfaces during ISM rollovers when transient cluster-state exceptions (`ShardNotFoundException`, `IllegalIndexShardStateException`, etc.) bubble up and are misclassified as privilege denials.

* Category: **Bug fix**

The `instanceof OpenSearchException` check is over-broad. `OpenSearchException` is the base class of essentially every domain exception in the project, only a small subset of which represent permission denials. Privilege denials are audited at their [decision point](https://github.com/opensearch-project/security/blob/9e62714db8b8df1918a6cc70e6e0e1f639c01e11/src/main/java/org/opensearch/security/filter/SecurityFilter.java#L402-L414) before the exception is thrown or delivered via `listener.onFailure`. The catch-site emission was at best redundant, at worst spurious.

**Old:** Any `OpenSearchException` reaching the catch triggers a `MISSING_PRIVILEGES` audit, regardless of whether it represents a permission failure. Transient runtime exceptions during normal cluster operations (ISM rollovers, shard recoveries, retention-lease cleanup, etc.) produce phantom audits attributed to whichever user happens to be issuing requests at the time.

**New:** The `MISSING_PRIVILEGES` branch is removed. `OpenSearchException` instances reaching this catch are debug-logged with the action and exception class for diagnostic visibility. The non-`OpenSearchException` branch (SSL handling) is unchanged. Real privilege denials continue to be audited at their decision points.

### Issues Resolved
Resolves #6150

Is this a backport? No.

Do these changes introduce new permission(s) to be displayed in the static dropdown on the front-end? No

### Testing
New `AuditLogSslExceptionHandlerTest.shouldNotEmitMissingPrivilegesForOpenSearchException` asserts `logMissingPrivileges` is never invoked when an `OpenSearchException` reaches `logError(type=0)`. Fails on `main` (Mockito reports the spurious invocation from `AuditLogSslExceptionHandler.java:72`); passes after the fix.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/auditlog/AuditLogSslExceptionHandler.java
+++ b/src/main/java/org/opensearch/security/auditlog/AuditLogSslExceptionHandler.java
@@ -26,6 +26,9 @@
 
 package org.opensearch.security.auditlog;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
 import org.opensearch.OpenSearchException;
 import org.opensearch.security.filter.SecurityRequestChannel;
 import org.opensearch.security.ssl.SslExceptionHandler;
@@ -33,6 +36,8 @@ import org.opensearch.tasks.Task;
 import org.opensearch.transport.TransportRequest;
 
 public class AuditLogSslExceptionHandler implements SslExceptionHandler {
+
+    private static final Logger log = LogManager.getLogger(AuditLogSslExceptionHandler.class);
 
     private final AuditLog auditLog;
 
@@ -69,7 +74,12 @@ public class AuditLogSslExceptionHandler implements SslExceptionHandler {
         switch (type) {
             case 0:
                 if (t instanceof OpenSearchException) {
-                    auditLog.logMissingPrivileges(action, request, task);
+                    log.debug(
+                        "Exception caught at SSL request handler (action={}, class={}); not audited.",
+                        action,
+                        t.getClass().getName(),
+                        t
+                    );
                 } else {
                     auditLog.logSSLException(request, t, action, task);
                 }
--- a/src/test/java/org/opensearch/security/auditlog/AuditLogSslExceptionHandlerTest.java
+++ b/src/test/java/org/opensearch/security/auditlog/AuditLogSslExceptionHandlerTest.java
@@ -1,0 +1,60 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.auditlog;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import org.opensearch.OpenSearchException;
+import org.opensearch.tasks.Task;
+import org.opensearch.transport.TransportRequest;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Tests for {@link AuditLogSslExceptionHandler#logError(Throwable, TransportRequest, String, Task, int)}.
+ *
+ * The {@code type=0} branch funnels exceptions caught by the generic
+ * {@code catch (Exception e)} in {@code SecuritySSLRequestHandler.messageReceived}. Privilege
+ * denials are audited at their decision point (e.g. {@code SecurityFilter.handleUnauthorized})
+ * before the {@link OpenSearchException} is thrown or delivered. The catch-site no longer emits a
+ * {@code MISSING_PRIVILEGES} audit; it debug-logs for diagnostic visibility instead.
+ */
+public class AuditLogSslExceptionHandlerTest {
+
+    private AuditLog auditLog;
+    private AuditLogSslExceptionHandler handler;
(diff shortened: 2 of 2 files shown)


src/main/java/org/opensearch/security/auditlog/AuditLogSslExceptionHandler.java
```diff
@@ -69,7 +74,12 @@ public void logError(Throwable t, TransportRequest request, String action, Task
         switch (type) {
             case 0:
                 if (t instanceof OpenSearchException) {
-                    auditLog.logMissingPrivileges(action, request, task);
```
cwperks, line 72: Thank you @shiv0408 , this totally makes sense to remove and not quite sure why its originally here when its audit logged elsewhere at the decision point.

May be out of scope for this PR, but I would love to invert this block to check specifically for superclass of ssl exception and otherwise do a debug log.

cwperks (approved):

cwperks:
Other high-level comment, is that we may have an existing test suite for this new test to be added. Not too much overhead to have a dedicated suite either way.

shiv0408:
Thanks for your approval @cwperks. 
I can open an issue to invert this catch block  to check for `SSLConfigException` (I believe, correct me if I'm wrong). 
Also, I could not find any suite test for this that's why created a new block.

shiv0408:
@cwperks All CI checks are passing now. I don't see an option to merge this PR, can you help with that? Thanks!

RyanL1997 (approved):

PR #6224, opened 2026-06-17
Enforce 256 character limit on all text inputs

DarshitChanpura:
Add global MAX_STRING_LENGTH (256) enforcement:
- Add validateStringLength to validation chain for all PUT/PATCH requests
- Recursively checks all string values in request body JSON
- Add entity name length check in EndpointValidator.withRequiredEntityName()
- Add STRING_EXCEEDS_MAX_LENGTH validation error

### Check List


RyanL1997 (approved):

rursprung:
@DarshitChanpura:
i just tried upgrading from 3.7.0 to 3.8.0 and noticed that we can't `PUT` one of our `role`s anymore. that role happens to have a fairly large DLS query.
suddenly we got an HTTP 400 with "String value exceeds the maximum allowed length of 256 characters"

=> since this is not configurable this is a breaking change. also there's no workaround if you need a longer DLS query => i think this needs to be rolled back

thanks @reta for tracing the error back to this PR

i've raised #6434 for this

PR #6244, opened 2026-06-23
support query-based terms lookup queries in DLS

### LINKED ISSUE #6243: [FEATURE] support TLQ with `query` in DLS
since 3.2.0 TLQs can have a `query` instead of an `id`, this was introduced in https://github.com/opensearch-project/OpenSearch/pull/18195
this is currently unsupported in DLS, if you try to use it you get an NPE.

support `query` based TLQ in DLS

n/a

original support for TLQ in DLS was introduced in https://github.com/opensearch-project/security/issues/1508

rursprung:
### Description
OpenSearch introduced the possibility to use a `query` rather than just
an `id` in Terms Lookup Queries in https://github.com/opensearch-project/OpenSearch/pull/18195 in release 3.2.0.
However, support for this was never added to the security plugin for
DLS/FLS.

When trying to use `query` based TLQ in a DLS this failed for two
reasons:
- the visible failure is an NPE because `DlsFilterLevelActionHandler`
  unconditionally dereferenced `termsLookup().id()`, which returns
  `null` for query-based lookups.
- additionally, the `DocumentAllowList` privilege bypass only handled
  `GetRequest`, but query-based TLQ internally resolves via
  `SearchRequest` and additionally uses a `GetSettingsRequest`, causing
  the privileges evaluator to block the lookup even after the NPE is
  fixed.

This commit adds proper support for `query` based TLQ in DLS, both for
the old and the new evaluator, by allowlisting a wildcard entry for the
whole index in question rather than a single document in case `query` is
used. additionally, `SearchRequest` and `GetSettingsRequest` are now
also supported in `DocumentAllowList#isAllowed`.

Furthermore, the legacy
`PrivilegesEvaluatorImpl#checkDocAllowListHeader` now delegates to
`DocumentAllowList#isAllowed` since the old method body was a
hand-inlined copy of the same logic (parse header, check request);
delegating removes the duplication and picks up `SearchRequest` support
without repeating the new code.

Header parsing in `DocumentAllowList#isAllowed` is now delegated to
`DocumentAllowList#get` to centralise error handling.

The wildcard allowlist pattern is already established for Dashboards
multi-tenancy and is scoped to the request's thread context, so it
cannot grant persistent or write access beyond the TLQ resolution.
Existing code with the wildcard has been migrated to use the new
constant to make it easier to find it.

additionally, `DlsTermLookupQueryV4EvaluatorTest` has been added (in a separate commit) to run the tests from `DlsTermLookupQueryTest` also against the new evaluator (this unearthed that `GetSettingsRequest` had to be whitelisted as well; the old evaluator didn't block this!).

### Issues Resolved
fixes #6243

Do these changes introduce new permission(s) to be displayed in the static dropdown on the front-end?
no

### Testing
unit & integration testing + manual testing

### Check List
- [ ] New functionality has been documented - shouldn't be needed since everyone will just expect this to work?

### DIFF
--- a/src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
+++ b/src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
@@ -517,8 +517,14 @@ public class DlsFilterLevelActionHandler {
 
                 for (QueryBuilder queryBuilder : queryBuilders) {
                     TermsQueryBuilder termsQueryBuilder = (TermsQueryBuilder) queryBuilder;
+                    final var lookupIndex = termsQueryBuilder.termsLookup().index();
+                    final var lookupId = termsQueryBuilder.termsLookup().id();
 
-                    documentAllowlist.add(termsQueryBuilder.termsLookup().index(), termsQueryBuilder.termsLookup().id());
+                    if (lookupId != null) {
+                        documentAllowlist.add(lookupIndex, lookupId);
+                    } else {
+                        documentAllowlist.add(lookupIndex, DocumentAllowList.ANY_DOCUMENT_ID);
+                    }
                 }
             }
 
--- a/src/main/java/org/opensearch/security/privileges/DocumentAllowList.java
+++ b/src/main/java/org/opensearch/security/privileges/DocumentAllowList.java
@@ -18,7 +18,9 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import org.opensearch.action.ActionRequest;
+import org.opensearch.action.admin.indices.settings.get.GetSettingsRequest;
 import org.opensearch.action.get.GetRequest;
+import org.opensearch.action.search.SearchRequest;
 import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.security.support.ConfigConstants;
 
@@ -31,6 +33,8 @@ import org.opensearch.security.support.ConfigConstants;
 public class DocumentAllowList {
 
     private static final Logger log = LogManager.getLogger(DocumentAllowList.class);
+
+    public static final String ANY_DOCUMENT_ID = "*";
 
     public static DocumentAllowList get(ThreadContext threadContext) {
         String header = threadContext.getHeader(ConfigConstants.OPENDISTRO_SECURITY_DOC_ALLOWLIST_HEADER);
@@ -48,34 +52,51 @@ public class DocumentAllowList {
     }
 
     public static boolean isAllowed(ActionRequest request, ThreadContext threadContext) {
-        String docAllowListHeader = threadContext.getHeader(ConfigConstants.OPENDISTRO_SECURITY_DOC_ALLOWLIST_HEADER);
+        final var documentAllowList = DocumentAllowList.get(threadContext);
 
-        if (docAllowListHeader == null) {
+        if (documentAllowList.isEmpty()) {
             return false;
         }
 
-        if (!(request instanceof GetRequest)) {
-            return false;
-        }
-
-        try {
-            DocumentAllowList documentAllowList = DocumentAllowList.parse(docAllowListHeader);
-            GetRequest getRequest = (GetRequest) request;
-
+        // GetRequest: id-based TLQ resolves via GET; match exact (index, id) entry.
+        // SearchRequest: query-based TLQ resolves via SEARCH; match wildcard entry.
+        // GetSettingsRequest: query-based TLQ retrieves the index setting (during the fetch phase).
+        // Other request types (including writes) are never allowlisted.
+        if (request instanceof GetRequest getRequest) {
             if (documentAllowList.isAllowed(getRequest.index(), getRequest.id())) {
-                if (log.isDebugEnabled()) {
-                    log.debug("Request " + request + " is allowed by " + documentAllowList);
-                }
-
+                log.debug("Request {} is allowed by {}", request, documentAllowList);
                 return true;
-            } else {
+            }
+            return false;
+        } else if (request instanceof SearchRequest searchRequest) {
+            if (isIndicesAllowlisted(documentAllowList, searchRequest.indices())) {
+                log.debug("Request {} is allowed by {}", request, documentAllowList);
+                return true;
+            }
+            return false;
+        } else if (request instanceof GetSettingsRequest getSettingsRequest) {
+            if (isIndicesAllowlisted(documentAllowList, getSettingsRequest.indices())) {
+                log.debug("Request {} is allowed by {}", request, documentAllowList);
+                return true;
+            }
+            return false;
+        }
+
+        return false;
+    }
+
+    // allMatch semantics: at privilege-evaluation time, every index in the request must be
+    // covered. DlsFlsValveImpl uses anyMatch+size()==1 for a different purpose (per-shard bypass).
+    private static boolean isIndicesAllowlisted(DocumentAllowList documentAllowList, String[] indices) {
+        if (indices == null || indices.length == 0) {
+            return false;
+        }
+        for (String index : indices) {
+            if (index == null || !documentAllowList.isAllowed(index, ANY_DOCUMENT_ID)) {
                 return false;
             }
-
-        } catch (Exception e) {
-            log.error("Error while handling document allow list: " + docAllowListHeader, e);
-            return false;
         }
+        return true;
     }
 
     private static final DocumentAllowList EMPTY = new DocumentAllowList();
--- a/src/main/java/org/opensearch/security/privileges/actionlevel/legacy/PrivilegesEvaluatorImpl.java
+++ b/src/main/java/org/opensearch/security/privileges/actionlevel/legacy/PrivilegesEvaluatorImpl.java
@@ -62,7 +62,6 @@ import org.opensearch.action.bulk.BulkItemRequest;
 import org.opensearch.action.bulk.BulkRequest;
 import org.opensearch.action.bulk.BulkShardRequest;
 import org.opensearch.action.delete.DeleteAction;
-import org.opensearch.action.get.GetRequest;
 import org.opensearch.action.get.MultiGetAction;
 import org.opensearch.action.index.IndexAction;
 import org.opensearch.action.search.MultiSearchAction;
@@ -699,31 +698,7 @@ public class PrivilegesEvaluatorImpl implements PrivilegesEvaluator {
     }
 
     private boolean checkDocAllowListHeader(User user, String action, ActionRequest request) {
-        String docAllowListHeader = threadContext.getHeader(ConfigConstants.OPENDISTRO_SECURITY_DOC_ALLOWLIST_HEADER);
-
-        if (docAllowListHeader == null) {
-            return false;
-        }
-
-        if (!(request instanceof GetRequest)) {
-            return false;
-        }
-
-        try {
-            DocumentAllowList documentAllowList = DocumentAllowList.parse(docAllowListHeader);
-            GetRequest getRequest = (GetRequest) request;
-
-            if (documentAllowList.isAllowed(getRequest.index(), getRequest.id())) {
-                log.debug("Request {} is allowed by {}", request, documentAllowList);
-                return true;
-            } else {
-                return false;
-            }
-
-        } catch (Exception e) {
-            log.error("Error while handling document allow list: {}", docAllowListHeader, e);
-            return false;
-        }
+        return DocumentAllowList.isAllowed(request, threadContext);
     }
 
     private List<String> toString(List<AliasMetadata> aliases) {
--- a/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
+++ b/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java
@@ -211,7 +211,7 @@ public class DlsFlsValveImpl implements DlsFlsRequestValve {
             && resolvedIndices.local()
                 .namesOfIndices(context.clusterState())
                 .stream()
-                .anyMatch(index -> documentAllowList.isAllowed(index, "*"))) {
+                .anyMatch(index -> documentAllowList.isAllowed(index, DocumentAllowList.ANY_DOCUMENT_ID))) {
             // The documentAllowList is needed here for Dashboards multi tenancy which can redirect index accesses to indices for which no
             // normal index privileges are present
             // If we would not use the documentAllowList here, the index would appear to be protected
@@ -481,7 +481,7 @@ public class DlsFlsValveImpl implements DlsFlsRequestValve {
                 // - DLS rules which use "term lookup queries" and thus need to access indices for which no privileges are present
                 // - Dashboards multi tenancy which can redirect index accesses to indices for which no normal index privileges are present
 
-                if (!dlsRestriction.isUnrestricted() && documentAllowList.isAllowed(index, "*")) {
+                if (!dlsRestriction.isUnrestricted() && documentAllowList.isAllowed(index, DocumentAllowList.ANY_DOCUMENT_ID)) {
                     dlsRestriction = DlsRestriction.NONE;
                     log.debug("Lifting DLS for {} due to present document allowlist", index);
                 }
--- a/src/main/java/org/opensearch/security/configuration/SecurityFlsDlsIndexSearcherWrapper.java
+++ b/src/main/java/org/opensearch/security/configuration/SecurityFlsDlsIndexSearcherWrapper.java
@@ -163,7 +163,7 @@ public class SecurityFlsDlsIndexSearcherWrapper extends SystemIndexSearcherWrapp
                 // - DLS rules which use "term lookup queries" and thus need to access indices for which no privileges are present
                 // - Dashboards multi tenancy which can redirect index accesses to indices for which no normal index privileges are present
 
-                if (!dlsRestriction.isUnrestricted() && documentAllowList.isAllowed(index.getName(), "*")) {
+                if (!dlsRestriction.isUnrestricted() && documentAllowList.isAllowed(index.getName(), DocumentAllowList.ANY_DOCUMENT_ID)) {
                     dlsRestriction = DlsRestriction.NONE;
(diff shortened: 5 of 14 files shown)


src/main/java/org/opensearch/security/privileges/actionlevel/legacy/PrivilegesEvaluatorImpl.java
rursprung, line 122: FYI: the presence of this is why the legacy evaluator allowed `GetSettingsRequest`

rursprung:
the CI failure is a false reject (can't fetch maven snapshots). i've seen this happen a lot lately!

rursprung:
CC @srikanthpadakanti (you implemented `query` support in TLQ in OpenSearch) and @jochenkressin & @nibix (you implemented/contributed the original TLQ-in-DLS support)

rursprung:
from what i can see all CI failures are false rejects: they either failed due to not being able to fetch maven artefacts (i've seen that a lot now on PRs on this repo - what's going on here?) - or completely unrelated tests failing => probably nothing that i can help sort out here?

src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
```diff
@@ -517,8 +517,14 @@ private boolean modifyQuery(String localClusterAlias) throws IOException {
...
                 for (QueryBuilder queryBuilder : queryBuilders) {
                     TermsQueryBuilder termsQueryBuilder = (TermsQueryBuilder) queryBuilder;
+                    final var lookupIndex = termsQueryBuilder.termsLookup().index();
+                    final var lookupId = termsQueryBuilder.termsLookup().id();
 
-                    documentAllowlist.add(termsQueryBuilder.termsLookup().index(), termsQueryBuilder.termsLookup().id());
+                    if (lookupId != null) {
+                        documentAllowlist.add(lookupIndex, lookupId);
+                    } else {
+                        documentAllowlist.add(lookupIndex, DocumentAllowList.WILDCARD_DOCUMENT_ID);
```
reta: I think usage of `*` (DocumentAllowList.WILDCARD_DOCUMENT_ID) is widely permissive, shouldn't we limit documents to `termsQueryBuilder.termsLookup().query()` instead?
```suggestion
            dlsQueryBuilder.should(QueryBuilders.boolQuery()
                .must(QueryBuilders.termQuery("_index", lookupIndex))
                .must(termsQueryBuilder.termsLookup().query()));
```

src/main/java/org/opensearch/security/privileges/DocumentAllowList.java
```diff
@@ -32,6 +34,8 @@ public class DocumentAllowList {
 
     private static final Logger log = LogManager.getLogger(DocumentAllowList.class);
 
+    public static final String WILDCARD_DOCUMENT_ID = "*";
```
reta: May be replace wildcard with any:
```suggestion
    public static final String ANY_DOCUMENT_ID = "*";
```

src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
rursprung: i don't see how appending the TLQ subquery to the `dlsQueryBuilder` as a `should` clause would help? in my understanding this `dlsQueryBuilder` constructs the query which goes against the main index to apply the DLS and it uses `should` since the way DLS works with multiple roles is that the most permissive role wins (otherwise it'd have to use `must` for the queries).
the subquery can't go against this index since that's a different index (as visible by your example using a term query on `_index = lookupIndex` vs. `_index = prefixedIndex` in the other places).

we don't need more / other restrictions on the main index - we need access to the documents on the referenced index of the subquery. and that one is allowed to be read during the DLS evaluation.

(i did nevertheless give your example a quick spin and the tests fail as expected => this doesn't work; i currently don't see another way besides allowlisting, but i might not be aware of better alternatives available since i'm not super familiar with this code base)

nibix: The DocumentAllowList is indeed necessary as normally a user for which TLQ DLS rules apply does not have access privileges for the index with the lookup data. 

Still, I am a bit concerned that this gives such users a vector to read the full TLQ lookup index, i.e. also entries which do not belong to them. I don't have a concrete exploit for that ready, but I could imagine that this could work via a second user controlled TLQ query.

This could be solved by reasoning that the information in the lookup index is not sensitive and can be safely read by anyone. If so, this should be however clearly stated in the docs.

On the other hand, however, one could then also reason that users which are subject to TLQ DLS rules, could be also just given completely normal read privileges to the lookup index.

nibix:
@rursprung 

Thank you for this and sorry for the late response; at the moment, I have quite a few different things going on.

src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
rursprung: isn't it a major design flaw if a user could influence the DLS query? the DLS query can only access user attributes which are _usually_ not under the control of a user / attacker - and if they are the attacker can just give themselves a role which grants them admin access anyway.

and if the user cannot influence the DLS query then they cannot get more information about the content of the index accessed by the TLQ because the allowlisting is only valid during the DLS execution, where the DLS query is considered "safe".

(caveat: in #6265 i plan to add user attributes which technically can be influenced by the user - but these have to be used with care and will be documented as such)

nibix: > and if the user cannot influence the DLS query then they cannot get more information about the content of the index accessed by the TLQ because the allowlisting is only valid during the DLS execution, where the DLS query is considered "safe".

At the moment, so called "filter level" DLS works by combining user query and DLS query to one large restricted query. This query needs to have access to the index accessed by the TLQ query inside the DLS query. However, it is (AFAIK) only possible to give the whole query access to the index which provides TLQ information. In the current implementation, this is limited to a single document identified by a user attribute inside the DLS query. However, the change of the DocumentAllowlist in this PR would give complete index access.

This could possibly be fixed by actually executing two distinct queries, first the DLS/TLQ query in a privileged mode, retrieve results and then rewrite the original user query with the results from the DLS/TLQ query.

nibix: Alternatively, one could also think about extending DocumentAllowlist to accept queries and apply it like a lucene DLS rule when it is active.

reta: > Alternatively, one could also think about extending DocumentAllowlist to accept queries and apply it like a lucene DLS rule when it is active.

+1, I think that would be best option, I (mistakenly) assumed that `dlsQueryBuilder` query would take care of it

rursprung: i'm not quite sure that i follow: what query should `DocumentAllowlist` accept in this case? if i add the whole subquery as a DLS for the index targeted by the sub-query that'd just mean that query would contain it twice, right? (because then the subquery gets executed and wrapped in a DLS query with the same query again)
the important part here would be that the _index_ gets whitelisted - which is the same as i'm doing now with the wildcard.

rursprung:
if we can't reach an understanding of how the whitelisting for the index targeted by the subquery should work before we hit the code freeze for 3.8.0 would it also be ok if i'd split this in two: this PR would only add support for running query-based TLQs in DLS and the authorization part would be a new issue. then it'd already be possible to use query-based TLQ in DLS as long as the user is granted read access to the index targeted by the subquery (a workable workaround which could be documented as such).

src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
reta: >i'm not quite sure that i follow: what query should DocumentAllowlist accept in this case? 

I think disconnect that I (and @nibix seems like) have is:
 - termsLookup needs special treatment 
 - the id are extracted and passed to `DocumentAllowlist`

Now, if the termsLookup need special treatment for id, does it need the same for query? @cwperks may be you could help us here? 

Thanks!

rursprung: the rule with TLQ in DLS is that the user running the original query doesn't need to have explicit access granted to the index referenced in the TLQ in the DLS. instead, the system automatically whitelists access to that for the DLS execution.

so far, the only TLQ which existed was with a specific ID - accordingly, the whitelisting was only done for that one documented in the referenced index.

since 3.2.0 it is now possible to have a TLQ which contains a subquery - accordingly, there now isn't a single ID which we can whitelist, instead it needs access to whatever documents match that subquery - so essentially it needs access to all documents in the referenced index so that it can evaluate that subquery.

reta: > so essentially it needs access to all documents in the referenced index so that it can evaluate that subquery.

thanks, so the id match **is** a query by itself (just not freestyle like `query()` but by id), so why do we need this `DocumentAllowlist` at all since those would be filtered out anyway? (at least, this is my understanding) this asymmetry in treating id and query is not clear (if we need index whitelist, just `lookupIndex` is enough)

rursprung: > thanks, so the id match **is** a query by itself

i don't think so - it's a read of a single field of a single document in a referenced index.

> so why do we need this `DocumentAllowlist` at all since those would be filtered out anyway

what do you mean with "those would be filtered out anyway"? `DocumentAllowlist ensures that this read on that referenced document is allowed even though the user usually doesn't have access to that index / document.

> this asymmetry in treating id and query is not clear (if we need index whitelist, just `lookupIndex` is enough)

so far it was enough to whitelist a single document because it was clear which document was referenced. with the query all documents in the referenced index need to be accessed by the query so that it can filter them. that's the big difference.
and sure, we could now say that we just whitelist the whole index also for an `id` based TLQ - that'd work technically. but it'd widen the granted accesses unnecessarily. when designing this from the ground up we might do this, but since we already have this per-document whitelisting, why throw it out of the window?

reta: > what do you mean with "those would be filtered out anyway"? `DocumentAllowlist ensures that this read on that referenced document is allowed even though the user usually doesn't have access to that index / document.

Aha, this statements clears things out to me, thanks. So we just have no choice (in case of query), now I understand, adding query support to `DocumentAllowlist` would not be light option

src/main/java/org/opensearch/security/privileges/DocumentAllowList.java
```diff
@@ -48,34 +52,51 @@ public static DocumentAllowList get(ThreadContext threadContext) {
...
+                return true;
+            }
+            return false;
+        } else if (request instanceof SearchRequest searchRequest) {
+            if (isIndicesAllowlisted(documentAllowList, searchRequest.indices())) {
+                log.debug("Request {} is allowed by {}", request, documentAllowList);
+                return true;
+            }
+            return false;
+        } else if (request instanceof GetSettingsRequest getSettingsRequest) {
```
reta, line 77: This part looks very suspicious: 
 - we guard `GetSettingsRequest` behind `indices:monitor/settings/get` (if I am not mistaken), so with this check we will bypass that?
 - also, the index settings have nothing to do with documents, do we have a gap somewhere else in the evaluation chain?

reta: (just saw your comment https://github.com/opensearch-project/security/pull/6244/changes#r3467807886)

rursprung:
the CI failure seems to be an infrastructure issue. i've [reported it on slack](https://opensearch.slack.com/archives/C04UTNM338A/p1783673582951099).
the push was just to rebase the PR.

src/main/java/org/opensearch/security/privileges/DocumentAllowList.java
rursprung: so with my comment this is clear to you now?
query-based TLQ does fetch the index settings (i forgot again why it's doing that), so we need to whitelist that.

reta: > so with my comment this is clear to you now?

Yes, it is, so termsLookup.query() did introduce get settings call (see please https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/TermsQueryBuilder.java#L600), so we have no choice here as to let is pass :(

src/main/java/org/opensearch/security/privileges/DocumentAllowList.java
```diff
@@ -48,34 +52,51 @@ public static DocumentAllowList get(ThreadContext threadContext) {
...
 
-        if (docAllowListHeader == null) {
+        if (documentAllowList.isEmpty()) {
             return false;
         }
 
-        if (!(request instanceof GetRequest)) {
+        // GetRequest: id-based TLQ resolves via GET; match exact (index, id) entry.
+        // SearchRequest: query-based TLQ resolves via SEARCH; match wildcard entry.
+        // GetSettingsRequest: OpenSearch reads index settings before executing query-based TLQ search.
```
reta:
```suggestion
        // GetSettingsRequest: query-based TLQ retrieves the index setting (during the fetch phase) 
```

reta:
> if we can't reach an understanding of how the whitelisting for the index targeted by the subquery should work before

I think I cleared everything out for myself, just to summarize for others, we do have:
  - the terms lookup by id -> just GetRequest, covered by document allow list with (index, id) pairs
  - the terms lookup by query -> settings + search requests, both need index wide permissions (since the ids are not available), covered by document allow list with (index, *) pairs

Thanks @rursprung for patience, I just have one minor comment to update the comments section, attributing settings call, LGTM otherwise!

rursprung:
thanks @reta! i've incorporated your feedback & rebased the PR.

is there a need to raise a docs request for this at all? the docs already state that TLQ is supported in DLS and didn't state anywhere that `query` based TLQ wasn't supported (because this seems to have a been an oversight and not intentional in the first place) => i wouldn't know what to add to the docs beyond "this works too"?

reta:
> is there a need to raise a docs request for this at all? the docs already state that TLQ is supported in DLS and didn't state anywhere that `query` based TLQ wasn't supported (because this seems to have a been an oversight and not intentional in the first place) => i wouldn't know what to add to the docs beyond "this works too"?

Thanks @rursprung , I think no need for doc changes, this is indeed an oversight (== bugfix), thanks!

reta (approved):

rursprung:
@reta: is there anything missing to get this merged before the code freeze? CI is happy and i see an approval on the PR (i'm not a maintainer, so i don't have the rights to hit merge)

reta:
> @reta: is there anything missing to get this merged before the code freeze? CI is happy and i see an approval on the PR (i'm not a maintainer, so i don't have the rights to hit merge)

Not from my side, @cwperks @nibix anything from your side? Thanks!

cwperks:
@reta I don't have any objections, but I do think our coverage of FGAC for TLQs is lacking in general. I think this change looks sensible to me, and separately I would also like someone to take on adding FGAC focused tests for TLQs since you can "join" with another index. I certainly hope the subquery is being executed within the same context as the top level search.

cwperks:
@rursprung additionally I would like to see coverage in scenarios where the TLQ is done on an index where the user has DLS/FLS restrictions. I haven't dug deep on this, but I don't think this repo has a lot of coverage in this feature that was added in 3.2.

rursprung:
> @rursprung additionally I would like to see coverage in scenarios where the TLQ is done on an index where the user has DLS/FLS restrictions. I haven't dug deep on this, but I don't think this repo has a lot of coverage in this feature that was added in 3.2.

@cwperks but that's in `DlsTermLookupQueryTest` - or what are you looking for? that's a whole class just testing TLQ in DLS. and i've extended it with tests for `query` based TLQ in DLS in this PR (+ added coverage for the V4 evaluator since that was missing so far).

cwperks:
> @cwperks but that's in DlsTermLookupQueryTest - or what are you looking for? that's a whole class just testing TLQ in DLS. and i've extended it with tests for query based TLQ in DLS in this PR (+ added coverage for the V4 evaluator since that was missing so far).

Ah, my bad. I'm glad we have tests in place for a number of scenarios already. At a glance, this seemed like an area that could have issues given that it can link to another index to query and the user running the query on the primary index may have restrictions on the second index. In that event, I wanted to make sure the restrictions applied, either if it is getting by id or doing a search to find terms to replace in the primary query.

rursprung:
ah, my bad. now i see what you're looking for: the TLQ on another index which is itself protected by DLS. yes, i think that's not currently covered by tests.
but the way i understood it from @nibix (who was involved in the original TLQ-in-DLS implementation) the whole point of this is that the restrictions on the referenced index fall away for the DLS.

IIUC there is one caveat here (and i wasn't aware of this for a long time and i don't think that it's clearly documented?): the DLS is executed as a wrapper around the main query, i.e. this becomes one big query in the end. thus the restriction is also lifted for the wrapped query. though i haven't tried that yet.

since i'm off now i won't be able to update this PR until i get back. i leave it up to you if you want to merge this as-is or wait for additional test coverage.
from what i see:
* without this `query` based TLQ-in-DLS isn't possible, so by merging this we're not changing existing (working) DLS queries and anyone starting to use it opts into it explicitly
* the allowlisting behaviour is essentially the same as for existing `id` based TLQ-in-DLS, so again, no surprises there (beyond the fact that maybe already that behaviour is surprising?)

cwperks:
@rursprung I merged this PR since its up to the cluster admin to write the DLS clause on each role, but I do think we should expand the coverage in this area. FYI DLS is adaptive by default. Take a look at https://github.com/opensearch-project/security/blob/28cc929ebed96e2dac4d5d289bcdddd27dc937a0/src/main/java/org/opensearch/security/configuration/DlsFlsValveImpl.java#L247-L269 which controls the flow of what adaptive means. DLS can be at the filter level (modify the query) or applied on the Lucene level.

PR #6253, opened 2026-06-27
Log default compliance salt warning once at startup

### LINKED ISSUE #6105: [BUG] warning about default FLS salt on startup
**What is the bug?**
every time the cluster starts the following warning is printed to the log:
```json
{"type": "server", "timestamp": "2026-04-22T13:53:45,361+0000", "level": "WARN", "component": "o.o.s.c.Salt", "cluster.name": "test-opensearch", "node.name": "test-opensearch-0", "message": "If you plan to use field masking pls configure compliance salt e1ukloTsQlOgPquJ to be a random string of 16 chars length identical on all nodes" }
{"type": "server", "timestamp": "2026-04-22T13:53:45,361+0000", "level": "WARN", "component": "o.o.s.c.Salt", "cluster.name": "test-opensearch", "node.name": "test-opensearch-0", "message": "If you plan to use field masking pls configure compliance salt e1ukloTsQlOgPquJ to be a random string of 16 chars length identical on all nodes" }
{"type": "server", "timestamp": "2026-04-22T13:53:45,588+0000", "level": "WARN", "component": "o.o.s.c.Salt", "cluster.name": "test-opensearch", "node.name": "test-opensearch-0", "message": "If you plan to use field masking pls configure compliance salt e1ukloTsQlOgPquJ to be a random string of 16 chars length identical on all nodes" }
```

note that the salt given is the default salt (though this is not obvious from the message):
https://github.com/opensearch-project/security/blob/48934186b6dcf3705b58a0d217745d298b484821/src/main/java/org/opensearch/security/support/ConfigConstants.java#L323

the warning comes from here:
https://github.com/opensearch-project/security/blob/48934186b6dcf3705b58a0d217745d298b484821/src/main/java/org/opensearch/security/configuration/Salt.java#L47-L52

this happens, because we do not define the salt as we're not using FLS and thus have no need for it.

start a fresh cluster where the salt for FLS is not explicitly defined.

if FLS is not used there should be no need to define a salt for it and thus there should be no warning for it

 - OS: linux
 - Version OpenSearch 3.6.0
 - Plugins security

n/a

the cluster still works, the warning can be ignored.

MdTanwer:
## Summary

- Logs the default compliance salt warning **once** at plugin startup via `Salt.warnIfDefaultComplianceSalt()`, instead of on every `Salt` construction (which produced duplicate WARN lines on startup).
- Keeps the warning **unconditional** when the default salt is in use, per maintainer feedback — even if no roles define `masked_fields` at bootstrap.
- Removes the warning from the `Salt` constructor so repeated `Salt.from()` calls stay silent.
- Leaves 4.0 bootstrap enforcement (`Salt.validateSaltSettings()`) as a TODO, gated on `plugins.security.allow_unsafe_democertificates`.

Fixes #6105 (duplicate/noisy startup warnings; does not suppress the warning entirely when FLS is unused).

## Test plan

- [x] `./gradlew test --tests "org.opensearch.security.configuration.SaltTest"`
- [x] `./gradlew spotlessJavaCheck`
- [ ] Start a cluster without `plugins.security.compliance.salt` — single WARN at startup (not repeated)
- [ ] Configure a custom compliance salt — no WARN at startup


cwperks:
@MdTanwer IMO This warning should always be logged if this salt is unconfigured even if any roles at bootstrap don't have masking. In fact, in the next major version I'd like to fail bootstrap if its unconfigured unless `plugins.security.allow_unsafe_democertificates` is set to true. This is the setting that explicitly allows a cluster to run with demo security settings.

MdTanwer:
@cwperks Thanks for the feedback — I've revised the PR accordingly.


- The default compliance salt warning is now logged **unconditionally** at plugin startup (via `Salt.warnIfDefaultComplianceSalt()`), even when no roles define `masked_fields`.
- The warning was removed from the `Salt` constructor to fix the duplicate 3× WARNING.

This keeps the always-warn behavior . Please let me know if this looks good.

cwperks:
@MdTanwer can you please fix the code hygiene failing check?

PR #6254, opened 2026-06-27
Grant indices:admin/get to kibana_server role

### LINKED ISSUE #6103: [BUG] Missing indices:admin/get permission in kibana_server static role
**What is the bug?**
The `kibana_server` static role should provide all permissions required for OpenSearch Dashboards. However, a user with that role receives:

```json
{"type":"log","@timestamp":"2026-04-21T19:31:41Z","tags":["error","opensearch","data"],"pid":23005,"message":"[security_exception]: no permissions for [indices:admin/get] and User [name=opensearch-dashboards, backend_roles=[], requestedTenant=null]"}
```

Mapping the user to `all_access` works fine:
```json
{"type":"log","@timestamp":"2026-04-21T19:31:43Z","tags":["info","savedobjects-service"],"pid":23005,"message":"Creating index .kibana_1."}
{"type":"log","@timestamp":"2026-04-21T19:31:44Z","tags":["info","savedobjects-service"],"pid":23005,"message":"Pointing alias .kibana to .kibana_1."}
{"type":"log","@timestamp":"2026-04-21T19:31:44Z","tags":["info","savedobjects-service"],"pid":23005,"message":"Finished in 550ms."}
```

1. Use OpenSearch with security plugin and OpenSearch Dashboards 3.6.0.0
2. Map user `opensearch-dashboards` to role `kibana_server`
3. Start Dashboards with that user

OpenSearch Dashboard works with `kibana_server` role permissions.

 - OS: Arch Linux

The named permission is indeed missing from the static role: https://github.com/opensearch-project/security/blob/3.6.0.0/src/main/resources/static_config/static_roles.yml#L79

MdTanwer:
## Summary

- Add `indices:admin/get` to the `kibana_server` static role catch-all (`*`) index pattern so OpenSearch Dashboards can probe `.kibana` before saved-objects bootstrap on a fresh cluster.

Fixes #6103

### DIFF
--- a/src/main/resources/static_config/static_roles.yml
+++ b/src/main/resources/static_config/static_roles.yml
@@ -116,6 +116,7 @@ kibana_server:
         - "*"
       allowed_actions:
         - "indices:admin/aliases*"
+        - "indices:admin/get"
 
 logstash:
   reserved: true


cwperks (approved):
Approving this pr, but I'd like to do much deeper work in the future with dashboards. Most of the oddities around perms are because OpenSearch Dashboards is different from the rest of the OpenSearch ecosystem of plugins. Dashboards treats OpenSearch as merely a metadata store and does not expose any backend APIs which leads to these permissions listing out the dashboards indices. This is a read only perm so there's no danger of mutation, but it does now increase the scope of what kibana_server can do.

PR #6271, opened 2026-06-30
feat(audit): Add unified disabled_categories setting (#6222)

### LINKED ISSUE #6222: Add unified `disabled_categories` audit log setting and deprecate REST/transport split
### Is your feature request related to a problem?

The current audit log configuration requires users to manage two separate settings for disabling categories:

- `disabled_rest_categories`
- `disabled_transport_categories`

Most users don't understand or care about the REST vs transport layer distinction. The REST layer is the HTTP entry point, while the transport layer is the internal action execution layer. Some categories only fire at one layer (e.g., `GRANTED_PRIVILEGES`, `MISSING_PRIVILEGES`, `CLUSTER_SETTINGS_CHANGED`, `INDEX_SETTINGS_CHANGED` are transport-only), making it confusing to know which setting to use.

This complexity leads to misconfiguration—for example, a user might add `CLUSTER_SETTINGS_CHANGED` to `disabled_rest_categories` expecting it to be suppressed, but it has no effect because that category only fires on the transport layer.

### Proposed solution

1. Add a new unified `disabled_categories` setting that applies to both REST and transport layers
2. Mark `disabled_rest_categories` and `disabled_transport_categories` as deprecated
3. Precedence: if `disabled_categories` is present, it takes precedence over the split settings. If absent, fall back to the existing `disabled_rest_categories` / `disabled_transport_categories` behavior
4. Emit a deprecation warning when the old settings are detected

Example config with the new setting:

```yml
config:
  audit:
    disabled_categories:
      - AUTHENTICATED
      - GRANTED_PRIVILEGES
```

### Additional context

For advanced use cases where users genuinely need per-layer control (e.g., logging `AUTHENTICATED` on REST for source IPs but suppressing it on transport to reduce noise from node-to-node traffic), the deprecated settings would still function during the transition period. A future enhancement could add an override mechanism if demand exists.

Related: the new `CLUSTER_SETTINGS_CHANGED` and `INDEX_SETTINGS_CHANGED` categories added in #6113 are transport-only, further highlighting that the split is confusing for end users.

Taiwo435:
Adds a single 'disabled_categories' setting that applies to both REST and transport layers, replacing the confusing split configuration of 'disabled_rest_categories' and 'disabled_transport_categories'.

Behavior:
- When 'disabled_categories' is set, it takes precedence and applies to both layers at runtime
- When not set, falls back to the deprecated split settings with a deprecation warning encouraging migration
- Old settings continue to work for backwards compatibility
- REST API validation updated to accept the new field

The REST/transport category split was an internal implementation detail invisible to users and caused silent misconfigurations.

Resolves #6222

### Description

### Issues Resolved

### Testing

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -1736,7 +1736,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
             settings.add(
                 Setting.listSetting(
                     ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_CATEGORIES,
-                    disabledCategories,
+                    Collections.emptyList(),
                     Function.identity(),
                     Property.NodeScope
                 )
@@ -1809,7 +1809,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
                     case DISABLE_CATEGORIES:
                         return Setting.listSetting(
                             filterEntry.getKeyWithNamespace(),
-                            ConfigConstants.OPENDISTRO_SECURITY_AUDIT_DISABLED_CATEGORIES_DEFAULT,
+                            Collections.emptyList(),
                             Function.identity(),
                             Property.NodeScope
                         );
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -1735,14 +1735,6 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
             disabledCategories.add("GRANTED_PRIVILEGES");
             settings.add(
                 Setting.listSetting(
-                    ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_CATEGORIES,
-                    Collections.emptyList(),
-                    Function.identity(),
-                    Property.NodeScope
-                )
-            );
-            settings.add(
-                Setting.listSetting(
                     ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_TRANSPORT_CATEGORIES,
                     disabledCategories,
                     Function.identity(),
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -1735,6 +1735,14 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
             disabledCategories.add("GRANTED_PRIVILEGES");
             settings.add(
                 Setting.listSetting(
+                    ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_CATEGORIES,
+                    disabledCategories,
+                    Function.identity(),
+                    Property.NodeScope
+                )
+            );
+            settings.add(
+                Setting.listSetting(
                     ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_TRANSPORT_CATEGORIES,
                     disabledCategories,
                     Function.identity(),
@@ -1798,6 +1806,13 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
 
             Arrays.stream(FilterEntries.values()).map(filterEntry -> {
                 switch (filterEntry) {
+                    case DISABLE_CATEGORIES:
+                        return Setting.listSetting(
+                            filterEntry.getKeyWithNamespace(),
+                            ConfigConstants.OPENDISTRO_SECURITY_AUDIT_DISABLED_CATEGORIES_DEFAULT,
+                            Function.identity(),
+                            Property.NodeScope
+                        );
                     case DISABLE_REST_CATEGORIES:
                     case DISABLE_TRANSPORT_CATEGORIES:
                         return Setting.listSetting(
--- a/src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
+++ b/src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
@@ -130,11 +130,7 @@ public class AuditConfig {
      */
     @JsonInclude(JsonInclude.Include.NON_NULL)
     public static class Filter {
-        private static Set<String> FIELDS = new HashSet<>(DefaultObjectMapper.getFields(Filter.class)) {
-            {
-                add("disabled_categories");
-            }
-        };
+        private static Set<String> FIELDS = DefaultObjectMapper.getFields(Filter.class);
         @VisibleForTesting
         public static final Filter DEFAULT = Filter.from(Settings.EMPTY);
 
@@ -156,10 +152,7 @@ public class AuditConfig {
         private final WildcardMatcher ignoredAuditRequestsMatcher;
         private final WildcardMatcher ignoredCustomHeadersMatcher;
         private WildcardMatcher ignoredUrlParamsMatcher;
-        @com.fasterxml.jackson.annotation.JsonIgnore
-        private final boolean disabledCategoriesConfigured;
         @JsonProperty("disabled_categories")
-        @JsonInclude(JsonInclude.Include.NON_EMPTY)
         private final Set<AuditCategory> disabledCategories;
         @Deprecated
         private final Set<AuditCategory> disabledRestCategories;
@@ -180,7 +173,6 @@ public class AuditConfig {
             final Set<String> ignoredUrlParams,
             final Set<AuditCategory> disabledRestCategories,
             final Set<AuditCategory> disabledTransportCategories,
-            final boolean disabledCategoriesConfigured,
             final Set<AuditCategory> disabledCategories
         ) {
             this.isRestApiAuditEnabled = isRestApiAuditEnabled;
@@ -199,7 +191,6 @@ public class AuditConfig {
             this.ignoredUrlParamsMatcher = WildcardMatcher.from(ignoredUrlParams);
             this.disabledRestCategories = disabledRestCategories;
             this.disabledTransportCategories = disabledTransportCategories;
-            this.disabledCategoriesConfigured = disabledCategoriesConfigured;
             this.disabledCategories = disabledCategories;
         }
 
@@ -261,16 +252,13 @@ public class AuditConfig {
             final boolean logRequestBody = getOrDefault(properties, FilterEntries.LOG_REQUEST_BODY.getKey(), true);
             final boolean resolveIndices = getOrDefault(properties, FilterEntries.RESOLVE_INDICES.getKey(), true);
             final boolean excludeSensitiveHeaders = getOrDefault(properties, FilterEntries.EXCLUDE_SENSITIVE_HEADERS.getKey(), true);
-            final boolean disabledCategoriesConfigured = properties.containsKey(FilterEntries.DISABLE_CATEGORIES.getKey());
-            final Set<AuditCategory> disabledCategories = disabledCategoriesConfigured
-                ? AuditCategory.parse(
-                    getOrDefault(
-                        properties,
-                        FilterEntries.DISABLE_CATEGORIES.getKey(),
-                        ConfigConstants.OPENDISTRO_SECURITY_AUDIT_DISABLED_CATEGORIES_DEFAULT
-                    )
+            final Set<AuditCategory> disabledCategories = AuditCategory.parse(
+                getOrDefault(
+                    properties,
+                    FilterEntries.DISABLE_CATEGORIES.getKey(),
+                    Collections.emptyList()
                 )
-                : null;
+            );
             final Set<AuditCategory> disabledRestCategories = AuditCategory.parse(
                 getOrDefault(
                     properties,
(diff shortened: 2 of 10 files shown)


itsmevichu:
@Taiwo435 

Thanks for contributing! Really appreciate you taking the time to work on this.

Just a small note - I had opened a PR for the same change a few days earlier. It might be worth checking the open PRs before starting on a change, as it helps avoid duplicate work and ensures the effort someone has already put into an existing PR doesn't go to waste.

Would you mind taking a look at my PR? If they're essentially addressing the same thing, it might make sense to close this one and continue the discussion there. Of course, I'll leave it to you and the maintainers to decide what's best.

Thanks again!

cwperks:
@itsmevichu apologies, but this PR is related to an internship experience and will be the one I will review. I apologize for not assigning the other issue so that you knew it was taken.

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
```diff
@@ -129,7 +130,11 @@ public static AuditConfig from(final Settings settings) {
      */
     @JsonInclude(JsonInclude.Include.NON_NULL)
     public static class Filter {
-        private static Set<String> FIELDS = DefaultObjectMapper.getFields(Filter.class);
+        private static Set<String> FIELDS = new HashSet<>(DefaultObjectMapper.getFields(Filter.class)) {
```
cwperks: Is this change necessary? Why is it needed for this field and not others?

src/main/java/org/opensearch/security/dlic/rest/api/AuditApiAction.java
```diff
@@ -149,6 +149,19 @@ public class AuditApiAction extends AbstractApiAction {
     private final List<String> readonlyFields;
 
     public static class AuditRequestContentValidator extends RequestContentValidator {
+        public static final Set<AuditCategory> DISABLED_CATEGORIES = Set.of(
```
cwperks: Since this list is now composed of all, I think we may be able to do something like `AuditCategory.valueOf()` to see if the value supplied via REST API is a valid enum value.

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -212,10 +212,18 @@ public class ConfigConstants {
     public static final String OPENDISTRO_SECURITY_AUDIT_RESOLVE_INDICES = "opendistro_security.audit.resolve_indices";
     public static final String OPENDISTRO_SECURITY_AUDIT_ENABLE_REST = "opendistro_security.audit.enable_rest";
     public static final String OPENDISTRO_SECURITY_AUDIT_ENABLE_TRANSPORT = "opendistro_security.audit.enable_transport";
+    public static final String OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_CATEGORIES =
+        "opendistro_security.audit.config.disabled_categories";
     public static final String OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_TRANSPORT_CATEGORIES =
         "opendistro_security.audit.config.disabled_transport_categories";
     public static final String OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_REST_CATEGORIES =
         "opendistro_security.audit.config.disabled_rest_categories";
+    public static final List<String> OPENDISTRO_SECURITY_AUDIT_DISABLED_CATEGORIES_DEFAULT = ImmutableList.of(
```
cwperks: IMO we should leave a comment to revisit these defaults for 4.0. I think its odd that we are disabling categories by default and should instead look to create a sensible configuration for the demo security configuration while clearly documenting to users how to configure for their specific requirements.

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -1222,6 +1232,16 @@ boolean checkRestFilter(final AuditCategory category, final String effectiveUser
             return false;
         }
 
+        if (auditConfigFilter.isDisabledCategoriesConfigured()) {
```
cwperks: @Taiwo435 I think we should make the default for `getDisabledCategories` an empty list and then we don't need this `isDisabledCategoriesConfigured`. wdyt? I believe its only here since we set the default value for disabled_categories to a non-empty list which looks like its yanked from elsewhere.

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
Taiwo435: DefaultObjectMapper.getFields(Filter.class) uses Jackson's ClassIntrospector to reflectively discover all fields that would be serialized for the Filter class. It returns an ImmutableSet of field names and this set
 is used as an allowlist when validating incoming REST API requests so any key not in FIELDS gets rejected as invalid. The disabled_categories field is annotated with @JsonInclude(JsonInclude.Include.NON_EMPTY), meaning Jackson only serializes it when non-null/non-empty. Because of this (conditional serialization), the
 introspector may not consistently include it in the discovered field set. If it's missing from FIELDS, users would get their API requests rejected when trying to configure disabled_categories for the first time. So wrapping the result in new HashSet<>(...) because the original return type is an ImmutableSet, and then manually add "disabled_categories" to guarantee it's always recognized as a valid input
 field regardless of its current serialization state. For the other fields, they're unconditionally serialized and always picked up by reflection.

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
Taiwo435: if getDisabledCategories() always returned an empty set (instead of distinguishing between "not configured" and "configured as empty"), we'd lose the ability to tell states 1 and states 3 apart:

State 1: Not present in config at all
  
 User's audit config — no disabled_categories key anywhere
  audit:
    disabled_rest_categories: ["AUTHENTICATED"]
    disabled_transport_categories: ["FAILED_LOGIN"]
  
  This user hasn't migrated yet. They expect their split settings to keep working. We need to detect this field was never set and fall back to the split behavior. isDisabledCategoriesConfigured = false.
  
  State 2: Present and populated
  
  audit:
    disabled_categories: ["AUTHENTICATED", "GRANTED_PRIVILEGES"]
  
  User has migrated. Unified takes precedence. isDisabledCategoriesConfigured = true, set contains two categories.
  
  State 3: Present but empty
  
  audit:
    disabled_categories: []
  
 User explicitly says "I want everything audited, disable nothing." isDisabledCategoriesConfigured = true, set is empty.

We'd have to pick one interpretation for empty:
  
 - If empty means "not configured" → State 3 users can't disable nothing. Their empty list gets ignored and split settings kick in (or defaults). Their intent is lost.
 - If empty means "configured" → State 1 users (who never set the field) suddenly have unified taking precedence with an empty list, meaning nothing is disabled. Their split settings are silently ignored.
  Categories they thought were suppressed start flooding their logs.

However, we make disabledCategories nullable. null means "not configured" (fall back to split), any Set (even empty) means "configured. A null check already exists so we could drop the boolean and just rely on the null/non-null distinction directly. It's a clarity choice

src/main/java/org/opensearch/security/support/ConfigConstants.java
Taiwo435: Noted

src/main/java/org/opensearch/security/dlic/rest/api/AuditApiAction.java
Taiwo435: We could remove the DISABLED_CATEGORIES set and its containsAll check entirely. Validation already happens during deserialization. DefaultObjectMapper.readTree(jsonContent, AuditConfig.class) triggers
 AuditCategory.parse(), which calls AuditCategory.valueOf() on each category string. Invalid values throw IllegalArgumentException before the containsAll checks ever execute. The same applies to
  DISABLED_REST_CATEGORIES and DISABLED_TRANSPORT_CATEGORIES.

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
cwperks: > User has migrated. Unified takes precedence. isDisabledCategoriesConfigured = true, set contains two categories.

Is there really precedence though? I would think for simplicity we treat this independently from the existing settings - i.e. they can work in tandem - but if both `disable_categories` and either/both `disable_rest_categories` or `disable_transport_categories` are configured that we'd log a warning message, like the deprecation message in this PR. 

> However, we make disabledCategories nullable

This is also a good solution IMO.

Taiwo435: Good point, treating them independently is simpler. No precedence logic, no ambiguity about which setting wins. They just work in tandem, and we log a warning if both are configured to encourage migration. I'll rework it.

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
```diff
@@ -315,6 +341,20 @@ public static Filter from(Settings settings) {
             final Set<String> ignoredAuditUsers = fromSettingStringSet(settings, FilterEntries.IGNORE_USERS, DEFAULT_IGNORED_USERS);
             final Set<String> ignoreAuditRequests = fromSettingStringSet(settings, FilterEntries.IGNORE_REQUESTS, Collections.emptyList());
             final Set<String> ignoreHeaders = fromSettingStringSet(settings, FilterEntries.IGNORE_HEADERS, Collections.emptyList());
+
+            if (settings.hasValue(FilterEntries.DISABLE_CATEGORIES.getKeyWithNamespace())
+                && (settings.hasValue(FilterEntries.DISABLE_REST_CATEGORIES.getKeyWithNamespace())
+                    || settings.hasValue(FilterEntries.DISABLE_REST_CATEGORIES.getLegacyKeyWithNamespace())
+                    || settings.hasValue(FilterEntries.DISABLE_TRANSPORT_CATEGORIES.getKeyWithNamespace())
+                    || settings.hasValue(FilterEntries.DISABLE_TRANSPORT_CATEGORIES.getLegacyKeyWithNamespace()))) {
+                final DeprecationLogger deprecationLogger = DeprecationLogger.getLogger(AuditConfig.class);
```
cwperks: Can we do this log from a single place in this class?

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -212,6 +212,8 @@ public class ConfigConstants {
     public static final String OPENDISTRO_SECURITY_AUDIT_RESOLVE_INDICES = "opendistro_security.audit.resolve_indices";
     public static final String OPENDISTRO_SECURITY_AUDIT_ENABLE_REST = "opendistro_security.audit.enable_rest";
     public static final String OPENDISTRO_SECURITY_AUDIT_ENABLE_TRANSPORT = "opendistro_security.audit.enable_transport";
+    public static final String OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_CATEGORIES =
+        "opendistro_security.audit.config.disabled_categories";
```
cwperks: Let's change this to `plugins.security.audit.config.disabled_categories`. For the other one's in here we will want to also support `plugins.security...` while allowing use of legacy, but leaving deprecation messages to change the values in `opensearch.yml` to use the `plugins.security...` version of these settings. No need to update all the settings in this PR, but for this new setting let's use `plugins.security...` out of the gate.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1733,6 +1733,14 @@ public List<Setting<?>> getSettings() {
             final List<String> disabledCategories = new ArrayList<String>(2);
             disabledCategories.add("AUTHENTICATED");
             disabledCategories.add("GRANTED_PRIVILEGES");
+            settings.add(
+                Setting.listSetting(
+                    ConfigConstants.OPENDISTRO_SECURITY_AUDIT_CONFIG_DISABLED_CATEGORIES,
+                    Collections.emptyList(),
+                    Function.identity(),
+                    Property.NodeScope
```
cwperks: Not needed in this first PR, but we should make this dynamic as discussed earlier today. When adding Property.Dynamic as a setting property we also need to create an update consumer to consume changes to the setting.

cwperks (commented):
Thank you @Taiwo435, a few more minor comments but overall this PR looks good!

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
Taiwo435: Yes, I can extract it into a single private helper method that both from() methods call, so the logic isn't duplicated.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
Taiwo435: Noted. I've already implemented this on my standalone branch (dynamic settings with Property.Dynamic + addSettingsUpdateConsumer for most of the filter settings). I'll bring it in as a follow-up PR.

src/main/java/org/opensearch/security/support/ConfigConstants.java
Taiwo435: Noted.

cwperks (approved):

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -212,6 +212,7 @@ public class ConfigConstants {
     public static final String OPENDISTRO_SECURITY_AUDIT_RESOLVE_INDICES = "opendistro_security.audit.resolve_indices";
     public static final String OPENDISTRO_SECURITY_AUDIT_ENABLE_REST = "opendistro_security.audit.enable_rest";
     public static final String OPENDISTRO_SECURITY_AUDIT_ENABLE_TRANSPORT = "opendistro_security.audit.enable_transport";
+    public static final String SECURITY_AUDIT_CONFIG_DISABLED_CATEGORIES = "plugins.security.audit.config.disabled_categories";
```
DarshitChanpura, line 215: we'll have to add a documentation PR for this new setting:
https://github.com/opensearch-project/documentation-website

DarshitChanpura (approved):
Looks good let's follow up with doc PR to update the documentation website.

https://github.com/opensearch-project/documentation-website

src/main/java/org/opensearch/security/support/ConfigConstants.java
DarshitChanpura: resolving this to unblock the PR.

Issue #6015, opened 2026-03-16
[FEATURE] Add `plugins.security.dashboards.dashboards_url` as dynamic cluster setting


cwperks:
**Is your feature request related to a problem?**

Submitting this feature request to add a cluster setting for `kibana_url` as a duplicate to `config.dynamic.authc.saml_auth_domain.config.kibana_url`. Note that only a single SAML authenticator can be configured because it requires `challenge` to be set to true. 

I think it makes more sense to add this as a cluster setting and to implement some level of https://github.com/opensearch-project/security/issues/5219.

As for authz of https://github.com/opensearch-project/security/issues/5219, can we intercept cluster settings APIs check if any `plugins.security....` is being updated and enforce that the caller is a cluster security admin?

**What solution would you like?**

Add a dynamic cluster setting for this current securityconfig setting.

**What alternatives have you considered?**

Status quo

Rishav9852Kumar:
Hi @cwperks  I had a few questions I wanted to align on:

 -  Sensitive setting? / should we consider this setting-update as sensitive settings and be redacted in audit logs?
 - Precedence when both are configured / If someone has the static kibana_url in securityconfig and the new dynamic cluster setting, which should take precedence (my choice :- Dynamic cluster setting over Static config)

cwperks:
> Sensitive setting? / should we consider this setting-update as sensitive settings and be redacted in audit logs?

Sensitive, yes. There's actually a setting prop called sensitive that we should apply to this. Sensitive in this case though means only updateable by a cluster admin it does not need to be omitted from the audit log as this would be known to any user of OSD.

> Precedence

Prefer the one set in cluster settings to the one in the index and ensure we have log messages about this as well.

praveenvenkat06:
Hi @cwperks. I am interested to work on this.

PR #6275, opened 2026-07-01
Add support for dynamic Dashboards URL via cluster settings

praveenvenkat06:
### Description

**Enhancement** — Introduces a new dynamic cluster setting `plugins.security.dashboards.dashboards_url` that allows operators to update the OpenSearch Dashboards URL used in the SAML authentication flow at runtime, without modifying the security index or reloading configuration.

**Old behavior:** The Dashboards URL (`kibana_url`) was read once from the SAML authenticator's security-index configuration and required a config reload to change.

**New behavior:** The URL is resolved dynamically at every call site using the following precedence:
1. Dynamic cluster setting (`plugins.security.dashboards.dashboards_url`) — if set and non-empty
2. Fallback to `kibana_url` in the security configuration

The SAML settings cache (`Saml2SettingsProvider`) is automatically invalidated when the setting value changes, so updates take effect immediately with no restart required.

### Issues Resolved
[[6015](https://github.com/opensearch-project/security/issues/6015)]

### Testing
Manual testing is done and dynamic Dashboards URL changes are reflected at runtime. Unit tests have been added for the implementation.

### Check List
- [ ] New functionality has been documented - Will be done once the PR gets approved
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md) - Will be done once the PR gets approved


src/main/java/org/opensearch/security/auth/http/saml/AuthTokenProcessorHandler.java
```diff
@@ -394,7 +404,8 @@ private String getAbsoluteAcsEndpoint(String acsEndpoint) {
             if (acsEndpointUri.isAbsolute()) {
                 return acsEndpoint;
             } else {
-                return new URI(this.kibanaRootUrl).resolve(acsEndpointUri).toString();
+                String dashboardsUrl = dashboardsUrlSupplier.get();
```
Rishav9852Kumar, line 407: If both the dynamic cluster setting is cleared and kibana_url was never configured, this returns null. new URI(null) on the next line throws NullPointerException, which isn't caught by the URISyntaxException handler — resulting in an unhandled crash.

 Consider adding a null/empty guard:
```
if (dashboardsUrl == null || dashboardsUrl.isEmpty()) {
      log.error("Dashboards URL is not configured; cannot resolve relative acsEndpoint: {}", acsEndpoint);
      return acsEndpoint;
  }
```

praveenvenkat06: @Rishav9852Kumar I have added the null guard.

praveenvenkat06:
Hi @cwperks, can this be merged?

Issue #4031, opened 2024-02-08
[BUG] Field masking has inconsistent memory issues with certain queries


peternied:
**What is the bug?**
We've seen reports of heap memory usage spiking with field masking enabled.  With how masking is implemented at the leaf level its possible that certain types of queries cause the masked fields to be materialized even when they are not used.

**How can one reproduce the bug?**
- Checkout this branch https://github.com/opensearch-project/security/compare/main...peternied:masking-perf
- Run `./gradlew integrationTest --tests org.opensearch.security.MaskingTests`
- Analyze the results

## Baseline behavior

Baseline query with the following format:
```java 
final SearchSourceBuilder ssb = new SearchSourceBuilder();
ssb.size(0);
final SearchRequest request = new SearchRequest(INDEX_NAME_PREFIX + "*");
request.source(searchSourceBuilder);
```

### Creating 3 Indices with 5000 Documents

| Role    | Condition                   | Count | Attempts | Avg Heap Used | Max Heap Used | Min Heap Used | Std Deviation |
|---------|-----------------------------|-------|----------|---------------|---------------|---------------|---------------|
| admin   |                             | 100   | 106      | 888,814       | 3,621,248     | 191,136       | 480,198       |
| reader  |                             | 100   | 105      | 808,317       | 1,470,368     | 517,640       | 195,083       |
| reader  | ROLE_WITH_NO_MASKING        | 100   | 105      | 813,600       | 1,602,272     | 536,416       | 206,571       |
| reader  | MASKING_LOW_REPEAT_VALUE    | 100   | 105      | 936,948       | 1,631,088     | 618,704       | 195,653       |
| reader  | MASKING_RANDOM_LONG         | 100   | 105      | 919,187       | 1,593,456     | 548,264       | 236,892       |
| reader  | MASKING_RANDOM_STRING       | 100   | 105      | 951,826       | 1,656,432     | 564,784       | 191,008       |

### Creating 3 Indices with 50000 Documents

| Role    | Condition                   | Count | Attempts | Avg Heap Used | Max Heap Used | Min Heap Used | Std Deviation |
|---------|-----------------------------|-------|----------|---------------|---------------|---------------|---------------|
| admin   |                             | 100   | 110      | 873,379       | 4,872,552     | 575,208       | 508,483       |
| reader  |                             | 100   | 109      | 753,489       | 4,020,832     | 567,304       | 341,413       |
[...]

PR #6276, opened 2026-07-03
Solution (#4031): [BUG] Field masking has inconsistent memory issues with certain q

TFGSUMIT:
This pull request addresses the issue of inconsistent memory issues with field masking in certain queries. The changes include:

* Modifying the `QueryRewriter` class to rewrite queries to exclude masked fields when field masking is enabled.
* Adding an `isFieldMasked` method to the `Masking` class to check if a field is masked.
* Updating the `MaskingTests` class to test rewritten queries and verify that they do not materialize masked fields.

To test this PR, please follow these instructions:

1. Run the `MaskingTests` class to verify that the rewritten queries do not materialize masked fields.
2. Use a memory profiler to analyze the heap usage of the rewritten queries.
3. Verify that the rewritten queries do not cause memory issues.

Note: This PR assumes that the `Masking` class is properly configured to enable field masking. Please ensure that the `Masking` class is correctly configured before applying this PR.

### DIFF
--- a/h
+++ b/h
@@ -1,0 +1,1 @@
+mvn test -Dtest=org.opensearch.security.MaskingTests
--- a/org/opensearch/security/Masking.java
+++ b/org/opensearch/security/Masking.java
@@ -1,0 +1,19 @@
+import java.util.Set;
+
+public class Masking {
+
+    public static boolean isEnabled() {
+        // Check if field masking is enabled in the configuration
+        return getConfiguration().isEnabled();
+    }
+
+    public static boolean isFieldMasked(String field) {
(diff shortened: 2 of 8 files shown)


DarshitChanpura (commented):
hi @TFGSUMIT thank you for raising this PR.

The files added here seem to be incomplete and not referenced anywhere. Some files are already present in core engine (e.g. SearchRequest), and can be used from there. Could you file an issue describing the bug and repro if one doesn't exist already.

After these are addressed, can you:
- run "./gradlew spotlessApply"
- signoff commits using GPG sign

PR #6285, opened 2026-07-08
fix: emit FAILED_LOGIN audit event when Basic Auth rejects credentials in multi-auth (SAML + Basic) setup

### LINKED ISSUE #6221: [BUG] FAILED_LOGIN audit event missing for failed Basic Auth in multi-auth setup (SAML + Basic)
**What is the bug?**
  
  When using the [recommended multi-auth configuration](https://docs.opensearch.org/latest/security/authentication-backends/saml/#running-multiple-authentication-domains) (Basic Auth `challenge: false` + SAML `challenge: true`), failed login attempts via the OpenSearch Dashboards login form do not produce a `FAILED_LOGIN` audit event.
  
  Dashboards correctly shows "Invalid username or password", but the security audit trail has no record of the failed attempt. This makes brute-force detection via audit logs impossible in this configuration.
  
  1. Configure multi-auth as documented in [Running multiple authentication domains](https://docs.opensearch.org/latest/security/authentication-backends/saml/#running-multiple-authentication-domains):
  
  ```yaml
  authc:
    basic_internal_auth_domain:
      order: 1
      http_authenticator:
        type: basic
        challenge: false
      authentication_backend:
        type: internal
    saml_auth_domain:
      order: 3
      http_authenticator:
        type: saml
        challenge: true
      authentication_backend:
        type: noop
  ```
   Dashboards config:
  ```yaml
  opensearch_security.auth.type: ["saml", "basicauth"]
  opensearch_security.auth.multiple_auth_enabled: true
  ```
  2. Enable audit logging with FAILED_LOGIN category
  3. Open Dashboards login form and enter invalid credentials
  4. Observe: Dashboards shows "Invalid username or password" — but no FAILED_LOGIN event appears in the audit log
  
  What is the expected behavior?
  
  A FAILED_LOGIN audit event should be recorded with the rejected username when Basic Auth credentials are rejected.
  
  What is your host/environment?
  
  - OS: Linux (Kubernetes/EKS)
  - Version: Affected since 2.18.0
  - Plugins: OpenSearch Security Plugin
  
  Do you have any screenshots?
  
  N/A — the issue is the absence of an audit log entry.
  
  Do you have any additional context?

  This appears to be introduced by PR #4762 (fix for #4608). That PR suppressed false-positive FAILED_LOGIN events during normal SAML redirects, but it appears to also suppress the event in the case where Basic Auth has already rejected real credentials in a
  previous auth domain iteration.

mvanhorn:
### Description
* Category: Bug fix
* Why these changes are required? In the documented multi-auth configuration (Basic Auth with `challenge: false` at order 1, SAML with `challenge: true` later), a failed Basic Auth login through the Dashboards login form produced no FAILED_LOGIN audit event, so brute-force attempts were invisible to audit logs. The suppression added by #4762 (to silence false positives during normal SAML redirects) also fires when Basic Auth has already rejected real credentials, because the loop overwrites the previously extracted credentials with null in the SAML iteration and then returns from the suppressed branch without logging.
* What is the old behavior before changes and new behavior after changes? Before: rejected Basic Auth credentials in a multi-auth (SAML + Basic) setup logged nothing. After: `BackendRegistry.authenticate` preserves the last extracted credentials across auth-domain iterations (`authCredentials` is only overwritten when the new extraction is non-null), and the #4762 suppression is narrowed to apply only when no earlier domain rejected real credentials. A genuine failed Basic login now emits FAILED_LOGIN with the rejected username; an unauthenticated browser hitting the SAML redirect still logs nothing, preserving the #4608 fix.

### Issues Resolved
Fixes #6221

Is this a backport? No.

Do these changes introduce new permission(s) to be displayed in the static dropdown on the front-end? No.

### Testing
New `BackendRegistryTest` follows the existing `BackendRegistryGrpcAuthTest` pattern (mocked `AdminDNs`/`XFFResolver`/`AuditLog`/`ThreadPool`/`ClusterInfoHolder`, auth domains injected via a mocked `DynamicConfigModel`). Covers: rejected Basic credentials followed by a SAML challenge emits FAILED_LOGIN with the rejected username; a bare SAML redirect with no extracted credentials stays silent (the #4608 behavior); single-domain Basic rejection still logs as before.

### Check List
- [x] New functionality has been documented - N/A, no user-facing config or API change; behavior matches the documented audit expectations
- [x] New Roles/Permissions have a corresponding security dashboards plugin PR - N/A, no new roles/permissions
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md) - N/A, no API change

### DIFF
--- a/src/main/java/org/opensearch/security/auth/BackendRegistry.java
+++ b/src/main/java/org/opensearch/security/auth/BackendRegistry.java
@@ -372,7 +372,9 @@ public class BackendRegistry {
             2. If auth domain is challenging and no credentials are found -> present challenge.
             SAML and basic auth for example redirect/re-request credentials from clients.
              */
-            authCredentials = ac;
+            if (ac != null) {
+                authCredentials = ac;
+            }
             if (ac == null) {
                 // no credentials found in request
                 if (!gRPC && anonymousAuthEnabled && isRequestForAnonymousLogin(request.params(), request.getHeaders())) {
@@ -382,11 +384,17 @@ public class BackendRegistry {
                 if (authDomain.isChallenge()) {
                     final Optional<SecurityResponse> restResponse = httpAuthenticator.reRequestAuthentication(request, null);
                     if (restResponse.isPresent()) {
+                        final String authenticatorType = authDomain.getHttpAuthenticator().getType();
                         // saml will always hit this to re-request authentication
-                        if (!authDomain.getHttpAuthenticator().getType().equals(SAML_TYPE)) {
-                            auditLog.logFailedLogin("<NONE>", false, null, request);
+                        if (!authenticatorType.equals(SAML_TYPE) || authCredentials != null) {
+                            auditLog.logFailedLogin(
+                                authCredentials == null ? "<NONE>" : authCredentials.getUsername(),
+                                false,
+                                null,
+                                request
+                            );
                         }
-                        if (authDomain.getHttpAuthenticator().getType().equals(BASIC_TYPE)) {
+                        if (authenticatorType.equals(BASIC_TYPE)) {
                             log.warn("No 'Authorization' header, send 401 and 'WWW-Authenticate Basic'");
                         }
                         notifyIpAuthFailureListeners(request, authCredentials);
--- a/src/main/java/org/opensearch/security/auth/BackendRegistry.java
+++ b/src/main/java/org/opensearch/security/auth/BackendRegistry.java
@@ -371,6 +371,11 @@ public class BackendRegistry {
             Anonymous users are handled later outside of auth domain loop if no other user is authenticated.
             2. If auth domain is challenging and no credentials are found -> present challenge.
             SAML and basic auth for example redirect/re-request credentials from clients.
+
+            Retain credentials extracted by an earlier auth domain. In a SAML and Basic chain, Basic can reject the
+            credentials before HTTPSamlAuthenticator.extractCredentials() returns null because the request contains Basic
+            credentials. SAML still requires challenge: true in the current reRequestAuthentication flow so it can issue its
+            redirect; retaining the rejected credentials lets that challenge path audit the actual username.
              */
             if (ac != null) {
                 authCredentials = ac;
--- a/src/test/java/org/opensearch/security/auth/BackendRegistryTest.java
+++ b/src/test/java/org/opensearch/security/auth/BackendRegistryTest.java
@@ -1,0 +1,279 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.auth;
+
+import java.net.InetSocketAddress;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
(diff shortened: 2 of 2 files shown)


cwperks:
@mvanhorn Can you run `./gradlew spotlessApply` a commit the changes?

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -372,7 +372,9 @@ Authenticate users injected in the thread context by internal components (plugin
             2. If auth domain is challenging and no credentials are found -> present challenge.
             SAML and basic auth for example redirect/re-request credentials from clients.
              */
-            authCredentials = ac;
+            if (ac != null) {
```
cwperks, line 380: I see why this is being done since `extractCredentials` from `HTTPSamlAuthenticator` will return null when doing basic auth, but it does feel hacky. I think this really illustrates that the `reRequestAuthentication` flow is not structured well. Because of the design, we require `challenge: true` for authenticators like the SAML authenticator when `challenge` really only makes sense for HTTP Basic auth.

Can we add a comment in here explaining the rationale?

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -382,11 +384,17 @@ Authenticate users injected in the thread context by internal components (plugin
...
+                                false,
+                                null,
+                                request
+                            );
                         }
-                        if (authDomain.getHttpAuthenticator().getType().equals(BASIC_TYPE)) {
+                        if (authenticatorType.equals(BASIC_TYPE)) {
                             log.warn("No 'Authorization' header, send 401 and 'WWW-Authenticate Basic'");
                         }
                         notifyIpAuthFailureListeners(request, authCredentials);
```
DarshitChanpura, line 405: since authCredentials now retains the rejected username across iterations, notifyIpAuthFailureListeners call here will also receive the actual username instead of null when firing from the SAML challenge path which may end up improving failed login tracking by attributing user (if available) to an IP.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
DarshitChanpura: +1.

DarshitChanpura (commented):
thank you @mvanhorn for this improvment!

DarshitChanpura:
@mvanhorn Are you actively working on this PR? If not, it will be closed as stale during next cleanup.

mvanhorn:
@DarshitChanpura yes, still working on it. Sorry for the gap.

@cwperks added the rationale comment you asked for in ec4bc9a, at the credential-retention site. I kept it to describing why the current flow forces this rather than claiming to have restructured `reRequestAuthentication`; I agree with you that the underlying design is awkward, but that felt like a separate change from this one.

I also went through the three red checks, since only one of them is mine:

- **Spotless scan** was ours: an import-order violation in `BackendRegistryTest.java`. Fixed in the same commit. `./gradlew spotlessCheck` now exits 0 locally.
- **code-ql** is failing on `Loaded a configuration file for version '4.36.3', but running version '4.36.2'`, a CodeQL Action tooling mismatch in the runner rather than a finding against this code.
- **integration-tests (25, windows-latest)** dies on `BindTransportException: Failed to bind to [::1]:47300; Address already in use`. The SSL handshake errors after it are downstream of the failed bind. It is the only failing leg in the matrix, and CI on `main` is red on five of its last six runs, so I do not think either of these two is attributable to this PR.

`./gradlew test --tests "*BackendRegistryTest*"` passes locally. @DarshitChanpura on your line 400 note, agreed, that is the intended effect: retaining the rejected username means `notifyIpAuthFailureListeners` can attribute the failure to a user on the SAML challenge path instead of firing with null.

cwperks (approved):

Issue #6268, opened 2026-06-29
[FEATURE]  preserving original client identity in certificate based authentication  where a load balancer does initial SSL handshake


asifabashar:
**Is your feature request related to a problem?**
Where there a load balancer that itself does mTLS authentication, it forwards the request to opensearch where client is load balancer and actual client cert is no longer performing mtls , it is the load balancer that is doing cleint authentication with opensearch cluster. 

**What solution would you like?**
To preserve identity of the original client, we can have first opensearch do client authentication validation with load balancer certificate through TLS handshake and then load balancer can store the original client certificate in header as URL encoded value in header field x-client-cert. The load balancer will connect to opensearch cluster and forward the request with original client cert in header so that opensearch cluster is assume correct identity of the original client. The load balancer has to be trusted by the opensearch so that i can trust its header so load balancer certificate will be validated at every request by opensearch using its standard certficate based authentication.
It will be controlled by flag settings where by default this feature will be disabled.

**What alternatives have you considered?**

**Do you have any additional context?**
Many environments use load balancers in the infrastructure and kubernetes  deployments also use load balancers also in general in most cases.

asifabashar:
Please review , I have a change ready and will create pull request if you approve  the requirements above as valid.

asifabashar:
not a duplicate, its trying handle a different use case and applies for opensearch client auth, not only dashboard.

cwperks:
@asifabashar what do your solution look like? I'd be happy to review a PR? I believe we already have support for X-Forwarded-For header in this plugin with client cert auth.

asifabashar:
> [@asifabashar](https://github.com/asifabashar) what do your solution look like? I'd be happy to review a PR? I believe we already have support for X-Forwarded-For header in this plugin with client cert auth.
X-Forwarded-For is for IP ,  but in this case  , we would like to assign original client's cert's principle as user to OpenSearch by proxy server cert validation and principle comparison of proxy server's client cert.
[...]

PR #6287, opened 2026-07-08
read cert from header when trusted proxy sets it

asifabashar:
### Description
Where there a external load balancer/proxy that itself does mTLS authentication, it forwards the request to opensearch where client to opensearch is load balancer and original client cert is no longer performing client cert authentication with OpenSearch , it is the load balancer that is performing client authentication with opensearch cluster.

What solution would you like?
To preserve identity of the original client, we can have first opensearch do client authentication validation with external load balancer /proxy , during TLS handshake and then external load balancer/proxy can store the original client certificate in header as URL encoded value in header field x-client-cert (This is  configurable by setting  "plugins.security.ssl.http.clientauth_header_cert_name"). The external load balancer/proxy will connect to opensearch cluster and forward the request with original client cert in header so that opensearch cluster can assign original client's principle which is correct identity of the original client. The load balancer has to be trusted by the opensearch so that it can trust its header so load balancer certificate will be validated at every request by opensearch using its standard certficate based authentication and then will validate if the external load balancer certificate principle matches setting plugins.security.ssl.http.clientauth_header_cert_allowed_for_principle list of regex.

The feature  will be controlled by flag settings  "clientauth_header_cert_allowed_for_principle" where by default this feature will be disabled.

* Category (Enhancement)

Some of the general purpose load balancers are allowing the actual cert to be placed in header, but those load balancers are not setting x-forwarded-for field with the principle as its dedicated for IP. So allowing opensearch to extract its desired field from the certificate allows both opensearch and load balancer to work with their own requirements and configuration but still work together. In this scenerio,load balancer will validate original client and but it does not have the role mapping or correct principle extraction logic, so it will just pass the cert as is, Opensearch then will know the request is from proxy from proxy(external load balancer) by validating   client cert authentication of the load balancer/proxy,  so it will trust the request , but it should assume the actual client associated user that belongs to original client and not the load balancer , so it will then use certificate from header  to switch to that principle .

By default the flag to read from header cert is false, so it does not change default behaviour. 
If user sets this flag plugins.security.ssl.http.clientauth_use_header_cert to true, then a header cert will be checked when plugins.security.ssl.http.clientauth_header_cert_allowed_for_principle setting has the principle regex list that matches proxy server (external load balancer) client cert principle matching .

### Issues Resolved
[ #6268 ] 

- Not a backport
-No new permission
### Testing
[unit testing added]

### Check List
- [x ] New functionality includes testing
- [ x] New functionality has been documented - if approved, I will add a documentation PR

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/http/CertificateAuthenticationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/http/CertificateAuthenticationTest.java
@@ -9,13 +9,17 @@
 */
 package org.opensearch.security.http;
 
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.hc.core5.http.message.BasicHeader;
 import org.junit.ClassRule;
 import org.junit.Test;
 
+import org.opensearch.security.ssl.util.SSLConfigConstants;
 import org.opensearch.test.framework.TestSecurityConfig;
 import org.opensearch.test.framework.TestSecurityConfig.AuthcDomain;
 import org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.HttpAuthenticator;
@@ -66,7 +70,14 @@ public class CertificateAuthenticationTest {
 
     @ClassRule
     public static final LocalCluster cluster = new LocalCluster.Builder().nodeSettings(
-        Map.of("plugins.security.ssl.http.clientauth_mode", "OPTIONAL")
+        Map.of(
+            "plugins.security.ssl.http.clientauth_mode",
+            "OPTIONAL",
+            SSLConfigConstants.SECURITY_SSL_HTTP_USE_HEADER_CERT,
+            "true",
+            SSLConfigConstants.SECURITY_SSL_HTTP_HEADER_CERT_ALLOWED_PROXY_PRINCIPLE,
+            "DC=de,L=test,O=users,OU=bridge,CN=spock"
+        )
     )
         .clusterManager(ClusterManager.THREE_CLUSTER_MANAGERS)
         .anonymousAuth(false)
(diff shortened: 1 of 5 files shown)


asifabashar:
> ## PR Code Analyzer :exclamation:
> 
> **AI-powered '*Code-Diff-Analyzer*' found issues on commit 9eea16f3b50ef00ce0fa0c038fe90c0ddb075f01.**
> 
> <table><tr><th>Path</th><th>Line</th><th>Severity</th><th>Description</th></tr><tr><td>src/main/java/org/opensearch/security/ssl/util/HeaderClientCertResolver.java</td><td>67</td><td>high</td><td>The certificate parsed from the HTTP header is never validated against the configured CA trust store. `factory.generateCertificate()` only deserializes the DER/PEM bytes into an X509Certificate object — it does not verify the signature chain, check revocation, or confirm the cert was issued by a trusted CA. Any allowed proxy principal (i.e., any connection whose TLS cert matches the allowlist) can present a completely self-signed or forged certificate in the header and have its principal accepted as the authenticated identity. A compromised or impersonated proxy principal therefore gives an attacker full control over which user they authenticate as.</td></tr><tr><td>src/main/java/org/opensearch/security/ssl/util/HeaderClientCertResolver.java</td><td>44</td><td>medium</td><td>The feature introduces an authentication-override path driven by an untrusted HTTP header. Even though it is gated behind `SECURITY_SSL_HTTP_USE_HEADER_CERT=true` and an allowed-principal check, any misconfiguration (e.g., an overly broad wildcard in `SECURITY_SSL_HTTP_HEADER_CERT_ALLOWED_PROXY_PRINCIPLE`, or TLS termination that does not strip incoming headers before adding its own) allows a client to supply an arbitrary certificate identity and bypass the TLS handshake-based authentication entirely. The design fundamentally trusts user-controlled header data for authentication decisions.</td></tr><tr><td>src/main/java/org/opensearch/security/ssl/util/HeaderClientCertResolver.java</td><td>43</td><td>low</td><td>The default header name `x-client-cert` is a well-known, predictable name. If a deployment enables the feature but the upstream proxy or ingress controller does not explicitly strip this header from external requests before appending its own validated value, an external attacker can inject an arbitrary certificate identity. There is no warning in the code or documentation that the operator must configure the ingress to sanitize this header.</td></tr></table><p><i>The table above displays the top 10 most important findings.</i> <br/><br/><strong>Total: 3 | Critical: 0 | High: 1 | Medium: 1 | Low: 1</strong></p>
> 
> ***
> 
> **Pull Requests Author(s)**: Please update your Pull Request according to the report above.
> 
> **Repository Maintainer(s)**: You can `bypass diff analyzer` by adding label `skip-diff-analyzer` after reviewing the changes carefully, then `re-run failed actions`. To re-enable the analyzer, remove the label, then `re-run all actions`.
> 
> ***
> 
> :warning: Note: The *Code-Diff-Analyzer* helps protect against potentially harmful code patterns. Please ensure you have thoroughly reviewed the changes beforehand.
> 
> Thanks.

The ssl validation for client happens at the load balancers/proxy servers , the opensearch still validates load balancers/proxy certificates and trust that load balancer did correct validation. So it is dependent on load balancer/proxy to do original client validation.

asifabashar:
@cwperks please review

cwperks:
> @cwperks please review

@asifabashar apologies for delay, can you please make sure this has passing checks?

nibix:
I have one question: There is already the proxy auth method https://docs.opensearch.org/latest/security/authentication-backends/proxy/ which can be used to retrieve user information from HTTP headers in case trusted remote hosts are recognized.

This is quite close to the requirement stated in this PR, the only difference is - as far as I can see - that using the code in this PR the certificate is parsed on the OpenSearch side.

As you have already a TLS terminating proxy in the chain, wouldnt it be possible to extract principal information on the proxy side and pass it on to the OpenSearch side as plain text in HTTP headers and then do the authenticaton using the proxy auth method?

asifabashar:
> I have one question: There is already the proxy auth method https://docs.opensearch.org/latest/security/authentication-backends/proxy/ which can be used to retrieve user information from HTTP headers in case trusted remote hosts are recognized.
> 
> 
> 
> This is quite close to the requirement stated in this PR, the only difference is - as far as I can see - that using the code in this PR the certificate is parsed on the OpenSearch side.
> 
> 
> 
> As you have already a TLS terminating proxy in the chain, wouldnt it be possible to extract principal information on the proxy side and pass it on to the OpenSearch side as plain text in HTTP headers and then do the authenticaton using the proxy auth method?

Thanks for checking. In many cases , the proxy is third party app where custom principal parsing logic is not available, but opensearch security plugin recently has had some enhancements that allow for principle parsing from multiple fields in certificate such as SAN or Subject itself . So parsing cert at Opensearch provides enhanced options for authentication .

PR #6304, opened 2026-07-13
feat(audit): Implement standalone audit logging for SSL-only mode

### LINKED ISSUE #6303: [FEATURE] Standalone audit logging for SSL-only mode
### Description
  
Enable audit logging as an independent capability in SSL-only mode (`plugins.security.ssl_only: true`) without requiring full security (FGAC/authentication/authorization).
  
### Background
  
Currently, audit logging only produces events when the full security stack is active. Users running OpenSearch with TLS-only (for log analytics, observability, internal clusters) have no audit trail. Compliance frameworks (SOC2, HIPAA, PCI-DSS) require audit logs regardless of access control configuration.
  
This is the first phase of a multi-phase effort toward a fully standalone audit plugin:
  
  1. **SSL-only mode** Only TLS encryption, no authentication or authorization 
  2. **Disabled mode** No security at all
  3. **Standalone audit plugin** (extracted from security plugin) — future
  
### Scope

- New `REQUEST_AUDIT` category for all REST/transport actions
- Client certificate identity enrichment (mTLS)
- Request body logging with sensitive header exclusion
- Index resolution and bulk per-item event tracking
- Ignore-users and ignore-requests filtering
- Transport-layer interception
- Document-level compliance tracking (COMPLIANCE_DOC_WRITE, COMPLIANCE_DOC_READ)
- All existing sinks supported (internal_opensearch, log4j, webhook, Kafka)
- 17 settings dynamically configurable via `PUT _cluster/settings`
  
### Configuration
  
  ```yaml
  plugins.security.ssl_only: true
  plugins.security.audit.type: log4j  # or internal_opensearch, webhook
  ```
  
Related
  
- Broader vision: https://github.com/opensearch-project/.github/issues/501 
- Upstream PR (unified disabled_categories): https://github.com/opensearch-project/security/pull/6271

Taiwo435:
Enable audit logging as an independent capability in SSL-only mode (no auth/RBAC required). Users running OpenSearch without full security can now have a complete audit trail for compliance (SOC2, HIPAA, PCI-DSS).

Key features:
- New REQUEST_AUDIT category for all REST/transport actions
- Client certificate identity enrichment (mTLS CN/SAN as effective_user)
- Request body logging with sensitive header exclusion
- Index resolution (wildcards expanded to actual indices)
- Bulk per-item event tracking (configurable granularity)
- Ignore-users and ignore-requests filtering
- Transport-layer interception (shard-level operations)
- Document-level compliance tracking (COMPLIANCE_DOC_WRITE and COMPLIANCE_DOC_READ) via IndexingOperationListener and lightweight ComplianceReadIndexSearcherWrapper
- All sinks supported (internal_opensearch, log4j, webhook, Kafka)
- 17 dynamic settings configurable at runtime via PUT _cluster/settings without node restart

New production files:
- AuditActionFilter: lightweight ActionFilter for non-FGAC modes
- AuditTransportInterceptor: captures inter-node transport traffic
- ComplianceReadIndexSearcherWrapper: field-level read tracking
- SecuritySettings: dynamic Setting constants for cluster settings

New test files:
- AuditActionFilterTest: 11 unit tests
- AuditTransportInterceptorTest: 11 unit tests
- StandaloneAuditLoggingTest: 21 integration tests (basic events)
- StandaloneAuditFilterFeaturesTest: 25 integration tests
- StandaloneAuditComplianceDocTest: 14 integration tests
- StandaloneAuditDynamicFilterSettingsTest: 10 integration tests
- StandaloneAuditDynamicComplianceSettingsTest: 7 integration tests
- StandaloneAuditDynamicConfigTest: 2 integration tests
- StandaloneAuditTransportInterceptTest: 2 integration tests
- StandaloneAuditMtlsTest: 1 integration test
- StandaloneAuditDisabledCategoryTest: 1 integration test
- StandaloneAuditSinksTest: 3 integration tests
- StandaloneAuditWebhookSinkTest: 1 integration test

Total tests: 128 (43 unit + 85 integration)

### Description
  
  **Category:** New feature
  
  Users running OpenSearch in SSL-only mode (TLS without authentication/authorization) currently have zero audit trail. Compliance frameworks (SOC2, HIPAA, PCI-DSS, GDPR) require audit logs even without access
  control. This change makes audit logging an independent capability that works without FGAC.
  
  **What is the old behavior before changes and new behavior after changes?**
  
  - **Before:** Audit logging only works when full security (FGAC) is active. In SSL-only mode, no audit events are produced regardless of configuration.
  - **After:** In SSL-only mode (`plugins.security.ssl_only: true`), configuring an audit sink (`plugins.security.audit.type: log4j`) now enables full audit logging — capturing all REST/transport actions, request
  bodies, client cert identity, compliance doc read/write tracking, with 17 dynamically configurable settings.
  
  ### Issues Resolved
  
  Resolves https://github.com/opensearch-project/security/issues/6303
  
  ### Testing
  
  - **Unit tests:** 43 tests (AuditActionFilterTest, AuditTransportInterceptorTest, DisabledCategoriesTest, AuditConfigFilterTest)
  - **Integration tests:** 85 tests across 11 test classes covering all features, edge cases, dynamic config toggling, compliance tracking, transport interception, identity, sinks
  - **Manual testing:** Docker cluster (SSL-only mode, log4j sink) verified all phases — basic events, identity (mTLS), index resolution, bulk per-item, ignore patterns, compliance doc write/read, dynamic config
  toggle, transport intercept
  
  ### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/StandaloneAuditDynamicFilterSettingsTest.java
+++ b/src/integrationTest/java/org/opensearch/security/StandaloneAuditDynamicFilterSettingsTest.java
@@ -327,4 +327,142 @@ public class StandaloneAuditDynamicFilterSettingsTest {
             client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.exclude_sensitive_headers\": true}}");
         }
     }
+
+    // =====================================================================
+    // Transport interceptor — verify it reads live filter settings
+    // =====================================================================
+
+    @Test
+    public void shouldSuppressTransportEventsWhenCategoryDisabledAtRuntime() {
+        // First verify TRANSPORT_AUDIT events are produced
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson("transport-dynamic-test/_doc/1?refresh=true", "{\"val\": \"before-disable\"}");
+        }
+
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> msg.getCategory() == AuditCategory.TRANSPORT_AUDIT);
+
+        // Now disable TRANSPORT_AUDIT dynamically
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson(
+                "_cluster/settings",
+                "{\"persistent\": {\"plugins.security.audit.config.disabled_transport_categories\": [\"TRANSPORT_AUDIT\"]}}"
+            );
+        }
+
+        auditLogsRule.waitForAuditLogs();
+
+        // Index again — TRANSPORT_AUDIT should be suppressed
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson("transport-dynamic-test/_doc/2?refresh=true", "{\"val\": \"after-disable\"}");
+        }
+
+        // Should still get REQUEST_AUDIT but no TRANSPORT_AUDIT for this second write
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.REQUEST_AUDIT) return false;
+            Map<String, Object> fields = msg.getAsMap();
+            Object indices = fields.get(AuditMessage.INDICES);
+            if (indices == null) return false;
+            String[] indexArr = (String[]) indices;
+            for (String idx : indexArr) {
+                if ("transport-dynamic-test".equals(idx)) return true;
+            }
+            return false;
+        });
+
+        // Reset
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.disabled_transport_categories\": []}}");
+        }
+    }
+
+    @Test
+    public void shouldRespectIgnoreRequestsOnTransportEventsAtRuntime() {
+        // Dynamically add a pattern to ignore_requests that matches shard-level writes
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson(
+                "_cluster/settings",
+                "{\"persistent\": {\"plugins.security.audit.config.ignore_requests\": [\"indices:data/write/*\"]}}"
+            );
+        }
+
+        auditLogsRule.waitForAuditLogs();
+
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson("transport-ignore-req/_doc/1?refresh=true", "{\"val\": \"filtered\"}");
+        }
+
+        // TRANSPORT_AUDIT events for write actions should be suppressed
+        auditLogsRule.waitForAuditLogs();
+        auditLogsRule.assertExactlyScanAll(0, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.TRANSPORT_AUDIT) return false;
+            Map<String, Object> fields = msg.getAsMap();
+            String privilege = (String) fields.get(AuditMessage.PRIVILEGE);
+            return privilege != null && privilege.contains("indices:data/write");
+        });
+
+        // Reset
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.ignore_requests\": []}}");
+        }
+    }
+
+    // =====================================================================
+    // Transport interceptor — re-enable after disable proves bidirectional
+    // =====================================================================
+
+    @Test
+    public void shouldResumeTransportEventsWhenCategoryReenabledAtRuntime() {
+        // Disable TRANSPORT_AUDIT
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson(
+                "_cluster/settings",
+                "{\"persistent\": {\"plugins.security.audit.config.disabled_transport_categories\": [\"TRANSPORT_AUDIT\"]}}"
+            );
+        }
+
+        auditLogsRule.waitForAuditLogs();
+
+        // Re-enable by clearing disabled categories
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.disabled_transport_categories\": []}}");
+        }
+
+        auditLogsRule.waitForAuditLogs();
+
+        // Index a doc — TRANSPORT_AUDIT should now appear again
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson("transport-reenable-test/_doc/1?refresh=true", "{\"val\": \"back\"}");
+        }
+
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> msg.getCategory() == AuditCategory.TRANSPORT_AUDIT);
+    }
+
+    // =====================================================================
+    // disabled_rest_categories should NOT suppress TRANSPORT_AUDIT
+    // =====================================================================
+
+    @Test
+    public void shouldNotSuppressTransportAuditWhenOnlyRestCategoryDisabled() {
+        // Disable TRANSPORT_AUDIT via REST categories only (should not affect transport interceptor)
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson(
+                "_cluster/settings",
+                "{\"persistent\": {\"plugins.security.audit.config.disabled_rest_categories\": [\"TRANSPORT_AUDIT\"]}}"
+            );
+        }
+
+        auditLogsRule.waitForAuditLogs();
+
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson("transport-rest-cat-test/_doc/1?refresh=true", "{\"val\": \"still-logged\"}");
+        }
+
+        // TRANSPORT_AUDIT events should still appear — REST category disable doesn't affect transport
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> msg.getCategory() == AuditCategory.TRANSPORT_AUDIT);
+
+        // Reset
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.disabled_rest_categories\": []}}");
+        }
+    }
 }
--- a/src/integrationTest/java/org/opensearch/security/StandaloneAuditDynamicFilterSettingsTest.java
+++ b/src/integrationTest/java/org/opensearch/security/StandaloneAuditDynamicFilterSettingsTest.java
@@ -43,7 +43,10 @@ public class StandaloneAuditDynamicFilterSettingsTest {
                 TestRuleAuditLogSink.class.getName(),
                 // Start with body enabled (default)
                 ConfigConstants.OPENDISTRO_SECURITY_AUDIT_LOG_REQUEST_BODY,
-                true
+                true,
+                // Enable client cert auth so mTLS tests can present a DN
+                "plugins.security.ssl.http.clientauth_mode",
+                "OPTIONAL"
             )
         )
         .sslOnly(true)
@@ -101,19 +104,44 @@ public class StandaloneAuditDynamicFilterSettingsTest {
 
     @Test
     public void shouldIgnoreUsersAddedAtRuntime() {
-        try (TestRestClient client = cluster.getRestClient()) {
-            // Add a wildcard ignore pattern at runtime
-            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.ignore_users\": [\"*\"]}}");
-
-            // This should be suppressed (wildcard matches all identified users)
-            // But since SSL-only with no client cert = null user, it will still log
-            // So we verify by checking that the setting was accepted (no error)
-            TestRestClient.HttpResponse response = client.get("_cluster/settings");
-            response.assertStatusCode(200);
+        // Step 1: Verify events WITH cert DN ARE produced before ignore
+        try (TestRestClient client = cluster.getRestClient(cluster.getTestCertificates().getAdminCertificateData())) {
+            client.get("ignore-user-before/_search");
         }
 
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.REQUEST_AUDIT) return false;
+            return msg.getEffectiveUser() != null && msg.getEffectiveUser().contains("CN=kirk");
+        });
+
+        // Step 2: Dynamically add the cert DN pattern to ignore_users
+        try (TestRestClient client = cluster.getRestClient(cluster.getTestCertificates().getAdminCertificateData())) {
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.ignore_users\": [\"*kirk*\"]}}");
+        }
+
+        auditLogsRule.waitForAuditLogs();
+
+        // Step 3: Issue a request with the same cert — should be suppressed
+        try (TestRestClient client = cluster.getRestClient(cluster.getTestCertificates().getAdminCertificateData())) {
+            client.get("ignore-user-after/_search");
+        }
+
+        auditLogsRule.waitForAuditLogs();
+        auditLogsRule.assertExactlyScanAll(0, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.REQUEST_AUDIT) return false;
+            if (msg.getPrivilege() == null || !msg.getPrivilege().contains("indices:data/read/search")) return false;
+            Map<String, Object> fields = msg.getAsMap();
+            Object indices = fields.get(AuditMessage.INDICES);
+            if (indices == null) return false;
+            String[] indexArr = (String[]) indices;
+            for (String idx : indexArr) {
+                if ("ignore-user-after".equals(idx)) return true;
+            }
+            return false;
+        });
+
         // Reset
-        try (TestRestClient client = cluster.getRestClient()) {
+        try (TestRestClient client = cluster.getRestClient(cluster.getTestCertificates().getAdminCertificateData())) {
             client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.ignore_users\": []}}");
         }
     }
--- a/src/integrationTest/java/org/opensearch/security/StandaloneAuditDynamicFilterSettingsTest.java
+++ b/src/integrationTest/java/org/opensearch/security/StandaloneAuditDynamicFilterSettingsTest.java
@@ -465,4 +465,53 @@ public class StandaloneAuditDynamicFilterSettingsTest {
             client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.disabled_rest_categories\": []}}");
         }
     }
+
+    // =====================================================================
+    // ignore_requests — unified suppression across REST and transport layers
+    // =====================================================================
+
+    @Test
+    public void shouldSuppressBothRestAndTransportEventsWithSameIgnorePattern() {
+        // Add a pattern that matches write actions — affects both layers
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson(
+                "_cluster/settings",
+                "{\"persistent\": {\"plugins.security.audit.config.ignore_requests\": [\"indices:data/write/*\"]}}"
+            );
+        }
+
+        auditLogsRule.waitForAuditLogs();
+
+        // Send a write — should be suppressed at BOTH REQUEST_AUDIT and TRANSPORT_AUDIT
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson("unified-ignore-test/_doc/1?refresh=true", "{\"data\": \"suppressed\"}");
+        }
+
+        auditLogsRule.waitForAuditLogs();
+
+        // No REQUEST_AUDIT events for write actions on this index
+        auditLogsRule.assertExactlyScanAll(0, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.REQUEST_AUDIT) return false;
+            if (msg.getPrivilege() == null || !msg.getPrivilege().startsWith("indices:data/write")) return false;
+            Map<String, Object> fields = msg.getAsMap();
+            Object indices = fields.get(AuditMessage.INDICES);
+            if (indices == null) return false;
+            String[] indexArr = (String[]) indices;
+            for (String idx : indexArr) {
+                if ("unified-ignore-test".equals(idx)) return true;
+            }
+            return false;
+        });
+
+        // No TRANSPORT_AUDIT events for write actions either
+        auditLogsRule.assertExactlyScanAll(0, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.TRANSPORT_AUDIT) return false;
+            return msg.getPrivilege() != null && msg.getPrivilege().startsWith("indices:data/write");
+        });
+
+        // Reset
+        try (TestRestClient client = cluster.getRestClient()) {
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.ignore_requests\": []}}");
+        }
+    }
 }
--- a/src/integrationTest/java/org/opensearch/security/StandaloneAuditDynamicFilterSettingsTest.java
+++ b/src/integrationTest/java/org/opensearch/security/StandaloneAuditDynamicFilterSettingsTest.java
@@ -37,10 +37,13 @@ public class StandaloneAuditDynamicFilterSettingsTest {
         .loadConfigurationIntoIndex(false)
         .nodeSettings(
             Map.of(
-                ConfigConstants.SECURITY_SSL_ONLY, true,
-                "plugins.security.audit.type", TestRuleAuditLogSink.class.getName(),
+                ConfigConstants.SECURITY_SSL_ONLY,
+                true,
+                "plugins.security.audit.type",
+                TestRuleAuditLogSink.class.getName(),
                 // Start with body enabled (default)
-                ConfigConstants.OPENDISTRO_SECURITY_AUDIT_LOG_REQUEST_BODY, true
+                ConfigConstants.OPENDISTRO_SECURITY_AUDIT_LOG_REQUEST_BODY,
+                true
             )
         )
         .sslOnly(true)
@@ -57,8 +60,7 @@ public class StandaloneAuditDynamicFilterSettingsTest {
     public void shouldStopLoggingBodyWhenDisabledAtRuntime() {
         try (TestRestClient client = cluster.getRestClient()) {
             // Disable request body logging dynamically
-            client.putJson("_cluster/settings",
-                "{\"persistent\": {\"plugins.security.audit.config.log_request_body\": false}}");
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.log_request_body\": false}}");
 
             // This request's body should NOT appear in audit
             client.postJson("dynamic-body-test/_search", "{\"query\": {\"match_all\": {}}}");
@@ -72,8 +74,7 @@ public class StandaloneAuditDynamicFilterSettingsTest {
 
         // Re-enable for subsequent tests
         try (TestRestClient client = cluster.getRestClient()) {
-            client.putJson("_cluster/settings",
-                "{\"persistent\": {\"plugins.security.audit.config.log_request_body\": true}}");
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.log_request_body\": true}}");
         }
     }
 
@@ -81,8 +82,7 @@ public class StandaloneAuditDynamicFilterSettingsTest {
     public void shouldResumeLoggingBodyWhenReenabledAtRuntime() {
         try (TestRestClient client = cluster.getRestClient()) {
             // Ensure body logging is enabled
-            client.putJson("_cluster/settings",
-                "{\"persistent\": {\"plugins.security.audit.config.log_request_body\": true}}");
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.log_request_body\": true}}");
 
             client.postJson("dynamic-body-resume/_search", "{\"query\": {\"term\": {\"status\": \"active\"}}}");
         }
@@ -103,8 +103,7 @@ public class StandaloneAuditDynamicFilterSettingsTest {
     public void shouldIgnoreUsersAddedAtRuntime() {
         try (TestRestClient client = cluster.getRestClient()) {
             // Add a wildcard ignore pattern at runtime
-            client.putJson("_cluster/settings",
-                "{\"persistent\": {\"plugins.security.audit.config.ignore_users\": [\"*\"]}}");
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.ignore_users\": [\"*\"]}}");
 
             // This should be suppressed (wildcard matches all identified users)
             // But since SSL-only with no client cert = null user, it will still log
@@ -115,8 +114,7 @@ public class StandaloneAuditDynamicFilterSettingsTest {
 
         // Reset
         try (TestRestClient client = cluster.getRestClient()) {
-            client.putJson("_cluster/settings",
-                "{\"persistent\": {\"plugins.security.audit.config.ignore_users\": []}}");
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.ignore_users\": []}}");
         }
     }
 
@@ -128,24 +126,26 @@ public class StandaloneAuditDynamicFilterSettingsTest {
     public void shouldIgnoreRequestsAddedAtRuntime() {
         try (TestRestClient client = cluster.getRestClient()) {
             // Dynamically ignore search requests
-            client.putJson("_cluster/settings",
-                "{\"persistent\": {\"plugins.security.audit.config.ignore_requests\": [\"indices:data/read/search\"]}}");
+            client.putJson(
+                "_cluster/settings",
+                "{\"persistent\": {\"plugins.security.audit.config.ignore_requests\": [\"indices:data/read/search\"]}}"
+            );
 
             // Search should now be suppressed
             client.postJson("dynamic-ignore-req/_search", "{\"query\": {\"match_all\": {}}}");
         }
 
         auditLogsRule.waitForAuditLogs();
-        auditLogsRule.assertExactlyScanAll(0, (AuditMessage msg) ->
-            msg.getCategory() == AuditCategory.REQUEST_AUDIT
+        auditLogsRule.assertExactlyScanAll(
+            0,
+            (AuditMessage msg) -> msg.getCategory() == AuditCategory.REQUEST_AUDIT
                 && msg.getPrivilege() != null
                 && msg.getPrivilege().equals("indices:data/read/search")
         );
 
         // Reset
         try (TestRestClient client = cluster.getRestClient()) {
-            client.putJson("_cluster/settings",
-                "{\"persistent\": {\"plugins.security.audit.config.ignore_requests\": []}}");
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.ignore_requests\": []}}");
         }
     }
 
@@ -153,17 +153,19 @@ public class StandaloneAuditDynamicFilterSettingsTest {
     public void shouldStopIgnoringRequestsWhenCleared() {
         try (TestRestClient client = cluster.getRestClient()) {
             // First ignore, then clear
-            client.putJson("_cluster/settings",
-                "{\"persistent\": {\"plugins.security.audit.config.ignore_requests\": [\"indices:data/read/search\"]}}");
-            client.putJson("_cluster/settings",
-                "{\"persistent\": {\"plugins.security.audit.config.ignore_requests\": []}}");
+            client.putJson(
+                "_cluster/settings",
+                "{\"persistent\": {\"plugins.security.audit.config.ignore_requests\": [\"indices:data/read/search\"]}}"
+            );
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.ignore_requests\": []}}");
 
             // Search should now be logged again
             client.postJson("dynamic-unignore/_search", "{\"query\": {\"match_all\": {}}}");
         }
 
-        auditLogsRule.assertAtLeast(1, (AuditMessage msg) ->
-            msg.getCategory() == AuditCategory.REQUEST_AUDIT
+        auditLogsRule.assertAtLeast(
+            1,
+            (AuditMessage msg) -> msg.getCategory() == AuditCategory.REQUEST_AUDIT
                 && msg.getPrivilege() != null
                 && msg.getPrivilege().contains("indices:data/read/search")
         );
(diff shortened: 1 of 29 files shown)


cwperks:
@Taiwo435 please look into the test failures. Also address the code hygiene failure by running `./gradlew spotlessApply` and check in the formatted files.

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
```diff
@@ -545,6 +547,54 @@ public Set<AuditCategory> getDisabledCategories() {
             return disabledCategories;
         }
 
+        // Dynamic setters for cluster settings updates
+
+        public void setLogRequestBody(boolean logRequestBody) {
```
DarshitChanpura, line 552: have we tested that changes to these settings persist on cluster restart?

Also;   nit (concurrency): setWatchedReadFields (and any similar setter that updates multiple instance fields from a single logical config change) writes watchedReadFields, then readEnabledFields, then invalidates the cache as three separate operations. A concurrent reader thread can observe a state where the new matchers are visible but the old cache hasn't been invalidated yet (or vice versa), leading to stale/inconsistent compliance-read decisions.

Suggested fix — wrap the related fields in a single immutable holder and swap it atomically:

private static class WatchedReadState {
      final Map<String, List<String>> watchedReadFields;
      final Map<WildcardMatcher, Set<String>> readEnabledFields;

      WatchedReadState(Map<String, List<String>> watchedReadFields,
                       Map<WildcardMatcher, Set<String>> readEnabledFields) {
          this.watchedReadFields = watchedReadFields;
          this.readEnabledFields = readEnabledFields;
      }
  }

private volatile WatchedReadState watchedReadState =  new WatchedReadState(Map.of(), Map.of());

Then in the setter, build both maps in one pass and assign once:

this.watchedReadState = new WatchedReadState(Map.copyOf(parsed), Map.copyOf(enabled));
this.readEnabledFieldsCache.invalidateAll();

Readers only ever see a fully-consistent pair of maps. The cache invalidation after the swap is safe because any cache miss will recompute against the already-visible new state.

src/main/java/org/opensearch/security/auditlog/impl/AuditCategory.java
```diff
@@ -33,7 +33,22 @@ public enum AuditCategory {
     COMPLIANCE_INTERNAL_CONFIG_WRITE,
     CLUSTER_SETTINGS_CHANGED,
     INDEX_SETTINGS_CHANGED,
-    API_TOKEN_WRITE;
+    API_TOKEN_WRITE,
+    REQUEST_AUDIT,
+    TRANSPORT_AUDIT;
```
DarshitChanpura, line 38: what is the TRANSPORT_AUDIT category? Is it for node-to-node requests?

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -149,6 +149,10 @@ protected void onAuditConfigFilterChanged(AuditConfig.Filter auditConfigFilter)
         this.auditConfigFilter.log(log);
     }
 
+    public AuditConfig.Filter getFilter() {
```
DarshitChanpura, line 163: can we add javadoc for this method?

src/main/java/org/opensearch/security/compliance/ComplianceConfig.java
```diff
@@ -340,8 +342,10 @@ public static ComplianceConfig from(Settings settings, Supplier<DateTime> datePr
             false
         );
 
+        final boolean enabled = SecuritySettings.COMPLIANCE_ENABLED.get(settings);
+
         return new ComplianceConfig(
-            true,
+            enabled,
```
DarshitChanpura, line 363: we will need to document all these changes through doc website.

src/main/java/org/opensearch/security/compliance/ComplianceConfig.java
```diff
@@ -383,6 +387,46 @@ public boolean isEnabled() {
...
+                    split -> split.length == 1 ? List.of("*") : Arrays.stream(split).skip(1).collect(Collectors.toList())
+                )
+            );
+        this.watchedReadFields = parsed;
+        this.readEnabledFields = parsed.entrySet()
+            .stream()
+            .filter(entry -> !entry.getKey().isEmpty())
+            .collect(Collectors.toMap(entry -> WildcardMatcher.from(entry.getKey()), entry -> Set.copyOf(entry.getValue())));
+        this.readEnabledFieldsCache.invalidateAll();
+    }
```
DarshitChanpura, line 443:
```suggestion
 public void setWatchedReadFields(List<String> readFields) {
      // Parse format: "indexpattern,field1,field2,..." — empty field list means all fields ("*")
      final Map<String, List<String>> parsed = new HashMap<>(readFields.size());
      final Map<WildcardMatcher, Set<String>> enabled = new HashMap<>(readFields.size());

      for (final String entry : readFields) {
          final String[] split = entry.split(",");
          if (split.length == 0 || split[0].isEmpty()) {
              continue;
          }
          final List<String> fields = split.length == 1
              ? List.of("*")
              : List.of(Arrays.copyOfRange(split, 1, split.length));
          parsed.put(split[0], fields);                          // last entry wins on duplicate patterns
          enabled.put(WildcardMatcher.from(split[0]), Set.copyOf(fields));
      }

      this.watchedReadFields = Map.copyOf(parsed);
      this.readEnabledFields = Map.copyOf(enabled);
      this.readEnabledFieldsCache.invalidateAll();
  }
```

Here is a single pass approach compared to two pass above.

src/main/java/org/opensearch/security/support/SecuritySettings.java
```diff
@@ -57,4 +61,181 @@ public class SecuritySettings {
...
+        COMPLIANCE_PREFIX + "read_ignore_users",
+        Collections.emptyList(),
+        Function.identity(),
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
+
+    public static final Setting<List<String>> COMPLIANCE_WRITE_IGNORE_USERS = Setting.listSetting(
+        COMPLIANCE_PREFIX + "write_ignore_users",
+        Collections.emptyList(),
```
DarshitChanpura: for sake of correctness, have we double checked the default values that we supply over here in the settings?

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
```diff
@@ -0,0 +1,309 @@
...
+                msg.addEffectiveUser(user.getName());
+            } else if (principal != null) {
+                msg.addEffectiveUser(principal);
+            }
+
+            // REST headers (stashed by REST wrapper, filtered here)
+            Map<String, List<String>> headers = threadPool.getThreadContext().getTransient(ConfigConstants.SECURITY_AUDIT_REST_HEADERS);
+            if (headers != null && !headers.isEmpty()) {
+                Map<String, List<String>> filteredHeaders = new HashMap<>(headers);
+                if (filter.shouldExcludeSensitiveHeaders()) {
```
DarshitChanpura: bug (correctness): Header filtering behaves differently between AuditActionFilter and AuditTransportInterceptor.

In AuditActionFilter, filtering respects the config setting and uses WildcardMatcher:
```java
if (filter.shouldExcludeSensitiveHeaders()) {
    filteredHeaders.keySet().removeIf(AUTHORIZATION_HEADER);
}
```
In AuditTransportInterceptor, it unconditionally strips headers with a hardcoded string check:
```java
filteredHeaders.keySet().removeIf(k -> k.equalsIgnoreCase("authorization"));
```
Two problems:

1. The transport interceptor ignores the exclude_sensitive_headers setting entirely — if an operator sets it to false (e.g., for internal forensic debugging), transport events still strip the header.
2. The transport interceptor only matches the literal string "authorization". The action filter uses WildcardMatcher which handles case-insensitive matching via the matcher itself. If ignore_headers is ever extended to cover additional sensitive headers, the transport interceptor won't pick them up.

Suggested fix — reuse the same pattern in both places:
```java
// In AuditTransportInterceptor, same as AuditActionFilter:
if (filter.shouldExcludeSensitiveHeaders()) {
    filteredHeaders.keySet().removeIf(AUTHORIZATION_HEADER);
}
```

Or better, extract a shared utility method so the two paths can't drift apart:

```java
// e.g., in AuditConfig.Filter or a shared AuditHeaderUtils
public static Map<String, List<String>> filterHeaders( Map<String, List<String>> headers, AuditConfig.Filter filter) {
    if (headers == null || headers.isEmpty()) return Map.of();
    Map<String, List<String>> filtered = new HashMap<>(headers);
    if (filter.shouldExcludeSensitiveHeaders()) {
        filtered.keySet().removeIf(AUTHORIZATION_HEADER);
    }
    return filtered;
}
```

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
```diff
@@ -0,0 +1,309 @@
...
+            chain.proceed(task, action, request, listener);
+            return;
+        }
+
+        // Skip requests targeting the audit index (prevent self-referential loop)
+        if (request instanceof IndicesRequest) {
+            String[] indices = ((IndicesRequest) request).indices();
+            if (indices != null) {
+                for (String idx : indices) {
+                    if (idx != null && idx.startsWith(auditIndexPrefix)) {
```
DarshitChanpura: bug (correctness): In both AuditActionFilter and AuditTransportInterceptor, the self-loop guard checks:
```java
if (idx != null && idx.startsWith(auditIndexPrefix)) {
    chain.proceed(task, action, request, listener);
    return;
}
```

If auditIndexPrefix is null or empty (misconfiguration, setting not applied, or the operator clears it at runtime via dynamic settings), then idx.startsWith("") returns true for every string. This silently suppresses all audit events across the entire cluster with no error or warning logged — effectively a silent audit failure, which is the opposite of what a
compliance feature should do.

Suggested fix — guard against null/empty prefix and log a warning if the configuration looks broken:

```java
private boolean isAuditIndex(String idx) {
    if (idx == null || auditIndexPrefix == null || auditIndexPrefix.isEmpty()) {
        return false;
    }
    return idx.startsWith(auditIndexPrefix);
}
```

Then in the loop:
```java
for (String idx : indices) {
    if (isAuditIndex(idx)) {
        chain.proceed(task, action, request, listener);
        return;
    }
}
```

You could also fail-fast at construction time — if the filter is being created with an empty prefix, that's almost certainly a bug:
```java
public AuditActionFilter(..., String auditIndexPrefix) {
    if (auditIndexPrefix == null || auditIndexPrefix.isEmpty()) {
        log.warn("Audit index prefix is empty — self-loop guard will be disabled. "
            + "Verify 'plugins.security.audit.config.index' is configured.");
    }
    this.auditIndexPrefix = auditIndexPrefix;
    ...
}
```
Either approach prevents a blank prefix from accidentally muting the entire audit pipeline.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -778,7 +956,51 @@ public List<RestHandler> getRestHandlers(
...
+                                    String principal = new DefaultPrincipalExtractor().extractPrincipal(
+                                        (X509Certificate) certs[0],
+                                        PrincipalExtractor.Type.HTTP
+                                    );
+                                    if (principal != null) {
+                                        threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_SSL_PRINCIPAL, principal);
+                                    }
+                                }
+                            } catch (Exception e) {
+                                // No peer cert available — client didn't present one (wantClientAuth case)
```
DarshitChanpura: bug (observability): The SSL certificate extraction path swallows exceptions with an empty catch block. When mTLS is misconfigured (expired cert, wrong trust chain, missing CN/SAN), the audit event will silently have no effective_user field — and there's nothing in the logs to explain why.

In a compliance context this is particularly bad: an operator sees audit events with a blank identity and has no way to tell whether the client genuinely had no cert or whether extraction failed silently. These are very different situations from a compliance standpoint.

Suggested fix — log at debug level at minimum, warn for the first occurrence:
```java
} catch (Exception e) {
    log.debug("Failed to extract client certificate identity — "
        + "audit event will have no effective_user: {}", e.getMessage());
}
```

If you want to avoid log spam when a misconfigured client is making repeated requests, we can use a rate-limited or once-per-interval warning:
```java
} catch (Exception e) {
    if (log.isDebugEnabled()) {
        log.debug("Failed to extract client certificate identity", e);
    } else {
        log.warn("Failed to extract client certificate identity: {}. "
            + "Enable DEBUG logging for full stack trace.", e.getMessage());
    }
}
```

The key thing is that production operators should have some signal when identity extraction fails — otherwise mTLS issues become undebuggable and compliance auditors will flag blank-identity events as gaps.

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
```diff
@@ -0,0 +1,309 @@
...
+                        msg.addRestHeaders(filteredHeaders, false, null);
+                    }
+                    if (filter.shouldLogRequestBody() && innerRequest instanceof IndexRequest) {
+                        IndexRequest ir = (IndexRequest) innerRequest;
+                        if (ir.source() != null) {
+                            msg.addTupleToRequestBody(new Tuple<MediaType, BytesReference>(ir.getContentType(), ir.source()));
+                        }
+                    }
+
+                    auditLog.logRequestAudit(msg);
```
DarshitChanpura: nit (maintainability): The bulk-resolution path in AuditActionFilter has two nearly identical loops — one for BulkShardRequest and one for BulkRequest. The only difference is that BulkShardRequest adds a shardId. The rest (remote address, action, effective user, headers, request body extraction, calling auditLog.logRequestAudit) is copy-pasted.

This means any future fix (e.g., the header-filtering fix from the other comment, or adding a new field to bulk audit events) needs to be applied in two places — easy to miss one.

Suggested fix — extract a shared helper:
```java
private void logBulkItem(
        DocWriteRequest<?> innerRequest,
        String action,
        Task task,
        String effectiveUser,
        TransportAddress remoteAddress,
        Map<String, List<String>> filteredHeaders,
        ShardId shardId  // nullable — only set for BulkShardRequest
) {
    AuditMessage msg = new AuditMessage(
        AuditCategory.REQUEST_AUDIT, clusterService, Origin.REST, Origin.TRANSPORT);

    msg.addRemoteAddress(remoteAddress);
    msg.addPrivilege(action);
    msg.addRequestType(innerRequest.getClass().getSimpleName());
    msg.addIndices(new String[] { innerRequest.index() });
    msg.addId(innerRequest.id());

    if (shardId != null) {
        msg.addShardId(shardId);
    }
    if (task != null) {
        msg.addTaskId(task.getId());
    }
    if (effectiveUser != null) {
        msg.addEffectiveUser(effectiveUser);
    }
    if (filteredHeaders != null) {
        msg.addRestHeaders(filteredHeaders, false, null);
    }
    if (filter.shouldLogRequestBody() && innerRequest instanceof IndexRequest) {
        IndexRequest ir = (IndexRequest) innerRequest;
        if (ir.source() != null) {
            msg.addTupleToRequestBody(
                new Tuple<>(ir.getContentType(), ir.source()));
        }
    }

    auditLog.logRequestAudit(msg);
}
```
Then the two loops become:
```java
if (request instanceof BulkShardRequest) {
    BulkShardRequest bsr = (BulkShardRequest) request;
    for (BulkItemRequest item : bsr.items()) {
        logBulkItem(item.request(), action, task, effectiveUser,
            remoteAddress, filteredHeaders, bsr.shardId());
    }
} else {
    BulkRequest br = (BulkRequest) request;
    for (DocWriteRequest<?> innerRequest : br.requests()) {
        logBulkItem(innerRequest, action, task, effectiveUser,
            remoteAddress, filteredHeaders, null);
    }
}
```

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
```diff
@@ -0,0 +1,309 @@
...
+        this.clusterService = clusterService;
+        this.threadPool = threadPool;
+        this.filter = filter;
+        this.resolver = resolver;
+        this.auditIndexPrefix = auditIndexPrefix;
+    }
+
+    @Override
+    public int order() {
+        return Integer.MIN_VALUE;
```
DarshitChanpura: This guarantees the audit filter runs first in the chain — but it's fragile. If any other plugin (or a future change to the security plugin itself) also uses Integer.MIN_VALUE, the relative ordering between them becomes undefined (depends on registration order, which isn't guaranteed across restarts or plugin load sequences).

For an audit filter, running early makes sense — you want to capture the request before other filters modify or reject it. But "first" and "as negative as mathematically possible" aren't the same thing.

Suggestion — use a named constant with a documented rationale and a less extreme value:
```java
  /**
   * Run before authentication/authorization filters so audit captures the
   * original request even if a later filter rejects it. The existing
   * SecurityFilter uses order 0; we need to be well below that.
   */
  private static final int AUDIT_FILTER_ORDER = -200;

  @Override
  public int order() {
      return AUDIT_FILTER_ORDER;
  }
```
This leaves room for other "must run early" filters without colliding, and makes the intent explicit.

src/main/java/org/opensearch/security/filter/AuditTransportInterceptor.java
```diff
@@ -0,0 +1,173 @@
...
+        String auditIndexPrefix
+    ) {
+        this.auditLog = auditLog;
+        this.clusterService = clusterService;
+        this.threadPool = threadPool;
+        this.filter = filter;
+        this.auditIndexPrefix = auditIndexPrefix;
+
+        // Skip internal cluster noise by default
+        this.ignoreActionsMatcher = WildcardMatcher.from("internal:*", "cluster:monitor/*", "indices:monitor/*");
```
DarshitChanpura: can we make this configurable, something like:
```java
  // SecuritySettings.java
  public static final Setting<List<String>> AUDIT_TRANSPORT_IGNORE_ACTIONS = Setting.listSetting(
      AUDIT_CONFIG_PREFIX + "transport_ignore_actions",
      List.of("internal:*", "cluster:monitor/*", "indices:monitor/*"),
      Function.identity(),
      Setting.Property.NodeScope,
      Setting.Property.Dynamic
  );

  // AuditTransportInterceptor constructor
  this.ignoreActionsMatcher = WildcardMatcher.from(
      AUDIT_TRANSPORT_IGNORE_ACTIONS.get(settings)
  );
```

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
```diff
@@ -0,0 +1,309 @@
...
+        }
+
+        // Skip ignored requests (matches action name or request class name)
+        if (filter.isRequestAuditDisabled(action) || filter.isRequestAuditDisabled(request.getClass().getSimpleName())) {
+            chain.proceed(task, action, request, listener);
+            return;
+        }
+
+        // Bulk request handling — log each sub-operation separately
+        if (filter.shouldResolveBulkRequests() && (request instanceof BulkShardRequest || request instanceof BulkRequest)) {
```
DarshitChanpura: nit (performance): In the bulk-resolution path, logRequestAudit(msg) is called per bulk item, and the downstream AbstractAuditLog will discard each message if REQUEST_AUDIT is in the disabled categories. That's correct behavior — but if an operator has REQUEST_AUDIT disabled and resolve_bulk_requests: true, you're allocating and populating a full AuditMessage per item in the bulk (potentially thousands per _bulk call) only to immediately throw them away.

Short-circuit before the loop:
```java
if (filter.shouldResolveBulkRequests()
        && !filter.getDisabledCategories().contains(AuditCategory.REQUEST_AUDIT)
        && (request instanceof BulkShardRequest || request instanceof BulkRequest)) {
```

Cheap check, avoids hot-path allocation pressure on large bulk ingests when audit is partially disabled.

src/integrationTest/java/org/opensearch/security/StandaloneAuditFilterFeaturesTest.java
```diff
@@ -0,0 +1,782 @@
...
+                ConfigConstants.OPENDISTRO_SECURITY_AUDIT_RESOLVE_BULK_REQUESTS,
+                false
+            )
+        )
+        .sslOnly(true)
+        .build();
+
+    @Rule
+    public AuditLogsRule auditLogsRule = new AuditLogsRule();
+
```
DarshitChanpura, line 154: we should add a dynamic test that adds a cert DN to ignore_users at runtime and confirms suppression takes effect immediately.

src/integrationTest/java/org/opensearch/security/StandaloneAuditSinksTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+    // --- Webhook sink test infrastructure ---
+
+    private static HttpServer webhookServer;
+    private static int webhookPort;
+    private static final AtomicReference<String> capturedWebhookBody = new AtomicReference<>();
+    private static final CountDownLatch webhookLatch = new CountDownLatch(1);
+
+    @BeforeClass
+    public static void startWebhookServer() throws Exception {
```
DarshitChanpura: this seems like dead code. Can you verify if its being used?

DarshitChanpura (commented):
Thanks @Taiwo435. This is a solid first PR towards getting Audit Logging to work with SSL-only mode.

I've left a few comments around potential bugs, perf improvement and nits.

src/main/java/org/opensearch/security/auditlog/impl/AuditCategory.java
Taiwo435: TRANSPORT_AUDIT category is for shard-level transport actions (e.g., bulk[s][p], search[phase/query]). In a multi-node cluster these are node-to-node messages; in single-node they're local shard deliveries through the same transport path.

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
Taiwo435: yes we can, i'll do that

src/main/java/org/opensearch/security/compliance/ComplianceConfig.java
Taiwo435: noted

src/integrationTest/java/org/opensearch/security/StandaloneAuditFilterFeaturesTest.java
Taiwo435: noted

src/integrationTest/java/org/opensearch/security/StandaloneAuditSinksTest.java
Taiwo435: Yes, it's dead code. I originally wrote it to spin up a real HTTP server and verify the webhook actually receives the POST, but ran into an ordering issue because the cluster @ClassRule needs the webhook URL at build time, but the server port is only known after it starts. I pivoted to StandaloneAuditWebhookSinkTest which just proves the sink initializes and doesn't crash when the endpoint is unreachable. The actual HTTP POST serialization is covered by the existing WebhookAuditLogTest unit test. I'll remove it.

src/main/java/org/opensearch/security/filter/AuditTransportInterceptor.java
Taiwo435: We could but the transport interceptor already reads from AuditConfig.Filter.isRequestAuditDisabled() which is dynamically configurable via ignore_requests. The hardcoded ignoreActionsMatcher was
  an additional non-configurable baseline to guarantee internal cluster noise (internal:*, cluster:monitor/*, indices:monitor/*) is always suppressed.                        

Unless we want ignore_requests to remain a unified setting that applies to all actions regardless of which layer they appear in kinda similar to the unified disabled_categories setting I implemented earlier, where the rationale was that users do not necessarily know whether an action is REST-level or transport-level to suppress it. In that case I'd just remove the hardcoded matcher and let ignore_requests handle everything uniformly. The new transport_ignore_actions setting makes sense if we want to give users independent control per layer.

src/main/java/org/opensearch/security/support/SecuritySettings.java
Taiwo435: Default value should be ["kibanaserver"] not emptyList(), ComplianceConfig.from(Settings) uses AuditConfig.DEFAULT_IGNORED_USERS (["kibanaserver"]) as the default. I'll align them to use ["kibanaserver"] as the default in the Setting definition to match existing behavior.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
Taiwo435: I'll add the fix (once-per-interval warning), replacing the empty catch block with debug-level full stack trace and warn-level message for production visibility.

src/main/java/org/opensearch/security/compliance/ComplianceConfig.java
Taiwo435: i'll make the change and combine this single-pass approach with the atomic swap from your earlier concurrency suggestion. One loop builds both maps simultaneously, then swaps atomically into the WatchedReadState holder before invalidating the cache.

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
Taiwo435: noted

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
Taiwo435: Applied both, constructor logs a WARN at startup if the prefix is null/empty (early visibility for operators), and an isAuditIndex() helper at runtime that returns false for null/empty prefix so the guard degrades to "log everything" rather than silently suppressing all events. Covers both misconfiguration at boot and the (unlikely) case of someone clearing the setting at runtime.

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
Taiwo435: noted

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
Taiwo435: noted

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
Taiwo435: i'll go with the second approach and create a shred utility in the filter package since both classes live there

src/main/java/org/opensearch/security/filter/AuditTransportInterceptor.java
DarshitChanpura: Yes, let's have ignore_requests handle it in that case. We want to avoid hardcoded suppressors for now.

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
Taiwo435: Concurrency fix for AuditConfig.Filter dynamic setters:

1. Added volatile to ignoredUrlParams/ignoredUrlParamsMatcher — were invisible to reader threads
2.  Reordered setIgnoredAuditUsers() and setIgnoredAuditRequests() — build set + matcher in locals first, write matcher (the operationally-read field) last

Used reorder instead of a holder object because the Set<String> field is only for Jackson serialization and runtime filtering only reads the WildcardMatcher. No structural changes, no serialization risk.

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
```diff
@@ -0,0 +1,330 @@
...
+        String effectiveUser = user != null ? user.getName() : principal;
+        if (effectiveUser != null && filter.isAuditDisabled(effectiveUser)) {
+            chain.proceed(task, action, request, listener);
+            return;
+        }
+
+        // Skip ignored requests (matches action name or request class name)
+        if (filter.isRequestAuditDisabled(action) || filter.isRequestAuditDisabled(request.getClass().getSimpleName())) {
+            chain.proceed(task, action, request, listener);
+            return;
```
DarshitChanpura, line 158: AuditActionFilter.java — non-bulk path doesn't short-circuit on disabled REQUEST_AUDIT

  The bulk path now correctly bails early when REQUEST_AUDIT is disabled:

  if (filter.shouldResolveBulkRequests()
      && !filter.getDisabledCategories().contains(AuditCategory.REQUEST_AUDIT)
      && (request instanceof BulkShardRequest || request instanceof BulkRequest)) {

  ...but the single-request path below it doesn't. It builds the full AuditMessage (remote address, indices, wildcard resolution via resolver.concreteIndexNames, headers, request body) and only then calls auditLog.logRequestAudit(msg), relying on AbstractAuditLog to discard it downstream. When REQUEST_AUDIT is disabled, that's wasted allocation + an
  index-resolution call on every request.

  This also leaves it inconsistent with two things it should mirror:
  - AuditTransportInterceptor gates on disabled categories up front (getDisabledTransportCategories / getDisabledCategories for TRANSPORT_AUDIT) before building anything.
  - The existing FGAC pipeline — AbstractAuditLog.checkTransportFilter / checkRestFilter both check disabled categories before constructing the event.

  Suggest adding the same early gate before the non-bulk build:

  if (filter.getDisabledCategories().contains(AuditCategory.REQUEST_AUDIT)) {
      chain.proceed(task, action, request, listener);
      return;
  }

  Placed after the ignore-users / ignore-requests skips, this makes the REST path consistent with both its transport sibling and the bulk short-circuit, and avoids hot-path work when the category is off.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -778,7 +958,62 @@ public List<RestHandler> getRestHandlers(
...
+                                    log.debug("Failed to extract client certificate identity", e);
+                                }
+                            }
+                        }
+                    }
+
+                    // Store REST headers in ThreadContext for audit logging
+                    Map<String, List<String>> headers = request.getHeaders();
+                    if (headers != null && !headers.isEmpty()) {
+                        threadContext.putTransient(ConfigConstants.SECURITY_AUDIT_REST_HEADERS, headers);
```
DarshitChanpura, line 1006: we should not be adding all header right? Can we filter here instead of downstream?

src/integrationTest/java/org/opensearch/security/StandaloneAuditDynamicFilterSettingsTest.java
```diff
@@ -0,0 +1,517 @@
...
+
+    // =====================================================================
+    // ignore_users — add at runtime
+    // =====================================================================
+
+    @Test
+    public void shouldIgnoreUsersAddedAtRuntime() {
+        try (TestRestClient client = cluster.getRestClient()) {
+            // Add a wildcard ignore pattern at runtime
+            client.putJson("_cluster/settings", "{\"persistent\": {\"plugins.security.audit.config.ignore_users\": [\"*\"]}}");
```
DarshitChanpura: This test doesn't actually verify runtime suppression. It sets ignore_users: ["*"], but in SSL-only mode with no client cert the effective user is null, so — as the comment on the next line admits — events still log. The only assertion here is assertStatusCode(200), which just confirms the setting was accepted, not that a matching user's events get dropped.

  This was the ask from the earlier thread: add a specific cert DN to ignore_users at runtime and confirm suppression takes effect immediately. Could we mirror the shouldIgnoreRequestsAddedAtRuntime test right below (which correctly drives a request and asserts assertExactlyScanAll(0, ...)), but using an mTLS client presenting a known DN — set ignore_users to that DN, issue a request, and assert zero REQUEST_AUDIT events for it? That actually exercises the identified-user suppression path instead of the null-user no-op.

src/integrationTest/java/org/opensearch/security/StandaloneAuditFilterFeaturesTest.java
DarshitChanpura: https://github.com/opensearch-project/security/pull/6304/changes#r3618070623

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
DarshitChanpura: sounds good.

DarshitChanpura (commented):
Left a couple more comments.

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
Taiwo435: noted

src/integrationTest/java/org/opensearch/security/StandaloneAuditDynamicFilterSettingsTest.java
Taiwo435: got it

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
Taiwo435: We have a dynamic cluster setting for exclude_sensitive_headers that can be toggled at runtime via PUT _cluster/settings without a restart. If we filter upstream at the REST wrapper, we'd lose the raw headers permanently — so if a user disables the exclusion at runtime (e.g., for debugging), the headers are already gone from ThreadContext and can't be recovered. Keeping the filtering downstream in the audit filter means we always have the full set available and respect the live config value at the moment the audit event is constructed. 

The security risk of holding them in ThreadContext briefly is low — they're in-process memory only, request-scoped, and never persisted or transmitted unfiltered.

The only risk is that the Authorization header (containing credentials) sits in memory slightly longer than necessary (from REST wrapper → AuditActionFilter, a few milliseconds). But It's never written to the audit log (filtered before logging). It's never serialized or transmitted. Any process that can read JVM memory already has full access to everything anyway.

DarshitChanpura: sounds good. Let's worth capturing in perf test.

DarshitChanpura (approved):
LGTM!

src/main/java/org/opensearch/security/support/SecuritySettings.java
```diff
@@ -57,4 +62,181 @@ public class SecuritySettings {
         Setting.Property.Dynamic,
         Setting.Property.Sensitive
     );
+
+    public static final Setting<Boolean> AUDIT_ENABLED_SETTING = Setting.boolSetting(
```
cwperks, line 66: @DarshitChanpura @Taiwo435 Please add `Setting.Property.Sensitive` in the list of settings properties for all of these. When `Setting.Property.Sensitive` is present, the setting is only toggle-able by a user mapped to one of the roles from `plugins.security.restapi.roles_enabled: ["all_access","xyz_role"]` (colloquially referred to as a "security admin"). Its crude, but does at least move towards some semblance of fine grained controls for cluster settings. All of the settings in here should only be toggle-able by a security admin of the cluster.

Even in SSL-Only mode we theoretically could enforce that a cluster admin is the user that presents the admin certificate. So even when in SSL-Only mode (i.e. no users and roles) we can still differentiate between a trusted and non-trusted caller by presence of that certificate.

src/main/java/org/opensearch/security/filter/AuditTransportInterceptor.java
```diff
@@ -0,0 +1,187 @@
...
+                // Skip if TRANSPORT_AUDIT is disabled
+                if (!filter.getDisabledTransportCategories().contains(AuditCategory.TRANSPORT_AUDIT)
+                    && !filter.getDisabledCategories().contains(AuditCategory.TRANSPORT_AUDIT)) {
+                    // Skip ignored requests (action or class name)
+                    if (!filter.isRequestAuditDisabled(action) && !filter.isRequestAuditDisabled(request.getClass().getSimpleName())) {
+                        // Skip ignored users
+                        String principal = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_SSL_PRINCIPAL);
+                        User user = threadPool.getThreadContext().getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER);
+                        String effectiveUser = user != null ? user.getName() : principal;
+                        if (effectiveUser == null || !filter.isAuditDisabled(effectiveUser)) {
```
cwperks, line 110: IMO `!filter.isAuditDisabled` doesn't read naturally. Prefer `filter.isAuditEnabled()`

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
```diff
@@ -0,0 +1,353 @@
...
+            msg.addRemoteAddress(remoteAddress);
+
+            // Action name
+            msg.addPrivilege(action);
+
+            // Request type
+            msg.addRequestType(request.getClass().getSimpleName());
+
+            // Target indices
+            if (request instanceof IndicesRequest) {
```
cwperks, line 225: @DarshitChanpura @Taiwo435 Can we use the IndexResolverReplacer here? Can we please make sure to test with both legacy and v4 privilege evaluation code paths.

src/main/java/org/opensearch/security/compliance/ComplianceReadIndexSearcherWrapper.java
```diff
@@ -0,0 +1,293 @@
...
+
+/**
+ * A lightweight reader wrapper that provides compliance read tracking
+ * (COMPLIANCE_DOC_READ) without DLS/FLS. For use in non-FGAC modes
+ * where SecurityFlsDlsIndexSearcherWrapper is not registered.
+ *
+ * Reuses FieldReadCallback directly with FieldMaskingRule.ALLOW_ALL
+ * (no field masking since no FLS is active).
+ */
+public class ComplianceReadIndexSearcherWrapper implements CheckedFunction<DirectoryReader, DirectoryReader, IOException> {
```
cwperks, line 41: Is there any opportunity to make this more generic in case the same code can be re-used?

src/main/java/org/opensearch/security/support/SecuritySettings.java
Taiwo435: noted

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
Taiwo435: IndexResolverReplacer is marked Deprecated, the class says new code should use ResolvedIndices from ActionRequestMetadata. I'm thinking of using that instead over what I currently have (IndexNameExpressionResolver) because actionRequestMetadata.resolvedIndices() is already computed by core before the filter runs so no duplicate resolution, handles data streams/aliases/remote indices out of the box, and works with both legacy and v4 evaluation paths since it's framework-level. Would that work?

src/main/java/org/opensearch/security/filter/AuditTransportInterceptor.java
Taiwo435: Added isAuditEnabled(String user) to AuditConfig.Filter and updated the AuditTransportInterceptor call site to use it.

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
cwperks: It will need to work with both code paths so let's make it work with both.

Taiwo435: Done, switched to actionRequestMetadata.resolvedIndices(). It's populated by core before any plugin filter runs, so it's evaluator-agnostic.

src/main/java/org/opensearch/security/compliance/ComplianceReadIndexSearcherWrapper.java
Taiwo435: Yes, definitely possible.

PR #6305, opened 2026-07-14
improve the error message if unresolved user attrs are in DLS

rursprung:
### Description
if no user attributes are present then the error message looked weird, as it just didn't list anything and it wasn't clear whether that was a bug or if there really were no custom attributes available.

this now enhances it to explicitly show "[none]" if no user attributes are available.

### Issues Resolved
n/a

### Testing
unit test

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/privileges/dlsfls/DocumentPrivilegesTest.java
+++ b/src/integrationTest/java/org/opensearch/security/privileges/dlsfls/DocumentPrivilegesTest.java
@@ -67,6 +67,7 @@ import org.opensearch.test.framework.TestSecurityConfig;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
 import static org.opensearch.security.util.MockIndexMetadataBuilder.dataStreams;
 import static org.opensearch.security.util.MockIndexMetadataBuilder.indices;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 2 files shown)


rursprung:
the test failure seems to be an unrelated false reject, probably a flaky test or unstable infrastructure

rursprung:
for the AI review: the "[none]" is intentional to show that it isn't an attribute named `none` but really nothing. maybe we could write it completely differently if there are no available attributes, though. not sure how important that is...

cwperks:
@rursprung I'll check with others to see if this is something we can do. I suppose a user can approach their admin to say they are having issues with querying, but a non-admin user doesn't have any way of rectifying an error like this themselves as those attributes are taken from the external IdP (or added by an admin on internal user). That being said, I am usually open to introducing some changes behind a feature flag so that people in the community can use them even if its not default behavior.

rursprung:
> @rursprung I'll check with others to see if this is something we can do. I suppose a user can approach their admin to say they are having issues with querying, but a non-admin user doesn't have any way of rectifying an error like this themselves as those attributes are taken from the external IdP (or added by an admin on internal user). That being said, I am usually open to introducing some changes behind a feature flag so that people in the community can use them even if its not default behavior.

i'm not sure i follow? i'm just slightly improving an existing error message, i'm not sure why you'd need a feature flag for this?

cwperks:
> i'm not sure i follow? i'm just slightly improving an existing error message, i'm not sure why you'd need a feature flag for this?

In case any attribute names could be considered sensitive info? 

In general, error messages are meant to be actionable and the user receiving this error can't take action themselves (although they can communicate with cluster admin/IdP admin what error they receive). 

It may not be a problem. @nibix wdyt?

rursprung:
> > i'm not sure i follow? i'm just slightly improving an existing error message, i'm not sure why you'd need a feature flag for this?
> 
> In case any attribute names could be considered sensitive info?
> 
> In general, error messages are meant to be actionable and the user receiving this error can't take action themselves (although they can communicate with cluster admin/IdP admin what error they receive).
> 
> It may not be a problem. @nibix wdyt?

but the exception already contains the variable names today? the only thing i'm adding here is that i print `[none]` if there are none available. if any are available they're already being printed now.

cwperks:
> but the exception already contains the variable names today? the only thing i'm adding here is that i print [none] if there are none available. if any are available they're already being printed now.

ah my bad. yes the changes LGTM. 

I had previously thought the error message only contained the names of the unresolved attributes from the DLS clause itself...not from the user.

cwperks (approved):

PR #6307, opened 2026-07-14
Add support of ./gradlew run task to allow running the plugin locally

### LINKED ISSUE #6306: [BUG] `./gradlew run` doesn't work
**What is the bug?**
`./gradlew run` doesn't do anything

`./gradlew run`

it starts a cluster with the currently checked out version of the security plugin

n/a

n/a

acc. to @reta running `./gradlew run -PinstalledPlugins="['opensearch-security']"` in the OpenSearch repo should also work but seems to fail as well. but this is tangential since this ticket here is about launching an unmodified OpenSearch (possibly nightly/snapshot) with whatever code changes you've just done to the security plugin to test it while the other is about launching (a possibly modified) OpenSearch with an unmodified security plugin.

reta:
### Description
Add support of `./gradlew run` task to allow running the plugin locally. Most of the plugins support this feature in order to low the effort required to verify the plugin with OpenSearch distribution. It is also possible to run the plugin with debugger.

### Issues Resolved
Closes https://github.com/opensearch-project/security/issues/6306

### Testing
All the Gradle commands have been tested locally

### Check List

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -7,7 +7,8 @@ So you want to contribute code to OpenSearch Security? Excellent! We're glad you
     - [Native platforms](#native-platforms)
   - [Building](#building)
   - [Using IntelliJ IDEA](#using-intellij-idea)
-  - [Running integration tests](#running-integration-tests)
+  - [Running locally](#running-locally)
+  - [Running tests in the integrationTest package](#running-integration-tests)
     - [Bulk test runs](#bulk-test-runs)
     - [Checkstyle Violations](#checkstyle-violations)
   - [Authorization in REST Layer](#authorization-in-rest-layer)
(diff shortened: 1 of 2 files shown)


reta:
@rursprung wondering if you could give it a shot, thank you

cwperks (commented):
ty @reta ! Similar to https://github.com/opensearch-project/OpenSearch/pull/20393, but makes sense to live within the security repo.

reta:
> ty @reta ! Similar to [opensearch-project/OpenSearch#20393](https://github.com/opensearch-project/OpenSearch/pull/20393), but makes sense to live within the security repo.

Thanks @cwperks , yes, also mentioned https://github.com/opensearch-project/security/issues/6306#issuecomment-4970610602 , it seems like we would need minor docs update (on me)

rursprung:
> @rursprung wondering if you could give it a shot, thank you

i gave it a shot, it works, thanks!

> ty @reta ! Similar to [opensearch-project/OpenSearch#20393](https://github.com/opensearch-project/OpenSearch/pull/20393), but makes sense to live within the security repo.

yours does have the benefit of applying to all plugins rather than just the security plugin. and, if i understood correctly, yours would still run the security plugin with its own build (rather than fetching a snapshot) if used here? so maybe that'd still be the way to go?

reta:
> yours does have the benefit of applying to all plugins rather than just the security plugin. and, if i understood correctly, yours would still run the security plugin with its own build (rather than fetching a snapshot) if used here? so maybe that'd still be the way to go?

That is correct, the plugin is built (assemble task) from local, with all the local changes, only OpenSearch distribution is downloaded. That behavior / expectations are how the same task is implemented for other plugins.

cwperks (approved):

cwperks:
> yours does have the benefit of applying to all plugins rather than just the security plugin. and, if i understood correctly, yours would still run the security plugin with its own build (rather than fetching a snapshot) if used here? so maybe that'd still be the way to go?

No, that change in core would download snapshots. I submitted it for convenience for plugins when setting up CI checks for testing with security - which is notoriously cumbersome and large chunks of the same logic are copied across multiple repos. 

I definitely think it would be nice to have to be able to conveniently run security (w/ demo configuration) and multiple plugins from a single command from the core repo.

PR #6310, opened 2026-07-17
support attributes from request headers in DLS

### LINKED ISSUE #6265: [FEATURE] user attributes from HTTP headers
currently in DLS/FLS only user attributes from the authentication (e.g. claims from a JWT) can be referenced. however, sometimes other attributes are needed as well.

currently we grab all user attributes which we reference in our DLS from the JWT. we now have a use-case where my DLS needs to filter something based on an information not present in the JWT (the information is still secure even if it's not part of a signed JWT because i have the information in OpenSearch whether the value provided can be valid). i currently don't see any way of passing something like this in, so i was thinking if it'd be ok to extend the user attribute mapping to allow mapping HTTP headers to user attributes

extending the JWT is not an option in our case (the information is not known to the IdP)

n/a

rursprung:
### Description
* Category: Enhancement

with this it is now possible to specify request headers which should be
available as substitutions in DLS queries. both HTTP and gRPC headers
are supported.

the headers have to be configured under the config key
`plugins.security.unsupported.dls.allowed_request_headers`. this is a
map (the key doesn't matter) with the following content for each entry:
* `name`: the actual name of the gRPC / HTTP header (case-insensitive)
* `isMultiValue`: whether the header can have more than one value
  (default: `false`)
* `validationRegex`: a regex to ensure that the header value cannot be
  used for code injection into the DLS query
* `maxValueLength`: the maximum length each header value is allowed to
  have (default: 256)

since DLS query substitution is pure string substitution and headers,
unlike other user attributes, are fully under the control of the caller
(and thus a potential attacker) the content must be carefully validated
to ensure that it does not pose a risk. for this the `validationRegex`
needs to be used - by default it rejects all content, thus it must be
configured explicitly. it should be configured to only allow explicitly
the patterns which are absolutely needed.

due to the risk associated with this feature it is currently being
treated as unsupported/experimental and will also not be documented.

if you, dear reader, stumble upon this PR / commit please beware: only
use this is if you are absolutely sure that you know what you are doing!
you have been warned!

the substitution is done using `${attr.header.[header name]}`, e.g.
`${attr.header.x-example-header}`.
if `isMultiValue` is set to `true` then the substitution will always
contain quotes around the values and has to be treated as a list, i.e.
you should enclose it in `[]`:
```
{ "terms": { "testfield": [${attr.header.x-example-header-mv}] } }
```
while if `isMultiValue` is set to `false` then the value will be added
verbatim and you need to quote it:
```
{ "term": { "testfield": "${attr.header.x-example-header}" } }
```

to prevent the risk of DOS attacks through the header forwarding both a
limit on the length of header values has been introduced (configurable
per header, see `maxValueLength`; default: 256 characters) as well as a
global limit on the amount of headers (see
`DlsRequestHeadersUtil#MAX_HEADER_COUNT`; arbitrarily set to 256) has
been introduced.

the config options can only be set via the config file (they are
intentionally not marked as `Dynamic`) so that it has to be a clear
decision to set this. this restriction can be lifted at a later point.

once #6311 is implemented the risk posed by this feature will go down
since then it will no longer be possible to modify the query with a
crafted request (which this feature tries to prevent by having the admin
specify a regex for the validation).

the tests for gRPC have been moved to a separate test class due to the
bug reported in #6431 which leads to flaky gRPC tests with multi-node
test clusters.

### Issues Resolved
resolves #6265

### Testing
integration tests, manual tests

### Check List
- [ ] New functionality has been documented - intentionally undocumented!
- [ ] ~New Roles/Permissions have a corresponding security dashboards plugin PR~
- [ ] ~API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md)~

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/dlsfls/HeaderAttrInDlsIntegrationTest.java
+++ b/src/integrationTest/java/org/opensearch/security/dlsfls/HeaderAttrInDlsIntegrationTest.java
@@ -1,0 +1,178 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.dlsfls;
+
+import java.util.Arrays;
+import java.util.stream.Stream;
+
+import org.apache.hc.core5.http.Header;
+import org.apache.hc.core5.http.message.BasicHeader;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import org.opensearch.test.framework.TestSecurityConfig;
+import org.opensearch.test.framework.cluster.ClusterManager;
+import org.opensearch.test.framework.cluster.LocalCluster;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+import static org.opensearch.test.framework.TestSecurityConfig.Role.ALL_ACCESS;
+import static org.opensearch.test.framework.matcher.RestMatchers.isBadRequest;
+import static org.opensearch.test.framework.matcher.RestMatchers.isOk;
+
+public class HeaderAttrInDlsIntegrationTest {
+    static final String DLS_INDEX = "testindex";
+
+    private static final String DLS_INDEX_SETTINGS = """
+        {
+          "mappings": {
+            "properties": {
+              "testfield": {
+                "type": "text"
+              }
+            }
+          }
+        }""";
+
+    static final TestSecurityConfig.User ADMIN_USER = new TestSecurityConfig.User("admin").roles(ALL_ACCESS);
+
+    static final TestSecurityConfig.User SINGLE_VALUE_DLS_USER = new TestSecurityConfig.User("sv_dls_user").roles(
+        new TestSecurityConfig.Role("sv_dls_role").clusterPermissions("*")
+            .indexPermissions("*")
+            .dls("{ \"term\": { \"testfield\": \"${attr.header.x-example-header}\" } }")
+            .on(DLS_INDEX)
+    );
+
+    static final TestSecurityConfig.User MULTI_VALUE_DLS_USER = new TestSecurityConfig.User("mv_dls_user").roles(
+        new TestSecurityConfig.Role("mv_dls_role").clusterPermissions("*")
+            .indexPermissions("*")
+            .dls("{ \"terms\": { \"testfield\": [${attr.header.x-example-header-mv}] } }")
+            .on(DLS_INDEX)
+    );
+
+    static final TestSecurityConfig.User SHORT_VALUE_DLS_USER = new TestSecurityConfig.User("short_dls_user").roles(
+        new TestSecurityConfig.Role("short_dls_role").clusterPermissions("*")
+            .indexPermissions("*")
+            .dls("{ \"term\": { \"testfield\": \"${attr.header.x-short-header}\" } }")
+            .on(DLS_INDEX)
+    );
+
+    public static final LocalCluster.Builder clusterBuilder = new LocalCluster.Builder().authc(AUTHC_HTTPBASIC_INTERNAL)
+        .users(ADMIN_USER, SINGLE_VALUE_DLS_USER, MULTI_VALUE_DLS_USER, SHORT_VALUE_DLS_USER)
+        .nodeSetting("plugins.security.unsupported.dls.allowed_request_headers.x-example-header.name", "X-Example-Header")
+        .nodeSetting("plugins.security.unsupported.dls.allowed_request_headers.x-example-header.isMultiValue", "false")
+        .nodeSetting("plugins.security.unsupported.dls.allowed_request_headers.x-example-header.validationRegex", "[a-z\\-]+")
+        .nodeSetting("plugins.security.unsupported.dls.allowed_request_headers.x-example-header-mv.name", "X-Example-Header-MV")
+        .nodeSetting("plugins.security.unsupported.dls.allowed_request_headers.x-example-header-mv.isMultiValue", "true")
+        .nodeSetting("plugins.security.unsupported.dls.allowed_request_headers.x-example-header-mv.validationRegex", "[a-z\\-]+")
+        .nodeSetting("plugins.security.unsupported.dls.allowed_request_headers.x-invalid-header.name", "X-Invalid-Header")
+        .nodeSetting("plugins.security.unsupported.dls.allowed_request_headers.x-short-header.name", "X-Short-Header")
+        .nodeSetting("plugins.security.unsupported.dls.allowed_request_headers.x-short-header.validationRegex", "[a-z\\-]+")
+        .nodeSetting("plugins.security.unsupported.dls.allowed_request_headers.x-short-header.maxValueLength", "3");
+
+    @ClassRule
+    public static final LocalCluster cluster = clusterBuilder.clusterManager(ClusterManager.DEFAULT).build();
+
+    @BeforeClass
+    public static void createTestData() {
+        createTestData(cluster);
+    }
+
+    public static void createTestData(final LocalCluster cluster) {
+        try (final var client = cluster.getRestClient(ADMIN_USER)) {
+            client.putJson(DLS_INDEX, DLS_INDEX_SETTINGS);
+            client.postJson(DLS_INDEX + "/_doc?refresh=true", "{\"testfield\": \"foobar\"}");
+            client.postJson(DLS_INDEX + "/_doc?refresh=true", "{\"testfield\": \"foo\"}");
+            client.postJson(DLS_INDEX + "/_doc?refresh=true", "{\"testfield\": \"baz\"}");
+        }
+    }
+
+    @Test
+    public void testQueryWithSingleValueHeaderMatches() {
+        assertThat(runSearchAndGetTotalHitsForSingleValueHeader("foobar"), is(1));
+        assertThat(runSearchAndGetTotalHitsForSingleValueHeader("does-not-exist"), is(0));
+    }
+
+    @Test
+    public void testQueryWithMultiValueHeaderMatches() {
+        assertThat(runSearchAndGetTotalHitsForMultiValueHeader("foobar"), is(1));
+        assertThat(runSearchAndGetTotalHitsForMultiValueHeader("does-not-exist"), is(0));
+        assertThat(runSearchAndGetTotalHitsForMultiValueHeader("foobar", "foo"), is(2));
+    }
+
+    @Test
+    public void testFailureOnValueNotMatchingRegex() {
+        try (final var client = cluster.getRestClient(SINGLE_VALUE_DLS_USER)) {
+            final var response = client.get(DLS_INDEX + "/_search", new BasicHeader("X-Example-Header", "UPPERCASE-IS-INVALID"));
+            assertThat(response, isBadRequest());
+        }
+    }
+
+    @Test
+    public void testFailureOnUndefinedRegex() {
+        try (final var client = cluster.getRestClient(SINGLE_VALUE_DLS_USER)) {
+            final var response = client.get(DLS_INDEX + "/_search", new BasicHeader("X-Invalid-Header", "nothing matches here"));
+            assertThat(response, isBadRequest());
+        }
+    }
+
+    @Test
+    public void testLengthLimit() {
+        // short works
+        runSearchAndGetTotalHits(SHORT_VALUE_DLS_USER, new BasicHeader("X-Short-Header", "foo"));
+        // exceeding the limit fails it
+        try (final var client = cluster.getRestClient(SHORT_VALUE_DLS_USER)) {
+            final var response = client.get(DLS_INDEX + "/_search", new BasicHeader("X-Short-Header", "foobar"));
+            assertThat(response, isBadRequest());
+        }
+    }
+
+    @Test
+    public void testFailureOnMultipleValuesOnSingleValueConfig() {
+        try (final var client = cluster.getRestClient(SINGLE_VALUE_DLS_USER)) {
+            final var response = client.get(
+                DLS_INDEX + "/_search",
+                new BasicHeader("X-Example-Header", "a"),
+                new BasicHeader("X-Example-Header", "b")
+            );
+            assertThat(response, isBadRequest());
+        }
+    }
+
+    @Test
+    public void testFailureOnTooManyHeadersOverall() {
+        try (final var client = cluster.getRestClient(MULTI_VALUE_DLS_USER)) {
+            final var headers = Stream.iterate(new BasicHeader("X-Example-Header-MV", "a"), h -> h).limit(257).toArray(Header[]::new);
+            final var response = client.get(DLS_INDEX + "/_search", headers);
+            assertThat(response, isBadRequest());
+        }
+    }
+
+    int runSearchAndGetTotalHitsForSingleValueHeader(final String headerValue) {
+        return runSearchAndGetTotalHits(SINGLE_VALUE_DLS_USER, new BasicHeader("X-Example-Header", headerValue));
+    }
+
+    int runSearchAndGetTotalHitsForMultiValueHeader(final String... headerValues) {
+        final var headers = Arrays.stream(headerValues).map(s -> new BasicHeader("X-Example-Header-MV", s)).toArray(Header[]::new);
+        return runSearchAndGetTotalHits(MULTI_VALUE_DLS_USER, headers);
+    }
+
+    int runSearchAndGetTotalHits(final TestSecurityConfig.User user, final Header... headers) {
+        try (final var client = cluster.getRestClient(user)) {
+            final var response = client.get(DLS_INDEX + "/_search", headers);
+            assertThat(response, isOk());
+            return response.getIntFromJsonBody("/hits/total/value");
+        }
+    }
+}
--- a/src/main/java/org/opensearch/security/privileges/dlsfls/DlsRequestHeadersUtil.java
+++ b/src/main/java/org/opensearch/security/privileges/dlsfls/DlsRequestHeadersUtil.java
@@ -1,0 +1,223 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.privileges.dlsfls;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import org.opensearch.common.settings.Settings;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.security.DefaultObjectMapper;
+import org.opensearch.security.filter.SecurityRequestChannel;
+
+import tools.jackson.core.type.TypeReference;
+
+import static java.util.function.Function.identity;
+import static org.opensearch.security.support.ConfigConstants.OPENSEARCH_SECURITY_DLS_REQUEST_HEADERS;
+import static org.opensearch.security.support.ConfigConstants.OPENSEARCH_SECURITY_DLS_REQUEST_HEADERS_CONFIG;
+
+/**
+ * Utilities to handle request headers which are made available as for substitution in DLS queries.
+ * These headers must be whitelisted in the config (see {@link org.opensearch.security.support.ConfigConstants#OPENSEARCH_SECURITY_DLS_REQUEST_HEADERS_CONFIG})
+ * and are validated against the regex specified in the config. They are then stored in the thread context for usage in
+ * DLS queries and forwarded to other nodes if necessary.
+ *
+ * <p>
+ * This feature is in addition to the user attributes which are also made available for substitution, see {@link org.opensearch.security.privileges.UserAttributes}.
+ * </p>
+ */
+public class DlsRequestHeadersUtil {
+    private static final Logger log = LogManager.getLogger(DlsRequestHeadersUtil.class);
+
+    private DlsRequestHeadersUtil() {}
+
+    /**
+     * Arbitrary upper limit on the amount of supported headers to prevent a DOS attack.
+     * Since the headers are forwarded to requests to other shards sending a too big amount of headers could
+     * negatively impact other shards.
+     */
+    private static final int MAX_HEADER_COUNT = 256;
+
+    /**
+     * It is possible to reference HTTP / gRPC headers in DLS queries. To achieve this, they are extracted from the
+     * request and stored in the thread context in a manner where they will also be available if the request is
+     * forwarded.
+     */
+    public static void extractAndStoreDlsRequestHeaders(
+        final SecurityRequestChannel securityRequestChannel,
+        final ThreadContext threadContext,
+        final Settings settings
+    ) {
+        final var allowedDlsRequestHeaders = getDlsRequestHeaderSettings(settings);
+
+        final Map<String, List<String>> rawDlsRequestHeaders = securityRequestChannel.getHeaders() == null
+            ? Map.of()
+            : securityRequestChannel.getHeaders()
+                .entrySet()
+                .stream()
+                .filter(e -> allowedDlsRequestHeaders.keySet().stream().anyMatch(a -> a.equalsIgnoreCase(e.getKey())))
+                .collect(Collectors.toUnmodifiableMap(Map.Entry::getKey, Map.Entry::getValue));
+
+        final var totalHeaderCount = rawDlsRequestHeaders.values().stream().mapToInt(List::size).sum();
+
+        if (totalHeaderCount > MAX_HEADER_COUNT) {
+            throw new IllegalArgumentException(
+                String.format(
+                    "found %d headers for DLS variables which exceeds the global maximum of %d",
+                    totalHeaderCount,
+                    MAX_HEADER_COUNT
+                )
+            );
+        }
+
+        final var dlsRequestHeaders = rawDlsRequestHeaders.entrySet()
+            .stream()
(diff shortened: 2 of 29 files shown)


rursprung:
CC @nibix & @cwperks since we already discussed this in some details

if this could land in 3.8.0 that'd be great 🙏 (presuming there are no blockers)

src/integrationTest/java/org/opensearch/security/dlsfls/HeaderAttrInDlsIntegrationTest.java
```diff
@@ -0,0 +1,167 @@
...
+    static final TestSecurityConfig.User MULTI_VALUE_DLS_USER = new TestSecurityConfig.User("mv_dls_user").roles(
+        new TestSecurityConfig.Role("mv_dls_role").clusterPermissions("*").indexPermissions("*").dls(MULTI_VALUE_DLS_QUERY).on(DLS_INDEX)
+    );
+
+    @ClassRule
+    public static final LocalCluster cluster = new LocalCluster.Builder().clusterManager(ClusterManager.SINGLENODE)
+        .anonymousAuth(false)
+        .authc(AUTHC_HTTPBASIC_INTERNAL)
+        .users(ADMIN_USER, SINGLE_VALUE_DLS_USER, MULTI_VALUE_DLS_USER)
+        .nodeSetting("plugins.security.unsupported.dls.allowed_request_headers.x-example-header.name", "X-Example-Header")
```
reta, line 73: To be fair, this (to me) looks as very dangerous  and fragile feature, quite easily being misconfigured, just my opinion. In any case, the header size cap we currently use is 16KB (could be changed but this is a default). I think we should have reasonable default hard limit (256 characters?) to each allowed header, with possibility to be configured (like other settings), thanks

rursprung: if misconfigured it indeed can be dangerous. that's why it'll remain undocumented & unsupported in the first step until we gather some experience with it and possibly lay some groundwork to make it more secure (=> namely #6311)

i can add a max. length; though it is technically also possible to customise this via the regex (e.g. `^[a-z]{0,256}$` instead of just `^[a-z]+$`)

reta: >i can add a max. length; though it is technically also possible to customise this via the regex (e.g. ^[a-z]{0,256}$ instead of just ^[a-z]+$)

Correct, I was suggesting  default hard limit (256 characters), even if regex syntax supports it, so the cap is always there, thanks

rursprung: i've added `maxValueLength` as a config option with a default of 256.

i did not add a limit for how many headers can be present in total and i'm wondering now if i should add that as well? i can just hardcode a constant in `DlsRequestHeadersUtil` and check for that?

src/integrationTest/java/org/opensearch/security/dlsfls/HeaderAttrInDlsIntegrationTest.java
```diff
@@ -0,0 +1,195 @@
...
+    static final TestSecurityConfig.User MULTI_VALUE_DLS_USER = new TestSecurityConfig.User("mv_dls_user").roles(
+        new TestSecurityConfig.Role("mv_dls_role").clusterPermissions("*").indexPermissions("*").dls(MULTI_VALUE_DLS_QUERY).on(DLS_INDEX)
+    );
+
+    static final TestSecurityConfig.User SHORT_VALUE_DLS_USER = new TestSecurityConfig.User("short_dls_user").roles(
+        new TestSecurityConfig.Role("short_dls_role").clusterPermissions("*").indexPermissions("*").dls(SHORT_VALUE_DLS_QUERY).on(DLS_INDEX)
+    );
+
+    @ClassRule
+    public static final LocalCluster cluster = new LocalCluster.Builder().clusterManager(ClusterManager.SINGLENODE)
```
nibix: Let's make this test cluster have multiple nodes in order to properly test propagation via nodes

src/integrationTest/java/org/opensearch/security/dlsfls/HeaderAttrInDlsIntegrationTest.java
```diff
@@ -0,0 +1,195 @@
...
+        .anonymousAuth(false)
```
nibix: This is default, can be just removed

src/integrationTest/java/org/opensearch/security/dlsfls/HeaderAttrInDlsIntegrationTest.java
rursprung: i just went ahead and added that as well. arbitrarily set to 256 because why not ([4](https://xkcd.com/221/) felt a bit too low 😆)

src/integrationTest/java/org/opensearch/security/dlsfls/HeaderAttrInDlsIntegrationTest.java
rursprung: it already does the propagation on a single node. it goes through the transport layer every time (i was surprised to learn that when i was implementing & debugging the feature). should i also define sharding for the index so that there's a reason for the request to go to all nodes?
i guess you just want me to switch to `ClusterManager.DEFAULT`?

nibix: With single node, does it also go through `SecurityRequestHandler` and `SecurityInterceptor` ? 🤔

nibix: yes, ClusterManager.DEFAULT seems to be the most reasonable

rursprung: > With single node, does it also go through `SecurityRequestHandler` and `SecurityInterceptor` ? 🤔

yes, it does. i've spent a lot of time in there with the debugger in the last days 😆

rursprung: i am wondering whether it actually brings any benefit to switch from `SINGLENODE` to `DEFAULT` since already the single node tests the whole communication infrastructure (as written above) and - at least in my case - running `DEFAULT` locally fails (see my question on slack) + it just uses more resources without adding more benefit to the test?

rursprung:
i _think_ that the test failures which are now still happening are flaky tests? i see nodes which become unreachable (but no clear error as to why and i've seen these errors in other PRs before) and i see gradle failing to download a gradle plugin (=> network issues)

src/integrationTest/java/org/opensearch/security/dlsfls/HeaderAttrInDlsIntegrationTest.java
rursprung: i've switched to `DEFAULT` as @cwperks mentioned that there are still some differences around the handling of same-node transport compared to transport to other nodes. let's see if CI likes it 🤞

DarshitChanpura:
@rursprung Can you look into test failures:
Screenshot 2026-07-20 at 10 32 29 AM


also, is there a reason for using `.unsupported.` prefix?

rursprung:
> @rursprung Can you look into test failures: Screenshot 2026-07-20 at 10 32 29 AM
 
ok, i can see the failure in my test: i don't do `} finally { channel.shutdown(); }` in the GRPC test (and didn't notice that because i only ever ran the one, so it shut down afterwards anyway). i've added that.

i can indeed reproduce that `WhoAmITests` fails for me as well (though locally i have to run it as a single-node cluster; see slack for my problems with multi-node clusters in tests). `grantedPrivilegesMessages` has 5 entries on `main` but only 1 on my branch... and i have 0 explanation as to why that's the case. i don't see anything in my code which would indicate that it could cause this and debugging it i didn't see anything obvious either. and there are no exceptions in the logs as far as i can tell.
so for now i'm at a complete loss. also, i'm officially off. if anyone understand else has time to dig into this so that we have a chance to land it in this release that'd be awesome!

> also, is there a reason for using `.unsupported.` prefix?

as explained in the commit-msg & PR description this feature has some security implications and for now we want to keep it unsupported/experimental and gather some experience with it first, and maybe clean up some relevant functionality (e.g. how attribute substitution works) first before we stabilise it (if ever).

nibix:
So, I had a look into the test failures in `WhoAmITests`:

The changes in this PR has the effect that virtually any transport request carries the `_opensearch_security_dls_request_headers` header.

One example, generated by one test in `WhoAmITests`:

```
{
    "audit_cluster_name": "local_cluster_1",
    "audit_transport_headers": {
        "_opensearch_security_dls_request_headers": ""
    },
    "audit_node_name": "data_0",
    "audit_trace_task_id": "gpkyn_ViTr6T3l60GJ2Bow:77",
    "audit_transport_request_type": "GetSettingsRequest",
    "audit_category": "GRANTED_PRIVILEGES",
    "audit_request_origin": "REST",
    "audit_node_id": "gpkyn_ViTr6T3l60GJ2Bow",
    "audit_request_layer": "TRANSPORT",
    "@timestamp": "2026-07-22T14:26:02.652+00:00",
    "audit_format_version": 4,
    "audit_request_remote_address": "127.0.0.1",
    "audit_request_privilege": "indices:monitor/settings/get",
    "audit_node_host_address": "127.0.0.1",
    "audit_request_effective_user": "audit_log_verifier",
    "audit_node_host_name": "127.0.0.1"
}
```

This particular audit log message carried no headers before the change:

This influences a condition in `AuditLogsRule`:

https://github.com/opensearch-project/security/blob/06a098aa6e1b7c22cc5f3e6fa4ea860fc890e285/src/integrationTest/java/org/opensearch/test/framework/audit/AuditLogsRule.java#L147-L168

I do not really understand why the presence of `audit_transport_headers` influences into which bucket the messages are sorted.

But, this does not matter so much. Rather, IMHO we should avoid having any side effects of the feature when the feature is not in use. In this case, it is the presence of the empty valued transport header. We should strive to have such observable changes only when the feature is in use.

Thus, I would recommend to modify the code to only write such headers when the settings reflect the use of it.

rursprung:
thanks for the investigation @nibix!
i've updated the PR so that it now only adds the header if it really has any content. i did not expect this kind of side-effect of an empty header. the test now runs through with this change in place.

rursprung:
odd that `HeaderAttrInDlsIntegrationTest.testGrpcChannel` fails only on ubuntu-latest with JDK 25 but not any of the other three builds in the matrix?

```
2026-08-10T12:50:53.4658382Z 2026-08-10 12:50:53 opensearch[cluster_manager_0][grpc][T#4] ERROR SecurityFilter:376 - No user found for indices:data/read/search from null GRPC via null {_opensearch_security_dls_request_headers=[{"class":"SingleValueDlsRequestHeader","name":"x-example-header","value":"foobar"}]}
2026-08-10T12:50:53.5668078Z 
2026-08-10T12:50:53.5672498Z > Task :integrationTest
2026-08-10T12:50:53.5673608Z 
2026-08-10T12:50:53.5674368Z HeaderAttrInDlsIntegrationTest > testGrpcChannel FAILED
2026-08-10T12:50:53.5676859Z     io.grpc.StatusRuntimeException: INTERNAL: OpenSearchSecurityException[No user found for indices:data/read/search]; details={"error":{"root_cause":[{"type":"security_exception","reason":"No user found for indices:data/read/search"}],"type":"security_exception","reason":"No user found for indices:data/read/search"},"status":500}
2026-08-10T12:50:53.5679582Z         at app//io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:368)
2026-08-10T12:50:53.5680682Z         at app//io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:349)
2026-08-10T12:50:53.5681759Z         at app//io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:174)
2026-08-10T12:50:53.5683265Z         at app//org.opensearch.protobufs.services.SearchServiceGrpc$SearchServiceBlockingStub.search(SearchServiceGrpc.java:298)
2026-08-10T12:50:53.5684942Z         at app//org.opensearch.security.grpc.GrpcHelpers.doMatchAll(GrpcHelpers.java:334)
2026-08-10T12:50:53.5686607Z         at app//org.opensearch.security.dlsfls.HeaderAttrInDlsIntegrationTest.testGrpcChannel(HeaderAttrInDlsIntegrationTest.java:174)
2026-08-10T12:50:53.8655426Z 
```
i'll have a look tomorrow

src/integrationTest/java/org/opensearch/security/dlsfls/HeaderAttrInDlsIntegrationTest.java
```diff
@@ -0,0 +1,198 @@
...
+            "properties": {
+              "testfield": {
+                "type": "text"
+              }
+            }
+          }
+        }""";
+    private static final String SINGLE_VALUE_DLS_QUERY = "{ \"term\": { \"testfield\": \"${attr.header.x-example-header}\" } }";
+    private static final String MULTI_VALUE_DLS_QUERY = "{ \"terms\": { \"testfield\": [${attr.header.x-example-header-mv}] } }";
+    private static final String SHORT_VALUE_DLS_QUERY = "{ \"term\": { \"testfield\": \"${attr.header.x-short-header}\" } }";
```
nibix: That's just a nit and maybe just a personal pecularity, but IMHO string values which are used once (i.e., not re-used) should not be constants. This helps to avoid unnecessary clutter.

src/integrationTest/java/org/opensearch/security/dlsfls/HeaderAttrInDlsIntegrationTest.java
```diff
@@ -0,0 +1,198 @@
...
+        assertThat(runSearchAndGetTotalHitsForMultiValueHeader("foobar"), is(1));
+        assertThat(runSearchAndGetTotalHitsForMultiValueHeader("does-not-exist"), is(0));
+        assertThat(runSearchAndGetTotalHitsForMultiValueHeader("foobar", "foo"), is(2));
+    }
+
+    @Test
+    public void testFailureOnValueNotMatchingRegex() {
+        try (final var client = cluster.getRestClient(SINGLE_VALUE_DLS_USER)) {
+            final var response = client.get(DLS_INDEX + "/_search", new BasicHeader("X-Example-Header", "UPPERCASE-IS-INVALID"));
+            assertThat(response.getStatusCode(), is(HttpStatus.SC_BAD_REQUEST));
```
nibix: woudl recommed to use the matchers from `RestMatchers` here and in the other similar cases. This gives much richer error messages in case of assertion errors and thus facilitates debugging.

https://github.com/opensearch-project/security/blob/981dd864f5bd3b9be3043eddb58bc6ed9ebfd498/src/integrationTest/java/org/opensearch/test/framework/matcher/RestMatchers.java#L22-L44

nibix (commented):
I have rerun the tests, it looks like they are completely green now. Added just some minor nitpicks.

src/integrationTest/java/org/opensearch/security/dlsfls/HeaderAttrInDlsIntegrationTest.java
rursprung: done! though i've kept `DLS_INDEX_SETTINGS` since i think it's easier to see right on the top of the test the structure of the index rather than having to find the spot where it gets created.

src/integrationTest/java/org/opensearch/security/dlsfls/HeaderAttrInDlsIntegrationTest.java
rursprung: thanks, i wasn't aware of `RestMatchers`! adapted.

rursprung:
> I have rerun the tests, it looks like they are completely green now. Added just some minor nitpicks.

awesome, thanks! i've incorporated fixes for your findings and rebased. the test runs fine locally => should be good now (i guess any CI failures will just be flaky tests again)

src/main/java/org/opensearch/security/privileges/dlsfls/DlsRequestHeadersUtil.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        log.debug(
+            "found {} headers with a total of {} values which matched one of the {} allowlisted headers for DLS variables: {}",
+            dlsRequestHeaders::size,
+            () -> totalHeaderCount,
+            allowedDlsRequestHeaders::size,
+            rawDlsRequestHeaders::keySet
+        );
+        log.trace("allowlisted headers: {}", () -> allowedDlsRequestHeaders);
```
nibix: I dont really get the need for using lambdas in the log arguments. Usually, one uses lambdas to avoid costly computations if the log level is turned off. In these cases, however if I understand it correctly, dropping the lambda here and just passing the parameter value would actually reduce the number of created objects.

src/main/java/org/opensearch/security/privileges/dlsfls/DlsRequestHeadersUtil.java
```diff
@@ -0,0 +1,210 @@
...
+import org.opensearch.security.filter.SecurityRequestChannel;
+
+import tools.jackson.core.type.TypeReference;
+
+import static java.util.function.Function.identity;
+import static org.opensearch.security.support.ConfigConstants.OPENSEARCH_SECURITY_DLS_REQUEST_HEADERS;
+import static org.opensearch.security.support.ConfigConstants.OPENSEARCH_SECURITY_DLS_REQUEST_HEADERS_CONFIG;
+
+/**
+ * Utilities to handle request headers which are made available as attributes for substitution in DLS queries.
```
nibix: I think this comment should be made a bit more precise to avoid that this is confused with user attributes.  This should reference the config option which enables this and describe a bit the parallel nature of these attributes to user attributes.

nibix (approved):
Thank you. I have added two more tiny remarks (sorry!), but still the current state is already approvable.

src/main/java/org/opensearch/security/privileges/dlsfls/DlsRequestHeadersUtil.java
rursprung: i intentionally did it like this to minimize any possible performance impact (the alternative would've been to wrap the calls in `isDebugEnabled` / `isTraceEnabled` checks, which is a bit ugly since `debug`, `trace` & co. already do this internally)

i have now however removed the lambdas.

nibix: Ahh, I was actually just referring to the `() -> totalHeaderCount` lambda, not the `rawDlsRequestHeaders::keySet`. 

I think mostly it's fine to make the calls. You can argue about `keySet` because this will create a small heap allocation in some of the cases. But it will be tiny anyway.

Still, JFTR, I personally would leverage `isDebugEnabled()` and co for this purpose, the boilerplate is totally tolerable IMHO.

rursprung:
CI builds failed due to an infrastructure issue (i rebased to try to trigger them again, but that led to the same problem). i've [reported it on slack](https://opensearch.slack.com/archives/C04UTNM338A/p1787218841849159)

nibix:
Retriggered the jobs ...

nibix:
The CI now seems to be completely stuck, some jobs just seem to be hanging. Unfortunately I cannot retrigger jobs in this state.

Let's maybe wait a bit whether these will wake up again. Otherwise, @rursprung maybe force-push once again to retrigger the whole CI.

rursprung:
> maybe force-push once again to retrigger the whole CI.

done, though there are flaky tests again.

final results: one `test` run failed with the usual:
> java.io.IOException: Failed to load eclipse jdt formatter: java.lang.RuntimeException: java.net.SocketTimeoutException: timeout

and one `integration-tests` run with this:
> 2026-08-20T13:12:19.7276769Z Tests with failures:
2026-08-20T13:12:19.7277561Z  - org.opensearch.security.dlsfls.HeaderAttrInDlsIntegrationTest.testGrpcChannel
2026-08-20T13:12:19.7279069Z  - org.opensearch.security.StandaloneAuditSinksTest.internalOpenSearchSinkShouldCreateAuditIndex
2026-08-20T13:12:19.7280476Z  - org.opensearch.security.privileges.ApiTokenTest.testFlushCacheReloadsFromIndex

in both cases the other matrix entries worked fine => these are flaky

i checked the logs for `dlsfls.HeaderAttrInDlsIntegrationTest.testGrpcChannel` since i introduced it with this PR, but don't see why it'd fail with this (esp. because it worked fine on windows & both jdk21 runs and it runs fine locally and ran fine in a previous CI run before i fixed the nit-picking, which was unrelated / javadocs):

> 2026-08-20T13:11:49.8335783Z HeaderAttrInDlsIntegrationTest > testGrpcChannel FAILED
2026-08-20T13:11:49.8352078Z     io.grpc.StatusRuntimeException: INTERNAL: OpenSearchSecurityException[No user found for indices:data/read/search]; details={"error":{"root_cause":[{"type":"security_exception","reason":"No user found for indices:data/read/search"}],"type":"security_exception","reason":"No user found for indices:data/read/search"},"status":500}
2026-08-20T13:11:49.8354682Z         at app//io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:368)
2026-08-20T13:11:49.8355965Z         at app//io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:349)
2026-08-20T13:11:49.8357025Z         at app//io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:174)
2026-08-20T13:11:49.8358521Z         at app//org.opensearch.protobufs.services.SearchServiceGrpc$SearchServiceBlockingStub.search(SearchServiceGrpc.java:298)
2026-08-20T13:11:49.8360045Z         at app//org.opensearch.security.grpc.GrpcHelpers.doMatchAll(GrpcHelpers.java:334)
2026-08-20T13:11:49.8361847Z         at app//org.opensearch.security.dlsfls.HeaderAttrInDlsIntegrationTest.testGrpcChannel(HeaderAttrInDlsIntegrationTest.java:181)

it seems that in some cases it either doesn't do the user setup correctly or uses it wrong - either way i think that's probably an issue with the test infrastructure code or with the gRPC code causing it to be flaky.

re-running the test will probably make it pass

nibix:
> dlsfls.HeaderAttrInDlsIntegrationTest.testGrpcChannel

I think I also saw that failure during previous test runs. I am going to check whether I can identify any issue that might be causing this.

nibix:
Regarding the issue with dlsfls.HeaderAttrInDlsIntegrationTest.testGrpcChannel :

I noticed that I know very little about the GRPC implementation, so I used some help from my new virtual best friend.

So far, all GRPC tests configure the cluster to be SINGLENODE. However, the new test introduced in this PR has multiple nodes. This brings a bit of randomness into the game. 

The security plugin has the hacky GuiceHolder static singleton to get hold of certain dependencies:

https://github.com/opensearch-project/security/blob/a7f402723f6f6f5607bc9a543fe8a097b0a399cf/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L2803-L2817

Even though the comment is clearly stating it is a hack, its usage has been actually growing over time.

In the test framework, we have several nodes running in a single JVM. For an incoming GRPC request, the filter picks a backend registry instance from GuiceHolder. However, as this is a singleton, there is a 2:1 chance that it gets an instance from a wrong node. This will lead to the observed issue.

Short term fix would be: Change the GRPC test to a SINGLENODE cluster.

Long term would be somehow getting rid of the GuideHolder.

rursprung:
thanks @nibix! as discussed i've updated the test and split gRPC into a separate test class until #6431 gets resolved.

rursprung:
it'd be really great if CI wouldn't be completely and utterly broken... 🫤 
hopefully #6438 can sort this out

DarshitChanpura (commented):
Traced the substitution path end to end; the `validationRegex` is enforced at every client ingress and I don't see a bypass. A few minor points:

- `extractAndStoreDlsRequestHeaders`: the `MAX_HEADER_COUNT` check runs after every matched value has already been regex/length-validated, so it doesn't bound that work. Moving the count check ahead of the `toDlsRequestHeader` mapping would match the stated DOS-protection intent.
- `MultiValueDlsRequestHeader#serialize`: `parallelStream()` over a few header values adds fork-join overhead with no benefit — `stream()` is enough.
- Minor consistency: the new header/transient uses the `_opensearch_security_` prefix, whereas the other DLS/FLS transport headers use `_opendistro_security_`, which is the prefix the request-header guards (`HTTPHelper`/`SSLRequestHelper#containsBadHeader`) defend. Not reachable today since only registered headers are copied into the thread context, but aligning the prefix (or extending the guards) would keep the guard coverage complete if header handling ever changes.
- Worth a one-line comment on `serialize()` noting that query-safety of the substituted value depends entirely on the configured `validationRegex`.

rursprung:
thanks for the review @DarshitChanpura!
i've included everything except for the `_opensearch_security_`/`_opendistro_security_` header thing (in `ConfigConstants`): `OPENSEARCH_SECURITY_REQUEST_HEADERS` and `SECURITY_AUDIT_REST_HEADERS` also use `OPENSEARCH_SECURITY_CONFIG_PREFIX` (i.e. `_opensearch_security_`) as their prefix. thus i feel like going back to the legacy `_opendistro_security_` for something new is wrong. but i cannot judge the impact of extending `HTTPHelper`/`SSLRequestHelper#containsBadHeader`. is it ok if i just extend that with a check for `_opensearch_security_`?

i'll push my update once that's answered (no need to stress the CI even more 😅).

btw, the failing code hygiene is unrelated to this PR and #6435 will fix it.

DarshitChanpura:
Agreed on the prefix — `_opensearch_security_` is right for new code, no need to go back to `_opendistro_security_`.

For the guard: extend only the raw-request check, `HTTPHelper.containsBadHeader` (used by both the REST and gRPC filters). It inspects the incoming request headers, and all `_opensearch_security_*` headers are internal/server-set (`OPENSEARCH_SECURITY_REQUEST_HEADERS`, `SECURITY_AUDIT_REST_HEADERS`, the new DLS header), never sent by a client — so rejecting that prefix there is safe and is exactly the injection guard.

Don't add `_opensearch_security_` to `SSLRequestHelper.containsBadHeader(threadContext, …)`: by the time that runs in `checkAndAuthenticateRequest`, the thread context legitimately contains server-set `_opensearch_security_*` headers — `OPENSEARCH_SECURITY_REQUEST_HEADERS` (SecurityRestFilter#171) and the DLS header (SecurityRestFilter#194) are both put before it — so matching the prefix there would reject legitimate requests, including this feature's own.

Worth noting this is hardening rather than a live fix: only headers registered via `getRestHeaders()` are copied into the thread context and this one isn't, so a client-supplied copy doesn't reach the forwarding path today. The raw-request guard just keeps that explicit.

rursprung:
thanks @DarshitChanpura! i've now updated the PR and added a commit with the new check (and test coverage for it, of course). i took the liberty of rewriting `HTTPHelper.containsBadHeader` to make it cleaner and - hopefully - faster (fewer `toLowerCase` calls since i skip them for the constants)

rursprung:
CI is finally happy (thanks for fixing that @DarshitChanpura & @peterzhuamazon!)

can you please use a merge commit rather than squashing this when merging the PR? this PR contains 4 atomic commits with clean commit messages; it'd be a huge shame if they're lost

PR #6312, opened 2026-07-17
Scale legacy alias resolution with bounded indices-lookup access

DarshitChanpura:
On every non-_all request, privilege evaluation streamed the entire indices lookup (a TreeMap of all index abstractions) and filtered it with a WildcardMatcher to find matching aliases. On clusters with hundreds of thousands of aliases this is ~500ms of CPU per request even when the request names a single exact alias or a simple prefix.

Resolve exact names via a point lookup and simple prefix patterns via a bounded subMap range over the naturally-ordered lookup; fall back to the original full scan only for complex patterns (?, embedded/multiple *, regex). Result sets are identical to the previous behavior (proved by a differential test).

Hardening over the original approach: robust subMap upper-bound carry handling for prefixes ending in Character.MAX_VALUE (tailMap fallback) and a null-pattern guard, both of which the naive increment/point-lookup would have broken.

Supersedes [cwperks#110](https://github.com/cwperks/security/pull/110).

### Issues Resolved
- Address #5367 

### Testing
unit testing, integration testing

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexResolverReplacerAliasScaleIntegTests.java
+++ b/src/integrationTest/java/org/opensearch/security/privileges/int_tests/IndexResolverReplacerAliasScaleIntegTests.java
@@ -1,0 +1,85 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.privileges.int_tests;
+
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import org.opensearch.test.framework.TestSecurityConfig;
+import org.opensearch.test.framework.TestSecurityConfig.Role;
+import org.opensearch.test.framework.cluster.LocalCluster;
+import org.opensearch.test.framework.cluster.TestRestClient;
+import org.opensearch.test.framework.data.TestIndex;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+import static org.opensearch.test.framework.matcher.RestMatchers.isOk;
+
+public class IndexResolverReplacerAliasScaleIntegTests {
+
+    private static final int ALIAS_COUNT = 5_000;
+    private static final int EXACT_REQUEST_COUNT = 100;
+    private static final int PREFIX_REQUEST_COUNT = 25;
+    private static final String INDEX = "alias-scale-index";
+    private static final String ALIAS_PREFIX = "alias-scale-";
+
+    private static final TestIndex TEST_INDEX = TestIndex.name(INDEX).documentCount(1).build();
+    private static final TestSecurityConfig.User READER = new TestSecurityConfig.User("alias_scale_reader").roles(
+        new Role("alias_scale_reader_role").indexPermissions("read").on(INDEX, ALIAS_PREFIX + "*")
+    );
+
+    @ClassRule
+    public static LocalCluster cluster = new LocalCluster.Builder().authc(AUTHC_HTTPBASIC_INTERNAL)
+        .users(READER, TestSecurityConfig.User.USER_ADMIN)
+        .indices(TEST_INDEX)
+        .build();
+
+    @BeforeClass
+    public static void createAliasHeavyClusterState() {
+        StringBuilder body = new StringBuilder(ALIAS_COUNT * 80).append("{\"actions\":[");
+        for (int i = 0; i < ALIAS_COUNT; i++) {
+            if (i != 0) {
+                body.append(',');
+            }
+            body.append("{\"add\":{\"index\":\"").append(INDEX).append("\",\"alias\":\"").append(ALIAS_PREFIX).append(i).append("\"}}");
+        }
+        body.append("]}");
+
+        try (TestRestClient adminClient = cluster.getAdminCertRestClient()) {
+            assertThat(adminClient.postJson("_aliases", body.toString()), isOk());
+        }
+    }
+
+    @Test
+    public void repeatedExactConcreteIndexRequestsWithManyAliases() {
+        assertRepeatedSearches(INDEX, EXACT_REQUEST_COUNT);
+    }
+
+    @Test
+    public void repeatedExactAliasRequestsWithManyAliases() {
+        assertRepeatedSearches(ALIAS_PREFIX + (ALIAS_COUNT - 1), EXACT_REQUEST_COUNT);
+    }
+
+    @Test
+    public void repeatedPrefixAliasRequestsWithManyAliases() {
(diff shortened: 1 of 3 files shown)


DarshitChanpura:
## Benchmark numbers

JMH 1.35, `AverageTime`, **3 forks × 15 measurement samples** each, reported with 99.9% confidence intervals. Compares the bounded indices-lookup path (`IndexResolverReplacer#resolveMatchingAliases`:
exact = point lookup, simple prefix = `subMap` range) against the original full-scan +
`WildcardMatcher` path. Lower is better; `speedup = fullScan / bounded`.

| aliases | pattern | bounded (µs/op) | fullScan (µs/op) | speedup |
|--------:|---------|----------------:|-----------------:|--------:|
| 1,000   | exact   | 0.081 ± 0.000   | 13.44 ± 0.28     | ~167× |
| 1,000   | prefix  | 0.274 ± 0.001   | 15.72 ± 0.23     | ~57× |
| 1,000   | complex | 19.46 ± 0.33    | 19.55 ± 0.38     | ~1× (falls back) |
| 10,000  | exact   | 0.097 ± 0.002   | 255.8 ± 35.6     | ~2,630× |
| 10,000  | prefix  | 1.96 ± 0.04     | 231.1 ± 5.2      | ~118× |
| 10,000  | complex | 371.7 ± 42.2    | 348.0 ± 51.0     | ~1× (falls back) |
| 100,000 | exact   | 0.117 ± 0.001   | 3,908 ± 96       | ~33,468× |
| 100,000 | prefix  | 18.67 ± 0.22    | 2,938 ± 314      | ~157× |
| 100,000 | complex | 4,917 ± 113     | 4,978 ± 148      | ~1× (falls back) |
| **1,000,000** | **exact** | **0.129 ± 0.003** | **89,566 ± 2,892** | **~691,838×** |
| **1,000,000** | **prefix** | **379.6 ± 5.0** | **93,063 ± 3,766** | **~245×** |
| 1,000,000 | complex | 108,930 ± 2,857 | 111,564 ± 2,254  | ~1× (falls back) |

**Takeaways**
- Exact-name resolution is effectively constant time regardless of alias count — ~90 ms of CPU per
  call at 1M aliases collapses to ~0.13 µs.
- Simple prefix resolution scales with the size of the matched range, not the whole cluster.
- Complex patterns (`?`, embedded/multiple `*`, regex) intentionally fall back to the full scan and
  are within noise of the original path — **no regression**.

The benchmark's `@Setup` asserts `bounded.equals(fullScan)` for every (aliasCount, patternClass)
**before** measuring, so a semantic divergence fails the run rather than producing fast-but-wrong
numbers. It passed all 12 combinations.

I kept this out of the source tree (it's a standalone harness, not wired into Gradle/CI, so it would
just bit-rot in `main`). The differential unit test in this PR permanently guards the correctness
property the benchmark relies on. Full harness + reproduce steps below for anyone who wants to re-run.

<details>
<summary>AliasResolveBenchmark.java + reproduce steps</summary>

```java
package org.opensearch.security.privileges.actionlevel.legacy;

import java.util.HashSet;
import java.util.Map;
import java.util.Set;
import java.util.SortedMap;
import java.util.TreeMap;
import java.util.concurrent.TimeUnit;

import org.opensearch.cluster.metadata.IndexAbstraction;
import org.opensearch.security.util.MockIndexMetadataBuilder;

import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Level;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;
import org.openjdk.jmh.annotations.Param;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;
import org.openjdk.jmh.infra.Blackhole;

@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.MICROSECONDS)
@Warmup(iterations = 5, time = 1)
@Measurement(iterations = 5, time = 1)
@Fork(3)
@State(Scope.Benchmark)
public class AliasResolveBenchmark {

    private static final String INDEX = "bench-index";
    private static final String ALIAS_PREFIX = "alias-";

    @Param({ "1000", "10000", "100000", "1000000" })
    public int aliasCount;

    // exact  -> single existing alias name
    // prefix -> "alias-99*" simple prefix
    // complex-> "alias-9?9" forces the full-scan fallback
    @Param({ "exact", "prefix", "complex" })
    public String patternClass;

    private SortedMap<String, IndexAbstraction> lookup;
    private Set<String> patterns;

    @Setup(Level.Trial)
    public void setup() {
        MockIndexMetadataBuilder builder = MockIndexMetadataBuilder.indices(INDEX);
        for (int i = 0; i < aliasCount; i++) {
            builder.alias(ALIAS_PREFIX + i).of(INDEX);
        }
        lookup = new TreeMap<>(builder.build());

        String pattern;
        switch (patternClass) {
            case "exact":
                pattern = ALIAS_PREFIX + (aliasCount - 1);
                break;
            case "prefix":
                pattern = ALIAS_PREFIX + "99*";
                break;
            case "complex":
                pattern = ALIAS_PREFIX + "9?9";
                break;
            default:
                throw new IllegalArgumentException(patternClass);
        }
        patterns = Set.of(pattern);

        // Sanity: the two paths must agree, otherwise the benchmark is meaningless.
        Set<String> bounded = IndexResolverReplacer.resolveMatchingAliases(lookup, patterns);
        Set<String> full = fullScan(lookup, patterns);
        if (!bounded.equals(full)) {
            throw new IllegalStateException("Paths disagree for " + patternClass + "@" + aliasCount + ": " + bounded + " vs " + full);
        }
    }

    @Benchmark
    public void bounded(Blackhole bh) {
        bh.consume(IndexResolverReplacer.resolveMatchingAliases(lookup, patterns));
    }

    @Benchmark
    public void fullScan(Blackhole bh) {
        bh.consume(fullScan(lookup, patterns));
    }

    /** The original (pre-optimization) implementation. */
    static Set<String> fullScan(SortedMap<String, IndexAbstraction> lookup, Set<String> patterns) {
        if (patterns.isEmpty()) {
            return new HashSet<>();
        }
        WildcardMatcher matcher = WildcardMatcher.from(patterns);
        return lookup.entrySet()
            .stream()
            .filter(e -> e.getValue().getType() == IndexAbstraction.Type.ALIAS)
            .map(Map.Entry::getKey)
            .filter(matcher)
            .collect(Collectors.toSet());
    }
}
```

Reproduce (JMH 1.35 + deps are already resolved as transitive test deps in the Gradle cache):

```bash
./gradlew -q :compileTestJava
TESTCP=<sourceSets.test.runtimeClasspath>   # export via an init script or a printClasspath task
JMH="jmh-core-1.35.jar:jmh-generator-annprocess-1.35.jar:jopt-simple-5.0.4.jar:commons-math3-3.2.jar"

javac -proc:full -processorpath "$JMH" -cp "$TESTCP:$JMH" -d out AliasResolveBenchmark.java
java -cp "out:$TESTCP:$JMH" org.openjdk.jmh.Main -jvmArgs "-Xmx8g" -rf csv -rff jmh_results.csv
```

On 2.19 the class lives in `org.opensearch.security.resolver`; change the package and import
accordingly (the method under test is identical).

</details>

shikharj05 (approved):

cwperks:
@DarshitChanpura Can you open a manual backport to 2.19?

PR #6321, opened 2026-07-21
fix(audit): Address PR review feedback from #6304

Taiwo435:
- Add Setting.Property.Sensitive to all audit and compliance dynamic settings so only security admins can toggle them at runtime.
- Add isAuditEnabled(String user) method to AuditConfig.Filter for readability; update AuditTransportInterceptor to use it.
- Replace IndexNameExpressionResolver with actionRequestMetadata .resolvedIndices() in AuditActionFilter — framework-level resolution that is evaluator-agnostic and handles data streams/aliases. Remove unused resolver field and constructor parameter.
- Refactor ComplianceReadIndexSearcherWrapper into generic reusable components: FieldReadHandler interface, ReadInterceptIndexSearcherWrapper (generic wrapper accepting predicate + handler factory), and ComplianceReadIndexSearcherWrapper as a thin compliance-specific consumer. Correctly passes ShardId from reader to handler.

### Description

Follow-up to #6304 (Standalone Audit Logging for SSL-only mode) addressing review feedback from @cwperks.
  
## Changes
  
  1. **Setting.Property.Sensitive** — Added to all audit and compliance dynamic settings so only security admins can toggle them at runtime.
  
  2. **isAuditEnabled()** — Added `isAuditEnabled(String user)` to `AuditConfig.Filter` for readability. Updated `AuditTransportInterceptor` to use it instead of `!isAuditDisabled()`.
  
  3. **Use actionRequestMetadata.resolvedIndices()** — Replaced `IndexNameExpressionResolver` in `AuditActionFilter` with framework-level resolution from `ActionRequestMetadata`.
  Evaluator-agnostic, handles data streams/aliases, no duplicate work.
  
  4. **Generic ComplianceReadIndexSearcherWrapper** — Refactored into reusable components:
     - `FieldReadHandler` interface (generic field read interception contract)
     - `ReadInterceptIndexSearcherWrapper` (generic wrapper accepting predicate + handler factory)
     - `ComplianceReadIndexSearcherWrapper` (thin compliance-specific consumer)
  
  ## Testing
  
  - All unit tests pass (`AuditActionFilterTest`, `AuditTransportInterceptorTest`)
  - All integration tests pass (`StandaloneAudit*` — 85+ tests)
  - Existing FGAC tests unaffected

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
+++ b/src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
@@ -466,6 +466,15 @@ public class AuditConfig {
             return ignoredAuditUsersMatcher.test(user);
         }
 
+        /**
+         * Check if user is included in audit.
+         * @param user
+         * @return true if user is included in audit logging
+         */
+        public boolean isAuditEnabled(String user) {
+            return !ignoredAuditUsersMatcher.test(user);
+        }
+
         @VisibleForTesting
         WildcardMatcher getIgnoredAuditRequestsMatcher() {
             return ignoredAuditRequestsMatcher;
--- a/src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
+++ b/src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
@@ -459,7 +459,7 @@ public class AuditConfig {
 
         /**
          * Check if user is excluded from audit.
-         * @param user
+         * @param user effective user name (from FGAC user or SSL principal)
          * @return true if user is excluded from audit logging
          */
         public boolean isAuditDisabled(String user) {
@@ -468,7 +468,7 @@ public class AuditConfig {
 
         /**
          * Check if user is included in audit.
-         * @param user
+         * @param user effective user name (from FGAC user or SSL principal)
          * @return true if user is included in audit logging
          */
         public boolean isAuditEnabled(String user) {
--- a/src/main/java/org/opensearch/security/compliance/FieldReadHandler.java
+++ b/src/main/java/org/opensearch/security/compliance/FieldReadHandler.java
@@ -21,7 +21,13 @@ public interface FieldReadHandler {
 
     void stringFieldRead(FieldInfo fieldInfo, String value);
 
-    void numericFieldRead(FieldInfo fieldInfo, Number value);
+    void intFieldRead(FieldInfo fieldInfo, int value);
+
+    void longFieldRead(FieldInfo fieldInfo, long value);
+
+    void floatFieldRead(FieldInfo fieldInfo, float value);
+
+    void doubleFieldRead(FieldInfo fieldInfo, double value);
 
     /**
      * Called after all fields for a single document have been read.
--- a/src/main/java/org/opensearch/security/compliance/FieldReadHandler.java
+++ b/src/main/java/org/opensearch/security/compliance/FieldReadHandler.java
@@ -1,0 +1,31 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.security.compliance;
+
+import org.apache.lucene.index.FieldInfo;
+
+/**
+ * Generic interface for handling field-level read interceptions.
+ * Implementations decide what to do when a stored field is read
+ * (e.g., compliance audit logging, access tracking, analytics).
+ */
+public interface FieldReadHandler {
+
+    void binaryFieldRead(FieldInfo fieldInfo, byte[] value);
+
+    void stringFieldRead(FieldInfo fieldInfo, String value);
+
+    void numericFieldRead(FieldInfo fieldInfo, Number value);
+
+    /**
+     * Called after all fields for a single document have been read.
+     * Implementations should flush/finalize any pending work here.
+     */
+    void finished();
+}
--- a/src/main/java/org/opensearch/security/compliance/ReadInterceptIndexSearcherWrapper.java
+++ b/src/main/java/org/opensearch/security/compliance/ReadInterceptIndexSearcherWrapper.java
@@ -30,7 +30,7 @@ import org.opensearch.index.shard.ShardUtils;
  * to a {@link FieldReadHandler}. Reusable for any feature that needs to
  * observe field-level reads (compliance tracking, access logging, analytics, etc.).
  *
- * @see ComplianceReadIndexSearcherWrapper for the compliance-specific consumer
+ * @see ComplianceReadIndexSearcherWrapper — the current compliance-specific consumer
  */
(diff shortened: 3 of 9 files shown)


DarshitChanpura (commented):
Follow-up review — checked the four asks from #6304 and one adjacent item.

**Verified clean:**
- `Setting.Property.Sensitive` — all 13 `AUDIT_*` and 10 `COMPLIANCE_*` settings covered.
- `isAuditEnabled` refactor — clean rename, `null` short-circuit preserved.
- Generic `ReadInterceptIndexSearcherWrapper` — `ShardId` threaded through the factory correctly (preserves the fix from the earlier PR).
- `actionRequestMetadata.resolvedIndices()` switch — verified against OpenSearch core. Resolution is lazy inside `ActionRequestMetadata.resolvedIndices()` (server: `ActionRequestMetadata.java:50-60`) and delegates to core's `IndexNameExpressionResolver` via `TransportIndicesResolvingAction.resolveIndices(request)`, not the security plugin's resolver — so it works in SSL-only mode. All common audit-relevant `TransportAction`s (`Search`, `Bulk`, `ShardBulk`, `Get`, `Index`, `Update`, `Delete`, `MultiGet`, `FieldCapabilities`) implement `TransportIndicesResolvingAction` directly or via base classes, so no silent `Unknown` degradation on real actions.
- Happy-path coverage — `StandaloneAuditFilterFeaturesTest.java:400,435` asserts `RESOLVED_INDICES` for wildcard-expansion and concrete-index cases, and `StandaloneAuditDynamicFilterSettingsTest.java:284` covers the dynamic toggle. End-to-end coverage for the new lazy-resolution path exists.

**One out-of-diff nit worth flagging** (introduced in #6304, not this PR, but you're in the file):

`AuditActionFilter.java` around the new `AUDIT_FILTER_ORDER = -200` block — the Javadoc says "Run before authentication/authorization filters ... SecurityFilter uses Integer.MIN_VALUE; we use a less extreme value." ActionFilters run in ascending order, so `Integer.MIN_VALUE < -200` means `SecurityFilter` would run *first* and `AuditActionFilter` would run *after* it, not before. The comment describes the opposite of what the value does.

In practice this is a documentation issue, not a runtime bug: `OpenSearchSecurityPlugin.getActionFilters()` (lines 1213-1228) registers `SecurityFilter` and `AuditActionFilter` in an `if / else if` — FGAC mode gets `SecurityFilter`, SSL-only/disabled mode gets `AuditActionFilter`, never both. So the two filters can't collide and audit events aren't dropped by ordering. But someone reading the Javadoc in a future refactor will believe the ordering invariant holds when it can't (there's no `int` value less than `Integer.MIN_VALUE`).

Suggested fix — rewrite the Javadoc to state the actual invariant, e.g.:

```java
/**
 * Only registered in SSL-only / disabled modes where SecurityFilter is not
 * registered (see OpenSearchSecurityPlugin#getActionFilters). The two filters
 * are mutually exclusive by design. The low order value is chosen so audit
 * runs before any other action filter that might mutate the request; it does
 * not (and cannot) precede SecurityFilter's Integer.MIN_VALUE.
 */
private static final int AUDIT_FILTER_ORDER = -200;
```

Inline nits below.

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
```diff
@@ -466,6 +466,15 @@ public boolean isAuditDisabled(String user) {
             return ignoredAuditUsersMatcher.test(user);
         }
 
+        /**
+         * Check if user is included in audit.
+         * @param user
```
DarshitChanpura: nit: `@param user` has no description. Trivial, but easy to fix while you're here — e.g. `@param user effective user name (from FGAC user or SSL principal)`.

src/main/java/org/opensearch/security/compliance/FieldReadHandler.java
```diff
@@ -0,0 +1,31 @@
...
+ * Implementations decide what to do when a stored field is read
+ * (e.g., compliance audit logging, access tracking, analytics).
+ */
+public interface FieldReadHandler {
+
+    void binaryFieldRead(FieldInfo fieldInfo, byte[] value);
+
+    void stringFieldRead(FieldInfo fieldInfo, String value);
+
+    void numericFieldRead(FieldInfo fieldInfo, Number value);
```
DarshitChanpura: nit (performance): `numericFieldRead(FieldInfo, Number)` autoboxes on every `int`/`long`/`float`/`double` read. On a hot stored-fields read path this is measurable allocation pressure. The underlying `FieldReadCallback` already has typed methods — if this interface is genuinely meant to be reused (per the class Javadoc: "compliance tracking, access logging, analytics"), it's worth mirroring them here:

```java
void intFieldRead(FieldInfo fieldInfo, int value);
void longFieldRead(FieldInfo fieldInfo, long value);
void floatFieldRead(FieldInfo fieldInfo, float value);
void doubleFieldRead(FieldInfo fieldInfo, double value);
```

Or, if that's overkill for a single consumer today, at least drop a comment noting the autoboxing so a future perf-sensitive consumer knows to widen the interface.

src/main/java/org/opensearch/security/compliance/ReadInterceptIndexSearcherWrapper.java
```diff
@@ -0,0 +1,229 @@
...
+import org.opensearch.common.lucene.index.SequentialStoredFieldsLeafReader;
+import org.opensearch.core.index.shard.ShardId;
+import org.opensearch.index.shard.ShardUtils;
+
+/**
+ * A generic reader wrapper that intercepts stored field reads and delegates
+ * to a {@link FieldReadHandler}. Reusable for any feature that needs to
+ * observe field-level reads (compliance tracking, access logging, analytics, etc.).
+ *
+ * @see ComplianceReadIndexSearcherWrapper for the compliance-specific consumer
```
DarshitChanpura: nit: The class-level Javadoc says the wrapper is "reusable for any feature ... compliance tracking, access logging, analytics" but the `@see ComplianceReadIndexSearcherWrapper` on the very next line reads as if that's the sole consumer. Either drop the `@see` (so the reusability claim stands on its own) or reword it as "@see ComplianceReadIndexSearcherWrapper — the current compliance-specific consumer." Small thing, but the two lines are in tension as written.

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
Taiwo435: okay

src/main/java/org/opensearch/security/compliance/ReadInterceptIndexSearcherWrapper.java
Taiwo435: went with "@see ComplianceReadIndexSearcherWrapper — the current compliance-specific consumer."

src/main/java/org/opensearch/security/compliance/FieldReadHandler.java
Taiwo435: ok

DarshitChanpura (approved):

PR #6323, opened 2026-07-22
Support multiple ResourceProviders per shared resource index

DarshitChanpura:
## Summary

Extends the resource-sharing framework so multiple `ResourceProvider`s can share a single resource index (motivating case: alerting registers both `monitor` and `workflow` on `.opendistro-alerting-config`). Three related changes; each keeps existing single-provider-per-index plugins unchanged in behavior.

### 1. `ResourcePluginInfo.getResourceTypeForIndexOp` iterates all matching providers

The method previously used `findFirst()` on providers matching an index and asked only that provider's `typeField` to classify the incoming document. When two providers share an index, only one could classify docs correctly — the loser's `typeField` resolves to null and the `postIndex` listener silently skips recording a sharing entry.

Now iterates every provider whose `resourceIndexName` matches:
- No `typeField` declared: fall back to the provider's static `resourceType()` (preserves current single-type-per-index behavior).
- Otherwise, extract the provider's `typeField` from the index op; the first provider whose extraction yields a non-null value wins.

### 2. SPI: per-provider owner paths on `ResourceProvider`

`ResourceProvider` gains two default methods, both returning `null`:

```java
default String ownerNamePath()          { return null; }
default String ownerBackendRolesPath()  { return null; }
```

Existing providers (ml-commons, flow-framework, reporting, sample-resource-plugin) don't override them → they return null → the migrate endpoint falls back to the request-level `username_path` / `backend_roles_path` (still mandatory). Zero behavior change.

Multi-type-per-index plugins can now declare type-specific paths (e.g. `/monitor/user/name` on the monitor provider, `/workflow/user/name` on the workflow provider) so a single `POST /_plugins/_security/api/resources/migrate` call can attribute owners for both types without pre-processing.

### 3. Migrate endpoint changes

- **Per-doc classifier**: iterates every registered `typeField` (mirrors change #1 but on the JSON pointer path used at migrate time).
- **Per-doc owner paths**: uses the classified type's provider `ownerNamePath` / `ownerBackendRolesPath` when declared, falls back to the request-level paths otherwise.
- **Skip unclassifiable docs**: previously a single doc the classifier couldn't type would 400 the whole call. Now those docs are recorded under `MigrationStats#skippedNoType` and the migration proceeds — matches the existing behavior for docs classified but missing an owner.

### 4. Extracted helpers + unit tests

- `classifyDocType(...)` — the type-resolution ladder (per-type paths → access-level map fallback → single-registered-type inference).
- `jsonPointer(String)` — normalize a dot-notation or JSON pointer path.
- 8 unit tests for the migrate action + 6 for `getResourceTypeForIndexOp` covering: single/multiple providers, resolves/no-resolves, request-level fallback, unknown index.

## Backward compatibility

- All SPI additions default to null; existing plugins compile and run unchanged.
- Sample-plugin migrate ITs (16/16) pass, including the "missing `username_path`" negative test.
- Request schema unchanged: `username_path` and `backend_roles_path` remain mandatory (validated at `mandatoryKeys`).

## Motivating case

`opensearch-project/alerting#2180` registers both `monitor` and `workflow` on `.opendistro-alerting-config`. Verified locally end-to-end against this build: legacy monitor → RSC enabled → 403 → `POST /_plugins/_security/api/resources/migrate` → alice recovers access. Alerting no longer needs a plugin-side migration endpoint. See [design note on #4500](https://github.com/opensearch-project/security/issues/4500#issuecomment-5062337215).

## Companion PRs

- `opensearch-project/alerting#2180` — uses the new SPI methods, drops the alerting-side `_migrate_to_rsc` endpoint.
- `opensearch-project/documentation-website` (draft, will open after this merges) — updates the migrate API doc to note per-provider owner paths and the `skippedResources` semantics.

### Check List
- [x] New functionality has been documented (companion PR to opensearch-project/documentation-website)
- [ ] New Roles/Permissions have a corresponding security dashboards plugin PR (n/a)
- [ ] API changes companion pull request created (request schema unchanged; only SPI + response semantics extended)

### DIFF
--- a/src/main/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiAction.java
+++ b/src/main/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiAction.java
@@ -11,6 +11,7 @@ package org.opensearch.security.resources.api.migrate;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -157,6 +158,11 @@ public class MigrateResourceSharingInfoApiAction extends AbstractApiAction {
 
         // Extract fields - validation already done by RequestContentValidator framework via FieldConfiguration
         String sourceIndex = body.get("source_index").asText();
+
+        // Request-level owner paths (username_path / backend_roles_path) are still required at the
+        // schema layer, preserving backward compatibility for single-type-per-index callers.
+        // ResourceProviders may override them per-type via ownerNamePath() / ownerBackendRolesPath()
+        // when a single request needs to attribute owners for multiple types sharing an index.
         String userNamePath = body.get("username_path").asText();
         String backendRolesPath = body.get("backend_roles_path").asText();
 
@@ -178,6 +184,12 @@ public class MigrateResourceSharingInfoApiAction extends AbstractApiAction {
         // first non-null value.
         List<String> typePaths = new ArrayList<>();
 
+        // Per-type owner paths. When a provider declares ownerNamePath()/ownerBackendRolesPath()
+        // these take precedence over the request-level username_path/backend_roles_path so a
+        // single migrate call can attribute owners for multiple types sharing one index.
+        Map<String, String> typeToOwnerNamePath = new HashMap<>();
+        Map<String, String> typeToOwnerBackendRolesPath = new HashMap<>();
+
         // Validate each type + its accessLevel
         for (Map.Entry<String, String> entry : typeToDefaultAccessLevel.entrySet()) {
             String type = entry.getKey();
@@ -193,6 +205,12 @@ public class MigrateResourceSharingInfoApiAction extends AbstractApiAction {
 
             if (provider.typeField() != null) {
                 typePaths.add(provider.typeField());
+            }
+            if (provider.ownerNamePath() != null) {
+                typeToOwnerNamePath.put(type, provider.ownerNamePath());
+            }
+            if (provider.ownerBackendRolesPath() != null) {
+                typeToOwnerBackendRolesPath.put(type, provider.ownerBackendRolesPath());
             }
 
             // Allowed access-levels for this type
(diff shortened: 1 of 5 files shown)


src/test/java/org/opensearch/security/resources/ResourcePluginInfoTests.java
```diff
@@ -0,0 +1,188 @@
...
+        String result = resourcePluginInfo.getResourceTypeForIndexOp(".alerting-config", indexOp);
+        assertEquals("monitor", result);
+    }
+
+    @Test
+    public void testSingleProviderWithTypeField() {
+        registerProviders(List.of("monitor"), ".alerting-config", "monitor.type");
+        resourcePluginInfo.updateProtectedTypes(List.of("monitor"));
+
+        Engine.Index indexOp = mockIndexOp(new StringField("monitor.type", "monitor", Field.Store.NO));
```
cwperks: Thanks for capturing this in a test. Is it possible to create a small representative doc here that resembles the payload of what gets indexed into the alerting index for monitor/workflow?

RyanL1997 (approved):

src/main/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -157,6 +158,11 @@ private ValidationResult<ValidationResultArg> loadCurrentSharingInfo(RestRequest
 
         // Extract fields - validation already done by RequestContentValidator framework via FieldConfiguration
         String sourceIndex = body.get("source_index").asText();
+
+        // Request-level owner paths (username_path / backend_roles_path) are still required at the
+        // schema layer, preserving backward compatibility for single-type-per-index callers.
+        // ResourceProviders may override them per-type via ownerNamePath() / ownerBackendRolesPath()
+        // when a single request needs to attribute owners for multiple types sharing an index.
         String userNamePath = body.get("username_path").asText();
```
cwperks, line 166: What value wins in this case or does it cascade to find it? I would imagine the one on the ResourceProvider wins and that actually seems like a better general solution.

src/main/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiAction.java
```diff
@@ -172,7 +178,17 @@ private ValidationResult<ValidationResultArg> loadCurrentSharingInfo(RestRequest
...
 
-        String typePath = null;
+        // Collect the type-field JSON pointer for each registered type. When multiple providers
+        // share an index, each provider may declare a type-specific path (e.g. `monitor.type` /
+        // `workflow.type`); the per-doc classifier below tries each candidate and picks the
+        // first non-null value.
+        List<String> typePaths = new ArrayList<>();
+
+        // Per-type owner paths. When a provider declares ownerNamePath()/ownerBackendRolesPath()
+        // these take precedence over the request-level username_path/backend_roles_path so a
```
cwperks, line 188: Answered here.

cwperks (approved):

PR #6343, opened 2026-07-28
feat(audit): Add audit_request_id correlation field to all FGAC audit…

Taiwo435:
## Description

Add an `audit_request_id` field to all audit events so that multiple events generated by a single request (AUTHENTICATED → GRANTED_PRIVILEGES → COMPLIANCE_DOC_WRITE) can be correlated.

### How it works

- `SecurityRestFilter` reads the `X-Request-Id` header if present, otherwise generates a UUID
- Sanitizes client-provided values (strips control characters via pre-compiled pattern, relies on core's http.request_id.max_length validation for length)
- Stored in `ThreadContext` as a transient for local events
- `AuditLogImpl.save()` reads the transient first, falls back to `X-Request-Id` header (propagated by core across nodes)
- All events from the same request share the same ID

### Cross-node propagation

When a client sends `X-Request-Id`, OpenSearch core propagates it across nodes via `Task.REQUEST_HEADERS`. This means compliance events (`COMPLIANCE_DOC_WRITE`, `COMPLIANCE_DOC_READ`) on remote shard-hosting nodes also get the correlation ID.

**Known limitation:** Server-generated UUIDs (when no X-Request-Id header is sent) appear on user-initiated events (AUTHENTICATED, GRANTED_PRIVILEGES for the user action) but NOT on internally-spawned events COMPLIANCE_DOC_WRITE, auto-mapping INDEX_EVENT). This is because internal sub-actions run in a fresh ThreadContext where transients are cleared. When the client provides X-Request-Id, ALL events get the correlation ID because headers survive context changes. 

### References

- OpenSearch core `X-Request-Id` support: opensearch-project/OpenSearch#19798

### Testing

- Integration tests (`AuditRequestIdCorrelationTest`):
  - Same request → all events share one ID
  - Client-provided `X-Request-Id` header used verbatim
  - UUID generated when no header present
  - Different requests → different IDs
  - Compliance doc write gets the request ID (when `X-Request-Id` sent)
- Unit tests (`AuditMessageTest`): `addRequestId()` behavior
- Existing test fix (`WhoAmITests`): exclude `audit_request_id` from structural equality comparison

### Check List


src/integrationTest/java/org/opensearch/security/rest/WhoAmITests.java
```diff
@@ -322,6 +323,10 @@ private void assertCommonFields(Set<String> commonFields, Map<String, Object> re
             } else if (key.equals("audit_request_layer")) {
                 assertThat(restMsgFields.get(key).toString(), equalTo("REST"));
                 assertThat(transportMsgFields.get(key).toString(), equalTo("TRANSPORT"));
+            } else if (key.equals(AuditMessage.REQUEST_ID)) {
```
DarshitChanpura, line 326: Two things here:

- The comment ("Request IDs are unique per request") is backwards for this feature — within a single request the whole point is that the ID is **shared**, so REST and transport events should be *equal*, not merely non-null.
- This branch only executes when the key is present on **both** messages (it iterates the key intersection). If the transport event doesn't carry `audit_request_id`, the branch silently never runs — so as written it neither proves correlation nor catches its absence.

Suggest asserting `equalTo(transportMsgFields.get(key))` here, and separately asserting the transport message actually contains the key. If equality can't hold across layers, that's the thing to fix rather than assert around.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -408,4 +416,23 @@ private boolean restPathMatches(String requestPath, String handlerPath) {
...
+     */
+    public static String sanitizeRequestId(String rawRequestId) {
+        if (rawRequestId == null || rawRequestId.isEmpty()) {
+            return UUID.randomUUID().toString();
+        }
+        rawRequestId = rawRequestId.replaceAll("[\\p{Cntrl}]", "");
+        if (rawRequestId.isEmpty()) {
+            return UUID.randomUUID().toString();
+        }
+        if (rawRequestId.length() > 128) {
```
DarshitChanpura: The cap is hardcoded to 128, but OpenSearch core validates `X-Request-Id` against the **dynamic** setting `http.request_id.max_length` (default 128, range 16–1024). If an operator raises that setting, core admits a longer header and this truncates it — so `audit_request_id` no longer matches the `X-Request-Id` core echoes into responses and slow logs, breaking the cross-system correlation the field exists for. Consider reading the same setting (or dropping the truncation and relying on core's validation) so the two never diverge.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -408,4 +416,23 @@ private boolean restPathMatches(String requestPath, String handlerPath) {
...
     }
+
+    /**
+     * Sanitizes a raw X-Request-Id header value for use as an audit correlation ID.
+     * Strips control characters first (prevents log injection), then truncates to 128 chars.
+     * Falls back to a generated UUID if the input is null, empty, or entirely control characters.
+     */
+    public static String sanitizeRequestId(String rawRequestId) {
+        if (rawRequestId == null || rawRequestId.isEmpty()) {
+            return UUID.randomUUID().toString();
```
DarshitChanpura: Two minor performance notes on this hot path:
- Core's own convention for these IDs is `RequestUtils.generateID()` → `UUIDs.base64UUID()` (non-blocking). `UUID.randomUUID()` draws from a blocking `SecureRandom` on every header-less request; since this is a correlation tag, not a security token, `UUIDs.base64UUID()` is a better fit and matches core.
- `replaceAll("[\\p{Cntrl}]", "")` (line 429) recompiles the pattern on every call — hoist it to a `private static final Pattern`. The character class doesn't need the extra brackets: `\\p{Cntrl}` suffices.

src/main/java/org/opensearch/security/auditlog/impl/AuditLogImpl.java
```diff
@@ -134,6 +136,16 @@ public void close() throws IOException {
     @Override
     protected void save(final AuditMessage msg) {
         if (enabled) {
+            // Try transient first (coordinating node, already sanitized by SecurityRestFilter),
+            // fall back to X-Request-Id header (propagated by core to remote nodes — needs sanitization)
+            String requestId = getThreadContext().getTransient(ConfigConstants.SECURITY_AUDIT_REQUEST_ID);
+            if (requestId == null) {
+                String rawHeader = getThreadContext().getHeader(Task.X_REQUEST_ID);
```
DarshitChanpura, line 143: `sanitizeRequestId` mints a **fresh UUID** when its input reduces to empty. Core's validation uses `isBlank()`, which does not strip control characters, so a header made of control chars passes core, propagates across nodes, and here strips to empty → each node generates a *different* UUID for the same request, defeating correlation. Pathological, but the fallback should return `null` (→ no ID) rather than a new UUID when the header is unusable.

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -1123,6 +1123,13 @@ private Map<String, String> getThreadContextHeaders() {
         return threadPool.getThreadContext().getHeaders();
     }
 
+    /**
+     * Provides subclass access to the ThreadContext without exposing the full ThreadPool.
+     */
+    protected org.opensearch.common.util.concurrent.ThreadContext getThreadContext() {
```
DarshitChanpura: Nit: use an `import` for `ThreadContext` and the short name here — this file already works with `threadPool.getThreadContext()` elsewhere, so the fully-qualified inline type is inconsistent with the surrounding style.

src/integrationTest/java/org/opensearch/security/AuditRequestIdCorrelationTest.java
```diff
@@ -0,0 +1,188 @@
...
+        List<AuditMessage> userEvents = messages.stream()
+            .filter(msg -> msg.getCategory() == AuditCategory.AUTHENTICATED || msg.getCategory() == AuditCategory.GRANTED_PRIVILEGES)
+            .collect(Collectors.toList());
+        assertThat("Should produce a user-initiated event", userEvents.size(), greaterThanOrEqualTo(1));
+        String userEventRequestId = (String) userEvents.get(0).getAsMap().get(AuditMessage.REQUEST_ID);
+        assertThat("User event and compliance event must share the same request ID", userEventRequestId, equalTo(complianceRequestId));
+    }
+
+    @Test
+    public void shouldNotHaveRequestIdOnInternalTransportEvents() throws Exception {
```
DarshitChanpura: The test name says the opposite of what it asserts — the body and comments concede that on a single-node cluster everything goes through the REST path and *does* get an ID, so it only checks non-null. Either rename it to reflect what it verifies (e.g. `shouldStampRequestIdOnAdminCertPath`) or drop it; as written it's misleading.

DarshitChanpura (commented):
Thanks for this — the feature is well-scoped and the PR description is admirably honest about its limitations. The mechanism (transient stamped at the REST filter, read back in `AuditLogImpl.save()` with an `X-Request-Id` header fallback for remote nodes) is sound. I have one question that gates my approval and a few smaller items; none is a demonstrated blocker on its own.

**Main question — does the transport-layer event actually carry the shared ID on the common (no-header) path?**

`AuditLogImpl.save()` only *reads* the ID (transient, else non-empty header) and generates nothing on the transport path. For a request with no client `X-Request-Id`, the only source is the transient set by `SecurityRestFilter`. Transient headers don't cross the transport boundary, and the PR's own "Known limitation" says internally-spawned events don't get the ID absent a client header. So it looks like "all events from the same request share one ID" holds only among coordinating-node REST events in the common case.

The test suite can't currently detect this. The new branch in `WhoAmITests.assertCommonFields` runs inside a loop over the **intersection** of REST and transport keys — if the transport event lacks `audit_request_id`, the key isn't common, the branch never executes, and the test passes regardless. Could you confirm whether a transport-layer event carries `audit_request_id` when the client sends no header, and add a test that asserts the transport message **contains** the key and equals the REST event's value (outside the common-keys loop)?

Everything else is minor and inline.

src/integrationTest/java/org/opensearch/security/rest/WhoAmITests.java
```diff
@@ -326,5 +327,16 @@ private void assertCommonFields(Set<String> commonFields, Map<String, Object> re
                 assertThat(restMsgFields.get(key), equalTo(transportMsgFields.get(key)));
             }
         }
+
+        // Assert request ID correlation outside the common-keys loop:
```
DarshitChanpura: This new correlation block compares request IDs from **two different REST requests**, so the `equalTo` assertion will fail deterministically.

`testAuditLogSimilarityWithTransportLayer` issues two separate calls (neither sends `X-Request-Id`):

```java
client.get(WHOAMI_PROTECTED_ENDPOINT);   // → REST-layer GRANTED_PRIVILEGES event
...
client.get("_cat/indices");              // → TRANSPORT-layer GetSettingsRequest event
```

`verifyAuditLogSimilarity` then passes `restSet.get(0)` + `transportSet.get(0)` here, and the existing anchor assertions in `checkForStructuralSimilarity` confirm which events those are — the REST event is the `whoamiprotected` request (`audit_rest_request_path == /_plugins/_security/whoamiprotected`) and the transport event is the `_cat/indices` request (`audit_transport_request_type == GetSettingsRequest`). Two distinct HTTP requests.

Since neither request sends `X-Request-Id`, `SecurityRestFilter` generates an **independent `base64UUID` per request**. So:

```java
assertThat("REST and transport events must share the same request ID",
    restMsgFields.get(AuditMessage.REQUEST_ID),
    equalTo(transportMsgFields.get(AuditMessage.REQUEST_ID)));  // whoamiprotected UUID != _cat/indices UUID → fails
```

The comment ("on single-node, the transient is available to both layers") is only true *within a single request* — the shared-ID guarantee never applied across these two. The `notNullValue()` checks just above are fine (both requests do get an ID); it's specifically the `equalTo` that breaks.

`checkForStructuralSimilarity` intentionally pairs cross-request events to compare field *shape*, so a request-ID equality check doesn't belong in it. Suggest either:
- moving the equality check into a dedicated test that issues **one** request producing both a REST and a same-request transport event and asserts equality on those, or
- dropping the `equalTo` here and keeping only the two `notNullValue()` presence checks.

DarshitChanpura (commented):
One follow-up on commit 64e15b3 — the strengthened `audit_request_id` assertion here will fail deterministically. Details inline.

src/integrationTest/java/org/opensearch/security/rest/WhoAmITests.java
```diff
@@ -326,5 +327,16 @@ private void assertCommonFields(Set<String> commonFields, Map<String, Object> re
...
+        // Both messages must contain the key and their values must be equal
+        // (on single-node, the transient is available to both layers)
+        assertThat("REST event must contain audit_request_id", restMsgFields.get(AuditMessage.REQUEST_ID), notNullValue());
+        assertThat("Transport event must contain audit_request_id", transportMsgFields.get(AuditMessage.REQUEST_ID), notNullValue());
+        assertThat(
+            "REST and transport events must share the same request ID",
+            restMsgFields.get(AuditMessage.REQUEST_ID),
+            equalTo(transportMsgFields.get(AuditMessage.REQUEST_ID))
+        );
```
DarshitChanpura: This fixes both the loop failure at line 327 and removes the invalid cross-request equality block. `audit_request_id` differs by design here (two distinct requests), so we assert presence rather than equality — mirroring how `audit_request_layer` is already special-cased.

```suggestion
                if (key.equals(AuditMessage.REQUEST_ID)) {
                    // Two distinct requests are paired here for structural comparison
                    // (whoamiprotected vs _cat/indices), so each generates its own
                    // base64UUID — assert presence, not cross-request equality.
                    assertThat(restMsgFields.get(key), notNullValue());
                    assertThat(transportMsgFields.get(key), notNullValue());
                } else {
                    assertThat(restMsgFields.get(key), equalTo(transportMsgFields.get(key)));
                }
            }
        }
```

DarshitChanpura (commented):
Confirmed this is a deterministic failure — CI is red on all four `integration-tests` matrix combos (Java 21/25 × ubuntu/windows) and the log names it directly:

```
WhoAmITests > testAuditLogSimilarityWithTransportLayer FAILED
    java.lang.AssertionError:
    Expected: "S7Dgtp8BZ_cUB07XDA6W"
        at WhoAmITests.assertCommonFields(WhoAmITests.java:327)
        at WhoAmITests.checkForStructuralSimilarity(WhoAmITests.java:290)
```

One small correction to my earlier note on *which* line throws: because `audit_request_id` is now a **common key** in both messages, it gets compared inside the generic equality loop at line 327 and fails there *first* — before ever reaching the dedicated `equalTo` block at 336–340. Both are broken by the same root cause (the REST `whoamiprotected` request and the transport `_cat/indices` request each generate their own `base64UUID`), so the fix needs to address both spots.

Suggested patch below: skip cross-request equality for `audit_request_id` in the loop (treat it like `audit_request_layer`), assert presence only, and drop the invalid cross-request `equalTo` block. Verified `notNullValue` is already imported by this PR, so it compiles.

If you also want to keep a real *correlation* assertion, add a separate test that issues a single request producing both a REST and a same-request transport event (or sends an explicit `X-Request-Id` on the calls) and asserts equality there — that's the only scenario where the shared-ID guarantee actually holds.

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -408,4 +416,26 @@ private boolean restPathMatches(String requestPath, String handlerPath) {
         }
         return true;
     }
+
+    /** Pre-compiled pattern for control character stripping — avoids recompilation on every call. */
+    private static final Pattern CONTROL_CHARS = Pattern.compile("\\p{Cntrl}");
```
DarshitChanpura, line 421: this might be more useful in core but don't mind keeping it here.

DarshitChanpura (approved):

PR #6344, opened 2026-07-29
feat(audit): Add user_agent, user_roles, and auth_method to audit events

Taiwo435:
## Description

Enrich FGAC audit events with three fields that improve investigability:

- `audit_request_user_agent`: User-Agent string from REST headers
- `audit_request_user_roles`: mapped security roles from the authenticated user
- `audit_request_auth_method`: authentication backend that validated the user (e.g., `internal`, `ldap`)

### Why

Current audit events identify WHO (effective_user) and WHAT (action, indices) but not HOW they authenticated or what role granted them access. Incident investigators need these fields to answer: "Was this a service account or human?", "Which role allowed this?", "What client tool was used?"

### How it works

- Added `enrichWithUserContext()` helper to `AbstractAuditLog`
- Called from all FGAC audit methods: logFailedLogin, logSucceededLogin, logMissingPrivileges, logGrantedPrivileges, logIndexEvent, logClusterSettingsChange, logIndexSettingsChange, logSecurityIndexAttempt
- Graceful degradation: fields are absent (not null/empty) when User object is unavailable (e.g., failed login before auth completes, admin cert bypass)

### Testing

- Integration test: `AuditFieldEnrichmentTest` covering:
  - user_agent present on REST events
  - user_roles present after successful auth
  - auth_method present after successful auth
  - Fields absent on failed login (graceful degradation)
  - Fields absent on admin cert bypass

### Check List

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/AuditFieldEnrichmentMultiNodeTest.java
+++ b/src/integrationTest/java/org/opensearch/security/AuditFieldEnrichmentMultiNodeTest.java
@@ -1,0 +1,127 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.security;
+
+import java.util.Map;
+
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.opensearch.security.auditlog.impl.AuditCategory;
+import org.opensearch.security.auditlog.impl.AuditMessage;
+import org.opensearch.test.framework.AuditConfiguration;
+import org.opensearch.test.framework.AuditFilters;
+import org.opensearch.test.framework.TestSecurityConfig;
+import org.opensearch.test.framework.audit.AuditLogsRule;
+import org.opensearch.test.framework.cluster.ClusterManager;
+import org.opensearch.test.framework.cluster.LocalCluster;
+import org.opensearch.test.framework.cluster.TestRestClient;
+
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+
+/**
+ * Multi-node integration tests verifying that audit field enrichment (user_roles, auth_method)
+ * survives User serialization across nodes. Uses a 3-node cluster (1 cluster manager + 2 data nodes)
+ * to exercise the header-deserialization fallback path in enrichWithUserContext().
+ */
+public class AuditFieldEnrichmentMultiNodeTest {
+
+    private static final TestSecurityConfig.User ADMIN_USER = new TestSecurityConfig.User("admin").roles(
+        new TestSecurityConfig.Role("all_access").clusterPermissions("*").indexPermissions("*").on("*")
+    );
+
+    @ClassRule
+    public static LocalCluster cluster = new LocalCluster.Builder().clusterManager(ClusterManager.DEFAULT)
+        .users(ADMIN_USER)
+        .authc(AUTHC_HTTPBASIC_INTERNAL)
+        .audit(new AuditConfiguration(true).filters(new AuditFilters().enabledRest(true).enabledTransport(true)))
+        .build();
+
+    @Rule
+    public AuditLogsRule auditLogsRule = new AuditLogsRule();
+
+    /**
+     * Verifies that user_roles is present on shard-level audit events that fire on data nodes.
+     * The User object is serialized to a header on the coordinating node and deserialized on
+     * data nodes — securityRoles survives this round-trip because it is in serialPersistentFields.
+     * Asserts the concrete auth_method value ("basic") so a null-from-deserialization
+     * cannot hide behind a coordinator-side event.
+     */
+    @Test
+    public void shouldCaptureUserRolesOnCrossNodeShardEvents() {
+        try (TestRestClient client = cluster.getRestClient(ADMIN_USER)) {
+            // Create an index with 2 primary shards to ensure distribution across data nodes
+            client.putJson("multinode-test-index", "{\"settings\": {\"number_of_shards\": 2, \"number_of_replicas\": 1}}");
+            // Index a document — triggers shard-level transport operations on data nodes
+            client.putJson("multinode-test-index/_doc/1", "{\"field\": \"value\"}");
+        }
+
+        // GRANTED_PRIVILEGES events from shard-level ops should have user_roles and concrete auth_method
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.GRANTED_PRIVILEGES) return false;
+            Map<String, Object> fields = msg.getAsMap();
+            Object roles = fields.get(AuditMessage.USER_ROLES);
+            Object authMethod = fields.get(AuditMessage.AUTH_METHOD);
+            return roles != null && roles.toString().contains("all_access") && "basic".equals(String.valueOf(authMethod));
+        });
+    }
+
+    /**
+     * Verifies that auth_method is present on shard-level audit events that fire on data nodes.
+     * Before the serialization fix, authenticatedBy was transient and lost during cross-node
+     * User serialization — this test pins the corrected behavior by asserting the concrete
+     * value ("basic") so a null-from-deserialization cannot hide behind a coordinator-side event.
+     */
+    @Test
+    public void shouldCaptureAuthMethodOnCrossNodeShardEvents() {
+        try (TestRestClient client = cluster.getRestClient(ADMIN_USER)) {
+            // Create an index with 2 primary shards to ensure distribution across data nodes
+            client.putJson("multinode-auth-test-index", "{\"settings\": {\"number_of_shards\": 2, \"number_of_replicas\": 1}}");
+            // Index a document — triggers shard-level transport operations on data nodes
+            client.putJson("multinode-auth-test-index/_doc/1", "{\"field\": \"value\"}");
+        }
+
+        // GRANTED_PRIVILEGES events from shard-level ops should have concrete auth_method value
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.GRANTED_PRIVILEGES) return false;
+            Map<String, Object> fields = msg.getAsMap();
+            Object authMethod = fields.get(AuditMessage.AUTH_METHOD);
+            return "basic".equals(String.valueOf(authMethod));
+        });
+    }
+
+    /**
+     * Verifies both user_roles and auth_method are present on GRANTED_PRIVILEGES events triggered
+     * by the read-path (search). This exercises the two-arg enrichWithUserContext(msg, user) overload
+     * used for pre-resolved User objects in transport audit paths, and filters by the search action
+     * to provide distinct coverage beyond the write-path GRANTED_PRIVILEGES tests above.
+     */
+    @Test
+    public void shouldCaptureRolesAndAuthMethodTogetherOnCrossNodeEvents() {
+        try (TestRestClient client = cluster.getRestClient(ADMIN_USER)) {
+            client.putJson("multinode-combined-test-index", "{\"settings\": {\"number_of_shards\": 2, \"number_of_replicas\": 1}}");
+            client.putJson("multinode-combined-test-index/_doc/1", "{\"field\": \"value\"}");
+            // Search triggers read-path shard-level fan-out on data nodes
+            client.get("multinode-combined-test-index/_search?q=*");
+        }
+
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -1109,16 +1124,46 @@ private TransportAddress getRemoteAddress() {
...
+     * Enriches the audit message with a pre-resolved User object.
+     * Use this overload in loops to avoid redundant deserialization
+     * for bulk/multi-message transport paths.
+     */
+    private void enrichWithUserContext(AuditMessage msg, User user) {
+        if (user == null) {
+            return;
+        }
+        msg.addUserRoles(user.getSecurityRoles());
+        msg.addAuthMethod(user.getAuthenticatedBy());
```
DarshitChanpura, line 1200: `auth_method` will be inconsistent across nodes. `User.authenticatedBy` is a `transient` field and is not in `serialPersistentFields`, so it is not carried over the wire — `BackendRegistry` sets it only on the coordinating node, and `serialized/User.readResolve()` does not restore it.

- REST-origin events resolve the user on the coordinating node → `auth_method` present.
- Cross-node fan-out (shard-level ops that rebuild `User` from the header in `SecurityRequestHandler`) → `getAuthenticatedBy()` returns null → `auth_method` absent.

So the field appears or disappears depending on where the event fires, which is surprising for investigators. Options: document it as coordinator-path/best-effort, make `authenticatedBy` survive serialization (wire-compat care needed for mixed-version clusters), or at minimum add a multi-node test to pin the behavior. Note `user_roles` is unaffected — `securityRoles` is serialized and survives the round-trip.

src/main/java/org/opensearch/security/auditlog/impl/AuditMessage.java
```diff
@@ -408,6 +415,22 @@ void addRestRequestInfo(final SecurityRequest request, final AuditConfig.Filter
             addRestParams(request.params(), filter);
             addRestMethod(request.method());
 
+            // Extract User-Agent as a top-level field for easy filtering.
+            // Respects the ignore_headers configuration — if an admin excludes "user-agent",
+            // we skip extraction to stay consistent with the header exclusion in addRestHeaders().
+            Map<String, List<String>> headers = request.getHeaders();
+            if (headers != null && !filter.shouldExcludeHeader("user-agent")) {
```
DarshitChanpura: This exclusion gate can diverge from `addRestHeaders`. Here you pass a hardcoded lowercase literal `shouldExcludeHeader("user-agent")`, but `addRestHeaders` passes the actual header key (e.g. `"User-Agent"`). `shouldExcludeHeader` is backed by a case-sensitive matcher (`WildcardMatcher.Exact.test` → `pattern.equals`) and `ignore_headers` values aren't normalized.

So for the same request the two checks can disagree: with `ignore_headers: ["User-Agent"]` and an HTTP/1.1 client, `addRestHeaders` drops it from the headers map but this code still writes the top-level `user_agent` field — a partial leak of the header the admin excluded. (Finding the header is fine; the `equalsIgnoreCase` stream handles wire casing. It's only the exclusion decision that's off.)

Suggest gating on the same candidate the header map uses, or matching case-insensitively — so the comment's "stay consistent with the header exclusion in addRestHeaders()" actually holds.

src/integrationTest/java/org/opensearch/security/AuditFieldEnrichmentTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+    private static final TestSecurityConfig.User MULTI_ROLE_USER = new TestSecurityConfig.User("multirole").roles(
+        new TestSecurityConfig.Role("role_a").clusterPermissions("*").indexPermissions("*").on("*"),
+        new TestSecurityConfig.Role("role_b").clusterPermissions("*").indexPermissions("*").on("*")
+    );
+
+    private static final TestSecurityConfig.User NO_ROLES_USER = new TestSecurityConfig.User("noroles");
+
+    @ClassRule
+    public static LocalCluster cluster = new LocalCluster.Builder().clusterManager(ClusterManager.SINGLENODE)
```
DarshitChanpura, line 51: All tests run on `SINGLENODE`. This never exercises the header-deserialization fallback in `resolveUser()` — the exact path the two-arg `enrichWithUserContext` overload exists to optimize — and it's why the cross-node `auth_method` gap isn't caught. Please add a multi-node case so both the serialization behavior and the two-arg path are covered.

src/integrationTest/java/org/opensearch/security/AuditFieldEnrichmentTest.java
```diff
@@ -0,0 +1,251 @@
...
+        try (TestRestClient client = cluster.getRestClient(ADMIN_USER)) {
+            client.get("_cluster/health");
+        }
+
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.GRANTED_PRIVILEGES) return false;
+            Map<String, Object> fields = msg.getAsMap();
+            Object roles = fields.get(AuditMessage.USER_ROLES);
+            Object authMethod = fields.get(AuditMessage.AUTH_METHOD);
+            return roles != null && authMethod != null;
```
DarshitChanpura, line 166: This asserts `authMethod != null`, which holds on single-node but won't for shard-level events in a real multi-node cluster (see the `AbstractAuditLog` note on `authenticatedBy` not being serialized). Either the assertion is over-strong, or the underlying behavior needs fixing — worth resolving alongside that.

src/test/java/org/opensearch/security/auditlog/impl/AuditMessageTest.java
```diff
@@ -271,4 +271,120 @@ public void testToJsonSplitIndices() {
...
+    public void testAddUserAgentSpecialCharacters() {
+        String complexAgent = "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0";
+        message.addUserAgent(complexAgent);
+        assertThat(message.getAsMap().get(AuditMessage.USER_AGENT), is(complexAgent));
+    }
+
+    @Test
+    public void testUserAgentExtractionRespectsIgnoreHeaders() {
+        // When filter excludes "user-agent", addRestRequestInfo should NOT extract it
+        when(auditFilterMock.shouldExcludeHeader("user-agent")).thenReturn(true);
```
DarshitChanpura: This mocks `shouldExcludeHeader("user-agent") → true`, so the test only exercises the hardcoded-lowercase branch and never runs the real matcher against real header casing. It would pass even with the config-casing vs. wire-casing divergence flagged in `AuditMessage.java`. Consider a test that drives the actual `WildcardMatcher` with a canonically-cased `ignore_headers` value (e.g. `"User-Agent"`).

DarshitChanpura (commented):
Direction is solid and the change is cleanly scoped. Left inline notes on two consistency issues (cross-node `auth_method`, user-agent exclusion casing) and a couple of test-coverage gaps. Nothing blocks the happy path — these are consistency/coverage concerns to resolve or document before merge.

src/test/java/org/opensearch/security/auditlog/impl/AuditMessageTest.java
Taiwo435: The Filter constructor is package-private (config package) and addRestRequestInfo is package-private (impl package), so no single test class can access both to drive the real WildcardMatcher end-to-end. Instead, the test mocks shouldExcludeHeader("User-Agent") → true and shouldExcludeHeader("user-agent") → false — which proves the code now passes the actual header key (from entry.getKey()) rather than the old hardcoded lowercase literal. If the old code were still in place, this test would fail.

src/integrationTest/java/org/opensearch/security/AuditFieldEnrichmentMultiNodeTest.java
```diff
@@ -0,0 +1,118 @@
...
+            // Create an index with 2 primary shards to ensure distribution across data nodes
+            client.putJson("multinode-auth-test-index", "{\"settings\": {\"number_of_shards\": 2, \"number_of_replicas\": 1}}");
+            // Index a document — triggers shard-level transport operations on data nodes
+            client.putJson("multinode-auth-test-index/_doc/1", "{\"field\": \"value\"}");
+        }
+
+        // GRANTED_PRIVILEGES events from shard-level ops should have auth_method
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.GRANTED_PRIVILEGES) return false;
+            Map<String, Object> fields = msg.getAsMap();
```
DarshitChanpura, line 94: Minor: these assertions confirm `auth_method` is present on `GRANTED_PRIVILEGES` events, but they don't prove the value came through the *serialization* path specifically — on a coordinator-resolved event the field would also be present, so the test could pass even if the serialization fix regressed (as long as at least one matching event resolved on the coordinator). To make this a tight regression guard for the transient-field fix, consider asserting on an event you know fired on a data node, or assert the concrete value (e.g. `"basic"`/`internal`) so a null-from-deserialization can't hide behind a coordinator-side event. Not a blocker — the unit-level serialization round-trip is the real guarantee; this just tightens intent.

src/integrationTest/java/org/opensearch/security/AuditFieldEnrichmentMultiNodeTest.java
```diff
@@ -0,0 +1,118 @@
...
+     * Verifies both user_roles and auth_method are present together on cross-node events,
+     * exercising the two-arg enrichWithUserContext(msg, user) overload used for pre-resolved
+     * User objects in transport audit paths.
+     */
+    @Test
+    public void shouldCaptureRolesAndAuthMethodTogetherOnCrossNodeEvents() {
+        try (TestRestClient client = cluster.getRestClient(ADMIN_USER)) {
+            client.putJson("multinode-combined-test-index", "{\"settings\": {\"number_of_shards\": 2, \"number_of_replicas\": 1}}");
+            client.putJson("multinode-combined-test-index/_doc/1", "{\"field\": \"value\"}");
+            // Also do a search to trigger read-path shard-level fan-out
```
DarshitChanpura: Nit: `shouldCaptureRolesAndAuthMethodTogetherOnCrossNodeEvents` adds a `_search` to exercise read-path fan-out but still only asserts on `GRANTED_PRIVILEGES`, same as the two tests above — so it largely overlaps them. If the intent is to cover the read-path two-arg overload distinctly, consider asserting on the search's shard-level event category so the added `_search` call actually contributes coverage beyond the index path.

DarshitChanpura (commented):
Re-reviewed at `fe4e8b79`. Thanks for the thorough follow-up — all three findings from my previous review are addressed well:

- **`auth_method` cross-node gap** → fixed properly. `authenticatedBy` is now added to `serialPersistentFields` and restored in `serialized/User.readResolve()`. I verified both `User` classes share the same explicit `serialVersionUID` (`-5500938501822658596L`), so this is safe for mixed-version clusters — Java's `PutField`/`GetField` tolerates the field being absent (old→new reads null) or extra (new→old ignores it). No `InvalidClassException` risk. Good, minimal change.
- **User-agent exclusion casing** → fixed. The gate now uses the actual header key (`entry.getKey()`) instead of the hardcoded lowercase literal, matching `addRestHeaders()`. The refactor to a single `stream().filter(...).findFirst().ifPresent(...)` is also cleaner.
- **Test coverage** → new `AuditFieldEnrichmentMultiNodeTest` exercises the cross-node header-deserialization path on a real 3-node cluster, and `AuditMessageTest` now has `testUserAgentExclusionUsesActualHeaderKey` which would fail if the hardcoded-lowercase bug regressed. Both are exactly the right regression guards.

Two minor, non-blocking notes on the new multi-node test below. Happy to see this merge once those are considered.

src/integrationTest/java/org/opensearch/security/AuditFieldEnrichmentMultiNodeTest.java
Taiwo435: Investigated this. Search shard-level phases (search[phase/query], search[phase/fetch/id]) are registered as direct TransportService request handlers in OpenSearch core's SearchTransportService — they're not TransportAction subclasses, so they bypass ActionFilterChain and SecurityFilter entirely on data nodes. No
audit events fire on remote nodes for search fan-out. 

The indices:data/read/search GRANTED_PRIVILEGES event only fires at the coordinator level (before shard dispatch). So there's no distinct shard-level audit category for search.

To still give the _search call meaningful differentiation, I kept GRANTED_PRIVILEGES but added a filter on AuditMessage.PRIVILEGE to match only indices:data/read/search events. This ensures the test covers read-path enrichment at the coordinator level rather than overlapping with the write-path assertions in tests 1 and 2. The real cross-node serialization coverage comes from the write-path tests (where TransportShardBulkAction extends TransportAction and goes through SecurityFilter on data nodes).

DarshitChanpura (approved):

PR #6347, opened 2026-07-29
feat(audit): Add API path filtering for request body logging

Taiwo435:
## Description

Add API path filtering for request body logging in audit events. Users can now selectively suppress request body capture for specific actions/paths (e.g., bulk ingestion) while preserving all other audit fields — reducing log volume without losing visibility.

**New settings:**

- `plugins.security.audit.config.action_groups.<NAME>`: Define named groups of action/path patterns (static, in `opensearch.yml`)
- `plugins.security.audit.config.body_logging_exclusions`: Reference group names or raw patterns to exclude from body logging (dynamic, changeable at runtime via `_cluster/settings`)

**Example configuration:**

```yaml
plugins.security.audit.config.action_groups.BULK: "indices:data/write/bulk*,/_bulk"
plugins.security.audit.config.action_groups.SEARCH: "indices:data/read/search*,/_search"
plugins.security.audit.config.body_logging_exclusions:
  - BULK
```

## Why

Request body logging is valuable for compliance but expensive at scale. A cluster doing 100K+ bulk writes/sec generates enormous audit volume — most of it redundant data payloads. Today the only option is a global `log_request_body: false` toggle, which is all-or-nothing.

This feature gives operators granular control: suppress bodies for high-volume bulk ingestion while still capturing bodies for searches, index creation, and admin operations that matter for investigation.

## How it works

- **Action groups** map a name (e.g., `BULK`) to a list of action patterns (`indices:data/write/bulk*`) and/or REST paths (`/_bulk`). Supports wildcards.
- **Body logging exclusions** references group names or raw patterns. When an audit event's action or REST path matches, the body field is omitted — all other fields (user, IP, indices, timestamp) are preserved.
- **Pre-compiled matcher**: Exclusion patterns are compiled into a `WildcardMatcher` at configuration time (not per-request), ensuring zero allocation overhead on the hot path.
- **Order-independent configuration**: Both `setActionGroups()` and `setBodyLoggingExclusions()` trigger a matcher rebuild, so call order doesn't matter.
- **Persistence across security index reloads**: Node-level config stored as an atomic `NodeExclusionConfig` record in `AbstractAuditLog`, re-applied on every filter reload in FGAC mode.
- **Dynamic updates**: Exclusion list changeable at runtime via `PUT _cluster/settings` — no restart required.
- **Works in both modes**: FGAC (groups + exclusions persist across security index reloads) and SSL-only/standalone (groups in `opensearch.yml`, exclusions dynamic).

## Testing

**Unit tests** (`AuditConfigBodyExclusionTest` — 25 tests):
- Exact path matching, wildcard action matching
- Action group resolution (single, multiple, both excluded)
- Empty exclusions, null inputs
- Non-existent group treated as raw pattern
- Order independence (groups set before/after exclusions)
- Settings-based initialization from `opensearch.yml`
- Case sensitivity of group names
- Path with/without query parameters
- Malformed groups (empty string, whitespace-only)
- Null action groups + null exclusion list handling
- Concurrent setter/query thread safety (3 threads, 1000 iterations)
- Serialization round-trip (JSON output)

**Integration tests** (FGAC + standalone):
- `AuditBodyLoggingExclusionTest`: Body excluded for bulk, body present for non-excluded requests, REST path exclusion, search body logged when only bulk excluded, shard-level wildcard matching, non-existent group handling, empty exclusion behavior, global toggle interaction, multiple groups
- `StandaloneAuditBodyLoggingExclusionTest`: Same coverage in SSL-only mode with dynamic runtime updates (add exclusions, clear exclusions, verify body reappears)

## Check List

- [x] Commits are signed per the DCO 

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/AuditBodyLoggingExclusionTest.java
+++ b/src/integrationTest/java/org/opensearch/security/AuditBodyLoggingExclusionTest.java
@@ -1,0 +1,145 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.security;
+
+import java.util.Map;
+
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.opensearch.security.auditlog.impl.AuditCategory;
+import org.opensearch.security.auditlog.impl.AuditMessage;
+import org.opensearch.test.framework.AuditConfiguration;
+import org.opensearch.test.framework.AuditFilters;
+import org.opensearch.test.framework.TestSecurityConfig;
+import org.opensearch.test.framework.audit.AuditLogsRule;
+import org.opensearch.test.framework.audit.TestRuleAuditLogSink;
+import org.opensearch.test.framework.cluster.ClusterManager;
+import org.opensearch.test.framework.cluster.LocalCluster;
+import org.opensearch.test.framework.cluster.TestRestClient;
+
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+
+/**
+ * Integration tests for body logging exclusions feature.
+ * Verifies that request bodies are excluded from audit events
+ * when the action/path matches configured exclusion patterns.
+ */
+public class AuditBodyLoggingExclusionTest {
+
+    private static final TestSecurityConfig.User ADMIN_USER = new TestSecurityConfig.User("admin").roles(
+        new TestSecurityConfig.Role("all_access").clusterPermissions("*").indexPermissions("*").on("*")
+    );
+
+    @ClassRule
+    public static LocalCluster cluster = new LocalCluster.Builder().clusterManager(ClusterManager.SINGLENODE)
+        .users(ADMIN_USER)
+        .authc(AUTHC_HTTPBASIC_INTERNAL)
+        .nodeSettings(
+            Map.of(
+                // Audit sink type (required for node-level exclusion parsing)
+                "plugins.security.audit.type",
+                TestRuleAuditLogSink.class.getName(),
+                // Define action groups
+                "plugins.security.audit.config.action_groups.BULK",
+                "indices:data/write/bulk*,/_bulk",
+                "plugins.security.audit.config.action_groups.SEARCH",
+                "indices:data/read/search*,/_search",
+                // Exclude BULK from body logging
+                "plugins.security.audit.config.body_logging_exclusions",
+                "BULK"
+            )
+        )
+        .audit(new AuditConfiguration(true).filters(new AuditFilters().enabledRest(true).enabledTransport(true)))
+        .build();
+
+    @Rule
+    public AuditLogsRule auditLogsRule = new AuditLogsRule();
+
+    @Test
+    public void shouldExcludeBodyForBulkRequest() {
+        try (TestRestClient client = cluster.getRestClient(ADMIN_USER)) {
+            String bulkBody = "{ \"index\": { \"_index\": \"test-idx\", \"_id\": \"1\" } }\n" + "{ \"field\": \"value\" }\n";
+            client.postJson("_bulk?refresh=true", bulkBody);
+        }
+
+        // GRANTED_PRIVILEGES event for bulk should have NO body (excluded)
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.GRANTED_PRIVILEGES) return false;
+            if (msg.getPrivilege() == null || !msg.getPrivilege().contains("indices:data/write/bulk")) return false;
+            return msg.getRequestBody() == null;
+        });
+    }
+
+    @Test
+    public void shouldStillLogBodyForNonExcludedRequest() {
+        try (TestRestClient client = cluster.getRestClient(ADMIN_USER)) {
+            client.putJson("test-idx/_doc/1", "{\"field\": \"value\"}");
+        }
+
+        // Single index request is NOT excluded — body should be present
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.GRANTED_PRIVILEGES) return false;
+            if (msg.getPrivilege() == null || !msg.getPrivilege().contains("indices:data/write/index")) return false;
+            String body = msg.getRequestBody();
+            return body != null && body.contains("value");
+        });
+    }
+
+    @Test
+    public void shouldExcludeBodyOnRestPathForBulk() {
+        try (TestRestClient client = cluster.getRestClient(ADMIN_USER)) {
+            String bulkBody = "{ \"index\": { \"_index\": \"rest-test\", \"_id\": \"1\" } }\n" + "{ \"message\": \"hello\" }\n";
+            client.postJson("_bulk", bulkBody);
+        }
+
+        // AUTHENTICATED event (REST path) should also have no body for /_bulk
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.AUTHENTICATED) return false;
+            Map<String, Object> fields = msg.getAsMap();
+            Object path = fields.get(AuditMessage.REST_REQUEST_PATH);
+            if (path == null || !path.toString().contains("_bulk")) return false;
+            return msg.getRequestBody() == null;
+        });
+    }
+
+    @Test
+    public void shouldLogBodyForSearchWhenOnlyBulkExcluded() {
+        try (TestRestClient client = cluster.getRestClient(ADMIN_USER)) {
+            client.postJson("test-idx/_search", "{\"query\": {\"match_all\": {}}}");
+        }
+
+        // Search is NOT in the exclusion list — body should be logged
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.GRANTED_PRIVILEGES) return false;
+            if (msg.getPrivilege() == null || !msg.getPrivilege().contains("indices:data/read/search")) return false;
+            String body = msg.getRequestBody();
+            return body != null && body.contains("match_all");
+        });
+    }
+
+    @Test
+    public void shouldMatchShardLevelActionWithWildcard() {
+        // Bulk produces shard-level actions like "indices:data/write/bulk[s][p]"
+        // Our pattern "indices:data/write/bulk*" should match it
+        try (TestRestClient client = cluster.getRestClient(ADMIN_USER)) {
+            String bulkBody = "{ \"index\": { \"_index\": \"shard-test\", \"_id\": \"1\" } }\n" + "{ \"field\": \"shard-level\" }\n";
+            client.postJson("_bulk?refresh=true", bulkBody);
+        }
+
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.GRANTED_PRIVILEGES) return false;
+            if (msg.getPrivilege() == null) return false;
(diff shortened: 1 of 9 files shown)


src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -266,7 +302,7 @@ public void logMissingPrivileges(String privilege, TransportRequest request, Tas
             resolver,
             clusterService,
             settings,
-            auditConfigFilter.shouldLogRequestBody(),
+            auditConfigFilter.shouldLogRequestBody() && !auditConfigFilter.isBodyExcluded(privilege != null ? privilege : action),
```
DarshitChanpura, line 305: The exclusion match runs against two different string namespaces depending on the layer: REST-layer events match `request.path()` (e.g. `/_bulk`), while every transport-layer event here matches the transport action (e.g. `indices:data/write/bulk[s][p]`). So a group has to contain **both** a path pattern and an action wildcard to suppress a request across both layers — which is why the `BULK` example is `indices:data/write/bulk*,/_bulk`.

Could we document this two-namespace model prominently? As written, setting only `/_bulk` silently leaves transport-level bodies logged, and paths for indexed requests (`/my-index/_doc/1`, `/my-index/_search`) contain the index name so a fixed path pattern won't match them — the user has to target the action instead.

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
```diff
@@ -322,7 +322,7 @@ private void logBulkItem(
         if (!filteredHeaders.isEmpty()) {
             msg.addRestHeaders(filteredHeaders, false, null);
         }
-        if (filter.shouldLogRequestBody() && innerRequest instanceof IndexRequest) {
+        if (filter.shouldLogRequestBody() && !filter.isBodyExcluded(action) && innerRequest instanceof IndexRequest) {
```
DarshitChanpura, line 328: This tests `isBodyExcluded(action)` against the **parent** bulk action, not the per-item action, so bulk body exclusion is all-or-nothing per request — you can't keep index-item bodies while dropping delete-item bodies. That's a reasonable limitation, but the "action path filtering" naming implies finer granularity than exists; worth a note in the docs.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -486,6 +509,32 @@ private void warnIfAuthCategoriesEnabled(Settings settings) {
         }
     }
 
+    /**
+     * Parses action groups from opensearch.yml settings. Each group is defined as:
+     * {@code plugins.security.audit.config.action_groups.<NAME> = pattern1,pattern2,...}
+     * Handles both list-style and comma-separated value formats.
+     */
+    private static Map<String, List<String>> parseActionGroups(Settings settings) {
```
DarshitChanpura, line 516: This parse-and-split logic is duplicated three times: here, in `AuditConfig.Filter.from(Settings)`, and (Map variant) in `from(Map)`. Suggest extracting a single static helper so the copies don't drift over time.

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
```diff
@@ -154,6 +162,14 @@ public static class Filter {
...
         private final WildcardMatcher ignoredCustomHeadersMatcher;
         private volatile WildcardMatcher ignoredUrlParamsMatcher;
+        private volatile Map<String, List<String>> actionGroups = Collections.emptyMap();
+        private volatile WildcardMatcher bodyExclusionMatcher = WildcardMatcher.NONE;
+        @JsonProperty("body_logging_exclusions")
+        @JsonInclude(JsonInclude.Include.NON_EMPTY)
+        private volatile List<String> bodyLoggingExclusions = Collections.emptyList();
+        @JsonProperty("action_groups")
+        @JsonInclude(JsonInclude.Include.NON_EMPTY)
+        private volatile Map<String, List<String>> actionGroupsSerialized = Collections.emptyMap();
```
DarshitChanpura: `actionGroups` and `actionGroupsSerialized` always hold the identical reference (`setActionGroups` assigns both). Only the annotated one serializes (private field + no getter + Jackson `PUBLIC_ONLY`), so there's no duplicate-key bug — but a single `@JsonProperty`-annotated field would be clearer.

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
```diff
@@ -133,6 +135,12 @@ public static AuditConfig from(final Settings settings) {
     public static class Filter {
         private static final Logger log = LogManager.getLogger(Filter.class);
         private static Set<String> FIELDS = DefaultObjectMapper.getFields(Filter.class);
+        /** Settings key for body logging exclusions — defined here to avoid circular init with SecuritySettings.
+         *  Must match {@code SecuritySettings.AUDIT_BODY_LOGGING_EXCLUSIONS}. */
+        static final String BODY_LOGGING_EXCLUSIONS_KEY = "plugins.security.audit.config.body_logging_exclusions";
```
DarshitChanpura, line 140: These constants match the `SecuritySettings` keys today. Since they're deliberately copied to dodge a static-init cycle, a future rename would break the feature with no error — a tiny unit test asserting `BODY_LOGGING_EXCLUSIONS_KEY.equals(SecuritySettings.AUDIT_BODY_LOGGING_EXCLUSIONS.getKey())` (and the prefix) would lock it down.

src/test/java/org/opensearch/security/auditlog/config/AuditConfigBodyExclusionTest.java
```diff
@@ -0,0 +1,319 @@
...
+    public void testSetBodyLoggingExclusionsNull() {
+        AuditConfig.Filter filter = AuditConfig.Filter.from(Settings.EMPTY);
+        filter.setActionGroups(Map.of("BULK", List.of("/_bulk")));
+        filter.setBodyLoggingExclusions(null); // null → treated as empty
+
+        assertFalse("nothing should be excluded when exclusions are null", filter.isBodyExcluded("/_bulk"));
+    }
+
+    @Test
+    public void testConcurrentSettersAndQuery() throws InterruptedException {
```
DarshitChanpura: This only asserts no exception is thrown; it doesn't verify the resulting matcher state is correct, so it over-claims thread-safety. Either strengthen the assertion (check `isBodyExcluded` returns the expected result after the updates settle) or rename to reflect that it's a "does not throw" smoke test.

Note: correctness today relies on the external invariant that only one setter is called on a *published* `Filter` instance, serially (the class isn't internally synchronized). A comment on the setters documenting that invariant would help prevent a future caller from wiring `action_groups` as a second dynamic setting and breaking it silently.

src/test/java/org/opensearch/security/auditlog/config/AuditConfigBodyExclusionTest.java
```diff
@@ -0,0 +1,319 @@
...
+        // "DOES_NOT_EXIST" is not a group name — treated as a literal pattern
+        filter.setBodyLoggingExclusions(List.of("DOES_NOT_EXIST"));
+
+        // Only matches the literal string "DOES_NOT_EXIST"
+        assertTrue("literal match should work", filter.isBodyExcluded("DOES_NOT_EXIST"));
+        assertFalse("/_bulk should NOT match", filter.isBodyExcluded("/_bulk"));
+    }
+
+    @Test
+    public void testBodyExcludedGroupsSetAfterExclusionsDoesNotAutoUpdate() {
```
DarshitChanpura: Name says "does not auto update" and the comment cites "lazy resolution," but the implementation is eager — `setActionGroups` rebuilds the matcher, which is why this passes. The test is correct; suggest renaming to e.g. `testGroupsSetAfterExclusionsStillResolves` and fixing the comment.

src/integrationTest/java/org/opensearch/security/AuditBodyLoggingExclusionTest.java
```diff
@@ -0,0 +1,257 @@
...
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.GRANTED_PRIVILEGES) return false;
+            if (msg.getPrivilege() == null || !msg.getPrivilege().contains("indices:data/write/index")) return false;
+            String body = msg.getRequestBody();
+            return body != null && body.contains("bodies-on");
+        });
+    }
+
+    @Test
+    public void shouldNotLogAnyBodyWhenGlobalToggleOff() {
```
DarshitChanpura: This asserts the opposite of its name — the comment notes the global toggle can't be flipped in this FGAC setup, so it just verifies a non-excluded request has a body. Same issue with `shouldLogBodyWhenGlobalToggleOffRegardlessOfExclusions` above. Rename to reflect what's actually checked, or move the real coverage to the Standalone file (which does toggle it via `_cluster/settings`).

src/integrationTest/java/org/opensearch/security/AuditBodyLoggingExclusionTest.java
```diff
@@ -0,0 +1,257 @@
...
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.GRANTED_PRIVILEGES) return false;
+            if (msg.getPrivilege() == null || !msg.getPrivilege().contains("indices:data/read/search")) return false;
+            String body = msg.getRequestBody();
+            return body != null && body.contains("match_all");
+        });
+    }
+
+    @Test
+    public void shouldHandleRawPatternWithoutGroup() {
```
DarshitChanpura: Because the FGAC cluster's exclusion config is fixed at class level, this and several sibling tests (`shouldLogAllBodiesWhenExclusionListEmpty`, `shouldHandleMultipleGroupsInExclusions`, `shouldTreatNonExistentGroupAsRawPattern`) all reduce to "a single index PUT has a body" and don't exercise the branch their name implies. The `Standalone` file already covers raw patterns, empty lists, non-existent groups, and multiple groups properly via dynamic `_cluster/settings` updates. Suggest trimming these to avoid overstating coverage.

DarshitChanpura (commented):
Nice feature — the design is clean, the hot path is a single pre-compiled matcher, and it fails open (over-logs rather than silently dropping audit data during a transient config gap), which is the right default here. The unit tests in `AuditConfigBodyExclusionTest` are genuinely thorough.

One substantive ask before merge (documenting the matching model — see inline), plus some cleanup and test-hygiene notes. Nothing architectural.

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
Taiwo435: i will add documentation

src/main/java/org/opensearch/security/filter/AuditActionFilter.java
Taiwo435: noted

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
Taiwo435: noted

src/main/java/org/opensearch/security/auditlog/config/AuditConfig.java
Taiwo435: noted

src/integrationTest/java/org/opensearch/security/AuditBodyLoggingExclusionTest.java
Taiwo435: Addressed

src/integrationTest/java/org/opensearch/security/AuditBodyLoggingExclusionTest.java
Taiwo435: Addressed

src/test/java/org/opensearch/security/auditlog/config/AuditConfigBodyExclusionTest.java
Taiwo435: Addressed

DarshitChanpura (approved):
LGTM!

DarshitChanpura:
@Taiwo435 Let's raise a doc PR for this.

Issue #6119, opened 2026-04-27
Backend workspace collaborator management via security plugin resource sharing


cwperks:
## Summary

Workspace collaborator management in OpenSearch Dashboards is currently implemented entirely in the frontend — permissions are stored as a field on the workspace saved object and enforced by OSD server-side wrappers. This means any direct API access to `.kibana` bypasses workspace access control entirely.

This issue tracks moving workspace collaborator management to the OpenSearch backend using the security plugin's resource sharing framework, ensuring authorization is enforced at the OpenSearch level regardless of the access path.

## Background

Today, workspace access control works as follows:
- Workspace permissions (collaborators + access levels) are stored in the `permissions` field on the workspace saved object in `.kibana`
- OSD's `WorkspaceSavedObjectsClientWrapper` reads this field and filters saved object operations based on the requesting user's permissions
- The `dashboardAdmin.users` setting in `opensearch_dashboards.yml` statically defines which users are workspace admins
- Advanced settings per workspace are managed via a workspace-scoped config saved object

This is a frontend-only security boundary. The OpenSearch backend has no awareness of workspace collaborators or access levels.

## Compatibility with Existing Workspace Properties

This proposal preserves the core properties of the current workspace implementation:

- **Saved objects in multiple workspaces** — The `workspaces` array field on saved objects is unchanged. A visualization can still belong to `["ws-analytics", "ws-executive"]`. The backend DLS query checks this field to determine visibility.
- **Saved objects across multiple data sources** — Workspaces can continue to associate with multiple OpenSearch clusters as data sources. The workspace collaborator management and access control lives on the **default (local) cluster** — the one running the security plugin. Remote data sources are not affected; user permissions on remote clusters are still governed by those clusters' own security configurations.
- **Default cluster is mandatory** — The local OpenSearch cluster with the security plugin installed is the authoritative source for workspace collaborator records. This is a requirement for backend enforcement. OSD must be connected to a local cluster with the security plugin and the `opensearch-workspaces` plugin installed for backend workspace access control to function.

### 1. Register workspace as a resource type in the security plugin

Create a backend plugin (`opensearch-workspaces`) that implements `ResourceSharingExtension` to register `workspace` as a protected resource type. When a workspace saved object is created in `.kibana`, the security plugin's `ResourceIndexListener` automatically creates a sharing record in the sharing index.

The sharing record stores:
- `created_by` — the workspace owner
- `share_with` — collaborators organized by access level

### 2. Define workspace access levels

[...]

PR #6348, opened 2026-07-29
Graduate resource sharing feature out of experimental

DarshitChanpura:
### Description

Graduates the **Resource Sharing & Access Control** feature out of experimental.

**Breaking change (no deprecated fallback):** the feature-flag settings are renamed to drop the `.experimental` segment:

| Before | After |
|---|---|
| `plugins.security.experimental.resource_sharing.enabled` | `plugins.security.resource_sharing.enabled` |
| `plugins.security.experimental.resource_sharing.protected_types` | `plugins.security.resource_sharing.protected_types` |

Only the string *values* in `ConfigConstants` change — the constant names (`OPENSEARCH_RESOURCE_SHARING_ENABLED`, `OPENSEARCH_RESOURCE_SHARING_PROTECTED_TYPES`) are unchanged, so downstream Java references are unaffected.

Also:
- Removes the `@opensearch.experimental` javadoc annotations from the resource-sharing SPI and implementation classes (leaves the unrelated `SecurityConfigExtension` marker in place).
- Updates `RESOURCE_SHARING_AND_ACCESS_CONTROL.md` (setting keys + drops the "experimental" note).

Because the rename is breaking and has no fallback, this must ship in **3.9** together with the downstream references. Coordinated changes are being made to:
- `documentation-website` (setting keys + removal of experimental warning callouts)
- plugin CIs that toggle the flag: `reporting`, `anomaly-detection`, `ml-commons`, `notifications`, `alerting`, `flow-framework`, `security-analytics`

Draft while those companion PRs are prepared and reviewed.

### Related Issues
Related to https://github.com/opensearch-project/security/issues/4500

### Check List

### DIFF
--- a/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
+++ b/RESOURCE_SHARING_AND_ACCESS_CONTROL.md
@@ -30,7 +30,7 @@ The **Resource Sharing and Access Control** feature in OpenSearch Security Plugi
 
 This feature ensures **secure** and **controlled** access to shareableResources while leveraging existing **index-level authorization** in OpenSearch.
 
-NOTE: This feature is marked as **`@opensearch.experimental`** and can be toggled using the feature flag: **`plugins.security.experimental.resource_sharing.enabled`**, which is **disabled by default**.
+NOTE: This feature can be toggled using the feature flag: **`plugins.security.resource_sharing.enabled`**, which is **disabled by default**.
 
 
 ## **2. What are the Components?**
@@ -142,8 +142,8 @@ integTest {
     ...
     node.setting("plugins.security.system_indices.enabled", "true")
     if (System.getProperty("resource_sharing.enabled") == "true") {
-        node.setting("plugins.security.experimental.resource_sharing.enabled", "true")
-        node.setting("plugins.security.experimental.resource_sharing.protected_types", "[\"anomaly-detector\", \"forecaster\"]")
+        node.setting("plugins.security.resource_sharing.enabled", "true")
+        node.setting("plugins.security.resource_sharing.protected_types", "[\"anomaly-detector\", \"forecaster\"]")
     }
     ...
 }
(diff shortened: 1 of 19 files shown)


cwperks:
@DarshitChanpura until there's a clear path with workspaces like proposed in https://github.com/opensearch-project/security/issues/6119 I don't think this should be promoted out of experimental.

Let's get a consistent platform level experience for sharing.

I'm very strongly supportive of the ownership model where owner gets to choose sharing and I think the fact that workspaces was developed in a way where workspaces have owners that invite collaborators is testament to that being the desired model.

Workspace collaborators being controlled by the frontend is a problem and we need to figure out the path forward that works with workspaces desire to be supported with MDS (Multiple Data Sources). If that problem can be solved then there's clear line-of-sight to graduating from experimental.

Edit: @DarshitChanpura can we focus the efforts on adding a share button inside a dashboards plugin (like reporting) that at least with multi-tenancy is tenant aware. idk if reporting plugin is workspace aware atm and I know some teams had this notion of workspace acl (how does that work?)

If we can show MDS working with resource sharing where the currently toggled cluster supports resource sharing and share button appears than that would probably be a good place to begin. We certainly are going to want to work backwards from UX so that these share APIs can be conveniently called from the frontend and we should also have something similar to https://github.com/opensearch-project/security/pull/6107 to power new experiences on Dashboards where we can enable/disable actions based on user permissions.

DarshitChanpura:
> @DarshitChanpura can we focus the efforts on adding a share button inside a dashboards plugin (like reporting) that at least with multi-tenancy is tenant aware.

Hahaa, here is the centralized share button PR: https://github.com/opensearch-project/security-dashboards-plugin/pull/2491. I've used AD as example but you get the gist.

> @DarshitChanpura until there's a clear path with workspaces

While I agree that workspaces could leverage from the ownership model, I disagree on not taking this feature out of experimental. We should not delay this feature by tying it with workspaces improvement.

cwperks:
> While I agree that workspaces could leverage from the ownership model, I disagree on not taking this feature out of experimental. We should not delay this feature by tying it with workspaces improvement.

No they need to be compatible and I would block on that. Let's figure out the path forward.

DarshitChanpura:
Related workspaces PR has been merged and this one is unblocked now

RyanL1997 (approved):

PR #6349, opened 2026-07-29
feat(audit): Enrich Log4j MDC (Mapped Diagnostic Context) with audit event attributes.

Taiwo435:
## Description

Enrich Log4j MDC (Mapped Diagnostic Context) with audit event attributes before each log call in `Log4JSink`, enabling operators to use Log4j's native `RoutingAppender` to split audit logs by category, user, action, or request type into separate files.

**Four MDC keys added:**
- `audit_category`: event type (e.g., `GRANTED_PRIVILEGES`, `FAILED_LOGIN`, `COMPLIANCE_DOC_WRITE`)
- `audit_action`: the privilege/action being audited (e.g., `indices_data_write_index`)
- `audit_user`: the effective user for the request
- `audit_request_type`: the transport request class (e.g., `IndexRequest`, `BulkShardRequest`)

### Why

Operators running high-volume clusters need to route different audit event types to different destinations — security-critical events (`FAILED_LOGIN`, `MISSING_PRIVILEGES`) to a SIEM, compliance events to long-term storage, and routine `GRANTED_PRIVILEGES` to a rotated local file. Today, all events go to the same log file, requiring post-processing to separate them.

With MDC enrichment, a standard Log4j `RoutingAppender` config handles this natively:

```xml
<Routing name="AuditRouter">
  <Routes pattern="$${ctx:audit_category}">
    <Route ref="SecurityAlerts" key="FAILED_LOGIN"/>
    <Route ref="ComplianceLogs" key="COMPLIANCE_DOC_WRITE"/>
    <Route ref="DefaultAudit"/>
  </Routes>
</Routing>
```

### How it works

- `Log4JSink.doStore()` sets 4 MDC keys before logging, removes them after (in a `finally` block)
- Values are sanitized for filename safety via a pre-compiled regex (colons, slashes, brackets, etc. → `_`)
- MDC keys are owned by the audit sink — only our 4 keys are removed after logging, preserving any pre-existing MDC entries from other plugins
- Disabled sinks skip MDC entirely (no unnecessary ThreadContext manipulation)

WARNING: audit_user should NOT be used as a RoutingAppender routing key for file paths — it is unbounded/attacker-influenceable and could cause inode/disk exhaustion. Use audit_category (bounded enum) instead.

### Testing

Unit tests (`Log4JSinkMdcTest`) — 12 tests covering:
- MDC contains correct category/action/user/request_type during log call
- Action values sanitized for filename safety (colons, slashes, brackets → underscores)
- Null fields default to `"unknown"`
- MDC keys removed after `doStore()` completes (not `clearMap()`)
- Different categories produce different MDC values (routing works)
- Concurrent threads do not leak MDC between events
- Long values preserved without truncation
- Special characters: only filesystem-dangerous chars sanitized
- Disabled sink skips MDC entirely
- Multiple sequential calls each get fresh MDC
- Multi-chunk split messages all share same MDC values
- MDC cleaned up even when serialization produces errors

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
+++ b/src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
@@ -11,19 +11,32 @@
 
 package org.opensearch.security.auditlog.sink;
 
+import java.util.regex.Pattern;
+
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.ThreadContext;
 
 import org.opensearch.common.settings.Settings;
 import org.opensearch.security.auditlog.impl.AuditMessage;
 
 public final class Log4JSink extends AuditLogSink {
 
+    // MDC key constants — single source of truth for put/remove consistency
+    static final String MDC_CATEGORY = "audit_category";
+    static final String MDC_ACTION = "audit_action";
+    static final String MDC_USER = "audit_user";
+    static final String MDC_REQUEST_TYPE = "audit_request_type";
+
+    /** Characters unsafe for filenames + control chars — pre-compiled for performance. */
+    private static final Pattern UNSAFE_MDC_CHARS = Pattern.compile("[:/\\\\*?\"<>|\\[\\]\\p{Cntrl}]");
+
     final Logger auditLogger;
     final String loggerName;
     final Level logLevel;
     final boolean enabled;
+    final boolean mdcRoutingEnabled;
     final Integer maximumIndexCharactersPerMessage;
 
     public Log4JSink(final String name, final Settings settings, final String settingsPrefix, AuditLogSink fallbackSink) {
@@ -36,6 +49,7 @@ public final class Log4JSink extends AuditLogSink {
             Integer.MAX_VALUE
         );
         enabled = auditLogger.isEnabled(logLevel);
+        mdcRoutingEnabled = settings.getAsBoolean(settingsPrefix + ".log4j.enable_mdc_routing", false);
     }
 
     public boolean isHandlingBackpressure() {
@@ -44,8 +58,43 @@ public final class Log4JSink extends AuditLogSink {
 
     public boolean doStore(final AuditMessage msg) {
         if (enabled) {
-            msg.toJsonSplitIndices(maximumIndexCharactersPerMessage).forEach(message -> auditLogger.log(logLevel, message));
+            if (mdcRoutingEnabled) {
+                try {
+                    // Push audit attributes into Log4j MDC so operators can use
+                    // RoutingAppender with $${ctx:audit_category} etc. to split logs.
+                    // WARNING: Do not use audit_user as a RoutingAppender routing key for file paths —
+                    // it is unbounded/attacker-influenceable and could cause inode/disk exhaustion.
+                    // Prefer audit_category (bounded enum) for file-based routing.
+                    ThreadContext.put(MDC_CATEGORY, sanitizeForMdc(msg.getCategory() != null ? msg.getCategory().name() : null));
+                    ThreadContext.put(MDC_ACTION, sanitizeForMdc(msg.getPrivilege()));
+                    ThreadContext.put(MDC_USER, sanitizeForMdc(msg.getEffectiveUser()));
+                    ThreadContext.put(MDC_REQUEST_TYPE, sanitizeForMdc(msg.getRequestType()));
+
+                    msg.toJsonSplitIndices(maximumIndexCharactersPerMessage).forEach(message -> auditLogger.log(logLevel, message));
+                } finally {
+                    ThreadContext.remove(MDC_CATEGORY);
+                    ThreadContext.remove(MDC_ACTION);
+                    ThreadContext.remove(MDC_USER);
+                    ThreadContext.remove(MDC_REQUEST_TYPE);
+                }
+            } else {
+                msg.toJsonSplitIndices(maximumIndexCharactersPerMessage).forEach(message -> auditLogger.log(logLevel, message));
+            }
         }
         return true;
+    }
+
+    /**
+     * Sanitizes a value for safe use in Log4j MDC, particularly when operators
+     * use MDC values in RoutingAppender file paths. Replaces characters that are
+     * problematic in filenames (e.g., colons in action names like "indices:data/write/index")
+     * and control characters that could enable log-forging attacks.
+     * These MDC keys are owned by the audit sink — no other code should set them.
+     */
+    static String sanitizeForMdc(String value) {
+        if (value == null || value.isEmpty()) {
+            return "unknown";
+        }
+        return UNSAFE_MDC_CHARS.matcher(value).replaceAll("_");
     }
 }
--- a/src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
+++ b/src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
@@ -20,6 +20,7 @@ import org.apache.logging.log4j.ThreadContext;
 
 import org.opensearch.common.settings.Settings;
 import org.opensearch.security.auditlog.impl.AuditMessage;
+import org.opensearch.security.support.ConfigConstants;
 
 public final class Log4JSink extends AuditLogSink {
 
@@ -49,7 +50,7 @@ public final class Log4JSink extends AuditLogSink {
             Integer.MAX_VALUE
         );
         enabled = auditLogger.isEnabled(logLevel);
-        mdcRoutingEnabled = settings.getAsBoolean(settingsPrefix + ".log4j.enable_mdc_routing", false);
+        mdcRoutingEnabled = settings.getAsBoolean(settingsPrefix + "." + ConfigConstants.SECURITY_AUDIT_LOG4J_ENABLE_MDC_ROUTING, false);
     }
 
     public boolean isHandlingBackpressure() {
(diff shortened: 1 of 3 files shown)


src/test/java/org/opensearch/security/auditlog/sink/Log4JSinkMdcTest.java
```diff
@@ -0,0 +1,369 @@
...
+    public void testConcurrentThreadsDoNotLeakMdc() throws Exception {
+        Log4JSink sink = createSink();
+
+        // Two threads logging simultaneously — MDC must not bleed between them
+        Thread thread1 = new Thread(() -> {
+            AuditMessage msg = new AuditMessage(AuditCategory.GRANTED_PRIVILEGES, clusterService, null, null);
+            msg.addEffectiveUser("user-thread-1");
+            sink.doStore(msg);
+            // Our MDC keys must be cleared on this thread after
+            assertThat(ThreadContext.get("audit_category"), is((String) null));
```
DarshitChanpura: **Swallowed in-thread assertions (verified empirically).** These `assertThat(...)` calls run on the spawned worker thread, so an `AssertionError` here never propagates to the JUnit thread and can't fail the test.

I confirmed this on the branch: inserting a guaranteed-false assertion into this worker-thread body still passes the test. So the in-thread cleanup checks (here and in `thread2`) are effectively dead.

The post-`join()` value assertions below *do* run, so value-bleed between events **is** covered — it's only the per-worker-thread cleanup that goes unverified. Suggestion: capture failures into a shared `AtomicReference<Throwable>` (or synchronized list) inside each thread and assert it's empty after `join()`.

src/test/java/org/opensearch/security/auditlog/sink/Log4JSinkMdcTest.java
```diff
@@ -0,0 +1,369 @@
...
+            .put("plugins.security.audit.config.log4j.level", "INFO")
+            .put("plugins.security.audit.config.log4j.maximum_index_characters_per_message", 255)
+            .build();
+        Log4JSink sink = new Log4JSink("test", settings, "plugins.security.audit.config", null);
+
+        AuditMessage msg = new AuditMessage(AuditCategory.GRANTED_PRIVILEGES, clusterService, null, null);
+        msg.addEffectiveUser("exception-test-user");
+        msg.addIndices(new String[] { "index-a", "index-b", "index-c" });
+
+        sink.doStore(msg);
```
DarshitChanpura: **This test never actually throws (verified empirically).** The name is `testMdcCleanedUpEvenWhenLoggingThrows`, but nothing on this path throws — it's a normal multi-index store. I confirmed on the branch by asserting that `doStore` throws: that assertion fails, so the `finally` exception path (the whole point of the test) is never exercised.

Suggestion: inject a failing appender/layout so `auditLogger.log(...)` actually throws and then assert the MDC keys are still removed, or rename to something like `testMdcCleanedUpAfterMultiChunkStore` to match what it verifies.

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
```diff
@@ -44,8 +50,36 @@ public boolean isHandlingBackpressure() {
...
     public boolean doStore(final AuditMessage msg) {
         if (enabled) {
-            msg.toJsonSplitIndices(maximumIndexCharactersPerMessage).forEach(message -> auditLogger.log(logLevel, message));
+            try {
+                // Push audit attributes into Log4j MDC so operators can use
+                // RoutingAppender with $${ctx:audit_category} etc. to split logs.
+                // Values are sanitized for safe use in filenames (colons in action names etc.)
+                ThreadContext.put("audit_category", sanitizeForMdc(msg.getCategory() != null ? msg.getCategory().name() : "UNKNOWN"));
+                ThreadContext.put("audit_action", sanitizeForMdc(msg.getPrivilege()));
+                ThreadContext.put("audit_user", sanitizeForMdc(msg.getEffectiveUser()));
```
DarshitChanpura: **Security note — routing key safety.** `audit_user` is unbounded and attacker-influenceable, so using it as a `RoutingAppender` key (`$${ctx:audit_user}`) lets a caller mint arbitrarily many distinct filenames -> inode/disk exhaustion. The example in the PR description safely routes on `audit_category` (a bounded enum); worth calling out explicitly in the docs/PR that `audit_user` should not be used as an unbounded routing key. (`audit_action`/privilege and `audit_request_type` are effectively bounded — action strings and request class names — so those are lower risk.)

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
```diff
@@ -44,8 +50,36 @@ public boolean isHandlingBackpressure() {
 
     public boolean doStore(final AuditMessage msg) {
         if (enabled) {
-            msg.toJsonSplitIndices(maximumIndexCharactersPerMessage).forEach(message -> auditLogger.log(logLevel, message));
+            try {
+                // Push audit attributes into Log4j MDC so operators can use
+                // RoutingAppender with $${ctx:audit_category} etc. to split logs.
+                // Values are sanitized for safe use in filenames (colons in action names etc.)
+                ThreadContext.put("audit_category", sanitizeForMdc(msg.getCategory() != null ? msg.getCategory().name() : "UNKNOWN"));
```
DarshitChanpura: **Fallback casing is inconsistent.** `audit_category` falls back to `"UNKNOWN"` (uppercase) here, while the other three fall back to `"unknown"` (lowercase, via `sanitizeForMdc`). An operator keying a route on the "missing" value has to remember both spellings. Passing a null category name straight through `sanitizeForMdc` (which already returns `"unknown"`) would unify it.

Minor, related: the four key names are string literals duplicated across the `put` here and the `remove` calls in the `finally`. Extracting `private static final String` constants removes any risk of put/remove drift and documents the owned key set in one place.

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
```diff
@@ -44,8 +50,36 @@ public boolean isHandlingBackpressure() {
 
     public boolean doStore(final AuditMessage msg) {
```
DarshitChanpura, line 60: **Optional — unconditional per-event cost.** This runs 4x `put` + 4x regex + 4x `remove` on every audit event regardless of whether any operator has configured a `RoutingAppender`. On high-volume clusters that's overhead nobody using the default config benefits from. Consider gating enrichment behind a setting (e.g. `...log4j.enable_mdc_routing`, default off) so those clusters pay nothing.

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
```diff
@@ -44,8 +50,36 @@ public boolean isHandlingBackpressure() {
...
+     * Sanitizes a value for safe use in Log4j MDC, particularly when operators
+     * use MDC values in RoutingAppender file paths. Replaces characters that are
+     * problematic in filenames (e.g., colons in action names like "indices:data/write/index").
+     * These MDC keys are owned by the audit sink — no other code should set them.
+     */
+    private static String sanitizeForMdc(String value) {
+        if (value == null || value.isEmpty()) {
+            return "unknown";
+        }
+        return UNSAFE_FILENAME_CHARS.matcher(value).replaceAll("_");
```
DarshitChanpura: **Optional — control chars not sanitized.** The regex strips filename-dangerous chars but not control characters (`\n`, `\r`). Since `audit_user` can be attacker-influenced, a value containing newlines flows through unchanged; if an operator surfaces `%X{audit_user}` in a pattern layout this becomes a log-forging vector. Low severity for a routing key, but consider stripping control characters too.

DarshitChanpura (commented):
Nice, focused feature — the `finally`-based cleanup, removing only the sink's own four keys (not `clearMap()`), and skipping work when the sink is disabled are all the right calls, and pre-compiling the `Pattern` is good.

Two test-quality issues below are worth fixing before merge — I pulled the branch and empirically confirmed both (details in the inline comments). There's also a security note on routing keys, plus some optional polish. Overall a useful addition.

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
Taiwo435: Addressed

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
Taiwo435: Addressed

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
Taiwo435: Addressed

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
Taiwo435: Addressed

src/test/java/org/opensearch/security/auditlog/sink/Log4JSinkMdcTest.java
Taiwo435: Addressed

src/test/java/org/opensearch/security/auditlog/sink/Log4JSinkMdcTest.java
Taiwo435: Addressed

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
```diff
@@ -36,6 +49,7 @@ public Log4JSink(final String name, final Settings settings, final String settin
             Integer.MAX_VALUE
         );
         enabled = auditLogger.isEnabled(logLevel);
+        mdcRoutingEnabled = settings.getAsBoolean(settingsPrefix + ".log4j.enable_mdc_routing", false);
```
cwperks: Can we refer to a whole constant here? This looks like something that should be added to https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/ConfigConstants.java.

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
```diff
@@ -44,8 +58,43 @@ public boolean isHandlingBackpressure() {
...
+                    // WARNING: Do not use audit_user as a RoutingAppender routing key for file paths —
+                    // it is unbounded/attacker-influenceable and could cause inode/disk exhaustion.
+                    // Prefer audit_category (bounded enum) for file-based routing.
+                    ThreadContext.put(MDC_CATEGORY, sanitizeForMdc(msg.getCategory() != null ? msg.getCategory().name() : null));
+                    ThreadContext.put(MDC_ACTION, sanitizeForMdc(msg.getPrivilege()));
+                    ThreadContext.put(MDC_USER, sanitizeForMdc(msg.getEffectiveUser()));
+                    ThreadContext.put(MDC_REQUEST_TYPE, sanitizeForMdc(msg.getRequestType()));
+
+                    msg.toJsonSplitIndices(maximumIndexCharactersPerMessage).forEach(message -> auditLogger.log(logLevel, message));
+                } finally {
```
cwperks, line 81: Let's not do this please. We really should only set headers once. If you need to do something like this then check if the header exists before doing put.

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
Taiwo435: yhh we can, i'll do that

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
Taiwo435: noted

src/main/java/org/opensearch/security/auditlog/sink/Log4JSink.java
```diff
@@ -44,8 +59,53 @@ public boolean isHandlingBackpressure() {
...
+            return "unknown";
+        }
+        return UNSAFE_MDC_CHARS.matcher(value).replaceAll("_");
+    }
+
+    /**
+     * Puts a value into the Log4j ThreadContext only if the key is not already set.
+     * This avoids blindly overwriting MDC headers that may have been set upstream.
+     */
+    private static void putIfAbsent(String key, String value) {
```
DarshitChanpura: **`putIfAbsent` preservation intent is defeated by the unconditional `remove` in `finally`.** The javadoc says this avoids "blindly overwriting MDC headers that may have been set upstream" — but if an upstream component *had* already set e.g. `audit_user`, `putIfAbsent` correctly declines to overwrite it, and then the `finally` block does `ThreadContext.remove(MDC_USER)` regardless of who set it. So the upstream value is gone after `doStore` returns — the collision case the helper claims to protect is handled incorrectly.

Low impact in practice (these `audit_*` keys are sink-owned by convention, and the class comment already says "no other code should set them"), but the code now explicitly advertises collision-safety it doesn't deliver. Two clean options:

1. **Commit to ownership (simplest):** drop `putIfAbsent`, use plain `ThreadContext.put`. The header comment already asserts these keys are the sink's.
2. **Commit to coexistence:** record which keys this call actually set, and in `finally` only remove those (restoring/leaving any pre-existing value).

Either is fine — just avoid the half-measure where put is conditional but remove is not.

DarshitChanpura (commented):
Thanks for the thorough revision — I re-reviewed at 660fceb and pulled the branch to run it (14 tests, all green). Everything from the last round is addressed: the concurrency test now propagates worker-thread failures via a synchronized list, `testMdcCleanedUpEvenWhenLoggingThrows` genuinely throws through a `ThrowingAppender` (`ignoreExceptions=false`) before asserting cleanup, control chars are stripped (`\p{Cntrl}`), the casing is unified through `sanitizeForMdc`, MDC key names are extracted to constants, and the whole path is gated behind `log4j.enable_mdc_routing` (default off). Nice work.

One new inconsistency the fix introduced — see the inline comment on `putIfAbsent`.

DarshitChanpura (approved):

PR #6352, opened 2026-07-30
feat(audit): Add audit logging for Resource Sharing feature

Taiwo435:
## Description

Add audit logging for the Resource Sharing feature (introduced in OpenSearch 3.3) so that all authorization events surrounding resource access and sharing mutations are captured. Three new audit categories provide full visibility into who accessed shared resources, who was denied, and who changed sharing permissions.

**Three new audit categories:**
- `RESOURCE_ACCESS_GRANTED`: user accessed a shared resource (allowed by ResourceAccessEvaluator)
- `RESOURCE_ACCESS_DENIED`: user tried to access a resource they don't have sharing access to
- `RESOURCE_SHARING_CHANGED`: sharing permissions were successfully modified (share, revoke, patch)

**Audit event fields added:**
- `audit_resource_id`: the specific resource being accessed or shared
- `audit_resource_type`: resource type (e.g., `saved_query`, `dashboard`)
- `audit_resource_index`: the backing index storing the resource
- `audit_resource_access_result`: `"granted"` or `"denied"`
- `audit_resource_sharing_action`: `"share"`, `"patch"`, or `"share_denied"` / `"patch_denied"` (on internal failure)
- `audit_resource_recipients_added`: who was added (for PATCH)
- `audit_resource_recipients_revoked`: who was revoked (for PATCH)
- `audit_resource_share_with`: full sharing target (for PUT)

### Why

Current audit events (`GRANTED_PRIVILEGES` / `MISSING_PRIVILEGES`) cannot distinguish between:
- "User was denied because they lack a role on that index" (normal RBAC)
- "User was denied because they don't have sharing access to that specific resource" (resource sharing)

Incident investigators need to answer: "Who accessed this shared dashboard?", "When was access revoked?", "Who changed the sharing permissions?", "Did someone try to share a resource they don't own?" 
Without dedicated categories, these events are buried in thousands of generic privilege events.

With resource sharing audit, operators can filter by `audit_resource_id` to reconstruct the full lifecycle:

```
10:00  RESOURCE_SHARING_CHANGED  | alice shared dashboard-456 with bob
14:00  RESOURCE_ACCESS_GRANTED   | bob accessed dashboard-456
16:00  RESOURCE_SHARING_CHANGED  | alice revoked bob's access
16:05  RESOURCE_ACCESS_DENIED    | bob tried to access dashboard-456
```

Failed sharing attempts (e.g., sharing a resource you don't own) are captured via `RESOURCE_ACCESS_DENIED` at the SecurityFilter level with `audit_transport_action: "cluster:admin/security/resource/share"` — the share request never reaches `ShareTransportAction` because SecurityFilter blocks it.

### How it works

**Hook 1: Access decisions (SecurityFilter)**
- In `SecurityFilter.apply0()`, inside the `resourceAccessEvaluator.evaluateAsync()` callback
- Fires `RESOURCE_ACCESS_GRANTED` or `RESOURCE_ACCESS_DENIED` **alongside** existing `GRANTED_PRIVILEGES` / `MISSING_PRIVILEGES` events (additive — zero regression)
- Extracts resource ID, type, and index from the `DocRequest`
- The `audit_transport_action` field distinguishes resource reads (`sampleresource:get`) from sharing attempts (`cluster:admin/security/resource/share`)
- Admin cert users bypass ResourceAccessEvaluator entirely (early return in SecurityFilter) — no resource access events for them

**Hook 2: Sharing mutations (ShareTransportAction)**
- Wraps the `ActionListener` in `doExecute()` to fire audit after successful mutations
- Only fires on PUT (share) and PATCH/POST (add/revoke) — not on GET (read-only)
- Fires on both success (`"share"` / `"patch"`) and internal failure (`"share_denied"` / `"patch_denied"`)
- Accesses `AuditLog` via `GuiceHolder.getAuditLog()` (interface not resolvable via Guice component binding)
- Null-safe: if `AuditLog` is unavailable, the share operation still completes

**What does NOT produce `RESOURCE_SHARING_CHANGED` events:**
- `GET /_plugins/_security/api/resource/share?resource_id=...` — read-only, no state change
- Failed sharing attempts blocked by SecurityFilter — these produce `RESOURCE_ACCESS_DENIED` instead (Hook 1 captures them)

### Design decisions

- **Additive approach**: existing `GRANTED_PRIVILEGES` and `MISSING_PRIVILEGES` events still fire unchanged — new events are additional
- **Disabled by default**: all three categories are in default `disabled_rest_categories` and `disabled_transport_categories` lists — operators must explicitly enable them
- **AUTH_ONLY_CATEGORIES**: new categories require FGAC mode (resource sharing doesn't exist without auth)
- **String parameters for recipients**: serialized at the call site via `toString()` to guarantee clean JSON output regardless of object complexity
- **checkTransportFilter on all methods**: respects `disabled_categories`, `ignore_users`, and `ignore_requests` consistently
- **GuiceHolder pattern**: `AuditLog` is an interface; Guice binds components by concrete class only, so constructor injection doesn't work — use static accessor instead

### Testing

**Unit tests (40 in AuditMessageTest):**
- Field setters: null/empty ignored, valid values stored for all 8 resource fields
- New categories parse from strings
- New categories in `AUTH_ONLY_CATEGORIES` set
- New categories in default disabled REST and transport lists
- Full message construction for all 3 categories with correct fields
- All resource fields populated simultaneously without interference
- `RESOURCE_ACCESS_DENIED` message has correct category, action, and user
- JSON serialization contains resource fields

**Integration tests (18 in ResourceSharingAuditTest):**
- PUT share → `RESOURCE_SHARING_CHANGED` with correct resource ID and action
- Shared user access → `RESOURCE_ACCESS_GRANTED`
- Unshared user access → `RESOURCE_ACCESS_DENIED`
- Additive regression: `GRANTED_PRIVILEGES` still fires alongside `RESOURCE_ACCESS_GRANTED`
- All expected fields present in sharing changed event
- GET sharing info → no `RESOURCE_SHARING_CHANGED` (read-only)
- PATCH add recipients → `RESOURCE_SHARING_CHANGED` with action "patch"
- PATCH revoke → `RESOURCE_SHARING_CHANGED`
- Failed sharing attempt (non-existent resource) → `RESOURCE_ACCESS_DENIED` from SecurityFilter
- Effective user correctly captured in granted event
- Effective user correctly captured in denied event
- Full lifecycle: share → access granted → revoke → access denied
- Resource type consistent across all event types
- Share operation succeeds even if audit is unavailable (no blocking)
- Owner accessing own resource produces access event
- Admin cert bypasses resource evaluator (no resource access events)
- Long resource IDs preserved without truncation
- Categories enabled → events produced (baseline verification)

**Regression:** Existing `ShareApiTests` pass unchanged — no break from pipeline changes.

### Check List


src/main/java/org/opensearch/security/resources/api/share/ShareTransportAction.java
```diff
@@ -53,14 +55,61 @@ protected void doExecute(Task task, ShareRequest request, ActionListener<ShareRe
...
+     * and cannot be injected by interface type.
+     */
+    private ActionListener<ResourceSharing> auditingListener(
+        ActionListener<ResourceSharing> delegate,
+        ShareRequest request,
+        Task task,
+        String sharingAction
+    ) {
+        return ActionListener.wrap(resourceSharing -> {
+            AuditLog auditLog = OpenSearchSecurityPlugin.GuiceHolder.getAuditLog();
```
DarshitChanpura: **The GuiceHolder rationale here looks incorrect, and the static hop is avoidable.**

The class comment says `AuditLog` "cannot be injected by interface type." But `GuiceHolder` itself does `@Inject ... AuditLogImpl auditLog` — the **concrete** `AuditLogImpl` is bound in Guice. `ShareTransportAction` is already `@Inject`-constructed, so you can inject `AuditLogImpl` straight into the constructor exactly the way `GuiceHolder` does:

```java
private final AuditLog auditLog;

@Inject
public ShareTransportAction(TransportService ts, ActionFilters af, ResourceAccessHandler h, AuditLogImpl auditLog) {
    super(ShareAction.NAME, ts, af, ShareRequest::new);
    this.resourceAccessHandler = h;
    this.auditLog = auditLog;
}
```

That removes the `OpenSearchSecurityPlugin` dependency, both null-checks, and makes the class unit-testable with a mock `AuditLog`. If there's a binding/lifecycle reason the static accessor is genuinely required, the comment should state the *real* reason rather than the injectability one. This also dissolves the null-safety concern that `shouldNotBlockShareWhenAuditLogUnavailable` is trying (but failing) to cover.

src/main/java/org/opensearch/security/resources/api/share/ShareTransportAction.java
```diff
@@ -53,14 +55,61 @@ protected void doExecute(Task task, ShareRequest request, ActionListener<ShareRe
...
+                );
+            }
+            delegate.onResponse(resourceSharing);
+        }, e -> {
+            AuditLog auditLog = OpenSearchSecurityPlugin.GuiceHolder.getAuditLog();
+            if (auditLog != null) {
+                auditLog.logResourceSharingChanged(
+                    request.id(),
+                    request.type(),
+                    sharingAction + "_denied",
```
DarshitChanpura: **Failure-path branch is untested and semantically questionable.**

Two issues:

1. Firing a category literally named `RESOURCE_SHARING_CHANGED` when the mutation *failed* (nothing changed) is misleading for anyone querying audit logs. Consider a distinct result field (e.g. `audit_resource_sharing_result: "failed"`) instead of overloading the action string with `_denied`.
2. **None of the 18 integration tests exercise this `e -> ...` branch.** They all cover the success path or the SecurityFilter-denied path. Please add a test that forces `share()`/`patchSharingInfo()` to fail *after* authorization (so it reaches `ShareTransportAction`), or remove the branch if that state isn't reachable in practice.

Separately: `logResourceSharingChanged` never calls `addAction(...)`, so these events carry no `audit_transport_action`, while the access events do — a minor inconsistency for investigators filtering by action.

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -1036,6 +1036,89 @@ public void logApiTokenRevoked(String tokenId, String revokedBy) {
...
+    public void logResourceSharingChanged(
+        String resourceId,
+        String resourceType,
+        String sharingAction,
+        String recipientsAdded,
+        String recipientsRevoked,
+        String shareWith,
+        Task task
+    ) {
+        if (!checkTransportFilter(AuditCategory.RESOURCE_SHARING_CHANGED, sharingAction, getUser(), null)) {
```
DarshitChanpura: **Passing `null` as the request silently skips `ignore_requests` filtering.**

In `checkTransportFilter`, the `isRequestAuditDisabled(action)` / `isRequestAuditDisabled(request.getClass().getSimpleName())` block is guarded by `request != null`. With `null` here, `RESOURCE_SHARING_CHANGED` events bypass `ignore_requests` entirely.

The PR description explicitly states "checkTransportFilter on all methods: respects `disabled_categories`, `ignore_users`, and `ignore_requests` consistently" — this path doesn't honor `ignore_requests`. You have the real `ShareRequest` available in `ShareTransportAction`; thread it through `logResourceSharingChanged` so the filter can see it. (Otherwise, drop that claim from the description.)

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
```diff
@@ -1036,6 +1036,89 @@ public void logApiTokenRevoked(String tokenId, String revokedBy) {
...
+            return;
+        }
+
+        AuditMessage msg = new AuditMessage(AuditCategory.RESOURCE_SHARING_CHANGED, clusterService, getOrigin(), Origin.TRANSPORT);
+        msg.addEffectiveUser(getUser());
```
DarshitChanpura, line 1162: **Please confirm thread-context propagation for the effective user.**

`getUser()` reads `OPENDISTRO_SECURITY_USER` from the thread-local `ThreadContext`. For `RESOURCE_SHARING_CHANGED`, this executes inside the `ActionListener` callback that runs *after* `resourceAccessHandler.share()/patchSharingInfo()` — async index operations that may resume on a different threadpool thread where the security-user transient isn't set. If so, `audit_request_effective_user` will be null.

`shouldIncludeResourceFieldsInSharingChangedEvent` asserts the user is non-null, but a single-JVM integration cluster can mask context-propagation gaps that only appear under a real transport hop. Please confirm the handler wraps its listener in a `ContextPreservingActionListener` (or equivalent). If it doesn't, this is a real data-quality bug.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/ResourceSharingAuditTest.java
```diff
@@ -0,0 +1,541 @@
...
+            return RESOURCE_TYPE.equals(msg.getAsMap().get(AuditMessage.RESOURCE_TYPE));
+        });
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.RESOURCE_ACCESS_GRANTED) return false;
+            return RESOURCE_TYPE.equals(msg.getAsMap().get(AuditMessage.RESOURCE_TYPE));
+        });
+    }
+
+    @Test
+    public void shouldNotBlockShareWhenAuditLogUnavailable() {
```
DarshitChanpura: This test's comment admits it verifies the null-`AuditLog` path only "implicitly" — but audit is fully *available* in this cluster, so it actually verifies nothing about that path. If the constructor-injection change (see `ShareTransportAction` comment) is adopted, the null-safety concern disappears and this test can go. Otherwise, inject a `NullAuditLog` or a failing audit sink to genuinely exercise the branch.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/ResourceSharingAuditTest.java
```diff
@@ -0,0 +1,541 @@
...
+
+        auditLogsRule.assertExactly(0, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.RESOURCE_ACCESS_GRANTED && msg.getCategory() != AuditCategory.RESOURCE_ACCESS_DENIED)
+                return false;
+            return resourceId.equals(msg.getAsMap().get(AuditMessage.RESOURCE_ID));
+        });
+    }
+
+    @Test
+    public void shouldHandleLongResourceIdWithoutError() {
```
DarshitChanpura: The name and comment are about long resource IDs and "no truncation," but the test uses a normal `api.createSampleResourceAs(...)` ID — it never creates a long one, so it asserts nothing about length handling. Either generate a genuinely long ID and assert the full value round-trips, or remove the test.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/ResourceSharingAuditTest.java
```diff
@@ -0,0 +1,541 @@
...
+        auditLogsRule.assertAtLeast(1, (AuditMessage msg) -> {
+            if (msg.getCategory() != AuditCategory.RESOURCE_SHARING_CHANGED) return false;
+            Map<String, Object> fields = msg.getAsMap();
+            // Full resource ID preserved — no truncation
+            return resourceId.equals(fields.get(AuditMessage.RESOURCE_ID));
+        });
+    }
+
+    @Test
+    public void shouldRespectDisabledCategoriesAtRuntime() {
```
DarshitChanpura: Name says "respect disabled categories at runtime," but the test enables all categories and asserts they appear — it never disables anything, so it tests the opposite of its name. Either rename it (e.g. `shouldProduceEventsWhenCategoriesEnabled`) or add a variant that disables a category and asserts its absence.

DarshitChanpura (commented):
**Overall this is a solid, well-scoped feature** — the additive design (new categories fire alongside `GRANTED_PRIVILEGES`/`MISSING_PRIVILEGES`), the disabled-by-default posture, and the complete plumbing through all four `AuditLog` implementations are done correctly, with the config-default serialize/deserialize tests updated in lockstep (the part people usually miss). A few things I'd like addressed before merge — nothing is a showstopper.

**Verified as safe:** the `(DocRequest) request` cast in `SecurityFilter.apply0()` is fine — `shouldEvaluate()` returns `false` unless `request instanceof DocRequest`, and the early return guarantees the cast only runs on that path.

Substantive items (see inline comments):
1. The `GuiceHolder` static-accessor rationale appears incorrect — `AuditLogImpl` (the concrete class) *is* Guice-bound, so `ShareTransportAction` can inject it directly.
2. `RESOURCE_SHARING_CHANGED` passes a `null` request to `checkTransportFilter`, which silently skips `ignore_requests` — contradicting a stated design decision.
3. The `share_denied`/`patch_denied` failure branch is semantically odd ("CHANGED" when nothing changed) and has **no test coverage**.
4. Effective-user is read from thread-local `ThreadContext` inside an async callback — please confirm context propagation so the user field isn't null under a real transport hop.

Plus three integration tests whose names/comments claim more than they assert (inline). Requesting changes on 1/3/4; the rest are quick.

src/main/java/org/opensearch/security/resources/api/share/ShareTransportAction.java
Taiwo435: Done. ShareTransportAction now injects AuditLogImpl directly in the constructor (same binding GuiceHolder uses), dropped the OpenSearchSecurityPlugin dependency and both null-checks, and rewrote the comment. The class is now unit-testable with a mock AuditLog.

src/main/java/org/opensearch/security/resources/api/share/ShareTransportAction.java
Taiwo435: All three addressed. Replaced the _denied action-string overload with a dedicated audit_resource_sharing_result field ("success"/"failed"), so the category name is no longer misleading. Added msg.addAction(sharingAction) so these events now carry audit_transport_action. And added unit coverage in 
ShareTransportActionTests — shouldAuditFailedResultWhenShareOperationFails / ...WhenPatchOperationFails force the mutation to fail after authorization, hitting the e -> … branch that integration tests couldn't reach.

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
Taiwo435: Fixed. logResourceSharingChanged now takes the real TransportRequest and passes it (not null) to checkTransportFilter, threaded through from ShareTransportAction. ignore_requests is now honored, so the PR-description claim holds.

src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java
Taiwo435: Confirmed and hardened. auditingListener now captures threadContext.newRestorableContext(true) and wraps the audit listener in ContextPreservingActionListener, so the security-user transient set at doExecute time is restored when the async callback fires after stashContext in ResourceSharingIndexHandler.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/ResourceSharingAuditTest.java
Taiwo435: Removed. As you noted, once the constructor-injection change lands the null-AuditLog path no longer exists, so the test has nothing to cover.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/ResourceSharingAuditTest.java
Taiwo435: Removed. The scenario (sharing a non-existent long id) is denied in SecurityFilter before reaching the sharing path, so it emits no sharing event to assert  against — and there's no truncation logic for it to guard. Took the "remove" option; happy to add a genuine long-id round-trip assertion instead if you'd prefer that path.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/ResourceSharingAuditTest.java
Taiwo435: Fixed to match its name. Moved into a dedicated ResourceSharingAuditRuntimeConfigTest (own cluster, to avoid leaking audit-config mutations across the shared-cluster tests).

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -420,16 +421,19 @@ private <Request extends ActionRequest, Response extends ActionResponse> void ap
             // require blocking transport threads leading to thread exhaustion and request timeouts
             // We perform the rest of the evaluation as normal if the request is not for resource-access or if the feature is disabled
             if (resourceAccessEvaluator.shouldEvaluate(request)) {
+                final DocRequest docRequest = (DocRequest) request;
                 resourceAccessEvaluator.evaluateAsync(request, action, ActionListener.wrap(response -> {
```
DarshitChanpura: let's change the signature of evaluateAsync to accept DocRequest; as the first step there is to cast request as DocRequest.
https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/privileges/ResourceAccessEvaluator.java#L84

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -420,16 +421,19 @@ private <Request extends ActionRequest, Response extends ActionResponse> void ap
...
                     if (handlePermissionCheckRequest(listener, response, action)) {
                         return;
                     }
                     if (response.isAllowed()) {
                         auditLog.logGrantedPrivileges(action, request, task);
```
DarshitChanpura: i think it should be safe to remove `auditLog.logGrantedPrivileges` logs as we are now logging with new category specific to resource-sharing.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
```diff
@@ -420,16 +421,19 @@ private <Request extends ActionRequest, Response extends ActionResponse> void ap
...
+                        auditLog.logResourceAccessGranted(action, docRequest.id(), docRequest.type(), docRequest.index(), request, task);
                         auditLog.logIndexEvent(action, request, task);
                         auditLog.logSettingsChange(action, request, task);
                         chain.proceed(task, action, request, listener);
                     } else {
+                        auditLog.logResourceAccessDenied(action, docRequest.id(), docRequest.type(), docRequest.index(), request, task);
                         handleUnauthorized.accept(response);
```
DarshitChanpura, line 435: same as above. We might need to tweak the method a little.
https://github.com/opensearch-project/security/blob/981dd864f5bd3b9be3043eddb58bc6ed9ebfd498/src/main/java/org/opensearch/security/filter/SecurityFilter.java#L405

src/main/java/org/opensearch/security/auditlog/AuditLog.java
```diff
@@ -91,6 +91,37 @@ public interface AuditLog extends Closeable {
...
+    void logResourceAccessDenied(
+        String action,
+        String resourceId,
+        String resourceType,
+        String resourceIndex,
+        TransportRequest request,
+        Task task
+    );
+
+    void logResourceSharingChanged(
```
DarshitChanpura, line 129: i see that this method logs success and failure both. Let's add a java-doc.

src/main/java/org/opensearch/security/resources/api/share/ShareTransportAction.java
```diff
@@ -53,14 +64,68 @@ protected void doExecute(Task task, ShareRequest request, ActionListener<ShareRe
...
+                "success",
+                request.getAdd() != null ? request.getAdd().toString() : null,
+                request.getRevoke() != null ? request.getRevoke().toString() : null,
+                request.getShareWith() != null ? request.getShareWith().toString() : null,
+                request,
+                task
+            );
+            delegate.onResponse(resourceSharing);
+        }, e -> {
+            auditLog.logResourceSharingChanged(
```
DarshitChanpura: code duplication with above with only one word change. can be extracted to a helper method.

DarshitChanpura (commented):
Left a few comments. please check Codecoverage.

src/main/java/org/opensearch/security/filter/SecurityFilter.java
Taiwo435: okay, i will remove auditLog.logGrantedPrivileges

src/main/java/org/opensearch/security/resources/api/share/ShareTransportAction.java
Taiwo435: noted

src/main/java/org/opensearch/security/filter/SecurityFilter.java
Taiwo435: noted

src/main/java/org/opensearch/security/filter/SecurityFilter.java
Taiwo435: i will remove logging from handleUnauthorized, let each caller log separately. 
The resource-access path log logResourceAccessDenied then call handleUnauthorized
 The normal RBAC path log logMissingPrivileges then call handleUnauthorized

DarshitChanpura (approved):

PR #6373, opened 2026-08-06
Create parent-linked sharing entries for child resources written without a user

DarshitChanpura:
### Description
`ResourceIndexListener.postIndex` requires an authenticated user in the thread context to create a resource-sharing entry. When none is present, it fails with an uncaught NPE (null user subject) or skips silently at debug level. Resources written in a genuinely user-less context — scheduled jobs running under job-scheduler (e.g. scheduled report instances), provisioning steps executed under system context — therefore never receive sharing entries and stay permanently invisible to the resource-sharing APIs, including to the owner of the parent resource that triggered them.

### Changes
- Null-safe user-subject extraction
- No user + provider declares a parent (`parentType`/`parentIdField`): the sharing entry is created by inheriting `tenant`/`created_by` from the parent's sharing record, linked via `parentType`/`parentId` so `ResourceAccessHandler` delegates evaluation to the parent
- No user + no parent: skip is logged at WARN (previously silent), making the failure mode visible to operators
- Entry-indexing failures now log at WARN instead of debug

### Testing
New integration tests (`SystemContextChildResourceTests`, sample plugin hierarchy, all passing) index resource documents through the internal node client — no user in context, mirroring plugin/system-subject writes:
- child resources receive a parent-linked entry inheriting the parent owner; sharing the parent grants access to the system-created child
- parent-less resources are skipped without an entry
- children referencing a missing parent record are skipped

### Investigation note (correction)
This PR was initially motivated by on-demand report instances missing sharing records on a 3.8.0 snapshot. Deeper investigation showed that case was actually caused by listener-attachment timing on that build: the listener attaches per-index in `onIndexModule` filtered by `protected_types`, so a type added to the dynamic setting after its index was already open never got a listener until restart (current main attaches based on the unfiltered registered set, so main appears immune). On-demand instance writes do carry the authenticated user and work once the listener is attached. The user-less gap addressed by this PR remains real for scheduled/system-context writes, as covered by the new tests.

### Companion PRs
- opensearch-project/reporting#1213 (report-instance → report-definition)
- opensearch-project/flow-framework#1455 (workflow_state → workflow)

### Category
Bug fix


DarshitChanpura:
Companion reporting PR: opensearch-project/reporting#

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resourcegroup/SystemContextChildResourceTests.java
```diff
@@ -0,0 +1,149 @@
...
+import static org.opensearch.sample.utils.Constants.RESOURCE_TYPE;
+import static org.opensearch.security.api.AbstractApiIntegrationTest.forbidden;
+import static org.opensearch.security.api.AbstractApiIntegrationTest.ok;
+import static org.opensearch.test.framework.TestSecurityConfig.User.USER_ADMIN;
+
+/**
+ * Tests sharing-entry creation for resources written WITHOUT an authenticated
+ * user in the thread context — the situation plugins are in when they index
+ * resource documents under a plugin/system subject (e.g. reporting's
+ * on-demand report instances via PluginClient, or scheduled jobs running
```
cwperks: on-demand runs in system context?

cwperks (approved):

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resourcegroup/SystemContextChildResourceTests.java
DarshitChanpura: Ah, this needs an update. No, on-demand report instances aren't user-less. PluginBaseAction stashes the context but then calls storedThreadContext.restore() before executeRequest, so the caller's authenticated user is present when the instance is indexed and postIndex attributes it normally (the missing-record report; originally chased was the listener-attachment timing issue in the investigation note, not a no-user problem). The class Javadoc was stale from that initial hypothesis; I've updated it to drop the on-demand-report example and cite the correct user-less writes this suite exercises (scheduled jobs under job-scheduler, system/provisioning-context).

PR #6374, opened 2026-08-08
Workspace aware sharing records

DarshitChanpura:
### Description

**Category:** Enhancement

Makes the resource-sharing framework **workspace-aware**, so a resource's visibility and access can be driven by the workspaces it belongs to. This is the compatibility prerequisite for graduating resource sharing (RP) out of experimental — workspaces converge onto the RP ownership model rather than the reverse.

**What changed**

- **Model:** `ResourceSharing` gains an optional, set-valued `workspaces` field (a resource may belong to multiple workspaces). Threaded through the builder, XContent (omitted when empty → byte-identical for existing records), and transport serialization. Also adds the previously-missing `StreamInput` reader + `NamedWriteable` registration so `ShareResponse` round-trips.
- **Read path (DLS):** introduces a `workspace:<id>` principal namespace. Workspace IDs on a resource are denormalized into `all_shared_principals`; a user's accessible workspaces are added to the DLS `terms` filter — so visibility via workspace membership works through the existing intersection with **no new query shape**.
- **Write path:** generalizes the single-parent access recursion in `hasPermission` into a container fan-out (hierarchical parent + workspaces). Workspaces are batched in a single `mget` (avoids N+1 GETs on the privilege hot path) and evaluated in-memory; an ancestor-scoped cycle guard prevents unbounded recursion.
- **Ingestion:** `ResourceIndexListener` reads the multi-valued workspaces field off the indexed doc; migrate API is workspace-aware and adds an idempotent backfill path for already-migrated records (transition support for workspaces that predate RP).
- **SPI:** new `ResourceSharingExtension.resolveWorkspacesForUser(...)` extension point for **trusted, server-set** workspace membership. Default returns empty (safe by default); the contract explicitly requires the source to be non-user-assertable and I/O-free.

**Old vs. new behavior:** non-workspace resources are unaffected — sharing records are byte-identical (the `workspaces` field is omitted when empty), and the workspace resolver is disabled unless a plugin opts in. The DLS filter's shape changes from a single `terms` to a `bool.should` (min_should_match=1); with an empty resolver the workspace clause is dropped, so it stays semantically equivalent (same documents match) though the query JSON differs. New behavior only activates for workspace-associated resources when a trusted resolver is registered.

### Issues Resolved

Relates to #6348 (graduate resource sharing out of experimental) and #6119 (backend workspace collaborator management). No issue fully closed by this PR — see the open items comment below.

Not a backport.

No new static front-end permissions introduced.

### Testing

- Unit tests across `ResourceSharing`, `ResourceAccessHandler`, `ResourcePluginInfo`, `ResourceSharingIndexHandler` (incl. the mget + backfill paths), and the migrate extraction.
- Integration tests on live clusters in `sample-resource-plugin`: migrate workspace-backfill (idempotent) and live-index ingestion (which empirically confirms the workspace field is read at index time on the default mapping).
- Full `resources` unit package green; Codecov patch passing.

### Check List
- [x] New functionality has been documented (design doc linked)


DarshitChanpura:
### Open items / scope notes for reviewers

**In-repo, tracked as follow-ups (not blocking this PR's mechanics):**
- A few changed lines are covered by integration tests only (e.g. the `ResourceSharingDlsUtils` workspace-principal loop, which runs under DLS integration, and `SampleResource`'s transport round-trip). Called out for visibility.
- `WORKSPACE_RESOURCE_TYPE` in the write-path fan-out is a placeholder constant; the authoritative type name comes from the workspace provider once it's registered (see cross-repo below). Degrades safely — if no provider registers that type, the workspace branch denies cleanly.

**Cross-repo dependencies (cannot land here):**
- **OpenSearch-Dashboards / `opensearch-workspaces` backend plugin** (per #6119): must register the saved-object index as a protected resource, declare its `workspacesField()`, register the `workspace` resource type, and implement the trusted `resolveWorkspacesForUser` SPI. Until then the read-path resolver is inert by design.
- **Materializing workspace collaborator records** from the existing frontend-managed ACLs (the `permissions` field on the workspace saved object) — the collaborator data source lives in OSD; this PR provides the record-creation + backfill machinery.
- **security-dashboards-plugin:** workspace share UI wiring.

**Decisions for the team (not code):**
- The exact graduation gate for #6348 — is workspace *compatibility* (this PR) sufficient, or is full end-to-end enforcement required first?
- MDS scope: there is currently no MDS/datasource handling in RP code; whether it's in scope for graduation is open.

Happy to split any of the above into tracking issues if preferred.

DarshitChanpura:
### Decision needed: where does the OpenSearch-side workspace backend live?

The `ResourceSharingExtension` SPI (registering the saved-object index as a protected resource, declaring `workspacesField()`, registering the `workspace` resource type, and providing the trusted `resolveWorkspacesForUser`) must be implemented by an **OpenSearch-side (Java) plugin**. None exists today — workspaces currently live entirely in the OSD Node.js layer — so new backend code is required regardless. The open question is *where*:

- **Option A — new `opensearch-workspaces` backend plugin** (per #6119): cleanest separation; owns workspace resource-type + membership resolution.
- **Option B — fold into an existing OpenSearch-side module** (e.g. core's `opensearch-dashboards` module): avoids a new plugin/repo but couples workspace backend logic there.

Not the security plugin's responsibility either way. Owner: workspaces team. This gates the read-path resolver going live (it is inert until implemented) and ties into the #6348 graduation gate.

DarshitChanpura:
### Follow-up: dynamic workspace membership updates (associate/dissociate)

OSD shares an object to another workspace by mutating its `workspaces` array via an **update** (`POST /api/workspaces/_associate`; `_dissociate` reverses it — verified in OpenSearch-Dashboards on 2.19 and main).

`ResourceIndexListener.postIndex` extracts workspaces only on the **create** branch. On update (`!result.isCreated()`) it calls `fetchAndUpdateResourceVisibility`, which recomputes `all_shared_principals` from the **sharing record**, not from the doc's changed `workspaces` field (this PR, `ResourceIndexListener` L94 + L136/195). So:

- associating a resource to a new workspace after creation does **not** add its `workspace:<id>` principal, and
- dissociating does **not** remove the stale one (over-sharing).

Not addressed in this PR (create + migrate-backfill are). Fix: on update, re-extract the object's `workspaces` and recompute principals (adds **and** removals). This is naturally the OpenSearch-side workspace backend's responsibility (see the "where does the backend live" decision). Flagging so it's tracked before enforcement is turned on.

cwperks:
@DarshitChanpura FYI I wanted to capture what we discussed earlier today. For the sake of this PR, I think storing associated list of workspaces on the sharing record is ok, but that being said I want outline that its an intentional decision to do this rather than having 1 sharing record per workspace.

I've started of thinking of workspaces more as "realms" where a realm is a security-entity where you can associate users and groups of users (roles) with the realm. By being associated with the realm, you can see resources globally visible within that realm (workspace).

When a user goes to share a resource, they should only be able to select users and roles associated with that realm and not other users (it actually obviates the need for https://github.com/opensearch-project/security-dashboards-plugin/issues/2285). So with a single sharing record that has a list of workspaces we can theoretically store target users and roles that are associated with one of the realms, but not the other. I think that is ok, but if we had separate sharing records per workspace (realm), then you can imagine a world where workspaces are used like multi-tenancy where each workspace has a disjoint set of collaborators.

cwperks:
> Read path (DLS): introduces a workspace:<id> principal namespace. Workspace IDs on a resource are denormalized into all_shared_principals; a user's accessible workspaces are added to the DLS terms filter — so visibility via workspace membership works through the existing intersection with no new query shape.

@DarshitChanpura you can assume that the resource doc would have a workspaces field already on it that we can filter on that field. Keep `all_shared_principals` to usernames and role names.

DarshitChanpura:
@cwperks agreed on both — captured in the design doc (§5 + §9.7a).

**Read path:** I'll keep `all_shared_principals` to usernames + role names only, and filter DLS on the resource doc's own `workspaces` field as a second `terms` clause OR'd with the principals clause:

```
should: [
  terms(all_shared_principals, [public, user:*, role:*, backend:*]),
  terms(workspaces, [<user's accessible workspace ids>])
]
```

Nice side effect: since we filter on the object's live `workspaces` field (OSD-maintained), **associate/dissociate are reflected automatically** — this resolves the dynamic-membership gap I flagged above (no listener re-projection, and we can drop the `workspace:<id>` denormalization + the migrate workspace-principal backfill).

**Follow-up code changes in this PR:** remove `workspace:<id>` from `getAllPrincipals()`/visibility seeding, drop the migrate workspace-principal backfill, and add the `workspaces` DLS clause. The write-path container fan-out is unaffected (still reads the object's `workspaces` to consult each workspace's own sharing record).

**Realm framing (Sept 3):** noted as an intentional decision — single sharing record carrying the workspace list (vs. one record per workspace). Recorded the trade-off (mixed collaborators across realms) in the doc. Makes sense that scoping share targets to the realm's users/roles obviates security-dashboards-plugin#2285.

DarshitChanpura:
Implemented the read-path change (commit 114fdc436), and merged latest `main`.

- DLS is now `bool.should[ terms(all_shared_principals), terms(workspaces) ]` (`minimum_should_match: 1`) — the `workspaces` clause filters the resource's own field against the user's trusted accessible workspaces, omitted when empty.
- `all_shared_principals` / `getAllPrincipals()` no longer emit `workspace:<id>` — usernames/roles only. The `workspaces` field on the sharing record is retained for the write-path access-level fan-out.
- Since DLS filters the live `workspaces` field, associate/dissociate are reflected automatically — no principal re-projection, which closes the dynamic-membership follow-up I flagged earlier.

Tests updated (unit + the two live-cluster integration tests: `all_shared_principals` = user/role only; `workspaces` present on both the sharing record and the resource doc). Design doc updated (§5, §9).

DarshitChanpura:
Follow-up consistency fix (a1dc64653): the DLS workspaces clause was filtering a hardcoded `workspaces` field, while ingestion reads the field named by `ResourceProvider.workspacesField()`. If a provider declared a different name, DLS would filter the wrong field and never match. DLS now resolves the declared field per index (`ResourcePluginInfo.workspacesFieldForIndex`) and omits the clause for indices that declare none. Unit + live-cluster integration tests green.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/MigrateApiTests.java
```diff
@@ -294,6 +294,91 @@ public void testMigrateTwice_shouldSkipSecondTime() {
         }
     }
 
+    @Test
+    public void testLiveIndexingStampsWorkspacesOnSharingRecord() {
+        // Steady-state: creating a resource with a workspaces field must trigger ResourceIndexListener to
+        // extract the (multi-valued) workspaces from the parsed doc via extractMultiValuedFieldFromIndexOp and
+        // store them on the sharing record (used by the write-path access-level fan-out). Workspace membership is
```
cwperks: nit: prefer to keep these comments pertinent to the code as-is and not necessarily reference a previous revision. Also conciseness with these multi-line comments.

sample-resource-plugin/src/integrationTest/java/org/opensearch/sample/resource/securityapis/MigrateApiTests.java
```diff
@@ -294,6 +294,91 @@ public void testMigrateTwice_shouldSkipSecondTime() {
...
+            // The resource doc keeps its own `workspaces` field -- this is what DLS filters on for read visibility.
+            ArrayNode docWs = (ArrayNode) resourceDoc.bodyAsJsonNode().get("_source").get("workspaces");
+            List<String> docWorkspaceIds = new ArrayList<>();
+            docWs.forEach(n -> docWorkspaceIds.add(n.asString()));
+            assertThat(docWorkspaceIds, containsInAnyOrder("ws-a", "ws-b"));
+        }
+    }
+
+    @Test
+    public void testMigrateBackfillsWorkspacesOntoExistingRecord() {
```
cwperks, line 336: If I'm reading this correctly, the migrate API can now be called multiple times? One from a version before this PR and one afterwards to get the workspace info?

spi/src/main/java/org/opensearch/security/spi/resources/ResourceProvider.java
```diff
@@ -79,4 +79,25 @@ default String ownerBackendRolesPath() {
...
     }
 
+    /**
+     * Returns the name of the field on documents of this type that holds the set of workspace IDs the
+     * resource belongs to. A single resource may belong to <em>multiple</em> workspaces, so — unlike
+     * {@link #parentIdField()}, which resolves a single parent — this field is expected to be
+     * multi-valued (for example a {@code keyword} array) and every value is captured.
+     *
+     * <p>When declared, the security plugin reads these workspace IDs at index time and projects them
+     * into the resource's denormalized {@code all_shared_principals} field as {@code workspace:<id>}
```
cwperks: Is this comment still accurate?

spi/src/main/java/org/opensearch/security/spi/resources/ResourceProvider.java
```diff
@@ -79,4 +79,25 @@ default String ownerBackendRolesPath() {
...
+     * resource through the existing DLS intersection.
+     *
+     * <p>Returning {@code null} (the default) means the resource type is not workspace-associated and
+     * behavior is unchanged. This keeps the change additive for all existing providers.
+     *
+     * @return the field name containing the resource's workspace IDs, or {@code null} if this provider
+     *         does not participate in workspace-based sharing
+     */
+    default String workspacesField() {
+        return null;
```
cwperks: I think `workspaces` would be a reasonable default here.

spi/src/main/java/org/opensearch/security/spi/resources/ResourceSharingExtension.java
```diff
@@ -37,4 +38,33 @@ public interface ResourceSharingExtension extends SecurityConfigExtension {
...
+     *
+     * <p>The default returns an empty set, which disables workspace-based DLS visibility for the plugin. That is
+     * intentional and safe: only plugins that own an authoritative workspace-membership source should override.
+     *
+     * @param username     the authenticated user's name; never {@code null}
+     * @param securityRoles the user's security roles; never {@code null}, may be empty
+     * @param backendRoles the user's backend roles; never {@code null}, may be empty
+     * @return the trusted workspace IDs the user belongs to, or an empty set if none / not implemented
+     */
+    default Set<String> resolveWorkspacesForUser(String username, Set<String> securityRoles, Set<String> backendRoles) {
```
cwperks, line 62: @DarshitChanpura With multi-tenancy, with a request destined for a tenant there is a `securitytenant` header. What is the equivalent with workspaces? If a request is for a particular workspace then there would only be a single workspace to filter by for security.

src/main/java/org/opensearch/security/resources/sharing/ResourceSharing.java
```diff
@@ -434,6 +496,12 @@ public List<String> getAllPrincipals() {
             principals.add("user:" + createdBy.getUsername());
         }
 
+        // NOTE: workspace membership is intentionally NOT projected into all_shared_principals. DLS visibility via
```
cwperks: nit: let's keep comments pertinent to the code as-is without referencing a previous revision in this particular PR.

src/main/java/org/opensearch/security/resources/ResourceAccessHandler.java
```diff
@@ -132,14 +135,49 @@ public void hasPermission(
...
+    ) {
+        // Entry point: start with an empty visited-set so container inheritance (parent + workspaces) cannot loop.
+        hasPermission(resourceId, resourceType, action, new HashSet<>(), listener);
+    }
+
+    /**
+     * Internal permission check that carries a {@code visited} set of {@code type:id} keys to prevent unbounded
+     * recursion when resources inherit access from containers (a hierarchical parent and/or workspaces). Container
+     * inheritance walks a graph that is expected to be acyclic, but a malformed graph (e.g. a workspace that
+     * transitively contains itself) would otherwise loop forever. The set tracks the current ancestor (parent)
```
cwperks: > e.g. a workspace that transitively contains itself.

This cannot happen. A lot of the logic in here looks overly complicated.

DarshitChanpura:
Thanks @cwperks — addressed in e2f4225bb:

- **ResourceAccessHandler ("overly complicated", cycle can't happen):** agreed. Dropped the ancestor cycle guard entirely — workspaces are leaf-evaluated (never recursed into), and parent inheritance is the single chain the pre-existing code already recursed without a guard, so the visited-set + path-scoping was guarding a case that can't occur. `hasPermission` is back to one method; the container fallback is just "any workspace grants (batched mget) OR the single parent grants (recursion)".
  - **Re your DM about `ResourceAccessEvaluator`:** that file is unchanged in this PR (no diff) — the write-path changes are all in `ResourceAccessHandler`, and should read much simpler now. Happy to walk through it.
- **`ResourceProvider.workspacesField()` default:** now defaults to `"workspaces"`; a doc without the field just belongs to no workspace. Removed the redundant override in the sample plugin.
- **`ResourceProvider` javadoc (still accurate?):** no — updated. It described the old `all_shared_principals` denormalization, which is gone; it now describes storing on the sharing record (write path) + DLS filtering the field (read path).
- **Comment nits (MigrateApiTests, ResourceSharing):** trimmed; removed references to prior revisions.
- **Migrate called multiple times (pre/post this PR)?** Yes — that's intended and safe. `backfillWorkspacesOnExisting` is idempotent: re-running merges workspace IDs onto the existing record and only writes when something's new, so an old (pre-PR) migrate followed by a new one picks up workspace info without duplicating or clobbering `created_by`/`share_with`.
- **Workspace request-scoping header (like `securitytenant`):** good question — this is the one open design item. There's no workspace equivalent wired yet; membership comes from the trusted `resolveWorkspacesForUser` SPI (resolved from server-set identity, not a request header). If workspaces should be *request-scoped* (a "current workspace" header, analogous to tenant), that's a distinct mechanism we'd define with the OpenSearch-side workspace backend (§11 in the design doc). I'll capture it there as an open question — want it request-scoped, membership-scoped, or both?

DarshitChanpura:
On request-scoping (your `securitytenant` question) — recommendation: **membership-scoped as the enforced authorization, request-scoping only as an optional narrowing filter later (not instead).**

- Membership = what you're *allowed* to see (authz); a "current workspace" header = what you *want* to see now (UX). A request header must only ever narrow, never widen — otherwise omitting/forging it could over-share.
- The `securitytenant` analogy breaks: tenants are disjoint + single-active, whereas workspaces are multi-membership and non-disjoint (a resource can be in several). Your "realm" framing is itself membership-based.
- Backend enforcement must hold for direct `.kibana`/API calls that carry no OSD workspace context, so membership has to be the floor regardless.

**No code change needed** — the current DLS already filters the `workspaces` field against `resolveWorkspacesForUser` (membership), with no request header. If the UI later needs a single-workspace view, we add a request filter that intersects with membership — tracked in design doc §11.5(4) as a fast-follow with the OpenSearch-side backend.

spi/src/main/java/org/opensearch/security/spi/resources/ResourceSharingExtension.java
DarshitChanpura: Deliberately no `securitytenant`-style header for workspaces — access is **membership-scoped**, not request-scoped.

The models differ:
- `securitytenant` works as a request selector because a saved object lives in exactly **one** tenant, and the tenant *is* the security boundary — the header picks which tenant's index you operate in.
- A workspace here is a **container a resource belongs to**, and a resource can be in **multiple** workspaces (1:N). The security boundary is the user's **membership across all their accessible workspaces**, resolved server-side via the trusted `resolveWorkspacesForUser` — not something the request asserts.

So on the read path, DLS filters the resource's `workspaces` field against the user's **whole** accessible-workspace set (a `terms` over the set), not a single header-selected workspace.

A "current workspace" (the workspace the UI is focused on) is the closest analogue to `securitytenant`, but it's a **UX/narrowing** concern, not a security one: such a header could only *narrow* results to that workspace (intersected with membership) — it can never *grant*, since membership is the enforced floor. DLS already works with no header, so it's an optional fast-follow if the UI wants server-side scoping to the active workspace.

Worth flagging: because membership is 1:N (unlike 1:1 tenant), "a single workspace to filter by" isn't quite the model even when a request targets one workspace — security still authorizes against the user's full membership; a single-workspace header would be an extra filter on top.

cwperks: But for requests like GET Report Definition, how does that request ferry along which workspace the user is toggled to in dashboards to the backend?

src/main/java/org/opensearch/security/resources/ResourceSharingDlsUtils.java
```diff
@@ -48,24 +50,48 @@ public static IndexToRuleMap<DlsRestriction> resourceRestrictions(
...
         }
 
-        XContentBuilder builder = null;
-        DlsRestriction restriction;
+        // Workspace visibility is expressed as a separate clause on the resource's own `workspaces` field (which OSD
+        // maintains), rather than by denormalizing workspace:<id> into all_shared_principals. Membership comes from
+        // ResourceSharingExtension.resolveWorkspacesForUser, whose SPI contract requires a trusted, server-set,
+        // I/O-free source (see that interface's javadoc). Filtering the live field means associate/dissociate are
+        // reflected automatically. If no extension implements the resolver, the set is empty and the clause is omitted.
+        Set<String> userWorkspaces = resourcePluginInfo == null ? Set.of() : resourcePluginInfo.resolveWorkspacesForUser(user);
```
cwperks, line 58: I think we need to figure this out. With multi-tenancy, we set `user.getRequestedTenant` to know which tenant a request is destined for.

cwperks: I think security-dashboards-plugin carries the tenant info via https://github.com/opensearch-project/security-dashboards-plugin/blob/d4db9d33dde68e8c62565bfbd7bea66e980399eb/server/auth/types/authentication_type.ts#L175-L193

DarshitChanpura:
## Follow-up on the `securitytenant` question — prototyped it

Short version: workspaces don't need a `securitytenant` equivalent for **security**, but a request-scoped header is a clean **UX narrowing**, and it works. Prototyped end-to-end to confirm. **Not proposing it for this PR** — separate follow-up (details at the bottom).

### Why it isn't load-bearing the way tenant is

Tenant is a storage **partition** — the header selects `.kibana_<tenant>`, so it's required to both route and isolate. Workspace membership is **1:N over one shared index**, so security already filters by the user's whole membership set. There's no index to pick, so no header is needed to secure anything.

What a header *adds* is narrowing to the workspace the UI is actually in. It can only intersect with membership — never grant.

### Prototype (3 layers)

- **OSD (`security-dashboards-plugin`)** — in the auth handler, forward `currentworkspace` from `getWorkspaceState(request).requestWorkspaceId`. Worth noting: the workspace id is **not** in a cookie (unlike tenant). Core parses it from the `/w/<id>` URL at `onPreRouting` and strips it before auth, so it has to be read from request state, not the URL.
- **Backend (`security`)** — `BackendRegistry` reads the header right next to the existing tenant block and re-emits it server-side on the thread context.
- **DLS** — `ResourceSharingDlsUtils` narrows the workspace set: `contains(requested) ? {requested} : {}`.

### Live integration test (the actual demo)

```java
// FULL_ACCESS_USER is a member of exactly one workspace; the resource is admin-owned
// and visible to them only via workspace membership (not via sharing).

// no header -> full membership -> visible
assertSearchResponse(ok(() -> api.searchResources(FULL_ACCESS_USER)), 1, "sample");

// header = the user's own workspace -> narrowed set still contains it -> visible
assertSearchResponse(ok(() -> searchWithWorkspace(FULL_ACCESS_USER, userWorkspace)), 1, "sample");

// header = a workspace the user is NOT a member of -> narrowed to empty -> hidden
assertSearchResponse(ok(() -> searchWithWorkspace(FULL_ACCESS_USER, "ws-not-a-member")), 0, null);
```

The third assertion is the proof: the header reached DLS, and a workspace the user isn't a member of narrows to empty rather than granting. Passes against a live cluster.

### One real bug it surfaced

`putHeader` does **not** survive the plugin's stashed system-subject context (the sample search runs via `PluginClient`). The narrowing case failed until I switched to `putPersistent` — the same channel that carries the authenticated `User` to the DLS valve.

### Not done / open

- The OSD half hasn't been TypeScript-compiled yet (the deep `src/core/server/utils` import is unvalidated).
- Requires `currentworkspace` in `opensearch.requestHeadersWhitelist`.
- **Inert until the OSD-side work lands** — `.kibana` isn't registered as a resource-sharing protected type, so in a real Dashboards run there's nothing for the header to filter.
- Design question for you: carry it on `User` like `requestedTenant` (survives to all nodes, cleaner downstream), or the persistent context entry I used?

Keeping this out of this PR since it's inert without the OSD side and would need its own review — will raise it separately.

src/main/java/org/opensearch/security/resources/ResourceIndexListener.java
```diff
@@ -175,15 +198,18 @@ public void postIndex(ShardId shardId, Engine.Index index, Engine.IndexResult re
...
                 .resourceType(resourceType)
                 .tenant(parentSharing.getTenant())
                 .createdBy(parentSharing.getCreatedBy())
                 .parentType(parentType)
-                .parentId(parentId)
-                .build();
-            this.resourceSharingIndexHandler.indexResourceSharing(resourceIndex, sharingInfo, listener);
+                .parentId(parentId);
+            // Stamp the child's own workspaces onto its record; ownership is still inherited from the parent above.
+            if (provider.workspacesField() != null) {
```
cwperks, line 209: I believe this is never null now so this will always be true. This is ok for now, but enabling workspaces is optional so we would eventually need to know if a cluster is workspace enabled or not.

src/main/java/org/opensearch/security/resources/ResourcePluginInfo.java
```diff
@@ -148,6 +168,29 @@ public static String extractFieldFromIndexOp(String fieldName, Engine.Index inde
...
+     * e.g. the set of workspace IDs a resource belongs to (see {@link ResourceProvider#workspacesField()}). Unlike
+     * {@link #extractFieldFromIndexOp} it collects every {@link IndexableField} for {@code fieldName}, which is how a
+     * {@code keyword} array surfaces (one field per element). Values only surface for indexed/stored mappings, not
+     * {@code doc_values}-only fields.
+     *
+     * @param fieldName the multi-valued field to extract; must not be {@code null}
+     * @param indexOp   the index op whose parsed document is inspected; must not be {@code null}
+     * @return the field's values, or empty if none
+     */
+    public static Set<String> extractMultiValuedFieldFromIndexOp(String fieldName, Engine.Index indexOp) {
```
cwperks, line 182: For any request that is creating in a workspace or retrieving from a workspace, I think we can assume the request itself will carry that info. With multi-tenancy we can always get the requested tenant with `User.getRequestedTenant()` so should we consider doing similar here?

cwperks (approved):

DarshitChanpura:
Thanks for the thorough reviews on this one @cwperks — all the items from the last few rounds are addressed and CI is green on `649432f6f` (66 passing, 0 failures). The only delta since your last pass is javadoc/inline-comment tightening, no logic change.

Anything further I'm happy to take as follow-up PRs rather than growing this one — specifically:

- the request-scoped current-workspace header (prototype in the comment above), which is inert until the OSD side lands anyway
- the OSD-side backend work: registering the saved-objects index as a resource-sharing protected type + the `workspace` container type
- the `User`-vs-thread-context design choice for carrying the current workspace, if you have a preference

If you're happy with where this stands, could you drop a formal approval when you get a minute? The PR currently shows `REVIEW_REQUIRED` and is blocked on that.

Issue #6339, opened 2026-07-28
Add pagination support to Security configuration collection APIs


cwperks:
### Description

The Security configuration collection APIs currently return every caller-visible configuration entity in a single response. Large deployments may have thousands of internal users, roles, role mappings, action groups, tenants, or node-DN groups, which can produce large payloads and add memory, serialization, and client-side processing overhead.

OpenSearch introduced a common cursor-based pagination model for collection APIs in [OpenSearch#14641](https://github.com/opensearch-project/OpenSearch/pull/14641) and [OpenSearch#14718](https://github.com/opensearch-project/OpenSearch/pull/14718). The Security plugin should adopt a consistent version of that model for its configuration collection APIs.

### Proposed scope

Add opt-in pagination to collection GETs for:

- `/_plugins/_security/api/internalusers` (and the legacy `/user` alias)
- `/_plugins/_security/api/roles`
- `/_plugins/_security/api/rolesmapping`
- `/_plugins/_security/api/actiongroups` (and the legacy `/actiongroup` alias)
- `/_plugins/_security/api/tenants`
- `/_plugins/_security/api/nodesdn`

Singleton configuration APIs such as `securityconfig`, `audit`, and `allowlist` do not need pagination. Other non-configuration collection APIs (for example API tokens, certificates, versions, and resource APIs) can adopt the contract separately if appropriate.

### Proposed API contract

Use the same concepts as the OpenSearch `_list` APIs:

- `size`: positive page size, defaulting to 100 when pagination is requested
- `next_token`: opaque cursor returned by the preceding page
- `sort`: `asc` or `desc`, sorting by configuration entity name and defaulting to `asc`

For example:

```http
GET /_plugins/_security/api/roles?size=2&sort=asc
```

```json
{
  "next_token": "<opaque token>",
  "roles": {
    "role_a": { "...": "..." },
    "role_b": { "...": "..." }
  }
}
```

The final page should return `"next_token": null`.

The cursor should be bound to at least the configuration type, sort direction, and last returned entity name. Invalid or tampered cursors, and cursors used with a different endpoint or sort direction, should return `400 Bad Request`. Name-based cursor continuation is preferable to offsets so additions or deletions before the cursor do not shift every later page.

### Backward compatibility

Requests without pagination parameters must retain the exact existing response shape and continue returning the complete visible collection. Single-entity GET endpoints must retain their current behavior; pagination parameters on those endpoints should be rejected with `400 Bad Request`.

[...]

PR #6378, opened 2026-08-08
Add pagination support for Collection APIs

itsmevichu:
### Description
* Category:  New feature
Large deployments can have thousands of security configuration entities (users, roles, mappings, etc.). Returning everything in one response creates unbounded payloads. This PR adds opt-in cursor-based pagination to the six Security configuration collection APIs, using the same surface contract (size, sort, next_token) as OpenSearch core's _list APIs.
Without pagination parameters, all six collection endpoints behave identically to before — fully backward compatible.
With the new opt-in parameters, responses use a paginated envelope:
```{ "next_token": "<cursor or null>", "roles": { "role_a": {}, "role_b": {} } }```
Affected endpoints: internalusers, roles, rolesmapping, actiongroups, tenants, nodesdn.
* Key guarantees:
Pagination applies after authorization and redaction — hidden entities cannot leak through page contents or cursor values.
Cursors are bound to endpoint and sort direction; misuse returns 400.
Pagination params on single-entity GETs return 400.
Lexicographic cursor continuation - safe across additions and deletions between page requests.

### Issues Resolved
https://github.com/opensearch-project/security/issues/6339

### Testing
Unit tests, Integration tests and manual testing.

### Check List


itsmevichu:
@cwperks @DarshitChanpura Could you please review this PR when you get a chance? It’s been inactive for a while, and I’d really appreciate your feedback. Thanks!

cwperks:
> @cwperks @DarshitChanpura Could you please review this PR when you get a chance? It’s been inactive for a while, and I’d really appreciate your feedback. Thanks!

I will take a look this week. Sorry for delay. Ty for taking this on, its certainly something that I think can be done more broadly for any category of configuration.

cwperks:
@itsmevichu one high-level comment:

How would pagination work for APIs that may support a filter?

I think that `internalusers` may have a param called `filterBy` to distinguish internal from service account users (note: service was created in context of extensions and not currently used so in effect the `internalusers` API has no filter mechanism), but what happens if we pass a filter in like `/_plugins/_security/api/logs*` that matches 10_ users. Do the changes in this PR work?

Same on `nodes_dn` where I see a param called `show_all`. I'm not familiar with this param, but let's make sure the logic honors the existing params.

PR #6390, opened 2026-08-11
fix(audit): dedup bulk audit events by only iterating BulkShardRequest

Taiwo435:
### Description

**Category:** Bug fix

A bulk request runs through `ActionFilters` twice, both times on the
coordinating node: first as a `BulkRequest`, then as a `BulkShardRequest` when
`TransportBulkAction` fans out to `shardBulkAction.execute()` (before
`TransportReplicationAction` reroutes to the primary shard). The reroute to the
(possibly remote) data node goes through the transport request handler, which
does not re-run `ActionFilters`, so the data node never re-audits.
`AuditActionFilter` previously iterated sub-items for **both** passes, emitting
duplicate per-item `REQUEST_AUDIT` events for every bulk sub-operation.

**Why these changes are required:**
Duplicate audit events inflate the audit log, distort audit-based metrics and
alerting, and make it appear that each bulk sub-operation happened twice. Audit
records must reflect each operation exactly once.

**Old behavior (before):**
The bulk-handling block matched both `BulkShardRequest` and `BulkRequest`. Both
passes iterated the sub-items, so every bulk sub-operation produced two per-item
`REQUEST_AUDIT` events — one from the `BulkRequest` pass and one from the
`BulkShardRequest` pass, both on the coordinating node.

**New behavior (after):**
Only `BulkShardRequest` (shard-level) iterates sub-items, so each sub-operation
is audited exactly once and carries its shard ID. A `BulkRequest` targeting
normal indices falls through to the standard path, producing one summary
`REQUEST_AUDIT` event listing the distinct target indices (the audit index is
filtered out of that list, so it never leaks into the summary event).

Because a single-doc write to the audit index arrives as a `BulkRequest` (via
`TransportSingleItemBulkWriteAction`) — and `BulkRequest` does not implement
`IndicesRequest` — the top-level self-loop guard was extended to inspect a
`BulkRequest`'s sub-item indices. It suppresses the request (zero events) only
when **every** sub-item targets the audit index (all-match semantics), so a
pure audit-index bulk write is dropped before reaching either the bulk or
summary path. A mixed `BulkRequest` (normal + audit-index items) is not
suppressed; it falls through to the summary path, where the audit index is
excluded from the summary's indices. The existing `BulkShardRequest` top-level
and per-item audit-index self-loop guards are preserved.

### Testing

Unit testing (`AuditActionFilterTest`) — added bulk-path tests, full suite green
(32 tests, 0 failures). Also ran `spotlessApply`, `compileJava`,
`compileTestJava`, and `integrationTest` — all clean. Tests cover:

- Per-item logging — `BulkShardRequest` emits one event per sub-item.
- Shard ID presence — per-item events carry the shard ID.
- `resolve_bulk_requests` disabled — `BulkShardRequest` falls through to the standard single-event path.
- Audit-index self-loop skipping:
  - `BulkShardRequest` top-level guard (shard belongs to the audit index → zero events).
  - `BulkShardRequest` per-item guard (mixed items skip only the audit-index item).
  - `BulkRequest` guard (all-match — a bulk write where every sub-item targets the audit index is suppressed → zero events).
- Empty items — no events, chain still proceeds.
- `BulkRequest` summary — normal-index `BulkRequest` produces one summary event listing distinct target indices, with no per-item iteration.
- `BulkRequest` mixed summary — a mixed `BulkRequest` (normal + audit-index items) produces one summary event whose indices exclude the audit index.
- Integration: `StandaloneAuditDisabledModeTest.shouldNotProduceEventsForAuditIndexWrites` passes (the regression this fix addresses).

### Check List

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following Developer
Certificate of Origin and signing off your commits, please check
[here](https://github.com/opensearch-project/security/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/src/main/java/org/opensearch/security/filter/AuditActionFilter.java
+++ b/src/main/java/org/opensearch/security/filter/AuditActionFilter.java
@@ -129,26 +129,31 @@ public class AuditActionFilter implements ActionFilter {
         }
 
         // Skip requests targeting the audit index (prevent self-referential loop).
-        // BulkRequest does not implement IndicesRequest, so its target indices are
-        // gathered from its sub-items; a single-doc write to the audit index arrives
-        // here as a BulkRequest (via TransportSingleItemBulkWriteAction) and must be
-        // guarded too, otherwise the standard summary path below would leak an event.
-        String[] targetIndices = null;
+        // IndicesRequest guard: any-match (e.g., BulkShardRequest whose shard belongs
+        // to the audit index is suppressed entirely — matches FGAC's behavior where
+        // internal/system writes are never audited).
+        // BulkRequest guard: all-match — only suppress when EVERY sub-item targets the
+        // audit index. A single-doc write to the audit index arrives here as a
+        // BulkRequest (via TransportSingleItemBulkWriteAction) with one item, so
+        // all-match catches it. Mixed BulkRequests (normal + audit-index items) still
+        // produce a summary event; the per-item guard in logBulkItem() and the
+        // IndicesRequest guard on the BulkShardRequest handle individual audit-index
+        // items at the shard level.
         if (request instanceof IndicesRequest) {
-            targetIndices = ((IndicesRequest) request).indices();
-        } else if (request instanceof BulkRequest) {
-            targetIndices = ((BulkRequest) request).requests()
-                .stream()
-                .map(DocWriteRequest::index)
-                .filter(idx -> idx != null)
-                .toArray(String[]::new);
-        }
-        if (targetIndices != null) {
-            for (String idx : targetIndices) {
-                if (isAuditIndex(idx)) {
-                    chain.proceed(task, action, request, listener);
-                    return;
+            String[] indices = ((IndicesRequest) request).indices();
+            if (indices != null) {
+                for (String idx : indices) {
+                    if (isAuditIndex(idx)) {
+                        chain.proceed(task, action, request, listener);
+                        return;
+                    }
                 }
+            }
+        } else if (request instanceof BulkRequest) {
+            List<? extends DocWriteRequest<?>> bulkItems = ((BulkRequest) request).requests();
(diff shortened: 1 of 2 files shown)


DarshitChanpura:
Two things:

**1. PR description / inline comment — mechanism is inaccurate.** The description and the comment at `AuditActionFilter.java:185-187` say the two passes happen "first as a `BulkRequest` at the coordinating node, then as a `BulkShardRequest` at the data node." Both passes actually run on the **coordinating node**: the `BulkShardRequest` pass fires when `TransportBulkAction` fans out to `shardBulkAction.execute()` (`TransportBulkAction.java:762`), *before* `TransportReplicationAction` reroutes to the primary. The reroute to the (possibly remote) data node goes through the transport request handler (`handleOperationRequest`), which does not re-run `ActionFilters`, so the data node never re-audits. The `2N -> N+1` dedup outcome is unchanged — just worth correcting the rationale in the description and comment.

**2. Audit index can leak into the summary event.** A mixed `BulkRequest` (normal + audit-index sub-items) gets past the all-match self-loop guard at `AuditActionFilter.java:152-158` and falls through to the summary path. There, `distinctIndices` (`AuditActionFilter.java:244-253`) filters only `null`, so the audit index name ends up in the summary event's `INDICES` field:

```java
String[] distinctIndices = bulkRequest.requests()
    .stream()
    .map(r -> r.index())
    .filter(idx -> idx != null)   // no isAuditIndex() exclusion
    .distinct()
    .toArray(String[]::new);
```

Adding `&& !isAuditIndex(idx)` to the filter would keep the audit index out of the summary event. (This block is pre-existing; the new all-match guard + fall-through path just widens when it is reached.)

DarshitChanpura (approved):

PR #6392, opened 2026-08-13
Expose dynamic audit config via cluster settings in SSL-only mode

Taiwo435:
### Description

Narrow the audit settings filter in SSL-only (standalone audit) mode so the non-secret dynamic audit configuration is readable via `GET _cluster/settings`, while continuing to hide the credential-bearing sink settings.

  In SSL-only mode there is no `.opendistro_security` index, so the audit configuration is stored in cluster settings and is managed through `PUT`/`GET _cluster/settings`. The standalone audit configuration panel in the security dashboards plugin reads its state back through `GET _cluster/settings`. `OpenSearchSecurityPlugin.getSettingsFilter()` stripped the entire `plugins.security.audit.*` subtree from settings responses, which also hid the ~17 non-secret dynamic config keys the panel needs. Writes worked (the filter only affects reads), but reads came back empty — the panel could not display the current configuration.

  * Old: `getSettingsFilter()` always added `plugins.security.audit.*`, so the whole subtree — including the non-secret dynamic config — was stripped from every settings response in every mode.
  * New: in SSL-only mode the filter strips only the credential-bearing group settings (`plugins.security.audit.endpoints.*` and `plugins.security.audit.routes.*`); the dynamic config (`plugins.security.audit.config.*` and `plugins.security.audit.compliance.*`) is now returned. FGAC and all other modes keep the original broad `plugins.security.audit.*` filter, because their real audit config lives in the security index, not cluster settings.

  Credential safety is preserved:
  * The default-endpoint secrets (`username`, `password`, `webhook.url`, `pem*`, `salt`) are registered with `Property.Filtered`, so OpenSearch core strips them from settings responses regardless of this filter.
  * The nested credentials on secondary endpoints/routes are not `Property.Filtered`, so they are kept out of responses by the retained `endpoints.*` / `routes.*` filters.

  Additionally, three static external-sink settings under `config.*` — `http_endpoints` (the backend audit host list), `enabled_ssl_ciphers`, and `enabled_ssl_protocols` — were previously hidden only by the broad `plugins.security.audit.*` wildcard and carry no `Property.Filtered` annotation. Narrowing the wildcard would have exposed this infrastructure topology to unauthenticated clients in SSL-only mode via `GET _nodes/settings`. These are static (non-dynamic) sink settings the audit panel does not manage, so they are now registered `Property.Filtered` alongside their credential siblings — closing the disclosure in all modes with no functional impact (the sink still reads them directly from settings).

 * Note for backport: Adding Property.Filtered to http_endpoints, enabled_ssl_ciphers, and enabled_ssl_protocols means these settings will also disappear from _nodes/settings output in FGAC/default mode (not just SSL-only). This is intentional — they are static external-sink infrastructure config that should not have been exposed to settings readers in any mode.

### Issues Resolved

Read-path gap for the standalone audit configuration panel in SSL-only mode.

### Testing

Added two integration tests under `src/integrationTest`:

* `StandaloneAuditSettingsFilterTest` (SSL-only mode):
  * `dynamicAuditConfigIsReadableViaClusterSettings` — writes a representative slice of the dynamic settings across both prefixes (`config.*` and `compliance.*`) and both value shapes (list + boolean) via `PUT _cluster/settings`, then asserts they come back through `GET _cluster/settings`. The filter is prefix-based, so covering both subtrees and both value shapes proves all ~17 dynamic keys flow through.
  * `auditSecretsAreNotExposedInNodeSettings` — statically configures the `Property.Filtered` default-endpoint secrets plus a secondary endpoint credential and a route credential, then asserts via `GET _nodes/settings` that the non-secret config surfaces while none of the secret values or the `endpoints.*` / `routes.*` keys appear. Also configures `http_endpoints`, `enabled_ssl_ciphers`, and `enabled_ssl_protocols` under `config.*` and asserts they stay hidden (regression guard for the `Property.Filtered` additions).

* `StandaloneAuditFgacFilterUnchangedTest` (FGAC mode):
  * `fgacModeStillFiltersAuditSettings` — statically configures a dynamic audit setting and a nested `endpoints.*` credential, then asserts the broad `plugins.security.audit.*` filter still strips both from `GET _nodes/settings`, confirming the narrowing is gated to SSL-only mode and does not change FGAC behavior. (The values are set statically rather than via `PUT` because in FGAC `SecurityFilter` blocks a runtime cluster-settings update to a sensitive key unless the caller holds a `restapi.roles_enabled` role — a write guard orthogonal to the read filter under test.)

### Check List
- [x] New functionality has been documented (code comments on the filter behavior)

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/StandaloneAuditFgacFilterUnchangedTest.java
+++ b/src/integrationTest/java/org/opensearch/security/StandaloneAuditFgacFilterUnchangedTest.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.security;
+
+import java.util.List;
+import java.util.Map;
+
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.test.framework.TestSecurityConfig;
+import org.opensearch.test.framework.cluster.ClusterManager;
+import org.opensearch.test.framework.cluster.LocalCluster;
+import org.opensearch.test.framework.cluster.TestRestClient;
+import org.opensearch.test.framework.cluster.TestRestClient.HttpResponse;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.not;
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+
+/**
+ * FGAC regression guard for the settings-filter narrowing.
+ *
+ * The narrowing in {@code OpenSearchSecurityPlugin.getSettingsFilter()} is gated to SSL-only mode; an FGAC
+ * node must keep filtering the entire {@code plugins.security.audit.*} subtree. In FGAC the real audit config
+ * lives in the {@code .opendistro_security} index (managed via the security REST API), not cluster settings,
+ * so nothing under that subtree should surface from a settings response.
+ *
+ * The audit config value is set statically (node settings) rather than via {@code PUT _cluster/settings}: the
+ * dynamic audit settings are registered {@code Property.Sensitive}, and in FGAC {@code SecurityFilter} blocks a
+ * runtime cluster-settings update to a sensitive key unless the caller holds a {@code restapi.roles_enabled}
+ * role. That write guard is orthogonal to the read filter under test here; setting the value statically
+ * exercises the read filter directly (it strips the subtree regardless of how the value was configured).
+ * Statically-set values surface via {@code GET _nodes/settings}, so that is where we assert the filter applies.
+ */
+public class StandaloneAuditFgacFilterUnchangedTest {
+
+    private static final String AUDIT_CONFIG_PREFIX = ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX; // plugins.security.audit.config.
+    // Values that must NOT surface in FGAC because the broad audit.* filter still applies.
+    private static final String FGAC_HIDDEN_TOKEN = "fgac_should_be_hidden";
+    // A nested endpoint credential (not Property.Filtered): only the endpoints.* pattern hides it. Setting it in
+    // FGAC guards against a future refactor accidentally applying the SSL-only narrow filter to all modes.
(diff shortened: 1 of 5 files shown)


DarshitChanpura (commented):
Nice PR — the write-up made this easy to follow, and the diagnosis is right. I checked the pieces I was worried about: `isSslOnlyMode()` is the instance field so that's fine, and `Property.Sensitive` only gates writes, it doesn't filter reads, so narrowing the filter is genuinely the right fix here. Pulling the three `Filtered` flags onto `http_endpoints`/ciphers/protocols in the same PR is correct too — without them the narrowing would leak those, so they belong here.

I'm good with the design tradeoff. In ssl-only mode the audit config only lives in cluster settings and there's no authz layer anyway, so exposing the non-secret dynamic config to settings readers is consistent with what that mode already is. Not going to bikeshed keeping `compliance.*` filtered.

One thing I do want before merging: the safety story is now "every secret under `config.*` must be individually `Property.Filtered`," since the wildcard no longer covers them in ssl-only. That's fine today, but it's a trap for whoever adds the next sink credential. Can you add a short comment at the `config.*` registration block spelling that out, plus a test that fails if a config.* key with a secret-ish name isn't Filtered? Something like:

```java
@Test
public void allSensitiveConfigSettingsAreFiltered() {
    Settings disabled = Settings.builder().put(ConfigConstants.SECURITY_DISABLED, true).build();
    Pattern secretish = Pattern.compile("password|username|token|webhook|pem|salt|jks|http_endpoints|ssl_");
    for (Setting<?> s : new OpenSearchSecurityPlugin(disabled, null).getSettings()) {
        String key = s.getKey();
        if (key.startsWith(ConfigConstants.SECURITY_AUDIT_CONFIG_DEFAULT_PREFIX) && secretish.matcher(key).find()) {
            assertThat(key + " must be Property.Filtered", s.getProperties().contains(Setting.Property.Filtered), equalTo(true));
        }
    }
}
```

(Building the plugin with `disabled=true` is the easy way to reach `getSettings()` without dragging in configPath/TLS setup — a plain `new OpenSearchSecurityPlugin(Settings.EMPTY, null)` throws. If it fights the unit-test module for any reason, just fold the same secret-not-present assertions into `StandaloneAuditSettingsFilterTest`.)

Couple of small things, non-blocking: the tests hardcode the `plugins.security.audit.compliance.` prefix — worth exposing the constant from `SecuritySettings` so it can't drift. And add the `skip-changelog` label. Also worth a line in the description that `http_endpoints`/ciphers/protocols will disappear from settings output in FGAC/default too, so we don't get surprised on backport.

DarshitChanpura (approved):
Thanks for the follow-ups — this addresses everything. The invariant comment at the sink block is exactly the context the next person needs, and I like that the tripwire regex is broader than my sketch with the "add to an exclusion list, don't weaken the pattern" note. Making `COMPLIANCE_PREFIX` public is a nice cleanup too.

CI is green now that #6407 (the HttpClient5 5.6.4 hostname-verification fix) is in — the earlier citest/sslTest failures were upstream core, not this PR.

LGTM.

DarshitChanpura (approved):

PR #6393, opened 2026-08-13
Bound concurrent password-hash verifications and cache repeated failed credential checks

pgtgrly:
### Description

Adds bounded concurrency and caching around password-hash verification in the internal authentication backend.

* **Category:** Enhancement
* **Why these changes are required?**

  The internal authentication backend performs a full password-hash verification (BCrypt) for every credential check, on the request-handling thread. BCrypt is intentionally expensive, so under high authentication volume these verifications can consume a large share of CPU and add latency to concurrent requests. There is currently no bound on how many run at once, and repeated identical failing credentials re-run the full verification every time.

* **What is the old behavior before changes and new behavior after changes?**

  *Old behavior:* every credential check runs a full hash verification with unbounded concurrency; repeated identical failures each pay the full cost.

  *New behavior:* three additions, all opt-out-able via settings and with conservative defaults:

  1. **Bounded concurrent hash verifications** — a semaphore limits how many verifications run at once (default `max(1, availableProcessors / 4)`). When the limit is reached, the request receives `503 SERVICE_UNAVAILABLE`, signalled internally by a new `AuthBackendThrottledException`. These responses are deliberately **not** reported to `auth_failure_listeners`, so transient load does not affect the IP/username rate limiters. Only the internal backend is limited; LDAP, SAML, Kerberos, JWT and PKI paths are untouched.

  2. **Short-lived cache of failed credential checks** — repeated identical failing credentials are answered from cache without re-running the verification. Only *definitive* credential failures populate it, signalled by a new typed `InvalidCredentialsException`. Transient errors (e.g. "backend not configured" during startup or a config reload) continue to throw the generic `OpenSearchSecurityException` and are never cached. The cache is cleared alongside the other auth caches on config reload.

  3. **Optional `AuthenticationBackend#userExists()` fast-path** — a default method (returns `Optional.empty()`, so existing custom backends are unaffected) letting a backend report whether a user is known, so requests for unknown users can be short-circuited before the hash verification runs.

  New optional settings, all with safe defaults:

  | Setting | Default |
  |---|---|
  | `plugins.security.auth.max_concurrent_bcrypt` | `max(1, availableProcessors / 4)` (`0` disables the limit) |
  | `plugins.security.cache.incorrect_credential_ttl_minutes` | `10` |
  | `plugins.security.cache.incorrect_credential_max_size` | `10000` |

### Issues Resolved

None — this is a standalone improvement to the authentication path and is not tied to an existing issue.

Is this a backport? No. Requesting a `backport 3.7` label so this also lands on the 3.7 branch.

Does this introduce new permission(s) to be displayed in the static dropdown on the front-end? No.

### Testing

* **Unit tests (added, all passing):**
  * `AuthBackendThrottledExceptionTest` (5 tests) — message preservation, that it is an unchecked `RuntimeException`, that the stack trace is suppressed (`fillInStackTrace` returns `this`, matching Netty's `StacklessClosedChannelException` pattern), and — importantly — that it does **not** extend `OpenSearchSecurityException`, so it is not absorbed by the generic auth-failure handling and can propagate to the 503 handler.
  * `InvalidCredentialsExceptionTest` (3 tests) — message preservation, that it remains assignable to `OpenSearchSecurityException` so existing callers are unchanged, and that it retains a stack trace for diagnostics.
  * `InternalAuthBackendTests` (+3 tests, 7 total) — `userExists()` returning `true`/`false`, and returning `Optional.empty()` when the `InternalUsersModel` is transiently `null` during startup/reload (must not throw).

* **Manual testing:** built the plugin against 3.7.0 and ran it in a single-node Docker cluster with the limit forced to 1 permit. Verified that: existing users authenticate normally in steady state; concurrent cold-cache logins receive `503` rather than `401` when the limit is reached; and those `503`s do **not** increment the IP rate limiter (a client on the same source IP continued to authenticate successfully throughout).

### Check List


src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -785,20 +875,87 @@ private User authcz(
...
                     }
-                    final User authenticatedUser = authBackend.authenticate(context);
-                    return authz(context, authenticatedUser, roleCache, authorizers);
+                });
+            } catch (UncheckedExecutionException e) {
+                // Guava wraps the Callable's RuntimeExceptions; unwrap so the
+                // throttle exception propagates to the caller for 503 handling
+                // and security exceptions follow the original null-return path.
+                if (e.getCause() instanceof AuthBackendThrottledException) {
+                    throw (AuthBackendThrottledException) e.getCause();
```
devardee, line 948: method description say's : `no auditlog, throw no exception, does also authz for all authorizers` . but we are now throwing exception's.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -785,20 +875,87 @@ private User authcz(
...
+                            try {
+                                authenticatedUser = authBackend.authenticate(context);
+                            } catch (InvalidCredentialsException e) {
+                                // Definitive credential failure (wrong user / wrong password /
+                                // empty password) thrown only by InternalAuthenticationBackend.
+                                // Transient errors (e.g. "not configured" during startup/reload)
+                                // keep throwing plain OpenSearchSecurityException and are NOT cached.
+                                incorrectCredentialCache.put(ac, Boolean.TRUE);
+                                throw e;
+                            }
```
devardee, line 934: this whole logic is scoped to Basic Auth alone, can we have this logic inside of `InternalAuthenticationBackend.authenticate()` method ?

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -785,20 +875,87 @@ private User authcz(
...
+            // with the cache-put path: only InternalAuthenticationBackend throws
+            // InvalidCredentialsException (and therefore is the only backend that can
+            // populate this cache). Reading it for other backends (LDAP, SAML, Kerberos)
+            // would incorrectly reject users who exist there but not in internal_users.
+            if (authBackend instanceof InternalAuthenticationBackend && incorrectCredentialCache.getIfPresent(ac) != null) {
+                if (log.isDebugEnabled()) {
+                    log.debug("Credentials for user {} found in incorrect-credential cache, rejecting", ac.getUsername());
+                }
+                return null;
+            }
```
devardee, line 889: Can this lead to timing attacks ?

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -527,6 +605,18 @@ Authenticate users injected in the thread context by internal components (plugin
...
             Populate thread context with anonymous user is anonymous login requested and no other credentials provided.
              */
+            if (authenticationThrottled) {
+                log.warn(
+                    "Authentication throttled due to BCrypt concurrency limit for {} from {}",
+                    authCredentials == null ? null : authCredentials.getUsername(),
+                    remoteAddress
+                );
+                request.queueForSending(
+                    new SecurityResponse(SC_SERVICE_UNAVAILABLE, "Authentication service temporarily unavailable, please retry later")
```
devardee, line 615: Instead of 503, we should return 429 to the caller (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/429)

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -527,6 +605,18 @@ Authenticate users injected in the thread context by internal components (plugin
             Handle anonymous auth.
             Populate thread context with anonymous user is anonymous login requested and no other credentials provided.
              */
```
devardee, line 607: move this comment to this if block : `if (authCredentials == null && anonymousAuthEnabled && isRequestForAnonymousLogin(request.params(), request.getHeaders())) {`

kkhatua:
@pgtgrly 
A couple of minor comments:
1. `max(1, availableProcessors / 4)` --> Why did we settle for this? Could you share some perf numbers around this?
2. If the cache's purpose is to block recomputation of already failed/invalid creds, wouldn't an attacker essentially keep trying different permutations?
3. Why are we holding the password in plaintext? Isnt the hash sufficient to confirm if the retry is a match? It is possible that a typo in the username and a valid password might be stored, and a heapdump would reveal this.

kkhatua:
One interesting thing is that the speed of response of a system could reveal to an attacker of the existence of a valid username, given that retries with the cache will respond faster. 
nit: This might feel like a bit of an overkill, but would it make sense to inject a delay similar to the time for a valid authentication?

cwperks:
> One interesting thing is that the speed of response of a system could reveal to an attacker of the existence of a valid username, given that retries with the cache will respond faster. nit: This might feel like a bit of an overkill, but would it make sense to inject a delay similar to the time for a valid authentication?

We've actually previously created an advisory on that: https://github.com/opensearch-project/security/security/advisories/GHSA-c6wg-cm5x-rqvj

We should have some test cases around this so let's see if they give assurance or need to be updated.

cwperks:
@pgtgrly Can you please fix the conflicts? Apologies for not being able to review this in the past 2 weeks.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -123,6 +128,33 @@ public void onRemoval(RemovalNotification<AuthCredentials, User> notification) {
             })
             .build();
 
+        incorrectCredentialCache = CacheBuilder.newBuilder()
```
cwperks, line 131: i think this is fine, but I'd really like to see enhancements to `auth_failure_listeners` which provides similar functionality.

Some of the potential enhancements:

- Block at the cluster-level when any node has reached the threshold for blocking.
- Provide APIs for cluster admins to view if any blocks exist and take admin action to unblock
- UX for configuring `auth_failure_listeners` to help more people discover the feature

I think this cache will have similar problem as `auth_failure_listeners` where it only has node-level visibility.

src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -780,20 +870,87 @@ private User authcz(
...
+                            log.trace(
+                                "Credentials for user {} not cached, return from {} backend directly",
+                                ac.getUsername(),
+                                authBackend.getType()
+                            );
+                        }
+                        final Semaphore semaphore = bcryptSemaphore;
+                        final boolean shouldAcquire = (semaphore != null && authBackend instanceof InternalAuthenticationBackend);
+                        if (shouldAcquire && !semaphore.tryAcquire()) {
+                            log.warn("BCrypt concurrency limit reached, rejecting auth for user {}", ac.getUsername());
```
cwperks, line 913: Why name these bcrypt specifically? Password hashing algorithm is configurable and bcrypt is only the default.

PR #6394, opened 2026-08-13
Add FIPS build mode and the *FipsTests/*FipsIT test-selection convention

beanuwave:
### Description
**Category:** Enhancement

Establishes how FIPS mode is decided and applied, with no behaviour change to the default build. Nothing in this PR turns FIPS on for users; it makes FIPS a build and test mode the rest of the stack can rely on.

### Key changes

- **Single source of truth.** `FipsMode.isEnabled()` (env `OPENSEARCH_FIPS_MODE=true`) replaces `CryptoServicesRegistrar.isInApprovedOnlyMode()` as the way the plugin asks "are we in FIPS mode". Intent is decoupled from provider state; later PRs cross-check the two at startup.
- **BCFIPS is always built.** `compileOnly` in both modes (the jars are shipped by core); the compile-time `FipsBuildParams` fork is gone. `bctls-fips` is added, and the `securityadmin` standalone bundles ship the BC FIPS jars under `deps/`, so that bundle's classpath stays FIPS-capable for a manual `java -cp 'deps/*'` launch (its own launcher has not been self-contained since upstream #6023).
- **FIPS engages through `java.security`, not code.** No provider is registered programmatically. The runtime launcher merges in `fips_java.security` when `OPENSEARCH_FIPS_MODE=true`; tests swap the file wholesale via `-Djava.security.properties==<file>`.
- **`gradle/fips.gradle`** is the single place where the mode is decided and applied to the test surface: which classes exist in each mode, and the JVM args the workers need. Keeping both together matters — selecting the FIPS classes without the matching JVM args produces a suite that looks FIPS-shaped but runs on ordinary providers.
- **Test convention.** A FIPS variant is `<Base>FipsTests` (unit) or `<Base>FipsIT` (integration): it extends its baseline, overrides what the FIPS providers change, and adds FIPS-only assertions. The two modes select disjoint sets — outside FIPS the variants are excluded from compilation and from every `Test` task; inside FIPS each baseline that has a variant is excluded so the variant *replaces* it. Baselines therefore stay free of `FipsMode.isEnabled()` branches. This follows core's convention (`gradle/fips.gradle` there).
- **`BCFipsEntropyDaemonFilter`.** Because `java_test.security` registers BouncyCastleFipsProvider in *every* test JVM including non-FIPS runs, any suite touching JCA leaves a "BC FIPS Entropy Daemon" thread behind, which RandomizedRunner reports as a leak. The framework's `BouncyCastleThreadFilter` does not cover it. Applied to the suites that trip over it here and reused by later PRs.

### Reviewer notes

- This PR lands **inert**: no variant classes exist yet, so `fips.gradle` selects nothing and the default build is unchanged. The machinery is exercised from PR 5 onward.
- `BCFipsEntropyDaemonFilter` arguably belongs in core's test framework rather than this plugin — see *Core / distribution follow-ups* at the end of this document.

### Testing

```bash
./gradlew test integrationTest                              # unchanged
OPENSEARCH_FIPS_MODE=true ./gradlew test integrationTest    # runs, no variants yet
```

### Check List


cwperks (commented):
ty for splitting these up @beanuwave ! This looks like a good place to begin with this series of PRs. Will any of these create any new CI checks to run existing tests in FIPS approved mode as a PR check?

src/main/java/org/opensearch/security/support/FipsMode.java
```diff
@@ -0,0 +1,26 @@
...
+public final class FipsMode {
+
+    static java.util.function.Supplier<String> envSupplier = () -> System.getenv("OPENSEARCH_FIPS_MODE");
+
+    public static boolean isEnabled() {
+        return "true".equalsIgnoreCase(envSupplier.get());
+    }
+
+    private FipsMode() {
+        throw new UnsupportedOperationException();
```
reta: Wondering why do we need that? The private constructor prohibits class instantiation (https://docs.oracle.com/javase/specs/jls/se7/html/jls-6.html#d5e8146)


```suggestion
```

src/test/java/org/opensearch/security/support/FipsModeTest.java
```diff
@@ -0,0 +1,55 @@
...
+        for (String disabled : new String[] { "false", null, "", "yes" }) {
+            FipsMode.envSupplier = () -> disabled;
+            assertThat("expected disabled for: " + disabled, FipsMode.isEnabled(), equalTo(false));
+        }
+    }
+
+    @Test
+    public void constructor_isNotInstantiable() throws Exception {
+        Constructor<FipsMode> constructor = FipsMode.class.getDeclaredConstructor();
+        constructor.setAccessible(true);
```
reta: The code outside the security module should not be able to widen accessibility rules (unless opened explicitly), I think we should be fine without reflection based tests ....

gradle/fips.gradle
```diff
@@ -0,0 +1,152 @@
...
+ *
+ *   non-FIPS  the variants are excluded from compilation and from every Test task,
+ *             so only the baseline classes run.
+ *   FIPS      the variants are compiled and run, and each baseline class that has a
+ *             variant is excluded, so the variant replaces it rather than adding to it.
+ *
+ * FIPS mode is driven by the OPENSEARCH_FIPS_MODE environment variable.
+ */
+
+ext.fipsTestsEnabled = 'true'.equalsIgnoreCase(System.getenv('OPENSEARCH_FIPS_MODE'))
```
reta: Could we reuse FipsBuildParams from OpenSearch build please: 

```suggestion
import org.opensearch.gradle.info.FipsBuildParams

ext.fipsTestsEnabled = FipsBuildParams.isInFipsMode()
```

reta:
Thanks @beanuwave , a few minor comments, looks otherwise, same question as  https://github.com/opensearch-project/security/pull/6394#pullrequestreview-5235108609 : do we need update test matrix on CI to run fips / non-fips tests?

beanuwave:
@reta @cwperks Thanks for the review!

GitHub Actions are **not** part of this series. Looks like it will be a combined effort of extending `.github/workflows/ci.yml` and the `opensearch-build` action / integ-test config.

The `opensearch-build` change is off by default, so it can go in independently. The `ci.yml` change doesn't really make sense until the whole series is merged - a good reason for a follow-up PR.

src/main/java/org/opensearch/security/support/FipsMode.java
beanuwave: You are right, it's unnesseccary since java provides `Unsafe.allocateInstance` and `ReflectionFactory.newConstructorForSerialization` and the supplier is the actual attack surface.

gradle/fips.gradle
```diff
@@ -34,7 +34,8 @@
  * FIPS mode is driven by the OPENSEARCH_FIPS_MODE environment variable.
  */
 
-ext.fipsTestsEnabled = 'true'.equalsIgnoreCase(System.getenv('OPENSEARCH_FIPS_MODE'))
+def fipsBuildParams = rootProject.buildscript.classLoader.loadClass('org.opensearch.gradle.info.FipsBuildParams')
```
reta, line 37: Wow, this is unexpected, the security plugin applies OpenSearch build tooling plugins, why we need that?

beanuwave: The root buildscript classpath isn't inherited by scripts applied with `apply from:` so adding `import org.opensearch.gradle.info.FipsBuildParams` to `fips.gradle` fails to compile with `unable to resolve class …`.

Looking up the class through the root build script's classloader works around this, since the class is resolved when the script executes rather than when it's compiled.

How would you prefer to reuse `FipsBuildParams` from the OpenSearch build?

reta: Thanks @beanuwave , you are right, it is (for some reasons) is not possible. Couple of suggestions here:

- in `build.gradle`, add `ext` closure to `allprojects`:
```
import org.opensearch.gradle.info.FipsBuildParams
...

allprojects {
    group = "org.opensearch"
    version = opensearch_build

    ext {
        isInFipsApprovedOnlyMode = FipsBuildParams::isInFipsApprovedOnlyMode
    }
}
```

- use it like this in `fips.gradle`:
```
ext.fipsTestsEnabled = ext.isInFipsApprovedOnlyMode()
```

reta (commented):
Thanks a lot @beanuwave !

beanuwave:
Thanks for the review! @reta

reta:
@cwperks anything from your side? thank you!

cwperks:
> @cwperks anything from your side? thank you!

The change looks good to me.

reta (approved):

PR #6395, opened 2026-08-13
Model key and trust store configurations as records and support PKCS#11

beanuwave:
### Description
**Category:** Enhancement, Refactoring

Independent of FIPS: this adds PKCS#11 (HSM) support for node TLS and replaces ad-hoc `isPkcs11()`-style branching with a type-driven model. Reviewable without any FIPS context.

### Key changes

- **Sealed records replace branching.** Key stores and trust stores are modelled as `pem` / `jdk` / `pkcs11` records, and PKCS#11 dispatch moves into the records themselves rather than being re-decided at each call site.
- **`StorePassword` wrapper.** Store passwords are wrapped in a type whose `toString()` redacts the value, so a password cannot reach a log through an accidental interpolation.
- **A PKCS#11 store has no file.** The key lives on the token, so the path becomes optional throughout: `KeyStoreUtils` loads such stores with a null stream, and error messages name the token instead of a non-existent file. `PemKeyReader` learns the `PKCS11` store type and validates that a PKCS#11 provider is actually registered, rather than failing obscurely later.
- **PKCS#11 keys are signed via SunJSSE, not BCJSSE.** A PKCS#11/HSM private key is non-exportable and BCJSSE cannot sign with it (`no encoding for key`); SunJSSE delegates the handshake signature to the key's own provider (SunPKCS11). So when the keystore is PKCS#11, `SslConfiguration` builds the `SSLContext` against SunJSSE (`SslContextBuilder.sslContextProvider(SunJSSE)`).
- **Ignored settings now warn.** Trust store settings that a PKCS#11 configuration disregards produce a warning naming the exact setting, instead of being silently dropped.

### Reviewer notes

- Reviewable as two commits: the records refactor (no capability change), then the PKCS#11 capability on top.
- One test placeholder: `PemSslCertificatesLoaderTest` uses `nextAlphanumeric(10)` where the final branch uses `PasswordValidator.FIPS_MIN_PASSWORD_LENGTH`. That constant does not exist until PR 5 — this is deliberate, please don't "fix" it here.

### Testing

```bash
./gradlew test integrationTest
```

<details>
<summary>Manual test: node TLS key on a PKCS#11 token (SoftHSM), non-FIPS</summary>

Exercises a token-resident node TLS key (signed via SunJSSE) with a PKCS#12 file trust store, on a plain non-FIPS JVM. `securityadmin` still authenticates with the PEM `kirk` cert here — PKCS#11 client keys in the CLI arrive with PR 5. All paths are relative to `$OPENSEARCH_HOME`.

```bash
# 1. Set up a test cluster with the demo configuration.
cd $OPENSEARCH_HOME
sh plugins/opensearch-security/tools/install_demo_configuration.sh -y -i -s

# 2. Init token - the --pin becomes the keystore password.
softhsm2-util --init-token \
  --free \
  --label opensearch \
  --so-pin 4321 \
  --pin 1234

# 3. Configure SunPKCS11 in a java.security override (config/pkcs11.java.security).
#    Index 12 is the JDK's unconfigured SunPKCS11 slot; this line configures it.
      security.provider.12=SunPKCS11 /path/to/config/softhsm-pkcs11.cfg
#    with softhsm-pkcs11.cfg in the same dir:
      name = SoftHSM
      library = /usr/lib/softhsm/libsofthsm2.so
      slotListIndex = 0

# 4. Import the node key WITH its chain (PKCS#12 -> token).
openssl pkcs12 -export \
  -inkey config/esnode-key.pem \
  -in config/esnode.pem \
  -certfile config/root-ca.pem \
  -name esnode-cert \
  -out /tmp/esnode.p12 \
  -passout pass:1234

jdk/bin/keytool \
  -importkeystore \
  -srckeystore /tmp/esnode.p12 \
  -srcstoretype PKCS12 \
  -srcstorepass 1234 \
  -srcalias esnode-cert \
  -destkeystore NONE \
  -deststoretype PKCS11 \
  -deststorepass 1234 \
  -destalias esnode-cert \
  -addprovider SunPKCS11 \
  -providerarg config/softhsm-pkcs11.cfg

# 5. Trust anchor -> PKCS#12 file.
jdk/bin/keytool -importcert -noprompt \
  -alias root-ca \
  -file config/root-ca.pem \
  -keystore config/root-ca.p12 \
  -storetype PKCS12 \
  -storepass changeit

# 6. In opensearch.yml: comment the demo *.pem*_filepath lines, keep admin_dn: CN=kirk,...,
#    and add for both transport and http:

# --- Node identity from the PKCS#11 token (no keystore_filepath: the key lives on the token) ---
plugins.security.ssl.transport.keystore_type: PKCS11
plugins.security.ssl.transport.keystore_alias: esnode-cert
plugins.security.ssl.transport.keystore_password: "1234"      # SoftHSM PIN
plugins.security.ssl.http.keystore_type: PKCS11
plugins.security.ssl.http.keystore_alias: esnode-cert
plugins.security.ssl.http.keystore_password: "1234"

# --- Trust anchor from a PKCS#12 file ---
plugins.security.ssl.transport.truststore_type: PKCS12
plugins.security.ssl.transport.truststore_filepath: root-ca.p12
plugins.security.ssl.transport.truststore_password: "changeit"
plugins.security.ssl.http.truststore_type: PKCS12
plugins.security.ssl.http.truststore_filepath: root-ca.p12
plugins.security.ssl.http.truststore_password: "changeit"

# 7. Start the node with the override (single '=' appends to the JDK's java.security).
OPENSEARCH_JAVA_OPTS="-Djava.security.properties=$OPENSEARCH_HOME/config/pkcs11.java.security" \
  ./bin/opensearch

# 8. The HTTP layer serves the token-resident esnode cert (verify chain against root-ca):
openssl s_client -connect localhost:9200 -CAfile config/root-ca.pem </dev/null 2>/dev/null \
  | grep -E "subject=|Verify return code"          # expect CN=node-0.example.com, "0 (ok)"

# 9. Apply security config over that TLS layer. Pass = 'Done with success'.
sh plugins/opensearch-security/tools/securityadmin.sh \
  -cd config/opensearch-security/ \
  -icl \
  -nhnv \
  -cacert config/root-ca.pem \
  -cert config/kirk.pem \
  -key config/kirk-key.pem
```

Negative checks (restart per check, restore afterwards):

- Wrong PIN (`keystore_password: "0000"`) — node refuses to start with `Failed to load keystore from the PKCS#11 token`.
- Start without `OPENSEARCH_JAVA_OPTS` (no PKCS#11 provider configured) — node refuses to start at config load with the same message, caused by `KeyStoreException: PKCS11 not found`, rather than failing later in a handshake.

</details>

### Check List


cwperks (commented):
ty @beanuwave the refactor generally looks good to me and nice to see support reinstated for pkcs11 format.

src/main/java/org/opensearch/security/ssl/config/KeyStoreConfiguration.java
```diff
@@ -51,124 +76,118 @@ default Set<X500Principal> getIssuerDns() {
...
+    /**
+     * A file-based key store in any type a registered provider offers, e.g. JKS, JCEKS, PKCS12 or BCFKS.
+     *
+     * @param path location of the key store file
+     * @param type store type, as resolved from the settings or detected from the file
+     * @param alias optional alias of the key entry to use
+     * @param keyStorePassword password of the store itself
+     * @param keyPassword password of the key entry
+     */
+    record JdkKeyStoreConfiguration(Path path, String type, String alias, StorePassword keyStorePassword, StorePassword keyPassword)
```
reta, line 105: May be let's have:

```
static final String TYPE = "JDK";
```

to have this configuration accessible and check here as well https://github.com/opensearch-project/security/pull/6395/changes#diff-b57966d9fb4b4b3fe9bb72446abebb0076136e0720dac30764dc93946f42326bR52 (otherwise it looks like we only care about PKCS11.

src/main/java/org/opensearch/security/ssl/config/KeyStoreUtils.java
```diff
@@ -104,23 +108,119 @@ public static X509Certificate[] x509Certificates(final Path file) {
...
+        final var list = listBuilder.build();
+        if (list.isEmpty()) {
+            throw new OpenSearchException("The truststore " + source + " does not contain any certificates");
+        }
+        return list;
+    }
+
+    public static KeyStore loadTrustStore(final Path path, final String type, final String alias, final char[] password) {
+        final var trustStore = loadKeyStore(path, type, password);
+        return alias != null ? narrowToAlias(trustStore, type, alias, path.toString()) : trustStore;
```
reta, line 203:
```suggestion
        return Strings.isEmpty(alias) == false ? narrowToAlias(trustStore, type, alias, path.toString()) : trustStore;
```

src/main/java/org/opensearch/security/ssl/config/KeyStoreUtils.java
```diff
@@ -104,23 +108,119 @@ public static X509Certificate[] x509Certificates(final Path file) {
...
+    }
+
+    /**
+     * Copies the certificate of a single alias into a new in-memory store of {@code targetType}, so that only
+     * that certificate is trusted. The source store is never modified.
+     */
+    public static KeyStore narrowToAlias(final KeyStore trustStore, final String targetType, final String alias, final String source) {
```
reta, line 210: May be nit, but `singleAliasCopy` (or `aliasCopy`) looks a bit more aligned with what method does:
```suggestion
    public static KeyStore copyAlias(final KeyStore trustStore, final String targetType, final String alias, final String source) {
```

src/main/java/org/opensearch/security/ssl/config/SslCertificatesLoader.java
```diff
@@ -61,109 +61,145 @@ public SslCertificatesLoader(final String sslConfigSuffix, final String extended
...
-                    resolvePassword(
-                        fullSslConfigSuffix + KEYSTORE_KEY_PASSWORD,
-                        settings,
-                        keyStorePassword != null ? String.valueOf(keyStorePassword) : null
-                    )
+        final var keyStoreType = settings.get(sslConfigSuffix + KEYSTORE_TYPE);
+        final var trustStoreType = settings.get(sslConfigSuffix + TRUSTSTORE_TYPE);
+        final boolean isPkcs11Keystore = PemKeyReader.PKCS11.equalsIgnoreCase(keyStoreType);
+        final boolean isPkcs11Truststore = PemKeyReader.PKCS11.equalsIgnoreCase(trustStoreType);
+        final boolean usesKeyStore = settings.hasValue(sslConfigSuffix + KEYSTORE_FILEPATH) || isPkcs11Keystore;
```
reta, line 68: Why we need to factor isPkcs11Keystore  / isPkcs11Truststore  in here? TrustStoreConfiguration and KeyStoreConfiguration are aware of PKCS11 / JDK types. If the TRUSTSTORE_FILEPATH / KEYSTORE_FILEPATH is not present but  KEYSTORE_TYPE is, the TrustStoreConfiguration / KeyStoreConfiguration will fail.

src/main/java/org/opensearch/security/ssl/config/StorePassword.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.util.Arrays;
+
+/**
+ * A key store password, key password or PKCS#11 token PIN - a critical security parameter. Wrapping the
+ * characters keeps them off the public surface of the store configurations: the array is unwrapped only
+ * within this package, at the JCA calls that insist on a {@code char[]}. Equality is by content, and
+ * {@link #toString()} never reveals the characters.
+ */
+public final class StorePassword {
```
reta, line 19: Could we use `com.password4j.SecureString` ? (we already use this library)

src/main/java/org/opensearch/security/ssl/config/SslCertificatesLoader.java
reta: I think I found the answer:

>  // A PKCS#11 store lives on the token, not on disk, so it is the one case with no path.

Could you please add a comment (as in other places)? thank you

beanuwave: good idea! Those boolean conditions should be addressed.

src/main/java/org/opensearch/security/ssl/config/KeyStoreUtils.java
beanuwave: We should stick to `null` here. `""` is a valid alias for `JKS`, `JCEKS`, `PKCS12` (both `SUN` and `BCFIPS`), and `BCFKS`, and it can't be ambiguous there because each alias maps to exactly one entry.

`SunPKCS11` never produces a `""` alias - objects without a label get `0x<CKA_ID>`. For `PKCS12` files without an alias: `SUN` names the entry `1` and `BCFIPS` uses a hex ID, never `""`.

src/main/java/org/opensearch/security/ssl/config/KeyStoreConfiguration.java
beanuwave: I might have misread your comment here. As far as I understand it, `JdkKeyStoreConfiguration` already covers all the types offered by the installed providers (`JKS`, `PKCS12`, `BCFKS`, `JCEKS`, ...), with the concrete type carried as a record field.

`PKCS#11` is the only case that needs its own implementation. Since `KeyStore.getInstance()` needs a concrete type anyway, having `TYPE = "JDK"` would just be another alias for `JdkKeyStoreConfiguration`. I'm not sure what we'd gain from that.

src/main/java/org/opensearch/security/ssl/config/StorePassword.java
beanuwave: I was looking into reusing some existing classes, but nothing seemed appropriate:
- `com.password4j.SecureString` exposes no char[], so every JCA call would need a fresh copy of the password on the heap.
- `org.opensearch.core.common.settings.SecureString` cannot represent `null` (no password) and also returns the plaintext from `toString()`.

Both alternatives are build around the idea to hold a secret briefly and then wipe it - that's the opposite of what we need, since we keep the password for the node's lifetime so certificates can be reloaded.

reta: I see, thanks @beanuwave

beanuwave:
@rata @cwperks Thank you both for the detailed review and for catching all those little things!

src/main/java/org/opensearch/security/ssl/config/KeyStoreConfiguration.java
reta: Yeah, sorry, I should have been more specific. Looking  for `SECURITY_SSL_HTTP_KEYSTORE_TYPE` settting, it is just arbitrary string:

```
settings.add(Setting.simpleString(SSLConfigConstants.SECURITY_SSL_HTTP_KEYSTORE_TYPE, Property.NodeScope, Property.Filtered));
```

We do have dedicated PCKS#11 value now, but we don't have any other valid keystore types. My idea was to include the TYPE (which should be more like TYPES) property into `JdkKeyStoreConfiguration` and validate there is at least one of the supported keystore type (unless it could be anything really), thanks

PR #6402, opened 2026-08-14
Add setting to ignore source security roles on CCS requests

### LINKED ISSUE #6401: Add a flag to allow remote cluster to independently compute mapped roles for CCS
Currently during cross-cluster search (CCS), the remote cluster's ConfigurableRoleMapper includes the source cluster's pre-computed security roles in its role evaluation (union of source roles + remote's own role mappings). This means the effective roles on the remote cluster are a combination of both clusters' configurations.

This proposal introduces a cluster setting (flag) that allows the remote cluster to independently compute mapped roles using only its own roles_mapping.yml, based on the incoming user's username and backend_roles, without including security roles pre-computed by the source cluster.

Proposed setting:

`plugins.security.ccs.ignore_source_security_roles: true|false (default: false)`

When enabled, the remote cluster skips source-propagated security roles during CCS requests and evaluates access using only its own role mapping configuration.

Rollout:

Phase 1: Setting defaults to false (no behavior change). Users opt in.

Phase 2 (4.0.0): Flip default to true.

sharathkanaka:
### Description

* **Category:** Enhancement

* **Why these changes are required?**

On cross-cluster search (CCS) requests, the remote cluster inherits the source cluster's pre-computed security roles for the user. This prevents the remote cluster from independently controlling what permissions CCS users receive based on its own configuration.

* **What is the old behavior before changes and new behavior after changes?**

**Old behavior:** Remote cluster always unions source-propagated `securityRoles` into its own role mapping result for CCS requests. The remote cannot independently control what permissions a CCS user receives.

**New behavior:** A new cluster setting `plugins.security.ccs.ignore_source_security_roles` (default: false) allows users to skip source cluster propagated securityRoles on CCS requests. When enabled, the remote cluster evaluates access through its own roles_mapping.yml. This gives the remote cluster independent control over CCS user permissions.

### Issues Resolved

Resolves https://github.com/opensearch-project/security/issues/6401

Not a backport. No new permissions introduced.

### Testing

- **Unit tests (4):** `ConfigurableRoleMapperTest.CcsSkipSourceSecurityRolesTest` :  covers flag on/off with and without CCS request context
- **Integration tests (3):** `CcsIgnoreSourceSecurityRolesIntTests` : end-to-end CCS with two remote clusters 

### Check List

---

Note: Documentation will be added in a follow-up PR to the [documentation-website](https://github.com/opensearch-project/documentation-website) repo once this change is merged.

### DIFF
--- a/src/integrationTest/java/org/opensearch/security/privileges/int_tests/CcsIgnoreSourceSecurityRolesIntTests.java
+++ b/src/integrationTest/java/org/opensearch/security/privileges/int_tests/CcsIgnoreSourceSecurityRolesIntTests.java
@@ -1,0 +1,195 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.security.privileges.int_tests;
+
+import java.util.List;
+
+import com.google.common.collect.ImmutableList;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import org.opensearch.security.support.ConfigConstants;
+import org.opensearch.test.framework.TestSecurityConfig;
+import org.opensearch.test.framework.certificate.TestCertificates;
+import org.opensearch.test.framework.cluster.ClusterManager;
+import org.opensearch.test.framework.cluster.LocalCluster;
+import org.opensearch.test.framework.cluster.TestRestClient;
+import org.opensearch.test.framework.data.TestIndex;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.opensearch.test.framework.TestSecurityConfig.AuthcDomain.AUTHC_HTTPBASIC_INTERNAL;
+import static org.opensearch.test.framework.matcher.RestMatchers.isForbidden;
+import static org.opensearch.test.framework.matcher.RestMatchers.isOk;
+
+/**
+ * Integration test for the CCS remote role recomputation setting.
+ *
+ * Proves:
+ * 1. With flag=true, source-propagated securityRoles are stripped (CCS query denied for unmapped user)
+ * 2. With flag=true, roles_mapping on the remote still grants access independently (mapped user succeeds)
+ */
+public class CcsIgnoreSourceSecurityRolesIntTests {
+
+    private static final String REMOTE_CLUSTER_FLAG_ON = "remote_flag_on";
+    private static final String REMOTE_CLUSTER_FLAG_OFF = "remote_flag_off";
+    private static final String INDEX_NAME = "index_r1";
+    private static final String MAPPED_USER_NAME = "mapped_user";
+    private static final String UNMAPPED_USER_NAME = "unmapped_user";
+    private static final String UNLIMITED_ROLE_NAME = "unlimited_role";
+    private static final String READ_ROLE_REMOTE_NAME = "read_role_remote";
+
+    static final TestIndex REMOTE_INDEX = TestIndex.name(INDEX_NAME).documentCount(10).seed(1).build();
+
+    // Role that grants access to ALL indices (assigned to user on local cluster via securityRoles)
+    static final TestSecurityConfig.Role UNLIMITED_ROLE = new TestSecurityConfig.Role(UNLIMITED_ROLE_NAME).clusterPermissions("*")
+        .indexPermissions("*")
+        .on("*");
+
+    // Role that grants read-only on the remote index (mapped via roles_mapping on remote)
+    static final TestSecurityConfig.Role READ_ROLE_REMOTE = new TestSecurityConfig.Role(READ_ROLE_REMOTE_NAME).clusterPermissions(
+        "cluster_composite_ops_ro",
+        "cluster_monitor"
+    ).indexPermissions("read", "indices_monitor", "indices:admin/shards/search_shards").on(INDEX_NAME);
+
+    // User with unlimited access via securityRoles (source cluster), mapped via roles_mapping on remote
+    static final TestSecurityConfig.User MAPPED_USER = new TestSecurityConfig.User(MAPPED_USER_NAME).referencedRoles(UNLIMITED_ROLE);
+
+    // Second user: also has unlimited on source, but NO roles_mapping on remote
+    static final TestSecurityConfig.User UNMAPPED_USER = new TestSecurityConfig.User(UNMAPPED_USER_NAME).referencedRoles(UNLIMITED_ROLE);
+
+    // Roles mapping on remote: maps "mapped_user" -> "read_role_remote"
+    // Note: "unmapped_user" intentionally has NO mapping
+    static final TestSecurityConfig.RoleMapping READ_ROLE_MAPPING = new TestSecurityConfig.RoleMapping(READ_ROLE_REMOTE_NAME).users(
+        MAPPED_USER_NAME
+    );
+
+    static final List<TestSecurityConfig.User> USERS = ImmutableList.of(MAPPED_USER, UNMAPPED_USER);
+
+    static final TestCertificates TEST_CERTIFICATES = new TestCertificates();
+
+    // Remote cluster with flag=TRUE: ignores source securityRoles, uses only its own roles_mapping
+    @ClassRule
+    public static final LocalCluster remoteClusterFlagOn = new LocalCluster.Builder().certificates(TEST_CERTIFICATES)
+        .clusterManager(ClusterManager.SINGLENODE)
+        .clusterName(REMOTE_CLUSTER_FLAG_ON)
+        .authc(AUTHC_HTTPBASIC_INTERNAL)
+        .privilegesEvaluationType("v4")
+        .users(USERS)
+        .roles(UNLIMITED_ROLE, READ_ROLE_REMOTE)
+        .rolesMapping(READ_ROLE_MAPPING)
+        .nodeSetting(ConfigConstants.SECURITY_CCS_IGNORE_SOURCE_SECURITY_ROLES, true)
+        .indices(REMOTE_INDEX)
+        .build();
+
+    // Remote cluster with flag=FALSE: legacy behavior, source securityRoles propagate through
+    @ClassRule
+    public static final LocalCluster remoteClusterFlagOff = new LocalCluster.Builder().certificates(TEST_CERTIFICATES)
+        .clusterManager(ClusterManager.SINGLENODE)
+        .clusterName(REMOTE_CLUSTER_FLAG_OFF)
+        .authc(AUTHC_HTTPBASIC_INTERNAL)
+        .privilegesEvaluationType("v4")
+        .users(USERS)
+        .roles(UNLIMITED_ROLE, READ_ROLE_REMOTE)
+        .nodeSetting(ConfigConstants.SECURITY_CCS_IGNORE_SOURCE_SECURITY_ROLES, false)
+        .indices(REMOTE_INDEX)
+        .build();
+
+    // Local cluster: connects to BOTH remotes
+    @ClassRule
+    public static final LocalCluster localCluster = new LocalCluster.Builder().certificates(TEST_CERTIFICATES)
+        .clusterManager(ClusterManager.SINGLE_REMOTE_CLIENT)
+        .remote(REMOTE_CLUSTER_FLAG_ON, remoteClusterFlagOn)
+        .remote(REMOTE_CLUSTER_FLAG_OFF, remoteClusterFlagOff)
+        .authc(AUTHC_HTTPBASIC_INTERNAL)
+        .privilegesEvaluationType("v4")
+        .users(USERS)
+        .roles(UNLIMITED_ROLE)
+        .doNotFailOnForbidden(true)
+        .build();
+
+    /**
+     * With flag=true: source's unlimited_role is stripped.
+     * But remote's roles_mapping maps the user -> read_role_remote (read on index_r1).
+     * CCS query should SUCCEED via the remote's own roles_mapping.
+     */
+    @Test
+    public void ccsQuery_withFlagOn_shouldSucceedViaRolesMapping() throws Exception {
+        try (TestRestClient restClient = localCluster.getRestClient(MAPPED_USER)) {
+            TestRestClient.HttpResponse response = restClient.get(REMOTE_CLUSTER_FLAG_ON + ":" + INDEX_NAME + "/_search");
+            assertThat(response, isOk());
+        }
+    }
+
+    /**
+     * With flag=true: source's unlimited_role is stripped.
+     * Unmapped user has NO roles_mapping entry on remote.
+     * CCS query should be FORBIDDEN — proves source securityRoles are actually stripped.
+     */
+    @Test
+    public void ccsQuery_withFlagOn_shouldBeForbidden_whenNoRolesMapping() throws Exception {
+        try (TestRestClient restClient = localCluster.getRestClient(UNMAPPED_USER)) {
+            TestRestClient.HttpResponse response = restClient.get(REMOTE_CLUSTER_FLAG_ON + ":" + INDEX_NAME + "/_search");
+            assertThat(response, isForbidden());
+        }
+    }
+
+    /**
+     * With flag=false (legacy behavior): source's securityRoles propagate through.
+     * Unmapped user has unlimited_role from source — which exists on remote's roles.yml.
+     * CCS query should SUCCEED — proves legacy union behavior is preserved.
+     */
+    @Test
+    public void ccsQuery_withFlagOff_shouldSucceed_whenSourceRolesPropagate() throws Exception {
+        try (TestRestClient restClient = localCluster.getRestClient(UNMAPPED_USER)) {
+            TestRestClient.HttpResponse response = restClient.get(REMOTE_CLUSTER_FLAG_OFF + ":" + INDEX_NAME + "/_search");
+            assertThat(response, isOk());
+        }
+    }
+
+    /**
+     * Dynamic setting update: flip flag from false to true at runtime via PUT _cluster/settings.
+     * CCS query that previously succeeded should now be forbidden.
+     */
+    @Test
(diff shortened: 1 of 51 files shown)


cwperks:
Failing CI checks will be resolved by https://github.com/opensearch-project/security/pull/6407. There was a breaking change from core that this repo needs to react to.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -1599,8 +1599,14 @@ public Collection<Object> createComponents(
 
         rsIndexHandler = new ResourceSharingIndexHandler(localClient, threadPool, resourcePluginInfo);
 
+        ConfigurableRoleMapper configurableRoleMapper = new ConfigurableRoleMapper(cr, settings, threadPool.getThreadContext());
+        clusterService.getClusterSettings()
+            .addSettingsUpdateConsumer(SecuritySettings.CCS_IGNORE_SOURCE_SECURITY_ROLES_SETTING, newValue -> {
+                log.info("CCS ignore source security roles dynamically set to {}", newValue);
```
DarshitChanpura: we should audit log this.

src/main/java/org/opensearch/security/privileges/ConfigurableRoleMapper.java
```diff
@@ -59,16 +68,27 @@ public ConfigurableRoleMapper(ConfigurationRepository configurationRepository, R
...
+        this.ccsIgnoreSourceSecurityRoles.set(value);
     }
 
     @Override
     public ImmutableSet<String> map(User user, TransportAddress caller) {
         CompiledConfiguration activeConfiguration = this.activeConfiguration.get();
 
         if (activeConfiguration != null) {
-            return activeConfiguration.map(user, caller);
+            boolean isTrustedClusterRequest = threadContext != null
```
cwperks: Can we use [HeaderHelper.isRemoteClusterNodeRequest](http://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/HeaderHelper.java#L88-L90) here?

sharathkanaka: Oh nice, we already have an helper function. I will use that

src/main/java/org/opensearch/security/privileges/ConfigurableRoleMapper.java
```diff
@@ -59,16 +75,31 @@ public ConfigurableRoleMapper(ConfigurationRepository configurationRepository, R
...
     }
 
     @Override
     public ImmutableSet<String> map(User user, TransportAddress caller) {
         CompiledConfiguration activeConfiguration = this.activeConfiguration.get();
 
         if (activeConfiguration != null) {
-            return activeConfiguration.map(user, caller);
+            boolean isTrustedClusterRequest = HeaderHelper.isRemoteClusterNodeRequest(threadContext);
+            boolean ignoreSourceRoles = ccsIgnoreSourceSecurityRoles.get();
```
cwperks: Since `ccsIgnoreSourceSecurityRoles` and threadcontext are available at the classlevel, can we reference them directly in the `map` method instead of creating another overload of the method?

src/main/java/org/opensearch/security/privileges/ConfigurableRoleMapper.java
```diff
@@ -42,8 +46,20 @@ public class ConfigurableRoleMapper implements RoleMapper {
     private final static Logger log = LogManager.getLogger(ConfigurableRoleMapper.class);
 
     private final AtomicReference<CompiledConfiguration> activeConfiguration = new AtomicReference<>();
+    private final AtomicBoolean ccsIgnoreSourceSecurityRoles = new AtomicBoolean(false);
+    private final ThreadContext threadContext;
+    private final Settings settings;
```
DarshitChanpura: Dead field: `settings` is assigned in the constructor and never read afterward. Read the value from the constructor param directly and drop the field.

src/main/java/org/opensearch/security/privileges/ConfigurableRoleMapper.java
```diff
@@ -59,16 +75,31 @@ public ConfigurableRoleMapper(ConfigurationRepository configurationRepository, R
...
     @Override
     public ImmutableSet<String> map(User user, TransportAddress caller) {
         CompiledConfiguration activeConfiguration = this.activeConfiguration.get();
 
         if (activeConfiguration != null) {
-            return activeConfiguration.map(user, caller);
+            boolean isTrustedClusterRequest = HeaderHelper.isRemoteClusterNodeRequest(threadContext);
+            boolean ignoreSourceRoles = ccsIgnoreSourceSecurityRoles.get();
+
+            return activeConfiguration.map(user, caller, isTrustedClusterRequest && ignoreSourceRoles);
```
DarshitChanpura: Worth calling out in the docs PR: this only strips *non-injected* source roles. Requests carrying injected roles short-circuit in `InjectedRoleMapper.map` and never reach here, so "ignore source security roles" isn't quite "ignore *all* source roles."

src/integrationTest/java/org/opensearch/security/privileges/int_tests/CcsIgnoreSourceSecurityRolesIntTests.java
```diff
@@ -0,0 +1,195 @@
...
+            assertThat(response, isOk());
+        }
+    }
+
+    /**
+     * Dynamic setting update: flip flag from false to true at runtime via PUT _cluster/settings.
+     * CCS query that previously succeeded should now be forbidden.
+     */
+    @Test
+    public void ccsQuery_withFlagDynamicallyEnabled_shouldBeForbidden() throws Exception {
```
DarshitChanpura: This flips a persistent setting on the shared `remoteClusterFlagOff` cluster; the `finally` reset covers it, but it leaves an ordering dependency with `ccsQuery_withFlagOff_shouldSucceed_whenSourceRolesPropagate`. A dedicated cluster or transient settings would decouple them.

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
sharathkanaka: This should take care of it     [private void logClusterSettingsChange(String action, ClusterUpdateSettingsRequest request, Task task) {](https://github.com/opensearch-project/security/blob/af1d7eea4d1fa67ad681f5e711cbc515a6240699/src/main/java/org/opensearch/security/auditlog/impl/AbstractAuditLog.java#L435) ?

src/main/java/org/opensearch/security/privileges/ConfigurableRoleMapper.java
sharathkanaka: thanks, wil update

src/main/java/org/opensearch/security/privileges/ConfigurableRoleMapper.java
sharathkanaka: Yeah, will do when I add doc update

src/integrationTest/java/org/opensearch/security/privileges/int_tests/CcsIgnoreSourceSecurityRolesIntTests.java
sharathkanaka, line 181: Although this is rare case, worth covering it. Will add dedicated cluster, changing to transient will not make a difference as the cluster is still up

src/main/java/org/opensearch/security/privileges/ConfigurableRoleMapper.java
sharathkanaka: i tried to avoid introducing any more additional outer fields in static class CompiledConfiguration, but should be fine to pass them as constructor params. Updating it

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
cwperks: ^ yes that does cover the case.

src/main/java/org/opensearch/security/privileges/ConfigurableRoleMapper.java
```diff
@@ -139,14 +175,29 @@ static class CompiledConfiguration implements RoleMapper {
         private List<WildcardMatcher> barMatchers;
         private List<WildcardMatcher> hostMatchers;
 
+        /**
+         * Convenience constructor for non-CCS contexts (CCS flag defaults to false).
+         */
         CompiledConfiguration(
```
cwperks, line 142: Is this old constructor used anywhere? Let's remove it if its unused.

sharathkanaka: yep, I was about send a new revision for this :)

src/main/java/org/opensearch/security/privileges/ConfigurableRoleMapper.java
```diff
@@ -42,8 +46,18 @@ public class ConfigurableRoleMapper implements RoleMapper {
...
+    public ConfigurableRoleMapper(
+        ConfigurationRepository configurationRepository,
+        ResolutionMode resolutionMode,
+        ThreadContext threadContext,
+        Settings settings
+    ) {
+        this.threadContext = threadContext;
+        this.ccsIgnoreSourceSecurityRoles.set(settings.getAsBoolean(ConfigConstants.SECURITY_CCS_IGNORE_SOURCE_SECURITY_ROLES, false));
 
-    public ConfigurableRoleMapper(ConfigurationRepository configurationRepository, ResolutionMode resolutionMode) {
```
cwperks, line 46: I was thinking more about this change and I think we should leave `ConfigurableRoleMapper` alone and not introduce state like threadcontext into this class.

It looks like we are already computing mapped roles on the remote cluster, but its taking a union of that and the mapped roles also ferried from the local cluster?

If so, then maybe we should consider sanitizing on the receiver end in the remote cluster?

i.e.

```
SecurityRequestHandler
    identifies trusted-remote request
    deserializes user
    applies RemoteClusterIdentityPolicy
    rewrites transient/persistent/header identity state
```

Where `RemoteClusterIdentityPolicy` is something like:

```
final class RemoteClusterIdentityPolicy {
    private volatile boolean ignoreSourceSecurityRoles;

    User sanitize(User user, ThreadContext threadContext) {
        if (!ignoreSourceSecurityRoles
            || !HeaderHelper.isRemoteClusterNodeRequest(threadContext)) {
            return user;
        }

        return user.withoutSecurityRoles();
    }
}
```

and `user.withoutSecurityRoles();` something like:

```
public User withoutSecurityRoles() {
    if (securityRoles.isEmpty()) {
        return this;
    }

    return new User(
        name,
        roles,
        ImmutableSet.of(),
        requestedTenant,
        attributes,
        isInjected
    );
}
```

This reminds me that the naming of `getRoles` and `getSecurityRoles` on the User class is confusing. I would be in favor of renaming these to `getBackendRoles` and `getOpenSearchRoles` respectively and updating any instances in this repo for clarity.

sharathkanaka: That makes sense.  Stripping it early is safer , that makes all the code path avoid consuming the source-propagated roles.  WIll make that change
I think renaming can happen in separate PR to avoid confusions.

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
```diff
@@ -206,7 +209,10 @@ protected void messageReceivedDecorate(
                         getThreadContext().putTransient(ConfigConstants.OPENDISTRO_SECURITY_INJECTED_USER, injectedUserHeader);
                     }
                 } else {
-                    user = user != null ? user : this.userFactory.fromSerializedBase64(userHeader);
+                    if (user == null) {
+                        user = this.userFactory.fromSerializedBase64(userHeader);
+                    }
+                    user = remoteClusterIdentityPolicy.sanitize(user, getThreadContext());
```
DarshitChanpura: This rewrites the transient `OPENDISTRO_SECURITY_USER`, but the persistent `AUTHENTICATED_USER` subject was already built from the *unsanitized* user just above (line 188), so the two diverge on CCS requests. `ResourceAccessHandler.getFlatPrincipals` reads that subject and derives principals from `user.getSecurityRoles()` (`ResourceAccessHandler.java:413`) — so with the flag on, resource-sharing access is still evaluated using the source-propagated roles this feature is meant to ignore. This matches the flow sketched earlier in the thread ("rewrites transient/**persistent**/header identity state") — the persistent part is missing. Suggest sanitizing once, before the subject is built, and using the sanitized user for both. (Scoped: main index/cluster authz uses the sanitized transient user and is fine; `DlsFlsValveImpl` only uses the subject for an admin-name check.)

src/main/java/org/opensearch/security/transport/RemoteClusterIdentityPolicy.java
```diff
@@ -0,0 +1,41 @@
...
+    public void setIgnoreSourceSecurityRoles(boolean value) {
+        this.ignoreSourceSecurityRoles = value;
+    }
+
+    /**
+     * Strips source-propagated securityRoles on trusted cluster (CCS) requests when
+     * {@code plugins.security.ccs.ignore_source_security_roles} is enabled.
+     * For non-CCS requests, the User is returned unchanged.
+     */
+    User sanitize(User user, ThreadContext threadContext) {
```
DarshitChanpura, line 40: `sanitize` is only reached on the non-injected user-header branch in `SecurityRequestHandler`, so requests carrying injected roles skip it. That keeps CCR safe (it authorizes via injected roles), but a source cluster propagating *injected* roles won't be stripped — worth a line in the docs so "ignore source security roles" isn't read as "ignore all source roles."

src/main/java/org/opensearch/security/transport/RemoteClusterIdentityPolicy.java
```diff
@@ -0,0 +1,41 @@
...
+        if (ignoreSourceSecurityRoles && HeaderHelper.isRemoteClusterNodeRequest(threadContext)) {
+            return user.withoutSecurityRoles();
```
DarshitChanpura: Silent strip. A `log.debug` (or audit hook) when roles are dropped would help operators confirm the flag is taking effect and trace CCS authorization changes.

src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java
sharathkanaka: Oh I see that. I will move the sanitizing part before the persistent user is built so both  get the sanitized user.

Side note: With immutable User, anyone adding transformations here in the future will hit the same ordering trap. Is this something the team has discussed? Also, kinda confused with choosing Transient and Persistent user, is there a room for simplicity here?

src/main/java/org/opensearch/security/transport/RemoteClusterIdentityPolicy.java
sharathkanaka: Will do in the doc update PR

DarshitChanpura (dismissed):

DarshitChanpura:
Minor / defense-in-depth (not a blocker): the sanitize happens at receive time on the transient + persistent user, but on a **multi-node** remote cluster the coordinator's onward fan-out reuses the original `OPENDISTRO_SECURITY_USER_HEADER`. In `SecurityInterceptor.sendRequestDecorate` the incoming header is copied into the forwarded `headerMap` (it's in the copy allowlist), and `ensureCorrectHeaders` only rebuilds `USER_HEADER` when it's absent — so it's reused as-is rather than re-derived from the sanitized user. On the receiving data node the hop is intra-cluster (`INTERCLUSTER_REQUEST`, not `TRUSTED_CLUSTER_REQUEST`), so `RemoteClusterIdentityPolicy.sanitize` no-ops and the source `securityRoles` are present again.

The feature's core gate is unaffected — the remote **coordinator** authorizes with the sanitized user, so access control still works on multi-node clusters. This is purely about the source roles physically still reaching remote data nodes, i.e. the "header identity state" mentioned earlier in the thread that isn't rewritten.

Suggestions (optional for this PR, could be a follow-up):
- Also rewrite/clear `OPENDISTRO_SECURITY_USER_HEADER` / `OPENDISTRO_SECURITY_AUTHENTICATED_USER_HEADER` from the sanitized user so onward propagation stays sanitized.
- Add a multi-node remote to the integration test — all three remotes are currently `ClusterManager.SINGLENODE`, so coordinator == data node and this cross-node path is never exercised.

DarshitChanpura (approved):
Approving as non-blocker can be addressed in a follow-up PR.

sharathkanaka:
Darshit, replying to your comment on Coordinator to Data node. That is a good callout. Though no more authorization happens in data node, the user will re-appear with the security roles.
The current implementation has one logic, where it strips out if the orignal requests is CCS, why should that not apply for intra cluster nodes? I mean when data node receives the request, that should sanitize user as well right? regardless authz happens or not, its about correctness, we do not want to have user objects in inconsistent state across nodes. 
We can pass the `_opendistro_security_ccs_origin ` header in the [headerMap](https://github.com/opensearch-project/security/blob/2b32dab43d96723cb6e0d344a1c68626e74e133e/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java#L191-L213) will make the data node to behave same way?

DarshitChanpura:
Reason it doesn't sanitize on the intra-cluster hop today: sanitize is gated on HeaderHelper.isRemoteClusterNodeRequest, which only returns true for TRUSTED_CLUSTER_REQUEST. The data-node hop comes in as INTERCLUSTER_REQUEST, so it no-ops and the source securityRoles re-appear. In theory it should sanitize the same way a single node does, but a test proving it strengthens the claim and future-proofs the behavioral change. We also filter out headers not part of the map before serializing and sending across, so to preserve the CCS-origin signal we'd need to pass it in the map. Adding as a suggestion since I'm not completely sure there's no edge case here leading to inconsistent state — fine to address in a follow-up.

cwperks (approved):

kkhatua:
@sharathkanaka  wouldn't this need backport to 2.19 branch too?

kkhatua:
> The backport to `2.19` failed. Please backport manually. See failed workflow run: https://github.com/opensearch-project/security/actions/runs/34258969954

@sharathkanaka  could you please look into this and help merge it in?

PR #6409, opened 2026-08-17
Support wait_for_completion on Security configuration write APIs (#6337)

### LINKED ISSUE #6337: Support asynchronous Security configuration APIs with wait_for_completion
## Description

Security configuration APIs such as create/update role currently execute synchronously and do not expose their work through the OpenSearch task framework. This makes it difficult for clients to submit configuration changes without holding an HTTP connection open, monitor progress, or distinguish an in-progress operation from a lost request.

Add support for the standard `wait_for_completion` query parameter to Security configuration write APIs, starting with role create/update and ideally using a shared implementation for other applicable Security REST APIs.

## Proposed behavior

- Preserve backwards compatibility by defaulting `wait_for_completion` to `true`.
- With `wait_for_completion=true`, wait for the operation to complete and return the existing API response.
- With `wait_for_completion=false`, submit the operation through the OpenSearch task framework and immediately return its task ID.
- Allow callers to query the task through the standard Tasks API, for example `GET /_tasks/{task_id}`.
- Ensure the task represents the full configuration update lifecycle, including persistence and any cluster-wide propagation/acknowledgement the API currently waits for.
- Define completed-task result/error retention using the established task framework conventions rather than a Security-specific status mechanism.

Example:

```http
PUT /_plugins/_security/api/roles/my_role?wait_for_completion=false
```

```json
{
  "task": "node-id:task-id"
}
```

## Initial scope

At minimum:

- create role
- update role

The implementation should identify other Security configuration write endpoints that can use the same infrastructure, such as users, role mappings, action groups, tenants, and configuration updates.

## Acceptance criteria

- Security configuration write APIs accept `wait_for_completion`.
- Existing clients observe no response behavior change when the parameter is omitted.
- `wait_for_completion=false` returns a valid task ID without waiting for the operation to finish.
- Task status, completion, failure, and cancellation semantics are exposed consistently through the OpenSearch task framework.
- Authorization for submitting and inspecting tasks does not expose Security configuration contents to unauthorized users.
- Integration tests cover synchronous success/failure, asynchronous success/failure, and task lookup.

nishthm:
### Description
**Category:** Enhancement

Adds support for the standard `wait_for_completion` query parameter to Security configuration write APIs. Callers can now submit a configuration change with `?wait_for_completion=false` and receive an OpenSearch task ID immediately (`{"task":"<nodeId>:<taskId>"}`), then poll the outcome via the standard Tasks API (`GET /_tasks/{task_id}`). Omitting the parameter — or setting it to `true` — preserves the existing synchronous behavior byte-for-byte.

**Why:** Security configuration writes today fan out a cluster-wide reload after the index write, and the caller has to hold the HTTP connection open for the entire lifecycle. If the connection drops there's no way to tell whether the change committed. Every other long-running OpenSearch API (reindex, update-by-query, force-merge, open-index…) solves this via the Tasks framework; Security did not. This change brings Security in line with that convention.

**Old vs new behavior:**

| `wait_for_completion` | Behavior |
|---|---|
| omitted / `true` (default) | Unchanged. Same response body and status code the endpoint returned before this PR. |
| `false` | Update is submitted through the task framework. HTTP 200 with body `{"task":"nodeId:taskId"}` is returned as soon as the task is registered — before the index write or fan-out reload has completed. |

**Task completion boundary.** The task completes only after `ConfigUpdatingActionListener` receives all-node acknowledgements of the reload — exactly the point at which the sync path returns today. Any per-node reload failure surfaces as a task-level error.

**Task result payload.** The stored task result in `.tasks` contains only `{"status":"...","message":"'entity' created."}` — never configuration contents. Standard `_tasks` authorization is therefore sufficient to satisfy the "does not expose Security configuration contents" acceptance criterion; no security-plugin-specific gating on task lookup was added.

**Cancellation.** Deliberately not supported. `SecurityConfigWriteRequest#createTask` returns a plain `Task` rather than a `CancellableTask`, so `POST /_tasks/{id}/_cancel` is rejected by `TransportCancelTasksAction` with `"task [...] doesn't support cancellation"`. Rationale: a mid-fan-out cancel could leave the index write committed but only a subset of nodes reloaded — worse than letting the request run to completion.

### Scope

**In this PR** — the shared plumbing plus opt-in for the 5 endpoints that a Terraform provider calls when provisioning a cluster:

- Roles (`RolesApiAction`)
- Role mappings (`RolesMappingApiAction`)
- Internal users (`InternalUsersApiAction`)
- Tenants (`TenantsApiAction`)
- Audit (`AuditApiAction`)

**Deferred:**
- `ActionGroupsApiAction`, `SecurityConfigApiAction` — same shape (config write + fan-out reload), share the same plumbing, but not called by the Terraform provider. Opt-in is a one-line change per subclass; follow-up PR.
- `AccountApiAction` (self-service password change) — explicitly excluded per feedback in the issue thread; async has near-zero value for an interactive endpoint.
- `AllowlistApiAction`, `NodesDnApiAction`, `RateLimitersApiAction`, `MultiTenancyConfigApiAction` — same pattern, follow-up as demand appears.

**Companion PRs to follow** (separate repositories):
- `opensearch-project/documentation-website` — document the new query parameter, response shape, task lookup.
- `opensearch-project/opensearch-api-specification` — add `wait_for_completion` to the affected endpoint schemas so generated client SDKs pick it up.

### Implementation

**New files (`src/main/java/org/opensearch/security/action/configupdate/`):**

- `SecurityConfigWriteAction` — `ActionType` with name `cluster:admin/opendistro_security/api/write_config`.
- `SecurityConfigWriteRequest` — carries `cType`, serialized config bytes, seqNo/primaryTerm for optimistic concurrency, security index name, task description, success message, and success status. `createTask` returns a plain (non-cancellable) `Task`. `getShouldStoreResult()` returns `true` so `TransportAction.execute` wraps the listener with `TaskResultStoringActionListener` automatically.
- `SecurityConfigWriteResponse` — `ActionResponse` + `ToXContentObject`; body is `{"status":"OK","message":"..."}` matching the sync path exactly. Deliberately does not include configuration contents.
- `TransportSecurityConfigWriteAction` — `HandledTransportAction` that performs the `IndexRequest` and then broadcasts `ConfigUpdateAction`; only completes the listener after all-node ack, mirroring the sync path's `ConfigUpdatingActionListener` chain.

**Modified files:**

- `OpenSearchSecurityPlugin.java` — registers the new transport action.
- `AbstractApiAction.java` — adds `supportsAsync()` hook (default `false`) and a private `maybeSubmitAsTask` helper wired via `withAsyncTaskSubmitter(...)`. Consumes `wait_for_completion` in `prepareRequest` directly (not inside `consumeParameters`) so subclass overrides that don't call `super` still don't reject the parameter as unrecognized.
- `RequestHandler.java` — introduces the `AsyncTaskSubmitter` functional interface and pre-branches `PUT`/`PATCH`/`DELETE` on it. If the endpoint doesn't opt in, `AsyncTaskSubmitter.NEVER` returns `false` and the sync path is unchanged. Response message and status are precomputed once from the loaded configuration state so the sync body and the stored task result are byte-identical.
- 5 endpoint classes — override `supportsAsync()` to return `true`.

**Backwards compatibility.** The pre-branch runs only when the endpoint has opted in *and* `wait_for_completion=false` is explicitly present. The sync response body was refactored to compute `(status, message)` once instead of twice, but produces the same JSON: `{"status":"CREATED","message":"'my_role' created."}` for a new PUT, `{"status":"OK","message":"'my_role' updated."}` for an update, `{"status":"OK","message":"'my_role' deleted."}` for a delete. The public static helper `AbstractApiAction.saveAndUpdateConfigsAsync` is left untouched because `RollbackVersionApiAction` and `ConfigUpgradeApiAction` still depend on its signature.

### Issues Resolved
Closes https://github.com/opensearch-project/security/issues/6337

Is this a backport? **No.** New feature, targets `main`.

Do these changes introduce new permission(s) to be displayed in the static dropdown on the front-end? **No** — the transport action's authorization runs through the same REST-admin check the existing sync path uses, at the REST layer, before task submission. No new plugin-specific action name needs a UI-side entry.

### Testing
**New integration test file:** `src/integrationTest/java/org/opensearch/security/api/WaitForCompletionRestApiIntegrationTest.java`

Covers each acceptance criterion in the issue:

- Sync success (parameter omitted, and parameter explicitly set to `true`) — response status and body match the pre-PR behavior.
- Sync failure — invalid body still returns 400 even when `wait_for_completion=false` is requested, because validation runs before task submission (no phantom task IDs for requests that never execute).
- Async success and task lookup — one test per opted-in endpoint (Roles, RolesMapping, InternalUsers, Tenants, Audit): submit with `wait_for_completion=false`, verify the returned task ID follows `<nodeId>:<taskId>` shape, poll `_tasks/{id}?wait_for_completion=true&timeout=30s`, assert `completed=true` and — for Roles and InternalUsers — that the stored task result contains the exact success message the sync path would have returned.
- Task cancellation refused — `POST /_tasks/{id}/_cancel` returns a response whose body contains `"doesn't support cancellation"`.

**Regression coverage:** the existing `RolesRestApiIntegrationTest` and `InternalUsersRestApiIntegrationTest` suites pass unmodified against these changes (17 tests, 0 failures).

**Manual verification** — run against a real multi-node local cluster built from this branch:

| # | Scenario | Expected | Actual |
|---|---|---|---|
| 1 | `PUT /roles/sync_role` (no param) | 201 + `"created"` | `HTTP 201  {"status":"CREATED","message":"'sync_role' created."}` |
| 2 | `PUT /roles/async_role?wait_for_completion=false` | 200 + `{"task":"nodeId:taskId"}` | `HTTP 200  {"task":"WkDLHW0bRzG7OmOx3bpeRg:133"}` |
| 3 | `GET /_tasks/{id}?wait_for_completion=true` | Stored result mirrors sync response | `completed:true, cancellable:false, action:"cluster:admin/opendistro_security/api/write_config", description:"roles/async_role", response:{"status":"CREATED","message":"'async_role' created."}` |
| 4 | `POST /_tasks/{id}/_cancel` | Refused | `caused_by: illegal_argument_exception "task [...] doesn't support cancellation"` |
| 5 | Invalid body + `wait_for_completion=false` | Sync 400, no phantom task | `HTTP 400  {"status":"error","reason":"Invalid configuration","invalid_keys":{"keys":"unknown_field"}}` |
| 6 | Async PUT on internal users | Task ID returned | `HTTP 200  {"task":"...:218"}` |
| 7 | Async PUT on **non**-opted-in endpoint (`actiongroups`) | Parameter silently ignored, sync response | `HTTP 201  {"status":"CREATED","message":"'manual_ag' created."}` |

  The task-manager output in row 3 also confirms `"cancellable":false` at the framework level, so `_tasks/{id}/_cancel` rejection in row 4 comes from
  `TransportCancelTasksAction` rather than any custom check — non-cancellability is a property of the `Task` type, not a plugin-level filter.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/security/dlic/rest/api/AbstractApiAction.java
+++ b/src/main/java/org/opensearch/security/dlic/rest/api/AbstractApiAction.java
@@ -38,6 +38,7 @@ import org.opensearch.core.common.bytes.BytesReference;
 import org.opensearch.core.common.transport.TransportAddress;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.xcontent.ToXContent;
+import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.xcontent.XContentHelper;
 import org.opensearch.index.engine.VersionConflictEngineException;
 import org.opensearch.rest.BaseRestHandler;
@@ -49,6 +50,8 @@ import org.opensearch.rest.RestRequestFilter;
 import org.opensearch.security.action.configupdate.ConfigUpdateAction;
 import org.opensearch.security.action.configupdate.ConfigUpdateRequest;
 import org.opensearch.security.action.configupdate.ConfigUpdateResponse;
+import org.opensearch.security.action.configupdate.SecurityConfigWriteAction;
+import org.opensearch.security.action.configupdate.SecurityConfigWriteRequest;
 import org.opensearch.security.dlic.rest.support.Utils;
 import org.opensearch.security.dlic.rest.validation.EndpointValidator;
 import org.opensearch.security.dlic.rest.validation.RequestContentValidator;
@@ -121,11 +124,99 @@ public abstract class AbstractApiAction extends BaseRestHandler implements RestR
     private void buildDefaultRequestHandlers(final RequestHandler.RequestHandlersBuilder builder) {
         builder.withAccessHandler(request -> securityApiDependencies.restApiAuthorizationEvaluator().isCurrentUserAdminFor(endpoint))
             .withSaveOrUpdateConfigurationHandler(this::saveOrUpdateConfiguration)
+            .withAsyncTaskSubmitter(this::maybeSubmitAsTask)
             .add(Method.POST, methodNotImplementedHandler)
             .add(Method.PATCH, methodNotImplementedHandler)
             .onGetRequest(this::processGetRequest)
             .onChangeRequest(Method.DELETE, this::processDeleteRequest)
             .onChangeRequest(Method.PUT, this::processPutRequest);
+    }
+
+    /**
+     * Endpoint-level opt-in for the {@code wait_for_completion} query parameter and the OpenSearch
+     * task framework. When {@code false} (the default), every change request is handled fully
+     * synchronously and {@code wait_for_completion} is ignored, exactly matching the pre-existing
+     * behavior. Subclasses that override this to {@code true} allow callers to pass
+     * {@code wait_for_completion=false}, in which case the update is submitted as a Task and
+     * {@code {"task":"nodeId:taskId"}} is returned immediately.
+     */
+    protected boolean supportsAsync() {
+        return false;
+    }
+
+    /**
+     * Pre-branch invoked from {@link RequestHandler.RequestHandlersBuilder#onChangeRequest} before
+     * the sync save path. Returns {@code true} only when the endpoint has opted in via
+     * {@link #supportsAsync()} <b>and</b> the caller passed {@code wait_for_completion=false} — the
+     * update is then dispatched through the task framework, a task-id response is written to the
+     * channel, and the caller is expected to skip the sync path.
+     *
+     * <p>Any other combination returns {@code false}, so the existing sync path handles the request
+     * unchanged.
+     */
+    private boolean maybeSubmitAsTask(
+        final RestChannel channel,
+        final RestRequest request,
+        final Client client,
+        final SecurityDynamicConfiguration<?> configuration,
+        final String entityName,
+        final String successMessage,
+        final RestStatus successStatus
+    ) {
+        if (!supportsAsync()) {
+            return false;
+        }
+        if (request.paramAsBoolean("wait_for_completion", true)) {
+            return false;
+        }
+        if (!(client instanceof NodeClient)) {
+            // Defensive: this handler always runs off a NodeClient today, so this branch is
+            // effectively unreachable — but avoid a runtime ClassCastException if that ever
+            // changes and gracefully fall back to sync execution.
+            LOGGER.debug("Client is not a NodeClient; cannot submit as task, falling back to sync path");
+            return false;
+        }
+        final NodeClient nodeClient = (NodeClient) client;
+        final CType<?> cType = getConfigType();
+
+        configuration.removeStatic();
+        final BytesReference content;
+        try {
+            content = XContentHelper.toXContent(configuration, XContentType.JSON, ToXContent.EMPTY_PARAMS, false);
+        } catch (final IOException e) {
+            throw ExceptionsHelper.convertToOpenSearchException(e);
+        }
+
+        final String description = entityName == null ? cType.toLCString() : cType.toLCString() + "/" + entityName;
+        final SecurityConfigWriteRequest updateRequest = new SecurityConfigWriteRequest(
+            cType.toLCString(),
+            content,
+            configuration.getSeqNo(),
+            configuration.getPrimaryTerm(),
+            securityApiDependencies.securityIndexName(),
+            description,
+            successMessage,
+            successStatus
+        );
+
+        // Persist the eventual task result in .tasks so callers can retrieve it via
+        // GET /_tasks/{task_id} after completion. The request itself always signals this via
+        // getShouldStoreResult() — see SecurityConfigWriteRequest.
+        final org.opensearch.tasks.Task task = nodeClient.executeLocally(
+            SecurityConfigWriteAction.INSTANCE,
+            updateRequest,
+            org.opensearch.tasks.LoggingTaskListener.instance()
+        );
+
+        try (final XContentBuilder builder = channel.newBuilder()) {
+            builder.startObject();
+            builder.field("task", nodeClient.getLocalNodeId() + ":" + task.getId());
+            builder.endObject();
+            channel.sendResponse(new BytesRestResponse(RestStatus.OK, builder));
+        } catch (final IOException e) {
+            throw ExceptionsHelper.convertToOpenSearchException(e);
+        }
+        return true;
     }
 
     protected final ValidationResult<SecurityConfiguration> processDeleteRequest(final RestRequest request) throws IOException {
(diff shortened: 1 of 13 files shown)


DarshitChanpura (commented):
Design's fine. Sync parity and the fails-closed transport path check out (header is stripped cross-node, so the privileged write can't be reached off the REST path). One thing to fix before merge, rest is minor.

**Fix:** `TransportSecurityConfigWriteAction#doExecute` re-implements the IndexRequest build and ConfigUpdate fan-out that already exist as `createIndexRequestForConfig` and `ConfigUpdatingActionListener`. Identical today, so a change to the write semantics hits sync only. Extract the shared builder/listener so they can't drift.

**Minor:**
- `SecurityConfigWriteResponse#writeTo` skips `super.writeTo(out)` but the read ctor calls `super(in)` — align them.
- `not_` helper cites a static-import collision, but `not` isn't imported here — use `CoreMatchers.not(...)`.
- Async DELETE/PATCH go through the submitter but only PUT is tested; add them.
- Task `description` is `cType/entityName`, readable via `_tasks` — names not contents, noting only.

Docs and api-specification companions are marked "to follow" — nothing blocks them, link the PRs.

nishthm:
@DarshitChanpura 

Pushed 086f134 on top of a rebase on latest main.

Required — reuse createIndexRequestForConfig and ConfigUpdatingActionListener. Done. Widened both to public static on
AbstractApiAction. SecurityConfigWriteRequest now carries a pre-built IndexRequest (constructed via the shared helper on the coordinator); TransportSecurityConfigWriteAction.doExecute collapses to a one-liner that wires it through ConfigUpdatingActionListener — the same call the sync save makes. Sync and async now share one path.

SecurityConfigWriteResponse#writeTo / read-ctor asymmetry. Kept super(in) and added a comment explaining why the asymmetry is inherent: Writeable.writeTo is abstract up through ActionResponse so there's no concrete super.writeTo to call, and TransportMessage(StreamInput) is documented as a no-op — nothing is actually asymmetric on the wire. Matches how every other ActionResponse in the repo is written (ConfigUpdateResponse in this same package included). On the Request side TransportRequest.writeTo IS concrete (writes the parent task id), so super.writeTo is called explicitly there.

not_ wrapper. Removed; static-imported CoreMatchers.not directly.

Async DELETE/PATCH not tested. Added asyncDeleteRole_returnsTaskIdAndRemoves and asyncPatchRole_returnsTaskIdAndUpdates — both verify task shape, completion, the stored task-result message, and the post-condition via sync GET.

Docs / api-spec companions. Will open both and link here.

cwperks:
> ActionGroupsApiAction, SecurityConfigApiAction — same shape (config write + fan-out reload), share the same plumbing, but not called by the Terraform provider. Opt-in is a one-line change per subclass; follow-up PR. 

This would also be good to do following up to this PR for consistency.

cwperks (approved):
Thank you @nishthm ! The changes in this PR look good to me. I like that this creates an opt-in model for the security endpoints that can be extended to other security endpoints in the future.

src/main/java/org/opensearch/security/dlic/rest/api/AbstractApiAction.java
```diff
@@ -591,6 +688,10 @@ protected final RestChannelConsumer prepareRequest(RestRequest request, NodeClie
         // consume all parameters first so we can return a correct HTTP status,
         // not 400
         consumeParameters(request);
+        // Consume the async opt-in flag centrally (not in consumeParameters), so subclasses that
+        // override consumeParameters — and don't call super — still don't reject
```
cwperks, line 692: Are there subclasses that don't call super? Should we consider adding to the subclasses?

Issue #1432, opened 2021-08-25
[BUG] Confusing index template permissions level


davidshtian:
**Describe the bug**
Try to add index template to role and map to a user, due to official [Access Control Permissions docs](https://opensearch.org/docs/security-plugin/access-control/permissions/), I tried to add **index permission** to the role.

For the new [index template] (https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html#index-templates), the setting _"indices:admin/index_template/get"_ (also this permission is not included in [Access Control Permissions docs](https://opensearch.org/docs/security-plugin/access-control/permissions/)) is okay.

However, for the [legacy template](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates-v1.html), the setting _"indices:admin/template/get"_ is not working, and I need to add this permission to cluster level permissions, and it works.

**To Reproduce**
1. Go to 'Security/Roles'
2. Click on 'Create role'
3. Add permission _"indices:admin/template/get"_ to **Index permissions**
4. See error below:
```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "security_exception",
        "reason" : "no permissions for [indices:admin/template/delete] and User [name=template, backend_roles=[], requestedTenant=__user__]"
      }
    ],
    "type" : "security_exception",
    "reason" : "no permissions for [indices:admin/template/delete] and User [name=template, backend_roles=[], requestedTenant=__user__]"
  },
  "status" : 403
}
```
5. Add permission _"indices:admin/template/get"_ to **Cluster permissions**, and the query executed successfully.

**Expected behavior**
Modify the _"indices:admin/index_template/get"_ permission to index level as its name showed.

**Plugins**
Security plugin.

**Screenshots**
Not working setting:
![image](https://user-images.githubusercontent.com/14228056/130817151-2e3fd869-4ff3-42a0-9ff4-9cb28b36d74f.png)
Working setting:
![image](https://user-images.githubusercontent.com/14228056/130817321-fab4b2d9-476e-4d5c-b87e-93f4e53eddd1.png)

**Host/Environment (please complete the following information):**
 - OS: Amazon Linux release 2
 - Version 1.0.0

**Additional context**
N/A

CEHENKLE:
Transferring issue to security plugin in.  Please let me know (@) if this is the wrong location.   Thanks!

scrawfor99:
[Triage] @cwperks could you take a minute to review this when you get a chance. Thank you.

[...]

Issue #3794, opened 2023-12-04
[BUG] DLS evaluation mode cannot be adapted on opensearch.yml


spapadop:
**What is the bug?**
OpenSearch documentation on [DLS](https://opensearch.org/docs/latest/security/access-control/document-level-security/#dls-evaluation-modes) states that one can tweak the DLS evaluation mode to `filter-level`, `lucene-level`, or `adaptive-level` on opensearch.yml

However, setting this on opensearch.yml produces an error, with the `plugins.security.dls.mode` setting not being recognised:
```
[2023-12-04T13:39:14,659][ERROR][o.o.b.OpenSearchUncaughtExceptionHandler] [node1] uncaught exception in thread [main]
org.opensearch.bootstrap.StartupException: java.lang.IllegalArgumentException: unknown setting [plugins.security.dls.mode] did you mean any of [plugins.security.disabled, plugins.security.audit.type, plugins.security.ssl_only, plugins.security.cert.oid]?
        at org.opensearch.bootstrap.OpenSearch.init(OpenSearch.java:184) ~[opensearch-2.11.1.jar:2.11.1]
        at org.opensearch.bootstrap.OpenSearch.execute(OpenSearch.java:171) ~[opensearch-2.11.1.jar:2.11.1]
        at org.opensearch.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:104) ~[opensearch-2.11.1.jar:2.11.1]
        at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138) ~[opensearch-cli-2.11.1.jar:2.11.1]
        at org.opensearch.cli.Command.main(Command.java:101) ~[opensearch-cli-2.11.1.jar:2.11.1]
        at org.opensearch.bootstrap.OpenSearch.main(OpenSearch.java:137) ~[opensearch-2.11.1.jar:2.11.1]
        at org.opensearch.bootstrap.OpenSearch.main(OpenSearch.java:103) ~[opensearch-2.11.1.jar:2.11.1]
```

**How can one reproduce the bug?**
1. Add `plugins.security.dls.mode: filter-level` on opensearch.yml
2. Restart the node

**What is the expected behavior?**
Either the DLS mode to be adapted accordingly, or the relevant documentation should be removed.

**What is your host/environment?**
 - OpenSearch v2.11.1

cwperks:
[Triage] @spapadop Thank you for filing this issue.

dgsardina:
[...]

PR #6416, opened 2026-08-20
Support hybrid search with adaptive DLS

sharp-pixel:
### Description

* Category: Bug fix

Hybrid search requests can fail when they are executed by a user with Document-Level Security (DLS). Neural Search requires a `HybridQuery` to remain the top-level Lucene query, while the Security plugin can wrap the user query in `BooleanQuery` and `ConstantScoreQuery` objects when enforcing DLS. Neural Search then rejects the query or fails while casting the wrapper to `HybridQuery`.

The original report also observed inconsistent behavior: adding a suggestion could make an otherwise equivalent request succeed because it selected a different DLS path.

#### Old behavior

In adaptive DLS mode, a top-level hybrid query could be wrapped by the Security plugin. This changed the query shape expected by Neural Search and could cause every shard to fail with errors such as:

* `hybrid query must be a top level query and cannot be wrapped into other queries`
* `BooleanQuery` or `ConstantScoreQuery` cannot be cast to `HybridQuery`

#### New behavior

For supported top-level hybrid requests in adaptive mode, the Security plugin now:

* Detects the hybrid query without adding a compile-time dependency on Neural Search.
* Applies the DLS query through `QueryBuilder.filter(...)`, preserving the hybrid query as the top-level query and propagating DLS into every hybrid subquery.
* Uses a separate internal marker so reader-level DLS remains active for global aggregations and other search paths that do not rely solely on the top-level query.
* Propagates that marker between nodes in the local cluster and removes it before cross-cluster search requests.
* Retains the existing star-tree safeguard before skipping query wrapping.

Non-hybrid queries and unsupported hybrid configurations continue through their existing DLS paths.

#### Supported scope

The special hybrid handling is enabled only when all of these conditions are true:

* DLS mode is `adaptive`.
* The request has DLS restrictions.
* The user query is a top-level `hybrid` query.
* The DLS rule does not contain a term-lookup query.
* The hybrid query contains no parent/child query clauses.
* The request targets only the local cluster.
* Every node in the cluster runs OpenSearch 3.9 or newer.

Within that scope, the fix supports:

* Single or multiple hybrid subqueries.
* Existing hybrid filters, which are composed with the DLS filter.
* DLS-filtered hits.
* Global aggregations protected by reader-level DLS.
* Searches containing suggestions.
* Multi-node local clusters.

#### Not supported by this change

The special handling is intentionally not enabled for:

* Explicit `filter_level` DLS mode.
* Explicit `lucene_level` DLS mode.
* Adaptive mode when the DLS rule contains a term-lookup query.
* Hybrid queries nested beneath another query.
* Hybrid queries containing parent/child clauses.
* Cross-cluster search.
* Mixed-version clusters containing a node older than OpenSearch 3.9.

These cases retain their existing DLS behavior and are not claimed to be compatible with hybrid search by this PR.

This change introduces no new setting, permission, REST API, or dependency on the Neural Search plugin.

### Issues Resolved

Addresses:

* https://github.com/opensearch-project/neural-search/issues/1303
* Original transferred issue: https://github.com/opensearch-project/ml-commons/issues/3815
* Original community report: https://forum.opensearch.org/t/hybrid-search-not-working-with-document-level-security/24239

Related prior work:

* https://github.com/opensearch-project/neural-search/pull/1432

Neural Search PR #1432 added support for recognizing and reconstructing the Security DLS wrapper shape available at the time. This PR complements that work on the Security side by avoiding the outer wrapper for supported hybrid requests.

This is not a backport.

These changes introduce no new permissions and require no Security Dashboards companion PR.

#### Security plugin unit tests

Added focused coverage for:

* Adaptive-mode selection and the DLS, version, and local-cluster gates.
* Requests without a source or query, non-search requests, and non-hybrid queries.
* Nested hybrid queries, parent/child clauses, and term-lookup DLS rules.
* Applying DLS through the hybrid filter method and composing an existing hybrid filter.
* Preserving reader-level DLS and protecting global aggregations.
* Retaining the star-tree safeguard.
* Propagating the internal marker to another local node and removing it before a remote-cluster request.

Executed:

```bash
./gradlew test \
  --tests org.opensearch.security.configuration.DlsFlsFilterLeafReaderTest \
  --tests org.opensearch.security.configuration.DlsFilterLevelActionHandlerTest \
  --tests org.opensearch.security.configuration.DlsFlsValveImplTest \
  --tests org.opensearch.security.privileges.dlsfls.DlsFlsBaseContextTest \
  --tests org.opensearch.security.transport.SecurityInterceptorTests
```

Result: passed.

```bash
./gradlew :precommit
```

Result: passed.

The Security plugin bundle was also built successfully.

#### Cross-plugin integration testing

Published the Security plugin ZIP to Maven Local:

```bash
cd /path/to/opensearch-security
./gradlew publishPluginZipPublicationToMavenLocal
```

Then ran the companion Neural Search test on a three-node cluster:

```bash
cd /path/to/neural-search
./gradlew :integTest \
  --tests org.opensearch.neuralsearch.query.HybridQueryDlsIT \
  -x spotlessApply \
  -Dsecurity.enabled=true \
  -PnumNodes=3 \
  --refresh-dependencies
```

Result: passed.

The test verifies the unrestricted administrator baseline, DLS-filtered hits, DLS-protected global aggregation counts and buckets, a suggestion-bearing request with an allowed-document suggestion, composition with an existing hybrid filter, a single-clause hybrid query, no duplicate hits, and no failed shards.

Running the same test with the unmodified Security plugin reproduced the original top-level `BooleanQuery`/`HybridQuery` failure.

The companion integration coverage is available in [opensearch-project/neural-search#1957](https://github.com/opensearch-project/neural-search/pull/1957). That draft depends on this Security PR and should merge only after this fix is available to the Neural Search integration-test build.

### Check List

- [x] New functionality has been documented — behavior and support boundaries are documented above and in code comments; no user-facing setting or API is introduced
- [x] New Roles/Permissions have a corresponding security dashboards plugin PR — N/A, no roles or permissions were added
- [x] API changes companion pull request created — N/A, no API changes were made

### DIFF
--- a/src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
+++ b/src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
@@ -89,8 +89,7 @@ public class DlsFilterLevelActionHandler {
         boolean applyDlsFilterToHybridQuery
     ) {
 
-        if (threadContext.getHeader(ConfigConstants.OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE) != null
-            || threadContext.getHeader(ConfigConstants.OPENDISTRO_SECURITY_DLS_QUERY_FILTER_APPLIED) != null) {
+        if (threadContext.getHeader(ConfigConstants.OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE) != null) {
             return true;
         }
 
@@ -177,10 +176,8 @@ public class DlsFilterLevelActionHandler {
         try (StoredContext ctx = threadContext.newStoredContext(true)) {
 
             threadContext.putHeader(
-                applyDlsFilterToHybridQuery
-                    ? ConfigConstants.OPENDISTRO_SECURITY_DLS_QUERY_FILTER_APPLIED
-                    : ConfigConstants.OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE,
-                "true"
+                ConfigConstants.OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE,
+                applyDlsFilterToHybridQuery ? ConfigConstants.OPENDISTRO_SECURITY_HYBRID_QUERY_DLS_DONE : "true"
             );
 
             try {
--- a/src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
+++ b/src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
@@ -24,6 +24,7 @@ import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.lucene.search.BooleanClause;
 
 import org.opensearch.OpenSearchSecurityException;
 import org.opensearch.action.ActionRequest;
@@ -50,6 +51,7 @@ import org.opensearch.index.get.GetResult;
 import org.opensearch.index.mapper.MapperService;
 import org.opensearch.index.query.BoolQueryBuilder;
 import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.QueryBuilderVisitor;
 import org.opensearch.index.query.QueryBuilders;
 import org.opensearch.index.query.TermsQueryBuilder;
 import org.opensearch.index.seqno.SequenceNumbers;
@@ -310,6 +312,10 @@ public class DlsFilterLevelActionHandler {
             if (ParentChildrenQueryDetector.hasParentOrChildQuery(query)) {
                 throw new OpenSearchSecurityException("Unable to handle filter level DLS for hybrid queries with parent or child clauses");
             }
+            List<QueryBuilder> originalSubqueries = directSubqueries(query);
+            if (originalSubqueries.isEmpty()) {
+                throw new OpenSearchSecurityException("Hybrid query does not expose subqueries for DLS verification");
+            }
             // Hybrid queries must remain top-level, so apply filter level DLS query directly
             QueryBuilder filteredHybridQuery = query.filter(filterLevelQueryBuilder);
             if (filteredHybridQuery == null) {
@@ -317,6 +323,9 @@ public class DlsFilterLevelActionHandler {
             }
             if (!isHybridQuery(filteredHybridQuery)) {
                 throw new OpenSearchSecurityException("Hybrid query was not preserved after applying the DLS filter");
+            }
+            if (!isDlsFilterAppliedToEverySubquery(filteredHybridQuery, filterLevelQueryBuilder, originalSubqueries)) {
+                throw new OpenSearchSecurityException("Hybrid query did not apply the DLS filter to every subquery");
             }
             searchSource.query(filteredHybridQuery);
         } else {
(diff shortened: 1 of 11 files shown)


src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
```diff
@@ -84,10 +85,12 @@ public static boolean handle(
...
         ClusterService clusterService,
         IndicesService indicesService,
-        ThreadContext threadContext
+        ThreadContext threadContext,
+        boolean applyDlsFilterToHybridQuery
     ) {
 
-        if (threadContext.getHeader(ConfigConstants.OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE) != null) {
+        if (threadContext.getHeader(ConfigConstants.OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE) != null
+            || threadContext.getHeader(ConfigConstants.OPENDISTRO_SECURITY_DLS_QUERY_FILTER_APPLIED) != null) {
```
nibix: Would it be possible to consider the hybrid query handling as an extension of "filter level DLS"? 

I would believe that this would enable you to re-use most of the existing branching logic, i.e., you would not need to add another condition here and at a number of other places.

nibix (commented):
Thank you for this!

I didnt have the time to completely understand the change yet; still, I have two questions/comments:

- At the moment, this adds further logic at the places where we already have logic gates distinguising between already applied filter level DLS and yet to be applied lucene level DLS. Thus, it kind of adds another mode. Could this be considered as an extension of the existing filter level DLS mode? I would hope that then lots of the logic for filter level DLS could be just re-used without too many additions.
- I do not understand the relation to cross cluster search (well, I might have an idea, but it is a bit vague). Could you elaborate on the need to additional cross cluster search handling?

src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
sharp-pixel: Trying to figure out the best option here.

My understanding of DLS is that the is a default mode (adaptive) and 2 explicit modes: lucene_level and filter_level. This would be specified in the node config, so I am assuming the default at the moment which is affecting users of Amazon OpenSearch Service for example.

Effectively, what it does is the following:
An hybrid query needs to be a top-level query. The previous implementation wrapped that in another clause, breaking that requirement. The fix is forcing the filter mode for hybrid queries: hybrid queries support a top-level filter which effectively propagates to the subqueries. This has been functionaly validated with user cases.

However there is a catch, which is that suggestions and aggregations, there needs to be a specific handling, which is why it touches more than I hoped it would.

So the approach can and should be adapted also to the filter_level mode. It is currently only enabled for adaptive. lucene_level is unsupported at this stage.

sharp-pixel: If we make it an extension of the filter level DLS, and the mode is adaptive, we still need to force filter_level on hybrid queries. Happy to discuss a redesign if that makes sense.

cwperks:
@sharp-pixel should we also create a companion PR on neural search repo that executes integration tests for this?

sharp-pixel:
> @sharp-pixel should we also create a companion PR on neural search repo that executes integration tests for this?

https://github.com/opensearch-project/neural-search/pull/1957

src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
sharp-pixel: For CCS, it's mainly a restriction at the moment to enable DLS on local cluster only. Validating all the touch points of CCS would require additional effort.

src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
```diff
@@ -262,6 +281,83 @@ public void onFailure(Exception e) {
...
+    }
+
+    /**
+     * Neural Search is an optional plugin, so Security identifies its hybrid query through the public query type name
+     * instead of depending on its query builder class. {@link QueryBuilder#getName()} is OpenSearch's unique query type
+     * identifier. A query builder registered as {@code hybrid} must honor {@link QueryBuilder#filter(QueryBuilder)} by
+     * applying the supplied filter to every subquery and must expose every subquery through its visitor. Reader-level DLS
+     * remains active whenever this special path is selected, independently of the query builder's filter implementation.
+     */
+    static boolean isHybridQuery(QueryBuilder query) {
```
cwperks, line 370: @sharp-pixel overall I am aligned with this PR, but had a couple ideas to make it more generic:

1. Should we consider introducing anything on the query filter in core as a marker? i.e. https://github.com/cwperks/OpenSearch/pull/385
2. Can we re-use the existing `OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE` header to carry this information about whether its a hybrid query? I was exploring around with that in https://github.com/cwperks/security/pull/114.

sharp-pixel: I agree. I investigated and I may require https://github.com/cwperks/OpenSearch/pull/385 to be merged to be able to use QueryBuilder.supportsTopLevelFilter(). Should we do everything in one go, or fix and improve after?

cwperks: I'm good with this PR if we can incorporate 2) into it. I'd prefer to have a single thread context header

sharp-pixel: Updated to incorporate 2).

src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
nibix: > If we make it an extension of the filter level DLS, and the mode is adaptive, we still need to force filter_level on hybrid queries.
> The fix is forcing the filter mode for hybrid queries

That's actually also the case for term lookup queries inside DLS queries (for which the mode switching was introduced):

Effectively, we can do DLS at the moment using two ways: 

1. on Lucene level (inside the index reader classes)
2. on filter level (by modifying queries)

The `adaptive` (default) mode just checks the requests and based on that chooses whether it then goes for Lucene level or for filter level. Or, in your words, it is forcing the filter mode for TLQ queries (only difference is that these are used on the DLS query side, but effectively that's not a too big difference).

If you set the setting either `filter_level` or `lucene_level` however, this check is not done and just the respective mode is chosen regardless. If `lucene_level` is selected, this has the effect that term lookup queries inside DLS queries just won't work.

To be honest, I doubt that there is anyone who has changed this setting from the default. The switch was rather added as a kind of workaround in case this feature makes trouble.

cwperks: Its actually not even wired properly https://github.com/opensearch-project/security/issues/3794

cwperks:
@sharp-pixel can you separate out hybrid query support and neural query support or does it make sense to address them in the same PR? Personally, I think I would find it easy for review.

FYI Github now has a Stacked PR feature which I haven't tried out, but could be useful here: https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

sharp-pixel:
> @sharp-pixel can you separate out hybrid query support and neural query support or does it make sense to address them in the same PR? Personally, I think I would find it easy for review.
> 
> FYI Github now has a Stacked PR feature which I haven't tried out, but could be useful here: https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

If it's for review only yes I can separate, but I cannot seem to be able to create a stacked PR on the source repo. I could do it on my fork though.

Let me split the PR in  2. I will revert the changes to neural and add them in a new PR.

DarshitChanpura:
Traced the full flow — design fails closed, which is the bar here. Reader-level `liveBits` is the real enforcement, the subquery filter push is strictly verified, and every deviation throws. No fail-open path found.

Before merge:
1. Rebase — branch is conflicting.
2. Green full CI run — unit tests mock `HybridQueryBuilder`, so `HybridQueryDlsIT` is where the real coverage lives.

Two things I checked:
- The hybrid contract matches current neural-search: `HybridQueryBuilder.NAME` is `"hybrid"` and `filter()` delegates per-subquery and returns the hybrid builder — exactly what the verification here expects, and it pairs with #1432. One caveat worth a comment in-code: `V_3_9_0` is a lower bound only, so a future change to neural-search's `filter()`/`visit()` would fail closed but silently break hybrid+DLS. Acceptable, just flagging.
- `DlsFlsLegacyHeaders.prepare(..., doFilterLevelDls && !applyDlsFilterToHybridQuery)` — the two flags are mutually exclusive (TLQ vs. no-TLQ), so the second term is dead. Drop it.

DarshitChanpura (approved):

PR #1957, opened 2022-07-20
Increments minor patch version number in build.gradle

DarshitChanpura:
### Description
In order to generate correct `.deb` and `.rpm` artifacts this version number increment is required.

* Category : Maintenance

### Testing
manual testing

### Check List
- [ ] New functionality includes testing
- [ ] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #6428, opened 2026-08-24
Support neural queries with hybrid DLS

sharp-pixel:
### Description

This is a follow-up to #6416, which preserves a hybrid query as the top-level
query while applying Document-Level Security (DLS) to its execution branches.

The verification introduced by #6416 handles query builders that apply filters
through the standard boolean-wrapper contract. Neural and k-NN queries apply
filters differently:

- `neural` stores the filter internally and returns the same query builder;
- `knn` returns a copied query builder containing the combined filter.

As a result, otherwise valid hybrid requests containing these branches can fail
closed after DLS application.

This PR extends hybrid DLS handling without compile-time dependencies on either
optional plugin and without reflecting on plugin-specific getter names.
Security now relies on the public `QueryBuilder#filter(QueryBuilder)` and
`QueryBuilder#visit(QueryBuilderVisitor)` contract:

- hybrid builders expose their execution branches through `visit()`;
- a branch that stores an embedded filter exposes it as a direct
  `BooleanClause.Occur.FILTER` child;
- `filter()` preserves the original query semantics while adding the supplied
  filter.

After applying DLS, Security verifies that:

- every original hybrid branch is preserved exactly once;
- the filtered hybrid query retains the same query type;
- filter-aware copied branches retain the same runtime query type and registered
  query name;
- the exact DLS query instance is present in each branch's filter path; and
- an existing embedded filter is composed structurally with DLS.

This contract-based approach remains valid for 3.9 and later releases without a
per-release upper-version allowlist. If a query builder does not expose its
branches or embedded filters through `visit()`, changes query type, drops a
branch, or fails to retain DLS, Security fails closed.

The change also:

- preserves standard query metadata (`boost` and query name) when a filter-aware
  query returns a copy;
- preserves implicit `minimum_should_match` behavior when DLS is added to an
  embedded boolean filter;
- rejects parent/child clauses anywhere in the exposed hybrid query tree;
- supports both in-place and copied filter-aware query builders; and
- keeps `neural_sparse` on the generic boolean-wrapper path.

No new setting, permission, REST API, optional-plugin dependency, or
plugin-specific reflection is introduced.

### Companion PR and merge relationship

This PR and
[opensearch-project/neural-search#1957](https://github.com/opensearch-project/neural-search/pull/1957)
are interlocked for dense neural hybrid DLS:

- Security #6428 needs Neural #1957's `NeuralQueryBuilder#visit()` override to
  discover the embedded neural filter. Without that traversal, Security fails
  closed rather than applying DLS incompletely.
- Neural #1957's cross-plugin DLS integration test needs the contract-based
  verification implemented by Security #6428.

This is not a compile-time dependency cycle: both repositories compile and run
their unit tests independently. The practical sequence is to merge and publish
Security #6428, validate Neural #1957 against that Security artifact, and then
merge Neural #1957. Both changes should ship together in OpenSearch 3.9.

### Related work

- Follow-up to #6416.
- Interlocked companion change and cross-plugin coverage:
  [opensearch-project/neural-search#1957](https://github.com/opensearch-project/neural-search/pull/1957).
- Related Neural Search issue:
  [opensearch-project/neural-search#1303](https://github.com/opensearch-project/neural-search/issues/1303).
- Independent dense-neural model-registration blocker: #6433.

### Testing

Focused Security regression tests:

```bash
./gradlew test \
  --tests "org.opensearch.security.configuration.DlsFilterLevelActionHandlerTest" \
  --tests "org.opensearch.security.configuration.DlsFlsValveImplTest" \
  --tests "org.opensearch.security.util.ParentChildrenQueryDetectorTest"
```

Result: passed.

Formatting and static analysis:

```bash
./gradlew spotlessJavaCheck checkstyleMain checkstyleTest
```

Result: passed.

JaCoCo reports 100% line coverage and 91.3% branch coverage across the modified
Security verification methods. The new direct-filter visitor has 100% line and
branch coverage.

The companion `HybridQueryDlsIT` suite was run against the locally published
Security plugin on a three-node Neural Search cluster. Five of six tests pass,
including direct k-NN, existing-filter, single-clause, suggestion, aggregation,
and hit-filtering coverage. The dense-neural test is currently blocked during
ML model registration by #6433, before the neural query executes.

### Check List

- [x] New functionality has been documented — behavior, contract, companion
  dependency, and support boundaries are documented above and in code comments;
  no user-facing setting or API is introduced
  — N/A, no roles or permissions were added
- [x] API changes companion pull request created — N/A, this consumes and
  completes an existing public `QueryBuilder` contract without adding API
  surface

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.


cwperks:
@sharp-pixel can you please fix the conflicts post-merge of hybrid query support?

sharp-pixel:
> @sharp-pixel can you please fix the conflicts post-merge of hybrid query support?

looking

DarshitChanpura:
**Multi-knn hybrid queries fail closed under DLS.** The `knn` fast path in `isDlsFilterAppliedToSubquery` (`DlsFilterLevelActionHandler.java:420-422`) matches purely on `getName()` with no reference anchor. Because the outer loop in `isDlsFilterAppliedToEverySubquery` is order-independent, a single filtered knn subquery matches *every* knn entry in `unmatchedOriginalSubqueries`, tripping the "must not merge multiple original branches" guard.

Repro — originals `[knn(a), knn(b)]`, filtered `[knn(a'), knn(b')]`:
```
unmatched = {knn(a):1, knn(b):1}
filtered knn(a'):
  (knn(a),1): name match -> preserved = knn(a)
  (knn(b),1): name match -> preserved != null -> return false
-> "Hybrid query did not apply the DLS filter to every subquery"
```
Any hybrid with 2+ knn clauses (multi-vector-field search) throws. It fails *closed*, so not a security hole — but it breaks a legitimate workflow, and `acceptsKnnHybridSubqueryWithFilterCopy` only covers the single-knn case so tests don't catch it. `neural`/`neural_sparse` are unaffected (reference-identity match disambiguates).

Verified against neural-search `main`: `HybridQueryBuilder.filter()` and `visit()` both walk the same `ArrayList` in order and replace in place, so subqueries stay 1:1 and positionally aligned. Simplest fix is to pair `filteredSubqueries[i]` with `originalSubqueries[i]` positionally instead of the name-based multiset match; if you want to keep order-independence, consume exactly one original per filtered on the knn path (break after first match) rather than treating a second candidate as a merge. Please add a 2+-knn test either way.

sharp-pixel:
I am reviewing in depth because several things are broken in their current state.

sharp-pixel:
I reworked the implementation

src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
```diff
@@ -76,11 +77,54 @@
...
+        Object.class
+    );
+    private static final Function<QueryBuilder, Object> KNN_RESCORE_CONTEXT_GETTER = ReflectiveAttributeAccessors.objectMethod(
+        "getRescoreContext",
+        Object.class
+    );
+    private static final Function<QueryBuilder, Object> KNN_EXPAND_NESTED_GETTER = ReflectiveAttributeAccessors.objectMethod(
+        "getExpandNested",
+        Object.class
+    );
```
DarshitChanpura: I'm wary of this reflection surface (DlsFilterLevelActionHandler.java:87-127). Reflected bindings to ~10 KNNQueryBuilder/NeuralQueryBuilder getters are fragile: they're concrete method names on optional-plugin internals with no version guard, so a rename or signature change silently disables knn/neural hybrid DLS (fails closed, but the feature just stops working) and unit tests can't catch it since they mock QueryBuilder.

Worth noting applying the filter already uses a core contract - QueryBuilder#filter(QueryBuilder) and #visit(...). The reflection only exists to verify the result. So the clean fix is a core extension point for that verification - e.g. a way to introspect the filter embedded in a leaf builder - so Security can confirm the DLS filter landed without reaching into plugin internals. I'd prefer that over reflection.

If a core change isn't feasible in this PR's timeline, please add a version guard (min k-NN/neural or OpenSearch version where these accessors are known-stable) so drift fails predictably at a known boundary rather than silently.

sharp-pixel: Thanks for raising this. Keeping in mind that we will need to backport the fix, version guard is a bit tricky.
I am removing the reflection and will propose a better alernative.

sharp-pixel: @DarshitChanpura Can you check the revised implementation?

DarshitChanpura:
Cross-repo gap: dense-`neural` depends on neural-search **https://github.com/opensearch-project/neural-search/pull/1957** (still open). Security reads a subquery's embedded filter via `visit(getChildVisitor(Occur.FILTER))`. `KNNQueryBuilder` exposes it that way and `neural_sparse` works via bool-wrap, but `NeuralQueryBuilder.filter()` stores into `queryfilter` / returns `this` with **no `visit()` override** on `main` — that override is in https://github.com/opensearch-project/neural-search/pull/1957. Until it ships, dense-`neural` hybrid+DLS fails closed (as `failsClosedWhenNeuralFilterIsNotExposedByVisitor` intends).

https://github.com/opensearch-project/neural-search/pull/1957 targets 3.9, same as this PR's `V_3_9_0` floor — no version bump needed, but the two must land in the same 3.9 release or dense neural is dead on arrival. Also note in the neural tests that they pass only because `stubFilterAwareQuery` simulates https://github.com/opensearch-project/neural-search/pull/1957 's unreleased `visit()`, so they aren't end-to-end proof for dense neural — the IT in https://github.com/opensearch-project/neural-search/pull/1957 is.

sharp-pixel:
Cross-plugin coordination note: after this Security change lands, hybrid queries containing NeuralQueryBuilder will temporarily fail closed against the current Neural Search implementation because its embedded filter is not yet exposed through QueryBuilderVisitor. This is an availability/compatibility limitation, not a DLS bypass: unauthorized documents are not returned.

opensearch-project/neural-search#1957 adds the required FILTER visitor traversal and cross-plugin integration coverage, enabling dense neural hybrid DLS. The two PRs are intended for a coordinated release, so the Security artifact should not be released without the companion Neural Search change.

DarshitChanpura (approved):

src/main/java/org/opensearch/security/configuration/DlsFilterLevelActionHandler.java
DarshitChanpura: looks good.

PR #6437, opened 2026-08-28
Make REST API request body max string length configurable (fixes #6434)

### LINKED ISSUE #6434: [BUG] 3.8.0 introduces hard limit of 256 character for all text fields
**What is the bug?**
#6224 introduced a hard limit of 256 characters for all text fields in requests. this breaks various use-cases, e.g. using a large(-ish) DLS query in a role.

HTTP requests get rejected with HTTP 400 and the error message "String value exceeds the maximum allowed length of 256 characters"

try to create a role with a large DLS query:
```bash
curl --location --request PUT 'http://localhost:9200/_plugins/_security/api/roles/testrole' \
--header 'Content-Type: application/json' \
--data-raw '{
    "index_permissions": [
        {
            "index_patterns": [
                "testindex"
            ],
            "dls": "{ \"term\": { \"testfield\": \"123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890\" } }",
            "allowed_actions": [
                "indices:data/read/search",
                "indices:data/read/search/template",
                "indices:data/read/get"
            ]
        }
    ]
}'
```

such requests do not get rejected.
at the very least the limit should be configurable.

opensearch 3.8.0

n/a

n/a

DarshitChanpura:
### Description

PR #6224 added a blanket recursive check (`RequestContentValidator#validateStringLength`) that rejects **any** string value longer than a hard-coded **256** characters *anywhere* in a Security REST API request body. While the intent was hardening against oversized inputs, 256 chars is far too small for legitimate free-form fields — most notably **DLS queries**, but also FLS/masked-field definitions, `description`, query templates, etc. On upgrade to 3.8.0.0 these requests started failing with:

```
HTTP 400 - String value exceeds the maximum allowed length of 256 characters
```

This is a breaking change for any cluster that manages roles with non-trivial DLS/FLS via the REST API (see #6434).

This PR:

- Introduces a configurable node setting **`plugins.security.restapi.max_string_length`** that controls the blanket body-wide limit.
- Raises the default from 256 to **4096**, restoring compatibility for realistic payloads while still bounding pathologically large inputs. (Jackson's own `StreamReadConstraints` remains the hard parser-level ceiling.)
- Leaves the **targeted** 256-char validators for semantically-bounded fields (entity names, principals, paths) unchanged — those limits are reasonable and are not the source of the regression.
- Makes the error response report the **effective configured limit** instead of a hard-coded 256.

Operators who want a stricter (or looser) cap can now set the value in `opensearch.yml`, e.g.:

```yaml
plugins.security.restapi.max_string_length: 8192
```

### Issues Resolved

Resolves #6434

### Testing

- Reworked the existing `validateStringLength` unit tests to drive the limit through the new setting and to assert the dynamic (limit-aware) error message.
- Added a regression test that a large free-form value is accepted under the default limit, and a test that a value beyond the default is rejected and reports the default limit.
- `./gradlew spotlessCheck checkstyleMain checkstyleTest test --tests "org.opensearch.security.dlic.rest.validation.RequestContentValidatorTest" --tests "org.opensearch.security.dlic.rest.validation.EndpointValidatorTest"` → all green, 102 tests, 0 failures.

### Backport

This should be backported to the `3.8` branch for a **3.8.0.1** patch release, since 3.8.0.0 is affected.

### Check List
- [x] New functionality has been documented (setting will be added to the documentation-website in a follow-up)

### DIFF
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -95,6 +95,7 @@ import org.opensearch.common.settings.SettingsFilter;
 import org.opensearch.common.util.BigArrays;
 import org.opensearch.common.util.PageCacheRecycler;
 import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.common.xcontent.XContentConstraints;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.action.ActionResponse;
 import org.opensearch.core.common.io.stream.NamedWriteableRegistry;
@@ -2587,6 +2588,7 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
                     ConfigConstants.SECURITY_RESTAPI_MAX_STRING_LENGTH,
                     ConfigConstants.SECURITY_RESTAPI_MAX_STRING_LENGTH_DEFAULT,
                     1,
+                    XContentConstraints.DEFAULT_MAX_STRING_LEN,
                     Property.NodeScope,
                     Property.Filtered
                 )
--- a/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
+++ b/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
@@ -2582,6 +2582,15 @@ public final class OpenSearchSecurityPlugin extends OpenSearchSecuritySSLPlugin
                     Property.Filtered
                 )
             );
+            settings.add(
+                Setting.intSetting(
+                    ConfigConstants.SECURITY_RESTAPI_MAX_STRING_LENGTH,
+                    ConfigConstants.SECURITY_RESTAPI_MAX_STRING_LENGTH_DEFAULT,
+                    1,
+                    Property.NodeScope,
+                    Property.Filtered
+                )
+            );
 
             // Compliance settings moved outside the gate — see below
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
```diff
@@ -2582,6 +2582,15 @@ public List<Setting<?>> getSettings() {
                     Property.Filtered
                 )
             );
+            settings.add(
```
reta, line 2586: May be worth aligning with OpenSearch XContent constraints (max): 
```
     settings.add(
                Setting.intSetting(
                    ConfigConstants.SECURITY_RESTAPI_MAX_STRING_LENGTH,
                    ConfigConstants.SECURITY_RESTAPI_MAX_STRING_LENGTH_DEFAULT,
                    1,
                    XContentConstraints.DEFAULT_MAX_STRING_LEN, 
                    Property.NodeScope,
                    Property.Filtered
                )
            );
```

DarshitChanpura:
The Checkstyle scan failure is pre-existing on main, not from this change: the `licenseHeaders` precommit task flags `sample-resource-plugin/src/main/java/org/opensearch/sample/SampleResourceGroupExtension.java`, which is missing its license header. It's resolved by https://github.com/opensearch-project/security/pull/6435 (adds the missing headers); I'll re-merge main once that lands.

willyborankin (approved):

src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java
DarshitChanpura: done.

reta (approved):

DarshitChanpura:
#6435 is a merge blocker.

PR #6451, opened 2026-09-01
Keep hybrid DLS safe across mixed versions

sharp-pixel:
## Summary

- Replace the hybrid DLS sentinel value in the filter-level completion header with a dedicated header.
~~- Continue accepting the legacy sentinel from unpatched nodes during rolling upgrades.~~
- Remove the cluster-wide version gate, allowing safe backports and mixed-version rollout behavior.
- Preserve reader-level DLS and strip both marker formats before dispatch to non-local clusters.

## Security review

No security finding under the mixed patched-node and unpatched-node cluster assumption.

- Untrusted REST callers cannot forge either internal header: security-prefixed headers are rejected at ingress.
- Unpatched coordinator to patched data node: the patched node recognizes the legacy sentinel as hybrid-query-only completion, does not treat it as full filter-level completion, and retains reader-level DLS.
- Patched coordinator to unpatched data node: the unpatched node ignores the unfamiliar dedicated marker and performs normal DLS processing. This can repeat DLS work but is fail-closed.
- Patched nodes preserve both marker formats during local fan-out. If an unpatched intermediate node drops the unfamiliar dedicated marker, downstream nodes fall back to normal DLS rather than bypassing it.
- Cross-cluster dispatch strips both marker formats on patched nodes. Unpatched nodes already strip the legacy sentinel and do not copy the unfamiliar dedicated header.
- Only the exact legacy sentinel is reclassified; the ordinary filter-level completion value continues to suppress reader-level DLS as before.

### DIFF
--- a/src/main/java/org/opensearch/security/support/ConfigConstants.java
+++ b/src/main/java/org/opensearch/security/support/ConfigConstants.java
@@ -72,6 +72,8 @@ public class ConfigConstants {
     public static final String OPENDISTRO_SECURITY_DOC_ALLOWLIST_TRANSIENT = OPENDISTRO_SECURITY_CONFIG_PREFIX + "doc_allowlist_t";
 
     public static final String OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE = OPENDISTRO_SECURITY_CONFIG_PREFIX + "filter_level_dls_done";
+    /** Legacy wire value accepted from unpatched nodes during rolling upgrades. Patched nodes must not write it. */
+    public static final String OPENDISTRO_SECURITY_HYBRID_QUERY_DLS_DONE = OPENDISTRO_SECURITY_CONFIG_PREFIX + "hybrid_query";
     public static final String OPENDISTRO_SECURITY_DLS_QUERY_FILTER_APPLIED = OPENDISTRO_SECURITY_CONFIG_PREFIX
         + "dls_query_filter_applied";
     public static final String OPENDISTRO_SECURITY_CONTAIN_PARENT_CHILD_QUERY = OPENDISTRO_SECURITY_CONFIG_PREFIX + "is_parent_child_query";
--- a/src/main/java/org/opensearch/security/support/ConfigConstants.java
+++ b/src/main/java/org/opensearch/security/support/ConfigConstants.java
@@ -72,8 +72,6 @@ public class ConfigConstants {
     public static final String OPENDISTRO_SECURITY_DOC_ALLOWLIST_TRANSIENT = OPENDISTRO_SECURITY_CONFIG_PREFIX + "doc_allowlist_t";
 
     public static final String OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE = OPENDISTRO_SECURITY_CONFIG_PREFIX + "filter_level_dls_done";
-    /** Legacy wire value accepted from unpatched nodes during rolling upgrades. Patched nodes must not write it. */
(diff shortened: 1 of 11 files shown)


cwperks:
> Continue accepting the legacy sentinel from unpatched nodes during rolling upgrades.

We don't need to do this if this code hasn't been released yet

sharp-pixel:
> > Continue accepting the legacy sentinel from unpatched nodes during rolling upgrades.
> 
> We don't need to do this if this code hasn't been released yet

I think we do to support mixed mode because the coordinator can either be an unpatched or a patched node.

Here is the analysis of GPT5.6 Daybreak Blue:
In a mixed patched/unpatched cluster, one direction is unsafe:
| Coordinator | Data node | Result |
| ----------- | --------- | ------ |
| Patched | Unpatched | Fail-closed. The old node ignores the new dedicated header and performs normal DLS again. It may duplicate filtering or fail the query, but should not expose documents. |
| Unpatched | Patched | Potential fail-open. The old coordinator sends the legacy hybrid sentinel in FILTER_LEVEL_DLS_DONE. The patched data node treats any value in that header as “DLS fully complete,” disabling reader-level DLS. Hits remain filtered, but global aggregations and other reader-level paths could expose unauthorized data. |

cwperks:
@sharp-pixel that would matter if your last PR (https://github.com/opensearch-project/security/pull/6416) was in a released version, but its only in mainline and unreleased. Given that, bwc checks like this block:

```
String filterLevelDlsDone = threadContext.getHeader(ConfigConstants.OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE);
return filterLevelDlsDone != null && !ConfigConstants.OPENDISTRO_SECURITY_HYBRID_QUERY_DLS_DONE.equals(filterLevelDlsDone);
```

Can be simplified to:

```
return threadContext.getHeader(ConfigConstants.OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE) != null;
```

since we don't actually check the value in that header, only that its not-null (or having the value of "true").

src/main/java/org/opensearch/security/support/ConfigConstants.java
```diff
@@ -72,7 +72,8 @@ public class ConfigConstants {
     public static final String OPENDISTRO_SECURITY_DOC_ALLOWLIST_TRANSIENT = OPENDISTRO_SECURITY_CONFIG_PREFIX + "doc_allowlist_t";
 
     public static final String OPENDISTRO_SECURITY_FILTER_LEVEL_DLS_DONE = OPENDISTRO_SECURITY_CONFIG_PREFIX + "filter_level_dls_done";
-    public static final String OPENDISTRO_SECURITY_HYBRID_QUERY_DLS_DONE = OPENDISTRO_SECURITY_CONFIG_PREFIX + "hybrid_query";
+    public static final String OPENDISTRO_SECURITY_DLS_QUERY_FILTER_APPLIED = OPENDISTRO_SECURITY_CONFIG_PREFIX
```
cwperks, line 75: nit: I think it would be useful to have more javadoc around these constants for context of someone newly looking at the repo.

cwperks (approved):

DarshitChanpura (approved):

src/main/java/org/opensearch/security/support/ConfigConstants.java
DarshitChanpura: I agree with this. Resolving this to merge and unblock this PR. @sharp-pixel can you address this in #6428?

PR #6454, opened 2026-09-01
Harden migrate tests for single-provider (no typeField) resources

DarshitChanpura:
### Description

Adds regression coverage for the resource-sharing migrate path used by plugins that register a **single `ResourceProvider` with no `typeField`** (for example, notifications' `notification_config`). This behavior currently has no unit coverage; the tests lock it so a future change to `classifyDocType` or the request-path validation can't silently break migration for these plugins.

Two cases:
- **`classifyNotificationConfigFallsBackToSingleAccessLevelKey`** — with no type paths, `classifyDocType` falls back to the sole `default_access_level` key, so a `notification_config` doc is classified (and migrated) rather than recorded under `skippedResources`. Uses a faithful `.opensearch-notifications-config` `_source` (`NotificationConfigDoc.toXContent` shape with the real `NotificationConstants` tag names).
- **`nonResolvingUsernamePathYieldsNoOwnerSoMigrationUsesDefaultOwner`** — `username_path` must be a valid, non-empty path (an empty `""` is rejected by `PATH_VALIDATOR#requireNonEmpty`), and a valid non-resolving path yields no owner, so migration attributes ownership to `default_owner` while `backend_roles_path` (`/metadata/access`) still supplies the shared backend roles.

Test-only; no production change. The live write-path counterpart is already covered by `MigrateApiTests` (a resource with no owner migrates into `resourcesWithDefaultOwner`).

### Testing

`./gradlew test --tests "org.opensearch.security.resources.api.migrate.MigrateResourceSharingInfoApiActionTests"` — 10 passed, 0 failed.

### Check List
- [x] New functionality includes testing.
- [x] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/src/test/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiActionTests.java
+++ b/src/test/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiActionTests.java
@@ -16,6 +16,7 @@ import java.util.Map;
 import org.junit.Before;
 import org.junit.Test;
 
+import org.opensearch.security.dlic.rest.validation.RequestContentValidator;
 import org.opensearch.security.resources.ResourcePluginInfo;
 
 import tools.jackson.databind.JsonNode;
@@ -23,6 +24,7 @@ import tools.jackson.databind.ObjectMapper;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThrows;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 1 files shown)


src/test/java/org/opensearch/security/resources/api/migrate/MigrateResourceSharingInfoApiActionTests.java
```diff
@@ -143,4 +145,55 @@ public void jsonPointerPreservesLeadingSlash() {
         assertEquals("/monitor/user/name", MigrateResourceSharingInfoApiAction.jsonPointer("/monitor/user/name"));
     }
 
+    @Test
+    public void classifyNotificationConfigFallsBackToSingleAccessLevelKey() throws Exception {
+        // A single-provider index with no typeField (e.g. notifications' notification_config) is
+        // classified by the sole default_access_level key, so the doc is migrated rather than skipped.
+        // Uses a faithful `.opensearch-notifications-config` _source (NotificationConfigDoc.toXContent:
+        // metadata + config, no envelope) with the real NotificationConstants tag names.
+        JsonNode notifDoc = mapper.readTree(
```
cwperks, line 154: nit: prefer to use Java TextBlocks

cwperks (approved):

PR #6457, opened 2026-09-02
[3.8] Unblock 3.8 CI: pin jspecify + guard Eclipse JDT formatter provisioning

DarshitChanpura:
### Description

Two `main`-only CI fixes never reached `3.8`, and together they deadlock every `3.8` PR (dependency resolution fails, and the Spotless scan times out fetching the formatter), so neither could be validated in isolation. This PR carries both:

**1. Pin jspecify (resolves `1.0.1` vs `1.0.0` conflict).** `io.projectreactor.netty:reactor-netty-core:1.3.7` pulls `org.jspecify:jspecify:1.0.1` while `com.google.guava:33.6.0-jre` and OpenSearch core pull `1.0.0`; the OpenSearch build plugin's `failOnVersionConflict` fails the build. `main` forces jspecify (added with the guava bump in `2314b027d`); this adds the catalog entry + `force libs.jspecify` on `3.8`. Verified: `integrationTestCompileClasspath` now resolves `1.0.0 -> 1.0.1 (forced)`.

**2. Guard Eclipse JDT formatter provisioning (backport of #6438).** Spotless provisions the JDT formatter from a P2 mirror during Gradle's configuration phase for every task, intermittently failing with `SocketTimeoutException`. Wire the Eclipse step only when a spotless task runs. Cherry-pick of `024622ffb`.

Supersedes the standalone backports #6458 and #6459 (folded here so a single PR has both fixes and can go green).

### Check List
- [x] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -498,6 +498,7 @@ configurations {
             force "io.netty:netty-transport-native-unix-common:${versions.netty}"
             force "com.github.luben:zstd-jni:${versions.zstd}"
             force libs.google.guava
+            force libs.jspecify
             force libs.lz4.java
             force libs.snappy.java
 
--- a/gradle/formatting.gradle
+++ b/gradle/formatting.gradle
@@ -1,5 +1,12 @@
 allprojects {
 	project.apply plugin: "com.diffplug.spotless"
+	// The Eclipse JDT formatter is provisioned by Spotless from a P2 update-site mirror during Gradle's
+	// configuration phase. That means every Gradle invocation - including test-only tasks that never format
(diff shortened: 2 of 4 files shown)


cwperks (approved):
@DarshitChanpura can we apply the Eclipse JDT format optimization to main brain as well? If it works well then @peterzhuamazon may want to expand this across all repos.

DarshitChanpura:
> @DarshitChanpura can we apply the Eclipse JDT format optimization to main brain as well? If it works well then @peterzhuamazon may want to expand this across all repos.

Yes that is applied via #6438 .

PR #6469, opened 2026-09-04
Support proxy_auth_domain on gRPC auxiliary transport

### LINKED ISSUE #6439: [FEATURE] proxy_auth_domain support on gRPC auxiliary transport
`proxy_auth_domain` is a first-class REST auth type but is silently skipped on the gRPC transport. `BackendRegistry` gates the gRPC path on a hard-coded allowlist: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/BackendRegistry.java#L89

```java
private static final Set<String> GRPC_SUPPORTED_AUTH = Set.of("jwt", "basic");
```   

We hit this while building authenticated gRPC access to OpenSearch behind an Envoy ext_authz sidecar. Envoy has already authenticated the caller and set an identity header, but OpenSearch's gRPC listener won't trust it. That forces one of two workarounds:

Validate the caller's JWT a second time inside OpenSearch (double crypto per request, forces the proxy to expose a JWT).
Run a per-pod sidecar that re-mints tokens with a locally-signed keypair just so OS has something to validate on the gRPC listener (adds a keypair, a local JWKS endpoint, and refresh handling).
Both are workarounds for a header-trust primitive that already exists on REST via proxy_auth_domain.

Add "proxy" to GRPC_SUPPORTED_AUTH. 

BackendRegistry.authenticate() already calls XFFResolver.resolve() unconditionally for both transports, so RemoteIpDetector sets OPENDISTRO_SECURITY_XFF_DONE on the thread context when the caller IP matches the configured internalProxies allowlist — exactly the same trust model REST proxy auth uses today.

GrpcRequestChannel.getRemoteAddress() returns the client InetSocketAddress via Grpc.TRANSPORT_ATTR_REMOTE_ADDR.
GrpcRequestChannel.getHeaders() returns headers in a case-insensitive map so HTTPProxyAuthenticator's user_header / roles_header lookups just work.

Verified locally against `main`: one-line change to `GRPC_SUPPORTED_AUTH`
 Added unit tests,  a full integration test on `SecureNetty4GrpcServerTransport`,  TLS
that exercises the SecurityGrpcFilter → BackendRegistry → HTTPProxyAuthenticator
→ roles_mapping chain with a real DocumentService.Bulk call. 

Also confirmed end-to-end against a stock `opensearchproject/opensearch:3.6.0` container by
overlaying the patched jar and hitting it with `grpcurl` — patched jar succeeds,
original jar returns `Authentication finally failed` on the same request.

Happy to send this as an away-team PR.

- JWT on gRPC (#5812, #5916): supported, but forces the ext_authz proxy to re-mint tokens with a locally-signed keypair, adds a second crypto validation per request and JWKS refresh logic to the sidecar.
- Basic auth on gRPC (#6005): supported, but not appropriate when the upstream proxy has already authenticated the caller — moves credentials over the wire per request.

Parent tracking issue: #5813. Would appreciate a checkbox added there once triaged.

abhinavp105:
Closes #6439

`BackendRegistry.GRPC_SUPPORTED_AUTH` gated the gRPC auth path on `{jwt, basic}`, silently dropping any configured `proxy_auth_domain`. This adds `"proxy"` to that set so `HTTPProxyAuthenticator` is included in the gRPC auth chain — enabling Envoy `ext_authz`-style header-based identity (`x-proxy-user` / `x-proxy-roles`) to be trusted on gRPC the same way it is on REST today.

No other wiring changes were needed:

- `BackendRegistry.authenticate()` already calls `XFFResolver.resolve()` unconditionally for both REST and gRPC, so `RemoteIpDetector` sets `OPENDISTRO_SECURITY_XFF_DONE` on the thread context when the caller IP matches the configured internal-proxies allowlist.
- `GrpcRequestChannel.getRemoteAddress()` returns the client `InetSocketAddress` from `Grpc.TRANSPORT_ATTR_REMOTE_ADDR`.
- `GrpcRequestChannel.getHeaders()` exposes ASCII headers case-insensitively so `user_header` / `roles_header` lookups work.

### Tests

- **`BackendRegistryGrpcAuthTest`** — added a success case (`testGrpcAuthenticateWithValidProxyAuthAndConfiguredDomain`) and a regression guard (`testGrpcProxyAuthDomainIsNotSkipped`) that fails if `"proxy"` is dropped from `GRPC_SUPPORTED_AUTH`.
- **`ProxyAuthGrpcTest`** (new) — full integration test on `SINGLENODE` + `SecureNetty4GrpcServerTransport` + TLS, exercising `SecurityGrpcFilter` → `BackendRegistry` → `HTTPProxyAuthenticator` → `roles_mapping` with a real `DocumentService.Bulk` call. Positive case, plus negatives for missing `x-forwarded-for` and missing user header.

Also updates the stale `"JWT only"` comment inside `authenticate()`.

### Check List
- [x] New functionality has been documented in the code (Javadoc / comments where relevant)


src/integrationTest/java/org/opensearch/security/grpc/ProxyAuthGrpcTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+    public static final TestSecurityConfig.AuthcDomain BASIC_FALLBACK = new TestSecurityConfig.AuthcDomain("basic", 0, true)
+        .httpAuthenticator("basic")
+        .backend("internal");
+
+    @ClassRule
+    public static final LocalCluster cluster = new LocalCluster.Builder().clusterManager(ClusterManager.SINGLENODE)
+        .certificates(TEST_CERTIFICATES)
+        .nodeSettings(SINGLE_NODE_SECURE_AUTH_GRPC_TRANSPORT_SETTINGS)
+        .xff(new XffConfig(true).internalProxiesRegexp("127\\.0\\.0\\.1"))
```
finnegancarroll, line 83: Can you add a second IT suite where 127.0.0.1 is not a trusted client IP to assert we correctly detect and reject proxy auth attempts from this source?

finnegancarroll:
Thanks for adding this support @abhinavp105 !
Added a minor comment on testing the negative case but otherwise LGTM.

finnegancarroll (changes requested):

src/integrationTest/java/org/opensearch/security/grpc/ProxyAuthGrpcTest.java
abhinavp105: Thanks @finnegancarroll for reviewing the change.
I have added a new test to address the comment. Can you please have a look again

finnegancarroll (approved):

PR #6506, opened 2026-09-14
Fix FIPS short Basic Auth password crash → 401 (#6504)

### LINKED ISSUE #6504: [BUG] Uncaught FipsUnapprovedOperationError in PBKDF2PasswordHasher causes JVM crash on short Basic Auth passwords
### Description
When OpenSearch runs in FIPS mode using the Bouncy Castle FIPS provider (`BC-FJA`), submitting an HTTP Basic Authentication request with a password shorter than 112 bits (14 characters) causes the OpenSearch Java process to crash immediately.

Because `org.bouncycastle.crypto.fips.FipsUnapprovedOperationError` extends `java.lang.Error` (not `java.lang.Exception`), it bypasses the `catch (Exception e)` handlers in `SecurityRestFilter` and `BackendRegistry`. The uncaught `Error` bubbles up to the Netty transport layer, tearing down the worker thread and terminating the JVM container.

### Steps to Reproduce
1. Deploy OpenSearch with FIPS mode enabled (using Bouncy Castle FIPS provider).
2. Submit an HTTP Basic Authentication request containing a password under 14 characters:
   ```bash
   curl -sk -u "admin:test123" https://localhost:9200/_cluster/health


### Expected Behavior
The security plugin should handle FIPS bit-length validation failures gracefully and return an HTTP 401 Unauthorized status.

### Actual Behavior
An uncaught FipsUnapprovedOperationError crashes the Netty HTTP transport pipeline and terminates the JVM process.

SatvikMishra08:
## Summary
Fixes #6504

When OpenSearch runs with Bouncy Castle FIPS, short Basic Auth passwords
(< 14 chars / 112 bits) cause `FipsUnapprovedOperationError` (extends
`Error`). That bypassed `catch (Exception)` and crashed the JVM.

This change:
- Adds `FipsErrors` to detect FIPS unapproved-operation Errors (FQCN or simple name)
- Converts them in `PBKDF2PasswordHasher` to `OpenSearchSecurityException`
- Treats them as auth failure in `BackendRegistry.authcz` / `SecurityRestFilter` → HTTP 401
- Does **not** swallow non-FIPS `Error`s (OOM, StackOverflow, …)

DCO: commits are Signed-off-by (git commit -s).

## Test plan
- [x] Unit: `./gradlew test --tests org.opensearch.security.hasher.FipsErrorsTests`
- [ ] Manual (FIPS JVM): `curl -sk -u admin:test123 https://localhost:9200/_cluster/health` → 401, process stays up


src/main/java/org/opensearch/security/auth/BackendRegistry.java
```diff
@@ -799,6 +800,14 @@ public User call() throws Exception {
                 log.debug("Can not authenticate {} due to exception", ac.getUsername(), e);
             }
             return null;
+        } catch (Error e) {
```
reta, line 803: Thanks @SatvikMishra08 , I am wondering we are adding this catch statements as precautionary measure or there are other flows (besides password hashing) that cause  `FipsUnapprovedOperationError`? Thank you

SatvikMishra08: Thanks for the question — not a blanket precaution.

The only concrete producer we know of is the FIPS PBKDF2 path: short Basic Auth passwords make Bouncy Castle throw `FipsUnapprovedOperationError` (extends `Error`, so it skips `catch (Exception)`). That Error escapes from hashing into the Basic Auth call stack (`BackendRegistry.authcz` → `SecurityRestFilter`), which is what was tearing down the Netty worker.

The extra catches are only there so that same FIPS `Error` is treated as an auth failure (401 / null auth) instead of crashing the JVM. Non-FIPS `Error`s (e.g. OOM) are still rethrown — we are not swallowing arbitrary Errors, and we are not claiming other independent code paths throw `FipsUnapprovedOperationError`.

src/main/java/org/opensearch/security/hasher/FipsErrors.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Converts a FIPS unapproved-operation {@link Error} into an
+     * {@link OpenSearchSecurityException}; otherwise rethrows the Error so
+     * fatal JVM conditions (OOM, StackOverflow, etc.) are not swallowed.
+     */
+    public static OpenSearchSecurityException asSecurityExceptionOrPropagate(Error e) {
+        if (isFipsUnapprovedOperationError(e)) {
+            return new OpenSearchSecurityException("password rejected by FIPS policy", e);
```
reta, line 65: I would think of other (non-password only) flows that may occur, may be we could add message as an argument:

```
OpenSearchSecurityException asSecurityExceptionOrPropagate(Error e, String message)
```

or make a method explicitly password specific: `asRejectedPasswordSecurityExceptionOrPropagate`

src/main/java/org/opensearch/security/filter/SecurityRestFilter.java
```diff
@@ -373,15 +374,27 @@ public void checkAndAuthenticateRequest(SecurityRequestChannel requestChannel) t
...
+                    org.apache.logging.log4j.ThreadContext.put(
+                        "user",
+                        ((User) threadContext.getTransient(ConfigConstants.OPENDISTRO_SECURITY_USER)).getName()
+                    );
+                }
+            } catch (Error e) {
+                if (FipsErrors.isFipsUnapprovedOperationError(e)) {
+                    log.warn("Authentication failed due to FIPS policy", e);
+                    org.apache.logging.log4j.ThreadContext.remove("user");
+                    requestChannel.queueForSending(
```
reta, line 392: @cwperks do you think it worth adding `AuditLog::logsFipsError` API to capture such issues (as we do we SSL, etc), could be separate pull request though, thanks

src/test/java/org/opensearch/security/hasher/FipsUnapprovedOperationError.java
```diff
@@ -0,0 +1,33 @@
...
+
+package org.opensearch.security.hasher;
+
+/**
+ * Test double for {@code org.bouncycastle.crypto.fips.FipsUnapprovedOperationError}.
+ * Extends {@link Error} (not {@link Exception}) so it mirrors Bouncy Castle FIPS
+ * provider behavior without requiring the FIPS JAR on the unit-test classpath.
+ * Detection keys off the simple class name {@code FipsUnapprovedOperationError}.
+ */
+public class FipsUnapprovedOperationError extends Error {
```
reta, line 20: I think you should be able to use `org.bouncycastle.crypto.fips.FipsUnapprovedOperationError` directly, all FIPS related dependencies are  available in all test scopes

PR #6521, opened 2026-09-17
[Backport 3.8] fix(audit): dedup bulk audit events by only iterating BulkShardRequest

DarshitChanpura:
### Description

Manual backport of https://github.com/opensearch-project/security/pull/6390 to 3.8. It merged to main on 14 Aug, after the 3.8.0.0 tag was cut on 3 Aug, and was never labelled for backport, so the released 3.8.0 artifact does not have it.

Without it, `AuditActionFilter`'s audit index guard misses the audit sink's own writes. The guard checks `IndicesRequest`, but a single-document write to the audit index arrives as a `BulkRequest` via `TransportSingleItemBulkWriteAction`, and `BulkRequest` does not implement `IndicesRequest`. The write gets audited, which produces another write, which gets audited.

Measured on the released `opensearchproject/opensearch:3.8.0` image with `plugins.security.disabled=true`, `disabled_categories=TRANSPORT_AUDIT` and no client attached: 145% CPU on an idle single node and 34,033 audit documents in three minutes, 33,975 of them `BulkRequest` events targeting `security-auditlog-2026.09.17` itself. On 3.7.0, which predates standalone audit logging, the index is never created and the node sits at 0.59%.

This is the REQUEST_AUDIT half of https://github.com/opensearch-project/OpenSearch/issues/22815. The TRANSPORT_AUDIT half is https://github.com/opensearch-project/security/pull/6520, which is labelled for 3.8; both are needed, since each sustains the loop on its own.

Cherry-picked clean. The only difference from the post-image on main is the absence of the `isBodyExcluded` calls from https://github.com/opensearch-project/security/pull/6347, which is not on 3.8.

### Testing

`AuditActionFilterTest` 32/32 green on this branch, including the 330 lines of tests the original PR added.

### Check List


DarshitChanpura:
Closing — there is no 3.8 patch release planned, and OpenSearch only cuts one for a previous minor on a critical CVE. This does not meet that bar.

3.9.0 is unreleased and already carries this change on both main and 3.9, so the fix reaches users there. For anyone on 3.8.0 in the meantime, the workarounds are on https://github.com/opensearch-project/OpenSearch/issues/22815: use `DISABLE_SECURITY_PLUGIN=true` rather than `plugins.security.disabled=true`, or unset `plugins.security.audit.type`.

